PR #1, opened 2016-04-21
C# codegen for retry

chrisdunelm:


vgen/src/main/java/io/gapi/vgen/csharp/CSharpContextCommon.java
```diff
@@ -92,18 +100,21 @@ public String underscoresToCamelCase(
   // This function is necessary to provide a static entry point for the same-named
   // member function.
   public static String s_underscoresToCamelCase(
-      String input, boolean capNextLetter, boolean preservePeriod) {
+      String input, boolean capNextLetter, boolean preservePeriod, boolean forceLower) {
```
jskeet: What does forceLower mean in this case? Basically *only* use _ to create upper-case characters? Any reason the caller can't just lower-case the input first?

Is this method still a port of csharp_helpers.cc (i.e. has the function changed there)? We should probably be clear about which methods have to maintain compatibility with protoc and which are for other purposes.
(I would suggest we should extract them to a separate class, too... and add plenty of unit tests.)

vgen/src/main/java/io/gapi/vgen/csharp/CSharpContextCommon.java
```diff
@@ -83,7 +83,15 @@ public String addImport(String namespace) {
   // the functionality, since snippets can't call static functions.
   public String underscoresToCamelCase(
       String input, boolean capNextLetter, boolean preservePeriod) {
-    return s_underscoresToCamelCase(input, capNextLetter, preservePeriod);
+    return s_underscoresToCamelCase(input, capNextLetter, preservePeriod, false);
+  }
+
+  public static String s_underscoresToCamelCase(String input) {
```
jskeet: What's with the s_ prefix here? (Not sure why it's in the method later, too...)

vgen/src/main/java/io/gapi/vgen/csharp/CSharpContextCommon.java
chrisdunelm: Good point, yes I'll change this to lower-case the input first, and revert the change to this function. Not currently sure about the state of its portyness.
And I agree about unit tests for functions like this. ...in a follow-up PR, etc, etc...

vgen/src/main/java/io/gapi/vgen/csharp/CSharpContextCommon.java
chrisdunelm: I suspect this is due to the Java-Formatter tool that has been run over all code in this repo.

pongad: I can provide some context. This was done when the big Dicovery merge happened. The motivation is that it is not possible to call static methods from the snippet file, but static methods are easier to use when programming in plain Java (assuming, of course, that it can be made static). So, the s_method is the static versions, the instance methods simply delegate the call.

vgen/src/main/java/io/gapi/vgen/csharp/CSharpContextCommon.java
chrisdunelm: Done.

chrisdunelm:
PTAL. This has now been Java-7-ised.

vgen/src/main/java/io/gapi/vgen/csharp/CSharpGapicContext.java
```diff
@@ -143,6 +151,94 @@ public ServiceInfo getServiceInfo(Interface service) {
   }
 
   @AutoValue
+  public static abstract class RetryDefInfo {
```
jskeet: Some comments would be helpful :)

jskeet:
LGTM

PR #2, opened 2016-04-21
Updated java main.snip

michaelbausor:
Removed throws from ApiCallable methods
Disable url encoding of template variables in Path Template


garrettjonesgoogle:
LGTM

PR #5, opened 2016-04-21
Support media-upload option

pongad:
Previously we have missed media-upload when parsing Discovery Doc.
Since Java API Client Libraries provide overloads for methods with
media-upload, one with the media content as a parameter and one without,
compilecheck never complained.

This change detects media-upload option in Discovery Doc, but
leaves the code samples unchanged.


garrettjonesgoogle:
LGTM

PR #7, opened 2016-04-21
Remove jdk7 from travis

jmuk:
Sorry, these jdks cause errors actually.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,5 +2,3 @@ sudo: false
 language: java
 jdk:
   - oraclejdk8
-  - oraclejdk7
-  - openjdk7


garrettjonesgoogle:
The problem isn't the JDKs, the problem is that we have Java 8 code. We actually want to support Java 7.

jmuk:
So should we fix the compile errors of Java 7 instead?

garrettjonesgoogle:
I think we should.

jmuk:
Ok, will abandon this PR.

tbetbetbe:
I've just raised #8 to track this.
If either of you want to jump and fix it before @chrisdunelm gets in tomorrow, I'm sure he'd appreciate it ;)

PR #9, opened 2016-04-22
Make vgen Java 7 compatible, and enforce this in gradle

chrisdunelm:


chrisdunelm:
Travis still fails, but it now fails in the same way for all three JDKs (oraclejdk7, oraclejdk8, openjdk7).
The failure is now not caused by Java 8 code being present.

vgen/src/main/java/io/gapi/vgen/csharp/CSharpGapicContext.java
```diff
@@ -378,25 +380,27 @@ public static PathTemplateInfo create(
...
-            })
-        .collect(Collectors.toList());
+    return FluentIterable.from(interfaceConfig.getCollectionConfigs())
+        .transform(new Function<CollectionConfig, PathTemplateInfo>() {
+          @Override public PathTemplateInfo apply(CollectionConfig collection) {
+            PathTemplate template = collection.getNameTemplate();
+            Set<String> vars = template.vars();
+            StringBuilder varArgDeclList = new StringBuilder();
+            StringBuilder varArgUseList = new StringBuilder();
+            for (String var : vars) {
```
jskeet: I suspect that Joiners would make this simpler, but I haven't checked for sure.

jskeet:
Would prefer not to have to do this, but code looks fine if we do. The suggestion around Joiners may or may not be worth investigating...

As this doesn't fix Travis yet, let's hold off for now - but consider this LGTM'd when we *do* want to.

tbetbetbe:
FYI: I re-ran the travis build after #11 was submitted, this builds OK now.
Given that jskeet@ LGTMed this, I will merge now to allow other builds to go green

PR #15, opened 2016-04-22
Updated Settings builder snip

michaelbausor:
Added default connection settings and builder to make useage more
concise
Made newBuilder return a blank builder
Updated tests

### DIFF
--- a/vgen/src/test/java/io/gapi/vgen/testdata/java_settings_library.baseline
+++ b/vgen/src/test/java/io/gapi/vgen/testdata/java_settings_library.baseline
@@ -250,13 +250,6 @@ public class LibraryServiceSettings extends ServiceApiSettings {
 
 
   /**
-   * Returns an instance of this class with recommended defaults.
-   */
-  public static LibraryServiceSettings defaultInstance() throws IOException {
-    return defaultBuilder().build();
-  }
-
-  /**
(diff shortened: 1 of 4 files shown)


vgen/src/test/java/io/gapi/vgen/testdata/java_settings_library.baseline
```diff
@@ -244,7 +253,14 @@ public class LibraryServiceSettings extends ServiceApiSettings {
...
    */
   public static LibraryServiceSettings defaultInstance() throws IOException {
-    return newBuilder().build();
+    return defaultBuilder().build();
+  }
+
+  /**
+   * Returns a builder for this class with recommended defaults.
+   */
+  public static Builder defaultBuilder() {
```
garrettjonesgoogle: If we have a defaultBuilder, I don't think we want a defaultInstance. I have been thinking about this and I have concluded that having a defaultBuilder is safer than defaultInstance because creating an instance actually creates a channel and an executor; having a build() step is less likely to be called indiscriminately (creating resources that won't be used) as compared to a defaultInstance() call, which implies something low-cost.

michaelbausor:
PTAL

garrettjonesgoogle:
LGTM

PR #23, opened 2016-04-27
Code Samples Properly Initialize and Use API Object

pongad:
Previously, code samples call API methods without an explicit receiver.
While such code is valid inside of the API class itself,
users cannot copy-paste it.
This commit makes code samples initialize a default instance
of the API object then call the methods through that object.

### DIFF
--- a/vgen/src/test/java/io/gapi/vgen/testdata/java_main_library.baseline
+++ b/vgen/src/test/java/io/gapi/vgen/testdata/java_main_library.baseline
@@ -86,9 +86,9 @@ import java.util.concurrent.ScheduledExecutorService;
  * <pre>
  * <code>
  * try (LibraryServiceApi libraryServiceApi = LibraryServiceApi.createWithDefaults()) {
- *   // make calls here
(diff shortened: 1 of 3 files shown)


vgen/src/test/java/io/gapi/vgen/testdata/java_main_library.baseline
```diff
@@ -292,8 +295,11 @@ public class LibraryServiceApi implements AutoCloseable {
    * Creates a shelf, and returns the new Shelf.
    *
    * <pre><code>
+   * LibraryServiceApi libraryServiceApi = LibraryServiceApi.defaultInstance();
```
garrettjonesgoogle: You should probably use the try-with-resources block.

pongad: You're right, we need to close it. Will do

pongad:
PTAL @garrettjonesgoogle

garrettjonesgoogle:
LGTM

PR #24, opened 2016-04-27
Fix a few ruby generator bugs

jmuk:
- PathTemplate now uses 'render' for the method name of creating
  the result. Generator should use it.
- render assumes keyword arguments, whose keys are symbols, not
  strings.
- similarly, headers keys should be symbol as well.
- fixes a bug of TYPE_ENUM in rubyType() method -- should invoke
  getEnumType() instead of getMessageType() for TYPE_ENUM.


geigerj:
LGTM

PR #25, opened 2016-04-27
Add Media-Upload Sample to NodeJS

pongad:
Fixes #10.

This commit adds 'media' object to requests that allows media upload.
This addresses some of compilecheck's warnings.

PR #26, opened 2016-04-27
Special-Case Inconsistencies in NodeJS Discovery

pongad:
In some cases, the generated API libraries do not completely agree with
Discovery docs.
Previously, we have special-cased inconsistencies when we implemented
Java Discovery code samples. However, we did this believing that other
languages will have similar inconsistencies.
However, they seem to have *different* inconsistencies.
Meaning each language's API library are inconsistent with not only
the Discovery doc but also each other.

This commit moves the special-cases implemented earlier so that they
become Java-specific and implements more special-cases as needed for
NodeJS. Together with #25, this solves all known compilecheck issues.


pongad:
cc @vchudnov-g

garrettjonesgoogle:
Your implementation is great. The problem makes me sad. LGTM.

PR #34, opened 2016-05-02
Add docgen for Ruby

jmuk:


vgen/src/main/java/io/gapi/vgen/ruby/RDocCommentFixer.java
```diff
@@ -28,15 +31,40 @@
   private static final Pattern CLOUD_LINK_PATTERN = Pattern.compile(
       "\\[([^\\]]+)\\]\\(([^\\)]+)\\)");
   private static final Pattern PROTO_LINK_PATTERN = Pattern.compile(
-      "\\[([^\\]]+)\\]\\[[^\\]]+\\]");
+      "\\[([^\\]]+)\\]\\[[^\\]]*\\]");
```
geigerj: Do you mind updating the Python as well? I know it's not strictly related. I still think we should eventually define these in a common location since they're cross-language.

vgen/src/main/java/io/gapi/vgen/ruby/RubyGapicContext.java
```diff
@@ -168,7 +191,7 @@ private String returnTypeComment(Method method) {
     if (msg.hasAttribute(ElementDocumentationAttribute.KEY)) {
       String text = RDocCommentFixer.rdocify(DocumentationUtil.getScopedDescription(msg));
       text = text.trim();
-      contentBuilder.append(text.replaceAll("\\s*\\n\\s*", "\n"));
+      contentBuilder.append(text.replaceAll("[ \\t\\r\u000B]*\\n", "\n"));
```
geigerj: Why this change?

vgen/src/main/java/io/gapi/vgen/ruby/RubyGapicContext.java
```diff
@@ -184,6 +207,53 @@ private String returnTypeComment(Method method) {
...
 
   /**
+   * Return comment lines for a given proto element.
+   */
+  public List<String> getDocComment(ProtoElement element) {
+    StringBuilder contentBuilder = new StringBuilder();
+    if (element.hasAttribute(ElementDocumentationAttribute.KEY)) {
+      String text = RDocCommentFixer.rdocify(DocumentationUtil.getScopedDescription(element));
+      text = text.trim();
+      contentBuilder.append(text.replaceAll("[ \\t\\r\u000B]*\\n", "\n"));
```
geigerj: Let's make this new whitespace regex a constant if it's being reused.

vgen/src/main/java/io/gapi/vgen/ruby/RDocCommentFixer.java
jmuk: Probably better to be done in another CL -- filed as https://github.com/googleapis/toolkit/issues/38

vgen/src/main/java/io/gapi/vgen/ruby/RubyGapicContext.java
jmuk: This actually contains two things;
- "\\n\\s*" wouldn't be great, it removes the whitespaces at the beginning of the lines (usually used for indneted codeblocks)
- "\\s" actually contains \n, therefore continuous newlines "\n\n" will be shortened to single "\n", which means removing blank lines unexpectedly.

Added comments.

vgen/src/main/java/io/gapi/vgen/ruby/RubyGapicContext.java
jmuk: Considering about this, moved this logic -- trimming and replacing -- to RDocCommentFixer, because it's always good to be done.

jmuk:
on the offline conversation, we agreed to remove this whitespace patterns (but keep trim()).

geigerj:
LGTM

PR #36, opened 2016-05-02
Updated settings.snip to match gax-java

michaelbausor:
Added getDefaultConnectionSettings and provideChannelWith overloads


garrettjonesgoogle:
LGTM (after the GAX change is submitted)

PR #46, opened 2016-05-03
Fix go codegen for primitive resource field types

### LINKED ISSUE #44: Go codegen fails on non struct page streaming types
Problem
======
If some page streaming function iterates on a resource field that is not a Message, the code generator fails unexpectedly. This is because the generator assumes incorrectly that all resource fields are Messages and cannot be primitive types.

Since the current solution is designed around struct types, slight renaming and modification is required.

Proposal
=======
-> Instead of using the Message type name as the prefix in the page streaming constructs, use the method's name: `ShelfPage -> ListShelvesPage`
-> Replace all struct references with primitive references if the resource field is a primitive.

saicheems:
Also adds a page streaming method to `library.proto` that has a
primitive type resource field for test coverage.

Fixes #44

### DIFF
--- a/vgen/src/main/java/io/gapi/vgen/go/GoGapicContext.java
+++ b/vgen/src/main/java/io/gapi/vgen/go/GoGapicContext.java
@@ -32,6 +32,7 @@ import com.google.protobuf.DescriptorProtos.FieldDescriptorProto.Type;
 import io.gapi.vgen.ApiConfig;
 import io.gapi.vgen.CollectionConfig;
 import io.gapi.vgen.GapicContext;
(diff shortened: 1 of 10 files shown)


saicheems:
The real relevant changes are in `.../go/main.snip`, `.../testdata/go_library.baseline`, and `.../go/GoGapicContext.java`

The rest are baseline changes due to the new proto method.
PTAL

saicheems:
@vchudnov-g 
@jmuk 
@jba 
@pongad

jmuk:
looks good to me

jba:
LGTM.

pongad:
lgtm

PR #49, opened 2016-05-03
Update Java codegen with PageAccessor.

shinfan:
- Also fixed a few missing imports.


michaelbausor:
LGTM once travis completes

PR #50, opened 2016-05-03
Add license

pongad:


garrettjonesgoogle:
LGTM.

Maybe we should add a license check of our source files as part of the tests in our build?

pongad:
Added #56

PR #54, opened 2016-05-04
Bootstrap snippet gen for PHP.

shinfan:
- Bootstrap PHP Language provider.
- Bootstrap PHP Discovery context.
- Currently the generated snippets are empty.


vgen/src/main/resources/io/gapi/vgen/php/php_discovery_gapic.yaml
```diff
@@ -0,0 +1,4 @@
+type: io.gapi.vgen.ConfigProto
```
garrettjonesgoogle: Remove "_gapic" from this filename

vgen/src/main/java/io/gapi/vgen/php/PhpDiscoveryLanguageProvider.java
```diff
@@ -0,0 +1,65 @@
...
+      throws IOException {
+    Api api = getService().getApis(0);
+    String root =
+        "autogenerated/"
+            + api.getName()
+            + "/"
+            + api.getVersion()
+            + "/"
+            + context.getApiRevision()
+            + "/";
```
garrettjonesgoogle: Thomas is moving the line to compose this string into DiscoveryContext (see https://github.com/googleapis/toolkit/pull/48/files)

shinfan: Acknowledged. I will switch to use the one in DiscoveryContext after Thomas's PR is merged.

vgen/src/main/resources/io/gapi/vgen/php/php_discovery_gapic.yaml
shinfan: Right now this naming pattern is consistent with Ruby and Java. If possible, I would prefer to leave it as-is for now and address this for all languages in a separate PR.

garrettjonesgoogle: This rename is part of Thomas' change, so if you submit before he does, then he would want to add the rename into his own PR; if you submit after, then you would want to include the rename.

garrettjonesgoogle:
LGTM

PR #58, opened 2016-05-05
Add license check for Java files

### LINKED ISSUE #56: Test that all source files have licenses

pongad:
Fixes #56.


build.gradle
```diff
@@ -291,4 +293,33 @@ project(':vgen') {
       args = clargs.split(',').toList()
     }
   }
+
+  task verifyLicense << {
+    def licenseText = '''/* Copyright 2016 Google Inc
```
garrettjonesgoogle: It would be nice if we could derive the check from a file that was outside of this build file.

pongad: I uploaded a change to move the license into its own file. It feels a little funny to have it in the top-level, but I'm not sure where else to put it. Let me know what you think.

java-license.txt
```diff
@@ -0,0 +1,14 @@
+/* Copyright 2016 Google Inc
```
garrettjonesgoogle: This file isn't the license for java; maybe call the filename "license-header-javadoc.txt"

pongad: That's what I meant, but I now see how the name is confusing. Will fix this.

pongad:
@garrettjonesgoogle PTAL

garrettjonesgoogle:
+1; have one more person review it.

saicheems:
lgtm

PR #59, opened 2016-05-05
Update go codegen

saicheems:
Fixes some build errors, removes the cloud package from the public
surface.
Also changed the snippet file to only use spaces. It makes the editing
process a lot easier, and `gofmt` can convert to tabs through a cleanup
step in the pipeline.

### DIFF
--- a/vgen/src/main/resources/io/gapi/vgen/go/main.snip
+++ b/vgen/src/main/resources/io/gapi/vgen/go/main.snip
@@ -52,14 +52,14 @@
 
     // Client is a client for interacting with {@service.getSimpleName}.
     type Client struct {
-    	// The connection to the service.
-    	conn *grpc.ClientConn
+        // The connection to the service.
+        conn *grpc.ClientConn
 
-    	// The gRPC API client.
-    	client {@context.getServiceClientName(service)}
+        // The gRPC API client.
+        client {@context.getServiceClientName(service)}
 
-    	// The map from the method name to the default call options for the method of this service.
-    	CallOptions map[string][]gax.CallOption
+        // The map from the method name to the default call options for the method of this service.
+        callOptions map[string][]gax.CallOption
     }
 
     {@constructor(service)}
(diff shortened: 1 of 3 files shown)


saicheems:
@jmuk 
@jba 
@pongad 
@vchudnov-g

vgen/src/main/resources/io/gapi/vgen/go/main.snip
```diff
@@ -210,8 +210,8 @@
 
                     {@methodComment(method, methodName)}
                     func (client *Client) {@methodName}(ctx context.Context, req {@inTypeName}, opts ...gax.CallOption) (*{@iteratorTypeName}, error) {
-                    	callOpts := append({@retryCodesMethodName}RetryCodes(), {@retryParamsMethodName}RetryParams(), client.CallOptions["{@methodName}"]...)
-                    	callOpts = append(callOpts, opts...)
+                    	defaultCallOpts := append({@retryCodesMethodName}RetryCodes(), {@retryParamsMethodName}RetryParams()...)
+                    	callOpts := append(client.CallOptions["{@methodName}"], opts...)
```
saicheems: Not super satisfied with all the appends thrown around... but I couldn't think of a better way (other than making a function that takes an arbitrary number of `[]CallOption`).

LMK

vgen/src/test/java/io/gapi/vgen/testdata/go_library.baseline
```diff
@@ -191,14 +189,14 @@ func (book Book) Path() (string, error) {
 
 // CreateShelf creates a shelf, and returns the new Shelf.
 func (client *Client) CreateShelf(ctx context.Context, req *google_example_library_v1.CreateShelfRequest, opts ...gax.CallOption) (*google_example_library_v1.Shelf, error) {
-	callOpts := append(nonIdempotentRetryCodes(), defaultRetryParams(), client.CallOptions["CreateShelf"]...)
-	callOpts = append(callOpts, opts...)
+	defaultCallOpts := append(nonIdempotentRetryCodes(), defaultRetryParams()...)
+	callOpts := append(client.CallOptions["CreateShelf"], opts...)
```
pongad: I am a little worried about this. If `client.CallOptions[s]` is not a filled slice, we will have a race condition if two goroutines call the same method on the same client concurrently. See example in https://play.golang.org/p/tgObgdGEsG .

It probably also worth documenting that the user may not assign/modify client.CallOptions concurrently with any calls on the client.

pongad:
There's a potential surprising behavior I'm worried about, other than that I think it looks good

vgen/src/test/java/io/gapi/vgen/testdata/go_library.baseline
saicheems: Wow yeah nice catch!

On inspection I think the right course of action is to not export `client.CallOptions` given that it really shouldn't be modifiable outside of initializing a new client struct.

There's still some work to be done on client configuration before it reaches parity with the python implementation... for now I'll make that struct var private.

vgen/src/test/java/io/gapi/vgen/testdata/go_library.baseline
```diff
@@ -109,24 +107,24 @@ type Client struct {
...
-	conn, err := transport.DialGRPC(ctx, o...)
+	conn, err := gax.DialGRPC(ctx, o...)
 	if err != nil {
 		return nil, err
 	}
 	return &Client {
 		conn: conn,
 		client: google_example_library_v1.NewLibraryServiceClient(conn),
-		CallOptions: map[string][]gax.CallOption{},
+		callOptions: map[string][]gax.CallOption{},
```
jmuk: This callOptions is now hidden and always empty... Why does this exist? Is there a way this can be updated?
You can simply remove it if you don't need it anymore.
(or, probably add a new gax.ClientOption to reset the default callOptions?)

vgen/src/test/java/io/gapi/vgen/testdata/go_library.baseline
```diff
@@ -191,14 +189,14 @@ func (book Book) Path() (string, error) {
 
 // CreateShelf creates a shelf, and returns the new Shelf.
 func (client *Client) CreateShelf(ctx context.Context, req *google_example_library_v1.CreateShelfRequest, opts ...gax.CallOption) (*google_example_library_v1.Shelf, error) {
-	callOpts := append(nonIdempotentRetryCodes(), defaultRetryParams(), client.CallOptions["CreateShelf"]...)
-	callOpts = append(callOpts, opts...)
+	defaultCallOpts := append(nonIdempotentRetryCodes(), defaultRetryParams()...)
+	callOpts := append(client.callOptions["CreateShelf"], opts...)
```
jmuk: This reminds me of a discussion in the internal code review. append() has an optimization that if the latter part (client.callOptions["CreateShelf"] in this case) has enough space, append() actually updates it rather than creating a new slice, thus this operation is not very safe sometimes. You may want to make a copy instead.

vgen/src/test/java/io/gapi/vgen/testdata/go_library.baseline
saicheems: My plan is to allow it to be instantiated through GAX (WithRetryOptionsOverride)
Was also considering populating it to supply the defaults...

I'll leave it for now

jmuk: I see, then that's good.

Please address my comment below about tricky behavior of append(). It actually depends on how callOptions will be constructed from the retry-options-override though.

saicheems: Yup, will do

vgen/src/test/java/io/gapi/vgen/testdata/go_library.baseline
jba: Yes, this is buggy. append does try to modify its first argument if there is space.

I suggest simple, linear code:

var callOpts []gax.CallOption
callOpts = append(callOpts, thing1)
callOpts = append(callOpts, thing2)
...

(This is not optimal in space or time, but it almost certainly won't matter.)

saicheems: Makes sense, fixed.

saicheems:
Introduced a lot of noise just now - I converted all tabs in the snippet file to spaces. I ran into a bug due to mix and matching... When `gofmt` is integrated into the pipeline, it will not be an issue.

All other concerns are addressed.

jmuk:
LGTM. I'm a bit concerned about using spaces, but actually artman runs gofmt, so you don't have to worry about this much (see https://github.com/googleapis/artman/blob/master/pipeline/pipelines/code_generation_pipeline.py#L274)

vgen/src/test/java/io/gapi/vgen/testdata/go_library.baseline
```diff
@@ -52,226 +52,236 @@ import (
...
-	}, callOpts...)
-	if err != nil {
-		return nil, err
-	}
-	return resp, nil
+    var callOpts []gax.CallOptions
+    callOpts = append(callOpts, nonIdempotentRetryCodes())
+    callOpts = append(callOpts, defaultRetryParams())
+    callOpts = append(callOpts, client.callOptions["CreateShelf"])
+    callOpts = append(callOpts, opts...)
```
pongad: I just realized we can use gax.callOptions if it were exported. I'm not sure if it's worth changing at this point.

saicheems:
Yeah the snippet language leaves us in an odd state where using all tabs is buggy, and a mix of tabs and spaces is annoying to develop with and also buggy...
I think it should be fine if artman runs `gofmt` (cool!)

vgen/src/test/java/io/gapi/vgen/testdata/go_library.baseline
saicheems: I'll leave it private in this CL, my plan right now is to make `ClientSettings`/`CallSettings` public in `gax`, and set `client.callOptions` during initialization.

There shouldn't really be a pathway imo for users to modify client level settings like retry overrides (the intended use case of `client.callOptions` outside of initialization.

PR #66, opened 2016-05-06
Draft snippet generation for PHP.

shinfan:
- Draft .snip
- Implement helper methods in PHP discovery context.
- Add baseline tests.


shinfan:
@garrettjonesgoogle @vchudnov-g  @pongad

vgen/src/main/resources/io/gapi/vgen/php/discovery_fragment.snip
```diff
@@ -3,7 +3,123 @@
 @end
 
 @snippet generateClass(method, body, imports)
+  @let api = context.getApi, \
+       apiName = api.getName, \
+       serviceTitle = context.getService.getTitle
+    <?php
```
pongad: is this tag closed?

pongad: ah, closed only when embedded in html. OK

vgen/src/main/resources/io/gapi/vgen/php/discovery_fragment.snip
```diff
@@ -3,7 +3,123 @@
...
+
+  @end
 @end
 
 @snippet generateBody(method)
+  @let apiName = context.getSimpleApiName, \
+       apiaryConfig = context.getApiaryConfig, \
+       serviceName = context.getServiceClassName, \
+       params = context.getFlatMethodParams(method)
+    function {@apiName}Example
```
pongad: Does PHP allow for just free-floating code? If it does, you don't need to wrap this in a function.

vgen/src/test/java/io/gapi/vgen/testdata/discoveries/php/php_appengine.v1beta4.json.baseline
```diff
@@ -0,0 +1,433 @@
...
+  $service = new Google_Service_Appengine($client);
+
+  // Part of `name`. Name of the resource requested. For example: "apps/myapp/modules/default".
+  $appsId = '';
+  // Part of `name`. See documentation of `appsId`.
+  $modulesId = '';
+
+  $response = $service->apps->modules->versions->listVersions($appsId, $modulesId);
+
+  foreach ($response['versions'] as $versions) {
```
pongad: Ideally the $versions should be singular. I think $element might work better.

pongad:
A couple of comments but it looks good. Do you have any idea how we might "compile" check this?

shinfan:
I probably need more time to look into the compile check, but I assume it should be similar to NodeJS since both PHP and Node are interpreted languages.
There is also a PHP method that does the syntax check: http://php.net/manual/en/function.php-check-syntax.php

pongad:
I am not worried about syntax check. Syntax is (at least I hope) easy to get right, and if we get it right for a few cases, we can be pretty sure we get it right for everything, since our sample gen is not that complicated.

I'm a lot more worried that due to some inconsistencies in API libraries, we generate sample with wrong types or wrong number of parameters, and calls will never succeed. We have encountered these in Java, and it's too time-consuming to verify all samples by hand.

shinfan:
TODO: Implement compile check https://github.com/googleapis/toolkit/issues/93

pongad:
lgtm

PR #69, opened 2016-05-06
Bump appengine to v1beta5

pongad:


vchudnov-g:
What's the rationale for bumping the version numbers? Are we exercising a new code path or capturing a bug fix?

It's  not obvious to me that we need to keep the test files that are testing our snippet-generating code in sync with the APIs they were originally derived from. The original inputs were valid Discovery files, so they remain valid test cases. I don't think we need to be in the business of making our tests always reflect the latest APIs; all they need to do is work correctly against all valid Discovery docs. Plus, by removing the previous version of the file there is a chance of a backward incompatibility if the new inputs cease exercising a code path that was previously invoked.

pongad:
I neglected to document this in the commit message. If the rationale looks good to you, I will include it in the commit message and push.

The rationale is that Ruby's method names are computed by either a non-trivial algorithm or looked up in a list of names in the Ruby repo. The list of names is therefore required for us to be able to generate correct samples, and so it might as well be an error if we can't look up the name. I copied the list of name from the Ruby repo into toolkit (in my own WIP branch, not yet in master). The list contains only contains the naming for some versions of the APIs. The two PRs bump our versions of discovery to these new versions so that they may be tested properly.

Of course, if you are concerned about backward incompatibility, we can just add the new versions and disable Ruby testing on the old ones. Other languages will still pass.

vchudnov-g:
OK, so let me recap to make sure I understand:
- For Ruby, we are looking up the method names in a list
- We want to test this, and error if we can't look up a name
- For this test, we need Discovery files
- The list of names for Ruby is no longer available for the versions of the APIs for which we currently have Discovery files checked in
- There's no need to have duplicate Discovery files for Ruby vs the other languages
- So we just update the Discovery files in testdata/ for use by all languages.

Is this correct? If so, I'm OK with this, but do include this logic in the commit message.

BTW, have you had any luck locating/figuring out the algorithm Ruby uses to generate the names? That seems like a more sustainable solution.....

vchudnov-g:
OK, talked offline.

LGTM, but please include the logic above in the commit message, as well as a note saying that the name list is generated by the Ruby client library generators (not us) and always lives in a canonical place (and specify where).

pongad:
Synced with @vchudnov-g offline. I'll merge this since he LGTM'ed.

PR #70, opened 2016-05-07
Introduce 'call_settings' parameter

jmuk:
This removes retrying_overrides and bundling_overrides, because
I think it's quite easy to modify the parameters before invoking
the constructor.

This fixes googleapis/gax-python#95 and googleapis/gax-ruby#5

### DIFF
--- a/src/main/resources/com/google/api/codegen/py/main.snip
+++ b/src/main/resources/com/google/api/codegen/py/main.snip
@@ -154,46 +154,37 @@
     @end
 @end
 
-@private constructSettings(prefix, config, service, ifaceConfig)
-    @if or(context.messages.filterPageStreamingMethods(ifaceConfig, service.getMethods), \
-           context.messages.filterBundlingMethods(ifaceConfig, service.getMethods))
-        {@prefix}api_callable.construct_settings(
-            '{@service.getFullName}',
-            {@config},
-            {},
-            {},
-            config.STATUS_CODE_NAMES,
-            timeout,
-            @if context.messages.filterBundlingMethods(ifaceConfig, service.getMethods)
-                @if context.messages.filterPageStreamingMethods(ifaceConfig, service.getMethods)
-                    bundle_descriptors=self._BUNDLE_DESCRIPTORS,
-                @else
-                    bundle_descriptors=self._BUNDLE_DESCRIPTORS)
-                @end
-            @end
-            @if context.messages.filterPageStreamingMethods(ifaceConfig, service.getMethods)
-                page_descriptors=self._PAGE_DESCRIPTORS)
-            @end
-    @else
-        {@prefix}api_callable.construct_settings(
-            '{@service.getFullName}',
-            {@config},
-            {},
-            {},
-            config.STATUS_CODE_NAMES,
-            timeout)
-    @end
-@end
-
 @private constructDefaults(service)
     @let ifaceConfig = context.getApiConfig.getInterfaceConfig(service), \
          jsonBaseName = {@context.upperCamelToLowerUnderscore(service.getSimpleName)}
         default_client_config = json.loads(pkg_resources.resource_string(
             __name__, '{@jsonBaseName}_client_config.json'))
-        {@constructSettings("self._defaults = ", "default_client_config", service, ifaceConfig)}
-        if client_config:
-            {@constructSettings("self._defaults.update(", "client_config", service, ifaceConfig)}
-            )
+        @if or(context.messages.filterPageStreamingMethods(ifaceConfig, service.getMethods), \
+               context.messages.filterBundlingMethods(ifaceConfig, service.getMethods))
+            self._defaults = api_callable.construct_settings(
+                '{@service.getFullName}',
+                default_client_config,
+                client_config,
+                config.STATUS_CODE_NAMES,
+                timeout,
+                @if context.messages.filterBundlingMethods(ifaceConfig, service.getMethods)
+                    @if context.messages.filterPageStreamingMethods(ifaceConfig, service.getMethods)
+                        bundle_descriptors=self._BUNDLE_DESCRIPTORS,
+                    @else
+                        bundle_descriptors=self._BUNDLE_DESCRIPTORS)
+                    @end
+                @end
+                @if context.messages.filterPageStreamingMethods(ifaceConfig, service.getMethods)
+                    page_descriptors=self._PAGE_DESCRIPTORS)
+                @end
+        @else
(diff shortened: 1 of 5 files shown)


vgen/src/main/resources/io/gapi/vgen/py/main.snip
```diff
@@ -212,18 +217,10 @@
...
-                custom RetryOptions objects, or to None. A value of None indicates
-                that the method in question should not retry.
-              bundling_override (dict[string, :class:`google.gax.BundleOptions`]): A
-                dictionary that overrides default bundling settings.
-                ``bundling_override`` maps bundling method names (e.g.,
-                'publish_foo') to custom BundleOptions objects, or to None. It is
-                invalid to have a key for a method that is not bundling-enabled. A
-                value of None indicates that the method in question should not
-                bundle.
+              client_config (dict or :class:`google.gax.CallSettings`):
```
geigerj: If a dict is specified, preferably we should use it to update the default client config. That way you can pass in a partial specification.

CallSettings also includes attribute that are not intended to be overridden by the user, like PageDescriptor and BundleDescriptor, so I'm not sure it's the right object to use here.

vgen/src/test/java/io/gapi/vgen/testdata/python_no_path_templates.baseline
```diff
@@ -98,18 +89,20 @@ class NoTemplatesServiceApi(object):
...
-        self._defaults = api_callable.construct_settings(
-            'google.example.v1.NoTemplatesService',
-            json.loads(client_config),
-            bundling_override,
-            retrying_override,
-            config.STATUS_CODE_NAMES,
-            timeout
-        )
+        if isinstance(client_config, google.gax.CallSettings):
+            self._defaults = client_config
```
geigerj: self._defaults must be a dict; the generated code contains such lines as

`settings = self._defaults['increment'].merge(options)`

vgen/src/test/java/io/gapi/vgen/testdata/python_no_path_templates.baseline
```diff
@@ -98,18 +89,20 @@ class NoTemplatesServiceApi(object):
...
+        else:
+            if not client_config:
+                client_config = json.loads(pkg_resources.resource_string(
+                    __name__, 'no_templates_service_client_config.json'))
+            self._defaults = api_callable.construct_settings(
```
geigerj: What if we do

```
client_config = json.loads(pkg_resources.resource_string(__name__, 'no_templates_service_client_config.json'))
client_config.update(user_client_config)
```

Then the user only needs to specify the options for the methods they are overriding. (Even better, do the update in the method dictionaries, so that you specify a bundling options and not override the retry options?)

vgen/src/test/java/io/gapi/vgen/testdata/python_no_path_templates.baseline
jmuk: Ow, right.
Then probably call_config parameter should be always dict() (Hash in Ruby), which is usually the return value of construct_settings, instead of raw JSON values.  Updated to do so -- both code and the comment.

jmuk:
So I think my code had a confusion about the JSON-parsed config vs the return value of `construct_settings()`, and probably we want to accept the return value of construct_settings rather than other data (like parsed result of JSON config files).

geigerj:
I still think that this means that to override any bundling/retry/page streaming behavior, even for a single method, you must provide an overly complex dictionary over every method in the API, and the keys of that dict must include non-configurable information like PageDescriptor.

jmuk:
You're right about the bundling descriptors and page descriptors.

I'm still suspicious about retrying_overrides and bundling_overrides though. With this client_config param, users can simply modify the input dict() (or JSON files) directly instead of passing override parameters, and I think we can omit them from the constructor parameters.

geigerj:
I agree that we should not have retrying_overrides or bundling_overrides, and should provide a single override parameter instead.

What I am unclear about is the structure of the dictionary being passed in. In my mind, it should

- be possible to override only some of the methods in an API
- be possible to override only some of the behaviors (retry, timeout, bundling, ...) of a particular method
- not be possible to override PageDescriptor or BundleDescriptor

jmuk:
Hm, I see. So you are saying to allow call_settings specify _a part_ of methods, only for customized ones. Uploaded a new one, which essentially invokes construct_settings() twice so that it can fall back to the default config.

Or, probably fix construct_settings interface?  Any ideas?

geigerj:
Yes, I think this would require a change to `construct_settings` to get the "parts of a method" level of granularity in overriding behavior.

I think there's also some more subtlety to the merging part of this setup that needs to be addressed in GAX. (For example, suppose the user defines a new `retry_params` whose name shadows a name in the default -- do we want the new `retry_params` to shadow all of the methods which reference the old `retry_params` name? I suspect in this case we'd only want the _overriden_ method to reference the new `retry_params`.)

jmuk:
Created a change on GAX to address the issue.  PTAL.

geigerj:
When you have a chance, could you update this to reflect the changes merged into GAX?

jmuk:
Updated for the new construct_settings behavior. PTAL

src/main/resources/com/google/api/codegen/py/main.snip
```diff
@@ -215,18 +212,11 @@
...
-                that the method in question should not retry.
-              bundling_override (dict[string, :class:`google.gax.BundleOptions`]): A
-                dictionary that overrides default bundling settings.
-                ``bundling_override`` maps bundling method names (e.g.,
-                'publish_foo') to custom BundleOptions objects, or to None. It is
-                invalid to have a key for a method that is not bundling-enabled. A
-                value of None indicates that the method in question should not
-                bundle.
+              client_config (dict):
+                A dictionary for call settings for each method. See
```
geigerj: call settings --> call options (or "configurable call options"? or "configurable call settings"). Hopefully this isn't confusing, since we have the separate objects CallSettings and CallOptions, neither of which corresponds exactly to the configuration for a method.

jmuk: done

geigerj: Sorry to go back on what I just said, but can we actually refer to the data structure as "call configuration"? Otherwise I don't like the collision with GAX objects CallOptions/CallSettings

geigerj:
Just an optional comment on updating a doc reference, but I don't want to block on this.

LGTM

PR #71, opened 2016-05-07
Refactor language providers

geigerj:
Each language provider instance now represents a single template
(i.e. code, doc, or fragment) for a given language. It has an internal
strategy object (*Viewing) to encapsulate template-specific generation
details. The structure of the language provider is simplified to have
a single generate method, corresponding to the generation of a single
template.

The structure of the GAPIC config YAML files has changed to push the
snippet definitions down to a per-template level, to avoid the
explosion of keys for various types of snippet files.

Pre-push hook installed.

Change-Id: Id682441e1b8a3f3209974e0d3ee3642d5a6ce610

### DIFF
--- a/vgen/src/main/java/io/gapi/vgen/CodeGenerator.java
+++ b/vgen/src/main/java/io/gapi/vgen/CodeGenerator.java
@@ -14,24 +14,15 @@
  */
 package io.gapi.vgen;
 
-import com.google.api.tools.framework.model.Field;
-import com.google.api.tools.framework.model.Interface;
-import com.google.api.tools.framework.model.MessageType;
-import com.google.api.tools.framework.model.Method;
 import com.google.api.tools.framework.model.Model;
-import com.google.api.tools.framework.model.ProtoFile;
+import com.google.api.tools.framework.model.ProtoElement;
 import com.google.api.tools.framework.model.stages.Merged;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Multimap;
-import com.google.protobuf.DescriptorProtos.FieldDescriptorProto.Type;
 
 import java.io.IOException;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
 import java.util.Map;
-import java.util.Set;
 
 import javax.annotation.Nullable;
 
@@ -40,14 +31,16 @@ import javax.annotation.Nullable;
  */
 public class CodeGenerator {
 
-  private final GapicLanguageProvider provider;
+  private final GapicLanguageProvider<ProtoElement> provider;
 
-  public CodeGenerator(GapicLanguageProvider provider) {
+  public CodeGenerator(GapicLanguageProvider<ProtoElement> provider) {
     this.provider = Preconditions.checkNotNull(provider);
   }
 
-  public static CodeGenerator create(ConfigProto configProto, Model model) {
-    return new CodeGenerator(GeneratorBuilderUtil.createLanguageProvider(configProto, model));
+  public static CodeGenerator create(ConfigProto configProto, TemplateProto template, Model model) {
+    InputElementView<ProtoElement> view = GeneratorBuilderUtil.createView(template, model);
+    return new CodeGenerator(
+        GeneratorBuilderUtil.createLanguageProvider(configProto, template, model, view));
   }
 
   /**
@@ -55,7 +48,9 @@ public class CodeGenerator {
    * Returns null if generation failed.
    */
   @Nullable
-  public Map<Interface, GeneratedResult> generate(SnippetDescriptor snippetDescriptor) {
+  public Map<ProtoElement, GeneratedResult> generate(SnippetDescriptor snippetDescriptor) {
+    Iterable<ProtoElement> elements = provider.getView().getElementIterable(provider.getModel());
+
     // Establish required stage for generation.
     provider.getModel().establishStage(Merged.KEY);
     if (provider.getModel().getErrorCount() > 0) {
@@ -63,13 +58,13 @@ public class CodeGenerator {
     }
 
     // Run the generator for each service.
-    ImmutableMap.Builder<Interface, GeneratedResult> generated = ImmutableMap.builder();
-    for (Interface iface : provider.getModel().getSymbolTable().getInterfaces()) {
-      if (!iface.isReachable()) {
+    ImmutableMap.Builder<ProtoElement, GeneratedResult> generated = ImmutableMap.builder();
+    for (ProtoElement element : elements) {
+      if (!element.isReachable()) {
         continue;
       }
-      GeneratedResult result = provider.generateCode(iface, snippetDescriptor);
-      generated.put(iface, result);
+      GeneratedResult result = provider.generate(element, snippetDescriptor);
+      generated.put(element, result);
     }
 
     // Return result.
@@ -79,43 +74,9 @@ public class CodeGenerator {
     return generated.build();
   }
 
-  @Nullable
-  public Map<String, GeneratedResult> generateDocs(SnippetDescriptor snippetDescriptor) {
-    Set<ProtoFile> files = new HashSet<ProtoFile>();
-    for (Interface iface : provider.getModel().getSymbolTable().getInterfaces()) {
-      for (Method method : iface.getMethods()) {
-        for (Field field : method.getInputType().getMessageType().getFields()) {
-          files.addAll(protoFiles(field));
-        }
-      }
-    }
-    Map<String, GeneratedResult> generated = new HashMap();
-    for (ProtoFile file : files) {
-      GeneratedResult result = provider.generateDoc(file, snippetDescriptor);
-      generated.put(result.getFilename(), result);
-    }
-    if (provider.getModel().getErrorCount() > 0) {
-      return null;
-    }
-    return generated;
-  }
-
-  private Set<ProtoFile> protoFiles(Field field) {
-    Set<ProtoFile> fields = new HashSet<ProtoFile>();
-    if (field.getType().getKind() != Type.TYPE_MESSAGE) {
-      return fields;
-    }
-    MessageType messageType = field.getType().getMessageType();
-    fields.add(messageType.getFile());
-    for (Field f : messageType.getNonCyclicFields()) {
-      fields.addAll(protoFiles(f));
-    }
-    return fields;
-  }
-
   /**
    * Delegates creating code to language provider. Takes the result list from
-   * {@link GapicLanguageProvider#outputCode(String, List)} and stores it in a language-specific
+   * {@link GapicLanguageProvider#output(String, Multimap)} and stores it in a language-specific
    * way.
    */
   public <Element> void output(String outputFile, Multimap<Element, GeneratedResult> results)
--- a/vgen/src/main/java/io/gapi/vgen/DiscoveryFragmentGenerator.java
+++ b/vgen/src/main/java/io/gapi/vgen/DiscoveryFragmentGenerator.java
@@ -48,12 +48,15 @@ public class DiscoveryFragmentGenerator {
 
     ApiaryConfig apiaryConfig = discovery.getConfig();
 
+    // TODO: Support multiple templates; don't hard-code the zero-index here.
+    Preconditions.checkArgument(configProto.getTemplatesCount() == 1);
     DiscoveryLanguageProvider languageProvider =
-        GeneratorBuilderUtil.createLanguageProvider(
-            configProto.getLanguageProvider(),
+        GeneratorBuilderUtil.createClass(
+            configProto.getTemplates(0).getLanguageProvider(),
             DiscoveryLanguageProvider.class,
             new Class<?>[] {Service.class, ApiaryConfig.class},
             new Object[] {discovery.getService(), apiaryConfig},
+            "discovery language provider",
             new GeneratorBuilderUtil.ErrorReporter() {
               @Override
               public void error(String message, Object... args) {
--- a/vgen/src/main/java/io/gapi/vgen/DiscoveryFragmentGeneratorApi.java
+++ b/vgen/src/main/java/io/gapi/vgen/DiscoveryFragmentGeneratorApi.java
@@ -19,6 +19,7 @@ import com.google.api.tools.framework.model.testing.SimpleDiag;
 import com.google.api.tools.framework.tools.ToolOptions;
 import com.google.api.tools.framework.tools.ToolOptions.Option;
 import com.google.common.base.Joiner;
+import com.google.common.base.Preconditions;
 import com.google.common.base.Splitter;
 import com.google.common.base.Strings;
 import com.google.common.collect.ArrayListMultimap;
@@ -101,7 +102,10 @@ public class DiscoveryFragmentGeneratorApi {
     }
 
     Multimap<Method, GeneratedResult> docs = ArrayListMultimap.create();
-    for (String snippetInputName : configProto.getFragmentFilesList()) {
+    
+    // TODO: Support multiple templates; don't hard-code the zero-index here.
+    Preconditions.checkArgument(configProto.getTemplatesCount() == 1);
+    for (String snippetInputName : configProto.getTemplates(0).getSnippetFilesList()) {
       SnippetDescriptor snippetDescriptor = new SnippetDescriptor(snippetInputName);
       Map<Method, GeneratedResult> code = generator.generateFragments(snippetDescriptor);
       if (code == null) {
--- a/vgen/src/main/java/io/gapi/vgen/GeneratorBuilderUtil.java
+++ b/vgen/src/main/java/io/gapi/vgen/GeneratorBuilderUtil.java
@@ -15,7 +15,9 @@
 package io.gapi.vgen;
 
 import com.google.api.tools.framework.model.Diag;
+import com.google.api.tools.framework.model.DiagCollector;
 import com.google.api.tools.framework.model.Model;
+import com.google.api.tools.framework.model.ProtoElement;
 import com.google.api.tools.framework.model.SimpleLocation;
 import com.google.api.tools.framework.model.stages.Merged;
 import com.google.common.base.Preconditions;
@@ -36,12 +38,24 @@ public final class GeneratorBuilderUtil {
     void error(String message, Object... args);
   }
 
-  public static GapicLanguageProvider createLanguageProvider(ConfigProto configProto, final Model model) {
+  /**
+   * Constructs a language provider from a configuration; the configuration cannot contain
+   * parameterized type information, so we must suppress the type safety warning here.
+   */
+  @SuppressWarnings("unchecked")
+  public static GapicLanguageProvider<ProtoElement> createLanguageProvider(
+      ConfigProto configProto,
+      TemplateProto template,
+      final Model model,
+      InputElementView<ProtoElement> view) {
     Preconditions.checkNotNull(model);
     Preconditions.checkNotNull(configProto);
 
     model.establishStage(Merged.KEY);
     if (model.getErrorCount() > 0) {
+      for (Diag diag : model.getDiags()) {
+        System.err.println(diag.toString());
+      }
       return null;
     }
 
@@ -50,11 +64,12 @@ public final class GeneratorBuilderUtil {
       return null;
     }
 
-    return GeneratorBuilderUtil.createLanguageProvider(
-        configProto.getLanguageProvider(),
+    return GeneratorBuilderUtil.createClass(
+        template.getLanguageProvider(),
         GapicLanguageProvider.class,
-        new Class<?>[] {Model.class, ApiConfig.class},
-        new Object[] {model, apiConfig},
+        new Class<?>[] {Model.class, ApiConfig.class, InputElementView.class},
+        new Object[] {model, apiConfig, view},
+        "language provider",
         new GeneratorBuilderUtil.ErrorReporter() {
           @Override
           public void error(String message, Object... args) {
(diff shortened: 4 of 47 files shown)


geigerj:
@bjwatson, who pair programmed this change with me. (Is there a way to attribute the commit to two authors?)

geigerj:
cc: @pongad (since this touches some Discovery files)

bjwatson:
@geigerj Nope, Github's not that clever. You're the primary. :)

vgen/src/main/java/io/gapi/vgen/CodeGenerator.java
```diff
@@ -40,36 +32,40 @@
...
     if (provider.getModel().getErrorCount() > 0) {
       return null;
     }
 
     // Run the generator for each service.
-    ImmutableMap.Builder<Interface, GeneratedResult> generated = ImmutableMap.builder();
-    for (Interface iface : provider.getModel().getSymbolTable().getInterfaces()) {
-      if (!iface.isReachable()) {
+    ImmutableMap.Builder<ProtoElement, GeneratedResult> generated = ImmutableMap.builder();
+    for (ProtoElement elt : elts) {
```
garrettjonesgoogle: Generally in Java, words will be fully spelled out, so use element/elements instead of elt/elts

vgen/src/main/java/io/gapi/vgen/DiscoveryFragmentGenerator.java
```diff
@@ -48,9 +48,10 @@ public static DiscoveryFragmentGenerator create(
 
     ApiaryConfig apiaryConfig = discovery.getConfig();
 
+    // TODO: Support multiple templates; don't hard-code the zero-index here.
```
garrettjonesgoogle: Maybe throw an exception if there is more than one, so that discovering the lack of support is easier when someone tries to use it down the road.

vgen/src/main/java/io/gapi/vgen/GapicLanguageProvider.java
```diff
@@ -26,24 +24,19 @@
...
-   * Generates the doc for the given proto file and snippet.
+   * Runs code generation.
    */
-  GeneratedResult generateDoc(ProtoFile file, SnippetDescriptor descriptor);
-
-  /**
-   * Generates a fragment for the given method.
-   */
-  GeneratedResult generateFragments(Method method, SnippetDescriptor snippetDescriptor);
+  GeneratedResult generate(T t, SnippetDescriptor snippetDescriptor);
```
garrettjonesgoogle: is 't' a template? Rename t to the full word.

vgen/src/main/java/io/gapi/vgen/ProtoElementViewing.java
```diff
@@ -0,0 +1,26 @@
...
+
+import com.google.api.tools.framework.model.Model;
+import com.google.api.tools.framework.model.ProtoElement;
+
+/**
+ * An implementation of ProtoElementViewing is a strategy object, encapsulating a strategy for
+ * retrieving a ProtoElement from a model.
+ */
+public interface ProtoElementViewing<T extends ProtoElement> {
+  public Iterable<T> getIterable(Model model);
```
garrettjonesgoogle: Maybe make it clearer what the iterable is for - e.g. 'getElementIterable'

vgen/src/main/java/io/gapi/vgen/ProtoFileViewing.java
```diff
@@ -0,0 +1,69 @@
...
+import com.google.api.tools.framework.model.Method;
+import com.google.api.tools.framework.model.Model;
+import com.google.api.tools.framework.model.ProtoFile;
+import com.google.protobuf.DescriptorProtos.FieldDescriptorProto.Type;
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * An file-based view of model, consisting of a strategy for getting the protocol buffer files from
```
garrettjonesgoogle: -> "A file-based view..."

vgen/src/main/java/io/gapi/vgen/py/PythonLanguageProvider.java
```diff
@@ -63,4 +63,4 @@
     Doc result = snippets.generateClass(element, importList);
     return GeneratedResult.create(result, pathPrefix + outputFilename);
   }
-}
+}
```
garrettjonesgoogle: revert file (the only change is a disappearing newline at the end)

vgen/src/main/resources/io/gapi/vgen/java/java_gapic.yaml
```diff
@@ -1,8 +1,10 @@
...
-fragment_files:
-- fragment.snip
+templates:
+- language_provider: io.gapi.vgen.java.JavaGapicCodeLanguageProvider
+  snippet_files:
+  - main.snip
+  - settings.snip
+- language_provider: io.gapi.vgen.java.JavaGapicFragmentLanguageProvider
+  snippet_files:
+  - fragment.snip
```
garrettjonesgoogle: This new structure implies to me that when I run the GAPIC codegen for Java, it will generate fragments too, which would be a change in behavior and not what we want. Have you tried running the artman pipeline for Java GAPIC to make sure it doesn't generate extra stuff?

vgen/src/main/java/io/gapi/vgen/CodeGenerator.java
```diff
@@ -40,36 +32,40 @@
...
 
   /**
    * Generates code for the model. Returns a map from service interface to code for the service.
    * Returns null if generation failed.
    */
   @Nullable
-  public Map<Interface, GeneratedResult> generate(SnippetDescriptor snippetDescriptor) {
+  public Map<ProtoElement, GeneratedResult> generate(SnippetDescriptor snippetDescriptor) {
+    ProtoElementViewing<ProtoElement> viewing = provider.getViewing();
+    Iterable<ProtoElement> elts = viewing.getIterable(provider.getModel());
```
garrettjonesgoogle: The term 'Viewing' is slightly odd to me. Why not just 'View'? Also, I'm not so sure the name needs 'ProtoElement' in it, since it is parameterized with the type 'ProtoElement' anyway. At a higher level, my understanding is that this ProtoElementViewing is used to acquire the elements that serve as the input for codegen.

With all that, maybe a better name might be 'InputElementProvider' or 'InputElementFilter' (it has to "filter" the proto elements from the model).

vgen/src/main/java/io/gapi/vgen/csharp/CSharpGapicCodeLanguageProvider.java
```diff
@@ -0,0 +1,34 @@
...
+package io.gapi.vgen.csharp;
+
+import com.google.api.tools.framework.model.Interface;
+import com.google.api.tools.framework.model.Model;
+
+import io.gapi.vgen.ApiConfig;
+
+public class CSharpGapicCodeLanguageProvider extends CSharpGapicLanguageProvider<Interface> {
+
+  private final CSharpInterfaceViewing viewing = new CSharpInterfaceViewing();
```
garrettjonesgoogle: To me it doesn't really make sense to make a subclass purely for returning a different concrete Viewing that the parent class uses. This approach has created a bunch of trivial concrete classes. I think that the correct viewing should just be created at instantiation time and passed into the existing <Language>GapicLanguageProvider class, which should hold the instance for itself.

garrettjonesgoogle: Ok I think I have figured out a little bit more of the intent here - you are preserving the property of the config files where only one class is specified, and that fully determines all of the behavior. This unfortunately creates the need to have a distinct class for every config file, though. I think a more flexible approach would be for the config file to specify two classes: the language provider (limited to the existing ones before this PR) plus a "ProtoElementViewing" (or whatever it ends up being called, per my other comment) class, which are combined together for the desired behavior.

vgen/src/main/java/io/gapi/vgen/csharp/CSharpMethodViewing.java
```diff
@@ -0,0 +1,29 @@
...
+import com.google.api.tools.framework.model.ProtoFile;
+
+import io.gapi.vgen.MethodViewing;
+
+public class CSharpMethodViewing extends MethodViewing
+    implements CSharpProtoElementViewing<Method> {
+
+  @Override
+  public ProtoFile getNamespaceFile(Method method) {
+    return method.getParent().getFile();
```
garrettjonesgoogle: This functionality seems generic enough to just lift up to ProtoElementViewing for use by any language, although slightly renamed, maybe 'getElementSourceFile'

vgen/src/main/java/io/gapi/vgen/GapicLanguageProvider.java
garrettjonesgoogle: After reading more of the code, it looks like the type might best be 'InputElementT' and the variable might best be 'element' or 'inputElement'.

vgen/src/main/resources/io/gapi/vgen/java/java_gapic.yaml
garrettjonesgoogle: Thinking about this more, I'm not sure why the GAPIC fragment generation isn't in its own file, similar to how the discovery fragment generation is it is own file. But then, I'm not even sure what process makes use of this.

garrettjonesgoogle:
I have finished my first pass. I like where this is heading, but I think it could be make more modular (specifically, my suggestion to specify multiple classes in the config files instead of just one).

vgen/src/main/java/io/gapi/vgen/csharp/CSharpGapicCodeLanguageProvider.java
geigerj: I agree that having <Language>x<ProtoElement>Viewing isn't ideal, since there are a large number of possible combinations. (Note, however, that the languages other than C# and Python don't have any special per-protoElement-subclass behavior, so there is no need for any JavaXXXViewing.) However, I'm not sure specifying two classes in the config file helps this situation -- wouldn't we still need to have this file?

geigerj: Ah, never mind. I see what you are saying.

garrettjonesgoogle: Yes, we would still need the file, but then the combinations would be in only one place, and we wouldn't need a class corresponding to each combination. With this current PR, you need both the config file and the new class for each new combination.

pongad:
I think I'm missing context on this change, but the changes to discovery looks good to me.

vgen/src/main/java/io/gapi/vgen/CodeGenerator.java
geigerj: Let's go with InputElementProvider -- it's nicely parallel with LanguageProvider once they're both in the config.

geigerj: Just kidding. Having a bunch of semi-unrelated files called *Provider is actually pretty confusing. Let's do InputElementView?

bjwatson: The "ing" part was my fault. It was a convention I used with Scala typeclasses, which are a combination of the Strategy pattern, the Decorator pattern, and a bit of implicit magic.

I agree it seems weird here is the context of a plain vanilla Java Strategy pattern.

The View part is a reference to SQL views: a pre-defined filtering of the model.

vgen/src/main/java/io/gapi/vgen/DiscoveryFragmentGenerator.java
bjwatson: Great idea.

vgen/src/main/java/io/gapi/vgen/csharp/CSharpGapicCodeLanguageProvider.java
bjwatson: Good point. Jacob and I discussed, and we agree to referencing two classes in the config file for each language-template.

vgen/src/main/java/io/gapi/vgen/csharp/CSharpMethodViewing.java
bjwatson: It doesn't seem generic to me right now. Every language seems to have its own custom code for implementing `GapicLanguageProvider.generate()`. I would prefer not to put a method into ProtoElementViewing / InputElementView unless it is used in every subclass.

vgen/src/main/java/io/gapi/vgen/py/PythonLanguageProvider.java
bjwatson: good catch

vgen/src/main/resources/io/gapi/vgen/java/java_gapic.yaml
bjwatson: I'm not sure either, but changing this would be out-of-scope for this PR. The fragment file was previously declared here, and we're just changing how it's done. I'm ramping up on vKit snippet gen for Python, and hope to better understand the use cases in a few days.

geigerj: When this change is pushed, artman _will_ generate fragments for Java when you run GAPIC. However, I think it is better to break up the configuration files than to change toolkit.

bjwatson:
@pongad The primary motivation is to constrain the number of unique keys in the YAML files (rather than creating a new one for every template type). It also reduces some code duplication.

vgen/src/main/resources/io/gapi/vgen/java/java_gapic.yaml
garrettjonesgoogle: @bjwatson , by including the fragment file in GAPIC generation, you are changing the behavior, and in an undesirable way (from the perspective of the Java codegen). I'd much prefer the behavior to stay the same with this refactoring.

vgen/src/main/java/io/gapi/vgen/CodeGenerator.java
garrettjonesgoogle: InputElementView sounds good to me.

vgen/src/main/java/io/gapi/vgen/csharp/CSharpMethodViewing.java
garrettjonesgoogle: I wouldn't necessarily make the rule "every" subclass - I'd probably go for "most subclasses, and 'feels' generic in nature." This one seems like a gray area to me - it's one subclass, but generic in nature. In any case, I will defer here.

vgen/src/main/resources/io/gapi/vgen/java/java_gapic.yaml
geigerj: This is tough. I see one of the key points of this refactoring is that code/doc/fragment/client config generation is -not- distinguished by any special logic in the toolkit itself. I don't see an easy way to make this work, other than re-introducing the concept of ["fragment generation" == "method view"] as a special case in CodeGenerator.

garrettjonesgoogle: @pongad reports that this isn't actually used anywhere yet. Vkit snippet generation (snippets within comments) is performed by the main Gapic codegen. I think we should either remove this fragment-gen config or put it in its own file - the process which would make use of this would run independently of GAPIC and its output would be used by different tools.

bjwatson: Sorry @garrettjonesgoogle. I misunderstood that aspect of our refactoring. Now that I look more closely, I see that you're right that we changed `CodeGeneratorApi` to also call the vKit `generateFragments` logic.

vgen/src/main/java/io/gapi/vgen/csharp/CSharpMethodViewing.java
bjwatson: You're right that the softer rule is okay, and it's plausible that we can define a namespace file for (almost) every language-template tuple. However, I don't see the use cases for this explicitly called out in the code right now. Only C# has a clear need for it at the moment.

garrettjonesgoogle: Well my suggestion was to take the 'Namespace' out of the name; that's just the context for how it is used in C#.  As I mentioned, I think it should be called 'getElementSourceFile'. The method could also be used by Python here:

https://github.com/googleapis/toolkit/blob/master/vgen/src/main/java/io/gapi/vgen/py/PythonGapicLanguageProvider.java#L63

The type knowledge embedded in the instanceof check there is known inside the ProtoElementViewing / InputElementView instance. 

Generally I could see this useful for a snippet to report what file it was generated from. We use "service.getFile()" in nearly every snippet, and this is arguably a more correct / more general way to provide that information.

bjwatson: Ok, I see what you mean. When I search from calls to ProtoElement.getFile(), I see it used for every language except Node.js (although it might not be implemented there yet). It appears to be in just the Context object for half of them (Go, Java, and PHP), but maybe this could leverage what you're suggesting here instead.

@geigerj Do you think it's worth implementing this method in InputElementView for all languages (except maybe Node.js) right now? I'm not talking about fixing all calls to ProtoElement.getFile() to use InputElementView.getElementSourceFile() just yet. That seems like another pass of refactoring.

geigerj: Yes, I think it is worth implementing this method in InputElementView, but I would rather file this as an issue for another round of refactoring. This PR is already very large, and as much as possible I'd like to avoid expanding the scope.

geigerj:
FYI: I will address feedback in a few successive commits, and then squash them at the end before merging. Hopefully that'll help keep track of what's changed in between feedback comments.

vgen/src/main/java/io/gapi/vgen/CodeGenerator.java
geigerj: Done.

vgen/src/main/java/io/gapi/vgen/csharp/CSharpGapicCodeLanguageProvider.java
geigerj: Done.

vgen/src/main/java/io/gapi/vgen/csharp/CSharpMethodViewing.java
garrettjonesgoogle: I can live with that, but keep in mind that as a consequence, you are creating a new class in this PR and then deleting the class in the following PR, because the only reason this class exists is to provide getNamespaceFile. (So  I would argue that my suggestion reduces the scope... but I can't keep my mouth shut :-P )

vgen/src/main/java/io/gapi/vgen/CodeGenerator.java
geigerj: Done

vgen/src/main/java/io/gapi/vgen/DiscoveryFragmentGenerator.java
geigerj: Done

vgen/src/main/java/io/gapi/vgen/GapicLanguageProvider.java
geigerj: Done

vgen/src/main/java/io/gapi/vgen/ProtoElementViewing.java
geigerj: Done

vgen/src/main/java/io/gapi/vgen/ProtoFileViewing.java
geigerj: Done

vgen/src/main/java/io/gapi/vgen/py/PythonLanguageProvider.java
geigerj: Done

vgen/src/main/resources/io/gapi/vgen/java/java_gapic.yaml
geigerj: Done.

(I moved the Java GAPIC fragment config out of main config and updated the test.)

geigerj:
Updated based on feedback, PTAL

vgen/src/main/java/io/gapi/vgen/CodeGenerator.java
```diff
@@ -40,36 +31,42 @@
...
   }
 
-  public static CodeGenerator create(ConfigProto configProto, Model model) {
-    return new CodeGenerator(GeneratorBuilderUtil.createLanguageProvider(configProto, model));
+  public static CodeGenerator create(ConfigProto configProto, TemplateProto template, Model model) {
+    InputElementView<ProtoElement> view = GeneratorBuilderUtil.createView(template, model);
+    GapicLanguageProvider<ProtoElement> provider =
+        GeneratorBuilderUtil.createLanguageProvider(configProto, template, model, view);
+    return new CodeGenerator(
+        GeneratorBuilderUtil.createLanguageProvider(configProto, template, model, view));
```
garrettjonesgoogle: You are creating a provider twice, and throwing the first one away.

vgen/src/main/java/io/gapi/vgen/DiscoveryFragmentGeneratorApi.java
```diff
@@ -101,7 +101,7 @@ protected void process() throws Exception {
     }
 
     Multimap<Method, GeneratedResult> docs = ArrayListMultimap.create();
-    for (String snippetInputName : configProto.getFragmentFilesList()) {
+    for (String snippetInputName : configProto.getTemplates(0).getSnippetFilesList()) {
```
garrettjonesgoogle: Also add the precondition check before this line for the number of templates.

vgen/src/main/java/io/gapi/vgen/GeneratorBuilderUtil.java
```diff
@@ -36,12 +38,24 @@ private GeneratorBuilderUtil() {}
...
+  /**
+   * Constructs a language provider from a configuration; the configuration cannot contain
+   * parameterized type information, so we must suppress the type safety warning here.
+   */
+  @SuppressWarnings("unchecked")
+  public static GapicLanguageProvider<ProtoElement> createLanguageProvider(
+      ConfigProto configProto,
+      TemplateProto template,
+      final Model model,
+      final InputElementView<ProtoElement> view) {
```
garrettjonesgoogle: The `view` parameter technically doesn't need to be final. (Not necessarily something that needs to be addressed, just calling it out)

vgen/src/main/java/io/gapi/vgen/csharp/CSharpGapicLanguageProvider.java
```diff
@@ -34,14 +35,19 @@
 /**
  * The LanguageProvider which runs Gapic code generation for C#.
  */
-public class CSharpGapicLanguageProvider implements GapicLanguageProvider {
+public class CSharpGapicLanguageProvider<InputElementT extends ProtoElement>
+    implements GapicLanguageProvider<InputElementT> {
 
   private final CSharpGapicContext context;
   private final CSharpLanguageProvider provider;
+  private CSharpProtoElementView<InputElementT> view;
```
garrettjonesgoogle: view can be made final

garrettjonesgoogle:
Just a couple minor comments on this round, but nothing blocking - take care of the ones you want to. LGTM.

vgen/src/main/java/io/gapi/vgen/GeneratorBuilderUtil.java
bjwatson: Isn't it generally good to make as many variables `final` as possible, to avoid accidental mutation?

garrettjonesgoogle: Well there isn't really much consequence to mutating function parameters, except for slightly hindering readability.

geigerj: I buy that it's generally a good idea to make parameters that don't need to be mutated `final` for readability, but I'm removing the `final` here for consistency with the other non-model parameters.

vgen/src/main/java/io/gapi/vgen/DiscoveryFragmentGeneratorApi.java
geigerj: Done

vgen/src/main/java/io/gapi/vgen/CodeGenerator.java
geigerj: Done

vgen/src/main/java/io/gapi/vgen/csharp/CSharpGapicLanguageProvider.java
geigerj: Done

geigerj:
Will merge when the change to googleapis (googleapis/googleapis#23) is approved

PR #75, opened 2016-05-09
Change version of clouduseraccounts to beta

pongad:


vchudnov-g:
Waiting to resolve comment in https://github.com/googleapis/toolkit/pull/69 about the need to bump versions.

vchudnov-g:
LGTM, but please include the same explanatory message as in #69

pongad:
I will close this PR and move its content into #69 , so that the explanation will end up in the same place.

PR #76, opened 2016-05-09
Provide support for `x-goog-api-client` header

### LINKED ISSUE #35: Support `x-goog-api-client` in go
Using this document to get opinions before a PR:

Background
=========
We plan to support an `X-Goog-Api-Client` header that allows us to track the version information of the code generator and GAX libraries for each client call.

The format is `<clientlib>/<semver> gapic/<semver> gax/<semver> language/<semver>`
For example, in python: `gcloud-python/<gcloud-python-version> gapic/0.1.0 gax/0.5.0 python/2.7`

Proposal
=======
In order to support the header, we need to be able to pass optional `app_name` and `app_version` strings to the client struct on initialization.

The method signature for `NewClient` is currently:
`func NewClient(ctx context.Context, opts ...cloud.ClientOption) (*Client, error)`

We propose instead:
`func NewClient(ctx context.Context, opts ...gax.ClientOption) (*Client, error)`
where the interface `gax.ClientOption` is provided by the following:

within `gax-golang`
-> `WithCloudClientOption(cloud.ClientOption) ClientOption`
-> `WithAppVersion(appVersion string) ClientOption`
-> `WithAppName(appName string) ClientOption`

This design also allows us to implement client level options specific to the gapic surface:
-> `WithCallOptionsOverride(map[string][]gax.CallOption) ClientOption`
-> `WithTimeout(duration time.Duration) ClientOption`

Now, within `NewClient` we can generate the header string as a private global variable and provide it to all calls via `grpc.CallOption`.

GAX will also support a global `Version string` that contains a semantic version tag. It's a workaround for go's package versioning limitations, and not perfect, but hopefully sufficient for now.
@tbetbetbe's suggestion was to use a script to generate a go file with a version tag (maybe the commit hash, or github release version).

saicheems:
Also refactors the `ClientSettings` implementation.

Resolves #35

### DIFF
--- a/src/main/java/com/google/api/codegen/go/GoGapicContext.java
+++ b/src/main/java/com/google/api/codegen/go/GoGapicContext.java
@@ -357,10 +357,13 @@ public class GoGapicContext extends GapicContext implements GoContext {
     TreeSet<GoImport> coreImports = new TreeSet<>();
     TreeSet<GoImport> imports = new TreeSet<>();
 
+    coreImports.add(GoImport.create("fmt"));
+
     // Add non-service-specific imports.
     imports.add(GoImport.create("golang.org/x/net/context"));
     imports.add(GoImport.create("google.golang.org/grpc"));
     imports.add(GoImport.create("google.golang.org/grpc/codes"));
+    imports.add(GoImport.create("google.golang.org/grpc/metadata"));
     imports.add(GoImport.create(GAX_PACKAGE_BASE, "gax"));
 
     if (!getApiConfig().getInterfaceConfig(service).getRetrySettingsDefinition().isEmpty()) {
--- a/src/main/resources/com/google/api/codegen/go/main.snip
+++ b/src/main/resources/com/google/api/codegen/go/main.snip
@@ -60,6 +60,9 @@
 
         // The map from the method name to the default call options for the method of this service.
         callOptions map[string][]gax.CallOption
+
+        // The metadata to be sent with each request.
+        metadata map[string]string
     }
 
     {@constructor(service)}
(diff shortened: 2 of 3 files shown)


saicheems:
@jba 
@pongad 
@jmuk 
@vchudnov-g 

PTAL: This change modifies the way default client settings/call settings are provided to reduce some verbosity. Also implements the client header.

Relies on changes in GAX that haven't been pushed, but are fairly innocuous - mostly making private structs public.

vgen/src/test/java/io/gapi/vgen/testdata/go_library.baseline
```diff
@@ -56,41 +57,98 @@ import (
...
+        AppVersion: gax.Version(),
+        Endpoint: "library-example.googleapis.com:443",
+        Scopes: []string{
+            "https://www.googleapis.com/auth/library",
+            "https://www.googleapis.com/auth/cloud-platform",
+        },
+        CallSettings: map[string][]gax.CallOption{
+            "CreateShelf": []gax.CallOption{
+                gax.WithRetryCodes(nonIdempotentRetryCodes()),
+                defaultRetryOptions()...
```
jmuk: I think this is invalid syntactically. Instead, probably
```
append([]gax.CallOption(gax.WithRetryCodes(nonIdempotentRetryCodes()), defaultRetryOptions()...)
```
?

See the compile errors of https://play.golang.org/p/HptLq_gC7j FYI.

saicheems: You're right - good catch
Fixed.

jmuk:
LGTM

pongad:
lgtm

PR #78, opened 2016-05-09
Fix Go Discovery Output Directory

pongad:
GoDiscoveryLanguageProvider did not honor the file heirarchy convention

autogenerated/api-name/api-version/api-revision/method-name.frag.go

This change fixes this buggy behavior.


vchudnov-g:
LGTM

PR #79, opened 2016-05-10
[discovery] Configuration not found should be an error

pongad:
Previously when we cannot read from the configuration file,
an error message is printed to stderr and we keep going.
This commit changes this behavior so we will just fail out.


vchudnov-g:
LGTM.

Are there any other types of error where we continue when we should be stopping? (It might be worth doing an audit for "cannot"...)

pongad:
After some audit with "cannot" "can't" "error", etc it looks like this is the only one we have. Of course, we'll be fixing them if we find them later.

PR #87, opened 2016-05-10
Add method samples for go

saicheems:


saicheems:
@pongad 
@vchudnov-g

vgen/src/main/java/io/gapi/vgen/go/GoGapicContext.java
```diff
@@ -168,6 +168,17 @@ public String typeName(TypeRef type) throws RuntimeException {
   }
 
   /**
+   * Returns the (dereferenced) Go type name for the specificed message TypeRef.
+   */
+  public String messageTypeName(TypeRef type) throws RuntimeException {
```
pongad: No need to declare the throws

vgen/src/main/java/io/gapi/vgen/go/GoGapicContext.java
```diff
@@ -168,6 +168,17 @@ public String typeName(TypeRef type) throws RuntimeException {
...
+      if (!type.isMessage()) {
+        throw new RuntimeException("Expected message type: " + type.toString());
```
pongad: Consider using IllegalArgumentException instead

vgen/src/test/java/io/gapi/vgen/testdata/go_library.baseline
```diff
@@ -610,6 +610,118 @@ func (iterator *StringIterator) Next() (string, error) {
...
+    opts := []gax.ClientOption{ /* Optional client parameters. */ }
+    client, err := library_service.NewClient(ctx, opts...)
+}
+
+func ExampleCreateShelf() {
+    ctx := context.Background()
+    req := &google_example_library_v1.CreateShelfRequest{ /* Data... */ }
+    opts := []gax.CallOption{ /* Optional call parameters. */ }
+    var resp *google_example_library_v1.Shelf
+    resp, err = client.CreateShelf(ctx, req, opts...)
```
pongad: Have you tried `go test`ing this? I'm not sure if these samples compile.
- You probably need to do something like `_, _ = resp, err` to avoid compilation error
- Where is `err` defined here?

saicheems:
Still a WIP, will tag you guys again when ready
Noticed some faults

vgen/src/main/java/io/gapi/vgen/go/GoGapicContext.java
saicheems: Fixed

vgen/src/main/java/io/gapi/vgen/go/GoGapicContext.java
saicheems: Done

saicheems:
@vchudnov-g 
@pongad 

retagged

vgen/src/test/java/io/gapi/vgen/testdata/go_library.baseline
saicheems: Fixed

vgen/src/main/java/io/gapi/vgen/go/GoGapicContext.java
```diff
@@ -401,6 +414,40 @@ private GoImport createMessageImport(MessageType messageType) {
...
+          imports.add(createMessageImport(resourceType.getMessageType()));
+        }
+      }
+    }
+    List<String> result = new ArrayList<>();
+    for (GoImport goImport : imports) {
+      result.add("    " + goImport.importString());
+    }
+    return result;
+  }
```
pongad: Seems to be a lot of duplication with `getImports`. Do you think it's worth refactoring so we only need to write the `// Add method request-type imports` section only once?

vgen/src/main/resources/io/gapi/vgen/go/example.snip
```diff
@@ -0,0 +1,46 @@
...
+        ctx := context.Background()
+        opts := []gax.ClientOption{ /* Optional client parameters. */ }
+        _, _ = {@context.upperCamelToLowerUnderscore(service.getSimpleName)}.NewClient(ctx, opts...)
+    }
+    @join method : service.getMethods
+        @let methodName = method.getSimpleName, \
+             inTypeName = context.messageTypeName(method.getInputType), \
+             outTypeName = context.typeName(method.getOutputType), \
+             methodConfig = context.getApiConfig.getInterfaceConfig(service).getMethodConfig(method), \
+             isPageStreaming = methodConfig.isPageStreaming
```
pongad: Are we doing anything with this? Do we want to give example of a loop if the method page-streams?

vgen/src/main/java/io/gapi/vgen/go/GoGapicContext.java
saicheems: Yeah it is, I'll refactor it

vgen/src/main/resources/io/gapi/vgen/go/example.snip
saicheems: Yeah I wasn't sure before... it sounds like a good idea to do so. Will include

saicheems:
Updated, PTAL

saicheems:
I can confirm that it compiles now

src/main/java/com/google/api/codegen/go/GoGapicContext.java
```diff
@@ -345,64 +357,102 @@ private GoImport createMessageImport(MessageType messageType) {
...
+   */
+  public Iterable<String> getTestImports(Interface service) {
+    TreeSet<GoImport> thirdParty = new TreeSet<>();
+
+    thirdParty.add(GoImport.create("golang.org/x/net/context"));
+    thirdParty.add(GoImport.create(GAX_PACKAGE_BASE, "gax"));
+    thirdParty.addAll(getMessageImports(service));
+
+    return formatImports(null, thirdParty);
+  }
```
pongad: You can avoid null by giving an empty list.

saicheems: One motivation for the null is it prevents a newline space before the thirdParty imports.
I can modify the format function to ignore the newline if the standard imports set is empty, lmk if that's preferred.

src/test/java/com/google/api/codegen/testdata/go_library.baseline
```diff
@@ -621,6 +621,181 @@ func (iterator *StringIterator) Next() (string, error) {
...
+func ExampleNewClient() {
+    ctx := context.Background()
+    opts := []gax.ClientOption{ /* Optional client parameters. */ }
+    client, err := NewClient(ctx, opts...)
+    _, _ = client, err
+}
+
+func ExampleClient_CreateShelf() {
+    ctx := context.Background()
+    client, _ := NewClient(ctx)
```
pongad: Users can't copy this code because they will have to actually write `library_service.NewClient`. Consider repackaging the test file into `library_service_test` then import `library_service`.

pongad:
Almost there!

src/test/java/com/google/api/codegen/testdata/go_library.baseline
saicheems: I'll make a note of this... trick is that it's not immediately available to the generator what the import path of the service is.
Will fix in a later follow up - still have to figure out a few things about the imports.

src/main/java/com/google/api/codegen/go/GoGapicContext.java
pongad: OK this makes sense. It's not too big of a problem in either case.

pongad:
LGTM

jba:
Hope I'm not too late: the file with the examples should be named
xxx_test.go, but its package should the same as xxx.go. No need to import.
(That way you can also test unexported parts of the package.)

saicheems:
Not too late! That's what was decided. Thanks for the confirmation!

PR #90, opened 2016-05-10
Added request object construction to doc gen

michaelbausor:
Construct request object in documentation
Updated tests

### DIFF
--- a/vgen/src/main/java/io/gapi/vgen/java/JavaContextCommon.java
+++ b/vgen/src/main/java/io/gapi/vgen/java/JavaContextCommon.java
@@ -221,37 +221,36 @@ public class JavaContextCommon {
 
       public abstract Builder setParams(ImmutableList<Variable> params);
 
-      public Builder setParams(JavaGapicContext languageProvider, Iterable<Field> fields) {
-        return setParams(fieldsToParams(languageProvider, fields));
+      public Builder setParams(JavaGapicContext context, Iterable<Field> fields) {
+        return setParams(fieldsToParams(context, fields));
       }
 
       public Builder setParamsWithFormatting(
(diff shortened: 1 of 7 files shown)


vgen/src/main/java/io/gapi/vgen/java/JavaContextCommon.java
```diff
@@ -212,7 +221,51 @@ public static Variable s_newVariable(
 
       public abstract Builder setParams(ImmutableList<Variable> params);
 
-      public Builder setParams(JavaGapicContext languageProvider, ImmutableList<Field> fields) {
+      public Builder setParams(JavaGapicContext languageProvider, Iterable<Field> fields) {
```
garrettjonesgoogle: `languageProvider` is no longer the correct name for the parameter - it should be `context`

vgen/src/main/resources/io/gapi/vgen/java/main.snip
```diff
@@ -626,14 +636,58 @@
...
 @private initValue(param, ApiName)
-  @if param.hasFormattingConfig() {
+  @if param.hasFormattingConfig()
     {@ApiName}.{@formatResourceFunctionName(param.getFormattingConfig())}({@formatResourceFunctionArgs(param.getFormattingConfig())})
   @else
     {@context.zeroValue(param.getType)}
   @end
 @end
 
+@snippet setCallNameFromParam(param)
```
garrettjonesgoogle: I would prefer not to repeat the structure of setCallName in common.snip.

michaelbausor:
PTAL

garrettjonesgoogle:
LGTM

PR #91, opened 2016-05-11
Add Node Discovery Samples Prerequisites

pongad:
This change add instructions for the users to set up auth
and install googleapis so that they can start using NodeJS client lib.


garrettjonesgoogle:
LGTM

PR #92, opened 2016-05-11
Package and class renaming

garrettjonesgoogle:
For https://github.com/googleapis/toolkit/issues/22. 

* move vgen/* to the root /
* rename "io.gapi.vgen" to "com.google.api.codegen"
* rename "io.gapi.vsync" to "com.google.api.codegen.sync"
* rename &lt;prefix&gt;LanguageProvider to just &lt;prefix&gt;Provider
 * e.g. PythonGapicLanguageProvider => PythonGapicProvider
 * reasons:
   * In all of these class names, the name of the language (e.g. Python) is redundant with the Language part of the name.
    * "LanguageProvider" as a parent concept doesn't exist as a class.
    * A lot of class names will be shorter as a result.
* rename the "language provider" concept to "codegen provider"
* rename LanguageContext to CodegenContext
* rename "&lt;language&gt;_discovery_gapic.yaml" to "&lt;language&gt;_discovery.yaml"
 * distinguishes discovery vs GAPIC


garrettjonesgoogle:
@bjwatson also

geigerj:
LGTM

src/main/java/com/google/api/codegen/DiscoveryContext.java
```diff
@@ -27,15 +26,15 @@
...
  */
-public abstract class DiscoveryContext extends LanguageContext {
+public abstract class DiscoveryContext extends CodegenContext {
 
   private final Service service;
   private final ApiaryConfig apiaryConfig;
 
   /**
-   * Constructs the abstract instance of the language provider.
+   * Constructs an abstract instance.
```
bjwatson: This is rather terse and unhelpful, although the previous version wasn't much better.

garrettjonesgoogle: The goal here was to turn a negative (incorrect comments) into a neutral (unhelpful comments). I slightly increase the scope of https://github.com/googleapis/toolkit/issues/62 to cover this issue.

src/main/java/com/google/api/codegen/csharp/CSharpGapicContext.java
```diff
@@ -88,7 +89,7 @@
   private CSharpContextCommon csharpCommon;
 
   /**
-   * Constructs the C# language provider.
+   * Constructs an instance.
```
bjwatson: This needs to be improved. The previous version was better, if not exactly accurate.

src/main/java/com/google/api/codegen/go/GoGapicContext.java
```diff
@@ -102,7 +101,7 @@
   private final GoContextCommon goCommon;
 
   /**
-   * Constructs the Go language provider.
+   * Constructs an instance.
```
bjwatson: This also needs to be improved.

geigerj:
Note that the old constructor docs were copy-pasted into the wrong files during the discovery merge -- they're inaccurate. I'm fine with leaving the unhelpful docs for a future PR, since it's not strictly related to the renaming.

bjwatson:
@garrettjonesgoogle The code changes all look fine to me. I'm mainly just finding Javadoc issues, which I know aren't in great shape anyway. I suggest that if we have nothing meaningful to say about a class or method right now, we just delete the Javadoc. That way we don't miss it when we go through and add Javadoc everywhere that it is missing per https://google.github.io/styleguide/javaguide.html#s7-javadoc.

bjwatson:
Tagging https://github.com/googleapis/toolkit/issues/62, which is relevant to the Javadoc issue.

src/main/java/com/google/api/codegen/DiscoveryContext.java
garrettjonesgoogle: I have removed the unhelpful javadocs.

bjwatson:
LGTM

PR #95, opened 2016-05-11
Unflake verifyLicense

pongad:
Previously verifyLicense looks for license-header-javadoc.txt by
looking in the process's working directory. This will most likely
work if the build is manually run.
When run by other programs such as an IDE, the working directory might
be different, causing the check to fail.

This commit looks for the license header within the project's root
directory instead. This should work generally.


geigerj:
LGTM

PR #96, opened 2016-05-11
Add client config generation

geigerj:
@tbetbetbe @bjwatson

### DIFF
--- a/src/main/java/com/google/api/codegen/clientconfig/ClientConfigGapicProvider.java
+++ b/src/main/java/com/google/api/codegen/clientconfig/ClientConfigGapicProvider.java
@@ -1,0 +1,91 @@
+/* Copyright 2016 Google Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen.clientconfig;
+
+import com.google.api.codegen.ApiConfig;
+import com.google.api.codegen.GapicProvider;
+import com.google.api.codegen.GeneratedResult;
+import com.google.api.codegen.InputElementView;
+import com.google.api.codegen.SnippetDescriptor;
+import com.google.api.tools.framework.model.Interface;
+import com.google.api.tools.framework.model.Model;
+import com.google.api.tools.framework.snippet.Doc;
+import com.google.api.tools.framework.snippet.SnippetSet;
+import com.google.api.tools.framework.tools.ToolUtil;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Multimap;
+
+import java.io.IOException;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+public class ClientConfigGapicProvider implements GapicProvider<Interface> {
+
+  private Model model;
+  private ApiConfig apiConfig;
+  private InputElementView<Interface> view;
+
+  public ClientConfigGapicProvider(
+      Model model, ApiConfig apiConfig, InputElementView<Interface> view) {
+    this.model = model;
+    this.apiConfig = apiConfig;
+    this.view = view;
+  }
+
+  @Override
+  public Model getModel() {
+    return model;
+  }
+
+  @Override
+  public <Element> void output(String outputPath, Multimap<Element, GeneratedResult> elements)
+      throws IOException {
+    Map<String, Doc> files = new LinkedHashMap<>();
+    for (GeneratedResult generatedResult : elements.values()) {
+      files.put(generatedResult.getFilename(), generatedResult.getDoc());
+    }
+
+    ToolUtil.writeFiles(files, outputPath);
+  }
+
+  /**
+   * The path to the root of snippet resources.
+   */
+  private static final String SNIPPET_RESOURCE_ROOT =
+      ClientConfigGapicProvider.class.getPackage().getName().replace('.', '/');
+
+  @SuppressWarnings("unchecked")
+  @Override
+  public GeneratedResult generate(Interface service, SnippetDescriptor snippetDescriptor) {
+    ClientConfigSnippetSet<Interface> snippets =
+        SnippetSet.createSnippetInterface(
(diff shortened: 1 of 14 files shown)


src/main/java/com/google/api/codegen/clientconfig/ClientConfigGapicProvider.java
```diff
@@ -0,0 +1,92 @@
...
+        SnippetSet.createSnippetInterface(
+            ClientConfigSnippetSet.class,
+            SNIPPET_RESOURCE_ROOT,
+            snippetDescriptor.getSnippetInputName(),
+            ImmutableMap.<String, Object>of(
+                "context", new ClientConfigGapicContext(model, apiConfig)));
+
+    String outputFilename = snippets.generateFilename(service).prettyPrint();
+
+    // Generate the body, which will collect the imports.
```
garrettjonesgoogle: This comment doesn't make sense for this context

geigerj: Whoops, fixed.

src/main/resources/com/google/api/codegen/clientconfig/client_gapic.yaml
```diff
@@ -0,0 +1,6 @@
+type: com.google.api.codegen.ConfigProto
+templates:
+- codegen_provider: com.google.api.codegen.clientconfig.ClientConfigGapicProvider
+  input_element_view: com.google.api.codegen.InterfaceView
+  snippet_files:
+  - main.snip
```
garrettjonesgoogle: I think it would be nice to set up this client config stuff so that other output formats could be included too, e.g. xml for C#. What classes/files would you have to add `Json` to later if we later added support for generating an xml client config file? I would consider adding `Json` to those things now.

src/test/java/com/google/api/codegen/PythonCodeGeneratorTest.java
```diff
@@ -67,6 +69,10 @@ protected Object run() {
       for (GeneratedResult result : docsResult) {
         builder.put(result.getFilename(), result.getDoc());
       }
+      for (GeneratedResult result : configResult) {
+        System.out.println(result.getFilename());
```
garrettjonesgoogle: Did you want to keep this println or is it just for debugging?

src/test/java/com/google/api/codegen/testdata/python_no_path_templates.baseline
```diff
@@ -151,3 +151,22 @@ class NoTemplatesServiceApi(object):
             self.stub.Increment, settings=settings)
         api_call(req, metadata=self._headers)
 
+============== file: no_templates_service_client_config.json ==============
+{
+  "interfaces": {
+    "google.example.v1.NoTemplatesService": {
+      "retry_codes": {
+        retry_codes_def:
```
garrettjonesgoogle: This structure doesn't look right here

src/test/java/com/google/api/codegen/testdata/ruby_library.baseline
```diff
@@ -1292,3 +1292,91 @@ module Google
...
+        },
+        "create_book": {
+          "retry_codes_name": "non_idempotent",
+          "retry_params_name": "default"
+        },
+        "publish_series": {
+          "retry_codes_name": "non_idempotent",
+          "retry_params_name": "default",
+          "bundling": {
+              "element_count_threshold": 6,
```
garrettjonesgoogle: Somehow this got extra-indented as compared to everything else.

src/main/resources/com/google/api/codegen/clientconfig/client_gapic.yaml
geigerj: I think you might rename main.snip to json.snip, and add an xml.snip. I doubt the Java classes would need to change.

garrettjonesgoogle: That's kind of what I was thinking - it didn't look like this scenario needed much specialized bootstrapping or context support in comparison to other cases. Could we rename main.snip to json.snip? (or maybe main_json.snip?)

src/test/java/com/google/api/codegen/testdata/ruby_library.baseline
geigerj: Done

src/test/java/com/google/api/codegen/testdata/python_no_path_templates.baseline
geigerj: How so? This test has a sort of degenerate GAPIC config, so I think something like this might be expected...

src/test/java/com/google/api/codegen/PythonCodeGeneratorTest.java
geigerj: Mistake -- fixed.

src/main/resources/com/google/api/codegen/clientconfig/client_gapic.yaml
geigerj: Done (renamed to json.snip)

src/test/java/com/google/api/codegen/testdata/python_no_path_templates.baseline
geigerj: Actually, I'm going to undo my change to no_path_templates_gapic.yaml. I see this as a useful test of "does this output valid json even if the input is incomplete?", and I think json.snip still needs a few "if" guards to make that work.

garrettjonesgoogle: Wouldn't you need this?

```
     "retry_codes": {
       retry_codes_def: {
         "non_idempotent": []
       }
     },
```

i.e. you are missing braces.

geigerj: Ah, I see now. Fixed.

src/main/resources/com/google/api/codegen/py/python_gapic.yaml
```diff
@@ -9,3 +9,7 @@ templates:
   input_element_view: com.google.api.codegen.py.PythonProtoFileView
   snippet_files:
   - message.snip
+- codegen_provider: com.google.api.codegen.clientconfig.ClientConfigGapicProvider
+  input_element_view: com.google.api.codegen.InterfaceView
+  snippet_files:
+  - json.snip
```
garrettjonesgoogle: Something we might want to consider addressing (separate from this PR): now that we're mixing providers from different directories, it could be somewhat confusing to find the right snippet file in the source tree corresponding to the provider, because the rule connecting the two is implicit, and we have multiple snippet files named the same thing (so many main.snip's!). I think we should consider fully qualifying the paths to our snippet files in the config yaml's.

garrettjonesgoogle:
LGTM

PR #99, opened 2016-05-12
Improved sample gen

michaelbausor:
Added while loop access for unpaged list callables
Added futureCall usage to other callables
Other tweaks in main.snip

### DIFF
--- a/src/main/java/com/google/api/codegen/java/JavaContextCommon.java
+++ b/src/main/java/com/google/api/codegen/java/JavaContextCommon.java
@@ -202,6 +202,12 @@ public class JavaContextCommon {
 
     public abstract String getReturnType();
 
+    public String getGenericAwareReturnType() {
+      String returnType = getReturnType();
+      if (returnType == null || returnType.isEmpty()) return "Void";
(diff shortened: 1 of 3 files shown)


michaelbausor:
@garrettjonesgoogle Re futureCall - this PR uses futureCall for callables returning a single object and for paged callables, but not for unpaged list callables (which use the while loop)
Callables with no return type (e.g. deleteTopic) use a ListenableFuture<Void>

src/main/java/com/google/api/codegen/java/JavaContextCommon.java
```diff
@@ -202,6 +202,12 @@ public static Variable s_newVariable(
 
     public abstract String getReturnType();
 
+    public String getGenericAwareReturnType() {
+      String returnType = getReturnType();
+      if (returnType == null || returnType.isEmpty()) return "Void";
```
garrettjonesgoogle: always use brackets for if statements

michaelbausor:
PTAL

garrettjonesgoogle:
LGTM

PR #100, opened 2016-05-12
Clean up file-getting logic

### LINKED ISSUE #82: Clean up ProtoElement file logic
There are numerous places where toolkit code needs to access the ProtoFile corresponding to a ProtoElement, or one of its ancestors if it's not possible to obtain the file directly from the element; for example see the C# utility `getNamespaceFile`. After #71, let's refactor this logic into `ProtoElementView`.

geigerj:
There is no need to get the element parent of proto elements in order
to determine the proto file; that logic is already contained in the
ProtoElement `getFile()` method.

Fixes #82


geigerj:
cc: @bjwatson @chrisdunelm

garrettjonesgoogle:
LGTM

geigerj:
Note: post-LGTM updates to C# references to the now-deleted CSharpXXXView in order to fix the build

Issue #40, opened 2016-05-03
Consolidate language-independent proto comment "fixers"


geigerj:
Both Python and Ruby have similar "comment fixers" ([Python](https://github.com/googleapis/toolkit/blob/1ab4b29ab60b84fd395582355ab222efaff637a9/vgen/src/main/java/io/gapi/vgen/py/PythonSphinxCommentFixer.java), [Ruby](https://github.com/googleapis/toolkit/blob/1ab4b29ab60b84fd395582355ab222efaff637a9/vgen/src/main/java/io/gapi/vgen/ruby/RDocCommentFixer.java)), and presumably other languages may want a similar construct.

We should factor out common data (e.g., regular expressions for matching links) and interfaces (e.g., fixProtoMarkdownLinks(), fixCloudMarkdownLinks()) into a common base class and let each language implement the methods.

This prevents issues like #38, where a fix in Ruby to a language-independed regex must be manually propagated to Python.

geigerj:
Fixed by #101

Issue #38, opened 2016-05-02
Python codegen should take care of link format


jmuk:
From the comment of @geigerj at https://github.com/googleapis/toolkit/pull/34/, where ruby reformatting is addressed slightly.

geigerj:
Also should allow the format `[display text][fully.qualified.proto.name]`

geigerj:
Fixed by #101

PR #101, opened 2016-05-12
Refactor Ruby/Python Comment fixers

geigerj:
Ruby and Python use similar regular expressions to make proto
comments idiomatic. Factor these regexes out into a utility class.

### DIFF
--- a/src/main/java/com/google/api/codegen/CommentPatterns.java
+++ b/src/main/java/com/google/api/codegen/CommentPatterns.java
@@ -1,0 +1,31 @@
+/* Copyright 2016 Google Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


geigerj:
Fixes #40 
Also affects #38

src/main/java/com/google/api/codegen/CommentPatterns.java
```diff
@@ -0,0 +1,14 @@
+package com.google.api.codegen;
+
+import java.util.regex.Pattern;
+
+/**
+ * Abstract base class for formatting source comments to follow language-idiomatic style.
```
garrettjonesgoogle: This isn't abstract, or a base class

geigerj: Whoops, bad comment. Fixed.

geigerj: (FWIW: I had originally considered making the methods of `RDocCommentFixer` and `PythonSphinxCommentFixer` nonstatic and deriving them from an abstract base class, but I don't know that there's any benefit to doing so -- that might just be an unnecessary constraint.)

garrettjonesgoogle: In general, I prefer to keep things more modular, and static utility classes are much more modular than abstract base classes.

geigerj:
Note: added distinction between relative and absolute markdown links to fix an issue in #103.

jmuk:
LGTM

PR #103, opened 2016-05-12
Sphinxify all Python comments

geigerj:
Previously there were still uses of the raw proto comment strings.
Instead, introduce a helper method that wraps the function that gets
proto comments and sphinxifies them.

Fixes googleapis/gax-python#83

### DIFF
--- a/src/test/java/com/google/api/codegen/testdata/python_library.baseline
+++ b/src/test/java/com/google/api/codegen/testdata/python_library.baseline
@@ -616,15 +616,15 @@ class FieldMask(object):
 
 class Any(object):
     """
-    `Any` contains an arbitrary serialized message along with a URL
+    ``Any`` contains an arbitrary serialized message along with a URL
     that describes the type of the serialized message.
 
 
     # JSON
 
-    The JSON representation of an `Any` value uses the regular
+    The JSON representation of an ``Any`` value uses the regular
     representation of the deserialized, embedded message, with an
-    additional field `@type` which contains the type URL. Example:
+    additional field ``@type`` which contains the type URL. Example:
 
         package google.profile;
         message Person {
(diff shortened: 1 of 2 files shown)


src/test/java/com/google/api/codegen/testdata/python_library.baseline
```diff
@@ -919,15 +919,15 @@ class Timestamp(object):
     0001-01-01T00:00:00Z to 9999-12-31T23:59:59.999999999Z.
     By restricting to that range, we ensure that we can convert to
     and from  RFC 3339 date strings.
-    See [https://www.ietf.org/rfc/rfc3339.txt](https://www.ietf.org/rfc/rfc3339.txt).
+    See `https://www.ietf.org/rfc/rfc3339.txt <https://cloud.google.comhttps://www.ietf.org/rfc/rfc3339.txt>`_.
```
geigerj: cc @jmuk 

This is wrong. We need to fix the link regex.

jmuk: Yes, this is wrong. IIRC PythonSphinxCommentFixer adds this cloud.google.com somehow, you need to fix there (and the same problem exists in RDocCommentFixer)

jmuk: minor correction: RDocCommentFixer had the same problem, but now it's fixed (it just checks if the link starts with http or not).

geigerj: I've put the fix in #101. I'll rebase this PR on that one once #101 is merged.

geigerj: rebased/fixed

geigerj:
Jun, do you mind taking a look at this?

jmuk:
LGTM

PR #104, opened 2016-05-12
Changes to match PublisherApi; generating API methods

garrettjonesgoogle:


src/main/resources/com/google/api/codegen/php/main.snip
```diff
@@ -311,13 +312,14 @@
             $this->scopes = $options['scopes'];
 
             $generatedCreateStub = function($hostname, $opts) {
-                return new {@service.getSimpleName}Client($hostname, $opts);
+                return new {@context.getGrpcClientName(service)}($hostname, $opts);
```
shinfan: You can just use {@grpcClientName} here

src/test/java/com/google/api/codegen/testdata/php_library.baseline
```diff
@@ -145,10 +121,53 @@ class LibraryServiceApi
...
+                    'requestPageTokenField' => 'page_token',
+                    'responsePageTokenField' => 'next_page_token',
+                    'resourceField' => 'books']);
+        $listStringsPageStreamingDescriptor =
+                new PageStreamingDescriptor([
+                    'requestPageTokenField' => 'page_token',
+                    'responsePageTokenField' => 'next_page_token',
+                    'resourceField' => 'strings']);
+
+        $pageStreamingDescriptors = [
```
shinfan: To be consistent with other static methods, we should make $pageStreamingDescriptors a static variable here so that if this method is invoked for multiple times, we only create it once.

src/test/java/com/google/api/codegen/testdata/php_library.baseline
```diff
@@ -55,59 +68,22 @@ class LibraryServiceApi
...
-            'listBooks' => $listBooksPageStreamingDescriptor,
-            'listStrings' => $listStringsPageStreamingDescriptor
-        ];
-    }
-
-    /**
      * Formats a string containing the fully-qualified path to represent
      * a shelf resource.
      */
     public static function formatShelfName($shelf)
```
shinfan: Have you considered using a separate instance to contains all the utility methods?

shinfan:
Overall looks very good to me. Just a few suggestions.

src/test/java/com/google/api/codegen/testdata/php_library.baseline
garrettjonesgoogle: That could make sense, but this construction is pretty low cost. The PathTemplate construction is higher cost (it has to parse the path template), so I wanted to make sure that it was only initialized once, but it's kind of a pain to do it for everything that is constant across instances - each thing needs to have an initializer method that initializes it only once.

src/test/java/com/google/api/codegen/testdata/php_library.baseline
garrettjonesgoogle: This was already covered in a separate discussion - we'll leave this as is for now, and have a broader discussion across languages.

shinfan: As discussed, we can hold this for now and have a broader discussion across the languages.

garrettjonesgoogle:
PTAL

src/main/resources/com/google/api/codegen/php/main.snip
garrettjonesgoogle: updated.

src/test/java/com/google/api/codegen/testdata/php_library.baseline
shinfan: I am not concerned about the cost but more about the potential risks of issues. For example, if we need to modify this $pageStreamingDescriptors and call getPageStreamingDescriptors(), the modification will not be preserved. (However since it is a static get method, I expect the modifications are preserved)
It is okay as-is here since pageStreamingDescriptors will not be modified.

shinfan:
LGTM

PR #117, opened 2016-05-16
Minor Fixes to Go Discovery Sample Gen

### LINKED ISSUE #112: Remove the comment about oauth2.NoContext

### LINKED ISSUE #114: Use better name than "service" for instance of client library

### LINKED ISSUE #116: Comment error handling instead of log.Fatal

pongad:
The HTTP client is renamed from "client" to "httpClient" and
"service" is renamed to "client". Fixes #114.

Remove the comment about oauth2.NoContext. Fixes #112.

Comment that an error should be handled instead of log.Fatal(err).
Fixes #116.

### DIFF
--- a/src/main/resources/com/google/api/codegen/go/discovery_fragment.snip
+++ b/src/main/resources/com/google/api/codegen/go/discovery_fragment.snip
@@ -31,11 +31,13 @@ func main() {
   # TODO(pongad): Every API except autoscaler has the string as API.CloudPlatformScope, find out why
   httpClient, err := google.DefaultClient(ctx, "https://www.googleapis.com/auth/cloud-platform")
   if err != nil {
-    _ = err // Handle error.
+    // TODO: Handle error.
+    _ = err
   }
   client, err := {@context.getApi.getName}.New(httpClient)
   if err != nil {
-    _ = err // Handle error.
+    // TODO: Handle error.
+    _ = err
   }
 
   {@initialize(method)}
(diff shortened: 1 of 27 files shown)


src/main/resources/com/google/api/codegen/go/discovery_fragment.snip
```diff
@@ -20,25 +20,22 @@ package main
...
 
 func main() {
-  // Use oauth2.NoContext if there isn't a good context to pass in.
   ctx := context.Background()
   # TODO(pongad): Every API except autoscaler has the string as API.CloudPlatformScope, find out why
-  client, err := google.DefaultClient(ctx, "https://www.googleapis.com/auth/cloud-platform")
+  httpClient, err := google.DefaultClient(ctx, "https://www.googleapis.com/auth/cloud-platform")
   if err != nil {
-    log.Fatal(err)
+    _ = err // Handle error.
```
vchudnov-g: I think the following might be more visible:

```
if err != nil {
  // TODO: Handle error properly.
  _ = err
}
```


WDYT?

src/test/java/com/google/api/codegen/testdata/discoveries/go/go_appengine.v1beta5.json.baseline
```diff
@@ -12,24 +12,21 @@ package main
...
+    _ = err // Handle error.
   }
-  service, err := appengine.New(client)
+  client, err := appengine.New(httpClient)
   if err != nil {
-    log.Fatal(err)
+    _ = err // Handle error.
   }
 
   // TODO: Change placeholders below to appropriate parameters values for the 'get' method:
```
vchudnov-g: While you're here:

s/parameters/parameter/

src/main/resources/com/google/api/codegen/go/discovery_fragment.snip
pongad: I don't have a strong opinion in either case. Will do.

vchudnov-g:
LGTM

PR #121, opened 2016-05-17
Fix minor errors of clientconfig

jmuk:
- retry_codes_def is not quoted, which is invalid in JSON spec
- construct_settings expect that the method names are in the
  original upper-camelcase format


geigerj:
LGTM

PR #123, opened 2016-05-18
Update iterator design

saicheems:


src/test/java/com/google/api/codegen/testdata/go_library.baseline
```diff
@@ -462,166 +462,153 @@ func (client *Client) ListStrings(ctx context.Context, req *google_example_libra
...
+
+    iterator := &StringIterator{}
+    iterator.apiCall = func() error {
+        if iterator.atLastPage {
+            return Done
+        }
+        var resp *google_example_library_v1.ListStringsResponse
+        err := gax.Invoke(ctx, func (c context.Context) error {
+            var err error
+            req.PageToken = iterator.nextPageToken
```
saicheems: Would appreciate a sanity check on this `nextPageToken` logic. It didn't seem there was a pathway before for the user to actually provide a value for this - I included a way via the `SetPageToken` method.

saicheems:
@pongad 
@vchudnov-g 
@jba 
@jmuk

src/test/java/com/google/api/codegen/testdata/go_library.baseline
jba: Looks OK to me, but I'd hold off. I'm not sure where the internal discussion is going to end up.

src/main/resources/com/google/api/codegen/go/main.snip
```diff
@@ -78,6 +78,8 @@
     )
 
     var (
+        Done = errors.New("iterator done")
```
jmuk: The reason why we use io.EOF is
- it's already builtin and known by programmers probably
- gRPC streaming is using the same error

Also -- defining a Done value per client seems redundant to me. If doing so, it would be better to define it in gax-golang, and use it here.

saicheems: This was a guideline in the iterator design doc Jonathan sent out. There is a rationale for using a special sentinel value there.

I'm not personally opposed to using `io.EOF`, but the argument is that it doesn't make sense in an iterator context, and a special sentinel value doesn't really belong in cloud or gax.

jmuk: Hm, I've missed the design doc, sorry -- inviting @jba to make sure.

Also --
- you should put comment above the definition, so godoc can show what this constant means.
- I prefer not to define this if the service does not have page-streaming methods (so far all services have page-streaming, but there can be)

saicheems: Good points, will do

saicheems: Done

jmuk:
LGTM from my side.

src/test/java/com/google/api/codegen/testdata/go_library.baseline
saicheems: Think the discussion is settled?

jba: Yes, or very nearly. I'm about to change the recommendation to:


   - Next() and NextPage() on the same iterator.
   - Iterator has methods:
   - SetPageSize(int32)
      - SetPageToken(string)
      - NextPageToken() string

saicheems:
Updated:
Purposefully leaving out `SetPageSize`, will implement in a follow up.

jba:
LGTM.

PR #128, opened 2016-05-19
Consolidating generated file writing logic

garrettjonesgoogle:


garrettjonesgoogle:
Note: the one case I didn't address was C#; it figures out its path based on the C# namespace specified in the proto file, which is different from the rest, which figure it out from the GAPIC config.

src/main/java/com/google/api/codegen/ruby/RubyProvider.java
```diff
@@ -38,24 +31,6 @@
...
-  public <Element> void output(
-      String packageRoot, String outputPath, Multimap<Element, GeneratedResult> elements)
-      throws IOException {
-    Map<String, Doc> files = new LinkedHashMap<>();
-    for (Map.Entry<Element, GeneratedResult> entry : elements.entries()) {
-      Element element = entry.getKey();
-      GeneratedResult generatedResult = entry.getValue();
-      String root;
-      if (element instanceof Method) {
-        root = ((Method) element).getFile().getFullName().replace('.', '/');
```
geigerj: Is this method-specific logic useful? Where did the logic go?

geigerj: Actually, I guess this would only be used by a Method-view codegen process, and there isn't one of those for Ruby?

geigerj:
One small bit of confusion about a line that was deleted in RubyProvider.java, but if that is dead code (as I suspect it is), then LGTM

src/main/java/com/google/api/codegen/ruby/RubyProvider.java
garrettjonesgoogle: It looked like it was inherited from Python, and Python doesn't actually use the analogous codepath, although Ruby might possibly. @pongad is checking.

garrettjonesgoogle: So, hilarious note: the Method that is imported here is com.google.api.tools.framework.model.Method, but the Method that is provided from RubyDiscoveryProvider is com.google.protobuf.Method. Consequently, this codepath can't be invoked and is thus dead.

PR #129, opened 2016-05-19
Fix Go Discovery Generation

pongad:
The bug was introduced with Python Discovery generation, and I missed
it.

This commit reverts the change to DiscoveryImporter: Now it will only
report the return type of "empty$" if the return type is completely
missing from the Discovery doc.

Most languages can ignore return type "Empty", so the string "Empty" is
especially handled. Go cannot ignore "Empty" type and is special-cased.

### DIFF
--- a/src/main/java/com/google/api/codegen/DiscoveryImporter.java
+++ b/src/main/java/com/google/api/codegen/DiscoveryImporter.java
@@ -409,11 +409,6 @@ public class DiscoveryImporter {
       return EMPTY_TYPE_NAME;
     }
     return response.get("$ref").asText();
-    // String typeUrl = response.get("$ref").asText();
-    // if (typeUrl.equals("Empty")) {
-    //   return EMPTY_TYPE_NAME;
-    // }
-    // return typeUrl;
   }
 
   /**
--- a/src/main/java/com/google/api/codegen/DiscoveryImporter.java
+++ b/src/main/java/com/google/api/codegen/DiscoveryImporter.java
@@ -408,11 +408,12 @@ public class DiscoveryImporter {
     if (response == null) {
       return EMPTY_TYPE_NAME;
     }
-    String typeUrl = response.get("$ref").asText();
-    if (typeUrl.equals("Empty")) {
-      return EMPTY_TYPE_NAME;
-    }
-    return typeUrl;
+    return response.get("$ref").asText();
+    // String typeUrl = response.get("$ref").asText();
+    // if (typeUrl.equals("Empty")) {
+    //   return EMPTY_TYPE_NAME;
+    // }
+    // return typeUrl;
   }
 
   /**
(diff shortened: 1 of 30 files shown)


src/main/java/com/google/api/codegen/DiscoveryImporter.java
```diff
@@ -408,11 +408,12 @@ private String responseTypeUrl(JsonNode root) {
     if (response == null) {
       return EMPTY_TYPE_NAME;
     }
-    String typeUrl = response.get("$ref").asText();
-    if (typeUrl.equals("Empty")) {
-      return EMPTY_TYPE_NAME;
-    }
-    return typeUrl;
+    return response.get("$ref").asText();
+    // String typeUrl = response.get("$ref").asText();
```
vchudnov-g: remove cruft

src/main/java/com/google/api/codegen/go/GoDiscoveryContext.java
```diff
@@ -110,6 +111,11 @@ private String typeName(Type type, Field field) {
   }
 
   @Override
+  public boolean isResponseEmpty(Method method) {
```
vchudnov-g: Why do we need to override this? Can't we just let DiscoveryContext.java handle this?

src/test/java/com/google/api/codegen/testdata/discoveries/go/go_clouddebugger.v2.json.baseline
```diff
@@ -183,10 +183,13 @@ func main() {
     breakpointId = ""
   )
 
-  if err = client.Debugger.Debuggees.Breakpoints.Delete(debuggeeId, breakpointId).Context(ctx).Do(); err != nil {
+  callResult, err := client.Debugger.Debuggees.Breakpoints.Delete(debuggeeId, breakpointId).Context(ctx).Do()
```
vchudnov-g: s/callResult/result/ ? or "response"?

src/main/java/com/google/api/codegen/DiscoveryImporter.java
pongad: whoops. sorry, done

src/main/java/com/google/api/codegen/go/GoDiscoveryContext.java
pongad: Added Javadoc to explain.

src/test/java/com/google/api/codegen/testdata/discoveries/go/go_clouddebugger.v2.json.baseline
pongad: Done.

src/main/java/com/google/api/codegen/go/GoDiscoveryContext.java
vchudnov-g: Oh, and I see my eyes glazed over the difference the first time around. OK.

vchudnov-g:
LGTM

PR #130, opened 2016-05-19
Renaming (Language)Provider classes to (Language)SnippetSetRunner

garrettjonesgoogle:
Now that the (Language)Provider classes only have a generate method in them, it will be much clearer for the classes to be named for their purpose, which is simply to run the corresponding SnippetSet. This also avoids confusion with the provider classes which are referenced in gapic.yaml files.


geigerj:
LGTM

"This also avoids confusion with the provider classes which are referenced in gapic.yaml files." 👍

PR #131, opened 2016-05-19
Reduce package name via regex

saicheems:
Removes `V[0-9]+` and `Service` from the end of the go package.


saicheems:
@jba 
@jmuk 
@pongad 
@vchudnov-g 

A thought - it's not necessarily robust... but is this something we still wanted to do?

For logging, it would produce:
`.../logging/client.go`
`.../metrics/client.go`
`.../config/client.go`

jba:
Seems fine. Internal guidelines say "In the rare case where an interface name would conflict with another name within the API, a suffix (for example Api or Service) should be used to disambiguate." So some uses of Service may be necessary. But I think most are not. Definitely putting the version in the name is wrong.

So LGTM, and we can revisit if there's ever a conflict.

PR #132, opened 2016-05-19
Remove error return from path template methods

saicheems:
Since the code generator ensures that the correct variable mapping is
passed to the `Instantiate` function, it's impossible for an error to be
returned.

Resolves #132

### DIFF
--- a/src/main/resources/com/google/api/codegen/go/main.snip
+++ b/src/main/resources/com/google/api/codegen/go/main.snip
@@ -184,12 +184,16 @@
              methodReceiver = context.lowerUnderscoreToLowerCamel(collectionConfig.getEntityName)
 
             // {@pathTypeName}Path returns the path for the {@humanReadableName} resource.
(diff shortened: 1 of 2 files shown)


saicheems:
@jba 
@pongad 
@jmuk 
@vchudnov-g

vchudnov-g:
Could you clarify the description of this issue? Do you mean that Instantiate() always returns a nil error?

saicheems:
The only failure mode for instantiate is if the mapping provided is incomplete:

so for the path `bar/{hello}/foo/{world}`, if I passed in `{"hello": "test"}`, it would fail because `"world"` was not provided. Since our generated code guarantees that this error path is not possible, we can ignore it.

saicheems:
Just updated the commit message for clarity.

jba:
Can we do the following instead?

s, err := ...
if err != nil {
   panic(err)
}

saicheems:
Done

jba:
LGTM.

PR #134, opened 2016-05-19
Add PathTemplate match helpers

geigerj:
Previously Python/Ruby codegen lacked helpers for the "match" method
in PathTemplate. This change conforms to the Java surface.


geigerj:
Fixes #134

jmuk:
LGTM

PR #136, opened 2016-05-20
Include package path in generated file path

saicheems:


saicheems:
@jba 
@jmuk 
@pongad 
@vchudnov-g

pongad:
lgtm

PR #145, opened 2016-05-24
Correct package names

saicheems:


saicheems:
@jba 
@vchudnov-g 
@jmuk 
@pongad

jba:
I don't think we've decided on the path yet.

saicheems:
This particular pull was more for convenience - so I don't have to edit the package names/import paths after regenerating to test the client.

Will hold off until a decision is made.

Issue #110, opened 2016-05-16
Allow page_size optional parameter


geigerj:
Currently [ServiceMessages.java](https://github.com/googleapis/toolkit/blob/4e19fd1bc9f4c1fd40d0b20878da2de9234a0daa/src/main/java/com/google/api/codegen/ServiceMessages.java#L80) suppresses the `page_size` field from being generated as a parameter in page streaming calls. In the languages that support required/optional parameters (e.g., Python), page_size should actually be present as an optional parameter.

geigerj:
Fixed by #147

PR #147, opened 2016-05-24
Add page_size param to codegen

geigerj:
Add page size to

- Python
- Ruby
- NodeJS
- PHP

code gen.


geigerj:
@jmuk 
Could you take a look at this? If it looks OK, I will make the corresponding change in the Ruby codegen as well.

geigerj:
FYI @tseaver @tbetbetbe 

This addresses the concern mentioned in googleapis/gax-python#94 and recorded as issue #110

geigerj:
This change conflicts with #153. I will finish this up once that change is merged.

geigerj:
@jmuk updated for other languages, PTAL
@garrettjonesgoogle This change adds `page_size` into optional parameters, which affects PHP. I've more-or-less tried to do the same thing to PHP codegen as to Python/Ruby/NodeJS, but could you check that this change makes sense in the context of PHP GAX?

src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
```diff
@@ -124,58 +122,77 @@ private String fieldComment(String comment, Field field) {
...
     if (optionalParams.iterator().hasNext()) {
       paramTypesBuilder.append("@param {?Object} otherArgs\n");
       for (Field field : optionalParams) {
-        paramTypesBuilder.append(fieldParamComment(field, true));
+        if (config.isPageStreaming()
+            && field.equals((config.getPageStreaming().getPageSizeField()))) {
+          paramTypesBuilder.append(
+              fieldParamComment(
+                  field,
+                  "The maximum number of resources contained in the underlying API\nresponse. If "
```
jmuk: nit: I like to break the line at "\n" instead of padding as far as possible -- so it can look same as what will appear in the actual code.

src/test/java/com/google/api/codegen/testdata/nodejs_library.baseline
```diff
@@ -710,6 +726,7 @@ LibraryServiceApi.prototype.moveBook = function moveBook() {
  *
  * @param {?Object} otherArgs
  * @param {String} otherArgs.name
+ * @param {number} otherArgs.pageSize
```
jmuk: Why doesn't this have comments?

geigerj: In the GAPIC config, this field is deliberately not configured to be part of page streaming. This is supposed to model the fact that the config distinguishes the semantics of the `page_size` field from the field name itself.

src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
geigerj: Good idea. Fixed.

src/test/java/com/google/api/codegen/testdata/nodejs_library.baseline
jmuk: ah, I see :smile:

jmuk:
LGTM

src/test/java/com/google/api/codegen/testdata/php_library.baseline
```diff
@@ -385,8 +385,13 @@ class LibraryServiceApi
...
      *
      * @param array $optionalArgs {
-     *     Optional. There are no optional parameters for this method yet;
-     *               this $optionalArgs parameter reserves a spot for future ones.
+     *     Optional.
+     *     @type integer $page_size
+     *           The maximum number of resources contained in the underlying API
+     *           response. If page streaming is performed per-resource, this
+     *           parameter does not affect the return value. If page streaming is
+     *           performed per-page, this "determines the maximum number of
```
garrettjonesgoogle: There is an extra quote character in here

src/test/java/com/google/api/codegen/testdata/php_library.baseline
```diff
@@ -385,8 +385,13 @@ class LibraryServiceApi
      * Lists shelves.
      *
      * @param array $optionalArgs {
-     *     Optional. There are no optional parameters for this method yet;
-     *               this $optionalArgs parameter reserves a spot for future ones.
+     *     Optional.
+     *     @type integer $page_size
```
garrettjonesgoogle: the parameter should be camel case

src/main/java/com/google/api/codegen/php/PhpGapicContext.java
```diff
@@ -175,86 +175,85 @@ public String getDescription(ProtoElement element) {
...
-          "__DIR__",
-          "__FILE__",
-          "__FUNCTION__",
-          "__LINE__",
-          "__METHOD__",
-          "__NAMESPACE__",
-          "__TRAIT__")
-      .build();
-
+          .add(
```
garrettjonesgoogle: why are we getting this huge formatting change?

geigerj: I have Eclipse configured to autoformat on save, and I think it made this change.

src/main/resources/com/google/api/codegen/php/main.snip
```diff
@@ -486,8 +486,17 @@
             {@""} * @@param array $optionalArgs {
              *     Optional.
             @join param : optionalFields
-                @join commentLine : context.php.getDocLinesWithPrefixes(context.getDescription(param), optionalParamDocLinePrefix(param), "          ")
-                    {@commentLine}
+                @if isPageSizeField(param, methodConfig)
+                    {@""} *     @@type integer ${@methodConfig.getPageStreaming.getPageSizeField.getSimpleName}
```
garrettjonesgoogle: So it looks like we're completely tossing away the proto-specified doc for page_size parameters. Is this what we intend? I think this is the first instance of us auto-generating API-specific behavioral documentation.

src/main/java/com/google/api/codegen/php/PhpGapicContext.java
garrettjonesgoogle: Why is the output different from the mass-reformatting that @bjwatson did?

src/main/resources/com/google/api/codegen/php/main.snip
geigerj: We do this only if the GAPIC yaml annotates that particular field as belonging to page streaming. This is necessary because VKit page streaming behavior is not the same as gRPC page streaming; in many cases the proto comments on this field simply do not apply to VKit code, for example, if you are doing per-resource page streaming.

I think doing this is OK because this is a field with common semantics across VKit libraries in the specific context of page streaming.

src/main/java/com/google/api/codegen/php/PhpGapicContext.java
geigerj: Looks like @bjwatson did the mass-reformatting (Apr. 20) before this was added (May 5)

src/test/java/com/google/api/codegen/testdata/php_library.baseline
geigerj: Done

src/test/java/com/google/api/codegen/testdata/php_library.baseline
geigerj: Done

geigerj:
@garrettjonesgoogle Should I wait for your LGTM before I merge this?

garrettjonesgoogle:
LGTM

tseaver:
@geigerj Sorry I missed seeing the issue earlier, but wouldn't it make more sense to have the `page_size` be passed as a member of `CallOptions` (like `page_token`), rather than as a separate parameter?

geigerj:
My original thinking was that

- `page_token` belongs in `CallOptions` because we've given it additional meaning beyond its use as field in the proto request message, in that `page_token` is a toggle for the response type being iterated over.
- in contrast, `page_size`, as a call parameter, is a simple wrapper for a proto field, no additional semantics.
- We'd have to consider that if both `page_token` and `page_size` are in `CallOptions`, do both of them act as toggles? Or just one of them? Or neither, with some other option acting being a separate toggle?

I'm now thinking that it's possible to do better than our current set-up; from #237, it sounds like @michaelbausor will be redesigning the way we specify the page streaming options to be more consistent across the vkit languages, which may involve moving where `page_size`/`page_token` are specified.

PR #148, opened 2016-05-24
Advanced sample code parameter initialization

garrettjonesgoogle:
* Implementing the initialization of complex structures in Vkit Java snippet code
* Refactoring how JavaDocConfig works


garrettjonesgoogle:
@pongad this might be interesting for you to review since it touches on stuff you bootstrapped.

garrettjonesgoogle:
@bjwatson

src/main/java/com/google/api/codegen/MethodConfig.java
```diff
@@ -147,6 +152,10 @@ public boolean apply(Field input) {
     ImmutableMap<String, String> fieldNamePatterns =
         ImmutableMap.copyOf(methodConfig.getFieldNamePatterns());
 
+    List<String> sampleCodeInitFields = new ArrayList<>();
+    sampleCodeInitFields.addAll(methodConfig.getRequiredFieldsList());
+    sampleCodeInitFields.addAll(methodConfig.getSampleCodeInitFieldsList());
```
michaelbausor: Your changes didn't cause this, but the methodConfig parameter is confusingly named since we are creating a MethodConfig object, perhaps rename to methodConfigProto?

src/main/java/com/google/api/codegen/java/JavaDocConfig.java
```diff
@@ -0,0 +1,161 @@
...
+
+    public abstract JavaDocConfig.Builder setPagedVariant(boolean paged);
+
+    public abstract JavaDocConfig.Builder setCallableVariant(boolean callable);
+
+    public abstract JavaDocConfig.Builder setResourcesFieldForUnpagedListCallable(Field field);
+
+    public abstract JavaDocConfig build();
+  }
+}
```
michaelbausor: Add newline

src/main/java/com/google/api/codegen/metacode/InitCodeGenerator.java
```diff
@@ -0,0 +1,167 @@
...
+    while (symbolTable.contains(actualName)) {
+      actualName = desiredName + i;
+    }
+    symbolTable.add(actualName);
+    return actualName;
+  }
+
+  private InitCodeLine generateSampleCodeInit(String suggestedName, TypeRef typeRef,
+      Object initFieldStructure) {
+    if (initFieldStructure instanceof InitValueConfig) {
```
michaelbausor: Having this first means we don't check for typeRef.isRepeated or typeRef.isMessage, which would be error conditions that we check for later. Perhaps this could be in an else if block after those checks?

src/main/java/com/google/api/codegen/metacode/InitCodeGenerator.java
```diff
@@ -0,0 +1,167 @@
...
+      InitValueConfig initValueConfig = (InitValueConfig) initFieldStructure;
+      String identifier = getNewSymbol(suggestedName);
+      return SimpleInitCodeLine.create(typeRef, identifier, initValueConfig);
+    }
+
+    if (typeRef.isMessage() && !typeRef.isRepeated()) {
```
michaelbausor: Place this block after the type.isRepeated() block below, then don't need to check that it is not repeated

src/main/resources/com/google/api/codegen/java/main.snip
```diff
@@ -670,54 +668,61 @@
...
+  @join line : initCodeSpec.getLines()
+    @switch line.getLineType.toString()
+    @case "StructureInitLine"
+      {@initLineStructure(line)}
+    @case "ListInitLine"
+      {@initLineList(line)}
+    @case "SimpleInitLine"
+      {@initLineSimple(line)}
+    @default
+      ??
```
bjwatson: This seems like it should fail fast (i.e. calling a non-existent function) rather than silently putting ?? into the snippet.

src/main/java/com/google/api/codegen/metacode/InitValueConfig.java
```diff
@@ -0,0 +1,48 @@
...
+    return new AutoValue_InitValueConfig(apiWrapperName, collectionConfig);
+  }
+
+  @Nullable
+  public abstract String getApiWrapperName();
+
+  @Nullable
+  public abstract CollectionConfig getCollectionConfig();
+
+  public boolean isEmpty() {
```
michaelbausor: Only one of isEmpty and hasFormattingConfig is necessary

src/main/java/com/google/api/codegen/metacode/ListInitCodeLine.java
```diff
@@ -0,0 +1,50 @@
...
+  @Override
+  public InitCodeLineType getLineType() {
+    return InitCodeLineType.ListInitLine;
+  }
+
+  @Override
+  public InitValueConfig getInitValueConfig() {
+    return InitValueConfig.create();
+  }
+}
```
michaelbausor: Add newline

src/main/resources/com/google/api/codegen/java/main.snip
```diff
@@ -670,54 +668,61 @@
...
-        {@initLineSimple(param, docConfig)}
-      @else
-        {@initLineMessage(param, docConfig)}
-      @end
-    @else
-      {@initLineSimple(param, docConfig)}
+@private initCode(initCodeSpec)
+  @join line : initCodeSpec.getLines()
+    @switch line.getLineType.toString()
+    @case "StructureInitLine"
```
michaelbausor: Maybe indent the case statements? Don't know what our standard is here...

src/main/java/com/google/api/codegen/metacode/InitCodeGenerator.java
garrettjonesgoogle: No, it is intentional for this to be first. If we reached this condition, it means that we have reached the end of a field spec, but there is more structure in the proto definition. The way will handle that at codegen time is to do Type.newBuilder().build(). If the type in the proto is a basic type, we will use the zero value for the type.

src/main/java/com/google/api/codegen/metacode/InitCodeGenerator.java
garrettjonesgoogle: I was trying to follow the same pattern everywhere: Map, List, simple type. (e.g. see FieldStructureParser.) I know that the condition checks might be slightly more optimized by putting it after, but I think the consistency of ordering is higher value.

src/main/java/com/google/api/codegen/metacode/InitValueConfig.java
garrettjonesgoogle: This is true now, but I wanted to allow for other types of config or data being added to this concept later. If something else is added, these would differ.

src/main/resources/com/google/api/codegen/java/main.snip
garrettjonesgoogle: The SnippetSet documentation shows them not being indented, so I followed that style:

https://github.com/googleapis/tools-framework/blob/master/src/main/java/com/google/api/tools/framework/snippet/SnippetSet.java#L219

src/main/java/com/google/api/codegen/MethodConfig.java
garrettjonesgoogle: Agreed, I renamed it.

src/main/java/com/google/api/codegen/java/JavaDocConfig.java
garrettjonesgoogle: Done.

src/main/resources/com/google/api/codegen/java/main.snip
garrettjonesgoogle: Done.

src/main/java/com/google/api/codegen/metacode/ListInitCodeLine.java
garrettjonesgoogle: Done.

garrettjonesgoogle:
PTAL - For all comments, I have either responded or made the suggested change.

michaelbausor:
LGTM

PR #149, opened 2016-05-24
Python GAPIC sample gen

bjwatson:
This does everything I think we currently need for GAPIC Python sample gen. I have also published a new version (0.7.6) of https://pypi.python.org/pypi/gax-google-pubsub-v1 with these auto-generated samples (the gcloud-python team needed the latest code).

Part of this PR is factoring out code from Java snippet gen that is also useful for Python. Another part is separating the part of Java main.snip that does GAPIC snippet gen into its own file (method_sample.snip), to make it easier to follow (Java main.snip currently has almost 800 lines of code on master; this change separates out the 150 lines that pertain to GAPIC snippet gen).


garrettjonesgoogle:
This PR and my PR (https://github.com/googleapis/toolkit/pull/148) were published *literally* 1 minute from each other (3:14 vs 3:15). There are a substantial number of conflicts. I will discuss with @bjwatson how to resolve them.

bjwatson:
@garrettjonesgoogle walked me through his PR. Since he will almost certainly merge first (he's functionality complete whereas I'm not), I will proactively merge his branch into mine and resolve the conflicts.

Nevertheless, this is ready for early feedback if y'all want to take a look.

src/main/java/com/google/api/codegen/DocConfig.java
```diff
@@ -0,0 +1,184 @@
...
+  public abstract String getApiName();
+
+  public abstract String getMethodName();
+
+  public abstract String getReturnType();
+
+  public String getGenericAwareReturnType() {
+    String returnType = getReturnType();
+    if (returnType == null || returnType.isEmpty()) {
+      return "Void";
```
geigerj: This looks Java-specific

src/main/java/com/google/api/codegen/py/PythonContext.java
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen.py;
+
+/**
+ * A PythonContext provides functionality specific to a context in Python.
```
geigerj: Usually, "context" refers to an object that is accessible within a particular snippet file. What does it mean for a context to provide functionality to a context?

src/main/java/com/google/api/codegen/py/PythonGapicContext.java
```diff
@@ -110,7 +118,10 @@ public String filePath(ProtoFile file) {
    * Returns type information for a field in Sphinx docstring style.
    */
   private String fieldTypeCardinalityComment(Field field, PythonImportHandler importHandler) {
-    TypeRef type = field.getType();
+    return fieldTypeCardinalityComment(field.getType(), importHandler);
+  }
+
+  private String fieldTypeCardinalityComment(TypeRef type, PythonImportHandler importHandler) {
```
geigerj: Rename this method, since it's no longer field-specific?

src/main/java/com/google/api/codegen/py/PythonGapicContext.java
```diff
@@ -253,6 +271,46 @@ private String returnTypeComment(Method method, PythonImportHandler importHandle
...
+  private String methodSnippet(Method method, PythonImportHandler importHandler) {
+    Interface service = (Interface) method.getParent();
+    MethodConfig methodConfig = getApiConfig().getInterfaceConfig(service).getMethodConfig(method);
+    String methodName = upperCamelToLowerCamel(method.getSimpleName());
+    ImmutableMap<String, String> fieldNamePatterns = methodConfig.getFieldNamePatterns();
+
+    Iterable<Field> requiredFields = methodConfig.getRequiredFields();
+    Iterable<Field> optionalFields = methodConfig.getOptionalFields();
+    List<Field> fields = new ArrayList<Field>();
+    for (Field field : requiredFields) fields.add(field);
```
geigerj: nit: Google style requires braces here

src/main/java/com/google/api/codegen/py/PythonSnippetSet.java
```diff
@@ -35,4 +35,9 @@
    * Generates the result module with a set of accumulated imports.
    */
   Doc generateModule(Element element, Doc body, Iterable<String> imports);
+  
+  /**
+   * Generate the example snippet for the code documentation.
+   */
+  Doc generateMethodSampleCode(PythonDocConfig config);
```
geigerj: It seems odd to require all Python snippets to have this method when it's not relevant for messages.snip or the discovery snippets.

(I see that we're already doing something similar with generateModule/generateBody where some of the implementations are empty. This also seems strange to me.)

src/test/java/com/google/api/codegen/testdata/python_no_path_templates.baseline
```diff
@@ -132,6 +132,11 @@ class NoTemplatesServiceApi(object):
         and trailing
         whitespace.
 
+        Example:
+          >>> from google.example.v1 import no_path_templates_pb2
+          >>> api = NoTemplatesServiceApi()
```
geigerj: NoTemplatesServiceApi needs a module

src/main/resources/com/google/api/codegen/py/method_sample.snip
```diff
@@ -0,0 +1,97 @@
...
+        @end
+      @end
+  @end
+@end
+
+# Optionally render the return value from the API method call
+@private callResultSampleCode(returnType)
+  @if returnType.isEmpty
+  @else
+    response = {@""}
```
geigerj: Is the empty string required for the snippet to render correctly here?

garrettjonesgoogle: I will interject: yes, it preserves the space between the '=' and the following expression.

src/main/resources/com/google/api/codegen/py/method_sample.snip
```diff
@@ -0,0 +1,97 @@
...
+  @end
+@end
+
+# Render the API method call itself
+@private methodCallSampleCode(docConfig)
+  @if docConfig.isUnpagedListCallableVariant
+    # FIXME: adapt for Python
+    api.{@methodCallName(docConfig)}().call({@argList(docConfig.getParams)})
+  @else
+    @if docConfig.isCallableVariant
```
geigerj: note: Python doesn't have a callable variant

src/main/java/com/google/api/codegen/py/PythonGapicContext.java
```diff
@@ -66,12 +67,19 @@
...
+  private PythonSnippetSet<?> pythonSnippetSet;
 
   public PythonGapicContext(Model model, ApiConfig apiConfig) {
     super(model, apiConfig);
     this.pythonCommon = new PythonContextCommon();
   }
 
+  @Override
+  public void resetState(PythonSnippetSet<?> pythonSnippetSet) {
+    this.pythonSnippetSet = pythonSnippetSet;
```
geigerj: What does this do?

(AFAIK, state in the snippet system comes only from the objects in the globals map that you pass into `createSnippetInterface`, so I might expect something called `resetState` to reinitialize those objects.)

geigerj:
Left a few comments, but mostly just nitpicky stuff, so feel free to ignore if you're looking for more structural feedback at this point.

src/main/java/com/google/api/codegen/DocConfig.java
bjwatson: Seems likely. I added a FIXME to verify when I get to how it's used.

src/main/java/com/google/api/codegen/py/PythonContext.java
bjwatson: Yeah, this is not really context in and of itself, but it provides common functionality to PythonGapicContext and PythonDiscoveryContext.

src/main/java/com/google/api/codegen/py/PythonGapicContext.java
bjwatson: Ok.

src/main/java/com/google/api/codegen/py/PythonGapicContext.java
bjwatson: Blar. Seems excessively verbose, but I will comply.

src/main/java/com/google/api/codegen/py/PythonSnippetSet.java
bjwatson: Yeah, I agree but mostly wanted to follow the recipe for vKit snippet gen already established for Java. There are definitely clean ups to be done in our OOD.

src/test/java/com/google/api/codegen/testdata/python_no_path_templates.baseline
bjwatson: Hmm... good catch. Anyone know why PythonImportHandler is not importing the class name (i.e. `from google.example.v1.no_path_templates_pb2 import NoTemplatesServiceApi`)?

The handwritten Python snippets for PubSub suggest that approach instead: http://pythonhosted.org/gax-google-pubsub-v1/google.pubsub.v1.publisher_api.html

src/main/resources/com/google/api/codegen/py/method_sample.snip
bjwatson: Ok, cool. That simplifies things a little.

src/main/java/com/google/api/codegen/py/PythonGapicContext.java
bjwatson: I copied this approach from the corresponding Java code, which has now been removed.

The call path in Python is a bit different from Java. For Java, most of the logic for calling `@snippet generateMethodSampleCode(...)` is sprinkled throughout `main.snip`, whereas Python is calling it from the PythonGapicContext.comments() -> methodComments() Java code path.

To address your concern, I'll rename this function to `setPythonSnippetSet()`.

bjwatson:
Thanks @geigerj!

src/test/java/com/google/api/codegen/testdata/python_no_path_templates.baseline
bjwatson: Nevermind, I figured it out myself. It was inappropriate of me to use PythonImportHandler, which is intended for surface import statements. This is a different use case, and not worth attempting to hijack PythonImportHandler to support it. Instead I'm directly using PythonImport from PythonGapicContext to generate the import statement I want.

src/main/java/com/google/api/codegen/py/PythonGapicContext.java
```diff
@@ -276,6 +294,60 @@ private String returnTypeComment(Method method, PythonImportHandler importHandle
...
     return pythonCommon.convertToCommentedBlock(contentBuilder.toString());
   }
+  
+  private String methodSnippet(Method method, PythonImportHandler importHandler) {
+    Interface service = (Interface) method.getParent();
+    List<String> importStrings = new ArrayList<>();
+    importStrings.add(
+        PythonImport.create(
+                ImportType.APP,
+                method.getFile().getProto().getPackage()
```
geigerj: Instead of the proto package, use the Python package, which is configured in the GAPIC yaml: `getApiConfig().getPackageName()`.

src/main/java/com/google/api/codegen/py/PythonGapicContext.java
```diff
@@ -276,6 +294,60 @@ private String returnTypeComment(Method method, PythonImportHandler importHandle
...
+                    + "."
+                    + PythonProtoElements.getPbFileName(method.getInputMessage()),
+                service.getSimpleName() + "Api")
```
geigerj: Consider using `getApiWrapperName()` (defined in GapicContext).

src/main/resources/com/google/api/codegen/java/method_sample.snip
```diff
@@ -0,0 +1,157 @@
+@extends "common.snip"
```
garrettjonesgoogle: Maybe consider not pulling out method_sample.snip for java for now, due to my open PR which also changes the corresponding snippets.

src/main/java/com/google/api/codegen/py/PythonGapicContext.java
bjwatson: Are you sure that's right for the snippet import statement? I spent some time studying what this does and it pulls the package name from language_settings in the vKit YAML. The other code only seems to use this for determining the path in which to place the generated files.

The code for generating package names and import statements seems to use other sources. I copied `method.getFile().getProto().getPackage()` from `PythonImportHandler`.

As an example, the `pubsub_gapic.yaml` definition in `googleapis` says:

```
language_settings:
  python:
    package_name: gcloud.pubsub
```

However, the generated docs with hand-written snippets (http://pythonhosted.org/gax-google-pubsub-v1/google.pubsub.v1.publisher_api.html) uses the package `google.pubsub.v1.publisher_api`.

src/main/java/com/google/api/codegen/py/PythonGapicContext.java
bjwatson: Awesome. Done.

src/main/resources/com/google/api/codegen/java/method_sample.snip
bjwatson: It's alright, I already fixed it in a holding branch (https://github.com/bjwatson/toolkit/tree/python-snippets-garrett). I carefully ensured that all your changes have been transferred into that branch, and I'll handle any incremental merges from your PR.

I find keeping method_sample.snip separate is easier than trying to visually distinguish the snippet gen code from the remainder of main.snip, even with the overhead of concurrent refactoring.

Once your PR is in master, I'll merge my holding branch into this one and then merge in master. It will be clean.

src/main/java/com/google/api/codegen/py/PythonGapicContext.java
geigerj: We generated the package for the handwritten snippets with a different config, which had the `package_name` set to `google.pubsub.v1.publisher_api`. But it's not generally true that the Python package in which the generated API object lives is the same as the proto package.

In fact, since the samples are inside the API object, do we need an import at all?

src/main/java/com/google/api/codegen/DocConfig.java
```diff
@@ -0,0 +1,185 @@
...
+import javax.annotation.Nullable;
+
+public abstract class DocConfig {
+  public abstract String getApiName();
+
+  public abstract String getMethodName();
+
+  public abstract String getReturnType();
+
+  // FIXME: is this Java-specific?
```
tcoffee-google: use "TODO": https://engdoc.corp.google.com/eng/doc/devguide/java/styleguide.shtml?cl=head#s4.8.6.2-todo-comments

src/main/java/com/google/api/codegen/py/PythonSnippetSet.java
tcoffee-google: The separation of generateModule/generateBody follows a logic similar to that for generateClass/generateBody in Java, to allow the snippet to generate imports for itself. It so happens that the Python Vkit client generation doesn't need this functionality since it imports *everything* from the proto source; but I can imagine circumstances where this could change. Similarly, we may want to use generateMethodSampleCode across the board to support fragment generation. I suggest we avoid further refactoring on this until our features have fleshed out.

src/test/java/com/google/api/codegen/testdata/python_no_path_templates.baseline
tcoffee-google: Note we use PythonImportHandler for generating method samples for discovery sample gen. The PythonImportHandler has a nontrivial mechanism for avoiding name conflicts, which is always a possibility (however unlikely) even for single-method samples.

bjwatson:
This is nearly done. It should be properly merged with the refactoring done by @garrettjonesgoogle. I still need to add support for Python Protobuf message arguments and update the Python library test.

src/main/java/com/google/api/codegen/DocConfig.java
```diff
@@ -0,0 +1,177 @@
...
+
+  public abstract InitCode getInitCode();
+
+  public abstract ImmutableList<InputParameter> getParams();
+
+  /**
+   * Generic variable definition
+   */
+  @AutoValue
+  public abstract static class Variable {
```
garrettjonesgoogle: Variable shouldn't exist any more - this was renamed to InputParameter

src/main/java/com/google/api/codegen/DocConfig.java
bjwatson: This is a deliberate violation to ensure that I address all FIXME comments prior to merging to master. Only TODOs will be part of that merge.

src/main/java/com/google/api/codegen/py/PythonSnippetSet.java
bjwatson: Agreed. Thanks for the context.

src/test/java/com/google/api/codegen/testdata/python_no_path_templates.baseline
bjwatson: I added a FIXME comment to revisit this and Jacob's comment (https://goto.google.com/huezs) before finalizing this PR.

src/main/java/com/google/api/codegen/DocConfig.java
bjwatson: My bad. Thanks.

src/test/java/com/google/api/codegen/testdata/python_no_path_templates.baseline
```diff
@@ -132,6 +132,11 @@ class NoTemplatesServiceApi(object):
         and trailing
         whitespace.
 
+        Example:
+          >>> from google.example.v1.no_path_templates_pb2 import NoTemplatesServiceApi
```
geigerj: This is still not right -- the module called "no_path_templates_pb2" is a compiled proto file, and doesn't have the class NoTemplatesServiceApi in it. (I'm still not sure this import make sense, since these examples are within the package you're "importing" from.)

If there is an import here, the package should derived be from the veneer config, and the module should be derived from `getApiWrapperName()`.

geigerj: Ah, sorry, I see that you had already responded above that you'll look at this. Ignore that comment.

src/main/java/com/google/api/codegen/py/PythonGapicContext.java
bjwatson: Ok, I understand what's going on. Your original comment was on the right track. I was trying to make it conform to the pythonhosted structure, like you said. I've pushed a fix for this.

src/test/java/com/google/api/codegen/testdata/python_no_path_templates.baseline
bjwatson: @tcoffee-google I looked briefly at how you're using PythonImportHandler in discovery sample gen, and it doesn't look like you're using it for the same use case as what I'm doing here: importing just the API class itself.

src/main/java/com/google/api/codegen/py/PythonGapicContext.java
bjwatson: BTW, I'm doing the import so that we can eventually get Sphinx doctest working. There are a few other issues with that, but I think they might be out of scope for this PR (I'm now running into what appears to be a surface bug, although it could be user error).

bjwatson:
This is functionally complete, but the devil's in the details. I've compared my generated samples against Tim's hand-written samples (http://pythonhosted.org/gax-google-pubsub-v1/), and identified a few remaining issues:

* Formatted var names are kind of funky (see `PublisherApi.delete_topic()` and `LibraryApi.merge_shelves()`)
* Protobuf messages should be properly imported (see FIXMEs in `PythonGapicContext.java` for `LibraryApi`)
* Tim’s formatted variable arguments are a little friendlier looking than mine (e.g. `‘my-foo’` rather than `"[FOO]"`)
* Should we skip optional parameters? (see `SubscriberApi.create_subscription()`)

I also need to review the usability study results to incorporate appropriate feedback, and finally update the baseline file for the Python library test. Looks like this will be done sometime next week.

bjwatson:
@tbetbetbe @geigerj I spent some time today getting Sphinx doctest working with the auto-generated snippets. It looks like it's triggering a bug in the surface, but it could be a user error (especially given that this has already been through a usability study).

* The error is in the call to `api = PublisherApi()` (which every snippet does)
* An example error occurs in `publisher_api.py, line 183, in __init__`
* The error is `TypeError: construct_settings() takes at least 6 arguments (7 given)`

Let's discuss next week to figure out what's going on.

garrettjonesgoogle:
Regarding formatting of variable arguments: we specifically chose the format `"[FOO]"` instead of `"my-foo"` to avoid the problem with project, where `my-project` is actually a real project that exists, so if people try to run the sample code with that project, they will get a permission denied error, which is confusing.

geigerj:
FYI @bjwatson @jmuk The `TypeError` Brian is seeing looks like it may come from googleapis/gax-python#103. It looks like the code generator was not updated after this change to GAX.

bjwatson:
Thanks @garrettjonesgoogle. I now remember that discussion. I will keep the current format for example values.

bjwatson:
@geigerj @jmuk Thanks for the analysis of the error I saw. When do you think it could be fixed?

jmuk:
Well, the code generator was updated already. I think that's because the gax package in PyPI wasn't updated yet.
For now, please fetch the HEAD of googleapis/gax-python and install it locally.

jmuk:
Also -- we need to update the package dependency on packman -- filed https://github.com/googleapis/packman/issues/37

geigerj:
Let me know if using HEAD for gax-python fixes the issue. If it does, I can push the latest version to PyPI.

jmuk:
gax-python repository is up to date
(see https://github.com/googleapis/gax-python/blob/master/google/gax/api_callable.py#L317), and the version is also updated to 0.11.0.

geigerj:
I've confirmed that updating GAX fixes the `TypeError` issue.

I've pushed Python GAX 0.11.0 to PyPI and filed PR googleapis/packman#38 to update the Packman configuration.

bjwatson:
@geigerj @jmuk Agreed about that fix. Now I'm seeing this error:

```
Document: gcloud.pubsub.publisher_api
-------------------------------------

Warning, treated as error:
WARNING: **********************************************************************
File "gcloud.pubsub.publisher_api.rst", line 10, in default
Failed example:
    api = PublisherApi()
Exception raised:
    Traceback (most recent call last):
      File "/usr/local/google/home/brianwatson/.linuxbrew/Cellar/python/2.7.11/lib/python2.7/doctest.py", line 1315, in __run
        compileflags, 1) in test.globs
      File "<doctest default[1]>", line 1, in <module>
        api = PublisherApi()
      File "/usr/local/google/home/brianwatson/src/gapi-dev/gapi-pubsub-python/.tox/docs/lib/python2.7/site-packages/gcloud/pubsub/publisher_api.py", line 189, in __init__
        pubsub_pb2.beta_create_Publisher_stub,
    AttributeError: 'module' object has no attribute 'beta_create_Publisher_stub'
```

geigerj:
Just to check, can you confirm that you are loading the right `pubsub_pb2.py`? There are actually two of these floating around after you've run generation. One is the "real" file, generated by `protoc`, and one is a "fake" file templated by `messages.snip`, which is used only to generate documentation (which therefore wouldn't have the expected attribute `beta_create_Publisher_stub`).

bjwatson:
Thanks @geigerj. I did not realize two of these are created. I'll take a closer look today to check.

bjwatson:
@geigerj Your theory is correct. I didn't realize this because I wasn't running `PythonGrpcClientPipeline`; just `PythonGapicClientPipeline`. Unfortunately, when I substitute the `pubsub_pb2.py` generated by the former into `gapi-pubsub-python`, I get different errors running `tox`:

```
brianwatson@brianwatson0:~/src/gapi-dev/gapi-pubsub-python$ tox -e docs
GLOB sdist-make: /usr/local/google/home/brianwatson/src/gapi-dev/gapi-pubsub-python/setup.py
docs inst-nodeps: /usr/local/google/home/brianwatson/src/gapi-dev/gapi-pubsub-python/.tox/dist/gax-google-pubsub-v1-1.0.13.zip
docs installed: alabaster==0.7.8,Babel==2.3.4,docutils==0.12,enum34==1.1.6,futures==3.0.5,gax-google-pubsub-v1==1.0.13,google-gax==0.11.0,googleapis-common-protos==1.1.0,grpc-google-pubsub-v1==1.0.15,grpcio==0.13.1,httplib2==0.9.2,imagesize==0.7.1,Jinja2==2.8,MarkupSafe==0.23,oauth2client==2.1.0,ply==3.8,protobuf==3.0.0b3,pyasn1==0.1.9,pyasn1-modules==0.0.8,Pygments==2.1.3,pytz==2016.4,PyYAML==3.11,rsa==3.4.2,six==1.10.0,snowballstemmer==1.2.1,Sphinx==1.4.2,sphinx-rtd-theme==0.1.9
docs runtests: PYTHONHASHSEED='411169849'
docs runtests: commands[0] | python -c import shutil; shutil.rmtree('docs/_build', ignore_errors=True)
docs runtests: commands[1] | python -c import shutil; shutil.rmtree('docs/generated', ignore_errors=True)
docs runtests: commands[2] | sphinx-build -W -b doctest -d docs/_build/doctrees docs docs/_build/html
Running Sphinx v1.4.2
making output directory...
loading pickled environment... not yet created
[autosummary] generating autosummary for: apis.rst, foo.conf.rst, gcloud.pubsub.publisher_api.rst, gcloud.pubsub.subscriber_api.rst, google.protobuf.timestamp_pb2.rst, google.pubsub.v1.pubsub_pb2.rst, index.rst, starting.rst
Failed to import 'google.pubsub.v1.pubsub_pb2': no module named google.pubsub.v1.pubsub_pb2
Failed to import 'gcloud.pubsub.publisher_api': no module named gcloud.pubsub.publisher_api
Failed to import 'gcloud.pubsub.subscriber_api': no module named gcloud.pubsub.subscriber_api
Failed to import 'google.protobuf.timestamp_pb2': no module named google.protobuf.timestamp_pb2

Warning, treated as error:
WARNING: [autosummary] failed to import u'gcloud.pubsub.publisher_api': no module named gcloud.pubsub.publisher_api

ERROR: InvocationError: '/usr/local/google/home/brianwatson/src/gapi-dev/gapi-pubsub-python/.tox/docs/bin/sphinx-build -W -b doctest -d docs/_build/doctrees docs docs/_build/html'
_________________________________________________________________________________________________ summary __________________________________________________________________________________________________
ERROR:   docs: commands failed
```

I've spent enough time on this today. I'm going to punt on doctest for now and finish up the other issues that have been raised for this PR. Perhaps we can discuss this issue in more detail on Friday.

geigerj:
Yup, let's chat about this Friday.

I'm not sure specifically what's causing your issue, but the short answer is that the "real" `pubsub_pb2` shouldn't be in `gapi-pubsub-python`; instead, it should get pulled in with the package [`grpc-google-pubsub-v1` as a dependency](https://pypi.python.org/pypi/grpc-google-pubsub-v1). So if you also have a module called `google.pubsub.v1.pubsub_pb2` in addition to the module of the same name in that package, your local package is likely shadowing the dependency.

bjwatson:
@geigerj and I got doctest working today, but it required a lot of manual changes. I think this can be automated, but it will take some time and is out of scope for this feature. I will write up a design document next week.

Now I'm focused on fixing the remaining issues with these snippets.

src/main/java/com/google/api/codegen/py/PythonGapicContext.java
```diff
@@ -295,12 +315,91 @@ private String returnTypeComment(
...
+     *    // FIXME: need to import Shelf with correct package
+     *    >>> shelf = library_pb2.Shelf();
+     *    >>> books = [];
+     *    >>> edition = 0;
+     *    >>> response = api.publishSeries(shelf, books, edition)
+     *
+     *  Example:
+     *    >>> from google.example.library.v1.library_pb2 import LibraryServiceApi
+     *    >>> api = LibraryServiceApi()
+     *    >>> formattedName = LibraryServiceApi.book_path("[SHELF]", "[BOOK]");
```
geigerj: Let's use single quotes throughout for consistency with the codegen

bjwatson:
@geigerj @garrettjonesgoogle All I have left to do tomorrow is fixing the imports for proto messages, and adding sample code to python_main_library.baseline (and re-enabling that test). Then this is ready for final review.

src/main/java/com/google/api/codegen/clientconfig/ClientConfigSnippetSetRunner.java
```diff
@@ -29,11 +29,10 @@
...
       ClientConfigSnippetSetRunner.class.getPackage().getName().replace('.', '/');
 
-  @SuppressWarnings("unchecked")
   @Override
-  public GeneratedResult generate(ElementT element,
-      String snippetFileName,
-      CodegenContext context) {
+  @SuppressWarnings("unchecked")
+  public GeneratedResult generate(
+      ElementT element, String snippetFileName, CodegenContext context) {
```
garrettjonesgoogle: Maybe revert files which only have formatting changes, and leave that to a separate change to reduce noise in this one?

src/main/java/com/google/api/codegen/py/PythonContext.java
```diff
@@ -0,0 +1,22 @@
...
+ */
+public interface PythonContext {
+  void setPythonSnippetSet(PythonSnippetSet<?> pythonSnippetSet);
```
garrettjonesgoogle: This function isn't needed. We have removed it from JavaContext & PhpContext.

src/main/java/com/google/api/codegen/clientconfig/ClientConfigSnippetSetRunner.java
bjwatson: Once @pongad's PR is in master and merged here, I think this will be a moot point.

src/main/java/com/google/api/codegen/py/PythonContext.java
bjwatson: I'll take a look and see how that was done. The call path is not exactly the same is in Python, because so much more of writing Python docstrings is done in the toolkit Java code, rather than in the template code.

bjwatson: Yeah, I checked and this is not easily avoided in Python.

For Java, you call `@snippet generateMethodSampleCode()` directly from other template code.

In Python, I had to call it from `PythonGapicContext.methodComments()` (well actually from a sub-method that I added in this PR). I think changing the structure of that is out-of-scope for this PR.

garrettjonesgoogle: Alrighty, makes sense!

bjwatson:
@pongad @garrettjonesgoogle @geigerj This is ready for final review!

src/main/java/com/google/api/codegen/DocConfig.java
```diff
@@ -0,0 +1,130 @@
...
+        initValueConfigMap.put(fieldNamePattern.getKey(), initValueConfig);
+      }
+      Map<String, Object> initFieldStructure =
+          FieldStructureParser.parseFields(
+              methodConfig.getSampleCodeInitFields(), initValueConfigMap.build());
+      return initFieldStructure;
+    }
+
+    public BuilderType setRequestObjectParam(
+        @SuppressWarnings("unused") GapicContext context,
```
geigerj: Why do we need these unused params?

src/main/java/com/google/api/codegen/py/PythonGapicContext.java
```diff
@@ -296,11 +321,88 @@ private String returnTypeComment(
...
+        return;
+      }
+    }
+  }
+
+  /**
+   * Generate the sample code for a method.
+   */
+  public String generateMethodSampleCode(PythonDocConfig config) {
+    return pythonSnippetSet.generateMethodSampleCode(config).prettyPrint();
```
geigerj: I am a little concerned about the complexity of running snippet codegen within snippet codegen -- I think it makes the code hard to follow, and removes the ability for the sample templates to "talk" to the main.snip templates about things like indentation and formatting.

I would suggest breaking the Python comment methods in PythonGapicContext into pre- and post- sample sections so that generateMethodSampleCode can be called directly from main.snip, like in Java.

(In the long term, I think we should try to adopt a more Java-like approach to comments by leveraging the snippet code rather than doing all of the comment generation in the context, but that's for another day.)

src/main/java/com/google/api/codegen/py/PythonContext.java
geigerj: See my comment below on `PythonGapicContext`'s `generateMethodSampleCode` method. I think it's possible to avoid this interface.

src/test/java/com/google/api/codegen/testdata/python_main_library.baseline
```diff
@@ -302,6 +315,22 @@ class LibraryServiceApi(object):
...
 
+        Example:
+          >>> from library.library_service_api import LibraryServiceApi
+          >>> from google.gax import CallOptions, INITIAL_PAGE
+          >>> api = LibraryServiceApi()
+          >>>
+          >>> # Iterate over all results
+          >>> for element in api.list_shelves():
+          >>>   pass
+          >>>   # process element
```
geigerj: Would prefer the comment go over the `pass` line, so that it's clear that `pass` means "process element" rather than "no-op".

bjwatson: Done.

src/main/java/com/google/api/codegen/py/PythonGapicContext.java
bjwatson: I've refactored the code as you suggested. PTAL

src/main/java/com/google/api/codegen/DocConfig.java
bjwatson: Good point. I removed them.

I guess thought certain language might need them in this base class definition, but they can add them back if needed in the future.

bjwatson:
I have addressed all of Jacob's comments. PTAL

src/main/java/com/google/api/codegen/py/PythonContext.java
bjwatson: Yep, I implemented Jacob's suggestion and got rid of this.

src/main/java/com/google/api/codegen/java/JavaDocConfig.java
```diff
@@ -68,101 +52,34 @@ public boolean isUnpagedListCallableVariant() {
...
-              .build();
-      return setParams(ImmutableList.of(param));
-    }
+    public abstract Builder setResourcesFieldForUnpagedListCallable(Field field);
 
-    public JavaDocConfig.Builder setEmptyParams() {
-      return setParams(ImmutableList.<InputParameter>of());
+    @Override
+    protected Builder setInitCodeProxy(InitCode initCode) {
+      return setInitCode(initCode);
```
garrettjonesgoogle: Are the proxy methods actually necessary? It seems like it would work without them.

garrettjonesgoogle:
LGTM. I just had a small optional thing; I think we should prioritize getting this committed soon because it enables other languages to do sample gen for vkit too.

geigerj:
LGTM

src/main/java/com/google/api/codegen/java/JavaDocConfig.java
bjwatson: Possibly, but I had trouble getting it to work without them. I tried declaring methods like `public abstract setInitCode()` in the DocConfig super-class, but autovalue was not picking them up and defining them in the @AutoValue annotated sub-classes. I spent about an hour trying to figure it out (and reading through the autovalue docs) before doing the proxy methods and moving on to fixing other issues.

If someone has a better idea for how to do this, let's fix it in a subsequent PR.

bjwatson:
Woo hoo! Squashing and merging now. :)

PR #150, opened 2016-05-24
Fix Python type documentation

### LINKED ISSUE #42: Python codegen: document as the python int type, there is no int32
From feedback on [subscriberApi.create_subscription](http://pythonhosted.org/gax-google-pubsub-v1/google.pubsub.v1.subscriber_api.html#google.pubsub.v1.subscriber_api.SubscriberApi.create_subscription) method

> ack_deadline_seconds is described as an (int32). there is no such type in python ...

this should just be described as an int

geigerj:
Python type hints should refer to Python types, not proto types.

Fixes #42


geigerj:
cc @tbetbetbe

jmuk:
LGTM

PR #153, opened 2016-05-25
Add page_size to config and Go iterator

saicheems:


saicheems:
@jba 
@jmuk 
@vchudnov-g 
@pongad 

@shinfan @garrettjonesgoogle Modified the config to include the page_size parameter, ptal? it didn't cause any test failures from what I could tell... seems a red flag.

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -204,8 +204,12 @@ message PageStreamingConfigProto {
 // `PageStreamingRequestProto` defines which fields match the paging pattern in
 // the request.
 message PageStreamingRequestProto {
```
jba: Out of scope for this CL, but "streaming" confuses me because there are no streaming RPCs involved.

src/test/java/com/google/api/codegen/testdata/go_library.baseline
```diff
@@ -537,6 +541,12 @@ func (iterator *ShelfIterator) Next() (*google_example_library_v1.Shelf, error)
     return result, nil
 }
 
+// SetPageSize sets the maximum size of the next page to be
+// retrieved.
+func (iterator *ShelfIterator) SetPageSize(int pageSize) {
```
jba: pageSize int32

saicheems: Should I specify the corresponding struct types to be int32 as well?

src/main/java/com/google/api/codegen/config/PageStreamingConfigGenerator.java
```diff
@@ -54,15 +56,16 @@
...
       if (fieldName.equals(PARAMETER_PAGE_TOKEN)) {
-        Map<String, Object> requestConfig = new LinkedHashMap<String, Object>();
         requestConfig.put(CONFIG_KEY_TOKEN_FIELD, fieldName);
-        return requestConfig;
+      } else if (fieldName.equals(PARAMETER_PAGE_SIZE)) {
+        requestConfig.put(CONFIG_KEY_PAGE_SIZE_FIELD, fieldName);
       }
     }
-    return null;
+    return requestConfig;
```
shinfan: Either this line should return null or you need to modify the null check in L45.

src/main/java/com/google/api/codegen/PageStreamingConfig.java
```diff
@@ -26,6 +26,7 @@
  * PageStreamingConfig represents the page streaming configuration for a method.
  */
 public class PageStreamingConfig {
+  private final Field requestPageSizeField;
```
geigerj: You can just call this `pageSizeField`, since it's not ambiguous, unlike `token`

src/main/java/com/google/api/codegen/PageStreamingConfig.java
```diff
@@ -77,20 +91,28 @@ public static PageStreamingConfig createPageStreaming(
               resourcesFieldName));
     }
 
-    if (requestTokenField == null || responseTokenField == null || resourcesField == null) {
+    if (requestPageSizeField == null || requestTokenField == null || responseTokenField == null || resourcesField == null) {
```
geigerj: `page_size` is not mandatory, so I don't think we should be doing this check.

src/main/java/com/google/api/codegen/config/PageStreamingConfigGenerator.java
saicheems: Done

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -204,8 +204,12 @@ message PageStreamingConfigProto {
...
+  // The name of the field in the request containing the desired number of items
+  // in the next page.
+  string page_size_field = 1;
+
   // The name of the field in the request containing the page token.
-  string token_field = 1;
+  string token_field = 2;
```
geigerj: nit: It's better not to change proto field numbers...

geigerj:
Sorry, jumping in -- I was working on essentially the same change for Ruby/Python (#147) and just saw this.

src/test/java/com/google/api/codegen/testdata/go_library.baseline
jba: Yes.

src/main/java/com/google/api/codegen/PageStreamingConfig.java
saicheems: Okay, I wasn't sure about that - will do

jba: This is a good point. None of the page-related methods of an iterator should exist unless the List request/response protos have all three pagination fields with the right types.

src/main/proto/com/google/api/codegen/config.proto
saicheems: Reverted

src/main/java/com/google/api/codegen/PageStreamingConfig.java
saicheems: Done

src/main/java/com/google/api/codegen/PageStreamingConfig.java
```diff
@@ -77,20 +83,37 @@ public static PageStreamingConfig createPageStreaming(
...
     this.requestTokenField = requestTokenField;
     this.responseTokenField = responseTokenField;
     this.resourcesField = resourcesField;
   }
 
   /**
+   * Returns true if page_size is not null.
+   */
+  public boolean hasPageSizeField() {
+    return pageSizeField != null;
```
saicheems: Note, this is really just so I can check for the field in snippet language... since null's aren't allowed

If someone knows a way around this that doesn't require a new method please let me know.

src/test/java/com/google/api/codegen/testdata/go_library.baseline

src/main/java/com/google/api/codegen/PageStreamingConfig.java
geigerj: Unfortunately, I don't think there's a way around this. There are similar methods in the other *Config classes.

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -206,6 +206,10 @@ message PageStreamingConfigProto {
 message PageStreamingRequestProto {
   // The name of the field in the request containing the page token.
   string token_field = 1;
+
+  // The name of the field in the request containing the desired number of items
+  // in the next page.
+  string page_size_field = 2;
```
geigerj: Optional: changes to the GAPIC yaml should also involve a change to the config generator. I've already done that in #147, so it should be OK if that change follows this one.

saicheems: Right, I did make a change to the config generator - I checked your pull and the implementation seems to be _nearly_ identical, so should be fine

geigerj: Ah, sorry, I see it now.

src/main/java/com/google/api/codegen/PageStreamingConfig.java
```diff
@@ -39,6 +40,11 @@
   @Nullable
   public static PageStreamingConfig createPageStreaming(
       DiagCollector diagCollector, PageStreamingConfigProto pageStreaming, Method method) {
+    // Since page_size is not required, no error should be raised if it's not present.
+    String pageSizeFieldName = pageStreaming.getRequest().getPageSizeField();
+    Field pageSizeField =
+        method.getInputType().getMessageType().lookupField(pageSizeFieldName);
```
geigerj: This should still cause an error.

- page_size is specified in config, page_size is not defined in proto ==> error
- page_size is not specified in config ==> pageSizeField is set to null

src/main/java/com/google/api/codegen/PageStreamingConfig.java
```diff
@@ -77,20 +83,37 @@ public static PageStreamingConfig createPageStreaming(
...
+   */
+  public boolean hasPageSizeField() {
+    return pageSizeField != null;
+  }
+
+  /**
+   * Returns the field used in the request to specify desired page size, and null if it's not
+   * present.
+   */
+  public Field getPageSizeField() {
```
geigerj: `@Nullable` annotation?

saicheems:
Rebased and ready

Issue #139, opened 2016-05-20
Reconsider the return value for page-streaming methods


jmuk:
In Ruby (and Python), page-streaming methods return either a stream object or a response object, based on the call options passed to.

I got a concern about this -- "Feels too subtle a mechanism & slightly error prone."  Maybe we want to reveal the interface for low-level API calls, which would be still retryable but simply receives a request object and returns a response object?

jmuk:
@tbetbetbe @geigerj thoughts?

geigerj:
- What does this mean exactly in terms of the surface? Does this mean that if per-resource page streaming is enabled, we construct a new response object? (I'm not sure this would be possible in Python proto.)

- We do not return a response object for page streaming methods. It's always an iterable; the switch toggles whether it is an iterable of pages, or an iterable of resources. (We do attach the response as an attribute to the iterable, but that's secondary.)

jmuk:
Hm, so that's Ruby-specific, but CallOptions can suppress page-streaming behavior (by passing is_page_streaming: false), it will return the response object in that case. Removed Python labels -- I'll fix this for Ruby (fix will be both on gax and toolkit).

I'm guessing that sometimes the user wants to see the response object or wants to control the low-level APIs of request->response. I might be wrong.

geigerj:
The low-level case is possible still; the response object is an attribute of the per-page iterable.

In Python:

Case 1: per resource
```
for foo in listFoos():
  print foo.name
```

Case 2: per page
```
for page in listFoos(callOptions(page_token=gax.INITIAL_PAGE)):
  for foo in page:
    print foo.name
```

Case 3: low level
```
resp1 = listFoos(callOptions(page_token=gax.INITIAL_PAGE)).response
for foo in resp1.foos:
  print foo.name
resp2 = listFoos(callOptions(page_token=resp1.next_page_token))
for foo in resp2.foos:
  print foo.name

```

jmuk:
Accessing the response object is possible in Ruby as well IIRC, but a bit obscure, it's like `listFoos().page.response`. Possibly better to bypass `.page`.

I've forgotten about the `page_token` call options -- that's not in Ruby yet. Filed https://github.com/googleapis/gax-ruby/issues/18 to track this.

PR #154, opened 2016-05-25
Fix wrong Ruby comments.

### LINKED ISSUE #140: Rename ssl_creds parameter
Ruby constructor uses 'ssl_creds' as the name of the parameter to the service constructor, which is actually a ChannelCredentials instance. And we use 'chan_creds' for create_stub method in gax. This should be chan_creds everywhere.

jmuk:
- ssl_creds is actually grpc::ChannelCredentials, better to name
  it as chan_creds. Fixed the comments.
- page-streaming methods now return PagedEnumerable always,
  and that class provides operations for per-page iterations
  and others. Correct the documents.

Fixes #140, #139


geigerj:
LGTM

PR #155, opened 2016-05-25
Documentation changes

### LINKED ISSUE #41: Python codegen: swap ',' for '->' when in documentation for dictionaries
> when you document a dictionary you are writing it as dict[ foo, bar ] which to my Python eyes implies that it is a dictionary who's keys are tuples (foo, bar) and does not tell me what the value is.  **Don't use a ',' there. A '->' more obviously implies mapping from foo -> bar.**

### LINKED ISSUE #43: Python codegen: document page-streaming methods using 'Returns:' rather than 'Yields'
> These have Yields: documentation that says they yield things as one might expect a generator to do. But then they go on to say the call options can change this - "If page streaming is disabled, a single XXX_pb2.XXX_Response instance is returned."

The docs should not use the 'Yields' keyword here.  What's going on is
-  api_call methods that support page streaming optionally return an iterable that happen to be implemented using a generator function
- they are not generator functions themselves, and should not be documented as such

For clarity the docs 
- should make it clear that an iterable is returned when page-streaming is enabled
- should avoid mentioning that the iterable is a generator so as not confuse readers by making them think that its a badly documented generator function

geigerj:
- Dictionary type hints are now in [key -> value] style. Previously
  they were in [key, value] style.
- Update page streaming return type to reflect changes to the page
  streaming implementation in GAX.

Fixes #41 
Fixes #43 

cc: @tbetbetbe

### DIFF
--- a/src/main/java/com/google/api/codegen/py/PythonGapicContext.java
+++ b/src/main/java/com/google/api/codegen/py/PythonGapicContext.java
@@ -138,7 +138,8 @@ public class PythonGapicContext extends GapicContext {
     String prefix;
     if (type.getCardinality() == Cardinality.REPEATED) {
       if (type.isMap()) {
-        prefix = String.format("dict[%s, ", fieldTypeComment(type.getMapKeyField(), importHandler));
+        prefix =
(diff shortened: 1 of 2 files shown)


jmuk:
LGTM

PR #157, opened 2016-05-25
Testing 'bytes' type in proto

garrettjonesgoogle:


src/test/java/com/google/api/codegen/testdata/library.proto
```diff
@@ -361,5 +361,5 @@ message Comment {
   string user_name = 1;
 
   // won't be filled in by the sample generator
```
michaelbausor: Does this comment need to change now, because the comment is filled and the user_name is not?

michaelbausor:
LGTM, comment is a nitpick

PR #158, opened 2016-05-25
User-friendly strings for Java Discovery snippets

pongad:
Equivalents for other languages can easily be implemented
once this lands.

This change only concerns itself with patterns of the form
  ^A/[^/]*/B/[^/]*$
There are only a handful other common patterns.
However, they are large and difficult to automate.
They can be special-cased instead.

### DIFF
--- a/src/main/java/com/google/api/codegen/ApiaryConfig.java
+++ b/src/main/java/com/google/api/codegen/ApiaryConfig.java
@@ -74,6 +74,13 @@ public class ApiaryConfig {
       HashBasedTable.<String, String, String>create();
 
   /**
+   * Specifies the pattern of the field. The pattern is expressed as a regular expression, like
+   * "^projects/[^/]*$". The table is indexed by (type name, field name).
+   */
+  private final Table<String, String, String> pattern =
+      HashBasedTable.<String, String, String>create();
+
+  /**
    * Records whether or not the method allows media upload.
    */
   private final Set<String> mediaUpload = new HashSet<>();
@@ -115,6 +122,10 @@ public class ApiaryConfig {
 
   public Table<String, String, String> getStringFormat() {
     return stringFormat;
+  }
+
+  public Table<String, String, String> getFieldPattern() {
+    return pattern;
   }
 
   public Map<String, Type> getTypes() {
--- a/src/main/java/com/google/api/codegen/ApiaryConfig.java
+++ b/src/main/java/com/google/api/codegen/ApiaryConfig.java
@@ -65,7 +65,7 @@ public class ApiaryConfig {
       HashBasedTable.<String, String, Boolean>create();
 
   /**
-   * Specifies the format of the field. A pair (type name, field name) is in this table if the type
+   * Specifies the format of each field. A pair (type name, field name) is in this table if the type
    * of the field is "string" and specific format is given in the discovery doc. The format is one
    * of {"int64", "uint64", "byte", "date", "date-time"}. Note: other string formats from the
    * discovery doc are encoded as types in the Service.
@@ -74,7 +74,7 @@ public class ApiaryConfig {
       HashBasedTable.<String, String, String>create();
 
   /**
-   * Specifies the pattern of the field. The pattern is expressed as a regular expression, like
+   * Specifies the pattern of each field. The pattern is expressed as a regular expression, like
    * "^projects/[^/]*$". The table is indexed by (type name, field name).
    */
   private final Table<String, String, String> pattern =
@@ -93,7 +93,8 @@ public class ApiaryConfig {
   /*
    * Maps method name to set of auth scope URLs, eg https://www.googleapis.com/auth/cloud-platform.
    */
-  private final ListMultimap<String, String> authScopes = ArrayListMultimap.<String, String>create();
+  private final ListMultimap<String, String> authScopes =
+      ArrayListMultimap.<String, String>create();
 
   /*
    * Maps (type, field name) to field.
--- a/src/main/java/com/google/api/codegen/discovery/DefaultString.java
+++ b/src/main/java/com/google/api/codegen/discovery/DefaultString.java
@@ -33,11 +33,12 @@ public class DefaultString {
   /**
    * Returns a default string from `pattern`, or null if the pattern is not supported.
    */
-  public static String of(String pattern) {
-    // We only care about patterns that look like
-    //  ^projects/[^/]*/subscriptions/[^/]*$
+  public static String forPattern(String pattern) {
+    // We only care about patterns that has alternating literal and wildcard like
+    //  ^foo/[^/]*/bar/[^/]*$
     // Ignore if what we get looks nothing like this.
-    if (!pattern.startsWith("^")
+    if (pattern == null
+        || !pattern.startsWith("^")
         || !pattern.endsWith("$")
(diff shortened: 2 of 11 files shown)


src/test/java/com/google/api/codegen/testdata/discoveries/java/java_cloudbilling.v1.json.baseline
```diff
@@ -232,7 +232,7 @@ public class CloudbillingExample {
 
     // TODO: Change placeholders below to values for parameters to the 'getBillingInfo' method:
     // * The resource name of the project for which billing information is retrieved. For example, `projects/tokyo-rain-123`.
-    String name = "";
```
garrettjonesgoogle: Don't use "MY-PROJECT". It already exists, and if people run this code without changing the project, they will get a permission denied error.

src/main/java/com/google/api/codegen/ApiaryConfig.java
```diff
@@ -74,6 +74,13 @@
       HashBasedTable.<String, String, String>create();
 
   /**
+   * Specifies the pattern of the field. The pattern is expressed as a regular expression, like
```
vchudnov-g: nit: both here and in the previous comment, I suggest s/of the field/of each field/ for clarity

src/main/java/com/google/api/codegen/discovery/DefaultString.java
```diff
@@ -0,0 +1,150 @@
...
+ * Creates default string from path patterns.
+ */
+public class DefaultString {
+
+  private static final String WILDCARD_PATTERN = "[^/]*";
+
+  /**
+   * Returns a default string from `pattern`, or null if the pattern is not supported.
+   */
+  public static String of(String pattern) {
```
vchudnov-g: nit: consider s/of/for/, which might read better

src/main/java/com/google/api/codegen/discovery/DefaultString.java
```diff
@@ -0,0 +1,150 @@
...
+    // We only care about patterns that look like
+    //  ^projects/[^/]*/subscriptions/[^/]*$
```
vchudnov-g: nit: when I first read this, I thought the literals were part of what you were expecting. Do you think using foo and bar instead would make it more obvious that the literals are just an example?

Maybe you could also describe in words the pattern a little bit, eg "path patterns that contain any number of WILDCARD_PATTERN and end with a WILDCARD_PATTERN, like this:" ?

src/main/java/com/google/api/codegen/discovery/DefaultString.java
```diff
@@ -0,0 +1,150 @@
...
+          .append(literal)
+          .append("/{MY-")
+          .append(Inflector.singularize(literal).toUpperCase())
+          .append('}');
+    }
+    return ret.substring(1);
+  }
+
+  /**
+   * Counts the number of non-overlapping instances of `needle` in `haystack`.
```
vchudnov-g: ;-)

src/main/java/com/google/api/codegen/discovery/DefaultString.java
```diff
@@ -0,0 +1,150 @@
...
+        pattern = pattern.substring(WILDCARD_PATTERN.length());
+      } else if ((slash = pattern.indexOf("/")) >= 0) {
+        elems.add(Elem.createLiteral(pattern.substring(0, slash)));
+        pattern = pattern.substring(slash);
+      } else {
+        elems.add(Elem.createLiteral(pattern));
+        pattern = "";
+      }
+
+      if (pattern.startsWith("/")) {
```
vchudnov-g: I think you could fold this into the else/ifs if you want:
while(...){
if (startsWith(WILDCARD)) {}
elseif (startsWIth("/")) {}
elseif (indexOf("/")>0) {}
else {}
}

src/main/java/com/google/api/codegen/discovery/DefaultString.java
```diff
@@ -0,0 +1,150 @@
...
+      }
+    }
+    return ImmutableList.<Elem>copyOf(elems);
+  }
+
+  /**
+   * Returns whether the pattern represented by the list is in a form we expect.
+   *
+   * A valid pattern must have the same number of literals and wildcards, alternating, and starts
+   * with a literal. Literals must be consisted only of letters.
```
vchudnov-g: nit: s/be consisted only of/consist of only/

src/test/java/com/google/api/codegen/discovery/DefaultStringTest.java
```diff
@@ -0,0 +1,55 @@
...
+import com.google.common.collect.ImmutableMap;
+import com.google.common.truth.Truth;
+
+import org.junit.Test;
+
+import java.util.Map;
+
+public class DefaultStringTest {
+  @Test
+  public void testInvalidString() {
```
vchudnov-g: would it be clearer to call this testInvalidPattern?

src/test/java/com/google/api/codegen/discovery/DefaultStringTest.java
```diff
@@ -0,0 +1,55 @@
...
+    String[] invalid =
+        new String[] {
```
vchudnov-g: Check null pattern as well?

src/main/java/com/google/api/codegen/ApiaryConfig.java
pongad: Done.

src/main/java/com/google/api/codegen/discovery/DefaultString.java
pongad: "for" doesn't work, it's a Java keyword. I changed it to "forPattern"

src/main/java/com/google/api/codegen/discovery/DefaultString.java
pongad: I see the confusion. We actually accept the pattern if it's alternates between literal and wildcard. Add clarification.

src/main/java/com/google/api/codegen/discovery/DefaultString.java
pongad: Done.

src/test/java/com/google/api/codegen/discovery/DefaultStringTest.java
pongad: Good catch. Done.

src/test/java/com/google/api/codegen/discovery/DefaultStringTest.java
pongad: Nice. Done.

src/main/java/com/google/api/codegen/discovery/DefaultString.java
pongad: I don't think this is right. The "if" is used to clean up the leading slash (if one exists) so that the next iteration can directly begin at the wildcard or literal. It might need to run following both "startsWith(WILDCARD)" or "indexOf(/) >= 0". Am I making any sense?

vchudnov-g: I was thinking more of having that cleanup take up another loop iteration. But either way is fine.

And you're fine with the second if possibly producing an empty literal if the index==0?

vchudnov-g:
LGTM

PR #164, opened 2016-05-26
Add page_size into GAPIC config

geigerj:


src/test/java/com/google/api/codegen/testdata/library_gapic.yaml
```diff
@@ -167,6 +168,7 @@ interfaces:
     page_streaming:
       request:
         token_field: page_token
+        page_size_field: page_size
```
saicheems: The `ListStrings` request also has a `page_size` field, but I intended to remove that one for tests where no `page_size` is specified.
Not super important, but if you remove the field from the proto then the yaml should match.

saicheems:
Other than that LGTM

src/test/java/com/google/api/codegen/testdata/library_gapic.yaml
geigerj: Ah -- I had intended ListStrings to be a test example for Python where `page_size` is in the proto but not the config. In this case, Python codegen should just treat `page_size` like an ordinary optional parameter. Since there are three page streaming requests, how about I remove `page_size` from a different one?

saicheems: Got it - that sounds good, but no need to do it in this CL

PR #167, opened 2016-05-31
Delete stray file and fix gRPC warning

pongad:
Baseline file ruby_dataflow.v1b3.json.baseline is not used and deleted.

The addition of AddComments RPC creates an HTTP path conflict,
generating a warning.
Change the path, so the warning disappears.


garrettjonesgoogle:
LGTM assuming tests pass

PR #168, opened 2016-05-31
Remove CallOption vararg

saicheems:
Requires all configuration to be performed on initialization.

### DIFF
--- a/src/main/resources/com/google/api/codegen/go/main.snip
+++ b/src/main/resources/com/google/api/codegen/go/main.snip
@@ -204,12 +204,6 @@
     @end
 @end
 
-@private callOptions(methodName)
-    var callOpts []gax.CallOption
-    callOpts = append(callOpts, client.callOptions["{@methodName}"]...)
-    callOpts = append(callOpts, opts...)
-@end
-
 @private methods(service)
     @join method : service.getMethods
         @let methodName = method.getSimpleName, \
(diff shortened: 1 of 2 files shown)


saicheems:
Based on email thread, lmk if this is reasonable

@pongad 
@vchudnov-g 
@jba 
@jmuk

jba:
I'm in favor of doing this, but there's more to it, isn't there? If the timeout is going to be passed in the context, doesn't the retry logic need to use that instead of the timeout in CallOptions?

saicheems:
The timeout in CallOptions is applied to the context in gax.Invoke
Any timeout applied to the context the user passes in is either ignored, or overrides the default provided:
https://github.com/googleapis/toolkit/blob/master/src/test/java/com/google/api/codegen/testdata/go_library.baseline#L119

jba:
On Tue, May 31, 2016 at 5:13 PM Sai Cheemalapati <notifications@github.com>
wrote:

> The timeout in CallOptions is applied to the context in gax.Invoke
>
But that is the problem — we now want the timeout to propagated via the
context, not the call options.

So if the user does

    ctx, _ = context.WithTimeout(ctx, time.Second)
    client.SomeMethod(ctx, arg)

then gax.Invoke should observe the 1-second timeout from the context.

I don't think we want timeout to be in CallOptions at all.

saicheems:
I see...but if we remove the Timeout option and the user doesn't apply a timeout to the context, then there's no default limit on the call.

The settings design for all other veneers includes a default timeout that the API owner can set for each call - so how can we enforce that without the option?

jba:
How about something like this:

if _, ok := ctx.Deadline(); !ok {
  // no deadline in ctx; use default
  ctx = context.WithTimeout(ctx, defaultTimeoutForMethod)
}

saicheems:
SGTM, I made a pull in gax-golang to change Invoke's behavior to give custom deadlines precedence

jba:
LGTM.

PR #172, opened 2016-06-01
Creating GapicProviderFactory

garrettjonesgoogle:
This change is a net removal of lines! Yay!

This change gives more power for customizing individual cases of code
generation. At the same time, I was able to consolidate repeated
patterns.

* Deleting [language]GapicProvider classes, replacing with a single
  CommonGapicProvider class
* Simplifying [language]_gapic.yaml files: now they only take a factory
  class and an id. This decouples googleapis from changes in the
  parameterization of providers. Additionally, in tests, automatically
  discovering the snippets to generate, so they don't have to be
  repeated in the tests.
* Leaving discovery provider classes as they are for now.
* Replacing SnippetDescriptor with just a String.
* Deleting classes related to FragmentGenerator, which is not used
  currently and is just deadweight.
* Tests: including the generated filename in the baseline file for all
  cases.
* I have started the transition of GAPIC class under
  com.google.api.codegen.gapic. I'm not moving everything now in order
  to not be too disruptive. Likewise, Discovery classes should move
  under com.google.api.codegen.discovery.
* Making sub path logic consistent: It is now computed in each given
  language's [language]GapicContext class.


src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -79,6 +79,9 @@ message ConfigProto {
 
   // The language of the generated code.
   string language = 5;
+  
+  // The configuration to create the providers.
+  GeneratorProto generator = 6;
```
geigerj: Should we remove the ability to configure `codegen_provider`/`input_element_view`/`snippet_files` altogether?

src/main/java/com/google/api/codegen/ApiConfig.java
```diff
@@ -61,6 +61,14 @@ public static ApiConfig createApiConfig(Model model, ConfigProto configProto) {
     }
   }
 
+  /**
+   * Creates an ApiConfig with no content. Exposed for testing.
+   */
+  public static ApiConfig createDummyApiConfig() {
```
geigerj: It looks like this is only used in order to extract the snippet file names from GapicProviderFactory. I wonder if it would be better expose the snippet file names directly, without exposing a test-only function in non-test code.

src/main/java/com/google/api/codegen/gapic/GapicProviderFactory.java
```diff
@@ -0,0 +1,202 @@
...
+
+/**
+ * GapicProviderFactory creates GapicProvider instances based on an id.
+ */
+public class GapicProviderFactory {
+
+  /**
+   * Create the GapicProviders based on the given id
+   */
+  public static List<GapicProvider<? extends Object>> defaultCreate(Model model,
```
geigerj: Java question: why `? extends Object` instead of `?`? Is there a difference?

src/main/java/com/google/api/codegen/gapic/GapicProviderFactory.java
```diff
@@ -0,0 +1,202 @@
...
+
+    } else if (id.equals("nodejs")) {
+      GapicProvider<? extends Object> mainProvider = CommonGapicProvider.<Interface>newBuilder()
+          .setModel(model)
+          .setView(new InterfaceView())
+          .setContext(new NodeJSGapicContext(model, apiConfig))
+          .setSnippetSetRunner(new NodeJSSnippetSetRunner<Interface>())
+          .setSnippetFileNames(Arrays.asList("main.snip"))
+          .build();
+      GapicProvider<? extends Object> clientConfigProvider = CommonGapicProvider.<Interface>newBuilder()
```
geigerj: Instead of repeating the clientConfigProvider instantiation multiple times, maybe refactor this as method?

src/main/java/com/google/api/codegen/gapic/CommonGapicProvider.java
```diff
@@ -0,0 +1,173 @@
...
+
+    // Run the generator for each service.
+    List<GeneratedResult> generated = new ArrayList<>();
+    for (ElementT element : view.getElementIterable(model)) {
+      GeneratedResult result = snippetSetRunner.generate(element, snippetFileName, context);
+
+      String subPath = outputSubPath;
+      if (subPath == null) {
+        // Note on usage of instanceof: there is one case (as of this writing)
+        // where the element is an Iterable<> instead of a ProtoElement.
```
geigerj: Why did we do this? That seems like a misuse of `InputElementView` -- the method is `getElementIterable`, not `getElementIterableIterable`...

src/main/java/com/google/api/codegen/ApiConfig.java
garrettjonesgoogle: The advantage of this route is that the GapicProviderFactory class can be written without worrying about exposing the snippet file names separately. If you expose them separately, it increases the complexity of GapicProviderFactory; information either has to be repeated, or shared variables need to be created.

src/main/java/com/google/api/codegen/gapic/CommonGapicProvider.java
garrettjonesgoogle: The Element is itself an Iterable :-P I don't see it as a misuse.

src/main/java/com/google/api/codegen/gapic/GapicProviderFactory.java
garrettjonesgoogle: The reason is that you can extract the objects out of a parameterized type when you know that it is an Object, but if it is only a ?, you can't even turn the elements into an Object to perform operations on them or cast them to anything. This is necessary in other parts of the code (I don't remember where now).

src/main/java/com/google/api/codegen/gapic/GapicProviderFactory.java
garrettjonesgoogle: I figured that since you are going to parameterize this specific class after this PR is submitted, I would leave that kind refactoring to you as a follow up step. Also, I kind of liked the simplicity and readability of this direct approach.

src/main/proto/com/google/api/codegen/config.proto
garrettjonesgoogle: Discovery still uses them. We can remove them after Discovery is refactored to this new pattern.

garrettjonesgoogle:
add/delete stats for code/yaml files (no testdata files): 1027 added, 1551 removed.

src/main/java/com/google/api/codegen/gapic/GapicProvider.java
```diff
@@ -12,34 +12,31 @@
...
    */
-  public InputElementView<InputElementT> getView();
+  List<String> getSnippetFileNames();
 
   /**
-   * Runs code generation.
+   * Runs code generation and puts the output in outputPath.
    */
-  GeneratedResult generate(InputElementT element, SnippetDescriptor snippetDescriptor);
+  void generate(String outputPath) throws Exception;
```
geigerj: Just a thought: this is not worse that what we had before, but it seems strange that a GapicProvider can implement arbitrary logic to generate and output, instead of creating an iterable of GeneratedResult (where the filename is set correctly), so that the XXXTool can do the outputting in a consistent manner. (I think usually the XXXTool class is intended to handle the direct I/O anyway?)

We don't need to change this now, though.

garrettjonesgoogle: The XXXTool class never handled the I/O itself, that was still always done in the GapicProvider. Previously, the XXXTool class just took the GeneratedResult and piped it back into the output function. The "loop-back" has now been pushed into the GapicProvider itself. I don't feel it's necessary to be one level above it.

src/main/java/com/google/api/codegen/gapic/GapicProviderFactory.java
```diff
@@ -0,0 +1,202 @@
...
+          .setModel(model)
+          .setView(new InterfaceView())
+          .setContext(new ClientConfigGapicContext(model, apiConfig))
+          .setSnippetSetRunner(new ClientConfigSnippetSetRunner<Interface>())
+          .setSnippetFileNames(Arrays.asList("json.snip"))
+          .build();
+
+      return Arrays.<GapicProvider<? extends Object>> asList(mainProvider, messageProvider,
```
geigerj: nit: no space after the parameterized type, here and below. (The Java compiler can't infer the here from the parameters?)

garrettjonesgoogle: Java 7 cannot, but Java 8 can.

src/test/java/com/google/api/codegen/GapicTestBase.java
```diff
@@ -0,0 +1,197 @@
...
+import java.util.regex.Pattern;
+
+import javax.annotation.Nullable;
+
+/**
+ * Base class for code generator baseline tests.
+ */
+public abstract class GapicTestBase extends ConfigBaselineTestCase {
+
+  private static final Pattern BASELINE_PATTERN = Pattern.compile("((?:-|\\w)+)\\[((?:-|\\w)+)\\]");
```
geigerj: Can you put an example or description here of which method names this should match?

geigerj: Ah, I see now this just came from GeneratorTestBase

garrettjonesgoogle: Yes, it does come from GeneratorTestBase, but I do think an example would be helpful :-)

src/test/java/com/google/api/codegen/GapicTestBase.java
```diff
@@ -0,0 +1,197 @@
...
+    ImmutableMap<String, Message> supportedConfigTypes =
+        ImmutableMap.<String, Message>of(
+            ConfigProto.getDescriptor().getFullName(), ConfigProto.getDefaultInstance());
+    ConfigProto configProto =
+        (ConfigProto) MultiYamlReader.read(model, inputNames, inputs, supportedConfigTypes);
+
+    return configProto;
+  }
+
+  public static List<Object[]> createTestedConfigs(String idForFactory, String[] gapicConfigFileNames) {
```
geigerj: Could you add a brief description of this method? The auto-creation of multiple test setups is a little magical, so it would be helpful to note that this is doing the work, and to give some sense of the structure of the return type.

garrettjonesgoogle: Good idea, will do.

geigerj:
Finished first pass, only minor comments.

- Let's make sure there is an assigned issue to get Discovery on the same set up as GAPIC. With this change, there is going to be both redundant configuration and test code, and it'll be much less confusing/error prone to consolidate that.
- I'm not sure how much I like the structure of `GapicProviderFactory.java`. On the one hand, it's simple and easy to read. On the other, it's inflexible and hard to access the data (e.g., the difficulty in extracting the snippet files for testing) -- in that sense, it almost seems like a step back from the storing that data in proto. I'm fine with it as is, but I also still wonder if it's worth giving it a little more structure to get rid of the `createDummyApiConfig`, which still seems awkward to me.

garrettjonesgoogle:
* Filed issue https://github.com/googleapis/toolkit/issues/177
* I wouldn't say that the "inflexibility" is a step back. The inflexibility you speak of only needs to be dealt with in one spot, and reduces the number of code locations that need to be changed to support a new use case. We can discuss more after this PR though.

garrettjonesgoogle:
Made tweaks for your minor feedback, PTAL

geigerj:
LGTM, but please let @bjwatson do a pass as well

src/main/java/com/google/api/codegen/InterfaceListView.java
```diff
@@ -18,26 +18,25 @@
 import com.google.api.tools.framework.model.Model;
 
 import java.util.ArrayList;
-import java.util.List;
 
 /**
  * An interface-based view of model, consisting of a strategy for getting the interfaces of the
  * model and returning them as a single list object.
```
pongad: It's not a list anymore.

src/main/java/com/google/api/codegen/InterfaceListView.java
```diff
@@ -18,26 +18,25 @@
...
  */
-public class InterfaceListView implements InputElementView<List<Interface>> {
+public class InterfaceListView implements InputElementView<Iterable<Interface>> {
 
   /**
    * Gets the reachable interfaces of the model.
    */
   @Override
-  public Iterable<List<Interface>> getElementIterable(Model model) {
+  public Iterable<Iterable<Interface>> getElementIterable(Model model) {
```
pongad: This can be done separately:
`return Collections.singleton(new InterfaceView().getElementIterable(model));` ?

garrettjonesgoogle: True, that would be some nice cleanup work to do separate from this PR.

src/main/java/com/google/api/codegen/InterfaceListView.java
garrettjonesgoogle: I'll adjust the doc.

bjwatson:
@garrettjonesgoogle Can you merge master? There's a conflict in src/test/java/com/google/api/codegen/testdata/python_library.baseline.

src/main/proto/com/google/api/codegen/config.proto
bjwatson: Could you add a TODO comment to indicate this intention, so we don't lose track of it?

src/main/java/com/google/api/codegen/ApiConfig.java
bjwatson: @geigerj More context: https://docs.oracle.com/javase/tutorial/java/generics/unboundedWildcards.html

src/main/java/com/google/api/codegen/gapic/GapicProviderFactory.java
bjwatson: @geigerj More context: https://docs.oracle.com/javase/tutorial/java/generics/unboundedWildcards.html

src/main/proto/com/google/api/codegen/config.proto
garrettjonesgoogle: Will do.

garrettjonesgoogle:
PTAL - I have merged from master and addressed a couple items of feedback.

bjwatson:
@garrettjonesgoogle I'm not as familiar with this code as you and @geigerj are, but I generally like what you're trying to achieve here. I agree that we should minimize references to Java objects in the YAML files. We do we even need the one remaining reference to `com.google.api.codegen.gapic.GapicProviderFactory`, since it's the same in all cases?

I second Jacob's concern about the current structure of GapicProviderFactory; I would add that it violates the Open/Closed principle: https://en.wikipedia.org/wiki/Open/closed_principle

That's not a big deal right now with our focus on the seven languages, but it will make the architecture less flexible for third-party developers who want to add support for more languages. We can tackle that issue at a later time. The correct architecture is not immediately obvious to me.

I'm going to check how it merges with my open branch before giving my final sign-off.

src/main/java/com/google/api/codegen/gapic/GapicProviderFactory.java
```diff
@@ -0,0 +1,202 @@
...
+
+  /**
+   * Create the GapicProviders based on the given id
+   */
+  public static List<GapicProvider<? extends Object>> defaultCreate(Model model,
+      ApiConfig apiConfig, String id) {
+
+    // Please keep the following IDs in alphabetical order
+
+    if (id.equals("client_config")) {
```
bjwatson: Can you move all these "magic numbers" to a list of constants?

src/main/java/com/google/api/codegen/gapic/GapicProviderFactory.java
```diff
@@ -0,0 +1,202 @@
...
+          .build();
+      return Arrays.<GapicProvider<? extends Object>>asList(provider);
+
+    } else if (id.equals("go")) {
+      GapicProvider<? extends Object> provider = CommonGapicProvider.<Interface>newBuilder()
+          .setModel(model)
+          .setView(new InterfaceView())
+          .setContext(new GoGapicContext(model, apiConfig))
+          .setSnippetSetRunner(new GoSnippetSetRunner<Interface>())
+          .setSnippetFileNames(Arrays.asList("main.snip", "example.snip", "doc.snip"))
```
bjwatson: It's arguable whether or not these are magic numbers. Maybe a better system would be to just scan the appropriate source directory for all .snip files. Just a thought for the future, rather than something to implement right now.

garrettjonesgoogle:
I have a single answer to your two seemingly unrelated comments: We need the remaining reference to the factory in order to enable non-Gapic cases to use their own factory without having to "open" up GapicProviderFactory (e.g. the discovery case will probably use its own factory). This design enables external usage cases where someone has only one use case; they can make a factory that only supports one id. Really, you can mix and match however you want.

src/main/java/com/google/api/codegen/gapic/GapicProviderFactory.java
garrettjonesgoogle: I don't think that really buys us anything; it creates an extra layer of indirection for something that will fail fast if it's wrong (the system won't even start).

src/main/java/com/google/api/codegen/gapic/GapicProviderFactory.java
garrettjonesgoogle: Scanning the source directory for snip files wouldn't work for our current file arrangement, because multiple use cases have snip files in the same directory and sometimes share a common.snip file. 

I was aiming for a "specify these files only once" solution: this is the only location where these files are named.

bjwatson:
I was suspecting the answer was something like that.

However, it looks like CodeGeneratorApi.process() currently assumes that the factory is of type GapicProviderFactory. Seems like this should instead refer to an Interface declaring the create() method, and GapicProviderFactory implements that Interface.

src/main/java/com/google/api/codegen/gapic/GapicProviderFactory.java
bjwatson: Naked strings scattered throughout the code are almost never a good thing (https://en.wikipedia.org/wiki/Magic_number_(programming)#Unnamed_numerical_constants).

Turning these into constants buys us an easy-to-read table of contents for expected values. It also encourages others to use the same constants, rather than hand-writing the same or similar strings (especially any other code that might ever call into this specific method, such as tests).

Having the compiler catch such errors is even better than failing at start time.

src/main/java/com/google/api/codegen/gapic/GapicProviderFactory.java
bjwatson: I thought of that, too. For languages like Java and Python, the files would need to be partitioned by sub-directories for the scanning solution to work.

This is okay. It's just one more thing a developer needs to remember to do when adding a .snip file, and it's in a different location from where they added the file.

src/main/java/com/google/api/codegen/gapic/GapicProviderFactory.java
garrettjonesgoogle: I understand the motivation behind string constants, and I have used a lot of them and encouraged others to do so too over the years. The thing here is that the only non-test location where these values are used is in non-Java files, which won't get verified until start time anyway. 

But anyway, since this is a shared codebase, I will be nice and make constants out of them.

garrettjonesgoogle:
Hmm you are correct about the type assumption. I have renamed GapiProviderFactory to MainGapiProviderFactory and created an interface named GapiProviderFactory. I also made the constants change. PTAL.

src/main/java/com/google/api/codegen/gapic/MainGapicProviderFactory.java
```diff
@@ -0,0 +1,202 @@
...
+
+import java.util.Arrays;
+import java.util.List;
+
+import org.apache.commons.lang3.NotImplementedException;
+
+/**
+ * GapicProviderFactory creates GapicProvider instances based on an id.
+ */
+public class MainGapicProviderFactory {
```
bjwatson: Shouldn't this implement `GapicProviderFactory<GapicProvider<? extends Object>>`?

src/main/java/com/google/api/codegen/CodeGeneratorApi.java
```diff
@@ -123,27 +121,26 @@ protected void process() throws Exception {
...
-        GapicProviderFactory.class,
-        new Class<?>[] {},
-        new Object[] {},
-        "generator", new GeneratorBuilderUtil.ErrorReporter() {
-        @Override
-        public void error(String message, Object... args) {
-          model.addDiag(Diag.error(SimpleLocation.TOPLEVEL, message, args));
-        }
-      });
+    GapicProviderFactory<GapicProvider<? extends Object>> provider = GeneratorBuilderUtil
```
bjwatson: Seems like the names `GapicProviderFactory` and `GapicProvider` are tying us to a specific set of use cases (e.g. not discovery sample gen). I'm wondering if better names might be `CodeGenProviderFactory` and `CodeGenProvider`, or just `ProviderFactory` and `Provider`?

src/main/java/com/google/api/codegen/gapic/GapicProviderFactory.java
bjwatson: Thanks @garrettjonesgoogle.

bjwatson:
Thanks. The structure looks good. I made a couple comments on the details.

bjwatson:
Also, the merge with my snippet gen branch (https://github.com/googleapis/toolkit/pull/149) seems relatively easy.

src/main/java/com/google/api/codegen/gapic/MainGapicProviderFactory.java
garrettjonesgoogle: Oops! I'll fix that.

src/main/java/com/google/api/codegen/CodeGeneratorApi.java
garrettjonesgoogle: The big problem is that Gapic and Discovery take different inputs, so their create methods can't have the same signature; Gapic uses ApiConfig, and Discovery uses ApiaryConfig. If you check over the existing Gapic and Discovery classes, you'll see that they don't share hierarchies for this reason.

garrettjonesgoogle:
I added the "implements" qualifier on MainGapicProviderFactory, and also used the language constants in the tests. PTAL

src/main/java/com/google/api/codegen/CodeGeneratorApi.java
bjwatson: I see. Good point. I guess it's best to wait until the Discovery case is added before thinking about any additional abstraction.

src/test/java/com/google/api/codegen/CSharpCodeGeneratorTest.java
```diff
@@ -43,7 +45,7 @@ public CSharpCodeGeneratorTest(
    */
   @Parameters(name = "{0}")
   public static List<Object[]> testedConfigs() {
-    return GapicTestBase.createTestedConfigs("csharp",
+    return GapicTestBase.createTestedConfigs(MainGapicProviderFactory.CSHARP,
```
bjwatson: Sweet! Good catch.

bjwatson:
LGTM

bjwatson:
@geigerj PTAL

geigerj:
LGTM

Could you update the `factory` references in googleapis/googleapis#38?

garrettjonesgoogle:
Updated https://github.com/googleapis/googleapis/pull/38 .

PR #173, opened 2016-06-01
Fix casting error

saicheems:


saicheems:
@jba 
@vchudnov-g

saicheems:
It doesn't show in the diff, but the specific error is the comparison
`iterator.currentIndex <= len(iterator.items)`

The length has to be casted to match int32, but I figure changing the type of currentIndex is a sufficient fix.

jba:
Yes, anything that's an array or slice index should be of type int.

LGTM.

PR #174, opened 2016-06-01
Change package path for client examples

### LINKED ISSUE #105: Go method samples should be in a unique package
`NewClient` should be called on the package name so users can copy paste code. The issue right now is the generator doesn't know the full import path.

saicheems:
Allows examples embedded in godoc to be copy-pastable.

Also fixes bug in `SetPageSize` method signature (unfortunate
side-effect of switching between Java and Go all day...)

Resolves #105

### DIFF
--- a/src/main/java/com/google/api/codegen/go/GoGapicContext.java
+++ b/src/main/java/com/google/api/codegen/go/GoGapicContext.java
@@ -336,6 +336,13 @@ public class GoGapicContext extends GapicContext implements GoContext {
   }
 
   /**
+   * Returns the package path for the client.
+   */
+  public String getClientPackagePath(Interface service) {
+    return getApiConfig().getPackageName() + "/generated/" + getReducedServiceName(service);
+  }
+
+  /**
    * Creates a Go import from the message import.
    */
   private GoImport createMessageImport(MessageType messageType) {
@@ -428,6 +435,7 @@ public class GoGapicContext extends GapicContext implements GoContext {
   public Iterable<String> getTestImports(Interface service) {
     TreeSet<GoImport> thirdParty = new TreeSet<>();
 
+    thirdParty.add(GoImport.create(getClientPackagePath(service)));
     thirdParty.add(GoImport.create("golang.org/x/net/context"));
     thirdParty.add(GoImport.create(GAX_PACKAGE_BASE, "gax"));
     thirdParty.addAll(getMessageImports(service));
--- a/src/test/java/com/google/api/codegen/testdata/go_library.baseline
+++ b/src/test/java/com/google/api/codegen/testdata/go_library.baseline
@@ -563,7 +563,7 @@ func (it *BookIterator) Next() (*google_example_library_v1.Book, error) {
 
 // SetPageSize sets the maximum size of the next page to be
 // retrieved.
-func (it *BookIterator) SetPageSize(int32 pageSize) {
+func (it *BookIterator) SetPageSize(pageSize int32) {
     it.pageSize = pageSize
 }
 
@@ -618,7 +618,7 @@ func (it *StringIterator) Next() (string, error) {
 
 // SetPageSize sets the maximum size of the next page to be
 // retrieved.
-func (it *StringIterator) SetPageSize(int32 pageSize) {
+func (it *StringIterator) SetPageSize(pageSize int32) {
     it.pageSize = pageSize
 }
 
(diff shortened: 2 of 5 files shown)


saicheems:
Resolves #105

saicheems:
@jba 
@pongad 
@jmuk 
@vchudnov-g

pongad:
Write "Fixes #105" in the commit message. GitHub will auto close the issue. I don't think it works if the message is just written in the thread.

saicheems:
You're right, thanks

src/main/java/com/google/api/codegen/go/GoGapicContext.java
```diff
@@ -428,6 +435,7 @@ private GoImport createMessageImport(MessageType messageType) {
   public Iterable<String> getTestImports(Interface service) {
     TreeSet<GoImport> thirdParty = new TreeSet<>();
 
+    thirdParty.add(GoImport.create(getClientPackagePath(service)));
     thirdParty.add(GoImport.create("golang.org/x/net/context"));
```
pongad: Not for this PR, but something to keep in mind. Go 1.7 is adding context to stdlib. This might need to change in the future.

src/test/java/com/google/api/codegen/testdata/go_library.baseline
```diff
@@ -635,54 +635,52 @@ func (it *StringIterator) NextPageToken() string {
...
-    client, err := NewClient(ctx, opts...)
-    _, _ = client, err
+    c, err := library.NewClient(ctx, opts...)
+    _, _ = c, err
 }
 
 func ExampleClient_CreateShelf() {
     ctx := context.Background()
-    client, _ := NewClient(ctx)
+    c, _ := library.NewClient(ctx)
```
pongad: We might want to check error. I think something like

```
c, err := library.NewClient(ctx)
_ = err // Handle error
```
should be appropriate

src/test/java/com/google/api/codegen/testdata/go_library.baseline
```diff
@@ -635,54 +635,52 @@ func (it *StringIterator) NextPageToken() string {
...
 
     req := &google_example_library_v1.CreateShelfRequest{ /* Data... */ }
-    opts := []gax.CallOption{ /* Optional call parameters. */ }
     var resp *google_example_library_v1.Shelf
-    resp, _ = client.CreateShelf(ctx, req, opts...)
+    resp, _ = c.CreateShelf(ctx, req)
```
pongad: Check error.

src/test/java/com/google/api/codegen/testdata/go_library.baseline
```diff
@@ -635,54 +635,54 @@ func (it *StringIterator) NextPageToken() string {
     return it.nextPageToken
 }
 ============== file: google.golang.org/cloud/library/generated/library/client_test.go ==============
-package library
+package library_test
 
 import (
     gax "github.com/googleapis/gax-golang"
     "golang.org/x/net/context"
     google_example_library_v1 "google.golang.org/cloud/library/generated/google/example/library/v1"
```
pongad: Not an immediate concern, but when we figure out the import path, we should give different names to the packages. It would be unfortunate if you have two packages called "library" and you always need both of them.

saicheems: To the proto packages?

src/test/java/com/google/api/codegen/testdata/go_library.baseline

src/test/java/com/google/api/codegen/testdata/go_library.baseline

src/test/java/com/google/api/codegen/testdata/go_library.baseline
pongad: I'm not sure how protobuf compiler chooses the name. However, if the proto package is called "library", we probably shouldn't also call the client library "library" unless we want to put the two in the same package.

pongad:
lgtm

src/test/java/com/google/api/codegen/testdata/go_library.baseline
saicheems: Gotcha. If I'm not mistaken, the proto compiler chooses the last component of the `go_package` string, so in this case it chooses `v2`.

I'm not sure if there is another option that can override the actual package name... I haven't seen it yet.

PR #178, opened 2016-06-02
User friendly strings for other languages

pongad:


src/main/java/com/google/api/codegen/php/PhpDiscoveryContext.java
```diff
@@ -166,19 +171,16 @@ public String typeDefaultValue(Type type, Field field) {
...
-            case "date":
-              return "\'1969-12-31\'";
-            case "date-time":
-              return "\'"
-                  + new DateTime(0L).toStringRfc3339()
-                  + "\'";
-            default:
-              // Fall through
-          }
+        if (STRING_DEFAULT_MAP.containsKey(stringFormat)) {
```
vchudnov-g: Is there a way to factor out the common cross-language code? One way would be to have the parent class implement typeDefaultValue(), and have it call methods that are overridden in the child classes to handle any initial exceptions for that language and to handle getting the correct empty string or null literal for that language. There are other ways, too.

pongad:
Some code paths are not used. I deleted them for the time being. We'll have to revisit them when improving defaults for struct fields. The implementation will most likely be significantly different from the one deleted.

vchudnov-g:
OK, sounds good. I'd like to minimize repeated code while maintaining clarity, but for anything that's not small, better to do it in a subsequent change.

LGTM

PR #181, opened 2016-06-02
Fix NodeJS page-streaming method

jmuk:
- As with https://github.com/googleapis/gax-nodejs/pull/14, callback
  for page-streaming methods are ignored. So callback parameters
  should be removed, and it should always return stream.
- Update the comment of stream to mention per-page streaming feature.
- Transfer the 'end' event. So it can finish repeating API calls
  when the user wants to end the stream.


geigerj:
LGTM

PR #191, opened 2016-06-07
Flatten service hierarchy

### LINKED ISSUE #187: Go: make the top-of-file editing message clearer.
The current one suggests that some edits can be made. It even ends with "Happy editing!".

I'm particularly concerned about giving people false hope, since these files will live under handwritten, editable code.

So can we change everything from "EDITING INSTRUCTIONS" to the end of the comment to:

AUTO-GENERATED CODE. DO NOT EDIT.

### LINKED ISSUE #188: Go: handle multi-service APIs better
Currently an API that has multiple proto services (e.g. logging) generates three separate packages, with three separate clients.

The most immediate problem is that each client always creates its own gRPC connection. We should be able to reuse the same connection for all clients, since they are all talking to the same endpoint.

But I think we should generate one VKit package per API, rather than per service. The HTTP surface does not distinguish by service, so using the VKit would be less pleasant than using the API directly in this respect, which isn't desirable.

saicheems:
Resolves #187, resolves #188.

### DIFF
--- a/src/main/java/com/google/api/codegen/go/GoGapicContext.java
+++ b/src/main/java/com/google/api/codegen/go/GoGapicContext.java
@@ -18,6 +18,7 @@ import com.google.api.Documentation;
 import com.google.api.codegen.ApiConfig;
 import com.google.api.codegen.CollectionConfig;
 import com.google.api.codegen.GapicContext;
+import com.google.api.codegen.InterfaceConfig;
 import com.google.api.codegen.LanguageUtil;
 import com.google.api.codegen.MethodConfig;
 import com.google.api.codegen.PageStreamingConfig;
@@ -101,6 +102,36 @@ public class GoGapicContext extends GapicContext implements GoContext {
     this.goCommon = new GoContextCommon();
   }
 
+  /**
+   * Returns the Go package name.
+   *
+   * Retrieved by splitting the package string on `/` and returning the second to last string if
+   * possible, and otherwise the last string in the resulting array.
+   *
+   * TODO(saicheems): Figure out how to reliably get the intended value...
+   */
+  public String getPackageName() {
+    String split[] = getApiConfig().getPackageName().split("/");
+    if (split.length - 2 >= 0) {
+      return split[split.length - 2];
+    }
+    return split[split.length - 1];
+  }
+
+  /**
+   * Returns the name of the service's client.
+   */
+  public String getClientName(Interface service) {
+    String name = getReducedServiceName(service);
+    return LanguageUtil.lowerUnderscoreToUpperCamel(name) + "Client";
+  }
+
+  /**
+   * Returns the service name with common suffixes removed.
+   *
+   * For example:
+   *  LoggingServiceV2 => logging
+   */
   public String getReducedServiceName(Interface service) {
     String name = service.getSimpleName().replaceAll("V[0-9]+$", "");
     name = name.replaceAll("Service$", "");
@@ -336,13 +367,6 @@ public class GoGapicContext extends GapicContext implements GoContext {
   }
 
   /**
-   * Returns the package path for the client.
-   */
-  public String getClientPackagePath(Interface service) {
-    return getApiConfig().getPackageName() + "/" + getReducedServiceName(service);
-  }
-
-  /**
    * Creates a Go import from the message import.
    */
   private GoImport createMessageImport(MessageType messageType) {
@@ -365,9 +389,6 @@ public class GoGapicContext extends GapicContext implements GoContext {
 
     if (!getApiConfig().getInterfaceConfig(service).getRetrySettingsDefinition().isEmpty()) {
       standardImports.add(GoImport.create("time"));
-    }
-    if (hasPageStreamingMethod(service)) {
-      standardImports.add(GoImport.create("errors"));
     }
     return standardImports;
   }
@@ -442,7 +463,7 @@ public class GoGapicContext extends GapicContext implements GoContext {
   public Iterable<String> getTestImports(Interface service) {
     TreeSet<GoImport> thirdParty = new TreeSet<>();
 
-    thirdParty.add(GoImport.create(getClientPackagePath(service)));
+    thirdParty.add(GoImport.create(getApiConfig().getPackageName()));
     thirdParty.add(GoImport.create("golang.org/x/net/context"));
     thirdParty.add(GoImport.create(GAX_PACKAGE_BASE, "gax"));
     thirdParty.addAll(getMessageImports(service));
@@ -451,10 +472,35 @@ public class GoGapicContext extends GapicContext implements GoContext {
   }
 
   /**
-   * Returns true if the speficied type is the Empty type.
+   * Returns true if the specified type is the Empty type.
    */
   public boolean isEmpty(TypeRef type) {
     return type.isMessage() && type.getMessageType().getFullName().equals("google.protobuf.Empty");
+  }
+
+  /**
+   * Returns true if the methodConfig's retry codes name refers to an empty list of retry codes.
+   */
+  public boolean hasEmptyRetryCodes(Interface service, MethodConfig methodConfig) {
+    InterfaceConfig config = getApiConfig().getInterfaceConfig(service);
+    for (String name : config.getRetryCodesDefinition().keySet()) {
+      if (name.equals(methodConfig.getRetryCodesConfigName())) {
+        return config.getRetryCodesDefinition().get(name).size() == 0;
+      }
+    }
+    return false;
+  }
+
+  /**
+   * Returns true if the API has a page streaming method.
+   */
+  public boolean hasPageStreamingMethod() {
+    for (Interface service : getModel().getSymbolTable().getInterfaces()) {
+      if (hasPageStreamingMethod(service)) {
+        return true;
+      }
+    }
+    return false;
   }
 
   /**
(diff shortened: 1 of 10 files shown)


src/main/java/com/google/api/codegen/go/GoGapicContext.java
```diff
@@ -101,6 +101,34 @@ public GoGapicContext(Model model, ApiConfig apiConfig) {
     this.goCommon = new GoContextCommon();
   }
 
+  /**
+   * Returns the Go package name.
+   *
+   * Retrieved by splitting the package string on `/` and returning the second to last string if
+   * possible, and otherwise the last string in the resulting array.
+   */
+  public String getPackageName() {
```
saicheems: This is slightly ugly... the trick is getting the package name for the service to be `logging` instead of `apiv2`, and the only way I could think of is to pull the second to last string of the import path.

saicheems:
@broady
@mcgreevy
@okdave

src/test/java/com/google/api/codegen/testdata/go_library.baseline
```diff
@@ -159,15 +127,15 @@ func NewClient(ctx context.Context, opts ...gax.ClientOption) (*Client, error) {
...
+func (c *LibraryClient) Close() error {
     return c.conn.Close()
 }
 
 // Path templates.
 
 // ShelfPath returns the path for the shelf resource.
-func ShelfPath(shelf string) string {
-    path, err := shelfPathTemplate.Render(map[string]string{
+func (c *LibraryClient) ShelfPath(shelf string) string {
```
saicheems: I've moved path template methods back under the client struct because they can conflict with other services. Each service defines a set of template strings which are available to the generator, which operates on a per-service basis. To prevent conflicts, the private vars are made unique by prefixing with the service name.

A possible solution to avoid this is to have a separate file that contains path template methods, which ignores equivalent template strings defined by multiple services (for example, 2 services could define a project template). I'm not sure about doing it that way though... on intuition I think it might be more confusing.

src/test/java/com/google/api/codegen/testdata/go_library.baseline
```diff
@@ -810,43 +794,40 @@ func ExampleClient_ListStrings() {
...
-//
-// Happy editing!
 
-// Package library provides the more language idiomatic and
-// developer friendly layer for the gRPC based API service.
-//
 // A simple Google Example Library API.
 package library
 
+import "errors"
```
saicheems: For now I've added the common constant and vars to `doc.go`.

src/test/java/com/google/api/codegen/testdata/go_library.baseline
```diff
@@ -810,43 +794,40 @@ func ExampleClient_ListStrings() {
...
-// preserve those manual additions. In order to notbreak the refresh
-// process, only certain types of modifications are allowed.
-//
-// Happy editing!
 
-// Package library provides the more language idiomatic and
-// developer friendly layer for the gRPC based API service.
-//
 // A simple Google Example Library API.
 package library
```
broady: There are multiple package level comments.

src/test/java/com/google/api/codegen/testdata/go_library.baseline
```diff
@@ -137,17 +105,17 @@ type Client struct {
     metadata map[string][]string
 }
 
-// NewClient creates a new API service client.
-func NewClient(ctx context.Context, opts ...gax.ClientOption) (*Client, error) {
-    s := defaultClientSettings()
+// NewLibraryClient creates a new API service client.
```
broady: Should this mention something about the library service?

src/test/java/com/google/api/codegen/testdata/go_library.baseline
```diff
@@ -634,26 +602,42 @@ func (it *StringIterator) SetPageToken(token string) {
 func (it *StringIterator) NextPageToken() string {
     return it.nextPageToken
 }
-============== file: google.golang.org/cloud/library/apiv1/library/client_test.go ==============
+============== file: google.golang.org/cloud/library/apiv1/library_client_test.go ==============
```
broady: Change filename to use the suffix `_example_test.go`

src/test/java/com/google/api/codegen/testdata/go_library.baseline
```diff
@@ -810,43 +794,40 @@ func ExampleClient_ListStrings() {
...
-// process will only be runnable by Google engineers. Manual additions are
-// allowed because the refresh process performs a 3-way merge in order to
-// preserve those manual additions. In order to notbreak the refresh
-// process, only certain types of modifications are allowed.
-//
-// Happy editing!
 
-// Package library provides the more language idiomatic and
-// developer friendly layer for the gRPC based API service.
-//
```
broady: What happened to the comment we came up with yesterday?

`Package library is an experimental, auto-generated ....`

src/main/java/com/google/api/codegen/go/GoGapicContext.java
jba: Isn't the name of the API represented somewhere?

src/test/java/com/google/api/codegen/testdata/go_library.baseline
saicheems: I'll follow up in another CL for the doc strings, didn't want to make too many changes in this one.

src/test/java/com/google/api/codegen/testdata/go_library.baseline
jba: The problem is, there is nothing the client provides to this operation. You're using it purely for scoping.

How about concatenating the service and resource names: LibraryShelfPath.

src/main/resources/com/google/api/codegen/go/doc.snip
```diff
@@ -1,39 +1,41 @@
...
-    // here through a refresh process. For the short term, the refresh
-    // process will only be runnable by Google engineers. Manual additions are
-    // allowed because the refresh process performs a 3-way merge in order to
-    // preserve those manual additions. In order to notbreak the refresh
-    // process, only certain types of modifications are allowed.
-    //
-    // Happy editing!
+@snippet generateBody(service)
+    const (
+        gapicNameVersion = "gapic/0.1.0"
```
saicheems: It's not ready in the generator yet... this is a placeholder until it is.

src/main/java/com/google/api/codegen/go/GoGapicContext.java
saicheems: No, unfortunately, not consistently. I think this will be fine temporarily. I think the only way around it is to add a field to the config, but it may also be a go only issue.
Will follow up in this.

src/test/java/com/google/api/codegen/testdata/go_library.baseline
saicheems: Yeah that makes sense, not sure why I didn't think that was a reasonable solution.
Will do.

src/test/java/com/google/api/codegen/testdata/go_library.baseline
saicheems: Actually I'll just do it now

src/test/java/com/google/api/codegen/testdata/go_library.baseline

src/test/java/com/google/api/codegen/testdata/go_library.baseline
saicheems: Fixed here:
https://github.com/googleapis/toolkit/pull/191/files#diff-c43de7347f2c96dbfdffd7320ce2e1d9R16

src/test/java/com/google/api/codegen/testdata/go_library.baseline

src/test/java/com/google/api/codegen/testdata/go_main_library.baseline
```diff
@@ -137,17 +93,30 @@ type Client struct {
     metadata map[string][]string
 }
 
-// NewClient creates a new API service client.
-func NewClient(ctx context.Context, opts ...gax.ClientOption) (*Client, error) {
-    s := defaultClientSettings()
+// NewLibraryClient creates a new library service client.
+//
+// This API represents a simple digital library.  It lets you manage Shelf
```
saicheems: @broady 

These comments used to be at the package level for each service. I tried moving them here so they're not lost, so let me know if you want them removed or it they're worth keeping on the constructor.

src/test/java/com/google/api/codegen/testdata/go_library.baseline

src/test/java/com/google/api/codegen/testdata/go_library.baseline

saicheems:
Updated with `<package_name>.go` that contains the common vars.

jba:
Can you put the doc in `<package_name>.go` if there is only one service, and in front of NewXXXClient if there are more than one?

saicheems:
Hmmm that should be possible, which doc are you referring to? The service comments?

There are two different levels of documentation in the proto iirc, there's the package level that I put above the declaration in `doc.go`, and the service level documentation I put on the client constructor in `xxx_client.go`

jba:
The problem is that I don't think all APIs follow a rule about those two kinds of doc (if there even is a rule). For instance, the library example puts the overall API doc on the service declaration (that is, there is no blank line between the doc comment and the service line). You're putting that comment on NewClient (that's another point, it should be called NewClient if there's only one) instead of in doc.go.

None of this is a showstopper, but if you could recognize whether there was more than one service in an API I think this would get better -- for most APIs it would be as it was before, with NewClient and the doc in doc.go.

saicheems:
I think I understand, it should be possible to do
I'll follow up in another PR - this one is starting to get bloated

jba:
OK, then LGTM.

PR #192, opened 2016-06-07
Fix JSON client config location

### LINKED ISSUE #126: clientconfig generates json files at toplevel
When I built a gapic client for logging/v2 API, I've noticed that the generated place for clientconfig is different from the code location. client config json files are always generated at the toplevel directory, while py files are in google/logging/v2 (rb files are lib/google/logging/v2).

But our generated code expects the JSON files in the directory where the code files reside.

This means client-config target directory has to be injected by some language-specific parameters -- I don't have concrete ideas to achieve this.

geigerj:
Previously the client config was generated at the root of the output.
This change adds custom GAPIC contexts to specify the proper output
location.

Fixes #126 
cc: @jmuk


src/test/java/com/google/api/codegen/testdata/ruby_json_library.baseline
```diff
@@ -1,4 +1,4 @@
-============== file: library_service_client_config.json ==============
+============== file: google/example/library/v1/library_service_client_config.json ==============
```
jmuk: Well, this does not look correct.  Probably not by this CL, but something went wrong during some refactoring...

Ruby code should be generated under lib/ directory. So this is lib/google/example/library/v1/library_service_client_config.json (and library_service_api.rb place is wrong right now)

jmuk: Filed https://github.com/googleapis/toolkit/issues/193 separately.

geigerj: Good catch. I have rebased on your fix.

src/main/java/com/google/api/codegen/gapic/MainGapicProviderFactory.java
```diff
@@ -64,139 +67,155 @@
...
 
     if (id.equals(CLIENT_CONFIG)) {
-      GapicProvider<? extends Object> provider = CommonGapicProvider.<Interface>newBuilder()
-          .setModel(model)
-          .setView(new InterfaceView())
-          .setContext(new ClientConfigGapicContext(model, apiConfig))
-          .setSnippetSetRunner(new ClientConfigSnippetSetRunner<Interface>())
-          .setSnippetFileNames(Arrays.asList("json.snip"))
-          .build();
+      GapicProvider<? extends Object> provider =
```
garrettjonesgoogle: I'd prefer the reformatting to be done in a different change - @pongad 's PR https://github.com/googleapis/toolkit/pull/175 will do this. Let's get his submitted before yours and rebase.

src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContextUtil.java
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen.nodejs;
+
+/**
+ * Utility methods for NodeJS GAPIC contexts.
+ */
+public class NodeJSGapicContextUtil {
```
garrettjonesgoogle: I don't think this functionality merits its own class. In general, I'd like to avoid an explosion of classes.

src/main/java/com/google/api/codegen/nodejs/NodeJSClientConfigGapicContext.java
```diff
@@ -0,0 +1,31 @@
...
+import com.google.api.tools.framework.model.ProtoElement;
+
+public class NodeJSClientConfigGapicContext extends ClientConfigGapicContext {
+  public NodeJSClientConfigGapicContext(Model model, ApiConfig config) {
+    super(model, config);
+  }
+
+  @Override
+  public String getOutputSubPath(ProtoElement element) {
+    return NodeJSGapicContextUtil.getOutputSubPath();
```
garrettjonesgoogle: CommonGapicProvider has a way to override the output sub path when it's being built. I think we should prefer that over making a new subclass just to change this one method's behavior.

garrettjonesgoogle:
Maybe we should just bite the bullet and add a "CodePathMapper" extraction. There are a couple ways we could do this - an interface plus a couple implementations, or a single implementations that can handle all of the cases.

Multiple implementations:
* FixedCodePathMapper: always returns the same string (NodeJS case - "lib", Go case - "", PHP - "")
* PackageCodePathMapper: returns the package path with '.' translated to '/' (Python, Java); maybe Ruby could fit here too (it seems to map '::' to '/')
* ProtoElementNamespacePathMapper: gets the namespace of the proto element, maps '.' to '/'

Single implementation:
* Fixed base
* Package path mapper
* Proto elem path mapper

Those three elements would get combined together for the complete path. 

This code path mapper would be passed into the GapicProvider at construction time in MainGapicProviderFactory. 

Thoughts?

src/main/java/com/google/api/codegen/gapic/MainGapicProviderFactory.java
geigerj: SGTM

geigerj:
I'll try doing this. I think what you suggested could work well if it can replace having "getOutputSubPath" as part of the context altogether, so that we don't need CommonGapicProvider to override the context.

My original concern with creating strategy objects was that I'm not sure it's significantly better in terms of "explosion of classes", since we have almost as many strategies (five) as languages (seven) -- they're, for the most part, single-use classes. From a readability point-of-view, it's harder to see exactly why each strategy exists than if you just override `getOutputSubPath` in the context.

garrettjonesgoogle:
Yes, it would replace getOutputSubPath from the context completely.

I was also initially hesitant to go the route of strategy objects for this problem, until I realized that there are only 3 (4?) algorithms in use for 7 languages (not five strategies...). 

W.r.t. overriding methods, if you overrode methods for every case where you needed differing functionality, you would end up with a combinatorial explosion of classes. If you have a strategy that is general and useful enough, you can re-use it in new contexts without having to create a new class. You cannot re-use code in classes that just exist to override other classes.

geigerj:
Updated; I rebased on the reformatting commit (there is still some minor noise from the difference between the Eclipse autoformat plugin and the formatter we use in the CI), and switched to a strategy object-based approach.

In the end I went with the "single implementation" route, which made things a little simpler for Ruby, which uses both a fixed prefix and and package-based directory structure. The C# case gets its own implementation, since it relies on a C#-specific proto annotation.

src/main/java/com/google/api/codegen/py/PythonContextCommon.java
```diff
@@ -181,4 +181,16 @@ public String wrapIfKeywordOrBuiltIn(String name) {
               "zip",
               "__import__")
           .build();
+
+  /**
+   * Converts the dot-separated Python package name from the GAPIC config to a slash-separated
+   * directory structure.
+   */
+  public String getOutputSubPath(String packageName) {
```
garrettjonesgoogle: Is this still needed here?

geigerj: Removed.

garrettjonesgoogle:
LGTM

PR #194, opened 2016-06-07
Fix the basepath for generated .rb code.

### LINKED ISSUE #193: Ruby generated code location seems wrong
It seems Ruby code is generated directly under the output directory ({OUTPUT}/google/logging/v2/logging_service_v2_api.rb, for example).

This is not intended, actually 'lib' is necessary, like {OUTPUT}/**lib**/google/logging/v2/logging_service_v2_api.rb.

jmuk:
The ruby code is generated under 'lib' directory.
Fixes #193.


geigerj:
LGTM

PR #198, opened 2016-06-08
C# discovery snippet generation

chrisdunelm:


src/main/java/com/google/api/codegen/csharp/CSharpContextCommon.java
```diff
@@ -31,11 +31,7 @@
    */
   // TODO(jonskeet): Handle naming collisions.
   private final TreeSet<String> imports = new TreeSet<>();
-  private final String serviceNamespace;
-
-  public CSharpContextCommon(String serviceNamespace) {
-    this.serviceNamespace = serviceNamespace;
-  }
+  //private final String serviceNamespace;
```
jskeet: Delete commented-out field?

src/main/java/com/google/api/codegen/csharp/CSharpDiscoveryContext.java
```diff
@@ -0,0 +1,232 @@
...
+  }
+
+  public String addImport(String namespace) {
+    return csharpCommon.addImport(namespace);
+  }
+  
+  private String packageNameAndImport() {
+    String packageName = CSharpContextCommon.s_underscoresToPascalCase(
+        getApiaryConfig().getServiceCanonicalName());
+    String versionName = getApiaryConfig().getServiceVersion().replace('.', '_');
```
jskeet: I assume this is following the existing codegen? Would prefer to stick to just major version numbers in the namespace, but we're not changing the past. (Just checking that we're on the same page.)

src/main/java/com/google/api/codegen/csharp/CSharpSnippetSetRunner.java
```diff
@@ -48,23 +49,20 @@ public GeneratedResult generate(
             snippetFileName,
             ImmutableMap.<String, Object>of("context", context));
 
-    // TODO don't depend on a cast here
-    CSharpGapicContext csharpContext = (CSharpGapicContext) context;
-
     String outputFilename = snippets.generateFilename(element).prettyPrint();
+    CSharpContextCommon csharpContextCommon = new CSharpContextCommon();
```
jskeet: I confess I have very little idea what these different contexts are about at this stage.

src/main/resources/com/google/api/codegen/csharp/wrapper.snip
```diff
@@ -19,7 +19,7 @@
 
     // Generated code. DO NOT EDIT!
 
-    @join import : imports
+    @join import : context.removeItem(imports, context.getNamespace(service.getFile))
```
jskeet: Why do we need to remove the import at all?

jskeet:
Do we have any way of testing that generated samples all compile? If so, definitely LGTM. If not, have you tried some random sample of them manually?

src/main/java/com/google/api/codegen/csharp/CSharpContextCommon.java
chrisdunelm: Done

src/main/java/com/google/api/codegen/csharp/CSharpDiscoveryContext.java
chrisdunelm: This is building the correct full name for the gRPC import, so cannot be changed,

src/main/java/com/google/api/codegen/csharp/CSharpSnippetSetRunner.java
chrisdunelm: It's all been fairly significantly changed over the last couple of months. I've essentially just used the Java snippet code and changed it so it does what we need in C#

src/main/resources/com/google/api/codegen/csharp/wrapper.snip
chrisdunelm: It's to remove the "using" statement that imports the namespace within which this code is generated.
We used to do this in a different way, but the underlying framework has been altered so we can't do that any more :(

chrisdunelm:
There is a framework to check all the generated snippets compile OK, but this PR doesn't integrate with it yet. Unfortunately it's a completely separate framework (written in go) that I don't yet have any understanding of :( I'll integrate with in as soon as I have time.
I will manually compile a few to check, then if all is OK commit this.

src/main/java/com/google/api/codegen/csharp/CSharpDiscoveryContext.java
jskeet: gRPC import? I thought this was for the REST API. The gRPC part *won't* include minor version numbers as far as I'm aware.

chrisdunelm:
There were a couple of problems with imports. I've checked some random snippets compile, so if travis passes I'll merge this.

src/main/java/com/google/api/codegen/csharp/CSharpDiscoveryContext.java
chrisdunelm: Oops, yes, my mistake - it's the REST import.
As I understand it, most services do only include the major version.

chrisdunelm: Yes, I've just checked, and most are major version only.
A counter-example is prediction, which is currently v1.6, but the namespace is Google.Apis.Prediction.v1_6. This code is to match that namespace when minor version is included.

src/main/java/com/google/api/codegen/ApiaryConfig.java
```diff
@@ -141,6 +151,22 @@
     return mediaUpload;
   }
 
+  public String getServiceCanonicalName() {
+    return serviceCanonicalName;
+  }
+
+  public void setServiceCanonicalName(String serviceCanonicalName) {
+    this.serviceCanonicalName = serviceCanonicalName;
+  }
```
garrettjonesgoogle: There weren't any set calls before your change here, why did you have to add them?

PR #200, opened 2016-06-08
Update C# codegen using new retry configuration

chrisdunelm:


src/main/resources/com/google/api/codegen/csharp/wrapper.snip
```diff
@@ -190,13 +190,13 @@
             /// </remarks>
             public CallSettings {@method.grpcName}Settings { get; set; } = new CallSettings
             {
-                RetrySettings = new RetrySettings
+                Timing = CallTiming.FromRetry(new RetrySettings
```
jskeet: So are we later going to add code that *doesn't* call FromRetry when there are no retries to make?

jskeet:
LGTM, but with a question.

src/main/resources/com/google/api/codegen/csharp/wrapper.snip
chrisdunelm: Do you mean then the retry configuration in the yaml has no response statuses that allow retry?
I *think* the idea is that even in this case we should still use the retry configuration, as it will have a suitable initial timeout. I'll check what other languages are doing.

PR #207, opened 2016-06-09
Updated config generator

michaelbausor:
Generate additional config fields:
- retry settings
- required_fields
- field name patterns
- generate_samples
- type
- request_method_object

Fix issue https://github.com/googleapis/toolkit/issues/202


src/main/java/com/google/api/codegen/config/FlatteningAndRequiredFieldsConfigGenerator.java
```diff
@@ -25,20 +25,21 @@
 import java.util.Map;
 
 /**
- * Config generator for method parameter flattening.
+ * Config generator for method parameter flattening and required fields.
  */
-public class FlatteningConfigGenerator implements MethodConfigGenerator {
+public class FlatteningAndRequiredFieldsConfigGenerator implements MethodConfigGenerator {
```
garrettjonesgoogle: "FieldConfigGenerator" would be a lot shorter, but I think is sufficiently clear.

src/main/java/com/google/api/codegen/Resources.java
```diff
@@ -69,22 +72,48 @@
...
+    }
+    return specs.values();
+  }
+
+  /**
+   * Returns a map from method name to a list of FieldSegment objects. Each FieldSegment will be
+   * composed of a field name and a path segment, representing an abstract resource name with
+   * wildcards.
+   */
+  public static Map<String, List<FieldSegment>> getAllFieldSegmentsFromHttpPathsPerMethod(
```
garrettjonesgoogle: maybe instead: `getMethodFieldSegmentMapFromHttpPaths`

michaelbausor:
PTAL

src/test/java/com/google/api/codegen/testdata/library_config.baseline
```diff
@@ -1,99 +1,242 @@
...
       groups:
       - parameters:
         - name
         - comments
+    required_fields:
+    - name
+    - comments
+    retry_codes_name: non_idempotent
+    retry_params_name: default
+    request_object_method: false
```
garrettjonesgoogle: This is wrong - request_object_method should be false only when there is 0 or 1 parameters, but true otherwise.

src/main/java/com/google/api/codegen/config/ConfigGeneratorApi.java
```diff
@@ -114,10 +134,47 @@ protected void process() throws Exception {
...
+    Map<String, Object> packageNameMap = new LinkedHashMap<String, Object>();
+    packageNameMap.put(CONFIG_KEY_PACKAGE_NAME, packageName.replace(".", separator));
+    return packageNameMap;
+  }
+
   private void dump(Map<String, Object> data) throws IOException {
     DumperOptions dumperOptions = new DumperOptions();
     dumperOptions.setDefaultFlowStyle(DumperOptions.FlowStyle.BLOCK);
-    dumperOptions.setPrettyFlow(true);
+    dumperOptions.setPrettyFlow(false);
```
shinfan: Does this boolean value affect the output format?

michaelbausor: The only effect is has is on empty lists or dictionaries, so when it is false you get this:
someemptything: []
and when it is true you get this:
someemptything: [
]

src/test/java/com/google/api/codegen/testdata/library_config.baseline
michaelbausor: Ah OK - I did not know, I just set it to false everywhere, will update

src/main/java/com/google/api/codegen/config/FieldNamePatternConfigGenerator.java
```diff
@@ -0,0 +1,53 @@
...
+    if (fieldSegmentMap.containsKey(method.getFullName())) {
+      Map<String, Object> nameMap = new LinkedHashMap<String, Object>();
+      for (FieldSegment fieldSegment : fieldSegmentMap.get(method.getFullName())) {
+        nameMap.put(fieldSegment.getFieldPath(), Resources.getEntityName(fieldSegment));
+      }
+      Map<String, Object> result = new LinkedHashMap<String, Object>();
+      result.put(CONFIG_KEY_FIELD_NAME_PATTERNS, nameMap);
+      return result;
+    } else {
+      return null;
```
shinfan: I notice some generator returns empty map but here it returns null. We probably should be consistent with the return value to avoid potential issues.

src/main/java/com/google/api/codegen/Resources.java
```diff
@@ -69,22 +72,48 @@
...
+          fieldSegmentsPerMethodMap.get(method.getFullName()).add(fieldSegment);
         }
       }
     }
 
-    return specs.values();
+    return fieldSegmentsPerMethodMap;
+  }
+
+  public static String getEntityName(FieldSegment segment) {
```
shinfan: Is this method still being used?

src/main/java/com/google/api/codegen/config/ConfigGeneratorApi.java
shinfan: oh I think it actually looks better when it is false :)

michaelbausor: I do too, that's why I changed it to false!

src/main/java/com/google/api/codegen/Resources.java
```diff
@@ -69,22 +72,48 @@
...
+    return specs.values();
+  }
+
+  /**
+   * Returns a map from method name to a list of FieldSegment objects. Each FieldSegment will be
+   * composed of a field name and a path segment, representing an abstract resource name with
+   * wildcards.
+   */
+  public static Map<String, List<FieldSegment>> getMethodFieldSegmentMapFromHttpPaths(
+      List<Method> methods) {
```
garrettjonesgoogle: I think it would simplify things to reduce the scope of this method to only operate on a single Method input. The getFieldSegmentsFromHttpPaths function could just iterate over methods itself and call this function for each of them; fieldSegmentMap could be removed from FieldNamePatternConfigGenerator, and it could call this method each time it needs the field segments for a method.

src/main/java/com/google/api/codegen/Resources.java
garrettjonesgoogle: @shinfan : it is still being used in CollectionConfigGenerator.generate(...).

src/main/java/com/google/api/codegen/Resources.java
```diff
@@ -69,22 +72,48 @@
...
       }
     }
 
-    return specs.values();
+    return fieldSegmentsPerMethodMap;
+  }
+
+  public static String getEntityName(FieldSegment segment) {
+    // TODO(shinfan): Consider finding a better way to determine the name if possible.
+    List<String> params = Lists.newArrayList(Resources.getParamsForResourceNameWildcards(segment));
```
garrettjonesgoogle: getParamsForResourceNameWildcards is only being used here. I think it would simplify things to just merge the methods - this getEntityName has to contort itself a little bit because of the return type of getParamsForResourceNameWildcards, which is silly.

src/main/java/com/google/api/codegen/config/FieldNamePatternConfigGenerator.java
michaelbausor: Returning null or returning an empty map should produce different results in the output, so sometimes we might want either one. I will update this method and others to be more careful about their return values in the empty cases, e.g. required_fields should be missing when parameter count is zero, not have an empty list.

src/main/java/com/google/api/codegen/Resources.java
```diff
@@ -69,22 +72,48 @@
...
+    }
+    return specs.values();
+  }
+
+  /**
+   * Returns a map from method name to a list of FieldSegment objects. Each FieldSegment will be
+   * composed of a field name and a path segment, representing an abstract resource name with
+   * wildcards.
+   */
+  public static Map<String, List<FieldSegment>> getMethodFieldSegmentMapFromHttpPaths(
```
garrettjonesgoogle: This whole class could be moved under com.google.api.codegen.config; the only reference to it outside of *.codegen.config is the isIdempotent method, but I think that can just be moved into GapicContext, the only place that calls it.

src/main/java/com/google/api/codegen/config/ConfigGeneratorApi.java
```diff
@@ -47,11 +49,30 @@
...
+
   private static final String CONFIG_KEY_SERVICE_NAME = "name";
   private static final String CONFIG_KEY_METHOD_NAME = "name";
-  private static final String CONFIG_KEY_INTERFACES = "interfaces";
   private static final String CONFIG_KEY_METHODS = "methods";
   private static final String CONFIG_KEY_COLLECTIONS = "collections";
+  private static final String CONFIG_KEY_PACKAGE_NAME = "package_name";
+  private static final String CONFIG_KEY_REQUEST_OBJECT_METHOD = "request_object_method";
+
+  private static final String CONFIG_PROTO_TYPE = "com.google.api.codegen.ConfigProto";
```
garrettjonesgoogle: We could actually get this from the ConfigProto class itself.

src/main/java/com/google/api/codegen/config/ConfigGeneratorApi.java
```diff
@@ -47,11 +49,30 @@
...
+
+  private static final Map<String, String> LANGUAGE_PACKAGE_SEPARATORS =
+      ImmutableMap.<String, String>builder()
+          .put("java", ".")
+          .put("python", ".")
+          .put("go", ".")
+          .put("csharp", ".")
+          .put("ruby", "::")
+          .put("php", "\\")
+          .build();
```
garrettjonesgoogle: Nothing for Node?

src/test/java/com/google/api/codegen/testdata/library_config.baseline
```diff
@@ -1,99 +1,242 @@
+type: com.google.api.codegen.ConfigProto
+generate_samples: true
+language_settings:
+  java:
+    package_name: google.example.library.v1
```
garrettjonesgoogle: we should probably add "com." to the beginning of the Java one.

src/test/java/com/google/api/codegen/testdata/library_config.baseline
```diff
@@ -1,99 +1,242 @@
...
+  python:
+    package_name: google.example.library.v1
+  go:
+    package_name: google.example.library.v1
```
garrettjonesgoogle: We should probably add "google.golang.org/" to the beginning of the Go one.

src/test/java/com/google/api/codegen/testdata/library_config.baseline
```diff
@@ -1,99 +1,242 @@
...
+  csharp:
+    package_name: google.example.library.v1
+  ruby:
+    package_name: google::example::library::v1
+  php:
+    package_name: google\example\library\v1
```
garrettjonesgoogle: We should probably capitalize Csharp/Ruby/Php.

src/main/java/com/google/api/codegen/config/ConfigGeneratorApi.java
michaelbausor: Node does not have anything in e.g. pubsub_gapic.yaml. It seems that Node packages are based on file location

michaelbausor:
PTAL

src/main/java/com/google/api/codegen/config/LanguageGenerator.java
```diff
@@ -0,0 +1,92 @@
...
+import com.google.common.collect.ImmutableMap;
+
+import java.util.LinkedHashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Config generator for retry codes and retry params. Static methods are used to generate retry
+ * definitions.
```
garrettjonesgoogle: The doc is referring to the wrong thing for this class.

michaelbausor:
PTAL

garrettjonesgoogle:
LGTM

PR #210, opened 2016-06-09
Re-enable formatting enforcement, with proper excludes

pongad:


bjwatson:
@pongad There was a small bug in this PR. The excludes need to be applied to `tasks.verifyGoogleJavaFormat` to take affect here. I left the block for `tasks.googleJavaFormat` in case we do compile-time reformatting in the future.

I also restored everything that was lost with the rebase in the previous PR (applying Eclipse formatting followed by the Gradle plugin formatting).

LGTM. PTAL.

PR #213, opened 2016-06-09
Update Ruby names

pongad:


src/test/java/com/google/api/codegen/testdata/discoveries/csharp/csharp_dataproc.v1.json.baseline
```diff
@@ -38,108 +38,6 @@ namespace DataprocSample
                 ApplicationName = "Google Cloud Platform Sample",
             });
```
vchudnov-g: What prompted all the deletions?

pongad: I simply updated it from https://www.googleapis.com/discovery/v1/apis/dataproc/v1/rest . As to what prompted the deletion from there, I have no idea.

vchudnov-g: So we're still generating the same baseline files, but the contents just happened to change? OK, I guess. Though weird. It seems like a breaking change to the API surface....

PR #215, opened 2016-06-10
Creating DiscoveryProviderFactory

garrettjonesgoogle:
Net removal of more than 500 lines :-)

* Also changing the result type of GapicProvider - now returning Map<String, Doc> so that the CodeGeneratorApi class can just call ToolUtil.writeFiles itself.


src/main/java/com/google/api/codegen/GeneratorBuilderUtil.java
```diff
@@ -14,13 +14,6 @@
  */
```
geigerj: The class doc comments here are no longer accurate. (They reference a "codegen provider".)

src/main/java/com/google/api/codegen/gapic/GapicProvider.java
```diff
@@ -30,13 +31,13 @@
...
-   * Runs code generation and puts the output in outputPath.
+   * Runs code generation and returns a map relative file paths to generated Doc.
    */
-  void generate(String outputPath) throws Exception;
+  Map<String, Doc> generate();
 
   /**
-   * Generates code for a single snippet. Returns a map from service interface to code for the service.
-   * Returns null if generation failed.
+   * Runs code generation for a single snippet and returns a map relative file
```
geigerj: "map relative" ---> "map from relative"

src/main/java/com/google/api/codegen/GeneratorBuilderUtil.java
garrettjonesgoogle: I'll take care of that, and additionally I think I'll rename the class while I'm at it.

src/main/java/com/google/api/codegen/gapic/GapicProvider.java
```diff
@@ -30,13 +31,13 @@
   List<String> getSnippetFileNames();
 
   /**
-   * Runs code generation and puts the output in outputPath.
+   * Runs code generation and returns a map relative file paths to generated Doc.
```
geigerj: "map relative" ---> "map from relative"

geigerj:
Hooray for net removal of code! Looks good to me other than a few comments.

garrettjonesgoogle:
I have addressed your comments, PTAL.

pongad:
LGTM if test passes.

geigerj:
LGTM

PR #219, opened 2016-06-10
Add Connection getter

### LINKED ISSUE #218: Go: allow several clients to share the same connection
When a client uses multiple services for the same API, it should only need to have one connection. 

One way to support this would be to provide a way to get the connection from a client and pass it to another. E.g.

c1 := vkit.NewClient(...)
c2 := vkit.NewConfigClient(gax.WithConnection(c1.Connection()))

saicheems:
Resolves #218 with googleapis/gax-go#25


jba:
LGTM.

PR #220, opened 2016-06-11
Update Go paging iterators

jba:
The slight differences between Next and NextPage were hard to get right
using the existing division of labor. Refactored so that apiCall does
not deal with when to return Done, it just calls the List method and
sets some iterator fields. Deciding when to return Done and what to
return along with it is left to Next and NextPage.

This change was tested against the iterator tests of the logging client.

### DIFF
--- a/src/main/resources/com/google/api/codegen/go/main.snip
+++ b/src/main/resources/com/google/api/codegen/go/main.snip
@@ -205,9 +205,6 @@
                         ctx = metadata.NewContext(ctx, c.metadata)
                         it := &{@iteratorTypeName}{}
                         it.apiCall = func() error {
-                            if it.atLastPage {
-                                return Done
-                            }
                             var resp {@outTypeName}
                             err := gax.Invoke(ctx, func (ctx context.Context) error {
                                 var err error
@@ -223,9 +220,8 @@
                             }
                             if resp.{@responseTokenFieldName} == {@tokenZeroValue} {
                                 it.atLastPage = true
-                            } else {
-                                it.nextPageToken = resp.{@responseTokenFieldName}
                             }
+                            it.nextPageToken = resp.{@responseTokenFieldName}
                             it.items = resp.{@resourceFieldName}
                             return nil
                         }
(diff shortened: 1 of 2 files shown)


jba:
@saicheems

src/main/resources/com/google/api/codegen/go/main.snip
```diff
@@ -301,19 +297,28 @@
...
             }
 
             // Next returns the next element in the stream. It returns Done at
             // the end of the stream.
             func (it *{@iteratorTypeName}) Next() ({@resourceFieldTypeName}, error) {
                 for it.currentIndex >= len(it.items) {
-                    _, err := it.NextPage()
-                    if err != nil {
+		    if it.atLastPage {
+		       return nil, Done
```
saicheems: Nit: 4 spaces

saicheems:
Ah, I couldn't figure out for a while why the spacing looked weird in the diff.
Due to some failures in the template language, mixing spaces and tabs doesn't work well - could you run a text replace on tab => 4 spaces?

To get the tests working, the generator outputs the actual baseline to `/tmp/com.google...`
Copy that into the `test/com.google.../testdata` directory.

saicheems:
Thanks for making this change! It is tricky to get some functional subtleties across via docs and email

jba:
I fixed the spacing, but I can't find the file go_main_library.baseline anywhere.

saicheems:
Hmm... It should be here `/tmp/com.google.api.codegen_testdata/go_main_library.baseline`

If it's not, I ran the build for your branch on my machine:
https://gist.githubusercontent.com/saicheems/83509f4ca04a2c76a3093e6b2ba3e11a/raw/0ac2596205d59ba270ae07747fffb1801e132eb7/gistfile1.txt

jba:
The only file I see in that gist is google.golang.org/cloud/library/apiv1/library_client.go. Also it's v1, not v2.

saicheems:
Right, it's the full baseline: `go_main_library.baseline`
If you copy that into `src/test/java/com/google/api/codegen/testdata` the travis build will pass

jba:
Looks like it worked. Thanks!

saicheems:
lgtm

PR #223, opened 2016-06-13
Fix path to MainDiscoveryProviderFactory in the *_discovery.yaml files.

vchudnov-g:


pongad:
LGTM if tests pass

vchudnov-g:
gradlew check passes
Though it didn't catch this error before. We need to follow up with a test.

PR #225, opened 2016-06-13
Updated PHP generated surface

michaelbausor:
- Altered surface to match php-cs-fixer
- Added sample generation

### DIFF
--- a/src/main/resources/com/google/api/codegen/php/method_sample.snip
+++ b/src/main/resources/com/google/api/codegen/php/method_sample.snip
@@ -1,0 +1,119 @@
+@snippet methodSampleDoc(docConfig)
+    @let ApiName = docConfig.getApiName, \
+            apiName = context.upperCamelToLowerCamel(ApiName), \
+            methodName = docConfig.getMethodName(), \
+            returnType = docConfig.getReturnType(), \
+            initCodeSpec = docConfig.getInitCode()
+        try {
+            ${@apiName} = new {@ApiName}();
+            {@initCode(initCodeSpec)}
+            @if docConfig.isPagedVariant()
+                for ({@methodCallSampleCode(methodName, apiName, initCodeSpec)} as element) {
+                    // doThingsWith(element);
+                }
+            @else
+                @if returnType
+                    $response = {@methodCallSampleCode(methodName, apiName, initCodeSpec)};
+                @else
+                    {@methodCallSampleCode(methodName, apiName, initCodeSpec)};
+                @end
+            @end
+        } finally {
+            if (isset(${@apiName})) {
+                ${@apiName}->close();
+            }
+        }
+    @end
+@end
+
+@private methodCallSampleCode(methodName, apiName, initCodeSpec)
+    ${@apiName}->{@methodName}({@argList(initCodeSpec.getArgFields)})
+@end
+
+@private initCode(initCodeSpec)
+  @join line : initCodeSpec.getLines()
+    @switch line.getLineType.toString()
+    @case "StructureInitLine"
+      {@initLineStructure(line)}
+    @case "ListInitLine"
+      {@initLineList(line)}
+    @case "SimpleInitLine"
+      {@initLineSimple(line)}
+    @default
+      {@unhandledCase()}
+    @end
+  @end
+@end
+
+@private initLineStructure(line)
+  {@formattedIdentifier(line)} = new {@context.typeName(line.getType)}();
+  @join fieldSetting : line.getFieldSettings
+    @let setCallName = setCallName(fieldSetting)
+      {@formattedIdentifier(line)}->{@setCallName}({@formattedIdentifier(fieldSetting)});
+    @end
+  @end
+@end
+
+@private initLineList(line)
+  {@formattedIdentifier(line)} = [{@initList(line)}];
+@end
+
+@private initList(line)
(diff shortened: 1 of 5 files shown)


garrettjonesgoogle:
LGTM - have Shin look at it too before submitting.

src/test/java/com/google/api/codegen/testdata/php_main_library.baseline
```diff
@@ -616,23 +757,38 @@ class LibraryServiceApi
     /**
      * Lists books in a shelf.
      *
+     * Sample code:
+     * <pre><code>
+     * try {
+     *     $libraryServiceApi = new LibraryServiceApi();
+     *     $formattedName = LibraryServiceApi::formatShelfName("[SHELF]");
+     *     for ($libraryServiceApi->listBooks($formattedName) as element) {
```
shinfan: Should it be foreach?
Also s/element/$element/

src/main/resources/com/google/api/codegen/php/method_sample.snip
```diff
@@ -0,0 +1,119 @@
...
+    @let ApiName = docConfig.getApiName, \
+            apiName = context.upperCamelToLowerCamel(ApiName), \
+            methodName = docConfig.getMethodName(), \
+            returnType = docConfig.getReturnType(), \
+            initCodeSpec = docConfig.getInitCode()
+        try {
+            ${@apiName} = new {@ApiName}();
+            {@initCode(initCodeSpec)}
+            @if docConfig.isPagedVariant()
+                for ({@methodCallSampleCode(methodName, apiName, initCodeSpec)} as element) {
```
shinfan: We might want to differentiate simple array vs associative arrays here. If the array is associative, we should do foreach (xxx as $key => $value)

src/test/java/com/google/api/codegen/testdata/php_main_library.baseline
michaelbausor: Done. Need compile check!

shinfan:
Overall looks good. Just some comments about the iterable calls.

src/main/resources/com/google/api/codegen/php/method_sample.snip
michaelbausor: Is this possible? Because PHP does not have separate types, the "standard" way to tell if an array is associative is a runtime inspection... http://stackoverflow.com/questions/173400/how-to-check-if-php-array-is-associative-or-sequential
Maybe we should do the associative case by default?

shinfan: In apiary snippets we can tell if the resource should be a sequential or associative array by looking at the discovery doc. Correct me if I am wrong, I think in this case it is good enough to do the associative case by default since the list method returns a PageAccessor object.

michaelbausor: Does iterating the PageAccessor ever produce an associative array? I thought that it would always return a sequential array, but if it does then definitely going associative makes sense

shinfan: Currently PageAccessor produces an associative array where the key is the request: https://github.com/googleapis/gax-php/blob/master/tests/ApiCallableTest.php#L218

michaelbausor: And if so, how do other languages e.g. Java handle this? Because the java page accessor is an iterable, so can't really handle the map case unless it iterates over key-value pairs...

shinfan: Yea, the Java one is sequential. I think that's the difference between two languages, in PHP the key() method is required by the Iterator interface: http://php.net/manual/en/class.iterator.php

shinfan: My understanding is that in PHP associative array is a super set of sequential array which means the sequential array is a special case where the key is null.

michaelbausor: But will the key for a PageAccessor always be sequential? Or is there a case when the key could be something other than the numeric index?
If it can be a key other than sequential numeric, then we definitely want to use the key=>value style. But otherwise, I think it is better to use just the value style, because then it is clear to the user that the keys are sequential.

michaelbausor: The reason I mentioned Java is because if there is a case when the key might be something other than sequential, e.g. page streaming of a map object, I don't know how we are handling that in Java

garrettjonesgoogle: @shinfan actually I think that for a sequential array, the key is the index.

michaelbausor: That was my understanding as well. It think that PageAccessor will always have a sequential array, although I am still not fully sure. Is it possible to apply page streaming to a grpc map field? Or only to a repeated field?

shinfan: @garrettjonesgoogle Yea sorry you are right, the key will be the index for sequential array.

For now the key for PageAccessor will be the request object so it makes sense to have a key/value iteration in the sample code.

shinfan:
LGTM.

Note: We need to consult the PHP exports about the traversing problems of PageAccessor.

PR #226, opened 2016-06-13
Go Discovery: silence declare-and-not-use errors

pongad:


vchudnov-g:
LGTM. Please run a compile check before submitting.

pongad:
Ran and all pass

PR #227, opened 2016-06-13
Java, PHP: Moving code path prefix from artman to toolkit

garrettjonesgoogle:
Java's prefix is "src/main/java" and PHP's is "src". 

Also:

* Moving PrefixPackageCodePathMapper under gapic/ and renaming to CommonGapicCodePathMapper; it is Gapic-specific. 
* Switching the implementation to use a builder pattern - it is clearer for cases like this where the inputs are all optional
* Dealing with the package separator in an automatic way instead of making it explicit

Coordinates with https://github.com/googleapis/artman/pull/34 .


geigerj:
LGTM

PR #229, opened 2016-06-13
Update collection configs for unbound paths

michaelbausor:
Added naming of unbound segments in resource paths with _path suffix
Added name conflict resolution for collection configs
Added test method to library.proto, updated tests

### DIFF
--- a/src/main/java/com/google/api/codegen/config/LiteralWildcardSegment.java
+++ b/src/main/java/com/google/api/codegen/config/LiteralWildcardSegment.java
@@ -1,0 +1,60 @@
+/* Copyright 2016 Google Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen.config;
+
+import com.google.api.codegen.Inflector;
+import com.google.api.tools.framework.aspects.http.model.HttpAttribute.LiteralSegment;
+import com.google.api.tools.framework.aspects.http.model.HttpAttribute.WildcardSegment;
+import com.google.auto.value.AutoValue;
+
+/**
+ * Class to hold a LiteralSegment and WildcardSegment pair
+ */
+@AutoValue
+public abstract class LiteralWildcardSegment {
+
+  public static LiteralWildcardSegment create(
+      LiteralSegment literalSegment, WildcardSegment wildcardSegment) {
+    return new AutoValue_LiteralWildcardSegment(literalSegment, wildcardSegment);
+  }
+
+  public abstract LiteralSegment literalSegment();
+
+  public abstract WildcardSegment wildcardSegment();
+
+  public String collectionName() {
+    return literalSegment().getLiteral();
+  }
+
+  public String paramName() {
+    String paramName = Inflector.singularize(collectionName());
+    if (wildcardSegment().isUnbounded()) {
+      paramName += "_path";
+    }
+    return paramName;
+  }
+
+  public String resourcePathString() {
+    StringBuilder builder = new StringBuilder();
+    builder.append(collectionName());
+    builder.append("/{");
+    builder.append(paramName());
+    if (wildcardSegment().isUnbounded()) {
+      builder.append("=**");
+    }
+    builder.append("}");
+    return builder.toString();
+  }
+}
--- a/src/main/java/com/google/api/codegen/config/Resources.java
+++ b/src/main/java/com/google/api/codegen/config/Resources.java
@@ -81,7 +81,6 @@ public class Resources {
 
   public static String getEntityName(FieldSegment segment) {
     Preconditions.checkArgument(isTemplateFieldSegment(segment));
-    // TODO(shinfan): Consider finding a better way to determine the name if possible.
     List<LiteralWildcardSegment> literalWildcardSegments =
         getWildcardCollectionMapFieldSegment(segment);
     return literalWildcardSegments.get(literalWildcardSegments.size() - 1).paramName();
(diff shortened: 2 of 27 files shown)


src/test/java/com/google/api/codegen/testdata/library_config.baseline
```diff
@@ -16,10 +16,12 @@ language_settings:
 interfaces:
 - name: google.example.library.v1.LibraryService
   collections:
+  - name_pattern: archives/{archif_path=**}/books/{book}
```
garrettjonesgoogle: Update inflector (singular wrong - archif)

michaelbausor: Could take this dependency: https://github.com/atteo/evo-inflector
Otherwise, could build on existing implementation to cover this case

michaelbausor: Hmmm, that link doesn't singularize...
Maybe this one: http://www.java2s.com/Tutorial/Java/0040__Data-Type/Transformswordstosingularpluralhumanizedhumanreadableunderscorecamelcaseorordinalform.htm

michaelbausor: Updated inflector rather than take the dependency

michaelbausor:
PTAL

src/test/java/com/google/api/codegen/testdata/library_config.baseline
```diff
@@ -16,10 +16,12 @@ language_settings:
 interfaces:
 - name: google.example.library.v1.LibraryService
   collections:
+  - name_pattern: archives/{archive_path=**}/books/{book}
+    entity_name: book
```
garrettjonesgoogle: Why call the special archived one 'book' and the normal shelved one 'book_2'? I would expect the reverse.

src/main/java/com/google/api/codegen/config/Resources.java
```diff
@@ -80,9 +82,27 @@
   public static String getEntityName(FieldSegment segment) {
     Preconditions.checkArgument(isTemplateFieldSegment(segment));
     // TODO(shinfan): Consider finding a better way to determine the name if possible.
```
garrettjonesgoogle: Do we want to keep this TODO?

src/test/java/com/google/api/codegen/testdata/library_config.baseline
michaelbausor: It is just the order they get processed in. I could change it, but I don't know how to guarantee that I'm not causing the same problem in some other proto layout. Could possibly collect all conflicting names and order them by simplicity somehow, but this feels like overkill. This can be changed manually by users in the generated config file, which they would probably want to do anyway (e.g. as in library_gapic.yaml)

garrettjonesgoogle: Alrighty.

src/main/java/com/google/api/codegen/config/Resources.java
michaelbausor: Removed. I think it is hard to do much better than using the last item plus conflict resolution, and the gain is marginal considering the user can just rename.

src/main/java/com/google/api/codegen/config/LiteralWildcardSegment.java
```diff
@@ -0,0 +1,60 @@
...
+import com.google.api.codegen.Inflector;
+import com.google.api.tools.framework.aspects.http.model.HttpAttribute.LiteralSegment;
+import com.google.api.tools.framework.aspects.http.model.HttpAttribute.WildcardSegment;
+import com.google.auto.value.AutoValue;
+
+/**
+ * Class to hold a LiteralSegment and WildcardSegment pair
+ */
+@AutoValue
+public abstract class LiteralWildcardSegment {
```
garrettjonesgoogle: Maybe call 'NamedWildcard' instead. It is composed of two segments, so using Segment in the name is either misleading or awkward (e.g. LiteralAndWildcardSegments).

src/main/java/com/google/api/codegen/config/Resources.java
```diff
@@ -126,20 +146,22 @@ private static boolean isTemplateFieldSegment(PathSegment pathSegment) {
     return true;
   }
 
-  private static List<String> getWildcardCollectionNames(FieldSegment fieldSegment) {
-    List<String> collectionNames = new ArrayList<>();
+  private static List<LiteralWildcardSegment> getWildcardCollectionMapFieldSegment(
```
garrettjonesgoogle: maybe just call this `getNamedWildcards`

src/main/java/com/google/api/codegen/config/Resources.java
```diff
@@ -80,9 +82,27 @@
...
+        getWildcardCollectionMapFieldSegment(segment);
+    return literalWildcardSegments.get(literalWildcardSegments.size() - 1).paramName();
+  }
+
+  public static Map<String, String> getResourceToEntityNameMap(Iterable<FieldSegment> segments) {
+    Set<String> usedNameSet = new HashSet<>();
+    Map<String, String> nameMap = new LinkedHashMap<>();
+    for (FieldSegment segment : segments) {
+      String resourceNameString = Resources.templatize(segment);
+      String entityNameStringBase = Resources.getEntityName(segment);
```
garrettjonesgoogle: The `Resources.` part can be removed since these functions are in the same class.

src/main/java/com/google/api/codegen/config/Resources.java
```diff
@@ -80,9 +82,27 @@
...
+      String entityNameString = entityNameStringBase;
+      int index = 1;
```
garrettjonesgoogle: Maybe follow the same pattern as https://github.com/googleapis/toolkit/blob/master/src/main/java/com/google/api/codegen/metacode/InitCodeGenerator.java#L84 (separate function), or maybe even share code.

garrettjonesgoogle: Another alternative is instead of appending numbers when there is a clash, instead start adding earlier literals. In the library case, this would produce something like shelf_book and archive_book.

garrettjonesgoogle: After offline discussion, I think it's best to be simpler and stick with the numeric resolution for now.

michaelbausor:
PTAL

garrettjonesgoogle:
LGTM

Issue #230, opened 2016-06-14
C# should use FormatResourceName rather than GetResourceName


jskeet:
It's a nicer name, and doesn't sound like an RPC.

PR #232, opened 2016-06-14
C# rename GetResourName() -> FormatResourceName()

chrisdunelm:
Fixes issue #230


jskeet:
LGTM. We should think carefully about when we want to rerun codegen.

chrisdunelm:
Committing the new codegen won't automatically affect the nuget packages.
But I won't submit the codegen until we've discussed.

PR #234, opened 2016-06-14
Merge branch 'master' of github.com:googleapis/toolkit

tcoffee-google:
Merge to sync with working copy, due to unexpected nonidentification of earlier commits due to squashing.


tcoffee-google:
@pongad please sanity check --- thanks!

broady:
More descriptive PR would be great here.

@jonparrott (Python DPE) FYI.

garrettjonesgoogle:
This PR needs to incorporate the new structure of providers as set up in https://github.com/googleapis/toolkit/commit/3b60c778b59635c97d9a8db2582cf6f84c9ee7fa .

tcoffee-google:
@pongad @garrettjonesgoogle PTAL

Looks to me like some logic with Java strings was removed, this may need to be resolved (?).

src/main/java/com/google/api/codegen/DiscoveryContext.java
```diff
@@ -68,6 +68,15 @@ public Api getApi() {
...
 
+  public String getApiRevision() {
+    return service.getDocumentation().getOverview();
+  }
+
+  public String outputRoot() {
+    Api api = getApi();
+    return "autogenerated/" + api.getName() + "/" + api.getVersion() + "/" + getApiRevision() + "/";
+  }
+
```
garrettjonesgoogle: Revert

src/main/java/com/google/api/codegen/discovery/CommonDiscoveryProvider.java
```diff
@@ -43,15 +43,7 @@ public CommonDiscoveryProvider(
...
-    Api api = context.getApi();
-    String outputRoot =
-        "autogenerated/"
-            + api.getName()
-            + "/"
-            + api.getVersion()
-            + "/"
-            + context.getService().getDocumentation().getOverview();
-    String resultPath = outputRoot + "/" + result.getFilename();
+    String resultPath = context.outputRoot() + result.getFilename();
```
garrettjonesgoogle: Revert

tcoffee-google:
@pongad @garrettjonesgoogle PTAL

I refactored the string format logic (that was left only in Python on master) to cover all languages in an extensible way.

src/main/java/com/google/api/codegen/DiscoveryContext.java
```diff
@@ -156,11 +161,56 @@ public boolean isPatch(Method method) {
...
+    return line + "  // " + comment;
+  }
+
+  /*
+   * Returns default string (to end of line) for a type's string field. If ApiaryConfig specifies a
+   * string format, returns corresponding default string with inline comment. Otherwise, if
+   * ApiaryConfig specifies a pattern, returns corresponding default string. Otherwise, returns
+   * empty string.
+   */
   protected String getDefaultString(Type type, Field field) {
```
pongad: Now that we are doing this, let's put the formatting in individual languages' templates.

The way I'm imagining this, `getDefaultString` (or something similarly named) can return both the string and the comment. The template can figure out what to do with them.

tcoffee-google: It's sufficiently complicated that to do this, the templates would almost certainly end up calling back into Java, adding yet another cycle of back-and-forth. The proposed functionality appears sufficiently general to cover all the cases I've seen or imagined. Let's not generalize further unless there are cases we can't already cover cleanly.

pongad:
lgtm then

garrettjonesgoogle:
As @broady requested, can you make the commit message more informative?

src/main/java/com/google/api/codegen/discovery/CommonDiscoveryProvider.java
```diff
@@ -45,12 +45,8 @@ public CommonDiscoveryProvider(
 
     Api api = context.getApi();
     String outputRoot =
-        "autogenerated/"
-            + api.getName()
-            + "/"
-            + api.getVersion()
-            + "/"
-            + context.getService().getDocumentation().getOverview();
+        "autogenerated/" + api.getName() + "/" + api.getVersion() + "/" + context.getApiRevision();
```
garrettjonesgoogle: I suspect that you haven't run `./gradlew googleJavaFormat`, which uses the format of the existing code here.

src/main/java/com/google/api/codegen/DiscoveryContext.java
```diff
@@ -68,6 +69,10 @@ public Api getApi() {
     return this.getService().getApis(0);
   }
 
+  public String getApiRevision() {
+    return service.getDocumentation().getOverview();
+  }
+
```
garrettjonesgoogle: Delete this, it's not used

src/main/java/com/google/api/codegen/discovery/CommonDiscoveryProvider.java
pongad: I think it has already been run; test wouldn't pass otherwise. The line got shorter, GJF might pull it back to the same line.

garrettjonesgoogle: Ahh that makes sense.

tcoffee-google:
The commit message was changed to cover the substantive changes, it just doesn't show up in the PR title.

src/main/java/com/google/api/codegen/DiscoveryContext.java
tcoffee-google: It now is again (in CommonDiscoveryProvider): it was a somewhat arbitrary choice to use the Service documentation overview field to hold the API revision (for lack of a better option), so this reference seemed unclear otherwise.

tcoffee-google:
@chrisdunelm it looks like C# alone currently does not use the getDefaultString method in DiscoveryContext to generate default values for string variables ... any reason it shouldn't be changed to use logic similar to the other languages?

garrettjonesgoogle:
LGTM

vchudnov-g:
LGTM

PR #237, opened 2016-06-14
Removing the hiding of page_token as a parameter

garrettjonesgoogle:
Also:
* PHP: switching from html tags to three ticks for code blocks
* PHP: Switching repeated types to the type plus brackets

@jmuk @geigerj Please check Python/Ruby/Node and make sure the changes are acceptable.

### DIFF
--- a/src/main/java/com/google/api/codegen/MethodConfig.java
+++ b/src/main/java/com/google/api/codegen/MethodConfig.java
@@ -142,7 +142,7 @@ public class MethodConfig {
 
     Iterable<Field> optionalFields =
         Iterables.filter(
-            new ServiceMessages().flattenedFields(method.getInputType()),
+            method.getInputType().getMessageType().getFields(),
             new Predicate<Field>() {
               @Override
               public boolean apply(Field input) {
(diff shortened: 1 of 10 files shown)


geigerj:
Currently, in Python/Node/Ruby, we still want to hide the page_token. In those languages, page_token is field in the CallOptions. The presence/absence of a page token in the CallOptions is the switch to toggle page-streaming between per-resource and per-page, so it's not treated like just another optional parameter. ([Example](https://github.com/googleapis/toolkit/blob/06514fa0171e1751d59677dabaa853e00ec5c1b8/src/test/java/com/google/api/codegen/testdata/python_main_library.baseline#L405))

garrettjonesgoogle:
Oh ok. We should probably make PHP match that pattern too.

Although, I find it odd that a field from the RPC is treated differently like this; it doesn't quite match our policy of generally reflecting the API surface in a transparent way.

geigerj:
Agreed about transparency. I'm open to discussing an alternative approach for all four languages if you think this is a significant disadvantage.

garrettjonesgoogle:
I'm going to close this PR; @michaelbausor will come up with a cross-language proposal for page streaming enhancements relative to where we are now.

PR #238, opened 2016-06-15
PHP doc updates

garrettjonesgoogle:


src/main/java/com/google/api/codegen/php/PhpGapicContext.java
```diff
@@ -132,8 +132,10 @@ public String addImport(String typeName) {
    * Returns the PHP representation of a reference to a type.
    */
   public String typeName(TypeRef type) {
-    if (type.isMap() || type.isRepeated()) {
+    if (type.isMap()) {
       return "array";
+    } else if (type.isRepeated()) {
+      return basicTypeName(type) + "[]";
```
shinfan: This makes sense but I am concerned that this is not consistent with the current php client library

garrettjonesgoogle: Which current php client library?

michaelbausor:
LGTM, we should check if there is also an equivalent doc syntax that could be used for associative arrays, e.g. the optionalArgs parameter, although this will be tricky because it is mixed type.

garrettjonesgoogle:
I have an open question to gcloud-php what the syntax for associative arrays should be. I think there are two cases:

1. Configuration - mixed values. I think this is documented value-by-value and thus is just tagged as `array`.
2. An actual map of consistent key and value types. This is the one I have no idea about.

I wanted to at least update the known one (indexed array).

src/main/java/com/google/api/codegen/php/PhpGapicContext.java
shinfan: I was referring to: https://github.com/google/google-api-php-client-services/blob/master/Translate/Resource/Detections.php

But as long as this is recommended by the gcloud-php, I am okay.

shinfan:
LGTM

PR #244, opened 2016-06-15
Update nodejs codegen.

jmuk:
- supports the new ApiCallable structure
- adds 'arguments' at the end of arguejs -- that's necessary
  in the strict mode.
- adds default value for otherArgs, otherwise it fails when missing values.
- modifies the module structure bit -- this will depend on the
  grpc client package. The same package structure as python/ruby has.

### DIFF
--- a/src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
+++ b/src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
@@ -68,6 +68,13 @@ public class NodeJSGapicContext extends GapicContext implements NodeJSContext {
   }
 
   /**
+   * The package name of the grpc module for the API.
+   */
+  public String grpcClientName(Interface service) {
+    return "grpc-" + service.getParent().getFullName().replace('.', '-');
+  }
+
+  /**
    * Returns type information for a field in JSDoc style.
    */
   private String fieldTypeCardinalityComment(Field field) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
```diff
@@ -68,6 +68,13 @@ public String filePath(ProtoFile file) {
   }
 
   /**
+   * The package name of the grpc module for the API.
+   */
+  public String grpcClientName(Interface service) {
+    return "grpc-" + service.getParent().getFullName().replace('.', '-');
```
geigerj: optional: I prefer the syntax `service.getFile()` rather than `service.getParent()` (same for other proto elements) wherever possible because it seems more readable/descriptive.

src/main/resources/com/google/api/codegen/nodejs/main.snip
```diff
@@ -191,21 +191,34 @@
...
-      this.headers.set('x-goog-api-client', googleApiClient);
+      var metadata = new grpc.Metadata();
+      metadata.set('x-goog-api-client', googleApiClient);
+
+      {@constructDefaults(service)}
+
       this.stub = gax.createStub(
           servicePath,
           port,
-          google.example.library.v1.LibraryService,
```
geigerj: 👍

geigerj:
LGTM

src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
jmuk: done

PR #247, opened 2016-06-15
Tweak the comment of composer auto-loading

shinfan:


src/main/resources/com/google/api/codegen/php/discovery_fragment.snip
```diff
@@ -18,7 +18,7 @@
      *    https://cloud.google.com/sdk/ and run 'gcloud beta auth application-default login'
      */
 
-    // composer autoloading
+    // To install the client library with Composer
```
vchudnov-g: I think the suggestion was referring to tweaking pre-requisite #2, not this code code comment. 

This code comment is fine, though I would just change it to start with a verb for consistency with the other comments:

`// Autoload Composer.`

And while we're being nitpicky, the `// Create a new XXX service` comments should end with a period, too.

shinfan:
PTAL

vchudnov-g:
Thanks for addressing this!

LGTM

PR #249, opened 2016-06-16
Make default string more informative

pongad:
Previously, most strings are blank.
The only non-blank ones were in the slash-separated-resources format:
  projects/{MY-PROJECT}/topics/{MY-TOPICS}

This commit gives more informative default values.
* Some API-field-pattern tuples are special-cased.
  Eg, default string for zones is "us-central1-f"
* If not special-cased, slash-separated format still works
* Otherwise, we generate dummy strings
  Eg, default string for project name is "{MY-PROJECT-NAME}"

### DIFF
--- a/src/main/java/com/google/api/codegen/discovery/DefaultString.java
+++ b/src/main/java/com/google/api/codegen/discovery/DefaultString.java
@@ -103,13 +103,10 @@ public class DefaultString {
    */
   @VisibleForTesting
   static String forPattern(String pattern) {
-    // We only care about patterns that has alternating literal and wildcard like
+    // We only care about patterns that have alternating literal and wildcard like
     //  ^foo/[^/]*/bar/[^/]*$
     // Ignore if what we get looks nothing like this.
-    if (pattern == null
-        || !pattern.startsWith("^")
-        || !pattern.endsWith("$")
-        || substrCount(pattern, "/") != substrCount(pattern, WILDCARD_PATTERN) * 3 - 1) {
+    if (pattern == null || !pattern.startsWith("^") || !pattern.endsWith("$")) {
       return null;
     }
     pattern = pattern.substring(1, pattern.length() - 1);
@@ -128,22 +125,6 @@ public class DefaultString {
           .append('}');
     }
     return ret.substring(1);
-  }
-
-  /**
-   * Counts the number of non-overlapping instances of `needle` in `haystack`.
-   */
-  private static int substrCount(String haystack, String needle) {
-    int count = 0;
-    int fromIndex = 0;
-    for (; ; ) {
-      int index = haystack.indexOf(needle, fromIndex);
-      if (index < 0) {
-        return count;
-      }
-      fromIndex = index + needle.length();
-      count++;
-    }
   }
 
   /**
--- a/src/main/java/com/google/api/codegen/discovery/DefaultString.java
+++ b/src/main/java/com/google/api/codegen/discovery/DefaultString.java
@@ -15,8 +15,12 @@
 package com.google.api.codegen.discovery;
 
 import com.google.api.codegen.Inflector;
+import com.google.api.codegen.LanguageUtil;
 import com.google.auto.value.AutoValue;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.protobuf.Field;
+import com.google.protobuf.Type;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -28,12 +32,58 @@ import javax.annotation.Nullable;
  */
 public class DefaultString {
 
+  @AutoValue
+  abstract static class SampleKey {
+    abstract String getApiName();
+
+    abstract String getFieldName();
+
+    abstract String getRegexp();
+
+    static SampleKey create(String apiName, String fieldName, String regexp) {
+      return new AutoValue_DefaultString_SampleKey(apiName, fieldName, regexp);
+    }
+  }
+
+  private static final ImmutableMap<SampleKey, String> sampleStrings =
+      ImmutableMap.<SampleKey, String>builder()
+          .put(
+              SampleKey.create("compute", "zone", "[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?"),
+              "us-central1-f")
+          .put(
+              SampleKey.create("autoscaler", "zone", "[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?"),
+              "us-central1-f")
+          .put(
+              SampleKey.create("clouduseraccounts", "zone", "[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?"),
+              "us-central1-f")
+          .build();
+
+  public static String of(String apiName, String fieldName, String pattern) {
+    if (pattern != null) {
+      String ret = sampleStrings.get(SampleKey.create(apiName, fieldName, pattern));
+      if (ret != null) {
+        return ret;
+      }
+
+      ret = forPattern(pattern);
+      if (ret != null) {
+        return ret;
+      }
+    }
+    return String.format(
+        "{MY-%s}", LanguageUtil.lowerCamelToUpperUnderscore(fieldName).replace('_', '-'));
+  }
+
+  static String getSample(String apiName, String fieldName, String pattern) {
+    return sampleStrings.get(SampleKey.create(apiName, fieldName, pattern));
+  }
+
   private static final String WILDCARD_PATTERN = "[^/]*";
 
   /**
    * Returns a default string from `pattern`, or null if the pattern is not supported.
    */
-  public static String forPattern(String pattern) {
+  static String forPattern(String pattern) {
     // We only care about patterns that has alternating literal and wildcard like
     //  ^foo/[^/]*/bar/[^/]*$
     // Ignore if what we get looks nothing like this.
--- a/src/main/java/com/google/api/codegen/discovery/DefaultString.java
+++ b/src/main/java/com/google/api/codegen/discovery/DefaultString.java
@@ -17,6 +17,7 @@ package com.google.api.codegen.discovery;
 import com.google.api.codegen.Inflector;
 import com.google.api.codegen.LanguageUtil;
 import com.google.auto.value.AutoValue;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.protobuf.Field;
@@ -45,7 +46,23 @@ public class DefaultString {
     }
   }
 
-  private static final ImmutableMap<SampleKey, String> sampleStrings =
+  private final String declare;
+  private final String comment;
+
+  private DefaultString(String declare, String comment) {
+    this.declare = declare;
+    this.comment = comment;
+  }
+
+  public String getDeclare() {
+    return declare;
+  }
+
+  public String getComment() {
+    return comment;
+  }
+
+  private static final ImmutableMap<SampleKey, String> SAMPLE_STRINGS =
       ImmutableMap.<SampleKey, String>builder()
           .put(
               SampleKey.create("compute", "zone", "[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?"),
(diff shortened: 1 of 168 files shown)


src/test/java/com/google/api/codegen/testdata/discoveries/go/go_autoscaler.v1beta2.json.baseline
```diff
@@ -224,13 +224,13 @@ func main() {
   }
 
   // Project ID of Autoscaler resource.
-  project := "" // TODO: Update placeholder value.
+  project := "{MY-PROJECT}" // TODO: Update placeholder value.
 
   // Zone name of Autoscaler resource.
-  zone := "" // TODO: Update placeholder value.
+  zone := "{MY-ZONE}" // TODO: Update placeholder value.
```
vchudnov-g: This zone isn't one we can special-case, too? (have not looked at API def)

src/main/java/com/google/api/codegen/discovery/DefaultString.java
```diff
@@ -28,12 +32,58 @@
...
+              "us-central1-f")
+          .put(
+              SampleKey.create("autoscaler", "zone", "[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?"),
+              "us-central1-f")
+          .put(
+              SampleKey.create("clouduseraccounts", "zone", "[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?"),
+              "us-central1-f")
+          .build();
+
+  public static String of(String apiName, String fieldName, String pattern) {
```
vchudnov-g: I'd prefer s/of/for/ but for is a reserved keyword. :(

vchudnov-g:
Check my question about a possible additional special case.

LGTM

pongad:
@vchudnov-g
The zone in question does not specify a pattern regexp, so we don't make the substitution. This is probably a good thing in case "zone" might mean different things in different contexts.

Looking at this for a while, I'm starting to think samples like this looks a little funny:
```
String project = "{MY-PROJECT}";
String zone = "us-central1-f";
String operation = "{MY-OPERATION}";
```
The place holders like `{MY-PROJECT}` looks good but `us-central1-f` looks out of place. They are sparse and (at least to me) breaks the flow of the sample.

I'd be curious to know what you and tech writers think.

vchudnov-g:
I see what you mean. Some disparate thoughts come to mind:
1. I wonder whether something like `zone = "us-central1-f"` would not sufficiently prompt developers to change it to a more appropriate value, and so we'll find way too many API calls using the default calls.
2. I wonder whether saying `zone = "{us-central1-f}"` would be better: brackets always mean you need to change the placeholder
3. I wonder whether we could go the other way with strings like `project = "TODO-my-project"`

broady:
For zone, maybe moving it into a comment is better. The idea is to prompt
them as to the format of the value, right?

pongad:
I should have thought more fully about the design before writing all this code. On the other hand, I guess it was easy enough to prototype that we didn't loose significant amount of time :/

I like @broady 's suggestion. Maybe something like these three cases? (They are written in Python syntax, but it should trivially generalize to other languages.)

- Normal format
`operation = '{MY-OPERATION}'`

- Normal format with sample
`zone = '{MY-ZONE}' # eg. 'us-central1-f'`

- Special format
`topic = 'projects/{MY-PROJECT}/topics/{MY-TOPIC}'`
I feel creating another `eg` comment for the special format is overly redundant.

We already have TODO comments, I think putting in more TODOs might be an overkill. Curly brace, I feel, is sufficiently evocative of string interpolation syntax found in Perl and template libraries.

@vchudnov-g ?

broady:
@pongad LGTM. You might want a comment for the special format, though:

`resource = 'projects/{MY-PROJECT}/zone/{MY-ZONE}/foo/bar' # MY-ZONE is something like 'us-central1-f'`

vchudnov-g:
I like these last two suggestions, @pongad and @broady . They seem succinct while not making it too easy for users to simply use the sample values.

pongad:
@vchudnov-g @broady PTAL

tcoffee-google:
I continue to feel that we should handle these path names in a more structured fashion that parallels the initialization pattern for other (atomic) parameter values, e.g.:

```
# TODO: Change placeholders below to appropriate parameter values for the 'create' method:
# * The name of the topic. It must have the format `"projects/{project}/topics/{topic}"`. `{topic}` must start with a letter, and contain only letters (`[A-Za-z]`), numbers (`[0-9]`), dashes (`-`), underscores (`_`), periods (`.`), tildes (`~`), plus (`+`) or percent signs (`%`). It must be between 3 and 255 characters in length, and it must not start with `"goog"`.
project = ''
topic = ''
name = 'projects/' + project + '/topics/' + topic
```

vchudnov-g:
@pongad : Do you have an example of the "special format" case you described above? I could see for that one case having something that somewhat merges the two previous suggestions:

```
   zone = '{MY-ZONE}' # eg. 'us-central1-f'
   resource = 'projects/{MY-PROJECT}/zone/' + zone + '/foo/bar'
```

Otherwise, LGTM.

pongad:
@vchudnov-g There might be more "special formats" added over time, but for the time being, the only ones right now are the path-separated pattern `foo/{MY-FOO}`, etc. 

There's a bit of an implentation problem with your suggestion. In order to provide a good sample, we store samples in a map indexed by (API name, argument name, pattern) tuple. When the zone is an argument all by itself, it has a pattern we can see. When the zone is inside another string like this, we do not have a good pattern for it. As far as I can see, when things are "embedded" inside other strings, the pattern we have for it is just `[^/]*`.

@tcoffee-google I thought about that earlier. I thought it was a little more trouble than it's worth, since we will now have to make sure variable names don't collide.

Another way to do this is `String.format("projects/%s/topics/%s", "{MY-PROJECT}", "{MY-TOPIC}");` or something similar for compiled languages. The dynamic languages that support string interpolation can stay as is. The implementation does still get more complicated, but probably not outsized.

I'd appreciate your thoughts.

tcoffee-google:
@pongad I see two usability issues with the other suggestions that I was attempting to resolve with mine: (1) the user has to hunt inside strings to see where they need to provide values, then erase and replace just the right parts; and (2) the user has every reason to believe that the curly braces are required in the actual values.

We may need a general way to guard against name collisions, as we are currently relying on faith that our existing patterns don't cause collisions, which may not scale to all APIs (e.g., "credentials", "content" in Java, "<request>_body" in Python). Our snippets already handle collisions for imports, so implementing a similar handler for variable names would not be too hard. Alternatively, it might be even more logical in this case to provide a utility function with its own namespace:

```
# TODO: Change placeholders below to appropriate parameter values for the 'create' method:
# * The name of the topic. It must have the format `"projects/{project}/topics/{topic}"`. `{topic}` must start with a letter, and contain only letters (`[A-Za-z]`), numbers (`[0-9]`), dashes (`-`), underscores (`_`), periods (`.`), tildes (`~`), plus (`+`) or percent signs (`%`). It must be between 3 and 255 characters in length, and it must not start with `"goog"`.
def name():
  project = ''
  topic = ''
  return 'projects/' + project + '/topics/' + topic
...
request = service.projects().topics().create(name=name(), ...)
```

pongad:
@tcoffee-google I understand your misgivings about the braces, but they serve two purposes. They look like string interpolation, and so they will hopefully tell users with background in dynamic languages what to do. Perhaps more importantly, they are also illegal characters in many APIs we know. If a user forgets to replace `{MY-PROJECT}`, we can be reasonably sure that we don't try to create a new project with a funny name by mistake. According to @garrettjonesgoogle , a project called "MY-PROJECT" already exists, probably because of something like this.

Most likely, `createTopic("projects/MY-PROJECT/topics/MY-TOPIC")` will cause an unauthorized exception, since the user will probably not be the owner of the project "MY-PROJECT". This might confuse the user: Did I do auth correctly? Any setup step I missed? On the other hand `createTopic("projects/{MY-PROJECT}/topics/{MY-TOPIC}")` will cause a validation exception, which more directly points out that the user did something wrong.

The curly braces have their problems, but in my humble opinion, the benefits greatly outweighs the cost. Am I making any sense?

tcoffee-google:
That all makes sense, but in the construction I proposed, forgetting to replace the placeholders will result in ```createTopic("projects//topics/")```, which will also cause a validation exception, for reasons that are even more clear.

vchudnov-g:
What's the status of this PR? It's been inactive for a month. We should probably submit it soon, before the upcoming change #376 

For the placeholders, let's go with the ideas in the June 24 comments on this issue: inline braces and end-of-line comments:
`
resource = 'projects/{MY-PROJECT}/zone/{MY-ZONE}/foo/bar' # MY-ZONE is something like 'us-central1-f'
`

We can also keep our eyes open to see whether we encounter cases in which my July 12 idea is workable and would be helpful. If so, we can think about adding it in the future, but no need to worry about it now, and I bet it won't be necessary.

pongad:
@vchudnov-g I seem to have forgotten about this. I apologize.

I'll try to get this done for you by your Friday morning.

pongad:
I have rebased the commit. This PR leaves two concerns unaddressed. However, with the PR already so large, I would like to address them in follow up PRs. The two concerns are detailed below.

@vchudnov-g I think your June 24 suggestion is a good one. It is, however, a little difficult to implement. The reason is that we map tuple of (service name, variable name, pattern) into default string. For example, `("compute", "zone", "[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?") => "us-central1-f"` When we are dealing with slash-sepated path, however, the pattern looks like this: `^projects/[^/]*/logs/[^/]*$`. So, all that we know about individual elements is that they must not contain slashes. Should we look up only first two elements in the tuple?
After some audit, I think there is no value that both has a pattern and stored in slash-separated path. Eg, our discovery doc doesn't have `^projects/[^/]*/zone/[^/]*$`. So, even if we solve the problem right now, it won't be immediately useful. I would like the merge the PR as is, then figure out what to do with the special strings in slash-separated path later.

@broady The sample string makes some Go samples look funny:
`zone := "{MY-ZONE}"  // eg. "us-central1-f" // TODO: Update placeholder value.`
This affects Go uniquely since the TODO is on the same line as the declaration.
I have a good idea of how to fix it. Unfortunately it would touch a lot of code, and I am hesitant to heap more changes onto this already really big PR. I think we should either
- merge as is and fix later, it looks funny but shows up rarely (45 instances out of 100+ methods) and provide some value
- back out the change only for Go, fix, then pull the change back in.

I'm good with both options. Which would you prefer?

broady:
Best to move TODOs to a new line I think.

pongad:
As discussed with @broady offline, putting TODO on its own line for now.

vchudnov-g:
@pongad I think your comment is on my July 12 suggestion. Like I said, I'm fine deferring that.

On the TODOs, I'm fine with either the sample value as part of the inline TODO as cbroady@ suggested, or keeping them in separate lines.

src/main/java/com/google/api/codegen/discovery/DefaultString.java
```diff
@@ -28,12 +33,76 @@
...
+
   private static final String WILDCARD_PATTERN = "[^/]*";
 
   /**
    * Returns a default string from `pattern`, or null if the pattern is not supported.
    */
-  public static String forPattern(String pattern) {
+  @VisibleForTesting
+  static String forPattern(String pattern) {
     // We only care about patterns that has alternating literal and wildcard like
```
vchudnov-g: s/has/have/

IIUC, you never actually check that the alternating part is there. You just check that the number of slashes is consistent with the alternating pattern. I suppose you could check that all the substrings in between the "/[^/]*" separators either begin with a slash and have no other slashes or (for the last one) are empty.

src/main/java/com/google/api/codegen/discovery/DefaultString.java
```diff
@@ -28,12 +33,76 @@
...
+              "us-central1-f")
+          .put(
+              SampleKey.create("autoscaler", "zone", "[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?"),
+              "us-central1-f")
+          .put(
+              SampleKey.create("clouduseraccounts", "zone", "[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?"),
+              "us-central1-f")
+          .build();
+
+  public static DefaultString of(String apiName, String fieldName, String pattern) {
```
vchudnov-g: A test for this?

vchudnov-g:
Would be good to add a test, but otherwise LGTM.

We should also think about the pattern recognition issue I raised, but we can fix that in a separate CL.

src/main/java/com/google/api/codegen/discovery/DefaultString.java
pongad: Fix the has/have.

You're right. A more rigorous check was added in `validElems` but I didn't delete the old. The old check is deleted now.

src/main/java/com/google/api/codegen/discovery/DefaultString.java
pongad: Done.

PR #250, opened 2016-06-17
Added collection config support for complex paths

michaelbausor:
Restructured collection config generation
Added support for custom verbs, leading wildcards and literals without
corresponding wildcards
Added unit tests


michaelbausor:
Overview:
Previously we were manipulating collection configs in the form of FieldSegment objects. The FieldSegment object was not sufficient on its own to act as a collection config, as it required additional methods for templatization and name conflict resolution. This functionality was spread between static methods in the Resources class and the NamedWildcard class. The methods in the Resources class were almost exclusively for collection config-related purposes. There were several edge cases that were not handled correctly.

This PR:
- Creates a CollectionPattern class which wraps a FieldSegment object and provides the additional functionality required to generate collection configurations
- Creates a CollectionPatternsGenerator class which generates the collection config section for services and the field_name_pattern section for methods, and takes responsibility for name conflict resolution
- Removes the 4 files which previously contained this functionality
- Adds unit tests

src/main/java/com/google/api/codegen/config/CollectionPattern.java
```diff
@@ -0,0 +1,135 @@
...
+        String name = "unknown";
+        String suffix = "";
+        if (lastSegment != null && lastSegment instanceof LiteralSegment) {
+          name = Inflector.singularize(lastSegment.syntax());
+        }
+        if (((WildcardSegment) pathSegment).isUnbounded()) {
+          name += "_path";
+          suffix = "=**";
+        }
+        builder.add(new LiteralSegment(String.format("{%s%s}", name, suffix)));
```
garrettjonesgoogle: If you're just concatenating name and suffix here, but you're adding things to name earlier, you might as well just have a single string that you keep adding stuff to, and pass that one string into the format here.

michaelbausor: We need to assign name to simpleName without the suffix to keep track of the last templatized wildcard.

garrettjonesgoogle: Ahh, I see now.

src/main/java/com/google/api/codegen/config/CollectionPatternsGenerator.java
```diff
@@ -0,0 +1,126 @@
...
+      collectionMap.put(CONFIG_KEY_ENTITY_NAME, nameMap.get(resourceNameString));
+      output.add(collectionMap);
+    }
+    return output;
+  }
+
+  /**
+   * Returns a list of CollectionPattern objects.
+   */
+  public static List<CollectionPattern> getCollectionPatternsFromMethod(Method method) {
```
garrettjonesgoogle: Maybe make private? I don't see it being accessed outside of this class.

src/main/java/com/google/api/codegen/config/CollectionPatternsGenerator.java
```diff
@@ -0,0 +1,126 @@
...
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeMap;
+
+/**
+ * Class for collection config generator.
+ */
+public class CollectionPatternsGenerator implements MethodConfigGenerator {
```
garrettjonesgoogle: Conceptually this class is a little confusing and is not a cohesive abstraction. It has a generate(Method) method which implements the MethodConfigGenerator interface, but then it has a generateCollectionConfigs() method which needs to be called on this class directly. Maybe you need one unit that generates the name map, and separate units for generating the method-specific config and the interface-level collection config using that name map.

michaelbausor: Agreed, PTAL.

michaelbausor:
PTAL

src/main/java/com/google/api/codegen/config/ConfigGeneratorApi.java
```diff
@@ -32,14 +32,19 @@
 import org.yaml.snakeyaml.DumperOptions;
 import org.yaml.snakeyaml.Yaml;
 
+import autovalue.shaded.com.google.common.common.collect.ImmutableMap;
```
garrettjonesgoogle: Pull in the correct ImmutableMap

src/main/java/com/google/api/codegen/config/FieldNamePatternGenerator.java
```diff
@@ -0,0 +1,54 @@
...
+
+import com.google.api.tools.framework.model.Method;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+/**
+ * Class for collection config generator.
+ */
+public class FieldNamePatternGenerator implements MethodConfigGenerator {
```
garrettjonesgoogle: Why not use the old name (FieldNamePatternConfigGenerator)?

src/main/java/com/google/api/codegen/config/ConfigGeneratorApi.java
```diff
@@ -141,6 +146,53 @@ protected void process() throws Exception {
     return LanguageGenerator.generate(packageName);
   }
 
+  /**
+   * Generates a collection configurations section.
+   */
+  private static List<Object> generateCollectionConfigs(Map<String, String> nameMap) {
```
garrettjonesgoogle: Maybe move above generateMethodConfigs(), where the old generateCollectionConfigs used to be

michaelbausor:
PTAL

garrettjonesgoogle:
LGTM

PR #251, opened 2016-06-17
C# Support request field in REST snippet generator

chrisdunelm:


src/main/java/com/google/api/codegen/csharp/CSharpDiscoveryContext.java
```diff
@@ -188,6 +303,8 @@ public SampleInfo getSampleInfo(Method method) {
     String serviceTypeName = packageName + "Service";
     String serviceVarName = CSharpContextCommon.s_underscoresToCamelCase(packageName) + "Service";
     String methodName = CSharpContextCommon.s_underscoresToPascalCase(getSimpleName(rawMethodName));
+    boolean hasRequestField = hasRequestField(method);
+    String requestFieldName = "content";
```
jskeet: Is the aim to vary this somewhere based on the discovery doc? At the moment it might as well be a constant, unless I've missed something.

jskeet:
One question, but generally LGTM.

src/main/java/com/google/api/codegen/csharp/CSharpDiscoveryContext.java
chrisdunelm: Possibly later, possibly not; I've changed it to a constant for now.

PR #254, opened 2016-06-20
Enum extraction from discovery doc

chrisdunelm:
Add correct enum handling in C#, with all other languages treating enums like strings (as is their current behaviour

### DIFF
--- a/src/main/java/com/google/api/codegen/DiscoveryImporter.java
+++ b/src/main/java/com/google/api/codegen/DiscoveryImporter.java
@@ -20,6 +20,7 @@ import com.google.common.base.CaseFormat;
 import com.google.common.collect.ImmutableTable;
 import com.google.protobuf.Api;
 import com.google.protobuf.Field;
+import com.google.protobuf.Field.Kind;
 import com.google.protobuf.Method;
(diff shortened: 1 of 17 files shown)


jskeet:
This looks like it's probably okay - but do we have any tests that show the effect of this?

chrisdunelm:
Unfortunately the current test cases don't seem to include any methods with enum parameters.
So, no, I don't think there's a test case currently.
I've manually generated code in all 7 languages that includes an enum, and verified it's OK, but I guess I should really add a test case.

garrettjonesgoogle:
Adding @pongad for him to take a look.

pongad:
We'll probably want tests for this, but otherwise it looks good to me. This will most likely affect #249, which is OK since that's still a work in progress and there're still some design works do we done there.

chrisdunelm:
Thanks Michael,
I've added adexchangebuyer to the set of tested APIs, as this exercises the enum code.
Is this a sufficient test?

pongad:
Ah! Sorry I didn't see it. LGTM.

Hopefully we'll able to do more streamlined testing, but that's way out of the scope of this PR.

PR #259, opened 2016-06-21
Update Toolkit to be compatible with latest api-compiler

pongad:
There are a few changes:
- SimpleDiagCollector is not used, and a similar class is already
  implemented in api-compiler. It is removed.
- Model no longer implements DiagCollector but contains one.
  Code that converts Models to DiagCollectors are changed.
- YamlReader no longer has 'read' method.
  Code that uses this method is changed, but retains the old behavior.
- Visibility aspect is removed from api-compiler.
  It is also removed from toolkit.


pongad:
Looks like the change to api-compiler hasn't been uploaded yet :cry:

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "googleapis"]
```
garrettjonesgoogle: Does the .gitmodules file allow comments in it? If so, clarify that this is just used for tests.

garrettjonesgoogle:
LGTM

PR #261, opened 2016-06-22
Updates to sample initialization

michaelbausor:
Creating this PR to get feedback on the features implemented, and what test strategy to use.

Added support for map fields using {"key"} syntax
Added support for value initialization using ="value" syntax
Added support for "nested" maps and lists using chained syntax e.g.
myobject{"key"}[0].myfield

### DIFF
--- a/src/main/java/com/google/api/codegen/java/JavaGapicContext.java
+++ b/src/main/java/com/google/api/codegen/java/JavaGapicContext.java
@@ -332,4 +332,34 @@ public class JavaGapicContext extends GapicContext implements JavaContext {
   public JavaDocConfig.Builder newJavaDocConfigBuilder() {
     return JavaDocConfig.newBuilder();
   }
+
+  /**
+   * Get a Java formatted primitive value, given a primitive type and a string representation of
+   * that value. The value must be a valid example of that type. Values of type String and Bytes
+   * must contain quote characters. Values of type Bool must be either the string 'true' or 'false'
+   * (other capitalizations are permitted).
+   */
+  public String renderPrimitiveValue(TypeRef type, String value) {
+    Type primitiveType = type.getKind();
+    if (!PRIMITIVE_TYPE_MAP.containsKey(primitiveType)) {
+      throw new IllegalArgumentException(
+          "Initial values are only supported for primitive types, got type "
+              + type
+              + ", with value "
+              + value);
+    }
+    switch (primitiveType) {
+      case TYPE_BOOL:
+        return value.toLowerCase();
+      case TYPE_FLOAT:
+        return value + "F";
+      case TYPE_INT64:
+      case TYPE_UINT64:
+        return value + "L";
+      case TYPE_BYTES:
+        return "ByteString.copyFromUtf8(" + value + ")";
+      default:
+        return value;
+    }
+  }
 }
--- a/src/main/java/com/google/api/codegen/java/JavaGapicContext.java
+++ b/src/main/java/com/google/api/codegen/java/JavaGapicContext.java
@@ -335,9 +335,8 @@ public class JavaGapicContext extends GapicContext implements JavaContext {
 
   /**
    * Get a Java formatted primitive value, given a primitive type and a string representation of
-   * that value. The value must be a valid example of that type. Values of type String and Bytes
-   * must contain quote characters. Values of type Bool must be either the string 'true' or 'false'
-   * (other capitalizations are permitted).
+   * that value. The value must be a valid example of that type. Values of type Bool must be either
+   * the string 'true' or 'false' (other capitalizations are permitted).
    */
   public String renderPrimitiveValue(TypeRef type, String value) {
     Type primitiveType = type.getKind();
@@ -356,9 +355,12 @@ public class JavaGapicContext extends GapicContext implements JavaContext {
       case TYPE_INT64:
       case TYPE_UINT64:
         return value + "L";
+      case TYPE_STRING:
+        return "\"" + value + "\"";
       case TYPE_BYTES:
-        return "ByteString.copyFromUtf8(" + value + ")";
+        return "ByteString.copyFromUtf8(\"" + value + "\")";
       default:
+        // Types that do not need to be modified (e.g. TYPE_INT32) are handled here
         return value;
     }
   }
--- a/src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
+++ b/src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
@@ -32,7 +32,7 @@ import java.util.regex.Pattern;
  */
 public class FieldStructureParser {
 
-  private static Pattern fieldStructurePattern = Pattern.compile("(.+)[.]([^.]+)");
+  private static Pattern fieldStructurePattern = Pattern.compile("(.+)[.]([^.\\{\\[]+)");
   private static Pattern fieldListPattern = Pattern.compile("(.+)\\[([^\\]]+)\\]");
   private static Pattern fieldMapPattern = Pattern.compile("(.+)\\{([^\\}]+)\\}");
 
@@ -90,15 +90,15 @@ public class FieldStructureParser {
         Matcher mapMatcher = fieldMapPattern.matcher(toMatch);
         if (structureMatcher.matches()) {
           String key = structureMatcher.group(2);
-          topLevel = new AutoValue_FieldStructureParser_FieldSpec(key, topLevel);
+          topLevel = FieldSpec.create(key, topLevel);
           toMatch = structureMatcher.group(1);
         } else if (listMatcher.matches()) {
           String index = listMatcher.group(2);
-          topLevel = new AutoValue_FieldStructureParser_ListElementSpec(index, topLevel);
+          topLevel = ListElementSpec.create(index, topLevel);
           toMatch = listMatcher.group(1);
         } else if (mapMatcher.matches()) {
           String key = mapMatcher.group(2);
-          topLevel = new AutoValue_FieldStructureParser_MapElementSpec(key, topLevel);
+          topLevel = MapElementSpec.create(key, topLevel);
           toMatch = mapMatcher.group(1);
         } else {
           // No pattern match implies toMatch contains simple field (with no "." separators)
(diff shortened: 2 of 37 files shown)


src/main/java/com/google/api/codegen/java/JavaGapicContext.java
```diff
@@ -332,4 +332,34 @@ public String getMultilineHeading(String heading) {
   public JavaDocConfig.Builder newJavaDocConfigBuilder() {
     return JavaDocConfig.newBuilder();
   }
+
+  /**
+   * Get a Java formatted primitive value, given a primitive type and a string representation of
+   * that value. The value must be a valid example of that type. Values of type String and Bytes
+   * must contain quote characters. Values of type Bool must be either the string 'true' or 'false'
```
garrettjonesgoogle: regarding String/Bytes and quote characters: I see what you're doing here - piping the quote characters in from the input config. I don't think that generalizes to other languages though; e.g. for Python, we might want to use single quotes. I think you should probably strip out the quotes from the config, then add them back on for Java.

src/main/java/com/google/api/codegen/java/JavaGapicContext.java
```diff
@@ -332,4 +332,34 @@ public String getMultilineHeading(String heading) {
...
+  public String renderPrimitiveValue(TypeRef type, String value) {
+    Type primitiveType = type.getKind();
+    if (!PRIMITIVE_TYPE_MAP.containsKey(primitiveType)) {
+      throw new IllegalArgumentException(
+          "Initial values are only supported for primitive types, got type "
+              + type
+              + ", with value "
+              + value);
+    }
+    switch (primitiveType) {
```
garrettjonesgoogle: where is INT32?

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
```diff
@@ -152,6 +177,37 @@ private static Object merge(Object mergedStructure, Object unmergedStructure) {
                 + ", this index = "
                 + index);
       }
+    } else if (unmergedStructure instanceof MapElementSpec) {
```
garrettjonesgoogle: I know I created a pretty long function here, but maybe it's gone too far with this addition - it might be best to separate into multiple functions.

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
```diff
@@ -54,21 +63,37 @@
...
-          String name = listMatcher.group(1);
-          String index = listMatcher.group(2);
-          ListElementSpec listStructure = new ListElementSpec(index, topLevel);
-          topLevel = new FieldSpec(name, listStructure);
-          // TODO also support maps - use {key} notation
-        } else {
-          String name = stringParts[i];
-          topLevel = new FieldSpec(name, topLevel);
+        String toMatch = stringParts[i];
+        while (true) {
```
garrettjonesgoogle: You are hiding goto's in the syntax of a while loop

src/main/java/com/google/api/codegen/metacode/InitCodeGenerator.java
```diff
@@ -179,10 +192,83 @@ private InitCodeLine generateSampleCodeInit(
...
+                  + "; suggestedName = "
+                  + suggestedName
+                  + ", initFieldStructure = "
+                  + initFieldStructure);
+        }
+
+        Object elementInitStructure = thisFieldInitMap.get(keyString);
+        String suggestedElementName = suggestedName + "_item";
+        // Using the Optional cardinality replaces the Repeated cardinality
+        TypeRef elementType = typeRef.getMapValueField().getType();
```
garrettjonesgoogle: The comment on this doesn't make sense here

src/test/java/com/google/api/codegen/testdata/library.proto
```diff
@@ -371,4 +371,7 @@ message Comment {
 message GetBookFromArchiveRequest {
   // The name of the book to retrieve.
   string name = 1;
+
+  // Test map initialization
+  map<string, Comment> comment_map = 2;
```
garrettjonesgoogle: A GetX method isn't the most natural example for this kind of parameter

src/test/java/com/google/api/codegen/testdata/library_gapic.yaml
```diff
@@ -267,3 +267,5 @@ interfaces:
     request_object_method: false
     field_name_patterns:
       name: archived_book
+    sample_code_init_fields:
+      - comment_map{"comment_key"}.comment="default comment"
```
garrettjonesgoogle: Maybe do separate tests for map key and default value instead of testing them in a single field.

garrettjonesgoogle:
I have done a first pass. Overall I think it looks good.

src/main/java/com/google/api/codegen/java/JavaGapicContext.java
michaelbausor: INT32, UINT32, and others like the fixed size integers do not need to be modified from their numeric value, so they are handled in the default case.

garrettjonesgoogle: Oh! I didn't catch that. Maybe added a comment on the default case calling that out.

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
michaelbausor: The purpose is to recursively strip [] and {} from the end of the string. I will refactor.

michaelbausor:
PTAL

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
```diff
@@ -54,21 +73,37 @@
...
-        Matcher listMatcher = fieldListPattern.matcher(stringParts[i]);
-        if (listMatcher.matches()) {
-          String name = listMatcher.group(1);
+
+      String toMatch = equalsParts[0];
+      while (toMatch != null) {
+        Matcher structureMatcher = fieldStructurePattern.matcher(toMatch);
+        Matcher listMatcher = fieldListPattern.matcher(toMatch);
+        Matcher mapMatcher = fieldMapPattern.matcher(toMatch);
+        if (structureMatcher.matches()) {
```
garrettjonesgoogle: I have a suspicion that the pattern "myobj.mylist[0]" won't work right since the structure matcher is first (and I don't see a test for this particular pattern)

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
```diff
@@ -54,21 +73,37 @@
...
+          String key = structureMatcher.group(2);
+          topLevel = new AutoValue_FieldStructureParser_FieldSpec(key, topLevel);
```
garrettjonesgoogle: I would prefer this particular part of the code to not know that FieldSpec uses AutoValue; could you make create methods in the spec classes to encapsulate that?

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
```diff
@@ -85,8 +120,54 @@
...
+  }
+
+  interface Spec {
+    Object merge(Object mergedStructure);
+
+    Object populate();
+  }
+
+  @AutoValue
+  abstract static class FieldSpec implements Spec {
```
garrettjonesgoogle: Since the Spec classes have become heavy, I think we should pull them out into separate files.

src/test/java/com/google/api/codegen/testdata/python_main_library.baseline
```diff
@@ -839,3 +842,37 @@ class LibraryServiceApi(object):
...
+        """
+        Updates the index of a book.
+
+        Example:
+          >>> from library.library_service_api import LibraryServiceApi
+          >>> api = LibraryServiceApi()
+          >>> name = api.book_path('[SHELF]', '[BOOK]')
+          >>> index_name = ''
+          >>> index_map_item = ''
+          >>> index_map = {'default_key', index_map_item,}
```
garrettjonesgoogle: shouldn't there be a colon between key & value in python?

`index_map = {'default_key': index_map_item}`

src/test/java/com/google/api/codegen/testdata/java_main_library.baseline
```diff
@@ -1820,6 +1810,106 @@ public class LibraryServiceApi implements AutoCloseable {
...
+  /**
+   * Updates the index of a book.
+   *
+   * Sample code:
+   * <pre><code>
+   * try (LibraryServiceApi libraryServiceApi = LibraryServiceApi.createWithDefaults()) {
+   *   String formattedName = LibraryServiceApi.formatBookName("[SHELF]", "[BOOK]");
+   *   String indexName = "default index";
+   *   String indexMapItem = "";
+   *   Map&lt;String, UpdateBookIndexRequest.IndexMapEntry&gt; indexMap = new HashMap&lt;&gt;();
```
garrettjonesgoogle: `UpdateBookIndexRequest.IndexMapEntry` looks peculiar for the value type, is that right? (I haven't used map fields in proto-generated classes yet)

src/test/java/com/google/api/codegen/testdata/php_main_library.baseline
```diff
@@ -1194,6 +1196,63 @@ class LibraryServiceApi
...
+     *
+     * @throws Google\GAX\ApiException if the remote call fails
+     */
+    public function updateBookIndex($name, $indexName, $indexMap, $optionalArgs = [], $callSettings = [])
+    {
+        $request = new UpdateBookIndexRequest();
+        $request->setName($name);
+        $request->setIndexName($indexName);
+        // TODO make this work
+        putAll...
```
garrettjonesgoogle: egads

garrettjonesgoogle:
Ok I have finished my pass

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
michaelbausor: Good catch! Have updated regex

michaelbausor:
PTAL

garrettjonesgoogle:
LGTM

PR #262, opened 2016-06-22
Fix a typo

jmuk:
Without method name, it actually passes the entire mapping
instead of the call settings for a method... :-(


geigerj:
LGTM

PR #264, opened 2016-06-23
Get rid of '..' path in templates

### LINKED ISSUE #260: @extends is system-dependent
In many of our templates, we `@extends "../common.snip"`. The `..` path is unique to UNIX-like systems.

Consequently, this fails when we package the project into a JAR file (JARs are ZIP archives, which do not support `..` directories). It will most likely fail on Windows as well.

There are a few ways to solve this problem.
- The most obvious is to not go into other directories. Keep in mind that path separators are also system-dependent, which means we cannot generally go into child directories either.

- We can make a change to api-compiler and find a way to normalize the "import path". For instance, we might say that `..` always mean parent directory and paths are always slash separated. We can then translate these strings to system-dependent forms. [A paper by Rob Pike](http://plan9.bell-labs.com/sys/doc/lexnames.html) touched on this topic. (Search for `Cleaning names`.) This approach does have its problems: For example, if someone put files in a `..` directory in a system where `..` does not mean parent directory, we won't be able to access it.

pongad:
The path '..' is special to UNIX.
When running under non-UNIX-like file systems,

  @extends "../common.snip"

does not work since the directory ".." loses its special meaning.

"Non-UNIX-like file systems" may also appear on UNIX systems.
If the project is packaged into a JAR for distribution,
the '..' directory will still fail, since a JAR is a Zip archive,
and a Zip archive does not have '..' directories.

The '..' was used in the first place to reuse template code stored in
"codegen/common.snip", while the resource root is in subdirectories,
like "codegen/go".

This commit changes the resource root of all template renderings to
"codegen".
In this way, "codegen/common.snip" can simply be used by

  @extends "common.snip"

and files in subdirectories can be used by

  @extends "subdir/foo.snip".

Fixes #260.

### DIFF
--- a/src/main/java/com/google/api/codegen/discovery/CommonDiscoveryProvider.java
+++ b/src/main/java/com/google/api/codegen/discovery/CommonDiscoveryProvider.java
@@ -29,11 +29,13 @@ import java.util.Map;
  */
 public class CommonDiscoveryProvider implements DiscoveryProvider {
   private final DiscoveryContext context;
-  private final SnippetSetRunner<Method> snippetSetRunner;
+  private final SnippetSetRunner.Generator<Method> snippetSetRunner;
   private final String snippetFileName;
 
   public CommonDiscoveryProvider(
-      DiscoveryContext context, SnippetSetRunner<Method> snippetSetRunner, String snippetFileName) {
+      DiscoveryContext context,
+      SnippetSetRunner.Generator<Method> snippetSetRunner,
+      String snippetFileName) {
     this.context = context;
     this.snippetSetRunner = snippetSetRunner;
     this.snippetFileName = snippetFileName;
@@ -41,7 +43,9 @@ public class CommonDiscoveryProvider implements DiscoveryProvider {
 
   @Override
   public Map<String, Doc> generate(Method method) {
-    GeneratedResult result = snippetSetRunner.generate(method, snippetFileName, context);
+    GeneratedResult result =
+        snippetSetRunner.generate(
+            method, SnippetSetRunner.SNIPPET_RESOURCE_ROOT, snippetFileName, context);
 
     Api api = context.getApi();
     String outputRoot =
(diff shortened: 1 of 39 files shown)


src/main/java/com/google/api/codegen/discovery/CommonDiscoveryProvider.java
```diff
@@ -29,19 +29,23 @@
...
     this.snippetSetRunner = snippetSetRunner;
     this.snippetFileName = snippetFileName;
   }
 
   @Override
   public Map<String, Doc> generate(Method method) {
-    GeneratedResult result = snippetSetRunner.generate(method, snippetFileName, context);
+    GeneratedResult result =
+        snippetSetRunner.generate(
+            method, SnippetSetRunner.SNIPPET_RESOURCE_ROOT, snippetFileName, context);
```
garrettjonesgoogle: Since this never changes, why not pass it into the constructor instead?

pongad: @garrettjonesgoogle I thought about this a little more. The way that it looks right now, a Generator is essentially a function object. It holds no state, everything that affects `generate` goes through the parameter. If we put the root in the constructor, we suddenly loose this property. What do you think? I'm not strong opinionated in this case.

garrettjonesgoogle: If we put the the root in the constructor, it just becomes a closure.

pongad:
@garrettjonesgoogle PTAL

pongad:
Friendly ping @garrettjonesgoogle @vchudnov-g

vchudnov-g:
afaict, lgtm, but see what @garrettjonesgoogle has to say.

garrettjonesgoogle:
In my PR https://github.com/googleapis/toolkit/pull/268 , the treatment of snippet set runners is diverging; in the refactored model, there is only a single snippet set runner class, CommonSnippetSetRunner. I'm thinking that maybe we should wait on the changes for '..' and apply them to the refactored world.

vchudnov-g:
Hmm, this change seems relatively small enough that I would favor submitting it now, locking it down with tests, and then ensuring that in the refactored world makes the tests pass. (I assume the idea is that the baseline tests would carry over into the refactored world.)

garrettjonesgoogle:
I looked over it again and there shouldn't be any conflicts. LGTM.

pongad:
OK, I'll merge this. Of course, feel free to back this out if it causes problems :smile:

PR #265, opened 2016-06-23
Resprestsnippets

chrisdunelm:


chrisdunelm:
The first commit contains the actual modified code.
The second commit only contains the baseline file changes.

src/main/java/com/google/api/codegen/csharp/CSharpDiscoveryContext.java
```diff
@@ -170,13 +179,65 @@ public String addImport(String namespace) {
...
+            ? FluentIterable.of(new String[] {""}).append(fluentUsings.filter(isAlias))
+            : Collections.<String>emptyList();
+    return fluentUsings
+        .filter(Predicates.<String>not(isAlias))
+        .append(aliases)
+        .transform(
+            new Function<String, String>() {
+              @Override
+              public String apply(String using) {
+                return using == "" ? "" : ("using " + using + ";");
```
jskeet: Did you intend to use a reference comparison here? `using.isEmpty()` perhaps?

src/main/java/com/google/api/codegen/csharp/CSharpDiscoveryContext.java
```diff
@@ -170,13 +179,65 @@ public String addImport(String namespace) {
...
+    }
+    int typeIndex = nonGenericTypeName.lastIndexOf('.');
+    if (typeIndex < 0) {
+      throw new IllegalArgumentException("fullTypeName must be fully specified.");
+    }
+    addImport(nonGenericTypeName.substring(0, typeIndex));
+    return fullTypeName.substring(typeIndex + 1);
+  }
+
+  private String usingData(String dataRelativeTypeName) {
```
jskeet: I'd personally use a conditional operator here, but it's definitely a matter of personal preference.
(Likewise at line 222)

src/main/java/com/google/api/codegen/csharp/CSharpDiscoveryContext.java
```diff
@@ -236,6 +297,21 @@ public static ParamInfo create(
   }
 
   @AutoValue
+  public abstract static class PageStreamingInfo {
```
jskeet: Are we doing page streaming for the REST snippets? Gosh.

src/main/resources/com/google/api/codegen/csharp/discovery_fragment.snip
```diff
@@ -80,14 +80,33 @@
...
     {@sample.requestTypeName} request = {@sample.serviceVarName}.{@sample.resourcePath}.{@sample.methodName}({@sample.paramList});
-    @if sample.responseTypeName == ""
+    @if sample.responseTypeName == "void"
         request.Execute();
     @else
-        {@sample.responseTypeName} response = request.Execute();
+        @if sample.isPageStreaming
+            @let ps = sample.pageStreaming
+                {@sample.responseTypeName} response;
+                do
```
jskeet: Ah. It's manual page-streaming. Fair enough...

src/test/java/com/google/api/codegen/testdata/discoveries/csharp/csharp_adexchangebuyer.v1.4.json.baseline
```diff
@@ -146,13 +149,13 @@ namespace AdExchangeBuyerSample
             int id = 0;
 
             // TODO Add code to assign values to fields of 'content'.
```
jskeet: Nit - properties instead of fields? (Can be done later; it's not a change in this PR anyway.)

jskeet:
A few nits, but generally looks good. Would strongly suggest making the change to transform the using directives before adding in the blank line, but everything else is optional.

src/main/java/com/google/api/codegen/csharp/CSharpDiscoveryContext.java
chrisdunelm: Thanks, I make this mistake so many times :(
Moved the transform, as you suggested, so the comparison no longer required.

src/main/java/com/google/api/codegen/csharp/CSharpDiscoveryContext.java
chrisdunelm: Both done

src/main/resources/com/google/api/codegen/csharp/discovery_fragment.snip
chrisdunelm: Indeed :)

chrisdunelm:
Thanks, I've made the suggested changes, and will merge if travis passes

src/test/java/com/google/api/codegen/testdata/discoveries/csharp/csharp_adexchangebuyer.v1.4.json.baseline
chrisdunelm: Yes, that probably should be changed. I'll do it at some point...

PR #266, opened 2016-06-23
Python: convert entity name casing

geigerj:
This change ensures that Python entity names are consistent with the snake_case convention.


bjwatson:
LGTM

PR #267, opened 2016-06-24
Expand error handling in go examples

saicheems:


saicheems:
@borady

Discussion in progress... but I'll keep this updated

saicheems:
@jba updated

src/test/java/com/google/api/codegen/testdata/go_example_library.baseline
```diff
@@ -28,35 +28,53 @@ func ExampleNewClient() {
...
     ctx := context.Background()
     c, err := library.NewClient(ctx)
-    _ = err // Handle error.
+    if err != nil {
+        // TODO: Handle error.
+    }
 
     req := &google_example_library_v1.CreateShelfRequest{ /* Data... */ }
     var resp *google_example_library_v1.Shelf
     resp, err = c.CreateShelf(ctx, req)
```
broady: No need for var resp here.

`resp, err := c.CreateShelf(ctx, req)`

broady: Also for the paging example. Prefer `:=` to `=` for assignment.

src/test/java/com/google/api/codegen/testdata/go_example_library.baseline
```diff
@@ -138,45 +187,64 @@ func ExampleClient_ListBooks() {
             break
```
broady: `// Handle error`

src/test/java/com/google/api/codegen/testdata/go_example_library.baseline
saicheems: So the reason I originally added the `var resp` was because I wanted to hint the type of the response to the user in the example (admittedly, also because without the type, imports can occur that aren't necessary if the response type is in a different package than the request type, because of the way our import handler is designed).

src/test/java/com/google/api/codegen/testdata/go_example_library.baseline
saicheems: Done

src/test/java/com/google/api/codegen/testdata/go_example_library.baseline
```diff
@@ -28,185 +28,267 @@ func ExampleNewClient() {
...
     var resp *google_example_library_v1.Shelf
     for {
         resp, err = it.Next()
         if err != nil {
+            // TODO: Handle error.
             break
         }
     }
+    // TODO: Use resp.
     _ = resp
```
broady: Use resp should be within the for loop.

src/test/java/com/google/api/codegen/testdata/go_example_library.baseline
```diff
@@ -28,185 +28,267 @@ func ExampleNewClient() {
...
-    _ = err // Handle error.
+    if err != nil {
+        // TODO: Handle error.
+    }
 
     req := &google_example_library_v1.ListShelvesRequest{ /* Data... */ }
     it := c.ListShelves(ctx, req)
     var resp *google_example_library_v1.Shelf
     for {
         resp, err = it.Next()
```
broady: var resp to inside the for loop.

Also, this code leaks "err" to the outside scope. That's partly why I don't like "var resp" here.

We don't tell people the types for the discovery doc samples. We should be consistent. Also, they can just look it up in the godoc for "ListShelves".

saicheems: Got it, amended my import code to fix the potential issue.

Done.

src/test/java/com/google/api/codegen/testdata/go_example_library.baseline
saicheems: Done

saicheems:
Woohoo! Can finally assign you

src/main/resources/com/google/api/codegen/go/example.snip
```diff
@@ -41,30 +45,38 @@
             func Example{@context.getClientName(service)}_{@method.getSimpleName}() {
                 ctx := context.Background()
                 c, err := {@context.getPackageName}.New{@context.getClientName(service)}(ctx)
-                _ = err // Handle error.
+                if err != nil {
+                    // TODO: Handle error.
+                }
 
                 req := &{@inTypeName}{ /* Data... */ }
```
broady: Avoid `/* */` comments.

I think this should be

```
req := &typeName{
// TODO: fill request struct fields.
}
```

src/main/resources/com/google/api/codegen/go/example.snip
```diff
@@ -29,7 +29,11 @@
         ctx := context.Background()
         opts := []gax.ClientOption{ /* Optional client parameters. */ }
```
broady: It's unlikely anyone will actually use a slice like this. More likely they'll use

`NewClient(ctx, gax.FooOption(...))`

With that in mind, I think you could drop the options thing entirely.

broady:
This change LGTM, I made some comments unrelated to error handling.

src/main/resources/com/google/api/codegen/go/example.snip
saicheems: Good point, that was there from legacy.
I removed it anyway.

PR #272, opened 2016-06-28
Fix go err declaration in discovery snippets

saicheems:


src/main/resources/com/google/api/codegen/go/discovery_fragment.snip
```diff
@@ -67,7 +67,7 @@ func main() {
         // {@TODO()} Handle error.
       }
     @else
-      err := {@methodPath(method)}({@callArg(method)}).Context(ctx).Do()
+      err = {@methodPath(method)}({@callArg(method)}).Context(ctx).Do()
```
broady: if err := ...; err != nil {
}

vchudnov-g:
LGTM

Thanks!

src/main/resources/com/google/api/codegen/go/discovery_fragment.snip
vchudnov-g: Oh, yeah, I like this better.

broady: It's actually what we had before (and have published):
https://cloud-dot-devsite.googleplex.com/storage/docs/json_api/v1/bucketAccessControls/delete#examples

Not sure why that changed.

saicheems: I changed it to break up the line length some, but it doesn't really save that many characters.
I've put it back.

PR #276, opened 2016-06-29
Alter C# snippet wording fields->properties

chrisdunelm:
Change the comment from 'Add code to assign values to fields of' to 'Add code to assign values to properties of'


jskeet:
LGTM.

PR #277, opened 2016-06-29
Update NodeJS codegen

jmuk:
This is mostly for the result of refactoring of gax-nodejs.
The generated code is simplified, and it allows to return a value
directly from the function created by createApiCall.


geigerj:
LGTM

PR #278, opened 2016-06-29
Per-method timeout configuration

### LINKED ISSUE #127: remove timeout from constructor parameter
I am wondering why 'timeout' is a separate parameter for the constructor (in Python and Ruby).

It's defined in the client-config, and the parameter can be configurable by #70, possibly we can simply remove 'timeout' parameter, the 'default_timeout' constant, and some other related code?

geigerj:
This change permits API producers to configure default timeouts for all methods on a per-method basis. Previously, this was only possible for retrying methods.

Since this PR does not break any current codegen, we can update languages individually. Python GAPIC is updated here to demonstrate the impact; [a corresponding change to Python GAX](https://github.com/googleapis/gax-python/pull/116) will be submitted alongside this one.

cc: @jmuk 

Fixes #127

### DIFF
--- a/src/main/java/com/google/api/codegen/config/ConfigGeneratorApi.java
+++ b/src/main/java/com/google/api/codegen/config/ConfigGeneratorApi.java
@@ -128,7 +128,7 @@ public class ConfigGeneratorApi extends ToolDriverBase {
       Map<String, Object> methodConfig = new LinkedHashMap<String, Object>();
       methodConfig.put(CONFIG_KEY_METHOD_NAME, method.getSimpleName());
       for (MethodConfigGenerator generator : methodConfigGenerators) {
-        Map<String, ? extends Object> config = generator.generate(method);
+        Map<String, Object> config = generator.generate(method);
         if (config != null) {
           methodConfig.putAll(config);
         }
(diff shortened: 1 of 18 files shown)


jmuk:
Why it's not a field of retry_params?

geigerj:
It is specifically for calls that are not retrying, so I don't think it makes sense to group it with the retry settings.

jmuk:
ah, I see 😄

tbetbetbe:
LGTM, once the conflict is resolved.

src/main/java/com/google/api/codegen/config/ConfigGeneratorApi.java
```diff
@@ -122,7 +116,13 @@ protected void process() throws Exception {
             new FieldConfigGenerator(),
             new PageStreamingConfigGenerator(),
             new RetryGenerator(),
-            new FieldNamePatternConfigGenerator(collectionConfigNameMap));
+            new FieldNamePatternConfigGenerator(collectionConfigNameMap),
+            new MethodConfigGenerator() {
+              @Override
+              public Map<String, Object> generate(Method method) {
+                return ImmutableMap.of("timeout_millis", (Object) 30000);
```
garrettjonesgoogle: Make a constant out of the 30000

geigerj: Done

garrettjonesgoogle:
LGTM

PR #281, opened 2016-06-30
Update protobuf dependencies

bjwatson:


.gitignore
```diff
@@ -2,7 +2,7 @@
 bin
 build
 .gradle
-vgen/src-gen
+src-gen
```
geigerj: Do we need this line at all? I'm not sure what this is used for.

geigerj:
LGTM

.gitignore
bjwatson: there is some auto-gen source that goes there. I'm not 100% sure what does it.

PR #283, opened 2016-06-30
Fix resource name in java

shinfan:
The letter case was incorrect for the resources name in java if it contains camel case.
An example that we encountered in KMS: (entity name)keyRing -> (resource name)KeyringName

### DIFF
--- a/src/main/java/com/google/api/codegen/java/JavaGapicContext.java
+++ b/src/main/java/com/google/api/codegen/java/JavaGapicContext.java
@@ -120,6 +120,13 @@ public class JavaGapicContext extends GapicContext implements JavaContext {
   }
 
   /**
+   * Returns the resource name from the entity name.
+   */
+  public String getResourceName(String entityName) {
+    return lowerUnderscoreToUpperCamel(lowerCamelToLowerUnderscore(entityName)) + "Name";
+  }
+
+  /**
    * Adds the given type name to the import list. Returns an empty string so that the output is not
    * affected in snippets.
    */
--- a/src/main/java/com/google/api/codegen/java/JavaGapicContext.java
+++ b/src/main/java/com/google/api/codegen/java/JavaGapicContext.java
@@ -120,13 +120,6 @@ public class JavaGapicContext extends GapicContext implements JavaContext {
   }
 
   /**
-   * Returns the resource name from the entity name.
-   */
-  public String getResourceName(String entityName) {
-    return lowerUnderscoreToUpperCamel(lowerCamelToLowerUnderscore(entityName)) + "Name";
-  }
-
-  /**
    * Adds the given type name to the import list. Returns an empty string so that the output is not
    * affected in snippets.
    */
--- a/src/main/resources/com/google/api/codegen/java/common.snip
+++ b/src/main/resources/com/google/api/codegen/java/common.snip
@@ -158,7 +158,8 @@
 @end
 
 @snippet resourceName(collectionConfig)
-  {@context.getResourceName(collectionConfig.getEntityName)}
+  {@context.lowerUnderscoreToUpperCamel(\
+      context.lowerCamelToLowerUnderscore(collectionConfig.getEntityName))}Name
 @end
 
 @snippet elementTypeName(resourcesField)
--- a/src/main/resources/com/google/api/codegen/java/common.snip
+++ b/src/main/resources/com/google/api/codegen/java/common.snip
@@ -158,7 +158,7 @@
 @end
 
 @snippet resourceName(collectionConfig)
-  {@context.lowerUnderscoreToUpperCamel(collectionConfig.getEntityName)}Name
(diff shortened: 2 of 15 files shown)


src/main/resources/com/google/api/codegen/java/common.snip
```diff
@@ -158,7 +158,7 @@
 @end
 
 @snippet resourceName(collectionConfig)
-  {@context.lowerUnderscoreToUpperCamel(collectionConfig.getEntityName)}Name
+  {@context.getResourceName(collectionConfig.getEntityName)}
```
garrettjonesgoogle: Don't add a context method for this (it will create more trouble for my refactoring)

src/main/java/com/google/api/codegen/java/JavaGapicContext.java
```diff
@@ -120,6 +120,13 @@ public String getTypeName(String typeName) {
   }
 
   /**
+   * Returns the resource name from the entity name.
+   */
+  public String getResourceName(String entityName) {
+    return lowerUnderscoreToUpperCamel(lowerCamelToLowerUnderscore(entityName)) + "Name";
```
garrettjonesgoogle: why not just lowerCamelToUpperCamel? (if it doesn't exist, add it)

shinfan: It seems a bit weird to me but the entiryName can be something like: cryptoKey_path so neither lowerCamelToUpperCamel nor lowerUnderScoreToUpperCamel fit.
What I am doing here (as a work-around) is: cryptoKey_path -> crypto_key_path -> CryptoKeyPath

garrettjonesgoogle: oh weird, ok.

garrettjonesgoogle:
Why are there no baseline changes?

src/main/java/com/google/api/codegen/java/JavaGapicContext.java
michaelbausor: The "_path" component is coming from here:
https://github.com/googleapis/toolkit/blob/master/src/main/java/com/google/api/codegen/config/CollectionPattern.java#L60

I used lower underscore for the _path suffix because that is how the snippets treat that input. Maybe worth enforcing lower underscore here as well (I can submit in another PR)

shinfan:
PTAL

I found switching entity name from underscore to camel causing too many chaos in the generator since there are multiple places depending on this feature. The cleaner solution would be keeping the format and add a conversion in the config generator.

I also updated the config gen tests to cover this case to avoid this issue coming back in the future.

src/test/java/com/google/api/codegen/testdata/library.proto
```diff
@@ -105,7 +105,7 @@ service LibraryService {
 
   // Adds comments to a book
   rpc AddComments(AddCommentsRequest) returns (google.protobuf.Empty) {
-    option (google.api.http) = { post: "/v1/{name=shelves/*}/comments" body: "*" };
+    option (google.api.http) = { post: "/v1/{name=bookShelves/*}/comments" body: "*" };
```
michaelbausor: Maybe it is not worth changing now, but instead of renaming shelves->bookShelves you could rename archives->bookArchives in the proto method GetBookFromArchive below. That would cause baseline changes in a lot fewer places.

shinfan: Hmmm yes you are right but since I already made the change..I will leave it as is for now.

michaelbausor:
+1

src/test/java/com/google/api/codegen/testdata/library.proto
garrettjonesgoogle: What do you mean by "already made the change"? It hasn't been pushed to master - I'd personally prefer changes to master to be smaller than larger, given all else equal.

shinfan:
@michaelbausor I just found out changing archives would not meet my purpose since archives is not an entity name in the config. I need to test the format of a entity name so it could be either shelf or book.

michaelbausor:
Won't changing archives to bookArchives here:
https://github.com/googleapis/toolkit/pull/283/files#diff-043edd4e4c9d611d51d0b60b7f07de6dR113

Create an update in the config baseline here:
https://github.com/googleapis/toolkit/pull/283/files#diff-84565a089fee3c3303d4a8ad2cebfc88R19

michaelbausor:
As discussed offline, changing archive wouldn't affect the entity name, and changing the last part of the archive path (book->archiveBook) would mean we weren't testing the name conflict resolution (book and book_2)

michaelbausor:
So it seems sensible to keep the exisiting changes

shinfan:
@garrettjonesgoogle PTAL

garrettjonesgoogle:
LGTM

PR #284, opened 2016-07-01
Use open source api-compiler dependency

bjwatson:
Fixes https://github.com/googleapis/toolkit/issues/144


bjwatson:
Ugh, there seems to be a build failure to figure out. I noticed this on my workstation, but only when running `g build` twice without an intervening `g clean`.

bjwatson:
Related to https://github.com/googleapis/api-compiler/pull/9

shinfan:
Just a heads up that you probably need to update: https://github.com/googleapis/packman/blob/master/templates/java/build-alt.gradle.mustache to use beta-3 as well.

bjwatson:
@shinfan I did: https://github.com/bjwatson/packman/blob/745396d31685930da21b5a4c8e0725229aaf4636/templates/java/build-alt.gradle.mustache#L31

bjwatson:
@tbetbetbe This includes the fix for https://github.com/googleapis/toolkit/issues/144.

bjwatson:
@tbetbetbe @geigerj @jmuk @garrettjonesgoogle I'm at a loss for why the Travis runs are failing. I _can_ reproduce this on my workstation, but it's kind of round-a-bout:

1. Run `g clean`
2. Run `g build`
3. Edit one of the baseline files
4. Run `g build` again (do *not* run `g clean`)

That produces stack dumps like the following for apparently all of the broken tests:

```
java.lang.IllegalArgumentException: proto compilation failed with internal error: Cannot run program "null": error=2, No such file or directory
	at com.google.api.tools.framework.model.testing.TestConfig.compileProtos(TestConfig.java:263)
	at com.google.api.tools.framework.model.testing.TestConfig.<init>(TestConfig.java:98)
	at com.google.api.tools.framework.model.testing.ConfigBaselineTestCase.test(ConfigBaselineTestCase.java:145)
	at com.google.api.codegen.PythonCodeGeneratorTest$PythonLibraryBaseline.library(PythonCodeGeneratorTest.java:57)
```

The real exception seems to be happening in https://github.com/googleapis/api-compiler/blob/master/src/test/java/com/google/api/tools/framework/model/testing/TestConfig.java#L255. It thinks the first element of `commandLine` is `null`, which is initialized in https://github.com/googleapis/api-compiler/blob/master/src/test/java/com/google/api/tools/framework/model/testing/TestConfig.java#L241 (and indirectly in https://github.com/googleapis/api-compiler/blob/master/src/test/java/com/google/api/tools/framework/model/testing/TestConfig.java#L59).

I tried changing that definition to use a `private static` method, rather than a static initializer block, but it didn't help.

jmuk:
The protoc location for tests is set up through build.gradle -- see setupProtocEnvironment task.  Possibly this configuration might become invalid due to this update?

bjwatson:
The code at https://github.com/googleapis/api-compiler/blob/master/src/test/java/com/google/api/tools/framework/model/testing/TestConfig.java#L59 is pretty clear it should not define `null` for `commandLine[0]`, regardless of whether or not `protoc` exists.

Also, I can reproduce on my workstation (albeit in a strange way), which does have the correct version of `protoc` installed.

jmuk:
Again, it's set by build.gradle anyways. It could be a string `"null"` (generated wrongly from the gradle config) rather than the `null` value.

build.gradle
```diff
@@ -35,7 +35,7 @@ sourceCompatibility = 1.7
 group = "com.google.api"
 version = "0.0.0-SNAPSHOT"
 // url to the snapshot repository of a private maven repo.
-def privateSnapshotRepo = "http://104.197.230.53:8081/nexus/content/repositories/snapshots/"
+def privateSnapshotRepo = "https://oss.sonatype.org/content/repositories/snapshots/"
```
jmuk: Question: the new value doesn't look like a private maven repo.
Consider renaming -- also wondering do we really need to have a variable like this.

garrettjonesgoogle:
I will take on this change because I need to update the api-compiler dependency for another reason (mix-in support).

bjwatson:
Thanks @garrettjonesgoogle. Looks like it's GA now: https://github.com/google/protobuf/releases

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Thu Jun 30 16:37:08 PDT 2016
+#Sat Oct 03 17:00:57 PDT 2015
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.13-bin.zip
```
garrettjonesgoogle: why not 3.0?

bjwatson: IIRC, this is an auto-gen file for including `gradlew` in a repo. I was just rolling back to what's on master to avoid the CI errors that were previously plaguing this PR.

The reason for upgrading `gradlew` were driven by the choice of `protobuf-gradle-plugin` 0.7.7, which is now 0.7.3 as you did in your PR, so it doesn't seem like `gradlew` needs to be updated in this PR either.

build.gradle
```diff
@@ -32,8 +32,8 @@ sourceCompatibility = 1.7
 
 group = "com.google.api"
 version = "0.0.0-SNAPSHOT"
-// url to the snapshot repository of a private maven repo.
-def toolkitSnapshotRepo = "http://104.197.230.53:8081/nexus/content/repositories/snapshots/"
+// Nexus repository for publishing toolkit snapshots and downloading api-compiler snapshot.
+def nexusSnapshotRepo = "https://oss.sonatype.org/content/repositories/snapshots/"
```
bjwatson: @ethanbao Is there any reason why we need to continue publishing `toolkit` to the private Nexus repo vs. the official one? AFAIK, `artman` and `packman` are the only consumers of `toolkit`.

build.gradle
```diff
@@ -320,7 +318,7 @@ task showGrpcJavaPluginPath << {
 task showProtobufPath() << {
   DependencyResolver resolver = new DependencyResolver(project)
   println 'output: ' + resolver.extractArchive(
-    'com.google.protobuf:protobuf-java:3.0.0-beta-2');
+    'com.google.protobuf:protobuf-java:3.0.0-beta-3');
```
bjwatson: @garrettjonesgoogle I don't know that this makes much difference, but it seems good that this move in lockstep with `protobuf`. What do you think?

garrettjonesgoogle: oops

build.gradle
```diff
@@ -393,5 +391,6 @@ tasks.verifyGoogleJavaFormat {
   exclude 'build/**'
   exclude 'vgen/build/**'
   exclude 'vgen/.apt_generated/**'
+  exclude 'src-gen/**'
```
bjwatson: @garrettjonesgoogle I'm not 100% what creates this directory, but it appears on my workstation often (it might be running `artman` that does it). Without adding this line, `g test` fails.

garrettjonesgoogle: It's not happening on my machine... What is `g test`?

build.gradle
garrettjonesgoogle: https://github.com/googleapis/toolkit/pull/378

build.gradle
bjwatson: `gradlew test`

I have `g` setup as an alias for `gradlew`.

bjwatson:
This has been superseded by PRs from @garrettjonesgoogle.

PR #286, opened 2016-07-01
Support bundling in NodeJS

jmuk:


jmuk:
Please take a look at this too.

src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
```diff
@@ -151,7 +151,9 @@ private String returnTypeComment(Method method, MethodConfig config) {
         + callbackType
         + "} callback\n"
         + "  The function which will be called with the result of the API call.\n"
-        + "@returns {gax.EventEmitter} - the event emitter to handle the call\n"
+        + "@returns {"
+        + (config.isBundling() ? "gax.BundleEventEmitter" : "gax.EventEmitter")
+        + "} - the event emitter to handle the call\n"
```
geigerj: Is this correct if the call is bundling-enabled, but has `CallOptions.isBundling` turned off?

src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
```diff
@@ -316,6 +318,25 @@ public String getFieldType(Field field) {
...
+        return "gax.createByteLengthFunction(grpcClient."
+            + typeRef.getMessageType().getFullName()
+            + ")";
+      case TYPE_STRING:
+      case TYPE_BYTES:
+        return "function(s) { return s.length; }";
+      default:
+        // It's not quite easy how to compute the byte length for numeric
+        // values. Returns '8' (necessary for 64bit integer) always for now.
+        return "function(n) { return 8; }";
```
geigerj: This doesn't seem like a good permanent solution (what about 32-bit ints?), although we don't have any examples of bundled numeric fields at the moment. Let's add a TODO here at the minimum.

Optional: just throw an exception here, which will force us to implement this later if it ever comes up in an API.

geigerj:
LGTM

src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
jmuk: Yes, it's still a BundleEventEmitter because it does not change the class, but it's already scheduled, so its method (like runNow()) does nothing with the executor. Added comment to explain this.

src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
jmuk: Changed to throw an exception.

PR #294, opened 2016-07-11
Broaden the prereq comment in the go fragments

saicheems:
Followup to #282

Based on feedback that we weren't listing all of the necessary dependencies to install to run a sample. I feel that telling the user to call `go get -u` is more idiomatic than saying: call `go get github.com/...` and `go get github.com/...`

Let me know if there's a better way.


vchudnov-g:
lgtm, but @pongad and @broady should weigh in to confirm this is the most recommended way.

vchudnov-g:
Was this merged or folded into a separate PR?

saicheems:
Yes, the other on I have open

PR #296, opened 2016-07-12
Restructure Ruby Gapic codegen.

jmuk:
I've forgotten to update Ruby's codegen to support the new
structure which all callables are initialized in the constructor.

This also modifies some style checks usually rubocop raises warnings,
of parens and linebreak styles for invocations.

### DIFF
--- a/src/main/resources/com/google/api/codegen/ruby/main.snip
+++ b/src/main/resources/com/google/api/codegen/ruby/main.snip
@@ -57,9 +57,14 @@
        documentation = {@comments(context.defaultComments(service))}
     @if documentation
       {@documentation}
+    @#
 
     @end
+    @# @@!attribute [r] stub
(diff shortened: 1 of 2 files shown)


src/main/resources/com/google/api/codegen/ruby/main.snip
```diff
@@ -57,9 +57,14 @@
        documentation = {@comments(context.defaultComments(service))}
     @if documentation
       {@documentation}
+    @#
```
geigerj: indentation?

geigerj:
one style nit, LGTM

src/main/resources/com/google/api/codegen/ruby/main.snip
jmuk: done

PR #297, opened 2016-07-12
Add "options" into blacklisted param names

geigerj:
This is necessary because we use the `options` keyword in every generated method; if a proto request has an `options` field, we must mangle it.


geigerj:
This was done in https://github.com/googleapis/toolkit/pull/299

bjwatson:
Ok, cool.

PR #299, opened 2016-07-13
Avoid duplicated argument name with 'options'.

jmuk:
The generated code gets wrong if a message contains this name.
Although this is not a language keyword, Ruby/Python/NodeJS uses
this name for the 'CallOptions' parameter. NodeJS actually deals
with this, but Ruby/Python don't.


geigerj:
Ah, nice -- I actually opened https://github.com/googleapis/toolkit/pull/297 for Python yesterday. But this looks more complete anyway.

geigerj:
LGTM

PR #300, opened 2016-07-13
Use common language in prereq headers

saicheems:


saicheems:
I fear this will be painful to review, but the actual changes are relatively small.

I went through all the prereq comments in each of the discovery fragments and tried to use a unified language. Specifically, making sure step 1 and 2 were exactly the same in all languages, and that step 3 used a common language (imperative mood).

garrettjonesgoogle:
Implementing issue https://github.com/googleapis/toolkit/issues/201 would make changes like this less painful to review in the future...

src/main/resources/com/google/api/codegen/py/discovery_fragment.snip
```diff
@@ -15,9 +15,12 @@
...
            and check the quota for your project at
            https://console.developers.google.com/apis/api/{@apiName}
-        2. Install python client library for Google APIs using 'pip install --upgrade google-api-python-client'
-        3. This sample uses Application Default Credentials for Auth. If not already done, install the gcloud CLI from
-           https://cloud.google.com/sdk/ and run 'gcloud auth application-default login'
+        2. This sample uses Application Default Credentials for authentication. If
+           not already done, install the gcloud CLI from
+           https://cloud.google.com/sdk/ and run
+           'gcloud auth application-default login'
+        3. Install the python client library for Google APIs by running
```
vchudnov-g: In the .snips for the other languages, you don't mention the language name in step 3. Let's be consistent one way or the other. I lean toward including the language name, and capitalizing it: "the Python client library", etc.

saicheems: Ah, good catch

saicheems: Done

src/main/resources/com/google/api/codegen/go/discovery_fragment.snip
```diff
@@ -17,10 +17,11 @@ package main
...
 //    https://console.developers.google.com/apis/api/{@context.getApi.getName}
 // 2. This sample uses Application Default Credentials for authentication. If
-//    not already done, install the gcloud CLI from https://cloud.google.com/sdk/
-//    and run 'gcloud auth application-default login'
-// 3. To install the client library, run:
-//    go get -u google.golang.org/api/{@context.getApi.getName}/{@context.getApiVersion}
+//    not already done, install the gcloud CLI from
+//    https://cloud.google.com/sdk/ and run
+//    'gcloud auth application-default login'
+// 3. Install and update Go dependencies by running 'go get -u' in the project
```
vchudnov-g: s/update Go/update the Go/

vchudnov-g:
Just the one additional comment above, otherwise lgtm.

src/main/resources/com/google/api/codegen/go/discovery_fragment.snip
saicheems: http://english.stackexchange.com/questions/114467/omitting-article-the-in-front-of-plural-nouns

I've found a discussion on the topic...
Both are valid English, but if you were to ask me what "the Go dependencies" references without finishing the rest of the sentence, I might think you were referring to the standard library or the compiler's dependencies. Normally when I hear "the Go", I think of the language toolchain itself ("the Go authors") and maybe not an individual program.

My 2 cents. It might be a silly thing to take a stance on, but let me know what you think.

vchudnov-g:
The [definite article](https://en.wikipedia.org/wiki/Article_(grammar)) identifies something specific rather than a general set of things. "Install Go dependencies" reads like "install some random dependencies" or "this is how you install Go dependencies in general", whereas "Install the Go dependencies" implies much more "install the specific ones that will be needed". I still think including the article is more correct ;-)

We can be more specific and maybe avoid this discussion by saying "Install the needed Go dependencies" or "Install the Go dependencies we need" or some such....

The example in the link you posted is not quite the issue here. In a general discussion about how to run a general toolchain, I would say "To install Go dependencies you do X, but to install Java dependencies you do Y". These dependencies I refer to are not specific ones but rather any at all, and hence take no definite article (there's an implicit  "in general" in there). OTOH, if we're talking about specific dependencies, as we are here: "This code has some dependencies. To install the Go dependencies, do X" (and s/the/these/ would also be clear and correct). This is because the "some dependencies" refers to a specific set of dependencies.

saicheems:
Hahaha, one of the oddest tangents I've had in a PR...

You've convinced me to change it. Done.

Maybe it's one of those personal things... for some reason dropping the "the" in this context just *sounds* better to me.

vchudnov-g:
Yeah, it's frustratingly hard to argue concretely for things that just  _sound_ right or wrong to one's own ears.

lgtm

tcoffee-google:
Jon Parrott suggested replacing "PRE-REQUISITES" with non-jargon like "Before you run this sample", which may be particularly helpful for English language learners.

I have only one other nit: can the second "If not already done" also start on a new line?

saicheems:
@tcoffee-google Done for both.

Ran 's/PRE-REQUISITES/Before you run this sample/g'

vchudnov-g:
I'm not completely convinced about removing the PREREQUISITES header. I like the fact that the caps and the underline draw attention to it. If we do need to change the word prerequisites (which I would argue isn't jargon, but I could see possibly may be confusing to non-English speakers...perhaps), we could tweak the header to something like:

```
BEFORE RUNNING:
---------------
```

One concern I have is that we're getting opinions from several sources, so I want to make sure we're not thrashing on some of these stylistic issues. Let me loop in Andrew. In the meantime, I'm fine if we hold off on this "prerequisites" change and just submit the previous common language changes in this PR.

saicheems:
Changed it to `BEFORE RUNNING:` for the meantime.

saicheems:
Good to merge? I'll follow up with whatever feedback we get.

vchudnov-g:
LGTM

PR #302, opened 2016-07-14
Fix minor nodeJS errors

jmuk:
- missing comma at the end of null
- stub has to be bound, otherwise method call will lose 'this'.
- Adding 'grpcClient' to the enum value.


geigerj:
LGTM

PR #303, opened 2016-07-14
Fix the default value for Ruby Hash.

jmuk:


geigerj:
LGTM

geigerj:
What is the motivation for the second commit?

jmuk:
Oops, I've forgotten I sent a PR previously -- I'm revisiting this by a different approach. Close the PR -- will create another PR with the new approach. Sorry.

PR #304, opened 2016-07-14
Line-wrap method parameter descriptions consistently across languages

tcoffee-google:
Line-wrapping changes in Python fix compilecheck errors resulting from
manual line breaks in parameter descriptions in new version of Logging
API discovery doc.

Formatting of descriptions remains slightly different in C# due to
its use of separate code for line-wrapping.

### DIFF
--- a/src/main/java/com/google/api/codegen/DiscoveryContext.java
+++ b/src/main/java/com/google/api/codegen/DiscoveryContext.java
@@ -213,30 +213,46 @@ public abstract class DiscoveryContext extends CodegenContext {
     return lineEnding(stringLiteral(Strings.nullToEmpty(DefaultString.forPattern(stringPattern))));
   }
 
-  // Line wrap `str`, returning a list of lines. Each line in the returned list is guaranteed to
-  // not have new line characters.
+  // Line wrap `str`, returning a list of lines. Each line in the returned list is guaranteed to not
+  // have new line characters. The first line begins with `firstLinePrefix` (defaults to list
+  // bullet: "* "), while subsequent lines begin with a hanging indent of equal width.
+  public List<String> lineWrapDoc(String str, int maxWidth, String firstLinePrefix) {
+    return s_lineWrapDoc(str, maxWidth, firstLinePrefix);
+  }
+
   public List<String> lineWrapDoc(String str, int maxWidth) {
-    return s_lineWrapDoc(str, maxWidth);
+    return lineWrapDoc(str, maxWidth, "* ");
   }
 
   // For testing.
-  public static List<String> s_lineWrapDoc(String str, int maxWidth) {
+  public static List<String> s_lineWrapDoc(String str, int maxWidth, String firstLinePrefix) {
+    int indentWidth = firstLinePrefix.length();
+    String indent = Strings.repeat(" ", indentWidth);
+    maxWidth = maxWidth - indentWidth;
+
     List<String> lines = new ArrayList<>();
+    String prefix = firstLinePrefix;
 
     for (String line : str.trim().split("\n")) {
       line = line.trim();
 
       while (line.length() > maxWidth) {
         int split = lineWrapIndex(line, maxWidth);
-        lines.add(line.substring(0, split).trim());
+        lines.add(prefix + line.substring(0, split).trim());
         line = line.substring(split).trim();
+        prefix = indent;
       }
 
       if (!line.isEmpty()) {
-        lines.add(line);
+        lines.add(prefix + line);
       }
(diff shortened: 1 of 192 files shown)


tcoffee-google:
@vchudnov-g Note this does not seem to be the last of the Python compilecheck errors introduced by the latest Logging discovery doc; others to be resolved in a following PR.

src/test/java/com/google/api/codegen/testdata/discoveries/go/go_compute.v1.json.baseline
```diff
@@ -4041,10 +4041,10 @@ func main() {
     // TODO: Handle error.
   }
 
-  // Project ID for this request.
+  // * Project ID for this request.
   project := "" // TODO: Update placeholder value.
 
-  // The name of the zone where the instance group is located.
+  // * The name of the zone where the instance group is located.
```
vchudnov-g: Can we get rid of these asterisks?

src/test/java/com/google/api/codegen/testdata/discoveries/go/go_compute.v1.json.baseline
```diff
@@ -4090,13 +4090,13 @@ func main() {
     // TODO: Handle error.
   }
 
-  // Project ID for this request.
+  // * Project ID for this request.
```
vchudnov-g: In C# we have a general TODO above all the placeholders, and no TODOs for each parameter. Here, we do the opposite. Can we be consistent across languages? (AFAIK, there are no language-specific reasons to have diverging styles on this point, but correct me if I'm wrong)

src/main/java/com/google/api/codegen/DiscoveryContext.java
```diff
@@ -213,32 +213,48 @@ protected String getDefaultString(Type type, Field field) {
     return lineEnding(stringLiteral(Strings.nullToEmpty(DefaultString.forPattern(stringPattern))));
   }
 
-  // Line wrap `str`, returning a list of lines. Each line in the returned list is guaranteed to
-  // not have new line characters.
+  // Line wrap `str`, returning a list of lines. Each line in the returned list is guaranteed to not
+  // have new line characters. The first line begins with `firstLinePrefix` (defaults to list
+  // bullet: "* "), while subsequent lines begin with a hanging indent of equal width.
+  public List<String> lineWrapDoc(String str, int maxWidth, String firstLinePrefix) {
```
vchudnov-g: Why do we need to wrap s_lineWrapDoc instead of calling it directly? It seems like it would be less verbose to remove these two wrappers and then s/s_lineWrapDoc/lineWrapDoc/g. Is there a Java subtlety I'm missing?

tcoffee-google: The subtlety is that the snippet engine currently cannot call static methods.

src/test/java/com/google/api/codegen/testdata/discoveries/go/go_compute.v1.json.baseline
tcoffee-google: Yes, this inconsistency is on my fix list, but not in scope for this PR, which introduces minimal consistent changes to fix compilecheck.

src/test/java/com/google/api/codegen/testdata/discoveries/go/go_compute.v1.json.baseline
tcoffee-google: This only looks strange in Go because of the other inconsistency you mentioned, which will be fixed. When there is a general TODO above all the placeholders, it appears logical to offset the comments for each one so that the extent of the placeholders section is clear, as was previously done in the Python samples.

vchudnov-g: Could you add to this thread a link to the github issue where the asterisks will be fixed?

src/test/java/com/google/api/codegen/testdata/discoveries/go/go_compute.v1.json.baseline
vchudnov-g: Could you add to this thread a link to the github issue where this inconsistency will be fixed?

vchudnov-g:
Please add links to this github conversation to the github issues for the deferred issues we raised above, just so it's easy to keep track. After that, feel free to submit. LGTM.

tcoffee-google:
Remaining formatting consistency issues to be resolved under: https://github.com/googleapis/toolkit/issues/313

PR #305, opened 2016-07-14
Add shadow plugin to generate full JAR

### LINKED ISSUE #255: Add a JAR rule for runDiscoGen

saicheems:
Not much of a change, luckily! The plugin generates a working JAR by a
custom rule (`./gradlew discoJar`) and writes it to
`build/libs/discoGen-${version}.jar`.

Resolves #255

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,10 +15,12 @@ buildscript {
       url 'https://plugins.gradle.org/m2/'
     }
     mavenCentral()
+    jcenter()
   }
   dependencies {
-    classpath 'com.google.protobuf:protobuf-gradle-plugin:0.7.2-SNAPSHOT'
+    classpath "com.google.protobuf:protobuf-gradle-plugin:0.7.2-SNAPSHOT"
     classpath "gradle.plugin.com.github.sherter.google-java-format:google-java-format-gradle-plugin:0.2"
+    classpath "com.github.jengelman.gradle.plugins:shadow:1.2.3"
   }
 }
 
@@ -29,6 +31,7 @@ apply plugin: "idea"
 apply plugin: "eclipse"
 apply plugin: "com.google.protobuf"
 apply plugin: "com.github.sherter.google-java-format"
+apply plugin: "com.github.johnrengelman.shadow"
 
 sourceCompatibility = 1.7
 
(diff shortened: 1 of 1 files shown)


saicheems:
Note: I did a veeeery simple sanity check by checking that `java -cp path/to/jar com.google.foo.bar.DiscoveryFragmentGeneratorTool` failed as I expected (missing options).

vchudnov-g:
Do we need to specify the -cp flag when running? It looks from the [doc](http://imperceptiblethoughts.com/shadow/) that we should be able to run it as `java -jar build/libs/toolkit-0.0.0-SNAPSHOT-all.jar`

And this modified jar command line should be a drop-in replacement for how we're doing things currently, correct?

Cool.

saicheems:
Since we haven't specified the main class, we need to use the `-cp` flags.
We can specify it in the manifest for the jar, if that's preferred.

saicheems:
May be better to leave it as is - from what I can see in the docs, configuring the `Main-Class` attribute comes from either the jar rule or from the global `mainClassName` variable.

Should be sufficient to specify in any usage docs how the command works.

saicheems:
For the second question (`modified jar ... drop-in replacement ...?`): Yes

pongad:
I think it's OK to expect the main class. If we put in the Main-Class attribute on JAR generation, that JAR can't be used to do anything else. If we feel strongly about having a nice command for it, I'd suggest a shell script.

vchudnov-g:
I don't see a problem with having the JAR be customized for running main, since that's our only use case at the moment. It's a leaky abstraction to require the caller to fully specify the main class when trying to run it. Imagine when we refactor things--all the client have to change how they call things!

@saicheems , you say "either way makes it global". There's no way to have a jar-specific main declaration? Is that a limitation of the plugin? From my very cursory [reading](https://docs.oracle.com/javase/tutorial/deployment/jar/appman.html), it seems like we ought to be able to provide a manifest for this particular jar file, and a different one (or none at all) for other jar files we may choose to define.

saicheems:
With fresh eyes... I think there is a way to make a custom task. Will try to get it working.

saicheems:
Okay, found some subset of rules that generates a JAR with the proper main class attribute. The command is now `./gradlew discoJar` and the output is `build/libs/discoGen.jar`

vchudnov-g:
Great! Thanks!

LGTM

PR #306, opened 2016-07-14
Modify the handling of optional parameters for Ruby/NodeJS.

jmuk:
Previous: setting the default value unconditionally
Now: setting the value only when specified


jmuk:
Here you are: basically this skips setting optional values when unspecified, then we don't have to care about the default values anymore.

geigerj:
Just to confirm: given proto3 defaults, this is a style change, not a behavior change, right?

jmuk:
Right, as long as I know (everything is optional in proto3, right?)
I just want to stop caring about the default values anymore. Sometimes it's a bit tricky and I made mistakes (like using `nil` for a `Hash` field in Ruby). Probably this is the universal solution.

geigerj:
LGTM.

Probably we should do this in Python, too: #307

PR #310, opened 2016-07-14
Fix the file path / module path inconsistency issue.

### LINKED ISSUE #309: Gapic Ruby generated code should use package name for its module hierarchy
Right now the Ruby gapic generator will generate files into the path defined by package_name, but the generated code defines the modules derived from proto package name. This might cause some inconsistency.

Let's say two interfaces in an API have different proto package name (such as ',foo.v1.bar.BarService' and 'foo.v1.baz.BazService'), and the Gapic package name is 'Foo::V1'.

In that case, the files are generated under foo/v1/bar_service_api.rb, but the contents are:
```
module Foo
  module V1
    module Bar
      class BarServiceApi
           ...
```

This breaks the normal style of Ruby coding for no reasons, usually file path and module names match.
This should be `Foo::V1::BarServiceApi`, or file should be stored foo/v1/bar/bar_service_api.rb.

Also consider another case, where API relies on a common service such as google.longrunning.Operations.
In this case, the generated file like foo/v1/operations_api.rb will define a class `Google::Longrunning::OperationsApi`.  If the user installs multiple APIs depending on this, loading files will cause class redefinition and possibly an error or a warning (due to constant redefinition).
However, if this package generates google/longrunning/operaitons_api.rb instead of foo/v1, this means that multiple APIs will want to place the file into the same location, which causes another error.

Considering these cases, I think it's better to use Gapic 'package_name' for the generated Ruby classes rather than proto namespaces.

jmuk:
There are some confusions fo apiConfig's package name and
gRPC's namespace (i.e. fullName). This works well on simple
APIs, but does not work well on complicated APIs.

Now it uses:
- apiConfig's package_name is used for file path and module path
  of the generated api files.
- gRPC's namespace will be used to refer gRPC files (generated by
  protoc).

Fixes #309

### DIFF
--- a/src/main/java/com/google/api/codegen/ruby/RubyGapicContext.java
+++ b/src/main/java/com/google/api/codegen/ruby/RubyGapicContext.java
@@ -265,7 +265,7 @@ public class RubyGapicContext extends GapicContext implements RubyContext {
   /**
    * Returns the name of Ruby class for the given proto element.
    */
-  private String rubyTypeNameForProtoElement(ProtoElement element) {
+  public String rubyTypeNameForProtoElement(ProtoElement element) {
     String fullName = element.getFullName();
     int lastDot = fullName.lastIndexOf('.');
     if (lastDot < 0) {
@@ -328,6 +328,10 @@ public class RubyGapicContext extends GapicContext implements RubyContext {
       modules.add(lowerUnderscoreToUpperCamel(pkgName));
     }
     return modules;
+  }
+
+  public Iterable<String> getApiModules() {
+    return Splitter.on("::").splitToList(getApiConfig().getPackageName());
   }
 
   // Constants
--- a/src/main/java/com/google/api/codegen/ruby/RubyGapicContext.java
+++ b/src/main/java/com/google/api/codegen/ruby/RubyGapicContext.java
@@ -321,7 +321,7 @@ public class RubyGapicContext extends GapicContext implements RubyContext {
   /**
    * Returns the iterable of Ruby module names for the proto element.
    */
-  public Iterable<String> getModules(ProtoElement element) {
+  public Iterable<String> getGrpcModules(ProtoElement element) {
     String fullName = element.getFullName();
     List<String> modules = new ArrayList<>();
     for (String pkgName : Splitter.on(".").splitToList(fullName)) {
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/api/codegen/ruby/RubyGapicContext.java
```diff
@@ -330,6 +330,10 @@ public String rubyTypeName(TypeRef typeRef) {
     return modules;
   }
 
+  public Iterable<String> getApiModules() {
```
geigerj: Maybe let's rename `getModules` to `getGrpcModules`? This is the distinction, right -- that `getModules` works for types generated by `protoc`, and `getApiModules` works for types generated by GAPIC?

src/main/java/com/google/api/codegen/ruby/RubyGapicContext.java
```diff
@@ -330,6 +330,10 @@ public String rubyTypeName(TypeRef typeRef) {
...
+    return Splitter.on("::").splitToList(getApiConfig().getPackageName());
```
geigerj: Optional: This is sort of duplicated code -- it has the same functionality as the `pathMapper` that's passed into the Ruby `CommonGapicProvider`. I'm not sure it's worth going through the contortion of eliminating this duplicated code. But potentially you could pass the `pathMapper` into `RubyGapicContext` to get ride of the duplication.

geigerj:
Both comments optional. LGTM

src/main/java/com/google/api/codegen/ruby/RubyGapicContext.java
jmuk: Quickly reviewed `CommonGapicCodePathMapper`, but I am not sure it's worthwhile to extract the common pattern in this case, because the code is significantly different (it scans differently, and it needs to invoke toLowerCase).
Also having that in `CommonGapicCodePathMapper` makes me a bit nervous because it's unclear what that means in other languages.

I could extract `RubyCodePathMapper` class, but that sounds like another duplicated code among PathMappers -- and here is kinda 'one-liner' code, I think it's not worthwhile to creating a new path mapper class.

src/main/java/com/google/api/codegen/ruby/RubyGapicContext.java
jmuk: Done

Issue #313, opened 2016-07-15
Normalize discovery sample style


tcoffee-google:
The style of code and comments (wording, verbosity, etc.) have diverged between discovery snippets across different languages, and should be made uniformly polished and consistent.

Style issues and inconsistencies include:
- [ ] spacing between elements
- [ ] line-wrapping of comments (partially addressed by https://github.com/googleapis/toolkit/pull/300 )
- [ ] wording of comments/TODOs
- [ ] formatting of TODOs
- [ ] placeholder processing of response objects (e.g., pretty-printing)

Review should consider suggestions from external feedback by jonwayne: https://paste.googleplex.com/4972300784893952

@saicheems' list:
- [ ] Use `@context.lineWrapDoc` everywhere

@vchudnov-g's list:
- [ ] Consistent error handling across languages

hackathon list:
- [ ] Use `-->` to denote placeholder items instead of `*`. Make consistent across languages.
- [ ] Check `gcloud auth login`
- [ ] Add pom.xml to Java samples? It's taking Thomas and me a while to get started...
- [ ] Use JSON pretty print to show responses? It works well with Java + GSON.

tcoffee-google:
I'm closing this viz. #398 and #487, which cover the intended scope (pre-MVVM).

PR #315, opened 2016-07-18
Fix Go line wrapped comment prefix

saicheems:


saicheems:
Fixes another item in #313

vchudnov-g:
lgtm

PR #317, opened 2016-07-19
Refactoring PHP generation to use the MVVM pattern

garrettjonesgoogle:
This was prototyped in https://github.com/googleapis/toolkit/pull/268
for both Java and PHP; the present PR only includes PHP and shared
stuff.

The MVVM pattern uses the following steps:

1. The Model is converted to a ViewModel, which is a set of classes
containing the data to be used when rendering the templates.
2. The snippet engine renders the output using only data from the
ViewModel structure.

Other changes with the MVVM model:

* There is no context object when generating under the MVVM model. All
  data must be prepared beforehand.
* No casing changes are done in templates; all identifiers must be
  composed and cased correctly inside the view model.

There are a number of new utility classes created:

* Name: represents an identifier name which is casing-aware.
* TypeName: Represents a simple or complex type and keeps track of the
  aliases for the contributing types.
* TypeTable: manages the imports for a set of fully-qualified type
  names.
* SurfaceNamer: provides language-specific names or other strings.

There is also a change unrelated to MVVM:

* Using TreeSet for auth scopes to give a consistent ordering


src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
```diff
@@ -0,0 +1,260 @@
...
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * ApiMethodTransformer generates view objects from method definitions.
+ */
+public class ApiMethodTransformer {
+  private InitCodeTransformer initCodeTransformer;
```
shinfan: initCodeTransformer does not have to be a member. We can create it on demand.

src/main/java/com/google/api/codegen/metacode/InitCodeGenerator.java
```diff
@@ -119,13 +120,13 @@ private InitCodeLine generateSampleCodeInitStructure(
...
+    Name identifier = getNewSymbol(suggestedName);
     return StructureInitCodeLine.create(typeRef, identifier, fieldSettings);
   }
 
   private InitCodeLine generateSampleCodeInitList(
       String suggestedName, TypeRef typeRef, List<Object> thisFieldInitList) {
-    List<String> elementIdentifiers = new ArrayList<>();
+    List<Name> elementIdentifiers = new ArrayList<>();
     for (Object elementInitStructure : thisFieldInitList) {
       String suggestedElementName = suggestedName + "_element";
```
michaelbausor: Should a Name object be constructed here, instead of using string concatenation? And then the generateSampleCodeInit method be updated to accept a Name object instead of a string?

garrettjonesgoogle: Yes this is a good spot to use the Name class.

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
```diff
@@ -0,0 +1,260 @@
...
+        getMethodParamDocs(context, context.getMethodConfig().getRequiredFields());
+    paramDocs.add(getOptionalArrayParamDoc(context, context.getMethodConfig().getOptionalFields()));
+    paramDocs.add(getCallSettingsParamDoc(context));
+    docBuilder.paramDocs(paramDocs);
+    docBuilder.returnTypeName(
+        namer.getDynamicReturnTypeName(
+            context.getTypeTable(), context.getMethod(), context.getMethodConfig()));
+    docBuilder.throwsDocLines(new ArrayList<String>());
+
+    apiMethod.doc(docBuilder.build());
```
michaelbausor: Perhaps put docBuilder creation in its own method? And the same with methodParams?

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
```diff
@@ -0,0 +1,260 @@
...
+    List<RequestObjectParamView> optionalRequestObjectParams = new ArrayList<>();
+    for (Field field : context.getMethodConfig().getOptionalFields()) {
+      optionalRequestObjectParams.add(generateRequestObjectParam(context, field));
+    }
+    apiMethod.optionalRequestObjectParams(optionalRequestObjectParams);
+
+    return apiMethod.build();
+  }
+
+  public DynamicDefaultableParamView generateDefaultableParam(
```
michaelbausor: Do these methods (generateDefaultableParam and generateRequestObjectParam) need to be public? And, could they accept an Iterable<Field> and return a List, like getMethodParamDocs? This is only a nit, but makes the generateOptionalArrayMethod closer to being all one-liner apiMethod.<something>() calls.

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
garrettjonesgoogle: Sounds reasonable, this method got kind of long.

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
garrettjonesgoogle: That seems like a good approach, I'll try it out.

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
```diff
@@ -0,0 +1,260 @@
...
+    SimpleParamDocView.Builder retrySettingsDoc = SimpleParamDocView.newBuilder();
+    retrySettingsDoc.typeName(context.getNamer().getRetrySettingsClassName());
+
+    Name retrySettingsName = Name.from("retry", "settings");
+    Name timeoutMillisName = Name.from("timeout", "millis");
+
+    retrySettingsDoc.paramName(context.getNamer().varName(retrySettingsName));
+    String retrySettingsDocText =
+        String.format(
+            "Retry settings to use for this call. If present, then\n%s is ignored.",
```
michaelbausor: Is it necessary and/or desirable to have a newline here? This feels like either we should be relying on a formatting tool to insert the newline, or using the snippet language.

garrettjonesgoogle: The snippet language never inserts newlines into text provided to it - it doesn't have reshaping capabilities. The doc strings from proto have newlines in them, and we have to use utility methods (e.g. getDocLines()) to break those into a list of lines for the snippet language to process. Basically, those newlines are predetermined in the input.

Relying on a formatting tool might be reasonable, but this is a fixed string and we can already tell where a good break would be. One thing to keep in mind is that in general, formatting tools may not do a good job reformatting docs.

michaelbausor: When I said use the snippet language, I meant having the fixed string typed on multiple lines in the snippet rather than being encoded here. This goes somewhat against the spirit of the MVVM refactoring though.

I guess my objection is that this kind of formatting seems likely to be tedious work, especially if we wanted to insert or edit large blocks of text. If we are not expecting to be doing that, then possibly it isn't worth changing.

One advantage of doing this kind of manual newline insertion in the snip file instead of here is that it is at least kind of like a WYSIWYG editor, which makes it easier to figure out where to add the newlines.

garrettjonesgoogle: The problem with this particular text block is that it is rendered within parameter-rendering snippets, and you can't predict how long your lines are going to be anyway. I think the "right" solution is maybe to enhance the snippet language with reshaping capabilities. 

In any case, I don't know that it's an improvement to move this text into the snippet right now. The reference to the other variable name has to be correctly encoded in the text, and to reference the variable name, it would have to be explicitly encoded in the view, making things a little more complex. It would "break" the current fully general-purpose rendering of parameter docs.

shinfan: I agree with @michaelbausor here. Based on my understanding of MVVM, line wrapping is a part of rendering so it should be taken care by the views.

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -0,0 +1,232 @@
...
+        FieldStructureParser.parseFields(
+            context.getMethodConfig().getSampleCodeInitFields(), initValueConfigMap.build());
+    return initFieldStructure;
+  }
+
+  public List<InitCodeLineView> generateSurfaceInitCodeLines(
+      MethodTransformerContext context, InitCode initCode) {
+    List<InitCodeLineView> surfaceLines = new ArrayList<>();
+    for (InitCodeLine line : initCode.getLines()) {
+      switch (line.getLineType()) {
```
shinfan: line type string seems redundant here. Can we just use instanceof?

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -0,0 +1,232 @@
...
+          surfaceLines.add(generateMapInitCodeLine(context, (MapInitCodeLine) line));
+          continue;
+        default:
+          throw new RuntimeException("unhandled line type: " + line.getLineType());
+      }
+    }
+    return surfaceLines;
+  }
+
+  public StructureInitCodeLineView generateStructureInitCodeLine(
```
shinfan: Is it possible to have something polymorphism like:

public InitCodeLineView generateInitCodeLineView(MethodTransformerContext context, XXXCodeLine line) {
    ....
}

Then in your generateSurfaceInitCodeLines methond, you just need to call

surfaceLines.add(generateInitCodeLineView(context, line))

so you can just avoid all these switch cases.

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
garrettjonesgoogle: the line type enum is used in both the set of classes from init code and the views that represent them; it is more of a conceptual level than the exact implementing class.

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
garrettjonesgoogle: Not really, there needs to be a switch statement somewhere. Basically I am adding functionality to an existing class hierarchy, but outside of that hierarchy. I don't want the init code stuff to be aware of transformations, but I think it's appropriate for the transformations to be aware of the init code details. The alternative to a switch statement is the visitor pattern.

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
garrettjonesgoogle: I think it's easier to see the dependency this way. In the end I don't think it makes a huge difference since it holds no state.

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
garrettjonesgoogle: So at this point, do either of you have strong feelings on what adjustments should be made now vs deferred for later?

garrettjonesgoogle:
addressed comments by either making changes or responding to them - PTAL

src/main/java/com/google/api/codegen/viewmodel/ParamView.java
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen.viewmodel;
+
+public interface ParamView {}
```
michaelbausor: What is the purpose of this interface? Unlike ParamDocView, it doesn't seem to be used anywhere to enable polymorphism

garrettjonesgoogle: I think it is probably a holdover from an intermediate stage of my implementation when it was used to enable polymorphism, but you're right, it doesn't currently. I can remove it.

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
shinfan: I still think it is more preferable to rely on formatting tools since we do not know how long the line will be.  But I am okay with deferring this issue for later.

src/main/java/com/google/api/codegen/viewmodel/ApiMethodView.java
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen.viewmodel;
+
+public interface ApiMethodView {}
```
michaelbausor: Similarly, this has only one implementing class

michaelbausor: But perhaps this will be increased when additional languages are added?

src/main/java/com/google/api/codegen/util/php/PhpTypeTable.java
```diff
@@ -0,0 +1,168 @@
...
+import com.google.common.collect.BiMap;
+import com.google.common.collect.HashBiMap;
+import com.google.common.collect.ImmutableSet;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * The TypeTable for Java.
```
michaelbausor: s/Java/PHP

src/main/java/com/google/api/codegen/viewmodel/ApiMethodView.java
garrettjonesgoogle: It won't be used by Java, but I could see it being used by the other dynamic languages, which would use a different view from PHP's OptionalArrayMethodView.

src/main/java/com/google/api/codegen/util/php/PhpTypeTable.java
garrettjonesgoogle: oops, I will fix this.

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
garrettjonesgoogle: I will add a TO DO.

src/main/java/com/google/api/codegen/util/ViewNamer.java
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * Naming is composed of two steps:
+ *
+ * 1. Composing a Name instance with the name pieces
+ * 2. Formatting the Name for the particular type of identifier needed.
+ *
+ * This class delegates step 2 to the provided name formatter, which generally
+ * would be a language-specific namer.
+ */
+public class ViewNamer implements NameFormatter {
```
michaelbausor: I don't understand the purpose of this class. Does it exist only to be inherited from, such as by SurfaceNamer? In that case it could be abstract. If not, I don't understand why you would want a ViewNamer instead of a NameFormatter directly.

garrettjonesgoogle: Yeah it's meant to be inherited from; I'll make it abstract to make that clearer.

src/main/java/com/google/api/codegen/util/Name.java
```diff
@@ -0,0 +1,204 @@
...
+public class Name {
+  private List<NamePiece> namePieces;
+
+  /**
+   * Creates a Name from a sequence of lower-underscore strings.
+   *
+   * @throws IllegalArgumentException if any of the strings contain any characters that are not
+   * lower case or underscores.
+   */
+  public static Name from(String... pieces) {
```
michaelbausor: Why not name this lowerUnderscore, similar to lowerCamel below, and dotted in NamePath object

garrettjonesgoogle: I figured someone would bring this up - I was aiming for a shorter name in this case because it's the most common.

src/main/java/com/google/api/codegen/util/ViewNamer.java
shinfan: I just wanna mention that this class implements NameFormatter and also references a NameFormatter object so the structure looks a bit weird to me. Maybe it is more desirable to remove this class and just have SurfaceNamer which contains a language-specific name formatter.

garrettjonesgoogle: So essentially I'm trying to get a structure like a mix-in, but which isn't a mix-in since Java doesn't support them. In terms of these classes, I want the NameFormatter interface to be accessible through SurfaceNamer, but have the NameFormatter actually implemented separately. I could rename ViewNamer to NameFormatterMixin - would that help understanding?

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -39,7 +39,9 @@ public SurfaceNamer(NameFormatter languageNamer) {
     super(languageNamer);
   }
 
-  public static final String NOT_IMPLEMENTED = "$ NOT IMPLEMENTED $";
+  public String getNotImplementedString(String feature) {
```
michaelbausor: I recall in a previous PR that there was some discussion about failing the generation process in cases like this rather than outputting invalid code. Does that apply here? I don't actually have strong feelings about it, I can see a use case for wanting to generate a partial surface while you are building out a new language.

src/main/java/com/google/api/codegen/util/Name.java
michaelbausor: Hmm, OK - if you think the shorter name is worth it, I don't strongly disagree.

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
garrettjonesgoogle: IIRC that discussion was about this exact thing. The argument for failing the generation process immediately is that there would be a stack trace so you could know exactly where it failed; however, with the extra feature information I have added here, that particular aspect is mitigated. I think the full solution is to do a scan of the generated file before exiting the process, and if it contains one of these strings, then fail; that way, the overall output has still been written out and can be inspected.

src/main/java/com/google/api/codegen/util/ViewNamer.java
michaelbausor: Regarding naming - I assume the intention is for other classes (in addition to SurfaceNamer) to exist that will also inherit from this. If so, maybe the naming should reflect that relationship, perhaps as AbstractViewNamer, SurfaceViewNamer, OtherViewNamer, etc.

If there are no other classes planned, you could remove the class and roll the implementation into SurfaceNamer?

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
michaelbausor: Agreed - scanning the output and failing after the whole file has been generated is superior.

src/main/java/com/google/api/codegen/util/ViewNamer.java
shinfan: I agree that changing the name with "Mixin" would help the understanding. But after all, if there is no strong reason to have multiple sub-classes (other than SurfaceNamer), the structure would be much simpler if we roll the implementation of ViewNamer into SurfaceNamer. And the NameFormatter will remain accessible through SurfaceNamer in this case.

garrettjonesgoogle: I don't really think it would be "much simpler." For example, SurfaceNamer would implement NameFormatter (just like NameFormatterMixin now does) and forward calls to NameFormatter, which I think is one aspect that might be considered confusing about the current design, but which doesn't really change. It would just end up making SurfaceNamer a little bigger in order to move the NameFormatterMixin methods into it. I know there is no imminent sister class of SurfaceNamer, but conceivably the Discovery code might want to create such a class if they decide to use the MVVM pattern.

src/main/java/com/google/api/codegen/gapic/ViewModelGapicProvider.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    Map<String, Doc> docs = new TreeMap<>();
+    for (ViewModel surfaceDoc : surfaceDocs) {
+      if (snippetFileName != null && !surfaceDoc.templateFileName().equals(snippetFileName)) {
+        continue;
+      }
+      Doc doc = snippetSetRunner.generate(surfaceDoc);
+      if (doc == null) {
+        continue;
```
bjwatson: Will this happen often? Seems like something for which we might want to emit a warning, unless there's a common reason for this.

src/main/java/com/google/api/codegen/util/php/PhpTypeTable.java
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  public boolean hasImports() {
+    return !getImports().isEmpty();
+  }
+
+  /**
+   * A set of PHP keywords and built-ins. keywords: http://php.net/manual/en/reserved.keywords.php
+   */
+  private static final ImmutableSet<String> KEYWORD_BUILT_IN_SET =
```
bjwatson: Two comments, perhaps slightly at odds with each other:

1) This is not being used for PHP.
2) This seems like an important language-specific list. Why hide it as a private constant?

src/main/java/com/google/api/codegen/php/PhpContext.java
```diff
@@ -14,9 +14,11 @@
  */
 package com.google.api.codegen.php;
 
+import com.google.api.codegen.util.php.PhpTypeTable;
+
 /**
  * A PhpContext provides functionality specific to a context in PHP.
  */
 public interface PhpContext {
```
bjwatson: Will this package be removed once discovery moves to the MVVM pattern?

src/main/java/com/google/api/codegen/util/CommonRenderingUtil.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Utility class to process text in the templates.
+ */
+public class CommonRenderingUtil {
+
+  public static List<String> getDocLines(String text) {
+    // TODO: convert markdown to phpDoc
```
bjwatson: Do you have an issue for this?

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
```diff
@@ -0,0 +1,276 @@
...
+import java.util.List;
+
+/**
+ * ApiMethodTransformer generates view objects from method definitions.
+ */
+public class ApiMethodTransformer {
+  private InitCodeTransformer initCodeTransformer;
+
+  public ApiMethodTransformer() {
+    this.initCodeTransformer = new InitCodeTransformer();
```
bjwatson: Is there a deliberate reason to initialize here rather than at the point of field declaration? I generally prefer the latter unless there's complex logic or it depends on a constructor parameter, but it's a weak preference.

src/main/java/com/google/api/codegen/php/PhpSnippetSetRunner.java
```diff
@@ -48,15 +49,15 @@ public GeneratedResult generate(
             ImmutableMap.<String, Object>of("context", context));
 
     String outputFilename = snippets.generateFilename(element).prettyPrint();
-    PhpContextCommon phpContextCommon = new PhpContextCommon();
+    PhpTypeTable phpTypeTable = new PhpTypeTable();
 
     // TODO don't depend on a cast here
     PhpContext phpContext = (PhpContext) context;
```
bjwatson: Seems like this cast code is not used, and could be deleted.

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
```diff
@@ -0,0 +1,276 @@
...
+
+    Name optionalArgsName = Name.from("optional", "args");
+
+    paramDoc.paramName(context.getNamer().varName(optionalArgsName));
+    paramDoc.typeName(context.getNamer().getOptionalArrayTypeName());
+
+    List<String> docLines = null;
+    if (!fields.iterator().hasNext()) {
+      // TODO figure out a reliable way to line-wrap comments across all languages
+      // instead of encoding it in the transformer
```
bjwatson: 👍

src/main/resources/com/google/api/codegen/php/main.snip
```diff
@@ -108,464 +98,354 @@
...
+                '{@methodKey}' => $defaultDescriptors,
+            @end
+        ];
+        $pageStreamingDescriptors = self::getPageStreamingDescriptors();
+        foreach ($pageStreamingDescriptors as $method => $pageStreamingDescriptor) {
+            $this->descriptors[$method]['pageStreamingDescriptor'] = $pageStreamingDescriptor;
         }
 
-        return self::${@pathTemplateConstantName(collectionConfig)};
+        // TODO load the client config in a more package-friendly way
```
bjwatson: Is there an issue for this?

src/main/resources/com/google/api/codegen/php/main.snip
```diff
@@ -108,464 +98,354 @@
...
-            @let fullyQualifiedEntryName = context.fullyQualifiedName(field.getType), \
-                  entryName = context.getTypeName(fullyQualifiedEntryName)
-                foreach (${@paramLowerCamel} as $key => $value) {
-                    $request->add{@paramUpperCamel}((new {@entryName}())->setKey($key)->setValue($value));
-                }
-            @end
+@private setOptionalFieldCall(param)
+    if (isset($optionalArgs['{@param.name}'])) {
+        @if param.isMap
+            // TODO make this work
```
bjwatson: Is there an issue for this?

src/main/java/com/google/api/codegen/util/NameFormatterMixin.java
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * NameFormatterMixin is an abstract class that implements NameFormatter and
+ * simply forwards calls to another NameFormatter. This allows a child class
+ * to provide the interface of NameFormatter along with additional functionality.
+ *
+ * Note to future maintainers: This class should only contain methods which
+ * forward on to NameFormatter and nothing else; otherwise, it is no longer
+ * functioning in spirit as a mix-in.
+ */
+public abstract class NameFormatterMixin implements NameFormatter {
```
bjwatson: I don't love the Mixin name, since Java doesn't really have those (except maybe Java8 interfaces with default methods). How about NameFormatting or some other grammatical variant to indicate what concrete subclasses "is-a"?

bjwatson:
Are the context methods like `lowerUnderscoreToLowerCamel()` effectively dead code now that can be removed, thanks to the new namer classes?

src/main/java/com/google/api/codegen/gapic/ViewModelGapicProvider.java
garrettjonesgoogle: The reason for this is a failed generation. This is capture on the model itself (error count in the diag collector). We want it to work this way so that we can attempt generation of multiple files and see errors on all of them.

src/main/java/com/google/api/codegen/util/php/PhpTypeTable.java
garrettjonesgoogle: Other languages use an analogous list (as a private constant). I added it for PHP with the intent of using it at some point, we just haven't yet.

src/main/java/com/google/api/codegen/php/PhpContext.java
garrettjonesgoogle: Yes.

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
garrettjonesgoogle: Hmm I have generally weakly preferred initialization in the constructor. There isn't a particular reason.

src/main/java/com/google/api/codegen/php/PhpSnippetSetRunner.java
garrettjonesgoogle: It is used - it's necessary to call resetState on the phpContext.

src/main/java/com/google/api/codegen/gapic/ViewModelGapicProvider.java
bjwatson: I see. I'll take a look at that mechanism sometime, but seems like its out of scope for this PR.

bjwatson: Maybe it's worth a short comment here explaining this. I doubt I'm the only one who would wonder why we appear to be ignoring this null value.

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
bjwatson: That's fine. Just wondering. It's mainly a concern when there's a big swath of screen space between the declaration and definition, without a need to separate them.

src/main/java/com/google/api/codegen/php/PhpSnippetSetRunner.java
bjwatson: Doh! Overlooked that.

bjwatson: Hmm... didn't I get a bunch of grief about removing this pattern from the Python code? ;)

src/test/java/com/google/api/codegen/testdata/php_main_library.baseline
```diff
@@ -1207,7 +1213,7 @@ class LibraryServiceApi
      * try {
      *     $libraryServiceApi = new LibraryServiceApi();
      *     $formattedName = LibraryServiceApi::formatBookName("[SHELF]", "[BOOK]");
-     *     $indexName = "";
+     *     $indexName = "default index";
```
bjwatson: How did this happen? Was it previously supposed to be like this, and you fixed a bug in the init code handling?

src/test/java/com/google/api/codegen/testdata/php_main_library.baseline
```diff
@@ -42,20 +42,16 @@ use google\example\library\v1\GetBookRequest;
...
 use google\example\library\v1\MergeShelvesRequest;
 use google\example\library\v1\MoveBookRequest;
 use google\example\library\v1\PublishSeriesRequest;
-use google\example\library\v1\PublishSeriesResponse;
 use google\example\library\v1\Shelf;
 use google\example\library\v1\SomeMessage;
 use google\example\library\v1\StringBuilder;
 use google\example\library\v1\UpdateBookIndexRequest;
-use google\example\library\v1\UpdateBookIndexRequest\IndexMapEntry;
+use google\example\library\v1\UpdateBookIndexRequest.IndexMapEntry;
```
bjwatson: I don't know PHP, but this looks funny. Is it valid?

src/test/java/com/google/api/codegen/testdata/php_main_library.baseline
```diff
@@ -1243,7 +1249,7 @@ class LibraryServiceApi
         $request->setName($name);
         $request->setIndexName($indexName);
         foreach ($indexMap as $key => $value) {
-            $request->addIndexMap((new IndexMapEntry())->setKey($key)->setValue($value));
+            $request->addIndexMap((new UpdateBookIndexRequest.IndexMapEntry())->setKey($key)->setValue($value));
```
bjwatson: This looks connected to my `use` comment above.

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
shinfan: Sorry I missed this. As discussed offline, visitor pattern is probably overkill for this. I think we should leave it as-is.

bjwatson:
A general comment is that the `src/main/java/com/google/api/codegen/viewmodel` code could use a lot more documentation. These POJOs are central to how MVVM works, and explanation is needed for how they're structured.

This could wait until the Java PR, since I understand that there might be some active shuffling around of these fields right now.

src/main/java/com/google/api/codegen/gapic/ViewModelGapicProvider.java
garrettjonesgoogle: I'll add a short comment.

src/main/java/com/google/api/codegen/util/CommonRenderingUtil.java
garrettjonesgoogle: Filed one now: https://github.com/googleapis/toolkit/issues/331

src/main/java/com/google/api/codegen/php/PhpSnippetSetRunner.java
garrettjonesgoogle: This issue gets erased by MVVM :-P

src/main/resources/com/google/api/codegen/php/main.snip
garrettjonesgoogle: Filed an issue now: https://github.com/googleapis/toolkit/issues/332

src/main/resources/com/google/api/codegen/php/main.snip
garrettjonesgoogle: filed an issue now: https://github.com/googleapis/toolkit/issues/333

src/main/java/com/google/api/codegen/util/NameFormatterMixin.java
garrettjonesgoogle: I think the Mixin name conveys the intent best, because that's really what I'm trying to do. It gives the message to future maintainers that the class shouldn't have anything else in it besides calls that forward to NameFormatter. Other names lack that. Also, just doing a grammatical variant would make the situation confusing because it wouldn't be obvious what the relation was when just seeing the names.

src/test/java/com/google/api/codegen/testdata/php_main_library.baseline
garrettjonesgoogle: It looks like a feature that was added to toolkit but either wasn't implemented or wasn't correctly in PHP:

https://github.com/googleapis/toolkit/blob/master/src/test/java/com/google/api/codegen/testdata/library_gapic.yaml#L308

bjwatson:
There are a surprising number of public methods in `src/main/java/com/google/api/codegen/transformer`. I would expect that only the two methods of `ModelToViewTransformer` need to be public, since that is the interface with the rest of the `toolkit` code. Everything else should ideally be protected or package private.

src/test/java/com/google/api/codegen/testdata/php_main_library.baseline
bjwatson: Cool. Inadvertent bug fix. 👍

src/main/java/com/google/api/codegen/util/CommonRenderingUtil.java
bjwatson: A good habit is to link the TODO comment to the issue, so there's some additional context and notion of when it might be fixed.

src/main/java/com/google/api/codegen/util/php/PhpTypeTable.java
bjwatson: I'm still wondering about #2 (hiding it), but I see that you basically followed the same pattern as other languages.

src/main/java/com/google/api/codegen/util/NameFormatterMixin.java
bjwatson: This might more accurately be described as a [Proxy](https://en.wikipedia.org/wiki/Proxy_pattern). Mixin conveys the wrong thing to me, because I can't mix anything else into a `NameFormatterMixin`, due to Java's ban on multiple inheritance (at least until we move to Java 8).

garrettjonesgoogle: I don't think it matches the intent of Proxy, which is to stand in for an expensive resource. If real mixins were added to Java, I would want this class converted into a mixin. 

How about NameFormatterPseudoMixin?

src/main/java/com/google/api/codegen/util/php/PhpTypeTable.java
garrettjonesgoogle: As for # 2, it is hidden because it is hidden in the other languages.

src/main/java/com/google/api/codegen/util/CommonRenderingUtil.java
garrettjonesgoogle: I'll add a link in the code.

src/test/java/com/google/api/codegen/testdata/php_main_library.baseline
garrettjonesgoogle: I ignored this when I saw in the first time because I wasn't sure why it was happening. I have now figured out why and I'll push a fix.

src/test/java/com/google/api/codegen/testdata/php_main_library.baseline
garrettjonesgoogle: will fix (as responded above)

garrettjonesgoogle:
> A general comment is that the `src/main/java/com/google/api/codegen/viewmodel` code could use a lot more documentation. These POJOs are central to how MVVM works, and explanation is needed for how they're structured.

> This could wait until the Java PR, since I understand that there might be some active shuffling around of these fields right now.

These fields aren't likely to shuffle. Since the fields are simple POJOs, for the most part, any documentation would be a conversion from the camel case of the identifier into noun phrase form. Additional details would probably only talk about where they are used, which would only refer to other code, which isn't much of a value-add. My hope is that the naming of the fields is transparent such that they don't need documentation. 

> There are a surprising number of public methods in `src/main/java/com/google/api/codegen/transformer`. I would expect that only the two methods of `ModelToViewTransformer` need to be public, since that is the interface with the rest of the `toolkit` code. Everything else should ideally be protected or package private.

I will convert a few more methods to private that are only needed within their own class.

In general, as I argued in the previous PR, I don't think making things package private buys us very much in the context of toolkit where no one is depending on us. I would prefer to do a more thorough pass of access levels when we want to convert into a library for others to consume & extend.

src/main/java/com/google/api/codegen/viewmodel/ApiMethodView.java
garrettjonesgoogle: follow-up note - I just now started using this in the Java PR, so I'll keep it here.

garrettjonesgoogle:
Made some updates, PTAL

src/test/java/com/google/api/codegen/testdata/php_main_library.baseline
michaelbausor: I believe this feature was only implemented in Java (until now). See for example here:
https://github.com/googleapis/toolkit/blob/master/src/test/java/com/google/api/codegen/testdata/python_main_library.baseline#L856

src/main/java/com/google/api/codegen/util/NameFormatterMixin.java
bjwatson: The Proxy pattern can be used for more than just cacheing of expensive resources.

However, I realized this morning that this is actually a [Decorator](https://en.wikipedia.org/wiki/Decorator_pattern). It holds a reference to an interface rather than a concrete class, and it is told what concrete object to hold at runtime via its constructor rather than hard-coding it.

The intent of making this a mixin is fine, and we can do something like that in Java8. Nevertheless, it seems better to be as precise as possible about what this is. If you agree that this is a Decorator, then lets call it that.

bjwatson:
> In general, as I argued in the previous PR, I don't think making things package private buys us very much in the context of toolkit where no one is depending on us. I would prefer to do a more thorough pass of access levels when we want to convert into a library for others to consume & extend.

I don't think I ever responded to that. I think that in any large project, we should use the lowest level of access possible for just about everything (with the possible exception of constants). It's not just about us vs. them; it's also about what we can hold in our head at any given time, and it's helpful to see that the compiler enforces a limited interface between a model and the rest of the application.

We don't need to wait for a thorough audit to start moving in that direction now. Code cleanups, documentation, unit testing, etc. can and should happen incrementally as we do our other features.

src/test/java/com/google/api/codegen/testdata/php_main_library.baseline
```diff
@@ -230,6 +228,7 @@ class LibraryServiceApi
         return self::$archivedBookNameTemplate;
     }
 
+
```
bjwatson: Remove extra blank line

src/test/java/com/google/api/codegen/testdata/php_main_library.baseline
```diff
@@ -439,15 +439,17 @@ class LibraryServiceApi
      * @param string $options To test 'options' parameter name conflict.
      * @param array $optionalArgs {
      *     Optional.
-     *     @var SomeMessage $message Field to verify that message-type query parameter gets flattened.
+     *     @var SomeMessage $message
+     *          Field to verify that message-type query parameter gets flattened.
      *     @var StringBuilder $stringBuilder
+     *
```
bjwatson: Remove extra blank line

src/test/java/com/google/api/codegen/testdata/php_main_library.baseline
```diff
@@ -1049,19 +1055,20 @@ class LibraryServiceApi
      * @param array $optionalArgs {
      *     Optional.
      *     @var string $name
+     *
```
bjwatson: Remove extra blank line

src/main/java/com/google/api/codegen/util/NameFormatterMixin.java
shinfan: The @delegate annotation explained here is very close to this pattern: https://projectlombok.org/features/Delegate.html

bjwatson:
> These fields aren't likely to shuffle. Since the fields are simple POJOs, for the most part, any documentation would be a conversion from the camel case of the identifier into noun phrase form. Additional details would probably only talk about where they are used, which would only refer to other code, which isn't much of a value-add. My hope is that the naming of the fields is transparent such that they don't need documentation.

I wouldn't count on that. In some cases it's obvious, but in many cases it isn't. You're very familiar with this code in a way that others are not. Additional details could take the form of code examples in one or two languages, to show what it represents in concrete terms. This does provide value-add that an IDE cannot easily do for you, because it grounds the abstract concepts in the viewmodel to make them easily understandable for other developers who don't yet have the mental model that you do.

Like I said, I won't hold up this PR for documenting these classes, but I would like to see this improved in the near future while it's still fresh in your mind. A refactoring is a good opportunity to improve our game in terms of things like documentation and unit testing.

src/main/java/com/google/api/codegen/util/NameFormatterMixin.java
bjwatson: I'm not sure whether this is a pattern or tooling (like `@AutoValue`). The GoF did not have a Delegate pattern. Not that their catalogue is necessarily complete, but delegation was a building block of many of their patterns.

garrettjonesgoogle: This is not a Decorator. Its purpose is not to provide a facility to modify the behavior of the interface; it is a straight forwarding. The term Delegate makes a bit more sense to me than either Proxy or Decorator. I like it less than Mixin, but I would be willing to rename to NameFormatterDelegate.

garrettjonesgoogle:
>> In general, as I argued in the previous PR, I don't think making things package private buys us very much in the context of toolkit where no one is depending on us. I would prefer to do a more thorough pass of access levels when we want to convert into a library for others to consume & extend.

> I don't think I ever responded to that. I think that in any large project, we should use the lowest level of access possible for just about everything (with the possible exception of constants). It's not just about us vs. them; it's also about what we can hold in our head at any given time, and it's helpful to see that the compiler enforces a limited interface between a model and the rest of the application.

> We don't need to wait for a thorough audit to start moving in that direction now. Code cleanups, documentation, unit testing, etc. can and should happen incrementally as we do our other features.

I understand the idea of using the most restrictive visibility. The problem here is that the MVVM feature cuts across quite a few packages, all of which need to interact. There are components that I would ideally like to hide, but which need to be public because the need to be accessed in another package of MVVM. For example, the language specific transformers (which are in a different package as per the architectural restructuring suggested by you & Shin) need to access at least one public method in each of the general transformers. Thus, every transformer needs to be public. 

A good language feature would be to allow multiple packages to be part of a single visibility scope, but alas, we have no such thing. (C# seems to have an "internal" visibility level which limits visibility to an "assembly", which is close to what I want).

src/test/java/com/google/api/codegen/testdata/php_main_library.baseline
garrettjonesgoogle: will fix

src/test/java/com/google/api/codegen/testdata/php_main_library.baseline
garrettjonesgoogle: will fix

src/test/java/com/google/api/codegen/testdata/php_main_library.baseline
garrettjonesgoogle: will fix

src/main/java/com/google/api/codegen/util/NameFormatterMixin.java
bjwatson: A design pattern is defined in terms of a solution rather than a specific problem that it's trying to solve. This looks like a Decorator, unless you see some subtle reason why this _design_ doesn't match that pattern. Nevertheless, I understand that you want to convey the _intent_ in your class name.

I'm coming around to understanding why you designed this the way you did. It seems like you really want mixin capabilities to combine `SurfaceNamer` and `NameFormatter` functionality together in each concrete `<Language>SurfaceNamer` class, but Java7 forces you to do it in this indirect way. The problem is that it just doesn't look like a mixin and it can't really be used like one. Once we are allowed to use Java8, then I think we can properly realize your vision here.

I'm okay with Delegate, Delegator, Forwarder, or something like that as a compromise.

bjwatson:
> A good language feature would be to allow multiple packages to be part of a single visibility scope, but alas, we have no such thing. (C# seems to have an "internal" visibility level which limits visibility to an "assembly", which is close to what I want).

Okay, I think we're on the same page with this. Java unfortunately has no true concept of sub-packages, unlike Scala or C#. I think there might be ways to whittle down the visibility given that everything in the language-specific packages are subclasses of the common transformer types, but we can defer that argument until after this PR.

src/main/java/com/google/api/codegen/util/NameFormatterMixin.java
garrettjonesgoogle: > A design pattern is defined in terms of a solution rather than a specific problem that it's trying to solve. This looks like a Decorator, unless you see some subtle reason why this _design_ doesn't match that pattern. Nevertheless, I understand that you want to convey the _intent_ in your class name.

That view of design patterns is too narrow. There are cases of multiple design patterns having the same structure but having different intent - e.g. Strategy & Bridge: http://stackoverflow.com/questions/464524/what-is-the-difference-between-the-bridge-pattern-and-the-strategy-pattern . Thus, a design pattern is a design plus an intent. My intent is different from Proxy and Decorator, thus they don't fit. 

At a larger level, we shouldn't get to focused on design patterns. They are a tool that can be applied when appropriate, but we should never constrain our designs to fit the mold of design patterns when the intent does not match. 

> I'm coming around to understanding why you designed this the way you did. It seems like you really want mixin capabilities to combine `SurfaceNamer` and `NameFormatter` functionality together in each concrete `<Language>SurfaceNamer` class, but Java7 forces you to do it in this indirect way. The problem is that it just doesn't look like a mixin and it can't really be used like one. Once we are allowed to use Java8, then I think we can properly realize your vision here.

That is correct.

> I'm okay with Delegate, Delegator, Forwarder, or something like that as a compromise.

Great, I'll rename with one of those.

garrettjonesgoogle:
> Okay, I think we're on the same page with this. Java unfortunately has no true concept of sub-packages, unlike Scala or C#. I think there might be ways to whittle down the visibility given that everything in the language-specific packages are subclasses of the common transformer types, but we can defer that argument until after this PR.

Cool. 

I do want to note that for the key interaction I mentioned, between PhpGapicSurfaceTransformer and *Transformer in the general package, PhpGapicSurfaceTransformer has no structural relation to the *Transformer types; it only implements the interface ModelToViewTransformer. I have no idea how to give special access to PhpGapicSurfaceTransformer to make calls to the *Transformer types without contortions that would make the code more opaque and harder to maintain. We can discuss after the PR though.

src/main/java/com/google/api/codegen/util/NameFormatterMixin.java
bjwatson: > That view of design patterns is too narrow. There are cases of multiple design patterns having the same structure but having different intent - e.g. Strategy & Bridge: http://stackoverflow.com/questions/464524/what-is-the-difference-between-the-bridge-pattern-and-the-strategy-pattern . Thus, a design pattern is a design plus an intent. My intent is different from Proxy and Decorator, thus they don't fit.

I forgot about Strategy vs. Bridge. Thanks for reminding me. I think there's some nuance here between intent and solution, but we've spent enough time arguing philosophy for now. :)

> At a larger level, we shouldn't get to focused on design patterns. They are a tool that can be applied when appropriate, but we should never constrain our designs to fit the mold of design patterns when the intent does not match.

Agreed. When I was doing Scala, I barely thought about design patterns at all. I sometimes cynically think of design pattern as the contortions needed in OOP to do stuff that is natural in FP, but there are probably exceptions.

I only dug into design patterns here as a source of names to describe what you were trying to do.

bjwatson:
LGTM.

Let's have a 1:1 next week to discuss my remaining concerns around documentation and access modifiers. I'll put something on your calendar.

PR #318, opened 2016-07-21
Update iterator doc comments

### LINKED ISSUE #311: update iterator doc comments
We've standardized on documentation comments for iterators: see https://github.com/GoogleCloudPlatform/gcloud-golang/wiki/Iterator-Guidelines, search for "documentation comment". These should appear on each iterator method.

pongad:
Doc comments on iterator types now reflect the standard at
https://github.com/GoogleCloudPlatform/gcloud-golang/wiki/Iterator-Guidelines .

Fixes #311.

### DIFF
--- a/src/main/resources/com/google/api/codegen/go/main.snip
+++ b/src/main/resources/com/google/api/codegen/go/main.snip
@@ -313,8 +313,17 @@
                 apiCall       func() error
             }
 
-            // NextPage moves to the next page and updates its internal data.
-            // On completion, the last page is returned with Done.
+            // NextPage returns the next page of results.
+            @if hasPageSizeField
+              // It will return at most the number of results specified by the last call to SetPageSize.
+              // If SetPageSize was never called or was called with a value less than 1,
+              // the page size is determined by the underlying service.
+            @end
+            //
+            // NextPage may return a second return value of Done along with the last page of results. After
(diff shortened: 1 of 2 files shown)


src/main/resources/com/google/api/codegen/go/main.snip
```diff
@@ -313,8 +313,17 @@
                 apiCall       func() error
             }
 
-            // NextPage moves to the next page and updates its internal data.
-            // On completion, the last page is returned with Done.
+            // NextPage returns the next page of results.
+            @if hasPageSizeField
```
saicheems: Did you mean to have an empty comment here to separate?

saicheems:
+1
Aside: I'm guessing your editor is configured to 100 char line wrapping. Mine was configured to 80. Not a big deal, but better to be consistent

src/main/resources/com/google/api/codegen/go/main.snip
pongad: The doc puts them in the same paragraph. I thought I should do the same here.

jba:
LGTM. Thanks, Michael!

pongad:
The line breaks are actually copied from the doc, with the exception of `NextPage`.
The doc starts the second sentence right after the first, but I put it on the second line for the `@if` statement.

I actually stopped using automatic wrapping since Rob Pike influenced me to use variable width font.

Issue #245, opened 2016-06-15
Code sample in GAPIC API level is incorrect.


shinfan:
I found this issue when creating KMS API. The sample code in the API description comment section has the wrong return type for a page-streaming method.

Note: In the method-level, the sample code is correct for that method.

garrettjonesgoogle:
The code now picks a flattened method, so this is fixed.

PR #319, opened 2016-07-21
DO NOT COMMIT: Prototype getRepresentativeMethod

pongad:
Issue #245 is caused by the way we display a method sample in API doc.
We simply look for the first flattened method,
assume (most likely without meaning to) that it's not page-streaming,
and create a sample for it.

The issue itself can be resolved by properly detecting whether or not
the method page-streams. This can be done after MVVM to avoid duplicate
work.

But this begs the question: what the heck do we do if there is no
flattened method?

This commit prototypes a way to find a suitable method with the
following observations.
* we prefer flattened method; that's what users will probably use
* we prefer non-page-streaming methods; their samples are shorter
* we prefer methods with fewer arguments; initializations are shorter

How feasible do you think this approach is?

### DIFF
--- a/src/main/java/com/google/api/codegen/java/JavaGapicContext.java
+++ b/src/main/java/com/google/api/codegen/java/JavaGapicContext.java
@@ -16,6 +16,7 @@ package com.google.api.codegen.java;
 
 import com.google.api.codegen.ApiConfig;
 import com.google.api.codegen.GapicContext;
+import com.google.api.codegen.InterfaceConfig;
 import com.google.api.codegen.MethodConfig;
 import com.google.api.tools.framework.model.Field;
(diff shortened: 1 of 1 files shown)


garrettjonesgoogle:
Neat approach! I like it and I think it's worth rolling in.

garrettjonesgoogle:
Could you transition this over to JavaGapicSurfaceTransformer? There is a `getExampleApiMethod` which you could replace with this implementation.

pongad:
@garrettjonesgoogle I'm curious. Why is this Java-specific? Shouldn't the same method be chosen in any language?

garrettjonesgoogle:
Other languages don't have the flattened/non-flattened distinction.

pongad:
Superseded by #635

PR #323, opened 2016-07-25
Rename imports to cloud.google.com/go and add import checks

### LINKED ISSUE #320: change import paths to cloud.google.com/go
All import paths that used to begin "google.golang.org/cloud" should now be changed to "cloud.google.com/go". 

E.g. in metrics/client_test.go we currently have `"google.golang.org/cloud/logging/apiv2/metrics"`. It should be changed to `"cloud.google.com/go/logging/apiv2/metrics"`.

### LINKED ISSUE #321: add import check comments
The go tool supports a check to make sure packages are "go get"ed from the right import path. This is a comment that comes after the package declaration. Add these comments to generated code. They only need to be in one file, so add them to doc.go. They should use the new cloud.google.com/go path.

E.g. logging/doc.go should be changed to
```
package logging // import "cloud.google.com/go/logging/apiv2"
```

pongad:
This change affects two separate but related workflows.

Firstly, this change adds import check comment to the package
declaration of doc.go files.
The import path is determined by the package_name config in
GAPIC YAML file.
The YAML file used for baseline test is updated to reflect
this new convention.

While the GAPIC YAML file is meant to be edited by humans,
ConfigGenerator is responsible for creating a sensible default,
and therefore:
Secondly, this change modifies default config to also reflect
the new convention.

I still have some questions regarding the generator's behavior
when generating multiple interfaces;
I have reached out to @saicheems and @jba.
If further changes are required, I recommend that they are made
in their own issues and PRs.

Fixes #320. Fixes #321.

### DIFF
--- a/src/main/java/com/google/api/codegen/config/LanguageGenerator.java
+++ b/src/main/java/com/google/api/codegen/config/LanguageGenerator.java
@@ -17,6 +17,7 @@ package com.google.api.codegen.config;
 import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableMap;
 
+import java.util.Arrays;
 import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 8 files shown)


saicheems:
+1

jba:
+1

PR #325, opened 2016-07-25
Prevent NullPointerException in configgen

### LINKED ISSUE #316: configgen stops with NullPointerException if a service method doesn't have http option
It seems that configgen crashes when running to a service which does not have `option(google.api.http)` specification in a method. An example is https://github.com/googleapis/googleapis/blob/master/google/cloud/speech/v1/cloud_speech.proto

jmuk:
fixes #316


jmuk:
@garrettjonesgoogle , can you take a look at this?

garrettjonesgoogle:
LGTM. I made similar edits locally when playing around with the IAM mixin.

PR #326, opened 2016-07-25
Add page token support for php list methods

michaelbausor:


michaelbausor:
Note: this is the minimum change required to enable page_token usage. Because of the MVVM refactoring, I think it would be duplicated effort to try to restructure how this is handled in other languages, as @garrettjonesgoogle did here:
https://github.com/googleapis/toolkit/pull/237

PR #327, opened 2016-07-26
Redirect Go core proto path to google.golang.org/genproto/googleapis

pongad:
Go generation logic reads the import path of proto-generated types
from the protobuf file's go_package attribute.
If one doesn't exist, it falls back to a default.

This commit changes the default
and the go_package attribute of the baseline test
to reflect the new convention.


saicheems:
LGTM

PR #328, opened 2016-07-27
Stage a server-streaming method to library.proto

pongad:
This commit adds an example of server streaming gRPC method,
and generators in all languages are instructed to ignore it.
This is done so that streaming needs not be implemented
in every language at once.

### DIFF
--- a/src/main/java/com/google/api/codegen/GapicContext.java
+++ b/src/main/java/com/google/api/codegen/GapicContext.java
@@ -17,6 +17,7 @@ package com.google.api.codegen;
 import com.google.api.tools.framework.aspects.documentation.model.DocumentationUtil;
 import com.google.api.tools.framework.model.Field;
 import com.google.api.tools.framework.model.Interface;
+import com.google.api.tools.framework.model.Method;
 import com.google.api.tools.framework.model.Model;
 import com.google.api.tools.framework.model.ProtoElement;
 import com.google.common.base.Preconditions;
@@ -108,5 +109,18 @@ public class GapicContext extends CodegenContext {
       newFields.add(field);
     }
     return newFields;
+  }
+
+  /**
+   * Returns a list of simple RPC methods.
+   */
+  public List<Method> getSimpleRpcMethods(Interface service) {
+    List<Method> simples = new ArrayList<>(service.getMethods().size());
+    for (Method method : service.getMethods()) {
+      if (!method.getRequestStreaming() && !method.getResponseStreaming()) {
+        simples.add(method);
+      }
+    }
+    return simples;
   }
 }
--- a/src/main/java/com/google/api/codegen/GapicContext.java
+++ b/src/main/java/com/google/api/codegen/GapicContext.java
@@ -114,7 +114,7 @@ public class GapicContext extends CodegenContext {
   /**
    * Returns a list of simple RPC methods.
    */
-  public List<Method> getSimpleRpcMethods(Interface service) {
+  public List<Method> getNonStreamingMethods(Interface service) {
     List<Method> simples = new ArrayList<>(service.getMethods().size());
     for (Method method : service.getMethods()) {
       if (!method.getRequestStreaming() && !method.getResponseStreaming()) {
--- a/src/main/java/com/google/api/codegen/go/GoGapicContext.java
+++ b/src/main/java/com/google/api/codegen/go/GoGapicContext.java
@@ -415,7 +415,7 @@ public class GoGapicContext extends GapicContext implements GoContext {
     TreeSet<GoImport> messageImports = new TreeSet<>();
 
     // Add method request-type imports
-    // TODO(pongad): Change this back to service.getImports() once streaming is implemented.
+    // TODO(pongad): Change this back to service.getMethods() once streaming is implemented.
     //   imports for streaming methods are removed for now to not mess with tests.
     for (Method method : getNonStreamingMethods(service)) {
       MessageType inputMessage = method.getInputMessage();
--- a/src/main/java/com/google/api/codegen/go/GoGapicContext.java
+++ b/src/main/java/com/google/api/codegen/go/GoGapicContext.java
@@ -245,7 +245,7 @@ public class GoGapicContext extends GapicContext implements GoContext {
    */
   public Iterable<PageStreamingConfig> getPageStreamingConfigs(Interface service) {
     Map<String, PageStreamingConfig> streamingConfigs = new LinkedHashMap<>();
-    for (Method method : service.getMethods()) {
+    for (Method method : getSimpleRpcMethods(service)) {
       MethodConfig methodConfig =
           getApiConfig().getInterfaceConfig(service).getMethodConfig(method);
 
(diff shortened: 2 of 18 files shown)


src/main/java/com/google/api/codegen/GapicContext.java
```diff
@@ -90,4 +94,17 @@ public CollectionConfig getCollectionConfig(Interface service, String entityName
     }
     return result;
   }
+
+  /**
+   * Returns a list of simple RPC methods.
+   */
+  public List<Method> getSimpleRpcMethods(Interface service) {
```
garrettjonesgoogle: why not just name this getMethods()? Contexts for individual languages can override it and provide all methods once they support streaming. The naming change would be slightly less intrusive that way.

pongad: @garrettjonesgoogle . My thought is that methods with and without streaming will be sufficiently different that the template will want to treat them completely separately. We can implement `getStreamingMethods()` and template for each language can call it whenever it's ready.

garrettjonesgoogle: Ahh! Very good point and I agree. I think there might be a better name than "simple" though. In the MVVM view model layer, for Java methods, I am using "simple" to contrast with page streaming and bundling.

src/test/java/com/google/api/codegen/testdata/library_config.baseline
```diff
@@ -283,3 +283,14 @@ interfaces:
     field_name_patterns:
       name: book_2
     timeout_millis: 30000
+  - name: StreamShelves
+    request_object_method: false
+    page_streaming:
```
vchudnov-g: I wish this weren't called page_streaming, now.  :(

src/main/java/com/google/api/codegen/GapicContext.java
pongad: @garrettjonesgoogle I was going with the terminology in [gRPC doc](http://www.grpc.io/docs/tutorials/basic/java.html#simple-rpc), but I see how this can cause confusion. I'll change it to `getNonStreamingMethods`.

src/test/java/com/google/api/codegen/testdata/library_config.baseline
pongad: @vchudnov-g Somewhat interestingly, gRPC streaming and page streaming are not mutually exclusive. For example, an API might accept a page token and return a stream of (page token and an array of objects). We might want to "flatten" the stream. The behavior might depend on language, but the Go team has asked for this behavior.

vchudnov-g: I'm fine with them not being mutually exclusive. I just wish that the names of the two features were more distinct (ie didn't both have the word "streaming" in them). Like ("pagination", "streaming") rather than ("page streaming", "streaming"). But that's a non-trivial change at this point (and out of scope for this CL).

pongad: Ah that makes sense. Sorry I misunderstood you.

garrettjonesgoogle:
For PHP you'll need to modify transformer code.

pongad:
@garrettjonesgoogle Did I not do that at https://github.com/googleapis/toolkit/pull/328/files#diff-016a79750f6fb7f387bb0eae8cd3605dR153 ? I'm not sure if that's the best way to do it for not, but all tests pass without changing baselines.

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
```diff
@@ -147,19 +147,25 @@ private void addXApiImports(SurfaceTransformerContext context) {
...
   }
 
   private List<ApiMethodView> generateApiMethods(SurfaceTransformerContext context) {
-    List<ApiMethodView> apiMethods = new ArrayList<>();
+    List<ApiMethodView> apiMethods = new ArrayList<>(context.getInterface().getMethods().size());
 
     for (Method method : context.getInterface().getMethods()) {
+      if (method.getRequestStreaming() || method.getResponseStreaming()) {
+        continue;
+      }
```
garrettjonesgoogle: Make a single place with the logic of which methods to use.

garrettjonesgoogle:
> @garrettjonesgoogle Did I not do that at https://github.com/googleapis/toolkit/pull/328/files#diff-016a79750f6fb7f387bb0eae8cd3605dR153 ? I'm not sure if that's the best way to do it for not, but all tests pass without changing baselines.

@pongad Ahh I missed that somehow.

pongad:
@garrettjonesgoogle PTAL. I gave `SurfaceTransformerContext` its own `getNonStreamingMethods()`. It is largely a duplicate of the one in `GapicContext`, but since the latter should be removed soon, I don't think it greatly matters.

src/main/java/com/google/api/codegen/transformer/PageStreamingTransformer.java
```diff
@@ -31,7 +31,7 @@
   public List<PageStreamingDescriptorView> generateDescriptors(SurfaceTransformerContext context) {
     List<PageStreamingDescriptorView> descriptors = new ArrayList<>();
 
-    for (Method method : context.getInterface().getMethods()) {
+    for (Method method : context.getNonStreamingMethods()) {
```
garrettjonesgoogle: I was dubious about putting getNonStreamingMethods in the context, but after some thought, I have decided to embrace the approach. In this particular case, the loop is supposed to be over page streaming methods. So, could you add a getPageStreamingMethods() method in the context, use it here, and remove the isPageStreaming() check below?

src/main/java/com/google/api/codegen/transformer/SurfaceTransformerContext.java
```diff
@@ -71,4 +74,17 @@ public MethodTransformerContext asMethodContext(Method method) {
         method,
         getMethodConfig(method));
   }
+
+  /**
+   * Returns a list of simple RPC methods.
+   */
+  public List<Method> getNonStreamingMethods() {
+    List<Method> simples = new ArrayList<>(getInterface().getMethods().size());
```
garrettjonesgoogle: "simple" isn't a noun... how about `methods`?

garrettjonesgoogle:
Ok now you have to resolve with the Java MVVM change :-)

src/main/java/com/google/api/codegen/transformer/SurfaceTransformerContext.java
pongad: Done

src/main/java/com/google/api/codegen/transformer/PageStreamingTransformer.java
pongad: Done

pongad:
@garrettjonesgoogle PTAL. Hopefully this can get merged before the next MVVM lands. What does MVMM stand for anyway...?

src/main/java/com/google/api/codegen/go/GoGapicContext.java
```diff
@@ -415,7 +415,9 @@ private GoImport createMessageImport(MessageType messageType) {
     TreeSet<GoImport> messageImports = new TreeSet<>();
 
     // Add method request-type imports
-    for (Method method : service.getMethods()) {
+    // TODO(pongad): Change this back to service.getImports() once streaming is implemented.
```
garrettjonesgoogle: Do you mean "change back to service.getMethods()"?

garrettjonesgoogle:
LGTM after you fix the minor issue in the comments.

There are no other languages on deck for MVVM yet. I will coordinate that shortly.

MVMM doesn't stand for anything :-P MVVM stands for Model-View-ViewModel.

src/main/java/com/google/api/codegen/go/GoGapicContext.java
pongad: Done.

garrettjonesgoogle:
LGTM

pongad:
This seems to have caused unexpected failures. I'll back it out and submit a new PR

PR #334, opened 2016-07-28
Separate generation for py/rb docs

### LINKED ISSUE #239: Separate doc pipeline from GAPIC pipeline
For languages that generate "doc" files (i.e., Python, Ruby), there should be separate pipelines for generating documentation and generating packages. The GAPIC pipeline should not produce the doc files.

cc: @jmuk @bjwatson

geigerj:
This permits the codegen to be run without generating documentation-
only classes. For Python, this fixes an issue where the documentation-
only classes were shadowing real protocol buffer classes.

~~A change to artman will be needed to re-enable documenation
generation via artman.~~ Actually, no change to artman is necessary,
since the switch to turn doc generation on/off is handled entirely by the
GAPIC config.

Fixes #239.


src/main/java/com/google/api/codegen/gapic/MainGapicProviderFactory.java
```diff
@@ -235,10 +220,27 @@
               .setCodePathMapper(pythonPathMapper)
               .build();
 
+      if (id.equals(PYTHON)) {
+        System.out.println("stopping here");
```
garrettjonesgoogle: Do you still need the printouts?

src/main/java/com/google/api/codegen/gapic/MainGapicProviderFactory.java
```diff
@@ -235,10 +220,27 @@
...
+                  new PythonSnippetSetRunner<ProtoFile>(
+                      new PythonProtoFileInitializer(), SnippetSetRunner.SNIPPET_RESOURCE_ROOT))
+              .setSnippetFileNames(Arrays.asList("py/message.snip"))
+              .setCodePathMapper(CommonGapicCodePathMapper.defaultInstance())
+              .build();
       return Arrays.<GapicProvider<? extends Object>>asList(
           mainProvider, messageProvider, clientConfigProvider);
 
-    } else if (id.equals(RUBY)) {
+    } else if (id.equals(RUBY) || id.equals(RUBY_DOC)) {
```
garrettjonesgoogle: This logic is somewhat confusing to me. Is RUBY_DOC a superset of RUBY?

geigerj: Yes, "doc" is a superset. It includes an additional set of files that are not functional but that can be used to generate docs. For Python this is problematic because these files actually shadow real, useful files and break the client.

src/main/java/com/google/api/codegen/gapic/MainGapicProviderFactory.java
geigerj: Whoops, fixed

src/main/java/com/google/api/codegen/gapic/MainGapicProviderFactory.java
```diff
@@ -282,9 +286,7 @@
     }
   }
 
-  /**
-   * Create the GapicProviders based on the given id
-   */
+  /** Create the GapicProviders based on the given id */
```
bjwatson: This kind of stuff can be minimized if you configure Eclipse to only update styling on code regions you modify rather than updating the entire file.

src/test/java/com/google/api/codegen/testdata/python_doc_main_library.baseline
```diff
@@ -0,0 +1,876 @@
+============== file: library/library_service_api.py ==============
```
bjwatson: Why don't the baseline tests include the bogus proto files with documentation?

src/main/java/com/google/api/codegen/gapic/MainGapicProviderFactory.java
bjwatson: That's why I mark this kind of stuff with FIXME, so it's obvious it needs to go before approval.

src/main/java/com/google/api/codegen/gapic/MainGapicProviderFactory.java
geigerj: This is actually what I have set -- I'm not sure why it's happening. I've actually just completely disabled auto-reformatting to avoid this sort of thing in the future.

src/test/java/com/google/api/codegen/testdata/python_doc_main_library.baseline
geigerj: They do -- see `python_doc_message_library.baseline`. During one of the refactorings, there was a behavior change so that a single Java test corresponds to multiple baseline files -- one baseline file for each template.

geigerj: Ah, looks like the reason it's not showing up in the diff is that git treats the new file as a renaming from the old baseline test for the non-doc test.

bjwatson: Oh, I see that now.

bjwatson:
LGTM

jmuk:
I am thinking the name of the ID can be more composable rather than adding "<lang>_doc" as new IDs. Like, `id` to be an array, and normal Python generation will be simply `["python"]` while Python with docgen is going to be `["python", "doc"]`.  I am not particularly sure this is beneficial to this situation -- anyways it's only Python and Ruby probably.

Otherwise LGTM too.

geigerj:
Hm -- I'm not sure. Either you need to create more logic in toolkit to know that "doc" means Python doc when in an array with "python" and Ruby otherwise, or else you still need to call it as `["python", "python_doc"]` to disambiguate. In the latter case, I don't know if there's an advantage, since the doc file is not really currently that useful without the other files, too.

PR #335, opened 2016-07-28
Updates to PHP code generation

michaelbausor:
Add page token support for page streaming methods
Merge optionalArgs and callSettings optional parameters
Add credential and keyFile optional args


src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
```diff
@@ -95,6 +97,35 @@ private ApiMethodDocView generateOptionalArrayMethodDoc(MethodTransformerContext
     return docBuilder.build();
   }
 
+  /*
+   * Constructs a list of the optional fields for the method. Adds the page token field, which is
+   * missing from the optional fields list, and rearrange so that page token and page size are last.
```
garrettjonesgoogle: I think I'd really prefer the other languages to be the "odd ones out", where they have to skip the page token. We shouldn't have to contort PHP to get the full list of optional fields.

michaelbausor: Uploaded a fix. I removed the reordering of optional args as well. Not sure if we want to leave that in?

src/main/java/com/google/api/codegen/py/PythonGapicContext.java
```diff
@@ -250,7 +250,8 @@ private String returnTypeComment(
 
     // parameter types
     contentBuilder.append("Args:\n");
-    for (Field field : this.messages().flattenedFields(method.getInputType())) {
+    for (Field field :
+        removePageTokenFromFields(this.messages().flattenedFields(method.getInputType()), config)) {
```
garrettjonesgoogle: I don't think the call to flattenedFields is necessary here

michaelbausor: I'm not sure what you mean - what would it be replaced with?

garrettjonesgoogle: Ahh, I realize that was unclear. What I mean is that the implementation of flattenedFields could be substituted here:

removePageTokenFromFields(method.getInputType().getMessageType().getFields()), config)

And then maybe flattenedFields isn't needed anywhere anymore.

michaelbausor: Ah, got it - fixed

garrettjonesgoogle:
LGTM

PR #337, opened 2016-07-29
Refactoring Java generation to use the MVVM pattern

garrettjonesgoogle:


src/main/java/com/google/api/codegen/transformer/ApiCallableTransformer.java
```diff
@@ -0,0 +1,159 @@
...
+                  typeTable, pageStreaming.getResourcesField().getType());
+
+      pagedApiCallableBuilder.requestTypeName(
+          typeTable.getAndSaveNicknameFor(method.getInputType()));
+      pagedApiCallableBuilder.responseTypeName(pagedResponseTypeName);
+      pagedApiCallableBuilder.name(context.getNamer().getPagedCallableName(method));
+      pagedApiCallableBuilder.settingsFunctionName(
+          context.getNamer().getSettingsFunctionName(method));
+
+      apiCallables.add(pagedApiCallableBuilder.type(ApiCallableType.PagedApiCallable).build());
```
michaelbausor: nit, very minor - why use the builder's fluent api only in this last line? Maybe either put pagedApiCallableBuilder.type(...) on its own line, or use the fluent api for all the calls to the builder?

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
```diff
@@ -44,6 +54,237 @@ public ApiMethodTransformer() {
...
+    setCommonFields(context, methodViewBuilder);
+    methodViewBuilder.name(namer.getPagedCallableMethodName(context.getMethod()));
+    setListMethodFields(context, methodViewBuilder);
+    setCallableMethodFields(
+        context, namer.getPagedCallableName(context.getMethod()), methodViewBuilder);
+
+    return methodViewBuilder.type(ApiMethodType.PagedCallableMethod).build();
+  }
+
+  public StaticApiMethodView generateUnpagedListCallableMethod(MethodTransformerContext context) {
```
michaelbausor: nit: why List in this method name, but not e.g. generatePagedCallableMethod

src/main/java/com/google/api/codegen/transformer/BundlingTransformer.java
```diff
@@ -0,0 +1,105 @@
...
+    desc.responseTypeName(typeTable.getAndSaveNicknameFor(method.getOutputType()));
+    desc.bundledFieldTypeName(
+        typeTable.getAndSaveNicknameFor(bundling.getBundledField().getType()));
+    desc.subresponseTypeName(
+        typeTable.getAndSaveNicknameFor(bundling.getSubresponseField().getType()));
+
+    desc.partitionKeys(generatePartitionKeys(context));
+    desc.discriminatorFieldCopies(generateDiscriminatorFieldCopies(context));
+
+    desc.fnGetBundledField(namer.getGetFunctionCallName(bundling.getBundledField()));
```
michaelbausor: Nit: suggest renaming fnGetBundledField to bundledFieldGetFn or bundledFieldGetFunction, I found reading this line somewhat confusing, although that may be just me

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTransformer.java
```diff
@@ -0,0 +1,340 @@
...
+  @Override
+  public List<String> getTemplateFileNames() {
+    return Arrays.asList(
+        XAPI_TEMPLATE_FILENAME, XSETTINGS_TEMPLATE_FILENAME, PACKAGE_INFO_TEMPLATE_FILENAME);
+  }
+
+  @Override
+  public List<ViewModel> transform(Model model, ApiConfig apiConfig) {
+    List<ViewModel> surfaceDocs = new ArrayList<>();
+    SurfaceNamer namer = new JavaSurfaceNamer();
```
shinfan: I suggest we use language-specific namer instead of the general SurfaceNamer in the language-specific transformers. One benefit of doing this is to have the flexibility of adding naming method for only one language. For example, we can roll the builder related names from parent class into the JavaSurfaceNamer.

src/main/java/com/google/api/codegen/transformer/ApiCallableTransformer.java
garrettjonesgoogle: Good point. I'll put it on its own line.

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
garrettjonesgoogle: So it turns out that you were the one who introduced this terminology:

https://github.com/googleapis/toolkit/commit/a9805c835f81f39735814724c3b3a9ee03277be0#diff-e184ca12b1548f2c4cc6af9e753e70dc

In defense of the naming scheme, paged calls are always on list methods, whereas the term "unpaged" could theoretically be applied to both list and non-list methods.

michaelbausor: Oops... I think your defense is correct though, it should stay as generateUnpagedListCallableMethod to distinguish it from generateCallableMethod, which is also un-paged

src/main/java/com/google/api/codegen/transformer/BundlingTransformer.java
garrettjonesgoogle: I'll rename all of the methods that follow the fn* pattern.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTransformer.java
garrettjonesgoogle: True, but I think it's more flexible for future maintenance to go through SurfaceNamer. It allows multiple languages to use the same patterns if it makes sense to do so in the future. It also means there is one reference point for all of the naming patterns that we have (we just have a small set of language-specific overrides).

src/main/java/com/google/api/codegen/transformer/ApiCallableTransformer.java
```diff
@@ -0,0 +1,160 @@
...
+        namer.getNotImplementedString(notImplementedPrefix + "bundlingDescriptorName"));
+
+    if (methodConfig.isPageStreaming()) {
+      namer.addPageStreamingCallSettingsImports(typeTable);
+      settings.type(ApiCallableType.PagedApiCallable);
+      TypeRef resourceType = methodConfig.getPageStreaming().getResourcesField().getType();
+      settings.resourceTypeName(typeTable.getAndSaveNicknameForElementType(resourceType));
+      settings.pageStreamingDescriptorName(namer.getPageStreamingDescriptorConstName(method));
+    } else {
+      if (methodConfig.isBundling()) {
```
michaelbausor: Why not a normal else if?

garrettjonesgoogle: Given this context, it would make sense to use a normal else if. I think it ended up this way because in other places, there were some things shared by bundling & simple. I will change it here.

src/main/java/com/google/api/codegen/transformer/ApiCallableTransformer.java
```diff
@@ -0,0 +1,160 @@
...
+      }
+    }
+
+    settings.memberName(namer.getSettingsMemberName(method));
+    settings.settingsGetFunction(namer.getSettingsFunctionName(method));
+
+    return Arrays.asList(settings.build());
+  }
+
+  private BundlingConfigView generateBundlingConfig(MethodTransformerContext context) {
```
michaelbausor: It seems like this will be useful outside of this class in other languages that support bundling. I don't know where it could be placed at the moment that would make more sense though, maybe ApiMethodTransformer?

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
```diff
@@ -44,6 +54,237 @@ public ApiMethodTransformer() {
...
+      MethodTransformerContext context, StaticApiMethodView.Builder methodViewBuilder) {
+    SurfaceNamer namer = context.getNamer();
+    methodViewBuilder.apiRequestTypeName(
+        context.getTypeTable().getAndSaveNicknameFor(context.getMethod().getInputType()));
+    methodViewBuilder.apiClassName(namer.getApiWrapperClassName(context.getInterface()));
+    methodViewBuilder.apiVariableName(namer.getApiWrapperVariableName(context.getInterface()));
+    methodViewBuilder.settingsGetterName(namer.getSettingsFunctionName(context.getMethod()));
+  }
+
+  private void setListMethodFields(
```
michaelbausor: Why is this the only method that is private?

garrettjonesgoogle: Hmm that wasn't really intentional. I think it is an artifact of how I generated the method. I will make more methods in this class private.

src/main/java/com/google/api/codegen/transformer/ApiCallableTransformer.java
garrettjonesgoogle: Hmm maybe BundlingTransformer.

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
```diff
@@ -201,6 +441,16 @@ private RequestObjectParamView generateRequestObjectParam(
     return allDocs;
   }
 
+  public SimpleParamDocView getRequestObjectParamDoc(
```
michaelbausor: This method and getOptionalArrayParamDoc could both return a List<ParamDocView> object like getMethodParamDocs. This simplifies the calling code and gives all the get*ParamDoc methods a common return type.

src/main/java/com/google/api/codegen/transformer/ApiCallableTransformer.java
michaelbausor: Yes that makes sense. It seems like it could be a static method, although the MVVM restructure doesn't seem to use them very much so maybe that wouldn't fit?

garrettjonesgoogle: Well, technically, all of the MVVM methods could be static, because none of them need any state (the only state is held in the type table in the context). I made them all member methods for future flexibility.

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -117,10 +170,30 @@ public String getPageStreamingDescriptorName(Method method) {
...
+
+  public String getBundlingDescriptorConstName(Method method) {
+    return inittedConstantName(Name.upperCamel(method.getSimpleName()).join("bundling_desc"));
+  }
+
   public void addPageStreamingDescriptorImports(ModelTypeTable typeTable) {
     // do nothing
   }
 
+  public void addBundlingDescriptorImports(ModelTypeTable typeTable) {
```
michaelbausor: Is it ever valid to construct an instance of SurfaceNamer, rather than a language-specific child class? If not, it could be marked abstract, and these methods could be made abstract.

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
garrettjonesgoogle: I tried it out and I don't think it simplifies anything.

* setRequestObjectMethodFields: only needs to wrap the result of getRequestObjectParamDoc in Arrays.asList()
* generateOptionalArrayMethodDoc: just adds the result to an existing List that it is building up.

For both of these cases, they logically create a singleton doc. I don't think it makes sense for them to return a List.

garrettjonesgoogle:
I have addressed all outstanding comments, PTAL.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTransformer.java
```diff
@@ -0,0 +1,340 @@
...
+        pathTemplateTransformer.generateFormatResourceFunctions(context));
+    xapiClass.parseResourceFunctions(
+        pathTemplateTransformer.generateParseResourceFunctions(context));
+    xapiClass.apiMethods(methods);
+
+    // must be done as the last step to catch all imports
+    xapiClass.imports(context.getTypeTable().getImports());
+
+    String outputPath = pathMapper.getOutputPath(context.getInterface(), context.getApiConfig());
+    xapiClass.outputPath(outputPath + "/" + name + ".java");
```
michaelbausor: Maybe this could use a path builder of some sort, rather than hard-coded forward slash separator?

src/main/java/com/google/api/codegen/transformer/java/JavaModelTypeNameConverter.java
```diff
@@ -0,0 +1,236 @@
...
+    if (type.isRepeated()) {
+      return TypedValue.create(typeNameConverter.getTypeName("java.util.ArrayList"), "new %s<>()");
+    }
+    if (PRIMITIVE_ZERO_VALUE.containsKey(type.getKind())) {
+      return TypedValue.create(getTypeName(type), PRIMITIVE_ZERO_VALUE.get(type.getKind()));
+    }
+    if (type.isMessage()) {
+      return TypedValue.create(getTypeName(type), "%s.newBuilder().build()");
+    }
+    return TypedValue.create(new TypeName(""), "null");
```
michaelbausor: It may make no difference, but maybe this should pass in getTypeName(type) instead of an empty TypeName, similar to the PRIMITIVE_ZERO_VALUE case where it also makes no difference.

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceNamer.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * The SurfaceNamer for Java.
+ */
+public class JavaSurfaceNamer extends SurfaceNamer {
+
+  /**
+   * Standard constructor.
```
michaelbausor: Maybe remove this comment

src/main/java/com/google/api/codegen/viewmodel/StaticXApiView.java
```diff
@@ -0,0 +1,87 @@
...
+ */
+package com.google.api.codegen.viewmodel;
+
+import com.google.api.codegen.SnippetSetRunner;
+import com.google.auto.value.AutoValue;
+
+import java.util.List;
+
+@AutoValue
+public abstract class StaticXApiView implements ViewModel {
```
michaelbausor: Why are these called StaticXApiView and StaticXSettingsView, instead of just XApiView?

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
garrettjonesgoogle: In theory, it could be useful for a new language for SurfaceNamer to not be abstract. It would be a bit more burdensome to force all implementations to be provided during incremental development.

src/test/java/com/google/api/codegen/testdata/java_settings_library.baseline
```diff
@@ -386,7 +370,7 @@ public class LibraryServiceSettings extends ServiceApiSettings {
       new BundlingDescriptor<PublishSeriesRequest, PublishSeriesResponse>() {
         @Override
         public String getBundlePartitionKey(PublishSeriesRequest request) {
-          return request.getEdition() + "|" + request.getName();
+          return request.getEdition() + "|" + request.getName() + "|";
```
michaelbausor: Do we want a trailing '|'?

src/main/java/com/google/api/codegen/viewmodel/StaticXApiView.java
garrettjonesgoogle: Because there is a DynamicXApiView, although the dynamic languages have no corresponding SettingsView.

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
michaelbausor: I can see that, and I don't have a super strong opinion. The counter argument would be that while other methods can emit "Not Implemented" strings, these void-returning methods have no way to indicate that they aren't doing any work. So someone implementing a new language may not realize they need to implement the methods, because all that will go wrong is that imports will be missing.

src/test/java/com/google/api/codegen/testdata/java_settings_library.baseline
garrettjonesgoogle: It's easier to generate (in the snippet language) and doesn't hurt anything...

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
garrettjonesgoogle: True, but at the same time, it would be somewhat annoying if the new language actually didn't require any imports, and they had to implement an empty method because we forced them to. 

In the end I'm not super concerned with language implementers missing things because we are producing a mechanism to compile and run-check generated surfaces. Even if an implementer is forced to implement a method like this and they add some imports, that doesn't mean they would get the list right; it only means they are *slightly* more likely to get it right.

src/main/resources/com/google/api/codegen/java/settings.snip
```diff
@@ -116,310 +78,356 @@
...
+          }
+        };
+    {@""}
   @end
 @end
 
-@private partitionKeyCode(bundling)
-  {@context.partitionKeyCode(bundling.getDiscriminatorFields)}
+@private partitionKeyCode(bundlingDesc)
+  @join partitionKey : bundlingDesc.partitionKeys on " + "
```
michaelbausor: How about:
@join partitionKey : bundlingDesc.partitionKeys on " + {@partitionKey.separatorLiteral} + "

to fix the trailing "|"

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
michaelbausor: Agreed, happy to leave it as it is.

src/main/java/com/google/api/codegen/viewmodel/StaticXApiView.java
michaelbausor: OK yes that makes sense. When reading the name, I thought of static as in static methods. Do you think it would be an improvement to rename to StaticLanguageXApiView? (or StaticLangXApiView?) Or would this just be more verbose without additional clarity...

garrettjonesgoogle: Hmm I think it would help; it is indeed ambiguous here what "static" refers to. I'll rename them.

src/main/resources/com/google/api/codegen/java/settings.snip
michaelbausor: Hmmm... you could create a new function in the snippet that returned partitionKey.separatorLiteral with the plus signs on each end, and use that in the join statement. But now the whole thing is feeling convoluted, so I am happy to drop it.

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceNamer.java
garrettjonesgoogle: I will remove this pattern in all the places it appears.

src/main/java/com/google/api/codegen/transformer/java/JavaModelTypeNameConverter.java
garrettjonesgoogle: I'll do that, it makes the logic more consistent.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTransformer.java
garrettjonesgoogle: I'll just use File.separator for now.

michaelbausor:
LGTM

PR #338, opened 2016-08-01
Properly generate local import names for googleapis

pongad:
Go has different package structures than other languages.
This causes funny import names like

google_api "google.golang.org/genproto/googleapis/api/monitoredres"

because we directly copy the import path of the proto's go_package
option, but we derive the local name from the proto's package
declaration.
Something like this is reasonable in other languages, but not in Go.

This commit fixes the name by special-casing googleapis:
If an import is under google.golang.org/genproto,
we derive the local name from the import path instead.

This commit also removes import path guessing.
Since the package restructuring, it is essentially impossible to
guess the import path correctly.
Instead, we now require all imported proto packages to define
the go_package option.

### DIFF
--- a/src/main/java/com/google/api/codegen/go/GoGapicContext.java
+++ b/src/main/java/com/google/api/codegen/go/GoGapicContext.java
@@ -79,7 +79,7 @@ public class GoGapicContext extends GapicContext implements GoContext {
   /**
    * The import path for generated pb.go files for core-proto files.
    */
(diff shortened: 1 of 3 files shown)


pongad:
The new logic, while fixes the problem, is rather inelegant and clumsy. I'll write up a proposal for how we might be able to fix it for good later. My current idea relies on MVVM refactoring, so we can't really implement it immediately anyway.

saicheems:
+1

PR #340, opened 2016-08-01
Build API test generator with MVVM

shinfan:
- Built the first version of test generator with core MVVM components
- Baseline test


src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
```diff
@@ -0,0 +1,152 @@
...
+    typeTable.saveNicknameFor("java.io.IOException");
+    typeTable.saveNicknameFor("java.util.List");
+    typeTable.saveNicknameFor("com.google.api.gax.testing.ValueGenerator");
+    typeTable.saveNicknameFor("com.google.api.gax.testing.LocalServiceHelper");
+    typeTable.saveNicknameFor("com.google.protobuf.GeneratedMessage");
+    typeTable.saveNicknameFor("junit.framework.Assert");
+  }
+
+  private GapicSurfaceTestClassView createTestClassView(
+      Interface service, SurfaceTransformerContext context, ApiConfig apiConfig) {
```
garrettjonesgoogle: Don't pass service or apiConfig as separate parameters, they are already passed in the context

garrettjonesgoogle: (also the same in the other functions in this file)

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
```diff
@@ -0,0 +1,152 @@
...
+    }
+    return testCaseViews;
+  }
+
+  private GapicSurfaceTestCaseView createTestCaseView(
+      Method method, Interface service, SurfaceTransformerContext context, ApiConfig apiConfig) {
+    MethodConfig methodConfig = apiConfig.getInterfaceConfig(service).getMethodConfig(method);
+    MethodTransformerContext methodContext =
+        MethodTransformerContext.create(
+            service, apiConfig, context.getTypeTable(), context.getNamer(), method, methodConfig);
```
garrettjonesgoogle: I think it would be better to create the MethodTransformerContext in the calling function; simply use .asMethodContext() on SurfaceTransformerContext.

shinfan: Done

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
shinfan: Done

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
```diff
@@ -0,0 +1,154 @@
...
+
+  private GapicSurfaceTestCaseView createTestCaseView(
+      MethodTransformerContext methodContext) {
+    MethodConfig methodConfig = methodContext.getMethodConfig();
+    InitCodeTransformer initCodeTransformer = new InitCodeTransformer();
+    InitCodeView initCodeView =
+        initCodeTransformer.generateInitCode(methodContext, methodConfig.getRequiredFields());
+    List<GapicSurfaceTestAssertView> assertViews =
+        initCodeTransformer.generateTestAssertViews(
+            methodContext, methodConfig.getRequiredFields());
```
garrettjonesgoogle: I feel like it would be nice to not repeat the work of generating the init code (done by both generateInitCode and generateTestAssertViews). Don't make stuff hacky to accomplish it, though.

src/main/resources/com/google/api/codegen/java/test.snip
```diff
@@ -0,0 +1,87 @@
...
+  public void {@test.name}() {
+    {@initCode(test.initCode)}
+    {@methodCall(test)}
+
+    List<GeneratedMessage> requests = serviceHelper.getService().getRequests();
+    Assert.assertEquals(1, requests.size());
+    {@test.requestTypeName} request = ({@test.requestTypeName})requests[0];
+
+    @join assert : test.asserts
+      Assert.assertEquals(request.{@assert.actualValueIdentifier}(), {@assert.expectedValueIdentifier});
```
garrettjonesgoogle: actualValueIdentifer maybe should be renamed to indicate that it is a function name

src/main/resources/com/google/api/codegen/java/test.snip
```diff
@@ -0,0 +1,87 @@
...
+    @end
+  }
+@end
+
+@private methodCall(test)
+  @if {@test.isPageStreaming}
+    PageAccessor<{@test.resourceTypeName}> response = api.{@test.methodName}(\
```
garrettjonesgoogle: I feel like 'methodName' is not clear enough - it's ambiguous between the test method name and the method name being called. maybe call it 'surfaceMethodName'

src/test/java/com/google/api/codegen/testdata/java_test_test_library.baseline
```diff
@@ -0,0 +1,326 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.gcloud.pubsub.spi;
+
+import com.google.api.gax.testing.LocalServiceHelper;
+import com.google.api.gax.testing.ValueGenerator;
```
garrettjonesgoogle: Could you make a pull request with this class and any other edits to GAX?

src/test/java/com/google/api/codegen/testdata/java_test_test_library.baseline
```diff
@@ -0,0 +1,326 @@
...
+import org.junit.AfterClass;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+@javax.annotation.Generated("by GAPIC")
+public class LibraryServiceTest {
+  private static LocalServiceHelper serviceHelper;
+  private LibraryServiceApi api;
+  private ValueGenerator valueGenerator;
```
garrettjonesgoogle: When were you going to make use of valueGenerator?

src/test/java/com/google/api/codegen/testdata/java_test_test_library.baseline
```diff
@@ -0,0 +1,326 @@
...
+    List<GeneratedMessage> requests = serviceHelper.getService().getRequests();
+    Assert.assertEquals(1, requests.size());
+    GetBookFromArchiveRequest request = (GetBookFromArchiveRequest)requests[0];
+
+    Assert.assertEquals(request.getName(), formattedName);
+  }
+
+  @Test
+  @SuppressWarnings("all")
+  public void LibraryServiceTest() {
```
garrettjonesgoogle: All of the test cases are named the same thing... so I guess you haven't tried compiling and running this yet?

src/test/java/com/google/api/codegen/testdata/java_test_test_library.baseline
```diff
@@ -0,0 +1,326 @@
...
+    List<Comment> comments = Arrays.asList(commentsElement);
+    api.AddComments(formattedName, comments);
+
+    List<GeneratedMessage> requests = serviceHelper.getService().getRequests();
+    Assert.assertEquals(1, requests.size());
+    AddCommentsRequest request = (AddCommentsRequest)requests[0];
+
+    Assert.assertEquals(request.getName(), formattedName);
+    Assert.assertEquals(request.getComment(), comment);
+    Assert.assertEquals(request.getCommentsElement(), commentsElement);
```
garrettjonesgoogle: getComment() and getCommentsElement() are not valid members of request

src/test/java/com/google/api/codegen/testdata/java_test_test_library.baseline
```diff
@@ -0,0 +1,326 @@
...
+  }
+
+  @Test
+  @SuppressWarnings("all")
+  public void LibraryServiceTest() {
+
+    PageAccessor<String> response = api.ListStrings();
+    for (String element : response) {
+      // DO NOTHING
+    }
```
garrettjonesgoogle: I don't think you would want this here

garrettjonesgoogle:
I have finished my pass of the PR.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
shinfan: The repeat part is the InitCode which is the intermediate product produced by both method internally. This makes the reusing not very straightforward (it is not impossible but as you said, do not make stuff hacky :P). I am leaning towards to leave this as-is.

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -172,4 +176,16 @@ public String getOptionalArrayTypeName() {
...
+    return className(Name.upperCamel(service.getSimpleName(), "Settings"));
+  }
+
+  public String getTestClassName(Interface service) {
+    return className(Name.upperCamel(service.getSimpleName(), "Test"));
+  }
+
+  public String getMockServiceClassName(Interface service) {
+    return className(Name.upperCamel("Mock", service.getSimpleName()));
+  }
```
bjwatson: I'm a little concerned about an explosion of methods in this class that each are only used in one or two places in the code. Seems like this special case logic would be better defined in the file where it's used, unless it truly has broad applicability.

src/test/java/com/google/api/codegen/testdata/java_test_test_library.baseline
bjwatson: Let's not implement our own version of this unless we really have unique requirements. One of the QuickCheck implementations probably does what you need; there are four listed for Java here: https://en.wikipedia.org/wiki/QuickCheck

src/main/resources/com/google/api/codegen/java/test.snip
```diff
@@ -0,0 +1,87 @@
...
+      {@sampleMethodCallArgList(test.initCode.fieldSettings)});
+    for ({@test.resourceTypeName} element : response) {
+      // DO NOTHING
+    }
+  @else
+    api.{@test.methodName}(\
+      {@sampleMethodCallArgList(test.initCode.fieldSettings)});
+  @end
```
bjwatson: Do we really care about the response for this testing? Seems like we could just call every method the same way without assigning the return value to a variable, which might avoid the need for `@if` tests in this method.

garrettjonesgoogle: I think so - I think we want to check the plumbing.

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
garrettjonesgoogle: The issue is when you want to override one of these strings for a specific language; it would explode the complexity even more if this was spread over the code.

bjwatson: Oh, right. Are we actually going to override these for specific languages in a way that's not already handled by the NameFormatter?

garrettjonesgoogle: yes, examples: 
https://github.com/googleapis/toolkit/blob/master/src/main/java/com/google/api/codegen/transformer/php/PhpSurfaceNamer.java
https://github.com/garrettjonesgoogle/toolkit/blob/master/src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceNamer.java

bjwatson: Ok, I'm convinced. Nevertheless, some thought should be put into documenting these methods. This class already has 30+ methods, and is not necessarily obvious to developers which is used for what without searching for usage and reading through the dependent code to determine the use cases.

garrettjonesgoogle: In contrast with documenting the viewmodel classes (I'm dubious of the value of that), I can see more value in documenting these methods, particularly because they are a more likely point of exploration by a new implementer, and given that many concepts are in one place. I'll add comments in my Java MVVM PR ( https://github.com/googleapis/toolkit/pull/337 ).

bjwatson: SGTM

src/main/resources/com/google/api/codegen/java/test.snip
shinfan: It is a function name in Java but it may not be true for the other languages. I can rename it into actualValueGetter.

src/test/java/com/google/api/codegen/testdata/java_test_test_library.baseline
shinfan: I removed this variable for now since it is not blocking the basic functionality of this PR. QuickCheck looks like a good option.

src/test/java/com/google/api/codegen/testdata/java_test_test_library.baseline
shinfan: I am planing to separate the code into 3 parts: 
1. Test code generation <--- This is the current PR
2. Mock grpc service code generation
3. GAX

I cannot run the test until all 3 pieces are in place. Alternatively, I can work on a branch (maybe test-generator) of this repo instead of pushing into master.

src/test/java/com/google/api/codegen/testdata/java_test_test_library.baseline
shinfan: Will do

src/test/java/com/google/api/codegen/testdata/java_test_test_library.baseline
shinfan: Done.

src/test/java/com/google/api/codegen/testdata/java_test_test_library.baseline
shinfan: PageAccessor won't make any actual request until we iterate it so this loop is necessary. I have updated the comment to make this clear.

bjwatson: I think you have to do something with `element` to avoid JVM optimizations. See http://stackoverflow.com/questions/30116664/why-is-this-program-which-loops-many-times-taking-time-when-there-is-a-println

shinfan:
PR created for GAX related change: https://github.com/googleapis/gax-java/pull/92

src/main/resources/com/google/api/codegen/java/test.snip
shinfan: Done

src/test/java/com/google/api/codegen/testdata/java_test_test_library.baseline
shinfan: Done: googleapis/gax-java#92

src/main/java/com/google/api/codegen/gapic/MainGapicProviderFactory.java
```diff
@@ -139,6 +144,22 @@
 
       return Arrays.<GapicProvider<? extends Object>>asList(mainProvider, packageInfoProvider);
 
+    } else if (id.equals(JAVA_TEST)) {
+      GapicCodePathMapper javaPathMapper =
+          CommonGapicCodePathMapper.newBuilder()
+              .setPrefix("src/test/java")
```
michaelbausor: Nit: Maybe use a path builder of some sort, rather than hardcoding path separators?

garrettjonesgoogle: This is actually how all of the paths in this file are done...

src/test/java/com/google/api/codegen/testdata/java_test_test_library.baseline
```diff
@@ -111,11 +109,16 @@ public class LibraryServiceTest {
...
 
     PageAccessor<Shelf> response = api.ListShelves();
+
+    // Dumb value to avoid JVM optimization.
+    int requestCount = 0;
     for (Shelf element : response) {
-      // DO NOTHING
+      // PageAccessor will not make actual request until it is iterated.
+      // Iterate through the element here in order to make grpc requests.
+      requestCount += 1;
```
garrettjonesgoogle: I think I'd prefer something like:

List<Shelf> resources = new ArrayList<>();
resources.addAll(response.getPageValues());

src/main/java/com/google/api/codegen/gapic/MainGapicProviderFactory.java
michaelbausor: Fair enough, maybe I should create an issue for this instead

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -81,18 +82,26 @@ public InitCodeView generateRequestObjectInitCode(MethodTransformerContext conte
...
     Map<String, Object> initFieldStructure = createInitFieldStructure(context);
     InitCodeGenerator generator = new InitCodeGenerator();
+
     InitCode initCode =
         generator.generateRequestFieldInitCode(context.getMethod(), initFieldStructure, fields);
 
+    ArrayList<String> fieldNames = new ArrayList<>();
+    for (Field field : fields) {
+      fieldNames.add(Name.from(field.getSimpleName()).toLowerCamel());
+    }
```
garrettjonesgoogle: 1. Don't use casing functions in transformers
2. initCode.getArgFields() should give you what you need.

src/test/java/com/google/api/codegen/testdata/java_test_test_library.baseline
shinfan: Done

shinfan:
PTAL

src/main/java/com/google/api/codegen/gapic/MainGapicProviderFactory.java
garrettjonesgoogle: @shinfan could you file an issue?

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -423,4 +423,20 @@ public String getGetResourceListCallName(Field resourcesField) {
...
+
+  public String getGetFunctionCallName(Name name) {
```
garrettjonesgoogle: Add comments on each of your new functions here (following the pattern of the rest of the file as most recently merged)

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
```diff
@@ -0,0 +1,153 @@
...
+import com.google.api.tools.framework.model.Method;
+import com.google.api.tools.framework.model.Model;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/** A subclass of ModelToViewTransformer which translates model into API tests in Java. */
+public class JavaGapicSurfaceTestTransformer implements ModelToViewTransformer {
+
+  static String TEST_TEMPLATE_FILE = "java/test.snip";
```
garrettjonesgoogle: make private

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
```diff
@@ -0,0 +1,153 @@
...
+    GapicSurfaceTestClassView testClass =
+        GapicSurfaceTestClassView.newBuilder()
+            .packageName(context.getApiConfig().getPackageName())
+            .apiSettingsClassName(context.getNamer().getApiSettingsClassName(service))
+            .apiClassName(context.getNamer().getApiWrapperClassName(service))
+            .name(context.getNamer().getTestClassName(service))
+            .mockServiceClassName(context.getNamer().getMockServiceClassName(service))
+            .testCases(createTestCaseViews(context))
+            .outputPath(outputPath)
+            .imports(context.getTypeTable().getImports())
```
garrettjonesgoogle: Add a comment noting that the imports have to be after the other generation code is run because the type table is modified (see the comments to that effect in JavaGapicSurfaceTransformer)

src/main/resources/com/google/api/codegen/java/test.snip
```diff
@@ -0,0 +1,87 @@
...
+    {@test.requestTypeName} request = ({@test.requestTypeName})requests[0];
+
+    @join assert : test.asserts
+      Assert.assertEquals(request.{@assert.actualValueGetter}(), {@assert.expectedValueIdentifier});
+    @end
+  }
+@end
+
+@private methodCall(test)
+  @if {@test.isPageStreaming}
```
garrettjonesgoogle: I would prefer this to use a switch/case on a type string (which is done in other cases for page streaming vs not page streaming)

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    String requestTypeName =
+        methodContext
+            .getTypeTable()
+            .getAndSaveNicknameFor(methodContext.getMethod().getInputType());
+
+    return GapicSurfaceTestCaseView.newBuilder()
+        .name(methodContext.getNamer().getTestCaseName(methodContext.getMethod()))
+        .surfaceMethodName(methodContext.getMethod().getSimpleName())
```
garrettjonesgoogle: Never use string directly from the model - go through the surface namer. In particular, this has the wrong casing (it's upper camel and needs to be lower camel)

src/test/java/com/google/api/codegen/testdata/java_test_test_library.baseline
```diff
@@ -0,0 +1,327 @@
...
+    // PageAccessor will not make actual request until it is being used.
+    // Add all the pages here in order to make grpc requests.
+    List<String> resources = new ArrayList<>();
+    resources.addAll(response.getPageValues());
+
+    List<GeneratedMessage> requests = serviceHelper.getService().getRequests();
+    Assert.assertEquals(1, requests.size());
+    ListStringsRequest request = (ListStringsRequest)requests[0];
+
+  }
```
garrettjonesgoogle: You don't have any support here for checking responses. The tests should be able to inject a response, and have the API return it back so that it can be checked.

src/main/java/com/google/api/codegen/gapic/MainGapicProviderFactory.java
shinfan: Issue filed: https://github.com/googleapis/toolkit/issues/352

src/test/java/com/google/api/codegen/testdata/java_test_test_library.baseline
shinfan: Good thought. Done.

src/test/java/com/google/api/codegen/testdata/java_test_test_library.baseline
shinfan: Issue filed: https://github.com/googleapis/toolkit/issues/354

Will address in a separate PR

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
shinfan: Done

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
shinfan: Done

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
shinfan: Done

src/main/resources/com/google/api/codegen/java/test.snip
shinfan: Done

shinfan:
PTAL

garrettjonesgoogle:
LGTM

PR #341, opened 2016-08-03
Ignore streaming methods.

jmuk:
Currently Gapic generator does not support gRPC-streaming methods,
we will skip streaming methods for the time being (otherwise
the generated code will behavior incorrectly or lead to compile
errors).

Note that this skip can't be done solely on the configuration
yaml (i.e. not-listing methods in the yaml file), because
many of the generators silently assume the existence of
MethodConfig instances and end up with NullPointerExceptions.

Rather than supporting missing MethodConfig cases, I believe this
special-handling of gRPC-streaming would be better because
anyways this type of special-handling will be necessary when we start
supporting gRPC-streaming methods for a specific language without
affecting other languages.

### DIFF
--- a/src/main/java/com/google/api/codegen/ServiceMessages.java
+++ b/src/main/java/com/google/api/codegen/ServiceMessages.java
@@ -14,13 +14,12 @@
  */
 package com.google.api.codegen;
 
+import com.google.api.tools.framework.model.Interface;
 import com.google.api.tools.framework.model.Method;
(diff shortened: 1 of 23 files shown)


garrettjonesgoogle:
@pongad already did the same thing in https://github.com/googleapis/toolkit/pull/328 ...

jmuk:
Oh... good to know. Closing.

PR #344, opened 2016-08-04
Go: Properly use zeroValue

pongad:
StringIterator.Next() tries to return nil instead of "".
This commit makes it properly return "".

This change also applies to iterator over other non-ref types.


saicheems:
Lgtm

PR #351, opened 2016-08-04
Fix NodeJS codegen for gcloud-node integration.

jmuk:
This patch contains three significant changes:

- Previously we assume having a library for a gRPC API (like
  'grpc-google-logging-v1' package for logging gRPC), but
  that is not the case on gcloud-node. This patch changes
  the way to load grpc object (using 'google-proto-files'
  project) if it's going to be a part of gcloud-node.
- Update the file path: if it's going to be a part of
  gcloud-node, the file won't be directly under lib/ but
  lib/<api-name>.
- Remove gRPC dependency, and allow passing grpc module
  as a constructor argument.


jmuk:
gax change: https://github.com/googleapis/gax-nodejs/pull/30

src/test/java/com/google/api/codegen/NodeJSCodeGeneratorTest.java
```diff
@@ -26,31 +26,65 @@
...
+    @Test
+    public void library() throws Exception {
+      test("library");
+    }
   }
 
-  // Tests
-  // =====
+  @RunWith(Parameterized.class)
+  public static class NodeJSNoPathTGemplatesBaseline extends GapicTestBase {
```
geigerj: s/TGemplates/Templates

src/test/java/com/google/api/codegen/testdata/nodejs_main_no_path_templates.baseline
```diff
@@ -0,0 +1,165 @@
...
+ * @param {String} opts.servicePath
+ *   The domain name of the API remote host.
+ * @param {number} opts.port
+ *   The port on which to connect to the remote host.
+ * @param {Function} opts.getCredentials
+ *   Custom function to obtain the credentials.
+ * @param {grpc.ClientCredentials} opts.sslCreds
+ *   A ClientCredentials for use with an SSL-enabled channel.
+ * @param {Object} opts.grpc
+ *   When specified, this is used as the grpc module.
```
geigerj: Could you clarify this description to state that that if this is not specified, then the `grpc` module is loaded automatically?

src/test/java/com/google/api/codegen/testdata/nodejs_main_library.baseline
```diff
@@ -102,6 +88,8 @@ var ALL_SCOPES = [
  *   Custom function to obtain the credentials.
  * @param {grpc.ClientCredentials} opts.sslCreds
  *   A ClientCredentials for use with an SSL-enabled channel.
+ * @param {Object} opts.grpc
```
geigerj: So, once this is added to gcloud-node, if I'm a gcloud-node user, I have to set this parameter in order to make the library work?

src/test/java/com/google/api/codegen/NodeJSCodeGeneratorTest.java
jmuk: Done

src/test/java/com/google/api/codegen/testdata/nodejs_main_no_path_templates.baseline
jmuk: Done

src/test/java/com/google/api/codegen/testdata/nodejs_main_library.baseline
jmuk: No.

The consistency is the key -- the exactly same grpc module should be used among the API. Previous design was broken because this generated file loads grpc, and gax also loads grpc as well -- they happen to be same in a normal installation, but can be broken.

Now,
- if opts.grpc is not specified, GAX loads its own grpc module and that is used everywhere
- if opts.grpc is specified, GAX stops loading its own grpc and using specified one everywhere

So either way it should work (to make this work, we need to modify GAX as well, see https://github.com/googleapis/gax-nodejs/pull/30).

geigerj: I meant, when and end-user creates an API object, they need to set this parameter to prevent GAX from loading its own grpc module, right? That seems unexpected and potentially like a usability issue; what if the user forgets to set the parameter? Is there some way that we can auto-detect the "right" module?

jmuk: They don't have to prevent. GAX will loads its own gRPC module, but it will work as expected, API call will succeed.

Problem will happen when it's combined with _other_ API instances -- their gRPC module has to be same. In that case this `opt.grpc` parameter will be helpful.

jmuk:
BTW I uploaded a patch, which modifies the code path strategy a bit.  It always generate under `src`, and that will be combined with https://github.com/googleapis/googleapis/pull/72 and https://github.com/googleapis/packman/pull/80 (for non-gcloud APIs).

jmuk:
rebased

src/test/java/com/google/api/codegen/testdata/nodejs_main_library.baseline
geigerj: Note from offline discussion: this parameter only needs to be used to avoid type mismatches if the user wants to pass in custom gRPC objects (e.g., the channel or credentials), since there will be multiple gRPC modules (corresponding to different types) if this option is not enabled. It shouldn't need to be used in the simple case.

src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
```diff
@@ -73,6 +73,11 @@ public String grpcClientName(Interface service) {
     return "grpc-" + service.getFile().getFullName().replace('.', '-');
   }
 
+  public boolean isGcloud() {
+    String packageName = getApiConfig().getPackageName();
+    return !Strings.isNullOrEmpty(packageName) && packageName.indexOf("@google-cloud/") == 0;
```
geigerj: nit: `startsWith` instead of `indexOf`?

geigerj:
One nit, LGTM

src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
jmuk: done

PR #353, opened 2016-08-05
Ruby style fixes

jmuk:
Those styles are from gcloud-ruby:
- prefers "" rather than ''
- do not use () for def
- use : instead of => for simbol-keyed Hashes
- add rubocop suppression for line length and method length:

### DIFF
--- a/src/test/java/com/google/api/codegen/testdata/ruby_doc_main_library.baseline
+++ b/src/test/java/com/google/api/codegen/testdata/ruby_doc_main_library.baseline
@@ -23,11 +23,15 @@
 # The only allowed edits are to method and file documentation. A 3-way
 # merge preserves those additions if the generated source changes.
 
-require 'json'
-require 'pathname'
+require "json"
+require "pathname"
 
-require 'google/gax'
-require 'library_services'
+require "google/gax"
+require "library_services"
+
+# TODO: implement line-wrapping rules and remove this suppression.
+# rubocop:disable LineLength
+# rubocop:disable MethodLength
 
 module Library
   module V1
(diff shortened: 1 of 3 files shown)


src/test/java/com/google/api/codegen/testdata/ruby_doc_main_library.baseline
```diff
@@ -50,133 +54,133 @@ module Library
...
+          "next_page_token",
+          "strings")
       }.freeze
 
       private_constant :PAGE_DESCRIPTORS
 
       BUNDLE_DESCRIPTORS = {
-        'publish_series' => Google::Gax::BundleDescriptor.new(
-          'books',
+        "publish_series" => Google::Gax::BundleDescriptor.new(
```
geigerj: `=>` here (and for `PAGE_DESCRIPTORS`)

jmuk: Actually this is correct. This is confusing, but in Ruby, `{"foo": "bar"}` and `{"foo" => "bar"}` are different. 
- `{"foo" => "bar"}`: normal one, the key is as is seen (in this example, a String instance `"foo"`)
- `{"foo": "bar"}`: special syntax, the key is converted into a Symbol (in this example, `:foo`)

For page descriptors and bundle descriptors, we want to have string-keyed Hash, so this still needs to be `=>`.

geigerj: Ah, I see. Thanks.

geigerj:
LGTM

PR #355, opened 2016-08-05
Modify Ruby styles a bit again.

jmuk:
The Ruby's new hash syntax ({foo: "bar"}) is recommended, but
the quoted version ({"foo-bar": "baz"}) is not yet available
until Ruby 2.2.0. Therefore I've reverted the change (and
added a comment to suppress rubocop, Ruby's syntax checker).


geigerj:
LGTM if build passes

Landrito:
LGTM

PR #361, opened 2016-08-09
Adding missing arg for PathTemplate.validate

garrettjonesgoogle:


src/main/java/com/google/api/codegen/viewmodel/PathTemplateCheckView.java
```diff
@@ -23,6 +23,8 @@
 
   public abstract String paramName();
 
+  public abstract String methodName();
```
michaelbausor: Should this be called something that indicates its function, rather than its content? validationMessagePrefix, or similar?

garrettjonesgoogle: Good idea. I think I'll go with "validationMessageContext".

garrettjonesgoogle:
Updated, PTAL

michaelbausor:
LGTM once Travis completes

PR #362, opened 2016-08-09
Rename Go Imports and Properly Wrap API Doc

pongad:
This commit further simplifies Go import names.
Prior to this, we take the import path such as "googleapis/library/v1"
and replace the slashes with underscore.

This commit renames the primary proto package to just "pb".
Other proto packages are named in the format of "foopb".

Top-level package documents were not line-wrapped.
This commit also line wraps them.


src/main/java/com/google/api/codegen/go/GoGapicContext.java
```diff
@@ -167,12 +169,27 @@ public String getServiceClientConstructorName(Interface service) {
...
+    List<String> parts = Arrays.asList(goPackage.split("/"));
+    Collections.reverse(parts);
+    for (String part : parts) {
+      if (part.length() < 2 || part.charAt(0) != 'v' || !Character.isDigit(part.charAt(1))) {
+        name = part;
+        break;
+      }
+    }
+    if (name == null) {
+      name = parts.get(0);
```
saicheems: Could set this up there instead of doing the null check.

src/test/java/com/google/api/codegen/testdata/go_example_library.baseline
```diff
@@ -20,7 +20,7 @@ package library_test
 import (
     "cloud.google.com/go/library/apiv1"
     "golang.org/x/net/context"
-    googleapis_example_library_v1 "google.golang.org/genproto/googleapis/example/library/v1"
+    pb "google.golang.org/genproto/googleapis/example/library/v1"
```
saicheems: Was the `pb` name a suggestion by the Go team?

I'd generally prefer `library` or `library_v1`... I know there's a tendency for very short names, but I feel `pb` has lost most of the context behind the import. My first thought was the `google/protobuf` package.

saicheems:
Btw I send github emails to my personal email (intentionally)

Just so you know I'm not stalking your pulls on weekends and nights lol

src/main/java/com/google/api/codegen/go/GoGapicContext.java
pongad: Good point! Done.

src/main/java/com/google/api/codegen/go/GoGapicContext.java
```diff
@@ -167,12 +169,24 @@ public String getServiceClientConstructorName(Interface service) {
...
-    if (goPackage.startsWith(CORE_PROTO_BASE)) {
-      return goPackage.substring(CORE_PROTO_BASE.length()).replace("/", "_");
+    if (!goPackage.startsWith(CORE_PROTO_BASE)) {
+      return proto.getFile().getProto().getPackage().replace(".", "_");
     }
-    return proto.getFile().getProto().getPackage().replace(".", "_");
+    List<String> parts = Arrays.asList(goPackage.split("/"));
+    Collections.reverse(parts);
+    String name = parts.get(0);
+    for (String part : parts) {
```
saicheems: Some brief commentary on what this logic does would be helpful for future eyes. It's not immediately obvious what input corresponds to what output on inspection.

src/test/java/com/google/api/codegen/testdata/go_example_library.baseline
saicheems: Or maybe in this case `librarypb`

pongad: The `pb` name is a suggestion from @okdave. I think the reasoning is along the line of "we are writing client for pubsub, so pb is just protobuf of pubsub". I think it's the context of the code that makes it work.

okdave: `librarypb` would be okay, be it's common practice to make proto imports short, and they should always have the `pb` suffix.

src/main/resources/com/google/api/codegen/go/common.snip
```diff
@@ -23,12 +23,12 @@
         gapicNameVersion = "gapic/0.1.0"
     )
 
-    var (
-        @if {context.hasPageStreamingMethod()}
+    @if {context.hasPageStreamingMethod()}
+        var (
```
okdave: Since there's only ever a single variable here, maybe just

var Done = errors.New(etc.)

pongad: Done.

src/main/java/com/google/api/codegen/go/GoGapicContext.java
```diff
@@ -167,12 +169,36 @@ public String getServiceClientConstructorName(Interface service) {
...
+    // Inside CORE_PROTO_BASE, we are referencing code generated by curated protos.
+    // Import path of these protos might be versioned
+    //   google.golang.org/genproto/googleapis/example/library/v1
+    // or not
+    //   google.golang.org/genproto/googleapis/api/monitoredres
+    // We heuristically get the import name by looking for the right-most element,
+    // that is not a version number.
+    List<String> parts = Arrays.asList(goPackage.split("/"));
+    Collections.reverse(parts);
+    String name = parts.get(0);
```
jmuk: This section will end up with a weird result when full name consists of all v[0-9]+ patterns (like "v1/v2/b3"). Or if it's "golang.org/v1/v2", `name` becomes `golang.org` which will be invalid.  Such things would never happen in the realistic situation though.

Considering that, however, I think this line would be unnecessary; simply saying `String name = null;` is better, so it can fail with errors in case we encounter such erroneous situation.

src/main/java/com/google/api/codegen/go/GoGapicContext.java
```diff
@@ -167,12 +169,36 @@ public String getServiceClientConstructorName(Interface service) {
...
+    for (String part : parts) {
+      if (part.length() < 2 || part.charAt(0) != 'v' || !Character.isDigit(part.charAt(1))) {
```
jmuk: Should we care about the format like "v1beta2"?

src/main/java/com/google/api/codegen/go/GoGapicContext.java
pongad: This is a good point. Changed to `null` and added explicit check for it.

src/main/java/com/google/api/codegen/go/GoGapicContext.java
pongad: We already are right? `v1beta2` has length greater than or equal to 2, starts with "v", and its second character "1" is a digit. It fails all 3 tests.

saicheems:
lgtm

src/main/java/com/google/api/codegen/go/GoContextCommon.java
```diff
@@ -34,4 +35,34 @@
     }
     return result;
   }
+
+  public Iterable<String> getCommentLinesWrap(String text) {
```
saicheems: Add javadoc?

src/main/java/com/google/api/codegen/go/GoContextCommon.java
```diff
@@ -34,4 +35,34 @@
...
+    List<String> result = new ArrayList<>();
+    for (String line : Splitter.on(String.format("%n")).split(text)) {
+      for (String wrapped : wrapLine(line, 70)) {
+        result.add(line.isEmpty() ? "//" : "// " + wrapped);
+      }
+    }
+    return result;
+  }
+
+  private List<String> wrapLine(String line, int length) {
```
vchudnov-g: javadoc?

src/main/java/com/google/api/codegen/go/GoContextCommon.java
```diff
@@ -34,4 +35,34 @@
...
+    if (line.length() <= length) {
+      return Collections.<String>singletonList(line);
+    }
+    List<String> result = new ArrayList<>(line.length() / length + 1);
+    StringBuilder current = new StringBuilder();
+    for (String word : Splitter.onPattern("\\s").omitEmptyStrings().split(line)) {
+      current.append(" ");
+      current.append(word);
+      if (current.length() >= length) {
```
vchudnov-g: instead of checking that the new word begins within length, you could check the current.length() + word.length() <= length. It might make things more intuitive.

src/main/java/com/google/api/codegen/go/GoGapicContext.java
```diff
@@ -167,12 +169,39 @@ public String getServiceClientConstructorName(Interface service) {
...
+    if (!goPackage.startsWith(CORE_PROTO_BASE)) {
+      return proto.getFile().getProto().getPackage().replace(".", "_");
     }
-    return proto.getFile().getProto().getPackage().replace(".", "_");
+    // Inside CORE_PROTO_BASE, we are referencing code generated by curated protos.
+    // Import path of these protos might be versioned
+    //   google.golang.org/genproto/googleapis/example/library/v1
+    // or not
+    //   google.golang.org/genproto/googleapis/api/monitoredres
+    // We heuristically get the import name by looking for the right-most element,
```
vchudnov-g: nit: remove comma.

vchudnov-g:
I would suggest tests for the new functions.

Aside from that and the javadoc, lgtm.

src/main/java/com/google/api/codegen/go/GoContextCommon.java
pongad: Done.

src/main/java/com/google/api/codegen/go/GoContextCommon.java
pongad: Done

src/main/java/com/google/api/codegen/go/GoGapicContext.java
pongad: Done. I'm constantly surprised by your eagle-like eyes for punctuation.

broady:
Importing as `pb` isn't particularly idiomatic, but it's not horrible.

I'd certainly prefer `librarypb` to `pb`.

/cc @okdave

src/main/java/com/google/api/codegen/go/GoContextCommon.java
pongad: That would make things more intuitive, but it introduces the problem of having words that are by themselves longer than the limit. For this reason, I chose instead to go for a simpler algorithm that splits *after* the "limit" has been reached. We can solve both problems by using a more sophisticated algorithm, but I think this (at least for the time being) has more costs than benefits.

I have documented this in the Javadoc. Let me know if this works for you.

okdave:
So, internal-to-Google we `pb` by itself is very idiomatic but I don't see that so much outside. (Probably because, to date, not that many people expose protos in their package APIs). From code I've reviewed, `pb` and `lpb` would be common aliases in practice.

My thoughts is that making it as-short-as-possible will make (the godoc for the API)[https://godoc.org/cloud.google.com/go/logging/apiv2#Client.ListLogEntries] more legible. Also it starts the trend of suffixing imports with `pb`.

I can live with `librarypb` if there's a strong consensus that `pb` by itself is too naked. Either is miles better than`googleapis_logging_v2`.

pongad:
@saicheems @okdave @broady `librarypb` it is then!

okdave:
Yup, let's do that. There's nothing stopping us revising these later, as users code doesn't depend on the aliases except as guiding examples.

broady:
Great, either way – this is a huge improvement! Thank you!

pongad:
Despite the LGTM earlier, the last commit is rather big so I wanted to let everyone review again. Hopefully we'll converge on this soon. PTAL.

broady:
LGTM, but I only looked at the output in the first place. No comment on the Java code.

pongad:
@broady Thank you for the comments. I'll wait for another lgtm by a toolkit team member.

src/main/java/com/google/api/codegen/go/GoContextCommon.java
```diff
@@ -46,6 +51,18 @@
     return result;
   }
 
+  /**
+   * wrapLine splits `line` into a list of lines around `length` long at whitespaces.
+   *
+   * Current implementation first splits the line into words, then add the words,
```
saicheems: adds the words

saicheems: might just make this sentence into a delimited list, it would read easier

src/main/java/com/google/api/codegen/go/GoContextCommon.java
```diff
@@ -46,6 +51,18 @@
...
+   * separated by a single space, to the current line until the line is longer than `length`,
+   * then breaks the line.
+   * This eliminates the corner-case of `line` containing a word that is by itself longer than
+   * length. This issue crops up occasionally since many comments contain URLs.
+   * Consequently, each wrapped line, except from possibly the last, will be slightly longer
```
saicheems: except possibly the last

src/test/java/com/google/api/codegen/go/GoContextCommonTest.java
```diff
@@ -0,0 +1,54 @@
...
+  @Test
+  public void testLineWrapLong() {
+    List<String> words = Arrays.asList(stringRep("a", 50), stringRep("a", 20), stringRep("a", 3));
+    String in = Joiner.on(" ").join(words);
+    Truth.assertThat(ctx.getCommentLinesWrap(in))
+        .containsExactly("// " + words.get(0) + " " + words.get(1), "// " + words.get(2))
+        .inOrder();
+  }
+
+  private String stringRep(String s, int n) {
```
saicheems: Maybe document this? I know it's pretty self explanatory

saicheems:
re-lgtm

src/main/java/com/google/api/codegen/go/GoContextCommon.java
```diff
@@ -34,4 +35,51 @@
...
+  }
+
+  /**
+   * wrapLine splits `line` into a list of lines around `length` long at whitespaces.
+   *
+   * Current implementation first splits the line into words, then add the words,
+   * separated by a single space, to the current line until the line is longer than `length`,
+   * then breaks the line.
+   * This eliminates the corner-case of `line` containing a word that is by itself longer than
+   * length. This issue crops up occasionally since many comments contain URLs.
```
saicheems: might've wanted that length in backticks

src/main/java/com/google/api/codegen/go/GoContextCommon.java
```diff
@@ -34,4 +35,51 @@
...
+    return result;
+  }
+
+  /**
+   * wrapLine splits `line` into a list of lines around `length` long at whitespaces.
+   *
+   * Current implementation first splits the line into words, then add the words,
+   * separated by a single space, to the current line until the line is longer than `length`,
+   * then breaks the line.
+   * This eliminates the corner-case of `line` containing a word that is by itself longer than
```
vchudnov-g: "corner case" (no hyphen)

I don't think a very long word in a line by itself is the problem, nor does your implementation guard against that, IIUC (if you were starting a new line anyway and the next word is super-long, then it will wind up as the only word on the line). What it does guard against is a very *short* line followed by a line longer than `length` (or more accurately, a line of length j followed by a line of length k where j+k>=`length`, which is unattractive when j << k)

src/main/java/com/google/api/codegen/go/GoContextCommon.java
```diff
@@ -34,4 +35,51 @@
...
+        result.add(line.isEmpty() ? "//" : "// " + wrapped);
+      }
+    }
+    return result;
+  }
+
+  /**
+   * wrapLine splits `line` into a list of lines around `length` long at whitespaces.
+   *
+   * Current implementation first splits the line into words, then add the words,
```
vchudnov-g: s/Current/The current/
s/add/adds/

src/main/java/com/google/api/codegen/go/GoContextCommon.java
```diff
@@ -34,4 +35,51 @@
...
+   * separated by a single space, to the current line until the line is longer than `length`,
+   * then breaks the line.
```
vchudnov-g: s/then/and then/

src/main/java/com/google/api/codegen/go/GoContextCommon.java
vchudnov-g: See my comment above about what I think is really happening. (Update the comment if what I said is true)

I'm fine with this implementation since you've documented it.

src/main/java/com/google/api/codegen/go/GoGapicContext.java
```diff
@@ -167,12 +169,31 @@ public String getServiceClientConstructorName(Interface service) {
...
+  private String localPackageName(ProtoElement proto) {
     String goPackage = proto.getFile().getProto().getOptions().getGoPackage();
-    if (goPackage.startsWith(CORE_PROTO_BASE)) {
-      return goPackage.substring(CORE_PROTO_BASE.length()).replace("/", "_");
+    if (!goPackage.startsWith(CORE_PROTO_BASE)) {
+      return proto.getFile().getProto().getPackage().replace(".", "_");
     }
-    return proto.getFile().getProto().getPackage().replace(".", "_");
+    // Inside CORE_PROTO_BASE, we are referencing code generated by curated protos.
+    // Import path of these protos might be versioned
```
vchudnov-g: s/Import/The import/
s/versioned:/

src/main/java/com/google/api/codegen/go/GoGapicContext.java
```diff
@@ -167,12 +169,31 @@ public String getServiceClientConstructorName(Interface service) {
...
+    //   google.golang.org/genproto/googleapis/example/library/v1
+    // or not
```
vchudnov-g: s/not/not:/

src/main/java/com/google/api/codegen/go/GoContextCommon.java
```diff
@@ -34,4 +35,51 @@
     }
     return result;
   }
+
+  /**
+   * Converts and wraps the specified text into a comment block in the generated Go file.
+   *
+   * Lines are wrapped to 70 characters. See documentation of wrapLine for more details.
+   */
+  public Iterable<String> getCommentLinesWrap(String text) {
```
vchudnov-g: How about calling this "getCommentLinesWrapped" or "getWrappedCommentLines"? I think these (especially the latter) reads better.

src/test/java/com/google/api/codegen/go/GoContextCommonTest.java
```diff
@@ -0,0 +1,54 @@
...
+  private static GoContextCommon ctx = new GoContextCommon();
+
+  @Test
+  public void testLineWrapShort() {
+    Truth.assertThat(ctx.getCommentLinesWrap("aaa")).containsExactly("// aaa");
+  }
+
+  @Test
+  public void testLineWrapLong() {
+    List<String> words = Arrays.asList(stringRep("a", 50), stringRep("a", 20), stringRep("a", 3));
```
vchudnov-g: nit: alternate between "a", "b", and "c" to make it more even more obvious to humans these are distinct words. I know it doesn't matter to the correctness of the test.

vchudnov-g:
lgtm

src/main/java/com/google/api/codegen/go/GoContextCommon.java
pongad: Done.

src/test/java/com/google/api/codegen/go/GoContextCommonTest.java
pongad: Done.

src/main/java/com/google/api/codegen/go/GoContextCommon.java
pongad: Done

src/main/java/com/google/api/codegen/go/GoContextCommon.java
pongad: Done

src/main/java/com/google/api/codegen/go/GoContextCommon.java
pongad: Done

src/main/java/com/google/api/codegen/go/GoGapicContext.java
pongad: Done

src/main/java/com/google/api/codegen/go/GoGapicContext.java
pongad: Done

src/main/java/com/google/api/codegen/go/GoContextCommon.java
pongad: Done

src/test/java/com/google/api/codegen/go/GoContextCommonTest.java
pongad: Done.

src/main/java/com/google/api/codegen/go/GoContextCommon.java
pongad: Done

pongad:
@vchudnov-g Your analysis about the algorithm is correct. I have changed the wording of the documentation. Your suggestion has a similar corner case if the long word comes after the line split. I think that greedy algorithm like this won't handle cases like long words very well. I have sent an email to the team about an alternate algorithm that I think handles these cases beautifully.

Since you have given an LGTM, I assume that you are OK, if not completely satisfied, with the current algorithm. I'll merge this PR for now. Whether we decide to use the new algorithm or fix the existing one, we can do that in follow up PRs. I didn't think line breaking would cause as much churn as it did.

vchudnov-g:
@pongad Sounds good. Thanks!

PR #364, opened 2016-08-09
Use "cloud" in Python package name by default

geigerj:
Also,
- change regular expressions in the package naming logic so that
  "cloud" will only be added if it is not already present in the
  proto package name
- Modify no_path_templates test to serve as a "cloud" API and add
  as a test for the config generator


garrettjonesgoogle:
LGTM

PR #367, opened 2016-08-10
Python 3 compatibility fix

geigerj:
In Python 3, the json library distinguishes between string and bytes
types. Ensure that we are passing in string to the json loading
function.

This addresses https://github.com/GoogleCloudPlatform/gcloud-python/issues/1944


bjwatson:
LGTM

PR #369, opened 2016-08-11
Java generation: not importing map entry type

garrettjonesgoogle:


src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceNamer.java
```diff
@@ -41,6 +42,15 @@ public JavaSurfaceNamer() {
   }
 
   @Override
+  public boolean shouldImportRequestObjectParamElementType(Field field) {
+    if (Field.IS_MAP.apply(field)) {
```
shinfan: Does this check only apply to Java? If not we can move this up to the SurfaceNamer.

garrettjonesgoogle: Basically, I don't know. I'm thinking that if it turns out that PHP is the exception, we can move it up later.

shinfan:
LGTM assuming the conflict will be resolved.

PR #370, opened 2016-08-11
Go: Properly cast to int32

pongad:
This change was missed from #368.
After size-checking the pageSize argument,
we should actually cast the type to int32.


pongad:
@okdave

okdave:
LGTM

PR #374, opened 2016-08-15
Incremental dependency bump

garrettjonesgoogle:
This is only an incremental bump to limit the scope of changes.

* protobuf-gradle-plugin: 0.7.2-SNAPSHOT to 0.7.3 (getting off of a
  snapshot dependency)
* protobuf/protoc: 3.0.0-beta-2 to 3.0.0-beta-3 (consistent with the
  version we are using in the generated code stack)

Also, renaming "private" snapshot repo references to "toolkit". This
means the username and password entries in each dev's gradle.properties
file need to be updated, if they want to use 'uploadArchives'. The
basic problem is that multiple repos might use the same
keys (e.g. api-compiler), and thus would clash with each other.

* privateOssrhUsername to privateToolkitUsername
* privateOssrhPassword to privateToolkitPassword

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,18 +7,13 @@ buildscript {
   repositories {
     mavenLocal()
     maven {
-      // For 0.7.2-SNAPSHOT of protobuf gradle plugin. This should be removed once
-      // this version is releaed.
-      url 'https://oss.sonatype.org/content/repositories/snapshots/'
-    }
-    maven {
       url 'https://plugins.gradle.org/m2/'
     }
     mavenCentral()
     jcenter()
   }
   dependencies {
-    classpath "com.google.protobuf:protobuf-gradle-plugin:0.7.2-SNAPSHOT"
+    classpath "com.google.protobuf:protobuf-gradle-plugin:0.7.3"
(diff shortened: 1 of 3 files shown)


geigerj:
> The basic problem is that multiple repos might use the same keys (e.g. api-compiler), and thus would clash with each other.

I'm not familiar with the uploadArchives task. How does this change introduce this problem?

garrettjonesgoogle:
>> The basic problem is that multiple repos might use the same keys (e.g. api-compiler), and thus would clash with each other.

>I'm not familiar with the uploadArchives task. How does this change introduce this problem?

This change fixes this problem :-P The uploadArchives task uploads the compiled jars to a maven repository so that they can be downloaded by consuming code. I'm not sure who is even using this at all, but I know we'll need it in the future (and we'll want to upload to Sonatype/Maven Central later instead of our own snapshot repository). 

The username/password are looked up in the gradle.properties file located in the user's home directory; if those values are for a different account (e.g. for api-compiler's private repo instead of toolkit's private repo), then the upload will fail.

geigerj:
LGTM

PR #375, opened 2016-08-15
Add response check for Java automated testing.

shinfan:
GAX PR: https://github.com/googleapis/gax-java/pull/98

### DIFF
--- a/src/test/java/com/google/api/codegen/testdata/java_test_mock_service_impl_library.baseline
+++ b/src/test/java/com/google/api/codegen/testdata/java_test_mock_service_impl_library.baseline
@@ -15,6 +15,7 @@
 
 package com.google.gcloud.pubsub.spi;
 
+import com.google.common.collect.Lists;
 import com.google.example.library.v1.AddCommentsRequest;
 import com.google.example.library.v1.Book;
 import com.google.example.library.v1.CreateBookRequest;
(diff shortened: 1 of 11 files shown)


src/test/java/com/google/api/codegen/testdata/java_test_test_library.baseline
```diff
@@ -84,8 +89,16 @@ public class LibraryServiceTest {
   @Test
   @SuppressWarnings("all")
   public void createShelfTest() {
+    Shelf response = Shelf.newBuilder()
```
garrettjonesgoogle: all the variables for the expected response should have "expected" in the name, otherwise it's confusing what they are.

src/test/java/com/google/api/codegen/testdata/java_test_test_library.baseline
```diff
@@ -199,12 +267,19 @@ public class LibraryServiceTest {
   @Test
   @SuppressWarnings("all")
   public void listBooksTest() {
+    ListBooksResponse response = ListBooksResponse.newBuilder()
+      .build();
+    List<GeneratedMessage> responses = new ArrayList<>();
```
garrettjonesgoogle: There is an important way that the sample code init fields and the tests need to differ: sample code init fields should initialize as little as possible; tests should initialize as much as possible (to make sure it gets plumbed through). This test doesn't really verify much, because an empty list can't be distinguished between different API method calls.

src/test/java/com/google/api/codegen/testdata/java_test_mock_service_impl_library.baseline
```diff
@@ -46,154 +47,229 @@ import java.util.List;
...
+    responses = new ArrayList<>();
+  }
+
+  public List<GeneratedMessage> getRequests() {
+    return requests;
+  }
+
+  public void setResponses(List<GeneratedMessage> responses) {
+    this.responses = Lists.newArrayList(responses);
+  }
```
garrettjonesgoogle: Why is responses a list when you only ever use the first element?

src/test/java/com/google/api/codegen/testdata/java_test_mock_service_impl_library.baseline
```diff
@@ -46,154 +47,229 @@ import java.util.List;
...
+    responseObserver.onNext(response);
+    responseObserver.onCompleted();
+  }
+
+  @Override
+  public void getShelf(GetShelfRequest request,
+      StreamObserver<Shelf> responseObserver) {
+    Shelf response = Shelf.newBuilder().build();
+    if (responses != null && !responses.isEmpty()) {
+      response = (Shelf) responses.get(0);
```
michaelbausor: Is the reason to have a list of responses stored in the mock service (instead of just a single response) to allow different responses to be returned in sequence? If so, should this pop the response off the list?

src/test/java/com/google/api/codegen/testdata/java_test_mock_service_impl_library.baseline
shinfan: I would expect to use more responses for the page-streaming case after the response field populating work is done.  Updated the issue: https://github.com/googleapis/toolkit/issues/354

src/test/java/com/google/api/codegen/testdata/java_test_mock_service_impl_library.baseline
shinfan: Right. response removed from the list.

src/test/java/com/google/api/codegen/testdata/java_test_test_library.baseline
shinfan: As discussed offline, I would fix this in a separate PR.
Issue to track: https://github.com/googleapis/toolkit/issues/354

src/test/java/com/google/api/codegen/testdata/java_test_mock_service_impl_library.baseline
```diff
@@ -46,154 +47,229 @@ import java.util.List;
...
+    responses = new ArrayList<>();
+  }
+
+  @Override
+  public void createShelf(CreateShelfRequest request,
+      StreamObserver<Shelf> responseObserver) {
+    Shelf response = Shelf.newBuilder().build();
+    if (responses != null && !responses.isEmpty()) {
+      response = (Shelf) responses.get(0);
+    }
```
shinfan: If no response is injected, the default response object will be used.

src/test/java/com/google/api/codegen/testdata/java_test_test_library.baseline
shinfan: Done

src/test/java/com/google/api/codegen/testdata/java_test_mock_service_impl_library.baseline
garrettjonesgoogle: What I mean is, do we want to allow a default response object at all? It might be better to always require an injected response.

shinfan: Done.
After second thought I think you are right it is better to always require injected response since default response does not provide much testing value.

garrettjonesgoogle:
LGTM

garrettjonesgoogle:
That is, after the Travis test is fixed

src/test/java/com/google/api/codegen/testdata/java_test_mock_service_impl_library.baseline
```diff
@@ -46,154 +47,280 @@ import java.util.List;
...
+    requests = new ArrayList<>();
+    responses = new ArrayList<>();
+  }
+
+  @Override
+  public void createShelf(CreateShelfRequest request,
+      StreamObserver<Shelf> responseObserver) {
+    if (responses != null && !responses.isEmpty()) {
+      Shelf response = (Shelf) responses.get(0);
+      responses.remove(0);
```
bjwatson: You can do this in a single shot with the `remove()` call: https://docs.oracle.com/javase/7/docs/api/java/util/ArrayList.html#remove(int)

src/test/java/com/google/api/codegen/testdata/java_test_mock_service_impl_library.baseline
```diff
@@ -46,154 +47,280 @@ import java.util.List;
...
-
-   @Override
-   public void streamShelves(ListShelvesRequest request,
-       StreamObserver<ListShelvesResponse> responseObserver) {
-     requests.add(request);
-     responseObserver.onNext(ListShelvesResponse.newBuilder().build());
-     responseObserver.onCompleted();
-   }
+  private ArrayList<GeneratedMessage> requests;
+  private ArrayList<GeneratedMessage> responses;
```
bjwatson: Nit: A LinkedList might be more efficient, since you only seem to ever pull the head off `responses`, and push a new head onto `requests`. The pattern is quite stack-like.

garrettjonesgoogle: Isn't it more queue-like?

bjwatson: Dur, you're right, and LinkedList implements that interface: https://docs.oracle.com/javase/7/docs/api/java/util/LinkedList.html

shinfan: I have fixed this in https://github.com/googleapis/toolkit/pull/386

bjwatson: Thanks. I just reviewed it.

PR #376, opened 2016-08-15
Update discovery sample gen across languages.

tcoffee-google:
Earlier discovery sample gen implementations diverged in features and
style due to separate implementations. This change brings a consistent
feature set and style to discovery samples across all languages,
including all known changes planned for the samples themselves, except
for media upload segments.

Missing features added:
* page iteration in PHP
* page-element iteration in Node.js
* map-element iteration in all except Java, Python, and PHP
* placeholder pretty-printing of output in all except Python
* informative variable naming/typing in Node.js, PHP, Go, and C#
* distinguishing comments for patch methods

Features unified across language samples:
* naming and typing inference for variables (including backend logic,
  except for C#)
* variable naming for non-inferred names, where appropriate
* parameter description formatting, including cases of missing
  descriptions
* content and formatting of code comments, where appropriate (except for
  unique call comment in Ruby, currently needed to pass compile check)
* spacing between sample elements (except for exclusion of multiple
  blank lines in Go, for consistency with gofmt)
* clarification of strings vs. identifiers in code comments
* clarification of async method use cases in C#

Features unified across language templates (except for C#):
* logical organization of template and context code
* naming of template entities

Note that ORDERING of template elements will be unified in a following
change to avoid obscuring changes made here (for review).

Due to major differences in the implementation of sample gen for C#,
refactoring of template and context logic for C# will be deferred. This
may result in minor changes to the presentation of default string values
appearing in samples, but should otherwise leave sample output
unchanged.

### DIFF
--- a/src/main/java/com/google/api/codegen/DiscoveryContext.java
+++ b/src/main/java/com/google/api/codegen/DiscoveryContext.java
@@ -66,7 +66,7 @@ public abstract class DiscoveryContext extends CodegenContext {
   // so they can be accessed by templates.
 
   public Api getApi() {
-    return this.getService().getApis(0);
+    return getService().getApis(0);
   }
 
   public String getApiRevision() {
@@ -82,6 +82,108 @@ public abstract class DiscoveryContext extends CodegenContext {
 
   public String getSimpleName(String name) {
     return name.substring(name.lastIndexOf('.') + 1);
+  }
+
+  public String getSampleVarName(String typeName) {
+    return upperCamelToLowerCamel(getSimpleName(typeName));
+  }
+
+  /**
+   * Returns a name for a type's field's type.
+   */
+  public String typeName(Type type, Field field, String name) {
+    if (field.getCardinality() == Field.Cardinality.CARDINALITY_REPEATED) {
+      if (isMapField(type, field.getName())) {
+        return mapTypeName(field);
+      } else {
+        return arrayTypeName(field);
+      }
+    } else {
+      if (field.getKind() == Field.Kind.TYPE_MESSAGE) {
+        return objectTypeName(field);
+      } else {
+        return nativeTypeName(type, field, name);
+      }
+    }
+  }
+
+  /**
+   * Returns a name for an array field's type.
+   */
+  protected String arrayTypeName(Field field) {
+    return arrayTypeName(elementTypeName(field));
+  }
+
+  protected String arrayTypeName(String elementName) {
+    return String.format("%s_array", elementName);
+  }
+
+  /**
+   * Returns a name for a map field's type.
+   */
+  protected String mapTypeName(Field field) {
+    return mapTypeName(keyTypeName(field), valueTypeName(field));
+  }
+
+  protected String mapTypeName(String keyName, String valueName) {
+    return String.format("%s_to_%s_map", keyName, valueName);
+  }
+
+  /**
+   * Returns a name for an object field's type.
+   */
+  public String objectTypeName(Field field) {
+    return objectTypeName(field.getTypeUrl());
+  }
+
+  protected String objectTypeName(String typeName) {
+    return upperCamelToLowerUnderscore(typeName);
+  }
+
+  /**
+   * Returns a name for a natively-typed field's type.
+   */
+  protected String nativeTypeName(Type type, Field field, String name) {
+    return name;
+  }
+
+  /**
+   * Returns a name for an array field element's type.
+   */
+  public String elementTypeName(Field field) {
+    Type items = getApiaryConfig().getType(field.getTypeUrl());
+    if (field.getKind() == Field.Kind.TYPE_MESSAGE) {
+      Field elements = getField(items, DiscoveryImporter.ELEMENTS_FIELD_NAME);
+      if (elements != null) {
+        return typeName(items, elements, "item");
+      } else {
+        return objectTypeName(field);
+      }
+    }
+    return nativeElementTypeName(field);
+  }
+
+  /**
+   * Returns a name for a natively-typed array field element's type.
+   */
+  protected String nativeElementTypeName(Field field) {
+    return "item";
+  }
+
+  /**
+   * Returns a name for a map field key's type.
+   */
+  public String keyTypeName(Field field) {
+    Type items = getApiaryConfig().getType(field.getTypeUrl());
+    return typeName(items, getField(items, "key"), "name");
+  }
+
+  /**
+   * Returns a name for a map field value's type.
+   */
+  public String valueTypeName(Field field) {
+    Type items = getApiaryConfig().getType(field.getTypeUrl());
+    return typeName(items, getField(items, "value"), "value");
   }
 
   @Nullable
@@ -213,15 +315,34 @@ public abstract class DiscoveryContext extends CodegenContext {
     return lineEnding(stringLiteral(Strings.nullToEmpty(DefaultString.forPattern(stringPattern))));
   }
 
-  // Line wrap `str`, returning a list of lines. Each line in the returned list is guaranteed to not
-  // have new line characters. The first line begins with `firstLinePrefix` (defaults to list
-  // bullet: "* "), while subsequent lines begin with a hanging indent of equal width.
+  /**
+   * Returns description of type's field from {@link ApiaryConfig}, or field's name if no description
+   * is available.
+   */
+  public String getDescription(String typeName, String fieldName) {
+    String description = apiaryConfig.getDescription(typeName, fieldName);
+    if (description != null) {
+      return description;
+    } else {
+      return fieldName;
+    }
+  }
+
+  /**
+   * Line wrap `str`, returning a list of lines. Each line in the returned list is guaranteed to not
+   * have new line characters. The first line begins with `firstLinePrefix` (defaults to list
+   * bullet: "* "), while subsequent lines begin with a hanging indent of equal width.
+   */
   public List<String> lineWrapDoc(String str, int maxWidth, String firstLinePrefix) {
     return s_lineWrapDoc(str, maxWidth, firstLinePrefix);
   }
 
   public List<String> lineWrapDoc(String str, int maxWidth) {
     return lineWrapDoc(str, maxWidth, "* ");
+  }
+
+  public List<String> lineWrapDoc(String str) {
+    return lineWrapDoc(str, 100);
   }
 
   // For testing.
--- a/src/main/java/com/google/api/codegen/go/GoDiscoveryContext.java
+++ b/src/main/java/com/google/api/codegen/go/GoDiscoveryContext.java
@@ -41,6 +41,21 @@ public class GoDiscoveryContext extends DiscoveryContext implements GoContext {
           .put(Field.Kind.TYPE_DOUBLE, "0.0")
           .build();
 
+  @Override
+  protected String arrayTypeName(String elementName) {
+    return String.format("%sArray", elementName);
+  }
+
+  @Override
+  protected String mapTypeName(String keyName, String valueName) {
+    return String.format("%sTo%sMap", keyName, valueName);
+  }
+
+  @Override
+  protected String objectTypeName(String typeName) {
+    return upperCamelToLowerCamel(typeName);
+  }
+
   public String typeDefaultValue(Type type, Field field) {
     if (field.getCardinality() == Field.Cardinality.CARDINALITY_REPEATED) {
       return typeName(type, field) + "{}";
@@ -82,14 +97,14 @@ public class GoDiscoveryContext extends DiscoveryContext implements GoContext {
     String arrayPrefix = "";
 
     if (field.getCardinality() == Field.Cardinality.CARDINALITY_REPEATED) {
-      Type items = this.getApiaryConfig().getType(fieldTypeName);
+      Type items = getApiaryConfig().getType(fieldTypeName);
       if (isMapField(type, fieldName)) {
         return String.format(
             "map[%s]%s",
-            typeName(items, this.getField(items, "key")),
-            typeName(items, this.getField(items, "value")));
+            typeName(items, getField(items, "key")),
+            typeName(items, getField(items, "value")));
       }
-      Field elements = this.getField(items, "elements");
+      Field elements = getField(items, "elements");
       if (elements != null) {
         return "[]" + typeName(items, elements);
       }
--- a/src/main/resources/com/google/api/codegen/go/discovery_fragment.snip
+++ b/src/main/resources/com/google/api/codegen/go/discovery_fragment.snip
@@ -9,101 +9,136 @@
 @end
 
 @snippet generateClass(method, body)
-package main
+  @let api = context.getApi, \
+       apiName = api.getName, \
+       apiVersion = api.getVersion, \
+       serviceTitle = context.getService.getTitle
+    package main
+    
+    // BEFORE RUNNING:
+    // ---------------
+    // 1. If not already done, enable the {@serviceTitle}
+    //    and check the quota for your project at
+    //    https://console.developers.google.com/apis/api/{@apiName}
+    // 2. This sample uses Application Default Credentials for authentication.
+    //    If not already done, install the gcloud CLI from
+    //    https://cloud.google.com/sdk/ and run
+    //    `gcloud beta auth application-default login`
+    // 3. Install and update the Go dependencies by running `go get -u` in the
+    //    project directory.
+    
+    import (
+      "encoding/json"
+      "fmt"
+      "log"
+    
+      "golang.org/x/net/context"
+      "golang.org/x/oauth2/google"
+    
+      "google.golang.org/api/{@apiName}/{@apiVersion}"
+    )
+    
+    func main() {
+      ctx := context.Background()
+      # TODO(pongad): Every API except autoscaler has the string as API.CloudPlatformScope, find out why
 
-// BEFORE RUNNING:
-// ---------------
-// 1. If not already done, enable the {@context.getService.getTitle}
-//    and check the quota for your project at
-//    https://console.developers.google.com/apis/api/{@context.getApi.getName}
-// 2. This sample uses Application Default Credentials for authentication.
-//    If not already done, install the gcloud CLI from
-//    https://cloud.google.com/sdk/ and run
-//    'gcloud beta auth application-default login'
-// 3. Install and update the Go dependencies by running 'go get -u' in the
-//    project directory.
+      // Authentication is provided by the `gcloud` tool when running locally
+      // and by built-in service accounts when running on GAE, GCE, or GKE.
+      // For more information, see
+      // https://developers.google.com/identity/protocols/application-default-credentials
+      {@clientName()}, err := google.DefaultClient(ctx{@scopeArg(method)})
+      if err != nil {
+        log.Fatal(err)
+      }
 
-import (
-  "golang.org/x/net/context"
-  "golang.org/x/oauth2/google"
+      // Construct the {@apiName} service object for interacting with the API.
+      // You can browse other available API services and versions at
+      // https://github.com/google/google-api-go-client
+      {@serviceName()}, err := {@apiName}.New({@clientName()})
+      if err != nil {
+        log.Fatal(err)
+      }
 
-  "google.golang.org/api/{@context.getApi.getName}/{@context.getApiVersion}"
-)
-
-func main() {
-  ctx := context.Background()
-  # TODO(pongad): Every API except autoscaler has the string as API.CloudPlatformScope, find out why
-  hc, err := google.DefaultClient(ctx{@scopeArg(method)})
-  if err != nil {
-    // {@TODO()} Handle error.
-  }
-  c, err := {@context.getApi.getName}.New(hc)
-  if err != nil {
-    // {@TODO()} Handle error.
-  }
-
-  {@initialize(method)}
-
-  @if context.isPageStreaming(method)
-    {@iterableCall(method)}
-  @else
-    {@compactCall(method)}
+      {@initialize(method)}
+      @if context.isPageStreaming(method)
+        {@iterableCall(method)}
+      @else
+        {@compactCall(method)}
+      @end
+    }
   @end
-}
 @end
 
 @private scopeArg(method)
   @let scopeList = context.getAuthScopes(method)
     @if not(scopeList.isEmpty)
-    , {@context.getApi.getName}.{@scopeList.get(0)}Scope
+      , {@context.getApi.getName}.{@scopeList.get(0)}Scope
     @end
   @end
 @end
 
 @private compactCall(method)
   @if context.isResponseEmpty(method)
-    @if context.hasRequestField(method)
-      call := {@methodPath(method)}({@callArg(method)}).Context(ctx)
-      if err := call.Do(); err != nil {
-        // {@TODO()} Handle error.
-      }
-    @else
-      if err := {@methodPath(method)}({@callArg(method)}).Context(ctx).Do(); err != nil {
-        // {@TODO()} Handle error.
-      }
-    @end
-  @else
-    # TODO(pongad): generate a better name for "resp"
-    resp, err := {@methodPath(method)}({@callArg(method)}).Context(ctx).Do()
-    if err != nil {
-      // {@TODO()} Handle error.
+    if err := {@request(method)}.Context(ctx).Do(); err != nil {
+      log.Fatal(err)
     }
-    // {@TODO()} Use resp.
-    _ = resp
+  @else
+    {@responseName()}, err := {@request(method)}.Context(ctx).Do()
+    if err != nil {
+      log.Fatal(err)
+    }
+
+    // {@TODO()} Change code below to process the `{@responseName()}` object:
+    {@printName()}, err := json.MarshalIndent({@responseName()}, "", "  ")
+    if err != nil {
+      log.Fatal(err)
+    }
+    fmt.Println({@printName()})
   @end
 @end
 
 @private iterableCall(method)
   @let responseType = method.getResponseTypeUrl, \
-      responseTypeRef = context.getApiaryConfig.getType(responseType), \
-      resourceField = context.getFirstRepeatedField(responseTypeRef), \
-      resources = context.lowerCamelToUpperCamel(resourceField.getName)
-
-    call := {@methodPath(method)}({@callArg(method)})
-    if err := call.Pages(ctx, func(page *{@context.getApi.getName}.{@responseType}) error {
-      for _, v := range page.{@resources} {
-        // {@TODO()} Use v.
-        _ = v
-      }
-      return nil // NOTE: returning a non-nil error stops pagination.
+       responseTypeRef = context.getApiaryConfig.getType(responseType), \
+       resourceField = context.getFirstRepeatedField(responseTypeRef), \
+       resources = context.lowerCamelToUpperCamel(resourceField.getName)
+    {@requestName()} := {@request(method)}
+    if err := {@requestName()}.Pages(ctx, func(page *{@context.getApi.getName}.{@responseType}) error {
+      @if context.isMapField(responseTypeRef, resourceField.getName)
+        @let keyName = context.keyTypeName(resourceField), \ 
+	     valueName = context.valueTypeName(resourceField)
+          for {@keyName}, {@valueName} := range page.{@resources} {
+            // {@TODO()} Change code below to process each ({@keyName}: {@valueName}) element:
+            {@printName()}, err := json.MarshalIndent({@valueName}, "", "  ")
+            if err != nil {
+              return err
+            }
+            fmt.Println(fmt.Sprintf("%v: %v", {@keyName}, {@printName()}))
+          }
+        @end
+      @else
+        @let resourceName = context.elementTypeName(resourceField)
+          for _, {@resourceName} := range page.{@resources} {
+            // {@TODO()} Change code below to process each `{@resourceName}` resource:
+            {@printName()}, err := json.MarshalIndent({@resourceName}, "", "  ")
+            if err != nil {
+              return err
+            }
+            fmt.Println({@printName()})
+          }
+        @end
+      @end
+      return nil
     }); err != nil {
-      // {@TODO()} Handle error.
+      log.Fatal(err)
     }
   @end
 @end
 
-@private methodPath(method)
-  c.{@resourcePath(method)}.{@context.lowerCamelToUpperCamel(context.getMethodName(method))}
+@private request(method)
+  @let methodName = context.lowerCamelToUpperCamel(context.getMethodName(method))
+    {@serviceName()}.{@resourcePath(method)}.{@methodName}({@arguments(method)})
+  @end
 @end
 
 @private resourcePath(method)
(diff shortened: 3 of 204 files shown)


tcoffee-google:
This PR intends to unify discovery sample templates across languages, both in output and implementation, except for the implementation for C#, which was sufficiently different to defer until later. I'd appreciate review of individual languages' revisions by those familiar with the languages and original templates:

@pongad : Node.js, Go
@garrettjonesgoogle / @shinfan : PHP
@chrisdunelm : C#

I'm most concerned about catching accidental code mistakes in the sample output, since we do not yet have systematic run-time checking. Style comments are also welcome.

src/main/java/com/google/api/codegen/go/GoDiscoveryContext.java
```diff
@@ -41,6 +41,21 @@ public GoDiscoveryContext(Service service, ApiaryConfig apiaryConfig) {
           .put(Field.Kind.TYPE_DOUBLE, "0.0")
           .build();
 
+  @Override
+  protected String arrayTypeName(String elementName) {
```
pongad: Where is this used? Go's array of ints is `[]int` not `intArray`. I also don't see this reflected in baseline test for any language. (I ran git diff and regexp `^\+.*Array`).

src/main/resources/com/google/api/codegen/go/discovery_fragment.snip
```diff
@@ -113,35 +148,76 @@ func main() {
...
+      {@argumentList(params)}
     @end
-  @else
-    {@callArgNoBody(method)}
   @end
 @end
 
-@private callArgNoBody(method)
-  @join param : context.getFlatMethodParams(method) on ", "
+@private clientName()
```
pongad: The Go team was against naming this `client`, preferring a shorter name. See: https://github.com/googleapis/toolkit/pull/180

src/main/resources/com/google/api/codegen/go/discovery_fragment.snip
```diff
@@ -9,101 +9,136 @@
...
-    @end
+    if err := {@request(method)}.Context(ctx).Do(); err != nil {
+      log.Fatal(err)
+    }
   @else
-    # TODO(pongad): generate a better name for "resp"
-    resp, err := {@methodPath(method)}({@callArg(method)}).Context(ctx).Do()
+    {@responseName()}, err := {@request(method)}.Context(ctx).Do()
+    if err != nil {
+      log.Fatal(err)
```
pongad: If I remember right, `log.Fatal` was replaced with a comment with the reason that users will probably paste this into a non-main function and `Fatal` is OK for main but essentially nowhere else. #116 was made specifically to do this, but I seem to have lost the discussion that led up to it. I believe that it was @broady who raised this point.

src/main/resources/com/google/api/codegen/go/discovery_fragment.snip
```diff
@@ -9,101 +9,136 @@
...
+    }
+
+    // {@TODO()} Change code below to process the `{@responseName()}` object:
+    {@printName()}, err := json.MarshalIndent({@responseName()}, "", "  ")
```
pongad: I'm not against this change but want Go team member to sign off on this before merging. @broady Do you have an opinion?

pongad:
I commented on changes in Go above. I'm good with changes to Node. @jmuk (IIRC) reviewed my original commit for Node and might have some thoughts.

src/main/java/com/google/api/codegen/go/GoDiscoveryContext.java
tcoffee-google: (1) The terminology gets a bit confusing because Go is the only statically typed language where we rely on type inference. In the other statically typed languages (Java, C#), we use the "xxxTypeName" functions for explicit type names, while corresponding variable names may or may not be constructed based on the type name, depending on context. In the dynamically typed languages, these functions provide the same kind of information, but we use them only to construct informative variable names. Due to Go's idiomatic use of type inference, they serve the same function here, rather than providing actual Go type names.

(2) None of the Cloud API methods (which we currently baseline test) require sample gen to create new array variables (e.g., as method parameters). However, this possibility is specified by discovery, and IIRC does occur in some non-Cloud APIs.

src/main/resources/com/google/api/codegen/go/discovery_fragment.snip
tcoffee-google: https://github.com/googleapis/toolkit/pull/180 changed `httpClient` to `hc` where `client` was already used elsewhere. I now use `service` for the latter to be consistent with our usage across languages. So we can use simply `client` for the client without naming collision.

tcoffee-google: I see this reverses part of https://github.com/googleapis/toolkit/pull/117, but I don't understand why that change was made. If there's a good reason, we should consider a change across all languages.

src/main/resources/com/google/api/codegen/go/discovery_fragment.snip
tcoffee-google: After discussion with @saicheems , we both felt that `log.Fatal` was more usable for the most critical use case, that of trying to get an API call working for the first time. In the previous formulation, these error handlers generally accounted for a majority of TODOs in the Go samples. And in cases where they actually result in an error, the user would be required to develop them before seeing error messages. The current behavior seems more useful, providing a similar experience to the samples in the other languages (with exception-like error handling).

If others strongly disagree, I propose a combination as follows, which will preserve the desirable features while still alerting the user with additional TODOs:
```
if err != nil {
  // TODO: Change code below to handle error.
  log.Fatal(err)
}
```

saicheems: I think Thomas' proposal is reasonable. The reason we argued for `log.Fatal` was that in the case of an error, the program would crash more cleanly than with a fatal exception from trying to operate on a `nil` pointer or whatever came out. Plus, the user is more quickly prepared to deal with an error if it's printed for him/her, vs. editing the program to include a `fmt.Println` after the fact.

Either way, they have to change the placeholder for production.

src/main/resources/com/google/api/codegen/php/discovery_fragment.snip
```diff
@@ -31,99 +33,176 @@
...
+    {@request(method)};
+  @else
+    ${@responseName()} = {@request(method)};
+
+    // {@TODO()} Change code below to process the `${@responseName()}` object:
+    {@print(responseName())}
+  @end
+@end
+
+@private request(method)
```
shinfan: request is not a very descriptive name. Maybe makeRequest or methodCall?

src/main/resources/com/google/api/codegen/php/discovery_fragment.snip
```diff
@@ -31,99 +33,176 @@
...
       @end
-    @end
 
-    @if context.hasRequestField(method)
+    @end
+  @end
+  @if context.hasRequestField(method)
 
-      $postBody = new {@context.getRequestClassName(method)}($client);
+    ${@requestBodyName(method)} = new {@context.getRequestClassName(method)}($client);
```
shinfan: The php client call it "$postBody". I think it should be consistent here.

src/main/resources/com/google/api/codegen/go/discovery_fragment.snip
vchudnov-g: I like the combination solution, particularly in light of Sai's comments that "Either way, they have to change the placeholder for production."

src/main/java/com/google/api/codegen/go/GoDiscoveryContext.java
vchudnov-g: In the java example, these methods have the syntax of type declarations. Here in Go, they don't. Are you saying in your comment that for Go we only use the output of these functions for the variable names? If so, it seems like you're overloading the use of the functions. It may be clearer to have arrayType() vs arrayName(). Or maybe I'm misunderstanding this.

src/main/java/com/google/api/codegen/DiscoveryContext.java
```diff
@@ -84,6 +84,108 @@ public String getSimpleName(String name) {
     return name.substring(name.lastIndexOf('.') + 1);
   }
 
+  public String getSampleVarName(String typeName) {
+    return upperCamelToLowerCamel(getSimpleName(typeName));
```
vchudnov-g: The return value is correctly formatted for all languages? If this is meant to be overriden by each language, please comment that.

src/main/java/com/google/api/codegen/DiscoveryContext.java
```diff
@@ -84,6 +84,108 @@ public String getSimpleName(String name) {
...
+
+  /**
+   * Returns a name for an array field's type.
+   */
+  protected String arrayTypeName(Field field) {
+    return arrayTypeName(elementTypeName(field));
+  }
+
+  protected String arrayTypeName(String elementName) {
+    return String.format("%s_array", elementName);
```
vchudnov-g: Is this always idiomatic, as opposed to declaring the array type inline? Same below. Please comment if meant to be overriden by each language.

src/test/java/com/google/api/codegen/testdata/discoveries/csharp/csharp_clouduseraccounts.beta.json.baseline
```diff
@@ -832,46 +973,57 @@ using Google.Apis.Auth.OAuth2;
...
                 HttpClientInitializer = credential,
                 ApplicationName = "Google Cloud Platform Sample",
             });
 
-            // TODO: Change placeholders below to values for parameters to the Get() method:
 
-            // Project ID for this request.
+            // TODO: Change placeholders below to desired parameter values for the `Get` method:
+
+            // * Project ID for this request.
```
broady: NLGTM

broady:
-1 for Go. I can't properly review in GitHub, so here are my comments.

"Authentication is provided by the `gcloud` tool when running locally"

Please don't mention the gcloud tool. Just link to the docs on DAC.

`client, err := google.DefaultClient(ctx`

Please revert `client` back to `hc`.

```
+  // Construct the adexchangebuyer service object for interacting with the API.
+  // You can browse other available API services and versions at
+  // https://github.com/google/google-api-go-client
+  service, err := adexchangebuyer.New(client)
```

Please revert back to `c, err := ...` The comment is not useful. Please remove it.

`// * The account id` Where did this asterisk come from? Please revert.

`res, err := service.Accounts.Get(id).Context(ctx).Do()`

Please revert `res` back to `resp`.

```
// TODO: Change code below to process the `res` object:
+  out, err := json.MarshalIndent(res, "", "  ")
+  if err != nil {
+    log.Fatal(err)
+  }
+  fmt.Println(out)
```

Please revert. There's no need to import the json package for a simple sample.

The original was adequate and common:
```
// TODO: use resp
_ = resp
```

broady:
For what it's worth, I think it's a flawed approach to unify things across languages.

Go is not Java. Ruby is not JavaScript, etc. etc.

Java developers are not looking at JavaScript snippets - there's very little value to them being consistent.

src/main/resources/com/google/api/codegen/go/discovery_fragment.snip
broady: `log.Fatal` is reasonable, I don't feel strongly about it.

vchudnov-g:
@broady To clarify, I would agree there's a strong case for idiomatic style trumping cross-language consistency, but in places where (once we're already idiomatic) we have discretion (variable names and values, boilerplate comments, ...) we should try to be consistent across languages.

lesv:
Idiomatic Go has very specific naming guidelines -- short variable names.  Java, on the other hand, is exceedingly verbose.  @vchudnov-g you might wish to look again.

tcoffee-google:
Building on @vchudnov-g 's comment, I think it is fine to introduce differences between languages where those differences serve a purpose that is distinctive to a language, but gratuitous differences are not desirable. Pointless differences hurt users who use samples in one language to help understand those in another, and hurt our developers trying to maintain features across languages. It is especially important that samples across different languages demonstrate similar functionality: otherwise our samples convey a misleading impression of differences in support or preferred usage across languages.

That said, if anyone feels that the comments or functionality that are provided uniformly across languages should be different than they are (uniformly across languages), please indicate your reasons so we can discuss.

src/main/resources/com/google/api/codegen/php/discovery_fragment.snip
tcoffee-google: Will change corresponding template snippets to `methodCall`.

src/main/resources/com/google/api/codegen/php/discovery_fragment.snip
tcoffee-google: I see this name used in internal client library generation code, but I don't find it in documentation anywhere, and it's a very generic term. I feel the informative variable name is preferable here.

vchudnov-g:
@lesv Sure, for things like loop variables, etc, we should follow the Go convention and have ultra-short names (I include that under "being idiomatic"). For meaningful variables, I think the names should ideally be succinct and consistent as much as possible across languages (eg using "request" in all languages, for example), but maintaining idiomatic use.

src/main/java/com/google/api/codegen/DiscoveryContext.java
tcoffee-google: The latter is correct, though as the only current override is for C#, this may be refactored later. Will add comment for now.

src/main/java/com/google/api/codegen/DiscoveryContext.java
tcoffee-google: As above.

broady:
To be clear, Go style does not mandate or recommend short variable names. It recommends good names. That usually means shorter names for variables used in smaller scopes (e.g., within a function).

`request` is too wordy, `req` gets the point across. That is idiomatic to Go, though, and may not apply to other languages.

`res` is a bad name because it might mean "resource". That's why I chose `resp`.

Please, if you're going to change names, provide justification. I chose them carefully. Please use the same level of consideration when changing them.

tcoffee-google:
On variable names:

* We currently use `req` for Go instead of `request`; no argument there.
* From a look at the Go client library (code and docs), it appears to me that `res` is used much, much more frequently than `resp`; I was attempting to be consistent. That said, I am fine with `res`.
* It's hard for me to see how names like `hc` and `c` will be meaningfully parsed by the user, especially when the phrase "http client" appears nowhere in the code. It does not appear to me that the names `client` and `service` are verbose by Go standards, especially for an explanatory sample. We use `service` to mean the same thing everywhere across languages.

I think these are the only proposed changes.

tcoffee-google:
On gcloud tool: I can't tell if this issue is specific to Go or not. Context?

tcoffee-google:
On asterisks: these form a bulleted list indicating the extent of the TODO directing the user to fill in placeholders (similar in all languages). Without some offset, it is less clear that following code segments (such as the request body) do not fall under the scope of this TODO. We've tossed around various ideas, but I have not yet seen a more compelling approach: other ideas are welcome.

broady:
* `hc` is very often used for a variable of type `*http.Client`.
* `service` is fine, but `c` is better. I'd prefer `fooService`, actually. But I don't like `service` because it will likely conflict with samples for several APIs, including the `appengine` and `servicemanagement` APIs.

On the mention of gcloud: the comment only mentions 3 out of the 4 ways of getting credentials. Also, it doesn't mention which command to run from gcloud to put the credentials in the right place. If a comment is going here at all, then it should be fully accurate. Otherwise, just link to the docs that have the full explanation.

tcoffee-google:
On placeholder JSON output: both our instincts and reviewer feedback suggested that the samples should provide some indication of result behavior on successful execution. This is particularly important given that the samples do not run out-of-the-box and require users to correctly complete some elements, hence to find an indication they have done so. The simplest way we found to achieve meaningful example output across all APIs was through JSON conversion and pretty-printing, which can be done (without the "pretty" part in C#) similarly in all our languages without third-party libraries.

broady:
Just use log.Printf("%#v", resp) then.

vchudnov-g:
@broady To your point, https://talks.golang.org/2014/names.slide#1 . The use of `client` is pretty standard, though: https://golang.org/pkg/net/http/

Asterisks: I would take them out, @tcoffee-google . They don't look nice. Developers should be able to figure it out. We could include all the placeholders inside a single 
```
var (
...
)
```

broady:
`client` is common if you're visibly working with the `net/http` package. In this case, there are effectively two "clients": the `*http.Client` and the client of whatever API is being used.

That variable is only used when calling `New`. Why give it a relatively important name like "client"?

tcoffee-google:
> Just use log.Printf("%#v", resp) then.

So far as I can tell, this will generally produce single-line output, which is very hard to read for non-trivial response objects.

tcoffee-google:
Wow, I couldn't disagree more with Andrew's assertion:

> long names obscure what the code _does_

That may be true if one is unfamiliar with parsing whole words. Maybe we should all just grunt like cavemen so that we can better understand each other just by observing inflection :-)

Re: `service`, I would be happy to generate a context-specific name like `fooService` as we do for Java, with the understanding that `foo` could be as long as the name of the API.

Re: `client`, we should call it something intuitive, it's generated by a method called `DefaultClient`, and it's an HTTP client. Makes sense to me.

tcoffee-google:
> Asterisks: I would take them out

Are you proposing a special case for Go because there is an idiomatic way of indicating the extent of the relevant code segment?

jonparrott:
Golang aside, Python is also very opinionated in possibly the other direction. This PR is too large for me to effectively review, so I'm very concerned about the outcome of this.

saicheems:
Gosh, this is phenomenally difficult to review 😞

tcoffee-google:
Until such time as we resolve our github review issues, my suggestion for review is to cherry-pick the change to a local branch and use git difftool.

tcoffee-google:
> On the mention of gcloud: the comment only mentions 3 out of the 4 ways of getting credentials. Also, it doesn't mention which command to run from gcloud to put the credentials in the right place. If a comment is going here at all, then it should be fully accurate. Otherwise, just link to the docs that have the full explanation.

@broady I've been trying to figure out all that you have in mind here, but it's not immediately clear to me, since I'm not yet too familiar with all the auth mechanisms. In particular, does the command `gcloud beta auth application-default login` in our comment not cover what you're suggesting here for `gcloud`?

It's my understanding that we formulated samples based on application default credentials because that was expected to be the easiest way for a new user to get a method call to run successfully, which was our primary motivation. Our formulation seems consistent to me with what's documented here: https://developers.google.com/identity/protocols/application-default-credentials . What docs are you referring to?

tcoffee-google:
> The asterisks look wrong. They look like a mistake.

In earlier discussion with @saicheems which I'd forgotten, we had suggested that "bulleting" with `--->` instead of `*` would provide more obvious visual cues. @vchudnov-g how do you like this option?

```go
...
  service, err := compute.New(client)
  if err != nil {
    log.Fatal(err)
  }

  // TODO: Change placeholders below to desired parameter values for the `CreateSnapshot` method:

  // ---> Project ID for this request.
  project := ""

  // ---> The name of the zone for this request.
  zone := ""

  // ---> Name of the persistent disk to snapshot.
  disk := ""

  snapshotBody := compute.Snapshot{
    // TODO: Add desired fields of the `snapshotBody` variable.
  }

  res, err := service.Disks.CreateSnapshot(project, zone, disk, &snapshotBody).Context(ctx).Do()
  if err != nil {
    log.Fatal(err)
  }
...
```

I should note that in Go the extent of this code segment is made less clear than in the other languages due to `gofmt`'s intolerance of consecutive blank lines. I like the idea of using a `var` block to remedy this, but I still think the extra visual attention to each placeholder above makes the TODO easier for the user to process.

jonparrott:
> Until such time as we resolve our github review issues, my suggestion for review is to cherry-pick the change to a local branch and use git difftool.

That's pretty disappointing. Is there no way you can break this into smaller, easier to review pieces? Large code changes are difficult to review and typically lead to more issues sliding by.

broady:
I just pushed a bunch of smaller commits to another branch that separates into changes to the baseline test files for each language, and another one for changes to the codegen itself.

tcoffee-google:
@jonparrott Here are **all** the changes made to the Python samples:

* Quote marks delimiting variable names in code comments are changed from `'name'` to ``` `name` ```.
* An indent is added to the TODO comment directing completion of the request body, like so:
```python
xxx_request_body = {
    # TODO: Add desired entries of the `xxx_request_body` dict
}
```
* The comment ```See https://developers.google.com/identity/protocols/application-default-credentials for more information.``` is reworded to ```For more information, see https://developers.google.com/identity/protocols/application-default-credentials``` for better line-wrapping.
* When no description is available for a method parameter, the name of the parameter is used instead to avoid a confusing gap in the commenting, like so:
```python
# * project
project = ''
```

There are **no changes to any code** in the Python samples. 

On the backend, the **only** changes connected to the Python samples are:

* refactoring of 157 LoC in the template file `src/main/resources/com/google/api/codegen/py/discovery_fragment.snip`

* refactoring of 77 LoC in the context file
`src/main/java/com/google/api/codegen/py/PythonDiscoveryContext.java` into the parent context `src/main/java/com/google/api/codegen/DiscoveryContext.java`

* a change to a single function `getDescription` across `src/main/java/com/google/api/codegen/ApiaryConfig.java` and `src/main/java/com/google/api/codegen/DiscoveryContext.java`

I considered doing per-language PRs, but since several issues brought up will potentially impact all languages, that seemed as though it would get extremely confusing. Hope this helps.

broady:
@tcoffee-google Where did the idea of bullets come from? Did users have particular difficulty during user testing?

It doesn't look how someone would write code, which is why it looks wrong in sample code.

In my opinion, the biggest usability problem with the samples isn't anything that this CL might change, but that we don't have placeholder values for required fields in request bodies.

jonparrott:
Thanks @tcoffee-google - that is what I suspected, but it was difficult to verify.

> we don't have placeholder values for required fields in request bodies.

+1 to this.

```python
# * project
project = ''
```

Is basically useless and redundant

```python
project = 'your-project-id'
```

Is far more useful.

vchudnov-g:
Alright, a bunch of comments:

* Yes, this CL is unwieldy large, especially until we get a better Github review system. Using gitdiff should help, though, even if we miss the advantage of adding comments at a line

* Asterisks/bullets in the placeholder comments: I still say we take them out. We can group things together in a `var` block. That's idiomatic.

* Printing with `"%#v"`: I think this is standard and quick in Go, so I would go with this. Yes, it's not the most-humanly readable, but the purpose of our snippets is not to show super-pretty output; it's to demonstrate how to successfully call the code. We only output to demonstrate that we have something. If the user wants to prettify it, they can pull in the tools they wish to do that. I think this also speaks to another meta-point: while not being rigid about it, we should minimize the number of extra packages we import in our snippets.

* With the auth mechanism: our goal is explicitly NOT to show every single possible way of doing things, but to show one way of getting things working. We explicitly chose to use Application Default Credentials for Cloud APIs for all samples and all languages, even of other mechanisms are supported. When we expand to non-Cloud, we will show another auth mechanism instead, but still only one. (And for example, with OAuth, we will show just one scope that applies to the method, not all possible scopes.)

* With the long names: I personally am also in the camp of being more explicit rather than less, but I realize this is often at odds with the Go style we recommend and expose externally. In the interest of unblocking, I'm OK compromising on this. I still think `client` should be OK and should be short enough. And for service, if we're concerned about `fooService` (eg `someInterestingGoogleService`) becoming too long, one idea would be to abbreviate the `foo` part to arrive at a shorter name (eg `sigService`). Thoughts? (Again, personally I'm fine with the longer `fooService`.)

tcoffee-google:
@broady @jonparrott Yes, meaningful placeholder values have been on our to-do list from the beginning, but we currently have no idea how to automate this in general. We've made some recent progress with common patterns like resource paths and zones (see, e.g., https://github.com/googleapis/toolkit/pull/249).

broady:
yeah, let's just use the imported package name (e.g. `servicemanagement` as the prefix.)

The longest one is `acceleratedmobilepageurl`. `acceleratedmobilepageurlClient` is pretty ridiculous, but just fine for an autogenerated sample.

tcoffee-google:
Re: bullets, I think it might be okay to rely on spacing in non-Go languages and a `var` block in Go to make clear the extent of the TODO for placeholders. However, this extent is often visually unclear to me when I look back at the samples (even when I wrote them). The purpose of the bulleted list was to make it more obvious that this TODO (unlike the other TODOs) spans several chunks of code. Since the purpose of the samples is to get users up and running as quickly as possible, but they have several items to complete before that can happen, it serves the users best to make it exceedingly clear what they need to do before the sample will work.

tcoffee-google:
> I just pushed a bunch of smaller commits

@broady Thanks, I hope that will help other folks as well. Our eventual goal is to develop a comprehensive test API so we can dispense with the redundancy in the baseline tests. This did not cause any real problems until we migrated to Github.

vchudnov-g:
On the bullets, I still think using the var block should suffice. Another alternative is to have a TODO after the variable comment, before the variable definition, like this:
```
// The name of the VM
// TODO: Fill in
vmName := ""
```
but I believe we had purposefully moved away from that earlier in order to make the snippets less verbose....

broady:
I removed the var block because that's not what you'd write in actual code.

vchudnov-g:
@broady But this isn't manual code we're writing. These are autogenerated samples. We just want to make the various parts clear to the user.

The alternative is to keep the individual TODOs (and have them inline with the variables, I would say). I think it looks slightly better to have them grouped, but am fine with the individual TODOs. And it addresses the bulleting issue.

tcoffee-google:
@vchudnov-g The problems with inline TODOs like
```go
name := "" // TODO: Change placeholder to desired value
```
are that they (a) obscure the actual placeholder values that require completion between the `name := ` and the TODO, and (b) require much more editing for the user to make the sample into code they would actually write (on each line, the user must navigate to the placeholder and delete the comment). To me, they also seem excessively repetitive.

broady:
@vchudnov-g the fact that it's autogenerated is not relevant. We should autogenerate code that someone would write.

vchudnov-g:
I agree they're repetitive, which is why I was advocating the var block. I don't think the erasing work is onerous. If you think they obscure the placeholder, then maybe we do move them up to a separate line, just below the description, even though that makes the snippet longer in #LOC.

I'm just looking for a way out of this impasse. If everyone digs in their heels, we won't get anywhere ;-)

jonparrott:
> the fact that it's autogenerated is not relevant. We should autogenerate code that someone would write.

+1, I agree with @broady. The go samples should look like they were written by a go developer - which means, regardless of if they are autogenerated they should pass @broady's code review.

Same goes for other languages.

broady:
> I'm just looking for a way out of this impasse. If everyone digs in their heels, we won't get anywhere ;-)

broady:
Oops, wrong button. Sorry!

tcoffee-google:
@saicheems suggested a compromise to clarify the scope of the placeholder TODO without additional elements --- that we autogenerate a list of the variable names requiring placeholder completion, like so:

```go
...
  service, err := compute.New(client)
  if err != nil {
    log.Fatal(err)
  }

  // TODO: Change placeholders below to desired parameter values for the `CreateSnapshot` method: 
  // `project`, `zone`, `disk`.

  // Project ID for this request.
  project := ""

  // The name of the zone for this request.
  zone := ""

  // Name of the persistent disk to snapshot.
  disk := ""

  snapshotBody := compute.Snapshot{
    // TODO: Add desired fields of the `snapshotBody` variable.
  }

  res, err := service.Disks.CreateSnapshot(project, zone, disk, &snapshotBody).Context(ctx).Do()
  if err != nil {
    log.Fatal(err)
  }
...
```

tcoffee-google:
Re: reviewing, here's an easier stopgap to relieve github pain that I tried just now:
```
sudo apt-get install kompare
wget https://github.com/googleapis/toolkit/pull/376.patch
kompare 376.patch &
```

tcoffee-google:
@garrettjonesgoogle @shinfan @pongad @jmuk Any further feedback on PHP, Go, Node.js?

garrettjonesgoogle:
I will defer to @shinfan for PHP - I see that he has already taken a look.

jmuk:
Quickly took a look at the test expectation updates for Node and Ruby, and they look fine. LGTM.

# BTW it's a real pain to look at the change of this size in github. Sigh.

src/main/java/com/google/api/codegen/go/GoDiscoveryContext.java
tcoffee-google: Go is in some sense a special case here, so I'll add comments to that effect. The logic is identical across all languages, so it seems redundant to create another set of identically-structured functions.

pongad:
@tcoffee-google I think the exact format of the TODOs is a [bikeshed](https://en.wikipedia.org/wiki/Law_of_triviality). With that said, @broady and I spent a good part of an afternoon typing furiously over Hangout to arrive at the current design, so I want to document some of my reasoning. @broady might have different reasons.

Keep in mind that I code with no syntax highlighting (comments are still highlighted in blue) and variable-width font. At the risk of sounding ostentatious, my experience might be different from most people's.

I completely agree that the TODOs look repetitive, but repetition is sometimes good for automation. When I was running tasks (I think I used 4 methods in total) for compute to make sure the samples work, I barely used by text editor after copy-pasting the samples.

I ended up doing something like this. (Example typed up by hand, might get some details slightly wrong):
```
$ grep TODO runcompute.go | head -n 1
    // TODO: Handle error.
$ findrep -F '// TODO: Handle error.' 'log.Fatal(err)' runcompute.go
$ grep TODO runcompute.go | head -n 1
  project := "" // TODO: Update placeholder value.
$ findrep 'project.*placeholder.*' 'project := "projects/testing-project"' runcompute.go
```
`findrep` is a tool I wrote for myself.

I was actually annoyed that `// TODO: Use resp.` and `_ = resp` are on different lines. Hoisting the TODO comments, either by using the `var` block or listing variable names in a comment, reduces the number of "things" on the page, I agree. However, it makes getting the samples to actually work more difficult. This is true even if our users don't do things on the command line like I do. With some tooling, TODOs are easy to find and easy to fix. Why make users go through pointer indirections?

tcoffee-google:
@pongad Indeed we seem to be following [Wadler's Law](https://wiki.haskell.org/Wadler's_Law) to a T :-)

Thanks for sharing a detailed perspective on workflow. I can appreciate that repeated inline TODOs make for easier code editing using command-line regex tools, though I have to agree this is probably a less common use case.

I forgot to mention that inline TODO comments are currently incompatible with some types of default values (e.g., formatted date strings), which are generated with their own inline comments to inform the user of their format. These don't appear in Cloud APIs, and so haven't shown up yet in sample output.

Following discussion with @vchudnov-g and @saicheems , I'm inclined to settle for the simplest option thus far proposed, to which I don't think anyone else had objections:
```go
...
  computeService, err := compute.New(client)
  if err != nil {
    log.Fatal(err)
  }

  // TODO: Change placeholders below to desired parameter values for the `CreateSnapshot` method:

  // Project ID for this request.
  project := ""

  // The name of the zone for this request.
  zone := ""

  // Name of the persistent disk to snapshot.
  disk := ""

  snapshotBody := compute.Snapshot{
    // TODO: Add desired fields of the `snapshotBody` variable.
  }

  resp, err := computeService.Disks.CreateSnapshot(project, zone, disk, &snapshotBody).Context(ctx).Do()
  if err != nil {
    log.Fatal(err)
  }
...
```

In non-Go languages, I think my concerns are satisfactorily addressed by additional spacing.

tcoffee-google:
> We only output to demonstrate that we have something. If the user wants to prettify it, they can pull in the tools they wish to do that. 

I disagree that a mere success indicator adds the same value. If that were true, we could simply write
```go
if err != nil {
  log.Fatal(err)
} else {
  log.Println("Success!")
}
```
A major purpose of our samples as they exist is to make clear the structure of request components as used by the client library. As the user's task does not end with obtaining the response (they want to do something with it), it seems only reasonable that the sample should help make clear the structure of the response they receive to facilitate further development. 

This purpose may not be ideally served by the JSON serialization approach currently taken in Go, Java, C#, and Ruby, which lack a readily available direct means of recursively printing an object/structure. OTOH, it is probably too verbose in general to generate our own complete structured print of response objects. For the languages (Go, Java, C#) lacking an interactive REPL, elaborating the response still seems like a significant usability gain, particularly given that many users may have no idea how to do this themselves.

broady:
Just a few changes to that code:

* `client` -> `hc`
* `snapshotBody` -> `snapshot` (body is an implementation detail)
* Drop "of the snapshotBody variable" from the comment. It's already inside the struct that's being referenced. No need to say it twice.
* "method" -> "call"

I'm also generally -1 on the backticks in the comments, but don't feel strongly about it. 

Please also add back the `TODO`s. Many editors highlight TODO differently. Even without highlighting, the visual indicator of all-caps helps call out that action is required from the user.

Otherwise LGTM.

pongad:
> easier code editing using command-line regex tools, though I have to agree this is probably a less common use case
But Ctrl+F for TODO is very common use case right? I don't care so much whether they're on the same line. (Same line makes things a little easier for me, but if it's nicer for most people on different lines, I can cope.) I would however like to argue that there is great value in having one TODO per thing to change.

saicheems:
Merged in #398

PR #378, opened 2016-08-15
Fixing inconsistent proto versions

garrettjonesgoogle:


build.gradle
```diff
@@ -35,6 +35,8 @@ version = "0.0.0-SNAPSHOT"
 // url to the snapshot repository of a private maven repo.
 def toolkitSnapshotRepo = "http://104.197.230.53:8081/nexus/content/repositories/snapshots/"
 
+def protoVersion = "3.0.0-beta-3"
```
geigerj: nit: I guess this isn't directly related to this PR, since it's also in line 36, but should we stick to single-quoted strings unless the variable replacement feature is needed?

geigerj:
LGTM

build.gradle
bjwatson: :+1: Makes sense for bash, and also makes sense for Groovy.

bjwatson:
LGTM2. Your choice if you want to address the nit.

build.gradle
garrettjonesgoogle: There are a bunch of double-quoted strings in the file; I think it makes sense to do a separate pass for that.

PR #380, opened 2016-08-16
Preparing non-snapshot release 0.0.1

garrettjonesgoogle:


build.gradle
```diff
@@ -246,34 +251,68 @@ tasks.cleanEclipseJdt {
...
         }
 
         pom.project {
-          name 'Gapi Tools'
+          name 'Google API Tools'
           packaging 'jar'
-          description 'Google API tools frameworks.'
-          url 'http://www.googleapis.com'
+          artifactId 'toolkit'
+          description 'Google API tools framework.'
```
geigerj: Is this the right full description for `toolkit`? I find it somewhat confusing given that the API compiler contains the package `com.google.api.tools.framework` -- I think of toolkit as built on top of the framework, rather than being a framework itself.

build.gradle
```diff
@@ -289,6 +328,15 @@ allprojects {
 //
 // Command line args can be passed to the tool as a comma-separated list:
 //   ./gradlew runVGen '-Pclargs=--arg1=val1,--arg2=val2,--arg3=val3'
```
geigerj: Update usage

build.gradle
```diff
@@ -289,6 +328,15 @@ allprojects {
...
+task runCodeGen(type: JavaExec) {
+  classpath = sourceSets.main.runtimeClasspath
+  main = 'com.google.api.codegen.CodeGeneratorTool'
+  if (project.hasProperty('clargs')) {
+    args = clargs.split(',').toList()
+  }
+}
+
+// old task being retained for backwards-friendliness
```
geigerj: Are there other projects that call this task besides `artman`? If not, can you just update artman and remove the duplication?

build.gradle
garrettjonesgoogle: Good point, I will update the description.

build.gradle
garrettjonesgoogle: It should just be artman, but I think it would be less frustrating to the team in general if artman/toolkit can be updated at the leisure of each dev.

geigerj: In that case, can you update artman and leave a TODO to delete this code, say, in a month? Otherwise, it might just get forgotten.

bjwatson: Agreed. We should be proactive about our tech debt.

@ethanbao FYI. Once your `toolkit` dependency is updated, we can strip out this code.

garrettjonesgoogle:
Updated, PTAL

geigerj:
LGTM, but please wait for Brian's approval, too.

bjwatson:
LGTM2

PR #381, opened 2016-08-16
Code Gen and Config Gen: using api list from service yaml

garrettjonesgoogle:
This is in preparation for mixin support.

Before this change, any interface discovered in the proto path will have
a corresponding VKit wrapper generated. After this change, only
interfaces in the api list of the service config will have wrappers generated.

### DIFF
--- a/src/main/java/com/google/api/codegen/InterfaceView.java
+++ b/src/main/java/com/google/api/codegen/InterfaceView.java
@@ -16,6 +16,7 @@ package com.google.api.codegen;
 
 import com.google.api.tools.framework.model.Interface;
 import com.google.api.tools.framework.model.Model;
+import com.google.protobuf.Api;
 
 import java.util.ArrayList;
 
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/api/codegen/InterfaceView.java
```diff
@@ -26,15 +27,13 @@
...
 
   /**
-   * Gets the reachable interfaces of the model.
+   * Gets the interfaces for the apis in the service config.
    */
   @Override
   public Iterable<Interface> getElementIterable(Model model) {
     ArrayList<Interface> interfaces = new ArrayList<>();
-    for (Interface iface : model.getSymbolTable().getInterfaces()) {
-      if (iface.isReachable()) {
```
michaelbausor: Was isReachable doing any work before? And if so, do we need something equivalent in the updated code?

garrettjonesgoogle: I glanced through the code, and I'm not super sure. I think it is doing two things:

1. Making sure the interface is "referenced" directly or indirectly by the APIs in the API list
2. Not including interfaces which have visibility marked as internal

Use case 1 is basically cut out because we are going to the same data source; use case 2 is not a concern because we are already working on stripped protos. I'm not sure if there are other cases I am missing, but I'm not terribly concerned because we'll find out at generation time and we can fix it then.

michaelbausor:
LGTM

PR #384, opened 2016-08-17
Update Ruby codegen.

jmuk:
- Remove rubocop disable comments.
  Decided to exclude those generated files from the rubocop target,
  so these comments are not necessary (we still can check the styles
  by manually invoking rubocop).
- Update the require path for ruby.
  The new path (_pb suffix) is the style introduced in protobuf 3.0.
  See googleapis/gax-ruby#37


bjwatson:
It might be a good idea to also bump the toolkit dependency on protobuf: https://github.com/googleapis/toolkit/pull/374/files

geigerj:
LGTM

Landrito:
LGTM

PR #386, opened 2016-08-17
Fix issues in java test generation.

shinfan:
- Fix names
  - Handle duplicate test case name since we may have more than one test for each method.
  - requests -> actualRequests
  - list field getter: getXXXX() -> getXXXList()
- Use LinkedList to store mock requests for efficiency.
- In the initialization stage, initialize all the flattened fields instead of required fields.
- Ignore the response if the grpc method has no return value.

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
+++ b/src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
@@ -447,9 +447,17 @@ public class SurfaceNamer extends NameFormatterDelegator {
     return getNotImplementedString("SurfaceNamer.getAndSavePagedResponseTypeName");
   }
 
-  /** The test case name for the given method. */
-  public String getTestCaseName(Method method) {
-    return methodName(Name.upperCamel(method.getSimpleName(), "Test"));
+  /**
+   * The test case name for the given method.
+   * Use the given count value to produce unique test case names if there are multiple tests
+   * for each method
+   */
+  public String getTestCaseName(Method method, Integer count) {
+    if (count > 1) {
+      return methodName(Name.upperCamel(method.getSimpleName(), "Test" + Integer.toString(count)));
+    } else {
+      return methodName(Name.upperCamel(method.getSimpleName(), "Test"));
+    }
   }
 
   /** The test class name for the given API service. */
@@ -468,7 +476,11 @@ public class SurfaceNamer extends NameFormatterDelegator {
   }
 
   /** The method name of getter function call for the given name */
-  public String getGetFunctionCallName(Name name) {
-    return methodName(Name.from("get").join(name));
+  public String getGetFunctionCallName(Name name, TypeRef type) {
+    if (type.isRepeated() && !type.isMap()) {
+      return methodName(Name.from("get").join(name).join("list"));
+    } else {
+      return methodName(Name.from("get").join(name));
+    }
   }
 }
--- a/src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
+++ b/src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
@@ -449,8 +449,9 @@ public class SurfaceNamer extends NameFormatterDelegator {
 
   /**
    * The test case name for the given method.
+   *
    * Use the given count value to produce unique test case names if there are multiple tests
-   * for each method
+   * for one method.
    */
   public String getTestCaseName(Method method, Integer count) {
     if (count > 1) {
--- a/src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
@@ -54,7 +54,8 @@ public class JavaGapicSurfaceTestTransformer implements ModelToViewTransformer {
 
   private GapicCodePathMapper pathMapper;
 
-  // TODO(shinfan): Reuse this logic with InitCodeGenerator
+  // TODO: Reuse this logic with InitCodeGenerator
+  // Github issue: https://github.com/googleapis/toolkit/issues/395
   private class TestCaseNameTable {
     private HashMap<String, Integer> nameCount;
 
@@ -180,9 +181,10 @@ public class JavaGapicSurfaceTestTransformer implements ModelToViewTransformer {
           testCaseViews.add(createTestCaseView(methodContext, paramFields, nameTable));
         }
       } else {
-        // TODO(shinfan): Add support of non-flattening method
+        // TODO: Add support of non-flattening method
+        // Github issue: https://github.com/googleapis/toolkit/issues/393
         System.err.println(
-            "Non-flatenning method test is not supported yet for " + method.getSimpleName());
+            "Non-flattening method test is not supported yet for " + method.getSimpleName());
       }
     }
     return testCaseViews;
--- a/src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
@@ -54,6 +54,7 @@ public class JavaGapicSurfaceTestTransformer implements ModelToViewTransformer {
 
   private GapicCodePathMapper pathMapper;
 
+  // TODO(shinfan): Reuse this logic with InitCodeGenerator
   private class TestCaseNameTable {
     private HashMap<String, Integer> nameCount;
 
@@ -180,6 +181,8 @@ public class JavaGapicSurfaceTestTransformer implements ModelToViewTransformer {
         }
       } else {
         // TODO(shinfan): Add support of non-flattening method
+        System.err.println(
+            "Non-flatenning method test is not supported yet for " + method.getSimpleName());
       }
     }
     return testCaseViews;
--- a/src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
@@ -17,6 +17,7 @@ package com.google.api.codegen.transformer.java;
 import com.google.api.codegen.ApiConfig;
 import com.google.api.codegen.InterfaceView;
 import com.google.api.codegen.MethodConfig;
+import com.google.api.codegen.ServiceMessages;
 import com.google.api.codegen.gapic.GapicCodePathMapper;
 import com.google.api.codegen.transformer.InitCodeTransformer;
 import com.google.api.codegen.transformer.MethodTransformerContext;
@@ -35,11 +36,13 @@ import com.google.api.codegen.viewmodel.testing.MockGrpcMethodView;
 import com.google.api.codegen.viewmodel.testing.MockGrpcResponseView;
 import com.google.api.codegen.viewmodel.testing.MockServiceImplView;
 import com.google.api.codegen.viewmodel.testing.MockServiceView;
+import com.google.api.tools.framework.model.Field;
 import com.google.api.tools.framework.model.Interface;
 import com.google.api.tools.framework.model.Method;
 import com.google.api.tools.framework.model.Model;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 
 /** A subclass of ModelToViewTransformer which translates model into API tests in Java. */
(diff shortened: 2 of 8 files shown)


src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -447,9 +447,17 @@ public String getAndSavePagedResponseTypeName(ModelTypeTable typeTable, TypeRef
     return getNotImplementedString("SurfaceNamer.getAndSavePagedResponseTypeName");
   }
 
-  /** The test case name for the given method. */
-  public String getTestCaseName(Method method) {
-    return methodName(Name.upperCamel(method.getSimpleName(), "Test"));
+  /**
+   * The test case name for the given method.
+   * Use the given count value to produce unique test case names if there are multiple tests
+   * for each method
```
garrettjonesgoogle: End all sentences in docs with a period

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
```diff
@@ -51,6 +54,24 @@
...
+
+    public String getTestName(SurfaceNamer namer, Method method) {
+      String testedMethodName = method.getSimpleName();
+      Integer count = 1;
+      if (nameCount.containsKey(testedMethodName)) {
+        count = nameCount.get(testedMethodName) + 1;
+      }
+      nameCount.put(testedMethodName, count);
+      return namer.getTestCaseName(method, count);
+    }
```
garrettjonesgoogle: We solved a similar problem slightly differently here: https://github.com/googleapis/toolkit/blob/master/src/main/java/com/google/api/codegen/metacode/InitCodeGenerator.java#L106

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
bjwatson: It's also a good practice to put a blank line between paragraphs, especially since the first is the summary line for the rendered JavaDoc.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
```diff
@@ -147,21 +170,32 @@ private GapicSurfaceTestClassView createTestClassView(SurfaceTransformerContext
...
       MethodTransformerContext methodContext = context.asMethodContext(method);
-      testCaseViews.add(createTestCaseView(methodContext));
+      MethodConfig methodConfig = methodContext.getMethodConfig();
+      if (methodConfig.isFlattening()) {
+        for (List<Field> paramFields : methodConfig.getFlattening().getFlatteningGroups()) {
+          testCaseViews.add(createTestCaseView(methodContext, paramFields, nameTable));
+        }
+      } else {
+        // TODO(shinfan): Add support of non-flattening method
+      }
```
garrettjonesgoogle: I think what you need to do is match or share the logic that determines the list of output methods: https://github.com/googleapis/toolkit/blob/master/src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTransformer.java#L270

src/test/java/com/google/api/codegen/testdata/java_mock_service_impl_library.baseline
```diff
@@ -43,37 +43,38 @@ import com.google.protobuf.Empty;
...
     requests = new ArrayList<>();
-    responses = new ArrayList<>();
+    responses = new LinkedList<>();
   }
 
   public List<GeneratedMessage> getRequests() {
     return requests;
   }
 
   public void setResponses(List<GeneratedMessage> responses) {
```
bjwatson: The code below could be further streamlined by the making this method take varargs, and then using `Lists.newLinkedList(Arrays.asList(responses))`: https://docs.oracle.com/javase/7/docs/api/java/util/Arrays.html#asList(T...)

This would make the code below look cleaner, since a single response is the most common case.

src/test/java/com/google/api/codegen/testdata/java_mock_service_impl_library.baseline
```diff
@@ -297,8 +284,7 @@ public class MockLibraryServiceImpl implements LibraryService  {
   public void updateBookIndex(UpdateBookIndexRequest request,
       StreamObserver<Empty> responseObserver) {
     if (responses != null && !responses.isEmpty()) {
-      Empty response = (Empty) responses.get(0);
-      responses.remove(0);
+      Empty response = (Empty) responses.poll();
```
bjwatson: Is it appropriate to pull a response for a method that does not have a response, like this one?

Also, seems better to use `remove()` for the cases that should have a response, which throws a `NoSuchElementException` if there's nothing in `responses`. That's a lot clearer when it's broken than ending up with a `null` in `responseObserver`.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
```diff
@@ -147,21 +170,32 @@ private GapicSurfaceTestClassView createTestClassView(SurfaceTransformerContext
...
     for (Method method : context.getNonStreamingMethods()) {
       MethodTransformerContext methodContext = context.asMethodContext(method);
-      testCaseViews.add(createTestCaseView(methodContext));
+      MethodConfig methodConfig = methodContext.getMethodConfig();
+      if (methodConfig.isFlattening()) {
+        for (List<Field> paramFields : methodConfig.getFlattening().getFlatteningGroups()) {
+          testCaseViews.add(createTestCaseView(methodContext, paramFields, nameTable));
+        }
+      } else {
+        // TODO(shinfan): Add support of non-flattening method
```
bjwatson: `throw UnsupportedOperationException` or something like that here, so it's clear that this hasn't been implemented yet.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
shinfan: I think algorithm wise my approach is slightly more efficient (hashmap vs while loop) but there is probably no big difference. If you have a strong preference I could address this in a separate PR.

garrettjonesgoogle: My suggestion would be to create a name table that can be used in both spots. Your algorithm is fine, but the SurfaceNamer/Method stuff shouldn't be in the shared code. It's ok to address in a separate PR (but put a TODO here).

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
shinfan: Issue filed: https://github.com/googleapis/toolkit/issues/387
Will address in a separate PR

src/test/java/com/google/api/codegen/testdata/java_mock_service_impl_library.baseline
shinfan: It makes sense to use varargs here. But in order to make the full change I need to update the interface in gax-java so I would prefer to address this in a separate PR. Issue filed in gax: https://github.com/googleapis/gax-java/issues/100

src/test/java/com/google/api/codegen/testdata/java_mock_service_impl_library.baseline
shinfan: Done. As a mock service, I think it would be more appropriate to send the exception back using onError() than throwing exception inline.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
shinfan: I consider this as an addition to the existing functionality so I would prefer to not interrupt the execution even there is a non-flattening method.  Alternatively I could print a warning message here. Do you think it would be sufficient?

src/test/java/com/google/api/codegen/testdata/java_mock_service_impl_library.baseline
bjwatson: SGTM

src/test/java/com/google/api/codegen/testdata/java_mock_service_impl_library.baseline
bjwatson: Depends on whether it's due to a bug in our code (throwing an exception and breaking the flow of execution makes that clear), or a bug in the user test configuration (for which the graceful `onError()` mechanism is appropriate).

I like breaking hard and fast when we have a bug, unless it's a mission critical system like a web app. Even back when I worked in the Linux kernel, I saw that `panic()` was often used whenever data structures were found to be in a surprising state, to ensure that the underlying bug got fixed.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
bjwatson: Yeah, that makes sense. However, I saw @garrettjonesgoogle assert that no one reads warnings. ;)

I generally agree, but I also think that's a cultural problem, albeit one that's plagued most teams I've worked in.

garrettjonesgoogle: So in this particular scenario, our generation would fail if we threw an exception because we have some API methods with no flattening. I think just tracking this on the github issue makes the most sense.

bjwatson: Yeah, I agree with the warning approach here. I was just making social commentary.

src/test/java/com/google/api/codegen/testdata/java_mock_service_impl_library.baseline
shinfan: Yes I agree that we should break hard and fast especially in a test code. But correct me if I am wrong, here the mock service is shared by all the tests so if we throw an exception (and interrupt the execution) here, it would fail all the following tests. Ideally we would like only fail the test that caused the issue, which requires the mock service to continue executing.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
shinfan: Done. Warning message added.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
shinfan: Done. TODO added.

src/test/java/com/google/api/codegen/testdata/java_mock_service_impl_library.baseline
bjwatson: Your logic is correct when the exception is due to an error in the system under test (i.e. the API surface is improperly configured for response testing). If the `responses` queue could be unexpectedly empty due to this kind of user error, then we should be graceful.

OTOH, if the `responses` queue could _only_ be unexpectedly empty due to a bug in the unit test auto-generator (i.e. this `toolkit` code you're working on), then throwing an exception is more appropriate, so that it's obvious that we need to fix the bug rather than an API producer.

shinfan: Hmmm ok I see what you mean. It is hard to predict if the responses queue would be empty due to an user error but in my opinion it maybe relatively rare. It would be more common to be caused by a bug in the generator. With that said, we probably should directly throw an exception immediately in this case.

shinfan:
PTAL

src/test/java/com/google/api/codegen/testdata/java_mock_service_impl_library.baseline
shinfan: Done.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
```diff
@@ -180,6 +181,8 @@ private GapicSurfaceTestClassView createTestClassView(SurfaceTransformerContext
         }
       } else {
         // TODO(shinfan): Add support of non-flattening method
+        System.err.println(
+            "Non-flatenning method test is not supported yet for " + method.getSimpleName());
```
garrettjonesgoogle: spelling (flatenning)

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
garrettjonesgoogle: In TODOs, we should reference issues instead of usernames - add a link to #393

bjwatson: :+1:

src/test/java/com/google/api/codegen/testdata/java_mock_service_impl_library.baseline
bjwatson: Thanks. If we're wrong, then we can re-evaluate how we handle exceptions and/or make the code more robust against known user errors.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
shinfan: oops Done.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
shinfan: Done.

garrettjonesgoogle:
LGTM (let Brian approve too)

bjwatson:
LGTM

PR #388, opened 2016-08-17
Do not validate empty string for optional path template fields.

shinfan:
We should allow empty path for optional path template fields.

This should fix: https://github.com/googleapis/toolkit/issues/389

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
@@ -38,13 +38,11 @@ import com.google.api.tools.framework.model.Field;
 import com.google.api.tools.framework.model.TypeRef;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.Sets;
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
```diff
@@ -282,14 +282,19 @@ private void setStaticLangSyncReturnFields(
...
-
         pathTemplateChecks.add(check.build());
       }
     }
     return pathTemplateChecks;
   }
 
+  private boolean shouldAllowEmpty(MethodTransformerContext context, Field field) {
+    Set<Field> requiredFields = Sets.newHashSet(context.getMethodConfig().getRequiredFields());
+    return !requiredFields.contains(field);
```
garrettjonesgoogle: It seems wasteful to add the fields to a temporary set for a containment check.

shinfan: Done. Switched to use a for loop.

garrettjonesgoogle:
LGTM

PR #392, opened 2016-08-18
Modify NodeJS codegen to fit with gcloud-node design.

jmuk:
- The new style exports a function instead of the class. The
  exported function will take the auth context as a parameter
  and returns builder function.
- Files will be under a sub directory with versioned name,
  so that it can be exposed as a sub package of the API package.
- A few misc changes and style changes made to pass lint
  checker in gcloud.

### DIFF
--- a/src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
+++ b/src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
@@ -87,21 +87,20 @@ public class NodeJSGapicContext extends GapicContext implements NodeJSContext {
   }
 
   /**
-   * The name for the module for this vkit module.
+   * The name for the module for this vkit module. This assumes that the service's
+   * full name will be in the format of 'google.some.apiname.version.ServiceName',
+   * and extracts the 'apiname' and 'version' part and combine them to lower-camelcased
+   * style (like pubsubV1).
    */
   public String getModuleName(Interface service) {
     List<String> names = Splitter.on(".").splitToList(service.getFullName());
-    if (names.size() >= 3) {
-      return lowerUnderscoreToLowerCamel(
-          names.get(names.size() - 3) + "_" + names.get(names.size() - 2));
-    } else if (names.size() >= 2) {
-      return names.get(names.size() - 2);
-    }
-    return upperCamelToLowerCamel(service.getSimpleName());
+    return names.get(names.size() - 3) + lowerUnderscoreToUpperCamel(names.get(names.size() - 2));
   }
 
   /**
-   * Returns the major version part in the API namespace.
+   * Returns the major version part in the API namespace. This assumes that the service's
+   * full name will be in the format of 'google.some.apiname.version.ServiceName', and
+   * extracts the 'version' part.
    */
   public String getApiVersion(Interface service) {
     List<String> names = Splitter.on(".").splitToList(service.getFullName());
--- a/src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
+++ b/src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
@@ -74,8 +74,16 @@ public class NodeJSGapicContext extends GapicContext implements NodeJSContext {
   }
 
   public boolean isGcloud() {
-    String packageName = getApiConfig().getPackageName();
-    return !Strings.isNullOrEmpty(packageName) && packageName.startsWith("@google-cloud/");
+    return NodeJSUtils.isGcloud(getApiConfig());
+  }
+
+  /**
+   * The namespace (full package name) for the service.
+   */
+  public String getNamespace(Interface service) {
+    String fullName = service.getFullName();
+    int slash = fullName.lastIndexOf('.');
+    return fullName.substring(0, slash);
   }
 
   /**
@@ -104,6 +112,7 @@ public class NodeJSGapicContext extends GapicContext implements NodeJSContext {
     String fieldName = wrapIfKeywordOrBuiltIn(lowerUnderscoreToLowerCamel(field.getSimpleName()));
     if (isOptional) {
       fieldName = "otherArgs." + fieldName;
+      commentType = commentType + "=";
     }
     return fieldComment(
         String.format("@param {%s} %s", commentType, fieldName), paramComment, field);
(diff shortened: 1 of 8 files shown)


jmuk:
The generated result is reviewed at https://github.com/GoogleCloudPlatform/gcloud-node/pull/1476 and probably some details will change later.

jmuk:
It seems the auto-generated code would be fine. @geigerj please take a look (@swcloud is added FYI)

src/main/resources/com/google/api/codegen/nodejs/main.snip
```diff
@@ -131,118 +172,74 @@
...
      *   The default timeout, in seconds, for calls made through this client.
-     * @@param {number} opts.appName
+     * @@param {number=} opts.appName
      *   The codename of the calling service.
-     * @@param {String} opts.appVersion
+     * @@param {String=} opts.appVersion
      *   The version of the calling service.
      */
-    function {@serviceName}(opts) {
+    function {@serviceName}(gaxGrpc, grpcClient, opts) {
```
geigerj: Please add `gaxGrpc` and `grpcClient` to the documentation

src/test/java/com/google/api/codegen/testdata/nodejs_main_library.baseline
```diff
@@ -819,25 +800,40 @@ LibraryServiceApi.prototype.updateBookIndex = function updateBookIndex() {
...
   var req = {
-    'resource': args.resource,
-    'tag': args.tag
+    resource: args.resource,
+    tag: args.tag
   };
   return this._addTag(req, args.options, args.callback);
 };
+
+module.exports = function build(gaxGrpc) {
```
geigerj: Consider adding documentation to the builder function, too?

src/test/java/com/google/api/codegen/testdata/nodejs_main_library.baseline
```diff
@@ -819,25 +800,40 @@ LibraryServiceApi.prototype.updateBookIndex = function updateBookIndex() {
...
+  var grpcClient = gaxGrpc.load([{
+    root: require('google-proto-files')('..'),
+    file: 'library.proto'
+  }]);
+  var built = grpcClient.google.example.library.v1;
+
+  built.libraryServiceApi = function(opts) {
+    return new LibraryServiceApi(gaxGrpc, grpcClient, opts);
+  };
+  return built;
```
geigerj: So, `built` is the gRPC client, with an additional attribute, which is a builder for the GAX client?

jmuk: True, but I would say it's a union of the builder function and gRPC messages.

src/test/java/com/google/api/codegen/testdata/nodejs_main_library.baseline
jmuk: Well, the documentation might not be helpful in this case. This function is just a part of module initialization. Instead of that, the module index.js file should have the documentation of how the module should be initialized and how the builder function should be used.

src/main/resources/com/google/api/codegen/nodejs/main.snip
jmuk: Documentation for those parameters are not helpful actually. Moved the existing parameter documentation to the actual builder function.

jmuk:
Modified the code a bit. Please take another look.

Here are the key points:
The API class is now not exposed and users are expected to create an instance through a builder function, like this.
```javascript
var pubsubV1 = require('@google-cloud/pubsub').v1();
var publisher = pubsubV1.publisherApi();
publisher.createTopic(...);
```

So now, `publisherApi` (not a capitalized `PublisherApi`) is a function and it actually takes the all optional parameters. Thus,
- define a Builder class, and let this builder function (like `publisherApi`) to be a method of this class. This way, JSDoc will pick up the builder function in the generated HTML.
- move the documentation of parameters to the builder function. The parameters for the constructor is intentionally blank, because it's not expected to be built directly.
- add an example clause in the constructor document of how to create an instance.

src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
```diff
@@ -74,8 +74,38 @@ public String grpcClientName(Interface service) {
...
+
+  /**
+   * The name for the module for this vkit module.
+   */
+  public String getModuleName(Interface service) {
+    List<String> names = Splitter.on(".").splitToList(service.getFullName());
+    if (names.size() >= 3) {
+      return lowerUnderscoreToLowerCamel(
+          names.get(names.size() - 3) + "_" + names.get(names.size() - 2));
+    } else if (names.size() >= 2) {
```
geigerj: I'm a little concerned about the logic here. We're baking in the assumption that the service name looks like `{package}.v{X}.Service`. What would it mean here for `names.size() == 2` or `names.size() == 1`? (In fact, in the last case, `getApiVersion` would fail.)

IMO, it's probably better to just deal with the `names.size() >= 3` case and throw an exception otherwise.

src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
```diff
@@ -74,8 +74,38 @@ public String grpcClientName(Interface service) {
...
+      return names.get(names.size() - 2);
+    }
+    return upperCamelToLowerCamel(service.getSimpleName());
+  }
+
+  /**
+   * Returns the major version part in the API namespace.
+   */
+  public String getApiVersion(Interface service) {
+    List<String> names = Splitter.on(".").splitToList(service.getFullName());
```
geigerj: nit: you can just use `lastIndexOf` here to avoid constructing the list of strings

geigerj:
Two nits, otherwise LGTM

src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
jmuk: This wants to extract the second to the last part, so I will have to invoke `lastIndexOf` twice, and `Splitter` would be simpler here.

src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
jmuk: Okay, now the code simply assumes `names.size() >= 3` so it will fail otherwise.

geigerj:
LGTM if build passes

PR #396, opened 2016-08-18
Supporting mixins

garrettjonesgoogle:


src/main/java/com/google/api/codegen/go/GoGapicContext.java
```diff
@@ -128,7 +129,12 @@ public String getClientPrefix(Interface service) {
     String name = getReducedServiceName(service);
     // If there's only one service, or the service name matches the package name, don't prefix with
     // the service name.
-    if (getModel().getSymbolTable().getInterfaces().size() == 1 || name.equals(getPackageName())) {
+    int interfaceCount = 0;
+    for (@SuppressWarnings("unused")
```
michaelbausor: Maybe use Iterables.size()?

garrettjonesgoogle: Good idea, will do.

src/main/java/com/google/api/codegen/py/PythonImportHandler.java
```diff
@@ -60,7 +60,7 @@ public PythonImportHandler(Interface service) {
     // Add method request-type imports.
     for (Method method : service.getMethods()) {
       addImport(
-          method.getFile(),
+          method.getInputMessage().getFile(),
```
michaelbausor: Why is this required? Is it just because of the updated api compiler version?

garrettjonesgoogle: The original is actually incorrect, but the problem didn't surface until I added mixin support. Basically, with mixins, the file declaring the input message type can be distinct from the file declaring the RPC; the type here needs to match the file of the input message type.

garrettjonesgoogle:
I have addressed the feedback so far.

src/main/java/com/google/api/codegen/config/ConfigGeneratorApi.java
```diff
@@ -15,9 +15,9 @@
 package com.google.api.codegen.config;
 
 import com.google.api.client.util.Lists;
-import com.google.api.codegen.CodegenSetup;
 import com.google.api.codegen.ConfigProto;
 import com.google.api.tools.framework.aspects.http.model.HttpAttribute;
+import com.google.api.tools.framework.aspects.mixin.model.MixinAttribute;
```
michaelbausor: Remove unused imports

garrettjonesgoogle: I blame Eclipse. Regardless, I have fixed it now.

src/main/java/com/google/api/codegen/go/GoGapicContext.java
```diff
@@ -544,7 +547,7 @@ public boolean hasEmptyRetryCodes(Interface service, MethodConfig methodConfig)
    * Returns true if the API has a page streaming method.
    */
   public boolean hasPageStreamingMethod() {
-    for (Interface service : getModel().getSymbolTable().getInterfaces()) {
+    for (Interface service : new InterfaceView().getElementIterable(getModel())) {
```
michaelbausor: It would be nice if the `new InterfaceView().getElementIterable(model)` pattern could be a method somewhere, but I'm not sure if there is anywhere sensible...

garrettjonesgoogle: I couldn't think of anything that would be a significant improvement over what I have here.

michaelbausor:
LGTM

PR #397, opened 2016-08-18
Add code coverage analysis to toolkit.

bjwatson:
Following the same recipe as was done for gax-java.


garrettjonesgoogle:
Nice, we're already at 80% :-) LGTM.

bjwatson:
@garrettjonesgoogle I guess you're more optimistic than me. My first thought was "only 80%?", but you're right that it's not bad for not being monitored until now.

PR #398, opened 2016-08-19
Update discovery sample gen across languages

saicheems:
Earlier discovery sample gen implementations diverged in features and
style due to separate implementations. This change brings a consistent
feature set and style to discovery samples across all languages,
including all known changes required for the samples themselves, except
for media upload segments.

Missing features added:
* page iteration in PHP
* page-element iteration in Node.js
* map-element iteration in all except Java, Python, and PHP
* placeholder pretty-printing of output in all except Python
* informative variable naming/typing in Node.js, PHP, Go, and C#
* distinguishing comments for patch methods

Features unified across language samples:
* naming and typing inference for variables (including backend logic,
  except for C#)
* variable naming for non-inferred names, where appropriate
* parameter description formatting, including cases of missing
  descriptions
* content and formatting of code comments, where appropriate (except for
  unique call comment in Ruby, currently needed to pass compile check)
* spacing between sample elements (except for exclusion of multiple
  blank lines in Go, for consistency with gofmt)
* clarification of strings vs. identifiers in code comments
* clarification of async method use cases in C#

Features unified across language templates (except for C#):
* logical organization of template and context code
* naming of template entities

Note that ORDERING of template elements will be unified in a following
change to avoid obscuring changes made here (for review).

Due to major differences in the implementation of sample gen for C#,
refactoring of template and context logic for C# will be deferred. This
may result in minor changes to the presentation of default string values
appearing in samples, but should otherwise leave sample output
unchanged.

PR #405, opened 2016-08-22
Create Python wrappers from Proto enums

### LINKED ISSUE #275: Enum documentation is incorrect
In Python, the message name is not part of the enum's namespace. [See this example from the library baseline test.](https://github.com/googleapis/toolkit/blob/06514fa0171e1751d59677dabaa853e00ec5c1b8/src/test/java/com/google/api/codegen/testdata/python_message_library.baseline#L569)

**Observed:**
In the baseline test, we refer to the type:

```
:class:`google.example.library.v1.library_pb2.Book.Rating`
```
There is no documentation present for `Rating` itself.

**Expected:**
~~The actual, expected type is~~

~~`:class:`google.example.library.v1.library_pb2.Rating`~~

~~Also expect to see documentation detailing the possible values of `Rating`.~~

geigerj:
This change creates convenient aliases for protocol buffer enums to allow us to document enums succinctly and correctly.

Note that generation is done into a single file per model (not service) to avoid duplication.

Fixes #275

### DIFF
--- a/src/main/java/com/google/api/codegen/py/PythonGapicContext.java
+++ b/src/main/java/com/google/api/codegen/py/PythonGapicContext.java
@@ -151,7 +151,7 @@ public class PythonGapicContext extends GapicContext {
       case TYPE_MESSAGE:
         return ":class:`" + importHandler.elementPath(type.getMessageType(), true) + "`";
       case TYPE_ENUM:
-        return "enum from :class:`"
+        return "enum :class:`"
             + getApiConfig().getPackageName()
             + ".enums."
             + pythonCommon.wrapIfKeywordOrBuiltIn(type.getEnumType().getSimpleName())
--- a/src/main/java/com/google/api/codegen/py/PythonGapicContext.java
+++ b/src/main/java/com/google/api/codegen/py/PythonGapicContext.java
@@ -151,7 +151,11 @@ public class PythonGapicContext extends GapicContext {
       case TYPE_MESSAGE:
         return ":class:`" + importHandler.elementPath(type.getMessageType(), true) + "`";
       case TYPE_ENUM:
-        return ":class:`" + importHandler.elementPath(type.getEnumType(), true) + "`";
+        return "enum from :class:`"
+            + getApiConfig().getPackageName()
+            + ".enums."
+            + pythonCommon.wrapIfKeywordOrBuiltIn(type.getEnumType().getSimpleName())
+            + "`";
       default:
         if (type.isPrimitive()) {
           return PRIMITIVE_TYPE_NAMES.get(type.getKind());
@@ -162,8 +166,8 @@ public class PythonGapicContext extends GapicContext {
   }
 
   /** Returns a comment string for field, consisting of type information and proto comment. */
-  private String fieldComment(String name, String cardinality, String paramComment) {
-    String comment = String.format("  %s (%s)", name, cardinality);
+  private String fieldComment(String name, String type, String paramComment) {
+    String comment = String.format("  %s (%s)", name, type);
     if (!Strings.isNullOrEmpty(paramComment)) {
       if (paramComment.charAt(paramComment.length() - 1) == '\n') {
         paramComment = paramComment.substring(0, paramComment.length() - 1);
--- a/src/main/java/com/google/api/codegen/py/PythonGapicContext.java
+++ b/src/main/java/com/google/api/codegen/py/PythonGapicContext.java
@@ -20,6 +20,7 @@ import com.google.api.codegen.InterfaceConfig;
 import com.google.api.codegen.MethodConfig;
 import com.google.api.tools.framework.aspects.documentation.model.DocumentationUtil;
 import com.google.api.tools.framework.aspects.documentation.model.ElementDocumentationAttribute;
+import com.google.api.tools.framework.model.EnumType;
 import com.google.api.tools.framework.model.EnumValue;
 import com.google.api.tools.framework.model.Field;
 import com.google.api.tools.framework.model.Interface;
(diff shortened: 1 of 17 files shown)


src/test/java/com/google/api/codegen/testdata/python_doc_message_library.baseline
```diff
@@ -623,7 +623,7 @@ class Book(object):
       author (string): The name of the book author.
       title (string): The title of the book.
       read (bool): Value indicating whether the book has been read.
-      rating (:class:`google.example.library.v1.library_pb2.Book.Rating`): For testing enums.
+      rating (enum from :class:`google.example.library.v1.enums.Rating`): For testing enums.
```
geigerj: @bjwatson Not sure what the right wording here is -- does this seem sufficient? It's not quite accurate to say that it's an *instance* of `google.example.library.v1.enums.Rating`, since `Rating` is really a container for the enum values.

geigerj:
Tests pass locally; currently debugging Travis failure.

geigerj:
Looks like the failures were caused by Travis doing a bad rebase on top of #398, should be fixed now. I'm going to leave the commit to increase gradle's verbosity in `.travis.yml` because its difficult to debug these failures otherwise.

geigerj:
@jmuk  @tbetbetbe Adding some extra folks to this review in case @bjwatson needs to take ownership of this PR while I am OOO.

src/test/java/com/google/api/codegen/testdata/library_gapic.yaml
```diff
@@ -4,7 +4,7 @@ language_settings:
   java:
     package_name: com.google.gcloud.pubsub.spi
   python:
-    package_name: library
+    package_name: google.example.library.v1
```
jmuk: I am wondering what this means. Can you elaborate it?

Naming of 'google.example.library.v1' might be problematic, because it is same as the namespace for gRPC messages.  In many examples, it is not -- like `google.pubsub.v1` grpc vs `google.cloud.pubsub.v1` gapic, and I am still not sure where you intended to place the `enums.py` file.  Might be good to keep `library`, or use `google.cloud.example.library.v1`, so you can find potential filename bugs much earlier.

geigerj: Ah, there wasn't much thought put into this change, I just wanted the test code to look a little more realistic because previously the package strings in the samples were unrealistically short.

It is probably more accurate to change this to `google.cloud.example.library.v1` given our current practice in reality.

bjwatson: I agree with making this more realistic. BTW, the actual path would be something like `google.cloud.gapic.library.v1`.

src/test/java/com/google/api/codegen/testdata/python_doc_message_library.baseline
bjwatson: I would just say "enum" rather than "enum for". That parallels the Py34 documentation for official enums: https://docs.python.org/3.4/library/enum.html#enum.Enum

geigerj: Done.

Thanks for the link! I was actually trying to figure out what the nomenclature was when I wrote "enum from" -- I was confused by the fact that an "enum member" is not exactly an "instance" of an enum. Good to know what the precedent is here.

src/main/resources/com/google/api/codegen/py/enum.snip
```diff
@@ -0,0 +1,84 @@
...
+    @# limitations under the License.
+@end
+
+@private moduleDocstring()
+    """Wrappers for protocol buffer enum types."""
+@end
+
+@private shouldGenerateEnum(element)
+    @if element.isEnum
+        @if element.getEnumType.isReachable
```
jmuk: Why not `and(element.isEnum, element.getEnumType.isReachable)`?  Guessing arguments would be evaluated prior to `and` so that causing a NullPointerException or something?  If so, it's better to leave a comment about the situation.

geigerj: Your analysis is right. I added a comment.

geigerj:
@bjwatson My previous comment was incorrect. The samples currently generated are also wrong (for example, [this will not run](https://github.com/geigerj/toolkit/blob/c24d904c0aa21374021c9492cf380fd449dcb630/src/test/java/com/google/api/codegen/testdata/python_doc_main_library.baseline#L802).) Instead of splitting that off like I said earlier, I'm looking into making the change now.

bjwatson:
Thanks, @geigerj

geigerj:
The method samples now reference enums correctly. The enum import is missing if the enum is not a top-level field due to #496.

@bjwatson If you're OK with it, I'd like to keep #496 separate to avoid further expanding the scope of this PR. Depedning on how we want to prioritize it, I can pick it up #496 immediately after this is merged due to the fact that it leads to a missing import statement in the method samples.

src/main/java/com/google/api/codegen/py/PythonGapicContext.java
```diff
@@ -286,7 +301,19 @@ private String returnTypeComment(
...
 
-    return Splitter.on("\n").splitToList(contentBuilder.toString());
+    return splitToLines(contentBuilder.toString());
+  }
+
+  public List<String> enumValueComment(EnumValue value) {
+    String description = "";
+    if (value.hasAttribute(ElementDocumentationAttribute.KEY)) {
+      String sphinxified = getSphinxifiedScopedDescription(value);
+      description = sphinxified.replaceAll("\\s*\\n\\s*", "\n");
```
bjwatson: NIT: This is duplicated above. It should be broken out into a private method, but I'm not going to block 0.10.0 for this.

geigerj:
Good news -- post rebase allows the enum import in the method samples to work, so that's fixed now. I'll leave #496 open until we verify for all cases (subfields/optional fields). So I have no lingering concerns with this PR.

PR #406, opened 2016-08-22
Check response fields in Java automate tests.

shinfan:
In this PR:
- Populate fields of mock response object in test:
  - Primitive fields
  - Resources fields with one mock resource element
  - Implement TestValueGenerator
      - Use TestValueGenerator to generate deterministic unique values (except for booleans which cannot be unique)

- Refactor InitCodeGenerator to take InitCodeGeneratorContext which allows:
  - Reuse SymbolTable to generate new symbols.
  - Allows unique symbols between two generate calls by reusing the symbol table object.

### DIFF
--- a/src/main/java/com/google/api/codegen/metacode/InitCodeGeneratorContext.java
+++ b/src/main/java/com/google/api/codegen/metacode/InitCodeGeneratorContext.java
@@ -1,0 +1,64 @@
+/* Copyright 2016 Google Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen.metacode;
+
+import com.google.api.codegen.util.SymbolTable;
+import com.google.api.codegen.util.testing.TestValueGenerator;
+import com.google.api.tools.framework.model.Method;
+import com.google.auto.value.AutoValue;
+
+import java.util.Map;
+
+import javax.annotation.Nullable;
+
+/**
+ * Container class used by InitCodeGenerator.
+ *
+ * <p>Contains context information related to init code generation stage which includes: 1.
+ * SymbolTable to produce unique variable names. 2. TestValueGenerator to populate unique values. 3.
+ * Initialized fields structure configuration map. 4. The method to generate.
+ */
+@AutoValue
+public abstract class InitCodeGeneratorContext {
+  @Nullable
+  public abstract TestValueGenerator valueGenerator();
+
+  public abstract SymbolTable symbolTable();
+
+  public abstract Map<String, Object> initStructure();
+
+  public abstract Method method();
+
+  public boolean shouldGenerateTestValue() {
+    return valueGenerator() != null;
+  }
+
+  public static Builder newBuilder() {
+    return new AutoValue_InitCodeGeneratorContext.Builder();
+  }
+
+  @AutoValue.Builder
+  public abstract static class Builder {
+    public abstract Builder valueGenerator(TestValueGenerator val);
+
+    public abstract Builder symbolTable(SymbolTable val);
+
+    public abstract Builder initStructure(Map<String, Object> val);
+
+    public abstract Builder method(Method val);
+
+    public abstract InitCodeGeneratorContext build();
+  }
+}
--- a/src/main/java/com/google/api/codegen/metacode/InitCodeGeneratorContext.java
+++ b/src/main/java/com/google/api/codegen/metacode/InitCodeGeneratorContext.java
@@ -24,11 +24,13 @@ import java.util.Map;
 import javax.annotation.Nullable;
 
 /**
- * Container class used by InitCodeGenerator.
+ * Context class used by InitCodeGenerator.
  *
- * <p>Contains context information related to init code generation stage which includes: 1.
- * SymbolTable to produce unique variable names. 2. TestValueGenerator to populate unique values. 3.
- * Initialized fields structure configuration map. 4. The method to generate.
+ * Contains context information related to init code generation stage which includes:
+ * 1. SymbolTable to produce unique variable names.
+ * 2. TestValueGenerator to populate unique values.
+ * 3. Initialized fields structure configuration map.
+ * 4. The method to generate.
  */
 @AutoValue
 public abstract class InitCodeGeneratorContext {
--- a/src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
@@ -30,7 +30,7 @@ import com.google.api.codegen.metacode.SimpleInitCodeLine;
 import com.google.api.codegen.metacode.StructureInitCodeLine;
 import com.google.api.codegen.util.Name;
 import com.google.api.codegen.util.SymbolTable;
-import com.google.api.codegen.util.testing.JavaTestValueGenerator;
+import com.google.api.codegen.util.testing.TestValueGenerator;
 import com.google.api.codegen.viewmodel.FieldSettingView;
 import com.google.api.codegen.viewmodel.FormattedInitValueView;
 import com.google.api.codegen.viewmodel.InitCodeLineView;
@@ -60,7 +60,7 @@ public class InitCodeTransformer {
       MethodTransformerContext context,
       Iterable<Field> fields,
       SymbolTable table,
-      JavaTestValueGenerator valueGenerator) {
+      TestValueGenerator valueGenerator) {
     Map<String, Object> initFieldStructure = createSampleInitFieldStructure(context);
     InitCodeGeneratorContext initCodeContext =
         InitCodeGeneratorContext.newBuilder()
@@ -76,7 +76,7 @@ public class InitCodeTransformer {
   }
 
   public InitCodeView generateRequestObjectInitCode(
-      MethodTransformerContext context, SymbolTable table, JavaTestValueGenerator valueGenerator) {
+      MethodTransformerContext context, SymbolTable table, TestValueGenerator valueGenerator) {
     Map<String, Object> initFieldStructure = createSampleInitFieldStructure(context);
     InitCodeGeneratorContext initCodeContext =
         InitCodeGeneratorContext.newBuilder()
@@ -92,7 +92,7 @@ public class InitCodeTransformer {
   }
 
   public InitCodeView generateMockResponseObjectInitCode(
-      MethodTransformerContext context, SymbolTable table, JavaTestValueGenerator valueGenerator) {
+      MethodTransformerContext context, SymbolTable table, TestValueGenerator valueGenerator) {
     Map<String, Object> initFieldStructure = createMockResponseInitFieldStructure(context);
     InitCodeGeneratorContext initCodeContext =
         InitCodeGeneratorContext.newBuilder()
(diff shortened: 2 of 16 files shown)


src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -476,12 +470,8 @@ public String getMockGrpcServiceImplName(Interface service) {
...
-  /** The method name of getter function call for the given name */
-  public String getGetFunctionCallName(Name name, TypeRef type) {
-    if (type.isRepeated() && !type.isMap()) {
-      return methodName(Name.from("get").join(name).join("list"));
-    } else {
-      return methodName(Name.from("get").join(name));
-    }
+  /** The method call string which combines the given the callee and the called method name */
+  public String methodCall(String callee, String method, List<String> params) {
+    return callee + "." + method + "(" + Joiner.on(", ").join(params) + ")";
```
garrettjonesgoogle: This should not be here, it is language-specific. Only names should be rendered here, and syntax should be rendered in the template.

src/main/java/com/google/api/codegen/util/testing/TestValueGenerator.java
```diff
@@ -0,0 +1,45 @@
...
+        // If the value already exists regenerate using a deterministically different identifier.
+        identifier = identifier.join("1");
+        value = generateValue(type, identifier);
+      }
+      valueTable.put(identifier, generateValue(type, identifier));
+    }
+    return valueTable.get(identifier);
+  }
+
+  protected abstract String generateValue(TypeRef typeRef, Name identifier);
```
garrettjonesgoogle: I'd prefer composition here instead.

garrettjonesgoogle:
I did a superficial pass now; I'll do a deeper pass tomorrow.

shinfan:
I think https://github.com/googleapis/toolkit/pull/407/ accidentally deleted some of my early changes so I need to rebase and check. I will let you know once it's done. @garrettjonesgoogle

saicheems:
Really sorry about that Shin

shinfan:
No worries @saicheems

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
shinfan: Removed.

src/main/java/com/google/api/codegen/util/testing/TestValueGenerator.java
shinfan: I think this is a situation where abstract class can be used.  According to the wikipedia, we should choose composition over inheritance to achieve polymorphic behavior, which is not needed here. The functionality of this generator is rather simple where each language just needs to override the generateValue method.

From the perspective of complexity, the inheritance approach only requires one type of class (TestValueGenerator and its subclasses) but the composition approach requires two (TestValueGenerator, an composition interface and its implementing classes).

shinfan:
I have replied/addressed your comments. PTAL. @garrettjonesgoogle

src/main/java/com/google/api/codegen/util/testing/TestValueGenerator.java
garrettjonesgoogle: When you are using inheritance, you are depending on polymorphic behavior. The polymorphic part is the generateValue call, and you provide 1 form so far for Java, JavaTestValueGenerator. Each new language will provide a new form called through the same method, making it polymorphic. 

Preferring composition over inheritance is not about flexibility in my mind, it is about https://en.wikipedia.org/wiki/Loose_coupling . When you use inheritance, the derived class knows the details of the class it inherits from. When you use composition, the class implementing the interface has no knowledge at all of the class calling that interface. That interface also becomes available to other classes independent of the original code calling that interface. 

This loose coupling makes the system easier to change and evolve over time, because you can mix and match the independent components. As the wikipedia article on composition over inheritance states, "This approach more easily accommodates future requirements changes that would otherwise require a complete restructuring of business-domain classes in the inheritance model." If you use too much inheritance, you don't get pieces, you get a bunch of things tied together. This makes the system more complex to understand, even if it technically has fewer classes. Number of classes is not the end-all metric for determining how complex a system is.

src/main/java/com/google/api/codegen/util/SymbolTable.java
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen.util;
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * A utility class used to get and store unique symbols
```
garrettjonesgoogle: End this sentence with a period.

src/main/java/com/google/api/codegen/util/SymbolTable.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * If a symbol is already used, the table will try to append a index number to its end.
```
garrettjonesgoogle: "a index" -> "an index"; "to its end" -> "onto the end of it".

src/main/java/com/google/api/codegen/util/testing/JavaTestValueGenerator.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * A utility class used by the Java test generator which populates values for primitive fields.
+ */
+public class JavaTestValueGenerator extends TestValueGenerator {
+
+  @Override
+  protected String generateValue(TypeRef typeRef, Name identifier) {
+    String typeName = typeRef.getPrimitiveTypeName();
+    if (typeName == "string") {
```
garrettjonesgoogle: We should be using constants defined in api-compiler somewhere instead of things like "string" here

src/main/java/com/google/api/codegen/util/testing/JavaTestValueGenerator.java
```diff
@@ -0,0 +1,41 @@
...
+    } else if (typeName == "bool") {
+      return identifier.hashCode() % 2 == 0 ? "true" : "false";
+    } else if (typeName.contains("int")) {
+      return Integer.toString(identifier.hashCode());
+    } else if (typeName == "byte") {
+      byte lowByte = (byte) (identifier.hashCode());
+      return Byte.toString(lowByte);
+    } else {
+      return Double.toString(identifier.hashCode() / 10);
+    }
```
garrettjonesgoogle: I would feel better about having else-if for the double type, and having a final else catch-all which throws an exception.

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -52,27 +56,53 @@
  */
 public class InitCodeTransformer {
 
-  public InitCodeView generateInitCode(MethodTransformerContext context, Iterable<Field> fields) {
-    Map<String, Object> initFieldStructure = createInitFieldStructure(context);
+  public InitCodeView generateInitCode(
+      MethodTransformerContext context,
+      Iterable<Field> fields,
+      SymbolTable table,
+      JavaTestValueGenerator valueGenerator) {
```
garrettjonesgoogle: Why does this take a Java-specific generator? This should be a language-independent class.

src/main/java/com/google/api/codegen/metacode/InitCodeGeneratorContext.java
```diff
@@ -0,0 +1,64 @@
...
+import com.google.api.codegen.util.testing.TestValueGenerator;
+import com.google.api.tools.framework.model.Method;
+import com.google.auto.value.AutoValue;
+
+import java.util.Map;
+
+import javax.annotation.Nullable;
+
+/**
+ * Container class used by InitCodeGenerator.
```
garrettjonesgoogle: "Container" -> "Context"

src/main/java/com/google/api/codegen/metacode/InitCodeGeneratorContext.java
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * <p>Contains context information related to init code generation stage which includes: 1.
+ * SymbolTable to produce unique variable names. 2. TestValueGenerator to populate unique values. 3.
+ * Initialized fields structure configuration map. 4. The method to generate.
```
garrettjonesgoogle: For readability, put each number on its own line.

src/main/resources/com/google/api/codegen/java/test.snip
```diff
@@ -55,14 +55,23 @@
...
-    @join assert : test.asserts
-      Assert.assertEquals(actualRequest.{@assert.actualValueGetter}(), {@assert.expectedValueIdentifier});
+    @join assert : test.requestAsserts
+      Assert.assertEquals({@assert.expectedValueIdentifier}, \
+        actualRequest.{@assert.actualValueIdentifier}());
+    @end
+
+    @join assert : test.responseAsserts
+      Assert.assertEquals(expectedResponse.{@assert.expectedValueIdentifier}(), \
+        actualResponse.{@assert.actualValueIdentifier}());
```
garrettjonesgoogle: It's confusing for the names to be xIdentifier when they are actually get function call names.

src/test/java/com/google/api/codegen/testdata/java_test_library.baseline
```diff
@@ -109,19 +129,42 @@ public class LibraryServiceTest {
     Assert.assertEquals(1, actualRequests.size());
     CreateShelfRequest actualRequest = (CreateShelfRequest)actualRequests.get(0);
 
-    Assert.assertEquals(actualRequest.getShelf(), shelf);
+    Assert.assertEquals(shelf, actualRequest.getShelf());
+
+    Assert.assertEquals(expectedResponse.getName(), actualResponse.getName());
+    Assert.assertEquals(expectedResponse.getTheme(), actualResponse.getTheme());
+    Assert.assertEquals(expectedResponse.getInternalTheme(), actualResponse.getInternalTheme());
```
garrettjonesgoogle: The response should have its own equals() method, so I think you should be able to just compare expectedResponse & actualResponse.

src/main/java/com/google/api/codegen/util/testing/TestValueGenerator.java
```diff
@@ -25,21 +25,24 @@
  * <p>The populated values will be unique (except for boolean values) and deterministic based on the
  * populated fields and the sequence.
  */
-public abstract class TestValueGenerator {
-  protected HashMap<Name, String> valueTable = new HashMap<>();
+public class TestValueGenerator {
+  private HashMap<Name, String> valueTable = new HashMap<>();
+  private ValueProducer producer;
```
garrettjonesgoogle: make members final that don't need to be re-assigned

src/test/java/com/google/api/codegen/testdata/java_test_library.baseline
```diff
@@ -130,10 +130,6 @@ public class LibraryServiceTest {
     CreateShelfRequest actualRequest = (CreateShelfRequest)actualRequests.get(0);
 
     Assert.assertEquals(shelf, actualRequest.getShelf());
-
-    Assert.assertEquals(expectedResponse.getName(), actualResponse.getName());
-    Assert.assertEquals(expectedResponse.getTheme(), actualResponse.getTheme());
-    Assert.assertEquals(expectedResponse.getInternalTheme(), actualResponse.getInternalTheme());
   }
```
garrettjonesgoogle: You no longer have any check on the response...

src/main/java/com/google/api/codegen/util/testing/TestValueGenerator.java
shinfan: Fair enough. I agree that this is polymorphic behavior but it is a very simple one. I was thinking of using fewer classes since this ValueGenerator is a simple utility class so I was not very concerned about the decoupling and future changes. As you said, if the composition makes the code easier to understand and the number of classes is not an important matrix then composition maybe a better approach here.

src/main/java/com/google/api/codegen/util/SymbolTable.java
shinfan: Done

src/main/java/com/google/api/codegen/util/SymbolTable.java
shinfan: Done

src/main/java/com/google/api/codegen/util/testing/JavaTestValueGenerator.java
shinfan: Done. Switched to use Type instead of String for most cases. For integer and fixed types, a string match is much easier since those types contain lots of Type enums such as int32 and int64.

src/main/java/com/google/api/codegen/util/testing/JavaTestValueGenerator.java
shinfan: Done.

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
shinfan: Done

src/main/java/com/google/api/codegen/metacode/InitCodeGeneratorContext.java
shinfan: Done

src/main/java/com/google/api/codegen/metacode/InitCodeGeneratorContext.java
shinfan: Done.

src/main/resources/com/google/api/codegen/java/test.snip
shinfan: Yes, sometimes the actual value needs to be an variable name but sometimes it could be a getter method name. I am trying to make these assert views as general as possible so they could fit for different usages. I changed the name to "actual" and "expected" so it does not indicate if it is a var or a method. Does it sound okay to you?

src/test/java/com/google/api/codegen/testdata/java_test_library.baseline
shinfan: Done. Removed unnecessary checks.

src/main/resources/com/google/api/codegen/java/test.snip
garrettjonesgoogle: C# has different casing for vars and methods, so they need to be distinguished. (lower camel for vars, upper camel for methods)

src/test/java/com/google/api/codegen/testdata/java_test_library.baseline
garrettjonesgoogle: My mistake, it was always there, just further up.

src/main/java/com/google/api/codegen/util/testing/TestValueGenerator.java
shinfan: Done.

src/main/resources/com/google/api/codegen/java/test.snip
shinfan: Done.

shinfan:
PTAL

garrettjonesgoogle:
LGTM

PR #409, opened 2016-08-23
Java,PHP: Adding reroute_to_grpc_interface for Pubsub IAM hack

garrettjonesgoogle:
See https://github.com/googleapis/googleapis/pull/97 for how Pub/Sub will be configured once all languages have this hack implemented.


garrettjonesgoogle:
Travis failure fixed. PTAL.

garrettjonesgoogle:
This has now been rewritten to not require any changes to pubsub.proto or pubsub.yaml. PTAL.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
```diff
@@ -78,6 +85,27 @@ public JavaGapicSurfaceTestTransformer(GapicCodePathMapper javaPathMapper) {
     return views;
   }
 
+  public Iterable<Interface> getGrpcInterfacesToMock(Model model, ApiConfig apiConfig) {
```
shinfan: Should this be public or private? If this is public method some comments would be helpful.

garrettjonesgoogle: It should be private. Silly eclipse, automatically assuming public. I'll fix it.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
```diff
@@ -78,6 +85,27 @@ public JavaGapicSurfaceTestTransformer(GapicCodePathMapper javaPathMapper) {
...
+      for (MethodConfig methodConfig : interfaceConfig.getMethodConfigs()) {
+        String reroute = methodConfig.getRerouteToGrpcInterface();
+        if (!Strings.isNullOrEmpty(reroute)) {
+          Interface targetInterface = model.getSymbolTable().lookupInterface(reroute);
+          interfaces.put(reroute, targetInterface);
+        }
+      }
+    }
+
+    return interfaces.values();
```
shinfan: Shall we make sure that the returned elements are unique? This may contain duplicate values.

garrettjonesgoogle: That won't happen if my logic is correct. I put Interface instances into the interfaces map keyed by the full name of the interface. So, if an interface is encountered more than once, it will occur in the result only once.

src/main/java/com/google/api/codegen/RerouteLogic.java
```diff
@@ -0,0 +1,38 @@
...
+
+import com.google.api.client.repackaged.com.google.common.base.Strings;
+import com.google.api.tools.framework.model.Interface;
+
+public class RerouteLogic {
+  /**
+   * If rerouteToGrpcInterface is set, then looks up that interface and returns it, otherwise
+   * returns the value of defaultInterface.
+   */
+  public static Interface getTargetInterface(
```
shinfan: I think we can put this into InterfaceConfig as a static method so we do not have to create a new class.

src/main/java/com/google/api/codegen/GapicContext.java
```diff
@@ -123,4 +123,20 @@ public CollectionConfig getCollectionConfig(Interface service, String entityName
     }
     return simples;
   }
+
+  /**
+   * Returns a list of simple RPC methods, taking into account GRPC interface rerouting.
+   * TODO replace getNonStreamingMethods with this when all languages are migrated.
```
shinfan: Can you create an issue for this?

src/main/java/com/google/api/codegen/ApiConfig.java
```diff
@@ -112,11 +112,6 @@ private static String getPackageName(ConfigProto configProto) {
    * Returns the InterfaceConfig for the given API interface.
    */
   public InterfaceConfig getInterfaceConfig(Interface iface) {
-    InterfaceConfig interfaceConfig = getInterfaceConfigMap().get(iface.getFullName());
-    if (interfaceConfig == null) {
-      throw new IllegalArgumentException(
-          "no interface config for interface '" + iface.getFullName() + "'");
-    }
-    return interfaceConfig;
+    return getInterfaceConfigMap().get(iface.getFullName());
```
michaelbausor: Why remove the null check? Is getInterfaceConfigMap() now guaranteed to contain iface?

src/main/java/com/google/api/codegen/RerouteLogic.java
garrettjonesgoogle: Sure, I'll do that.

src/main/java/com/google/api/codegen/GapicContext.java
garrettjonesgoogle: This doesn't seem quite like something to file an issue for...

src/main/java/com/google/api/codegen/ApiConfig.java
garrettjonesgoogle: No, it will sometimes *not* contain it, and I want to allow for that.

src/main/java/com/google/api/codegen/GapicContext.java
```diff
@@ -123,4 +123,20 @@ public CollectionConfig getCollectionConfig(Interface service, String entityName
...
+   */
+  public List<Method> getNonStreamingMethodsV2(Interface service) {
+    InterfaceConfig interfaceConfig = getApiConfig().getInterfaceConfig(service);
```
michaelbausor: Given the change in ApiConfig, should you check for null here?

garrettjonesgoogle: Yes, good point. I'll add a check.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
```diff
@@ -104,6 +132,7 @@ private void addTestImports(SurfaceTransformerContext context) {
     typeTable.saveNicknameFor("java.util.Arrays");
     typeTable.saveNicknameFor("java.util.List");
     typeTable.saveNicknameFor("java.util.ArrayList");
+    typeTable.saveNicknameFor("java.util.Arrays");
```
michaelbausor: Duplicate entry

garrettjonesgoogle: Done

garrettjonesgoogle:
Comments addressed, PTAL

shinfan:
LGTM

Issue #410, opened 2016-08-23
[NodeJS] generate documentation of proto messages


jmuk:
In addition to provide the interface of the generated service APIs (and their methods), we will have to generate the documents of proto messages, like Python/Ruby does.

PR #411, opened 2016-08-23
Fix JSDoc styles.

jmuk:
This is a preparation for #410. Some existing comment does not
work well with JSDoc tools.
- If the type is within the same package, they should exist
  in the doc comment (that's the goal of #410). Because they
  can be referred directly in the package, the type name should
  omit the prefix namespace.
- If not, they are in an external package. In that case,
  those things should be noted as '@external' tag and referred
  with external: prefix. Also for the names with dots, they
  should be quoted as "foo.bar".
- The same rule should be applied to some values in GAX library.

### DIFF
--- a/src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
+++ b/src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
@@ -35,6 +35,7 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.protobuf.DescriptorProtos.FieldDescriptorProto.Type;
 
+import java.util.ArrayList;
 import java.util.List;
 
 import javax.annotation.Nullable;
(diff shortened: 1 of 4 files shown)


jmuk:
cc: @jcanizales

jmuk:
@geigerj, can you take a look?

src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
```diff
@@ -260,22 +265,22 @@ private String returnTypeComment(Method method, MethodConfig config) {
 
   /**
    * Return the doccomment for the message.
+   * public List<String> methodDocComment(MessageType msg) {
```
geigerj: Did you mean to delete this?

src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
```diff
@@ -141,13 +143,12 @@ private String fieldParamComment(Field field, String paramComment, boolean isOpt
 
   /**
    * Returns a JSDoc comment string for the field as an attribute of a message.
+   * private String fieldPropertyComment(Field field) {
```
geigerj: Delete this?

geigerj:
Please fix comments, then LGTM

jmuk:
Hmm, it turned out this is not the right thing to do. Let me close this PR. gcloud-node uses a tool called 'dox' to parse doc comments, and the specs are slightly different from jsdoc itself. For example, 'external:"..."' format is not supported there 😞

PR #412, opened 2016-08-23
Fix discoJar task

saicheems:
There's some conflict with the META-INF keys, excluding the directory in
the shadow task fixes it.


saicheems:
disco

saicheems:
I'm adding this to every discovery related change from now on hahaha

vchudnov-g:
LGTM

(but what does META-INF do and why do we need it?)

saicheems:
As best I understand, it's metadata pooled from all the individual jars associated with the project. The stuff I'm excluding is keys that are included from some of the third party jars that, when the jar is executed, cause a security mismatch due to invalid signature. Excluding them all prevents this failure.

PR #413, opened 2016-08-23
Added support for page size field in php

michaelbausor:
Also update the resources directory location

Depends on https://github.com/googleapis/gax-php/pull/39

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/php/PhpSurfaceNamer.java
+++ b/src/main/java/com/google/api/codegen/transformer/php/PhpSurfaceNamer.java
@@ -60,7 +60,7 @@ public class PhpSurfaceNamer extends SurfaceNamer {
 
   @Override
   public String getClientConfigPath(Interface service) {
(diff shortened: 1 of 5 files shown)


michaelbausor:
Fixed build, PTAL

src/main/java/com/google/api/codegen/transformer/php/PhpSurfaceNamer.java
```diff
@@ -60,7 +60,7 @@ public void addPageStreamingDescriptorImports(ModelTypeTable typeTable) {
 
   @Override
   public String getClientConfigPath(Interface service) {
-    return "./resources/"
+    return "/resources/"
```
garrettjonesgoogle: I would prefer this to not have the leading slash so that it doesn't imply an absolute path.

michaelbausor:
Fixed, PTAL

garrettjonesgoogle:
LGTM

PR #414, opened 2016-08-23
Support custom auth configs in discovery

saicheems:


saicheems:
Hey folks, this isn't nearly done, but I want to keep this pull updated with my progress so you can sanity check my work (since I'm unfamiliar with this bit of the codebase).

PTAL and see if this makes sense so far. I'll bug you time to time at different stages.

pongad:
Looks sane so far :smile:

src/main/java/com/google/api/codegen/ApiaryConfig.java
```diff
@@ -230,4 +244,29 @@ public Field getField(Type type, String fieldName) {
   public ListMultimap<String, String> getAuthScopes() {
     return authScopes;
   }
+
+  /*
+   * @return auth type supported by the method.
+   */
+  public AuthType getAuthType(String method) {
```
saicheems: @vchudnov-g PTAL at this method - it determines the auth type to use given a method.

I've opted for an enum instead of a separate class, I think it's a bit simpler.

src/main/java/com/google/api/codegen/ApiaryConfig.java
```diff
@@ -111,6 +111,20 @@
    */
   private String serviceVersion;
 
+  /*
+   * Supported auth types.
+   */
+  public enum AuthType {
+    API_KEY,
+    THREE_LEGGED_OAUTH,
```
vchudnov-g: nit: I would probably put `OAUTH` in the front. And I might abbreviate `THREE_LEGGED`. Something like `OAUTH_3LO`. Though I realize the last `O` is redundant.... Up to you.

src/main/java/com/google/api/codegen/ApiaryConfig.java
vchudnov-g: It might be conceivable that a single method does not list scopes at all though its siblings do (though I can't imagine). If we can guarantee this doesn't happen, we can set this for the entire service by looking at any method once.

Conceivably not every method lists the same scopes, though, so we would still need to disambiguate the acceptable scopes regardless.

src/main/java/com/google/api/codegen/ApiaryConfig.java
```diff
@@ -230,4 +244,29 @@ public Field getField(Type type, String fieldName) {
   public ListMultimap<String, String> getAuthScopes() {
```
vchudnov-g: QQ: Is it customary to return the ListMultimap itself, or is it bett to have getAuthScopes(method) return the list of scopes for that method?

PR #418, opened 2016-08-25
Change Java surface from PageAccessor to PagedListResponse

michaelbausor:
Preliminary PR with updates to surface using PagedListResponse, pending smoke testing of generated surface.

Depends on https://github.com/googleapis/gax-java/pull/105


michaelbausor:
Updated test generation, PTAL

src/main/java/com/google/api/codegen/metacode/InitCodeGenerator.java
```diff
@@ -206,6 +202,12 @@ private InitCodeLine generateCodeInit(
                   + initFieldStructure);
         }
         initValueConfig = initValueConfig.withInitialValue(validatedValue);
+      } else if (typeRef.isPrimitive()
```
shinfan: Is there any reason that you move the check to here?

src/main/java/com/google/api/codegen/util/java/JavaTypeTable.java
```diff
@@ -83,14 +83,24 @@ public NamePath getNamePath(String fullName) {
...
   @Override
-  public TypeName getContainerTypeName(String containerFullName, String elementFullName) {
+  public TypeName getContainerTypeName(String containerFullName, String... elementFullNames) {
     TypeName containerTypeName = getTypeName(containerFullName);
-    TypeName elementTypeName = getTypeName(elementFullName);
+    TypeName[] elementTypeNames = new TypeName[elementFullNames.length];
+    StringBuilder patternBuilder = new StringBuilder("%s<");
+    for (int i = 0; i < elementTypeNames.length; i++) {
+      elementTypeNames[i] = getTypeName(elementFullNames[i]);
+      if (i == 0) {
```
shinfan: It would probably be easier if you use String.join() here

src/main/java/com/google/api/codegen/metacode/InitCodeGenerator.java
michaelbausor: Yes, I moved the check for hasInitialValue() first so that it is possible to set the initial value for the page token in tests, when shouldGenerateTestValue() is true

src/main/java/com/google/api/codegen/transformer/PageStreamingTransformer.java
```diff
@@ -90,6 +90,14 @@ private PageStreamingDescriptorClassView generateDescriptorClass(
 
     desc.requestTokenSetFunction(
         namer.getFieldSetFunctionName(pageStreaming.getRequestTokenField()));
+    desc.requestPageSizeSetFunction(
```
shinfan: It is more readable if you do:

if (pageStreaming.hasPageSizeField()) {
  desc.requestPageSizeSetFunction(namer.getFieldSetFunctionName(pageStreaming.getPageSizeField())
}

src/main/java/com/google/api/codegen/util/java/JavaTypeTable.java
shinfan: String pattern = i == 0 ? "%i" : ",%i";
patternBuilder.append(pattern);

src/test/java/com/google/api/codegen/testdata/java_test_library.baseline
```diff
@@ -637,7 +631,7 @@ public class LibraryServiceTest {
     mockLibraryService.setResponses(expectedResponses);
 
     String formattedName = LibraryServiceApi.formatBookName("[SHELF]", "[BOOK]");
-    String indexName = "indexName746962392";
+    String indexName = "default index";
```
shinfan: We want to distinguished default values for samples from auto-generated values for tests. For example, here we would prefer an randomized unique index name instead of a default index.

michaelbausor: Yes good point. I am not sure about the best way to resolve this. At the moment this value is coming from the sample_code_init_fields in the library_gapic.yaml file. It is showing up now because I changed the ordering of the check InitCodeGenerator. Do you think it is best to undo that change, and find a different way to set the initial value of the page token? Or should I change the test generator to not load from the sample_code_init_fields, and just use the required fields? Note that the previous surface is using "default_key" in the indexMap field, which is from the sample_code_init_fields.

shinfan: I think it makes sense to make initial value override the value generator so we could set values when necessary (next_page_token is an use case). The best way I think is to change the generator to not load from the sample_code_init fields.

michaelbausor: I have implemented this, PTAL. The first commit ("Address PR feedback") addresses this issue, and other changes. In the second commit ("Changed recursive initialization depth to 1") there is a lot of changes to the baseline, as many new fields are initialized.

shinfan: As discussed earlier, we would prefer not to initialize all fields of the response since the marginal benefit of dong this is small. And it is preferable to keep the test short and efficient while the functionality is similar.

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -98,6 +98,27 @@ public InitCodeView generateRequestObjectInitCode(
     return buildInitCodeView(context, initCode);
   }
 
+  public InitCodeView generateTestMethodInitCode(
+      MethodTransformerContext context,
+      Iterable<Field> fields,
+      SymbolTable table,
+      TestValueGenerator valueGenerator) {
+    Map<String, Object> initFieldStructure = createTestRequestInitFieldStructure(context, fields);
```
shinfan: Is it possible to make a shared helper function to reduce the duplicate code between those generate methods?

michaelbausor: Added, PTAL

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -65,62 +65,63 @@ public InitCodeView generateInitCode(
...
-    return buildInitCodeView(context, initCode);
+    return buildInitCodeView(
+        context,
+        generateInitCodeHelper(
+            Lists.newArrayList(fields),
+            table,
+            valueGenerator,
+            createSampleInitFieldStructure(context),
+            Name.from("request"),
+            context.getMethod().getInputType()));
```
garrettjonesgoogle: I think I prefer the builder over the helper function because it makes clearer what all of the inputs are.

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -212,6 +227,42 @@ private GapicSurfaceTestAssertView createAssertView(String expected, String actu
...
+        fieldStrings.add(initString);
+        if (field.getType().isMessage() && depth > 0) {
+          createTestRequestInitFieldStructure(
+              context,
+              fieldStrings,
+              field.getType().getMessageType().getFields(),
+              initString + ".",
+              depth - 1);
+        }
+      }
```
garrettjonesgoogle: I am becoming increasingly troubled by the roundabout nature of this: we are using code to incrementally build string instructions which are used by other code to construct an object tree. Maybe it would be easier to operate directly on the object tree?

src/main/java/com/google/api/codegen/util/java/JavaTypeTable.java
```diff
@@ -83,14 +83,21 @@ public NamePath getNamePath(String fullName) {
...
+  public TypeName getContainerTypeName(String containerFullName, String... elementFullNames) {
     TypeName containerTypeName = getTypeName(containerFullName);
-    TypeName elementTypeName = getTypeName(elementFullName);
+    TypeName[] elementTypeNames = new TypeName[elementFullNames.length];
+    StringBuilder patternBuilder = new StringBuilder("%s<");
+    for (int i = 0; i < elementTypeNames.length; i++) {
+      elementTypeNames[i] = getTypeName(elementFullNames[i]);
+      String pattern = i == 0 ? "%i" : ",%i";
+      patternBuilder.append(pattern);
+    }
```
garrettjonesgoogle: maybe use Joiner.on()?

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
michaelbausor: I am going to think about how to improve the init sample code stuff simpler and easier to use for this case.

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
michaelbausor: I will rethink whether there is a clearer way to provide helpers, otherwise I will revert it.

src/main/java/com/google/api/codegen/util/java/JavaTypeTable.java
michaelbausor: Done.

michaelbausor:
Updated with changes to init sample code, PTAL

src/test/java/com/google/api/codegen/testdata/java_test_library.baseline
michaelbausor: I have removed the excessive initialization, but encountered another problem. If we don't use the sample_code_init_fields at all, then some generated tests will fail. For example, the Publish test will fail because it won't have at least one message with a data field set.

One potential fix would be to use the sample_code_init_fields but strip the values. However, I can see this causing a problem for some future test, when a field needs to be set to a particular value. But maybe this is not likely to happen?

garrettjonesgoogle: What was the problem with using the default values from samples? The motivation for auto-generated values for tests is to not just set everything to blank strings or other overlapping values. If we have information from sample_code_init_fields, we know more, and I think makes sense to use it.

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
```diff
@@ -0,0 +1,406 @@
...
+    for (InitCodeNode child : children) {
+      item.addChild(child);
+    }
+    return item;
+  }
+
+  /*
+   * Constructs a tree of objects to be initialized using the provided context, and returns the root
+   */
+  public static InitCodeNode createSpecItemTree(SpecItemParserContext parser) {
```
garrettjonesgoogle: `context` would be a more appropriate name than `parser`

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
```diff
@@ -0,0 +1,406 @@
...
+    List<InitCodeNode> subTrees = buildSubTrees(parser);
+    InitCodeNode rootTree =
+        createWithChildren(
+            "root",
+            InitCodeLineType.StructureInitLine,
+            subTrees.toArray(new InitCodeNode[subTrees.size()]));
```
garrettjonesgoogle: I might prefer an overload to having to call toArray outside of createWithChildren()

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
```diff
@@ -0,0 +1,406 @@
...
+    rootTree.updateTree(
+        parser.table(), parser.valueGenerator(), parser.rootObjectType(), parser.suggestedName());
+    rootTree.createInitCodeLines();
+    return rootTree;
+  }
+
+  /*
+   * Generates a flattened list from a tree in post-order.
```
garrettjonesgoogle: Isn't this pre-order?

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
```diff
@@ -0,0 +1,406 @@
...
+  }
+
+  private InitCodeNode(String key, InitCodeLineType nodeType, InitValueConfig initValueConfig) {
+    this.key = key;
+    this.lineType = nodeType;
+    this.initValueConfig = initValueConfig;
+    this.children = new LinkedHashMap<>();
+  }
+
+  private void addChild(InitCodeNode newChild) {
```
garrettjonesgoogle: maybe call this `mergeChild`

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
```diff
@@ -0,0 +1,406 @@
...
+    if (context.subTrees() != null) {
+      subTrees.addAll(context.subTrees());
+    }
+    return subTrees;
+  }
+
+  private void updateTree(
+      SymbolTable table, TestValueGenerator valueGenerator, TypeRef type, Name suggestedName) {
+
+    // Remove and re-add children with checked keys, call updateTree() recursively
```
garrettjonesgoogle: but why?

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
```diff
@@ -0,0 +1,406 @@
...
+      default:
+        throw new IllegalArgumentException("Unexpected ParsedNodeType: " + lineType);
+    }
+  }
+
+  /*
+   * Validate the lineType against the typeRef that has been set, and against child objects. In the
+   * case of no child objects being present, update the lineType to SimpleInitLine.
+   */
+  private void validateType() {
```
garrettjonesgoogle: I might prefer this to be static, and have it take lineType and typeRef as arguments

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
```diff
@@ -0,0 +1,406 @@
...
+          "Primitive type " + parentType + " cannot have children. Child key: " + key);
+    }
+  }
+
+  /**
+   * Validates that the provided value matches the provided type, and returns the validated value.
+   * For string and byte types, the returned value has quote characters removed. Throws an
+   * IllegalArgumentException is the provided type is not supported or doesn't match the value.
+   */
+  private static String validateValue(TypeRef type, String value) {
```
garrettjonesgoogle: maybe instead `getValidatedValue`

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -220,72 +168,75 @@ private GapicSurfaceTestAssertView createAssertView(String expected, String actu
...
-      // page token to empty, in order to indicate that no more pages are available.
+      // Initialize one resource element if it is page-streaming.
       PageStreamingConfig config = context.getMethodConfig().getPageStreaming();
-      fields.add(config.getResourcesField().getSimpleName() + "[0]");
-      fields.add(config.getResponseTokenField().getSimpleName() + "=\"\"");
+      String resourceFieldName = config.getResourcesField().getSimpleName();
+      InitCodeNode childItem = InitCodeNode.create("0");
+      InitCodeNode item =
+          InitCodeNode.createWithChildren(
+              resourceFieldName, InitCodeLineType.ListInitLine, childItem);
```
garrettjonesgoogle: Maybe for readability here, support this:

InitCodeNode item = InitCodeNode.createSingletonList(resourceFieldName);

which creates a InitCodeLineType.ListInitLine with a child item "0". 

And also use it for the bundling case below.

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
michaelbausor: It would also have to take the list of children. Or possibly the check of the children could be moved to a separate function

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
```diff
@@ -0,0 +1,406 @@
...
+
+  public Name getIdentifier() {
+    return identifier;
+  }
+
+  public InitCodeLine getInitCodeLine() {
+    return initCodeLine;
+  }
+
+  public static InitCodeNode create(String key) {
```
shinfan: Can we use the identifier as the key here? It is unclear from the code that what the key should be. Some comments would be helpful.

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
michaelbausor: Because the keys can change. I will update the comment

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
michaelbausor: No, I think it is post-order. Each node emits all of its children before itself.

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
shinfan: Given this is the *only* method to generate the tree, I would prefer the name to be less wordy. Maybe something like "createTree"?

src/main/java/com/google/api/codegen/metacode/SpecItemParserContext.java
```diff
@@ -0,0 +1,76 @@
...
+  @Nullable
+  public abstract Iterable<Field> initFieldSet();
+
+  public abstract Map<String, InitValueConfig> initValueConfigMap();
+
+  @Nullable
+  public abstract List<String> sampleCodeInitFields();
+
+  @Nullable
+  public abstract List<InitCodeNode> subTrees();
```
shinfan: It looks weird to have subTrees() inside a context object

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
```diff
@@ -17,18 +17,10 @@
...
-import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-/**
- * FieldStructureParser parses a dotted path specification into a map of String to Map, List, or
- * InitValueConfig.
- */
```
shinfan: Update the comment?

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
```diff
@@ -50,105 +42,70 @@ static Pattern getFieldMapPattern() {
...
-      if (mergedStructure instanceof InitValueConfig) {
-        InitValueConfig metadata = (InitValueConfig) mergedStructure;
-        if (metadata.isEmpty()) {
-          // Replace empty structure with unmergedStructure
-          return unmergedStructure;
-        }
-      }
-      throw new IllegalArgumentException(
-          "Inconsistent: found both substructure and initialization metadata");
+  private static InitCodeNode parsePartialFieldToInitCodeLineNode(
```
michaelbausor: Done

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
```diff
@@ -50,105 +42,70 @@ static Pattern getFieldMapPattern() {
...
+      toMatch = null;
     }
-  }
 
-  public static Object populate(Object unmergedStructure) {
-    if (unmergedStructure instanceof PathSpec) {
-      return ((PathSpec) unmergedStructure).populate();
+    InitCodeNode item;
+    if (prevItem != null) {
+      item = InitCodeNode.createWithChildren(key, prevType, prevItem);
```
michaelbausor: Done

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
michaelbausor: Done

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
michaelbausor: Done

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
michaelbausor: Done

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
michaelbausor: Done

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
michaelbausor: Done

src/main/java/com/google/api/codegen/metacode/SpecItemParserContext.java
```diff
@@ -0,0 +1,76 @@
...
+
+  @Nullable
+  public abstract TestValueGenerator valueGenerator();
+
+  public abstract TypeRef rootObjectType();
+
+  public abstract Name suggestedName();
+
+  @Nullable
+  public abstract Iterable<Field> initFieldSet();
```
shinfan: initFieldSet sounds like a Set object. Maybe just initFields?

src/test/java/com/google/api/codegen/testdata/java_test_library.baseline
michaelbausor: I have reverted to this, @shinfan what do you think?

michaelbausor:
PTAL

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -60,139 +53,110 @@
...
+  public InitCodeView generateSmokeTestInitCode(
+      MethodTransformerContext context, SymbolTable table) {
+    SmokeTestConfig testConfig = context.getInterfaceConfig().getSmokeTestConfig();
+    InitCodeNode rootNode =
+        InitCodeNode.createSpecItemTree(
+            SpecItemParserContext.newBuilder()
+                .table(table)
+                .rootObjectType(testConfig.getMethod().getInputType())
+                .initValueConfigMap(createInitValueMap(context))
+                .sampleCodeInitFields(testConfig.getInitFields())
```
shinfan: I am not sure if sampleCodeInitFields is the best name here. Since it is actually used by both sample and test gen

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
```diff
@@ -0,0 +1,406 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/*
+ * Represents a node in an tree of objects to be initialized.
+ */
+public class InitCodeNode {
+
+  private String key;
+  private InitCodeLineType lineType;
```
michaelbausor: You are right, they are duplicated at the moment. But once python is moved over to MVVM, the InitCodeLine will be unused, and we can delete it completely.

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
```diff
@@ -0,0 +1,406 @@
...
+  private InitValueConfig initValueConfig;
+  private Map<String, InitCodeNode> children;
+  private TypeRef typeRef;
+  private Name identifier;
+  private InitCodeLine initCodeLine;
```
michaelbausor: I wanted to... but the problem is that we need to modify theses values after construction in the update() function, and AutoValue doesn't allow that.

src/test/java/com/google/api/codegen/testdata/java_test_library.baseline
shinfan: @garrettjonesgoogle  We can use the information from the sample_code_init_fields to determine which fields to init. But we still need to generate unique values since the sample code does not require the values to be unique.

@michaelbausor This looks good to me. Just some comments about the refactoring.

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
michaelbausor: I will add comments. We can't use the identifier, because the key can be different. For a list element, the key is the index, e.g. 0, 1, ... while for a map element the key is entry key in the map.

michaelbausor:
Renamed a number of things and added comments, PTAL

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
```diff
@@ -0,0 +1,417 @@
...
+      default:
+        throw new IllegalArgumentException("Unexpected ParsedNodeType: " + lineType);
+    }
+  }
+
+  /*
+   * Validate the lineType against the typeRef that has been set, and against child objects. In the
+   * case of no child objects being present, update the lineType to SimpleInitLine.
+   */
+  private void getValidateType() {
```
garrettjonesgoogle: this isn't "getting" anything - maybe just "validateType()"?

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
garrettjonesgoogle: Maybe add a TODO to remove initCodeLine after Python is moved to MVVM.

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
```diff
@@ -0,0 +1,417 @@
...
+
+  private String key;
+  private InitCodeLineType lineType;
+  private InitValueConfig initValueConfig;
+  private Map<String, InitCodeNode> children;
+  private TypeRef typeRef;
+  private Name identifier;
+  private InitCodeLine initCodeLine;
+
+  public String getKey() {
```
garrettjonesgoogle: Document the public getters here. Specifically, it's hard to tell the difference between `key` and `identifier`.

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
garrettjonesgoogle: Oh, you're right.

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
```diff
@@ -0,0 +1,417 @@
...
+  }
+
+  private void updateTree(
+      SymbolTable table, TestValueGenerator valueGenerator, TypeRef type, Name suggestedName) {
+
+    // Check and update key values against type, then remove and re-add children with updated keys
+    // Call updateTree() recursively
+    for (InitCodeNode child : new ArrayList<>(children.values())) {
+      children.remove(child.key);
+      child.key = getValidatedKeyValue(type, child.key);
```
garrettjonesgoogle: Make it so that the key doesn't have to change.

michaelbausor:
Addressed PR and offline comments, PTAL

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
```diff
@@ -0,0 +1,439 @@
...
+      }
+      subTrees = newSubTrees;
+    }
+    if (context.additionalSubTrees() != null) {
+      subTrees.addAll(context.additionalSubTrees());
+    }
+    return subTrees;
+  }
+
+  private void updateTree(
```
garrettjonesgoogle: `updateTree` is too general. I recommend `resolveNamesAndTypes`.

PR #419, opened 2016-08-25
Correct default value comments

saicheems:
Redoes some parts of DefaultString to make it more accessible in
snippetgen. Also delegates line endings and comments to snippetgen so
none of that work is done in the contexts.

### DIFF
--- a/src/main/java/com/google/api/codegen/discovery/DefaultString.java
+++ b/src/main/java/com/google/api/codegen/discovery/DefaultString.java
@@ -46,16 +46,16 @@ public class DefaultString {
     }
   }
 
-  private final String declare;
+  private final String define;
   private final String comment;
 
-  private DefaultString(String declare, String comment) {
-    this.declare = declare;
+  public DefaultString(String define, String comment) {
+    this.define = define;
     this.comment = comment;
   }
 
-  public String getDeclare() {
-    return declare;
+  public String getDefine() {
+    return define;
   }
 
   public String getComment() {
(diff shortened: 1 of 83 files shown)


saicheems:
OK! This is a surprising amount of work for a simple thing 😞 

This change refactors a good amount of DefaultString and defaultValue so they don't print semicolons, commas, comments, and a bunch of formatting that should really be at the snippet layer. For all of this work, it fixes the special comments after default fields so they don't absorb semicolons at the end of lines and cause compiler errors.

This is ready for review.

saicheems:
I've only added the compute baseline for ease of review. As far as I can tell, it's the only type that in the discovery samples that has an "eg. " style comment.

pongad:
I admit that I'm a little upset so much change is necessary for this, but I don't see any other way out either. LGTM

src/main/java/com/google/api/codegen/discovery/DefaultString.java
```diff
@@ -75,25 +75,25 @@ public String getComment() {
...
+  public static String getSample(String apiName, String fieldName, String pattern) {
     String sample = null;
+    if (pattern != null) {
+      // If the pattern has a specially-recognized sample, use the sample.
+      sample = SAMPLE_STRINGS.get(SampleKey.create(apiName, fieldName, pattern));
+    }
+    return sample == null ? "" : sample;
+  }
 
+  public static String getDefault(String fieldName, String pattern) {
```
vchudnov-g: nit: getSample() vs getDefault() seems a bit confusing. I would s/getDefault/getPlaceholder/

src/main/java/com/google/api/codegen/go/GoDiscoveryContext.java
```diff
@@ -64,17 +64,12 @@ public String typeDefaultValue(Type type, Field field) {
       return DEFAULT_VALUES.get(field.getKind());
     }
     if (field.getKind() == Field.Kind.TYPE_STRING || field.getKind() == Field.Kind.TYPE_ENUM) {
-      return getDefaultString(type, field);
+      return getDefaultString(type, field).getDeclare();
```
vchudnov-g: nit: Is getDeclare actually a getDefine? You're defining the value, not declaring the type, IIC.

(We can fix the names later if need. I'm probably misunderstanding.)

vchudnov-g:
Some nits on naming, which we could defer.

If tests pass, LGTM.

src/main/java/com/google/api/codegen/discovery/DefaultString.java
saicheems: Done

src/main/java/com/google/api/codegen/go/GoDiscoveryContext.java
saicheems: Done

PR #421, opened 2016-08-25
Fix Go discovery bugs

saicheems:


src/test/java/com/google/api/codegen/testdata/discoveries/go/go_appengine.v1beta5.json.baseline
```diff
@@ -737,7 +737,7 @@ func main() {
     // TODO: Add the desired fields to be changed.
   }
 
-  resp, err := s.Apps.Services.Versions.Patch(appsId, servicesId, versionsId, &version).Context(ctx).Do()
+  resp, err := s.Apps.Services.Versions.Patch(appsId, servicesId, versionsId, version).Context(ctx).Do()
```
broady: Can we map these names to `appsID`? no big deal if not.

saicheems:
Sorry I think I reuploaded while you were reviewing

Removed `fmt` in favor of `log` in every case, and renamed the request body to `body` to avoid some frequent name clashes.

saicheems:
As for the `Id` thing... It's worth making note of

It's a bit tricky to detect acronyms, but we might be able to special case common ones.

broady:
yeah, don't bother with the `Id` thing.

saicheems:
Renamed `body` to `resource`

broady:
LGTM

PR #426, opened 2016-08-26
Init bundling element in the mock response.

shinfan:
Fixes: https://github.com/googleapis/toolkit/issues/417

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
@@ -162,8 +162,14 @@ public class InitCodeTransformer {
       }
     }
     if (context.getMethodConfig().isPageStreaming()) {
+      // Initialize one resource element if it is page-streaming.
       PageStreamingConfig config = context.getMethodConfig().getPageStreaming();
       fields.add(config.getResourcesField().getSimpleName() + "[0]");
+    }
+    if (context.getMethodConfig().isBundling()) {
+      // Initialize one bundling element if it is bundling.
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -162,9 +162,15 @@ private GapicSurfaceTestAssertView createAssertView(String expected, String actu
...
     if (context.getMethodConfig().isPageStreaming()) {
+      // Initialize one resource element if it is page-streaming.
       PageStreamingConfig config = context.getMethodConfig().getPageStreaming();
       fields.add(config.getResourcesField().getSimpleName() + "[0]");
     }
+    if (context.getMethodConfig().isBundling()) {
+      // Initialize one bundling element if it is bundling.
+      fields.add(
+          context.getMethodConfig().getBundling().getSubresponseField().getSimpleName() + "[0]");
+    }
```
garrettjonesgoogle: Why don't we just initialize the whole object tree, and including at least one element of every list?

shinfan: Yes the alternative would be init the all the fields but I feel the marginal benefits of initializing more fields is small. Personally I would prefer the generated code as short as possible given the similar functionaliry. Thoughts?

garrettjonesgoogle: Hmm I think you make a good point. Let's go with this.

garrettjonesgoogle:
LGTM

PR #427, opened 2016-08-26
Fix doc comments for NodeJS.

### LINKED ISSUE #410: [NodeJS] generate documentation of proto messages

jmuk:
This patch contains the following updates:
- fix on the parameter types -- now they are Object, with the
  link text to the message definition.
- add NODEJS_DOC generator type and generate message definitions
  in that case.

Fixes #410


src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
```diff
@@ -495,4 +578,16 @@ public String getByteLengthFunction(TypeRef typeRef) {
               "options",
               "callback")
           .build();
+
+  private static final ImmutableSet<String> COMMON_PROTO_PATHS =
```
geigerj: It doesn't seem clean to hardcode such information as this and `getFileUrl` into `NodeJSGapicContext`. Will there be a better solution under MVVM?

src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
```diff
@@ -182,6 +226,21 @@ private String fieldComment(String comment, String paramComment, Field field) {
       paramComment = JSDocCommentFixer.jsdocify(paramComment);
       comment += "\n  " + paramComment.replaceAll("(\\r?\\n)", "\n  ");
     }
+    if (field.getType().isMessage() && !field.getType().isMap()) {
+      if (!Strings.isNullOrEmpty(paramComment)) {
+        comment += "\n";
+      }
+      comment +=
+          "\n  This object should be in the same structure as "
```
geigerj: "be in the same structure" --> "have the same structure"

src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
geigerj: Note: you don't need to address this now.

src/main/resources/com/google/api/codegen/nodejs/message.snip
```diff
@@ -0,0 +1,127 @@
...
+       {@comments(documentation)}
+       *
+     @end
+     {@properties(message)}
+     * @@class
+     * @@see [{@message.getFullName} definition in proto format]{@@link {@context.getFileURL(message.getFile())}}
+     */
+    {@header}
+      // This is for documentation. Actual contents will be loaded by gRPC.
+      @if or(context.filterDocumentingMessages(message), message.getEnums())
```
geigerj: Why `or` instead of just two `if` conditions? I think this way will produce extraneous whitespace if one or the other is false.

src/main/resources/com/google/api/codegen/nodejs/message.snip
```diff
@@ -0,0 +1,127 @@
...
+        @let valueDoc = context.defaultComments(value)
+          @if documentation
+            /**
+             {@comments(documentation)}
+             */
+          @end
+          {@value.getSimpleName}: {@value.getNumber}
+        @end
+      @end
+    }
```
geigerj: nit: could you add a comment that this closes the curly brace from `{@header}`? Otherwise it is confusing why this is here.

src/main/resources/com/google/api/codegen/nodejs/message.snip
```diff
@@ -0,0 +1,127 @@
...
+       *
+     @end
+     * @@external "{@message.getFullName}"
+     {@properties(message)}
+     * @@see [{@message.getFullName} definition in proto format]{@@link {@context.getFileURL(message.getFile())}}
+     */
+  @end
+@end
+
+@private comments(lines)
```
geigerj: optional: Factor this out into a `common.snip` since it also appears in `main.snip`

src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
jmuk: Done

src/main/resources/com/google/api/codegen/nodejs/message.snip
jmuk: You're right, removed.

src/main/resources/com/google/api/codegen/nodejs/message.snip
jmuk: Added a comment.

src/main/resources/com/google/api/codegen/nodejs/message.snip
jmuk: Done

src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
jmuk: I do not have the answer to your question at this point. Let me investigate, but I'm afraid that MVVM will not be that great -- I mean, I suspect that this type of thing will be more specific to Node so either way logic like this would exist in somewhere specific to Node.

geigerj:
LGTM if build passes

Note: I reviewed only Java code and the templates, not the actual JS output -- presumably if needed it can be refined in a later PR.

PR #429, opened 2016-08-26
Add API key and 3LO scaffold to discovery

saicheems:
Adds 3 methods to ApiaryConfig that allow language contexts to print the
correct auth boilerplate.

### DIFF
--- a/src/main/java/com/google/api/codegen/ApiaryConfig.java
+++ b/src/main/java/com/google/api/codegen/ApiaryConfig.java
@@ -102,14 +102,58 @@ public class ApiaryConfig {
   private final Table<Type, String, Field> fields = HashBasedTable.<Type, String, Field>create();
 
   /*
-   * The service canonical name, or name if no canonical name
+   * The service canonical name, or name if no canonical name.
    */
   private String serviceCanonicalName;
 
   /*
-   * The service version string
+   * The service version string.
    */
   private String serviceVersion;
+
+  private Map<String, AuthType> authOverrides = new HashMap<>();
+
+  /*
+   * If present in the scope list, indicates that the API supports application default credentials
+   * based auth.
+   */
+  private static final String CLOUD_PLATFORM_SCOPE =
+      "https://www.googleapis.com/auth/cloud-platform";
+
+  /*
+   * Possible auth types supported by discovery.
+   */
+  public enum AuthType {
+    APPLICATION_DEFAULT_CREDENTIALS,
+    OAUTH_3L,
+    API_KEY
+  }
+
+  /*
+   * Returns the auth type supported by the service.
+   */
+  public AuthType getAuthType() {
+    String key = getServiceCanonicalName();
+    if (authOverrides.containsKey(key)) {
+      return authOverrides.get(key);
+    }
+    // This statement is based on the assumption that every method in a service contains all the
+    // scopes necessary to determine the correct auth mechanism for the entire service.
+    // Therefore, we use the scopes of the first method in the auth scopes array.
+    key = getAuthScopes().keySet().iterator().next();
+    List<String> scopes = getAuthScopes().get(key);
+    if (scopes.isEmpty()) {
+      // If there are no scopes, it's api key based.
+      return AuthType.API_KEY;
+    } else {
+      // If there are scopes, but cloud platform is one of them, then we can use ADC.
+      if (scopes.contains(CLOUD_PLATFORM_SCOPE)) {
+        return AuthType.APPLICATION_DEFAULT_CREDENTIALS;
+      }
+      // Otherwise it's 3 legged OAuth.
+      return AuthType.OAUTH_3L;
+    }
+  }
 
   public ListMultimap<String, String> getMethodParams() {
     return methodParams;
--- a/src/main/java/com/google/api/codegen/ApiaryConfig.java
+++ b/src/main/java/com/google/api/codegen/ApiaryConfig.java
@@ -126,7 +126,7 @@ public class ApiaryConfig {
   public enum AuthType {
     APPLICATION_DEFAULT_CREDENTIALS,
     OAUTH_3L,
-    API_KEY
+    API_KEY,
   }
 
   /*
@@ -134,23 +134,17 @@ public class ApiaryConfig {
    */
   public AuthType getAuthType() {
     String key = getServiceCanonicalName();
-    if (authOverrides.containsKey(key)) {
+    if (!Strings.isNullOrEmpty(key) && authOverrides.containsKey(key)) {
       return authOverrides.get(key);
     }
-    // This statement is based on the assumption that every method in a service contains all the
-    // scopes necessary to determine the correct auth mechanism for the entire service.
-    // Therefore, we use the scopes of the first method in the auth scopes array.
-    key = getAuthScopes().keySet().iterator().next();
-    List<String> scopes = getAuthScopes().get(key);
-    if (scopes.isEmpty()) {
-      // If there are no scopes, it's api key based.
+    // If the API has no scopes, then we know it's API key-based.
+    if (getAuthScopes().isEmpty()) {
       return AuthType.API_KEY;
     } else {
-      // If there are scopes, but cloud platform is one of them, then we can use ADC.
-      if (scopes.contains(CLOUD_PLATFORM_SCOPE)) {
+      // If there are scopes and cloud platform is one of them, then we can use ADC.
+      if (getAuthScopes().containsValue(CLOUD_PLATFORM_SCOPE)) {
         return AuthType.APPLICATION_DEFAULT_CREDENTIALS;
       }
-      // Otherwise it's 3 legged OAuth.
       return AuthType.OAUTH_3L;
     }
   }
--- a/src/main/java/com/google/api/codegen/discovery/MainDiscoveryProviderFactory.java
+++ b/src/main/java/com/google/api/codegen/discovery/MainDiscoveryProviderFactory.java
@@ -49,16 +49,29 @@ public class MainDiscoveryProviderFactory implements DiscoveryProviderFactory {
   public static final String PYTHON = "python";
   public static final String RUBY = "ruby";
 
-  private static final String DEFAULT_SNIPPET_FILE = "discovery_fragment.snip";
+  private static final String ADC_SNIPPET_FILE = "discovery_fragment.snip";
+  private static final String OAUTH_3L_SNIPPET_FILE = "3lo_discovery_fragment.snip";
+  private static final String API_KEY_SNIPPET_FILE = "api_key_discovery_fragment.snip";
 
   public static DiscoveryProvider defaultCreate(
       Service service, ApiaryConfig apiaryConfig, String id) {
+    String snippetFile;
+    // TODO(saicheems): Update this as new auth types are supported.
+    switch (apiaryConfig.getAuthType()) {
+      case APPLICATION_DEFAULT_CREDENTIALS:
+      case OAUTH_3L:
+      case API_KEY:
+        snippetFile = ADC_SNIPPET_FILE;
+        break;
+      default:
+        throw new RuntimeException("unsupported auth type");
+    }
     if (id.equals(CSHARP)) {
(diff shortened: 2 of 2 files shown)


saicheems:
I want to do this in a set of small PRs, I'll follow up many more that add an auth type to each language.

src/main/java/com/google/api/codegen/ApiaryConfig.java
```diff
@@ -102,15 +102,59 @@
...
+    // This statement is based on the assumption that every method in a service contains all the
+    // scopes necessary to determine the correct auth mechanism for the entire service.
+    // Therefore, we use the scopes of the first method in the auth scopes array.
+    key = getAuthScopes().keySet().iterator().next();
+    List<String> scopes = getAuthScopes().get(key);
+    if (scopes.isEmpty()) {
+      // If there are no scopes, it's api key based.
+      return AuthType.API_KEY;
+    } else {
+      // If there are scopes, but cloud platform is one of them, then we can use ADC.
```
vchudnov-g: s/, but/and

src/main/java/com/google/api/codegen/ApiaryConfig.java
```diff
@@ -102,15 +102,59 @@
...
+    if (authOverrides.containsKey(key)) {
+      return authOverrides.get(key);
+    }
+    // This statement is based on the assumption that every method in a service contains all the
+    // scopes necessary to determine the correct auth mechanism for the entire service.
+    // Therefore, we use the scopes of the first method in the auth scopes array.
+    key = getAuthScopes().keySet().iterator().next();
+    List<String> scopes = getAuthScopes().get(key);
+    if (scopes.isEmpty()) {
+      // If there are no scopes, it's api key based.
```
vchudnov-g: "API Key-based"

src/main/java/com/google/api/codegen/ApiaryConfig.java
```diff
@@ -102,15 +102,59 @@
...
+    String key = getServiceCanonicalName();
+    if (authOverrides.containsKey(key)) {
+      return authOverrides.get(key);
+    }
+    // This statement is based on the assumption that every method in a service contains all the
+    // scopes necessary to determine the correct auth mechanism for the entire service.
+    // Therefore, we use the scopes of the first method in the auth scopes array.
+    key = getAuthScopes().keySet().iterator().next();
+    List<String> scopes = getAuthScopes().get(key);
+    if (scopes.isEmpty()) {
```
vchudnov-g: Not sure we need a comment for every single case here. Maybe just a comment here on top: `Implement a simple heuristic for determining which AUTH_TYPE to use"

src/main/java/com/google/api/codegen/ApiaryConfig.java
```diff
@@ -102,15 +102,59 @@
...
+
+  /*
+   * Returns the auth type supported by the service.
+   */
+  public AuthType getAuthType() {
+    String key = getServiceCanonicalName();
+    if (authOverrides.containsKey(key)) {
+      return authOverrides.get(key);
+    }
+    // This statement is based on the assumption that every method in a service contains all the
```
vchudnov-g: I wonder whether this assumption can land us in trouble.

I also wonder whether it can be misleading. What if the first method is a write method, and so it doesn't list any `readonly` scopes? Then when we generate for a read method, we don't list the `readonly` scopes, even though it would work. Thus, the user thinks that any of the listed scopes work but doesn't realize that the `readonly` scope would, too.

Also, what if the first method has a set of resource-specific scopes, but there are other resources in the API? eg: `mail.thread.readonly` vs `mail.contact.readonly`, say. (I made that up, btw.)

I think it's safest to do this per method....

src/main/java/com/google/api/codegen/discovery/MainDiscoveryProviderFactory.java
```diff
@@ -49,48 +49,61 @@
   public static final String PYTHON = "python";
   public static final String RUBY = "ruby";
 
-  private static final String DEFAULT_SNIPPET_FILE = "discovery_fragment.snip";
+  private static final String ADC_SNIPPET_FILE = "discovery_fragment.snip";
```
vchudnov-g: would it be easier to read to have instead "discovery_fragment_3lo", etc. (the specifics being a suffix rather than prefix)?

src/main/java/com/google/api/codegen/ApiaryConfig.java
saicheems: I'm a bit confused, the method only determines the auth type for a service. It doesn't actually change any scopes that would be listed in the snippet. This can't change how anything down the road is generated.

Now that you mention it though, this can get us in trouble if there's a non-authenticated function we look at first.

I'll change it to use a method.

saicheems: Actually I'd really like to leave it in the service form... but I realized a slightly more robust way to do it:

If the API doesn't have a single scope (`getAuthScopes().isEmpty()`), then we know it's API-key based. If it has more than one scope, and one of the scopes is `"https://googleapis.com/auth/cloud-platform"`, then we know it's ADC based. Otherwise it's 3LO.

I don't think this can fail in the context of discovery, and since it doesn't change how code is generated down the line, none of the `readonly` or scope specific semantics makes a difference.

saicheems:
Updated

@vchudnov-g I changed the way that auth method worked, but I still want it to be service based. I improved the algorithm a bit, PTAL

src/main/java/com/google/api/codegen/ApiaryConfig.java
saicheems: Btw, doing this on a per method basis will not work - some methods, if the API uses 3LO, may not specify scopes. That is, they're unauthenticated calls. We can wrongly identify some methods as API key based if we don't look at the whole service.

src/main/java/com/google/api/codegen/discovery/MainDiscoveryProviderFactory.java
```diff
@@ -49,48 +49,61 @@
   public static final String PYTHON = "python";
   public static final String RUBY = "ruby";
 
-  private static final String DEFAULT_SNIPPET_FILE = "discovery_fragment.snip";
+  private static final String SNIPPET_FILE_ADC = "discovery_fragment.snip";
+  private static final String SNIPPET_FILE_3L0 = "discovery_fragment_3lo.snip";
```
vchudnov-g: s/3L0/3LO/ (letter, not digit)

src/main/java/com/google/api/codegen/ApiaryConfig.java
vchudnov-g: To clarify: you're saying that we look at the scopes to determine the type of template we use for all methods (3LO, ADC, API Key), and, if we use 3LO, for each method we only include in the snippet the scopes listed for that method? That sounds OK.  I was incorrectly assuming you would use for all methods the scopes you read from the first method.

Wrt unauthenticated calls: It would be good to mark them as unauthenticated so that it's obvious to the user. I assume with 3LO this can happen when we notice no scopes listed for the method. For API Keys, I'm not sure what form this can take in Discovery, so maybe we still pass the API Key along unnecessarily (but harmlessly). Thoughts?

saicheems: 1: Yes

2: It's tricky... I think we should start with assuming that all calls are authenticated and then be more clever.

src/main/java/com/google/api/codegen/discovery/MainDiscoveryProviderFactory.java
saicheems: Fixed

src/main/java/com/google/api/codegen/ApiaryConfig.java
```diff
@@ -102,15 +102,53 @@
...
+    String key = getServiceCanonicalName();
+    if (!Strings.isNullOrEmpty(key) && authOverrides.containsKey(key)) {
+      return authOverrides.get(key);
+    }
+    // If the API has no scopes, then we know it's API key-based.
+    if (getAuthScopes().isEmpty()) {
+      return AuthType.API_KEY;
+    } else {
+      // If there are scopes and cloud platform is one of them, then we can use ADC.
+      if (getAuthScopes().containsValue(CLOUD_PLATFORM_SCOPE)) {
```
pongad: Hoist into `else if`? Alternatively skip the `else`?

src/main/java/com/google/api/codegen/discovery/MainDiscoveryProviderFactory.java
```diff
@@ -49,48 +49,61 @@
   public static final String PYTHON = "python";
   public static final String RUBY = "ruby";
 
-  private static final String DEFAULT_SNIPPET_FILE = "discovery_fragment.snip";
+  private static final String SNIPPET_FILE_ADC = "discovery_fragment.snip";
+  private static final String SNIPPET_FILE_3LO = "discovery_fragment_3lo.snip";
+  private static final String SNIPPET_FILE_API_KEY = "discovery_fragment_api_key.snip";
```
pongad: I wonder if we really want to put these into separate files? Looking at Victor's doc, 3LO looks a little gnarly, but API key shouldn't be too hard to fit into existing file right? I fear that the number of files will increase template duplication. I think duplication in template is a lot worse than code duplication, since a lot of tools that we use to help us maintain source code doesn't really work on templates.

src/main/java/com/google/api/codegen/discovery/MainDiscoveryProviderFactory.java
```diff
@@ -49,48 +49,61 @@
...
+    // TODO(saicheems): Update this as new auth types are supported.
+    switch (apiaryConfig.getAuthType()) {
+      case APPLICATION_DEFAULT_CREDENTIALS:
+      case OAUTH_3L:
+      case API_KEY:
+        snippetFile = SNIPPET_FILE_ADC;
+        break;
+      default:
+        throw new RuntimeException("unsupported auth type");
+    }
```
pongad: If we really need each in its own file, consider just putting these strings into `AuthType`'s constructor. That way we know the RuntimeException can't happen.

pongad:
Sorry I'm late to the party :cry:

src/main/java/com/google/api/codegen/ApiaryConfig.java
vchudnov-g: Yeah, skip the `else`

src/main/java/com/google/api/codegen/discovery/MainDiscoveryProviderFactory.java
vchudnov-g: But can we factor our the common parts so that we don't have code duplication? I could imagine two ways of doing it:

1. have a main template file for each auth mechanism. Each of these files consists mostly of the auth boiler plate, and at the bottom it imports a common template for the actual method invocation. This is the way this PR seems currently heading.

1. have a common template that imports the correct boilerplate depending on the auth mechanism used.

Both of these should prevent code duplication. Since the boilerplate is so long, it might be clearer to go for the first option, but I don't have strong feelings one way or the other at the moment. What do you guys think?

src/main/java/com/google/api/codegen/discovery/MainDiscoveryProviderFactory.java
vchudnov-g: Hmm, but then you'd be giving `AuthType` an additional abstraction. If we do this, it would be better renamed `AuthTemplate`.

src/main/java/com/google/api/codegen/discovery/MainDiscoveryProviderFactory.java
saicheems: My intention was to put any actual logic in its own file, and import it below all the boilerplate. Everything can be in the same snippet file... but this is more to manage complexity (and my sanity) than anything.

saicheems: Anyway... going to hold off on this briefly while I see if MVVM improves the snippet files any

PR #434, opened 2016-08-29
Go: Use new retry mechanism

pongad:
This commit accompanies https://github.com/googleapis/gax-go/pull/36 .
It updates the client library generation to use the new retry mechanism.

A notable change:
defaultCallOptions and its friends now lookup the appropriate CallOption
in a map. This has two advantages:
- We should now initialize fewer anonymous functions.
  This is nice but does not greatly matter.
- In the YAML config, the names are normally in snake case.
  The previous implementation converts it into camel.
  We can get rid of this dance completely by just making the names
  opague strings.

### DIFF
--- a/src/main/resources/com/google/api/codegen/go/main.snip
+++ b/src/main/resources/com/google/api/codegen/go/main.snip
@@ -83,22 +83,24 @@
 
     func default{@context.getClientPrefix(service)}CallOptions() *{@context.getClientPrefix(service)}CallOptions {
         retry := map[[2]string][]gax.CallOption{
-            @join retryName : context.getRetryConfigNames(service) if context.getApiConfig.getInterfaceConfig(service).getRetryCodesDefinition.get(retryName.getCodesName)
-                @let retryCodes = context.getApiConfig.getInterfaceConfig(service).getRetryCodesDefinition.get(retryName.getCodesName), \
-                        retrySettings = context.getApiConfig.getInterfaceConfig(service).getRetrySettingsDefinition.get(retryName.getSettingsName)
-                    {"{@retryName.getSettingsName}", "{@retryName.getCodesName}"}: {
-                        gax.WithRetry(func() gax.Retryer {
-                            return gax.OnCodes([]codes.Code{
-                                @join code : retryCodes
-                                    codes.{@context.upperUnderscoreToUpperCamel(code.toString())},
-                                @end
-                            }, gax.Backoff{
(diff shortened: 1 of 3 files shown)


src/main/resources/com/google/api/codegen/go/main.snip
```diff
@@ -81,35 +81,35 @@
...
-        }
-    @end
-
     func default{@context.getClientPrefix(service)}CallOptions() *{@context.getClientPrefix(service)}CallOptions {
-        @join retryCodesDef : context.entrySet(context.getApiConfig.getInterfaceConfig(service).getRetryCodesDefinition) if retryCodesDef.getValue
-            with{@context.lowerUnderscoreToUpperCamel(retryCodesDef.getKey)}RetryCodes := gax.WithRetryCodes([]codes.Code{
-                @join code : retryCodesDef.getValue
-                    codes.{@context.upperUnderscoreToUpperCamel(code.toString())},
+        retry := map[[2]string][]gax.CallOption{
+            @join retryName : context.getRetryConfigNames(service) if context.getApiConfig.getInterfaceConfig(service).getRetryCodesDefinition.get(retryName.getCodesName)
```
jmuk: optional: assign `interfaceConfig` or `retryCodesDefinition` through `@let`? The repetition of `context.getApiConfig...` isn't quite readable.

src/main/resources/com/google/api/codegen/go/main.snip
```diff
@@ -81,35 +81,35 @@
...
+                @let retryCodes = context.getApiConfig.getInterfaceConfig(service).getRetryCodesDefinition.get(retryName.getCodesName), \
+                        retrySettings = context.getApiConfig.getInterfaceConfig(service).getRetrySettingsDefinition.get(retryName.getSettingsName)
```
jmuk: nit: I'm not sure this is correctly aligned. Normally it would be aligned as:
```
@let retryCoes = ..., \
     retrySettings = ...
```

src/main/resources/com/google/api/codegen/go/main.snip
pongad: Done

src/main/resources/com/google/api/codegen/go/main.snip
pongad: You're right. I have changed this to match other declarations in the file

pongad:
@jmuk PTAL

jmuk:
LGTM

PR #442, opened 2016-08-31
Draft the smoke test generator with MVVM.

shinfan:
This is the initial implementation of the automate smoke generator. The basic structure is implemented but some functionalities are missing

**MVVM changes:**
- Config Model
  - Add smoke test into config proto
  - Add SmokeTestConfig class which contains the config data
- Transformer
  - Reorganize JavaGapicSurfaceTestTransformer to improve readability.
  - Implement transformer for smoke test.
  - Implement init code transformer for smoke test.
- View Model
  - SmokeTestClassView
  - TestMethodClassView (This view can be shared by unit test as well, I will make the change in a following PR).
- View
  - smoke_test.snip

**TODOs:**
- Add new config rules so users could assign init values to the fields such as:
   - Support collection pattern
   - Support random value
- Make unit test transformer reuse TestMethodClassView.

### DIFF
--- a/src/main/java/com/google/api/codegen/InterfaceConfig.java
+++ b/src/main/java/com/google/api/codegen/InterfaceConfig.java
@@ -77,7 +77,8 @@ public class InterfaceConfig {
       methodConfigs = createMethodConfigs(methodConfigMap, interfaceConfigProto);
     }
 
-    SmokeTestConfig smokeTestConfig = createSmokeTestConfig(iface, interfaceConfigProto);
+    SmokeTestConfig smokeTestConfig =
+        createSmokeTestConfig(diagCollector, iface, interfaceConfigProto);
 
     if (collectionConfigs == null || methodConfigMap == null) {
       return null;
@@ -93,9 +94,10 @@ public class InterfaceConfig {
   }
 
   private static SmokeTestConfig createSmokeTestConfig(
-      Interface iface, InterfaceConfigProto proto) {
-    if (proto.hasSmokeTest()) {
-      return SmokeTestConfig.createSmokeTestConfig(iface, proto.getSmokeTest());
+      DiagCollector diagCollector, Interface iface, InterfaceConfigProto interfaceConfigProto) {
+    if (interfaceConfigProto.hasSmokeTest()) {
+      return SmokeTestConfig.createSmokeTestConfig(
+          iface, interfaceConfigProto.getSmokeTest(), diagCollector);
     } else {
       return null;
     }
--- a/src/main/java/com/google/api/codegen/InterfaceConfig.java
+++ b/src/main/java/com/google/api/codegen/InterfaceConfig.java
@@ -25,6 +25,8 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Sets;
 
+import org.joda.time.Duration;
+
 import io.grpc.Status;
 
 import java.util.ArrayList;
@@ -34,14 +36,13 @@ import java.util.List;
 
 import javax.annotation.Nullable;
 
-import org.joda.time.Duration;
-
 /**
  * InterfaceConfig represents the code-gen config for an API interface, and includes the
  * configuration for methods and resource names.
  */
 public class InterfaceConfig {
   private final List<MethodConfig> methodConfigs;
+  private final SmokeTestConfig smokeTestConfig;
   private final ImmutableMap<String, CollectionConfig> collectionConfigs;
   private final ImmutableMap<String, MethodConfig> methodConfigMap;
   private final ImmutableMap<String, ImmutableSet<Status.Code>> retryCodesDefinition;
@@ -76,15 +77,27 @@ public class InterfaceConfig {
       methodConfigs = createMethodConfigs(methodConfigMap, interfaceConfigProto);
     }
 
+    SmokeTestConfig smokeTestConfig = createSmokeTestConfig(iface, interfaceConfigProto);
+
     if (collectionConfigs == null || methodConfigMap == null) {
       return null;
     } else {
       return new InterfaceConfig(
           methodConfigs,
+          smokeTestConfig,
           collectionConfigs,
           methodConfigMap,
           retryCodesDefinition,
           retrySettingsDefinition);
+    }
+  }
+
+  private static SmokeTestConfig createSmokeTestConfig(
+      Interface iface, InterfaceConfigProto proto) {
+    if (proto.hasSmokeTest()) {
+      return SmokeTestConfig.createSmokeTestConfig(iface, proto.getSmokeTest());
+    } else {
+      return null;
     }
   }
 
@@ -217,11 +230,13 @@ public class InterfaceConfig {
 
   private InterfaceConfig(
       List<MethodConfig> methodConfigs,
+      SmokeTestConfig smokeTestConfig,
       ImmutableMap<String, CollectionConfig> collectionConfigs,
       ImmutableMap<String, MethodConfig> methodConfigMap,
       ImmutableMap<String, ImmutableSet<Status.Code>> retryCodesDefinition,
       ImmutableMap<String, RetrySettings> retryParamsDefinition) {
     this.methodConfigs = methodConfigs;
+    this.smokeTestConfig = smokeTestConfig;
     this.collectionConfigs = collectionConfigs;
     this.methodConfigMap = methodConfigMap;
     this.retryCodesDefinition = retryCodesDefinition;
@@ -233,6 +248,13 @@ public class InterfaceConfig {
    */
   public List<MethodConfig> getMethodConfigs() {
     return methodConfigs;
+  }
+
+  /**
+   * Returns the SmokeTestConfig.
+   */
+  public SmokeTestConfig getSmokeTestConfig() {
+    return smokeTestConfig;
   }
 
   public CollectionConfig getCollectionConfig(String entityName) {
--- a/src/main/java/com/google/api/codegen/InterfaceConfig.java
+++ b/src/main/java/com/google/api/codegen/InterfaceConfig.java
@@ -80,7 +80,7 @@ public class InterfaceConfig {
     SmokeTestConfig smokeTestConfig =
         createSmokeTestConfig(diagCollector, iface, interfaceConfigProto);
 
-    if (collectionConfigs == null || methodConfigMap == null || diagCollector.hasErrors()) {
+    if (diagCollector.hasErrors()) {
       return null;
     } else {
       return new InterfaceConfig(
--- a/src/main/java/com/google/api/codegen/InterfaceConfig.java
+++ b/src/main/java/com/google/api/codegen/InterfaceConfig.java
@@ -80,7 +80,7 @@ public class InterfaceConfig {
     SmokeTestConfig smokeTestConfig =
         createSmokeTestConfig(diagCollector, iface, interfaceConfigProto);
 
-    if (collectionConfigs == null || methodConfigMap == null) {
+    if (collectionConfigs == null || methodConfigMap == null || diagCollector.hasErrors()) {
       return null;
     } else {
       return new InterfaceConfig(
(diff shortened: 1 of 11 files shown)


src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
```diff
@@ -87,90 +102,64 @@ public JavaGapicSurfaceTestTransformer(GapicCodePathMapper javaPathMapper) {
     return views;
   }
 
-  private Iterable<Interface> getGrpcInterfacesToMock(Model model, ApiConfig apiConfig) {
-    Map<String, Interface> interfaces = new LinkedHashMap<>();
+  ///////////////////////////////////// Smoke Test ///////////////////////////////////////
```
shinfan: I reorder the sequence of the functions in the file to make it more readable. I think this is the section (Smoke Test) that contains the most new code. Other changes are mostly just moving.

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -117,6 +120,14 @@ message InterfaceConfigProto {
   repeated RetryParamsDefinitionProto retry_params_def = 31;
 }
 
+message SmokeTestConfigProto {
```
michaelbausor: Are we also going to include a way to check the return values? Will that be in a followup PR?

src/main/java/com/google/api/codegen/InterfaceConfig.java
```diff
@@ -76,18 +77,30 @@ public static InterfaceConfig createInterfaceConfig(
...
+          smokeTestConfig,
           collectionConfigs,
           methodConfigMap,
           retryCodesDefinition,
           retrySettingsDefinition);
     }
   }
 
+  private static SmokeTestConfig createSmokeTestConfig(
+      Interface iface, InterfaceConfigProto proto) {
```
garrettjonesgoogle: don't call this variable 'proto', that is really confusing.

src/main/java/com/google/api/codegen/SmokeTestConfig.java
```diff
@@ -0,0 +1,64 @@
...
+        testedMethod = method;
+        break;
+      }
+    }
+
+    if (testedMethod != null) {
+      return new SmokeTestConfig(testedMethod, proto.getInitFieldsList());
+    } else {
+      throw new RuntimeException("The configured smoke test method does not exist.");
+    }
```
garrettjonesgoogle: The general pattern for processing configs is to log errors on the diag collector and return null, so that multiple errors can be reported, instead of throwing an exception.

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -113,6 +115,44 @@ public InitCodeView generateMockResponseObjectInitCode(
...
+    SmokeTestConfig testConfig = context.getInterfaceConfig().getSmokeTestConfig();
+    Method method = testConfig.getMethod();
+    MethodTransformerContext methodContext =
+        context.asMethodContext(context.getInterfaceConfig().getSmokeTestConfig().getMethod());
+    Map<String, Object> initFieldStructure =
+        createSmokeTestInitMap(testConfig.getInitFields(), methodContext);
+
+    ArrayList<Field> fields = new ArrayList<>();
+    for (Field field : method.getInputMessage().getFields()) {
+      System.out.println(field.getSimpleName());
```
garrettjonesgoogle: Do we need to keep this line?

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
```diff
@@ -87,90 +102,64 @@ public JavaGapicSurfaceTestTransformer(GapicCodePathMapper javaPathMapper) {
...
-      for (MethodConfig methodConfig : interfaceConfig.getMethodConfigs()) {
-        String reroute = methodConfig.getRerouteToGrpcInterface();
-        if (!Strings.isNullOrEmpty(reroute)) {
-          Interface targetInterface = model.getSymbolTable().lookupInterface(reroute);
-          interfaces.put(reroute, targetInterface);
-        }
-      }
-    }
+  private SmokeTestClassView createSmokeTestClassView(SurfaceTransformerContext context) {
+    context = context.withNewTypeTable();
```
garrettjonesgoogle: I would prefer the creation of contexts to be consistent; for the other views, the creation of a new context is done in transform().

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
```diff
@@ -87,90 +102,64 @@ public JavaGapicSurfaceTestTransformer(GapicCodePathMapper javaPathMapper) {
...
-    typeTable.saveNicknameFor("java.util.Arrays");
-    typeTable.saveNicknameFor("java.util.List");
-    typeTable.saveNicknameFor("com.google.api.gax.testing.MockServiceHelper");
-    typeTable.saveNicknameFor("com.google.api.gax.testing.MockGrpcService");
-    typeTable.saveNicknameFor("com.google.api.gax.core.PageAccessor");
-    typeTable.saveNicknameFor("com.google.common.collect.Lists");
-    typeTable.saveNicknameFor("com.google.protobuf.GeneratedMessage");
-  }
+    ApiMethodType methodType = ApiMethodType.FlattenedMethod;
+    if (context.asMethodContext(method).getMethodConfig().isPageStreaming()) {
```
garrettjonesgoogle: If this function (createSmokeTestMethodView) is limited to a single method, it should probably take a method context as its parameter instead of a surface context.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
```diff
@@ -289,6 +263,29 @@ private MockGrpcResponseView createMockResponseView(
     return MockGrpcResponseView.newBuilder().typeName(typeName).initCode(initCodeView).build();
   }
 
+  private String getTestName(SymbolTable symbolTable, SurfaceNamer namer, Method method) {
+    Name name = symbolTable.getNewSymbol(Name.lowerCamel(namer.getTestCaseName(method)));
```
garrettjonesgoogle: Don't put casing calls in the transformers. Move this into SurfaceNamer.

src/main/resources/com/google/api/codegen/java/smoke_test.snip
```diff
@@ -0,0 +1,59 @@
...
+        System.out.println("ok");
+      } catch (Exception e) {
+        System.err.println("Failed with exception:");
+        e.printStackTrace(System.err);
+        System.exit(1);
+      }
+    }
+
+    public static void executeNoCatch(String args[]) throws Exception {
+      @let method = smokeTest.method
```
garrettjonesgoogle: Don't use let's like this in MVVM snippets - call off to a new snippet function

src/main/resources/com/google/api/codegen/java/smoke_test.snip
```diff
@@ -0,0 +1,59 @@
...
+        try ({@smokeTest.apiClassName} api = {@smokeTest.apiClassName}.create()) {
+          {@initCode(method.initCode)}
+
+          @switch method.type
+          @case "PagedFlattenedMethod"
+            PageAccessor<{@method.responseTypeName}> pageAccessor =
+                api.{@method.name}(\
+                  {@sampleMethodCallArgList(method.initCode.fieldSettings)});
+            System.out.println(ReflectionToStringBuilder.toString(pageAccessor));
```
garrettjonesgoogle: Is this print line necessary?

garrettjonesgoogle: Oh I see that we want this. Does it actually print the contents of the list?

src/main/proto/com/google/api/codegen/config.proto
shinfan: It is possible to have auto-generated checks but I would consider it as a future enhancement. Right now the return object will be recursively printed as JSON strings to stdout so the nested values can be checked manually.

src/main/resources/com/google/api/codegen/java/smoke_test.snip
shinfan: Yes, the print is recursive.

src/main/java/com/google/api/codegen/InterfaceConfig.java
shinfan: Done

src/main/java/com/google/api/codegen/SmokeTestConfig.java
shinfan: Done

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
shinfan: Done.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
shinfan: Done.

src/main/resources/com/google/api/codegen/java/smoke_test.snip
shinfan: Done.

shinfan:
PTAL

src/main/java/com/google/api/codegen/InterfaceConfig.java
```diff
@@ -76,18 +77,32 @@ public static InterfaceConfig createInterfaceConfig(
       methodConfigs = createMethodConfigs(methodConfigMap, interfaceConfigProto);
     }
 
+    SmokeTestConfig smokeTestConfig =
+        createSmokeTestConfig(diagCollector, iface, interfaceConfigProto);
+
```
garrettjonesgoogle: There needs to be a check for smokeTestConfig being null (it needs to be one of the conditions in the 'if' below)

garrettjonesgoogle: I take it back, 'null' is not the right signal. I think we should change the check below to check whether there are errors on diagCollector (and return 'null' if so).

src/test/java/com/google/api/codegen/testdata/java_smoke_test_library.baseline
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.gcloud.pubsub.spi;
+
+import autovalue.shaded.org.apache.commons.lang.builder.ReflectionToStringBuilder;
```
garrettjonesgoogle: We shouldn't be using imports of shaded classes.

garrettjonesgoogle:
Just a couple minor things to address.

src/main/java/com/google/api/codegen/InterfaceConfig.java
shinfan: Done. Add check for errors.

src/test/java/com/google/api/codegen/testdata/java_smoke_test_library.baseline
shinfan: Done. We need to added the org.apache.commons.lang dependency into the pom file.

garrettjonesgoogle:
LGTM, one non-blocking suggestion

PR #443, opened 2016-08-31
Fix NodeJS style violations

jmuk:
- unnecessary semicolon at the end of module initializer function
- keys of dictionaries are usually not quoted (i.e.
  not `{'foo': foo}` but `{foo: foo}`)

See https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1548


swcloud:
LGTM

PR #444, opened 2016-08-31
Update configgen package name for Ruby/NodeJS

jmuk:
- Ruby has the same naming policy of Google::Cloud as other
  languages have.
- NodeJS -- uses @google-cloud/... for cloud APIs.

### DIFF
--- a/src/main/java/com/google/api/codegen/config/LanguageGenerator.java
+++ b/src/main/java/com/google/api/codegen/config/LanguageGenerator.java
@@ -15,6 +15,7 @@
 package com.google.api.codegen.config;
 
 import com.google.common.base.Joiner;
+import com.google.common.base.Splitter;
 import com.google.common.collect.ImmutableMap;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -28,7 +29,7 @@ import java.util.Map;
  */
 public class LanguageGenerator {
 
-  private static final String DEFAULT_PACKAGE_SEPARATOR = "\\.";
+  private static final String DEFAULT_PACKAGE_SEPARATOR = ".";
 
   private static final String CONFIG_KEY_PACKAGE_NAME = "package_name";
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/api/codegen/config/LanguageGenerator.java
```diff
@@ -132,6 +135,23 @@ public String getFormattedPackageName(String packageName) {
     }
   }
 
+  private static class NodeJSLanguageFormatter extends SimpleLanguageFormatter {
```
garrettjonesgoogle: Don't extend SimpleLanguageFormatter - you're overriding the only method defined. Just use "implements LanguageFormatter".

src/main/java/com/google/api/codegen/config/LanguageGenerator.java
```diff
@@ -132,6 +135,23 @@ public String getFormattedPackageName(String packageName) {
...
+    public NodeJSLanguageFormatter() {
+      super("-", null, false);
+    }
+
+    public String getFormattedPackageName(String packageName) {
+      List<String> nameComponents = Splitter.on(DEFAULT_PACKAGE_SEPARATOR).splitToList(packageName);
+      int size = nameComponents.size();
+      if (size < 3
+          || !nameComponents.get(0).equals("google")
+          || !nameComponents.get(size - 1).startsWith("v")) {
```
garrettjonesgoogle: make a static function for this if logic, and call it also from GoLanguageFormatter

src/main/java/com/google/api/codegen/config/LanguageGenerator.java
```diff
@@ -43,14 +44,17 @@
         Arrays.asList(new RewriteRule("^google(?!\\.cloud)", "google.cloud"));
     List<RewriteRule> pythonRewriteRules =
         Arrays.asList(new RewriteRule("^google(?!\\.cloud)", "google.cloud"));
+    List<RewriteRule> rubyRewriteRules =
+        Arrays.asList(new RewriteRule("^google(?!\\.cloud)", "google.cloud"));
```
garrettjonesgoogle: With three instances of this RewriteRule, I think it's worth creating it once and passing it into each of the three languages.

src/main/java/com/google/api/codegen/config/LanguageGenerator.java
jmuk: Done

src/main/java/com/google/api/codegen/config/LanguageGenerator.java
jmuk: Done

jmuk:
Is there anything I should fix?

src/main/java/com/google/api/codegen/config/LanguageGenerator.java
```diff
@@ -76,6 +76,16 @@ private static String firstCharToUpperCase(String string) {
     return Character.toUpperCase(string.charAt(0)) + string.substring(1);
   }
 
+  /**
+   * Returns true if it is a Google Cloud API.
+   */
+  private static boolean IsApiGoogleCloud(List<String> nameComponents) {
```
garrettjonesgoogle: Use lower-camel for methods in Java

garrettjonesgoogle:
LGTM other than the casing issue.

src/main/java/com/google/api/codegen/config/LanguageGenerator.java
jmuk: Done

PR #445, opened 2016-09-01
Add basic 3LO and API key auth support in discovery

saicheems:
Adds an auth placeholder in the Java sample depending on the API's auth
type.

Adds `auth_url` option to `DiscoveryFragmentGeneratorTool.java` to pass
custom auth instructions URLs to the fragments.

Adds `AuthType` enum and logic to `ApiaryConfig.java`
- Selects API key if there are no scopes
- Selects ADC if there are scopes and cloud is one of them
- Selects 3LO otherwise

### DIFF
--- a/src/main/java/com/google/api/codegen/ApiaryConfig.java
+++ b/src/main/java/com/google/api/codegen/ApiaryConfig.java
@@ -102,14 +102,63 @@ public class ApiaryConfig {
   private final Table<Type, String, Field> fields = HashBasedTable.<Type, String, Field>create();
 
   /*
-   * The service canonical name, or name if no canonical name
+   * The service canonical name, or name if no canonical name.
    */
   private String serviceCanonicalName;
 
   /*
-   * The service version string
+   * The service version string.
    */
   private String serviceVersion;
+
+  /*
+   * The auth instructions URL.
+   */
+  private String authInstructionsUrl;
+
+  /*
+   * Maps API names to an AuthType override.
+   */
+  private Map<String, AuthType> authOverrides = new HashMap<>();
+
+  /*
+   * If present in the scope list, indicates that the API supports application default credentials
+   * based auth.
+   */
+  private static final String CLOUD_PLATFORM_SCOPE =
+      "https://www.googleapis.com/auth/cloud-platform";
+
+  /*
+   * Possible auth types supported by discovery.
+   */
+  public enum AuthType {
+    APPLICATION_DEFAULT_CREDENTIALS,
+    OAUTH_3L,
+    API_KEY,
+  }
+
+  /*
+   * Returns the auth type supported by the service.
+   */
+  public AuthType getAuthType() {
+    String key = getServiceCanonicalName();
+    if (!Strings.isNullOrEmpty(key) && authOverrides.containsKey(key)) {
+      return authOverrides.get(key);
+    }
+    // If the API has no scopes, then we know it's API key-based.
+    if (getAuthScopes().isEmpty()) {
+      return AuthType.API_KEY;
+    }
+    // If there are scopes and cloud platform is one of them, then we can use ADC.
+    if (getAuthScopes().containsValue(CLOUD_PLATFORM_SCOPE)) {
+      return AuthType.APPLICATION_DEFAULT_CREDENTIALS;
+    }
+    return AuthType.OAUTH_3L;
+  }
+
+  public String getAuthInstructionsUrl() {
+    return Strings.nullToEmpty(authInstructionsUrl);
+  }
 
   public ListMultimap<String, String> getMethodParams() {
     return methodParams;
@@ -165,6 +214,10 @@ public class ApiaryConfig {
 
   public void setServiceVersion(String serviceVersion) {
     this.serviceVersion = serviceVersion;
+  }
+
+  public void setAuthInstructionsUrl(String authInstructionsUrl) {
+    this.authInstructionsUrl = authInstructionsUrl;
   }
 
   /**
@@ -229,5 +282,12 @@ public class ApiaryConfig {
    */
   public ListMultimap<String, String> getAuthScopes() {
     return authScopes;
+  }
+
+  /*
+   * @return true if method has any auth scopes.
+   */
+  public boolean hasAuthScopes(String methodName) {
+    return authScopes.containsKey(methodName);
   }
 }
--- a/src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorApi.java
+++ b/src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorApi.java
@@ -74,7 +74,7 @@ public class DiscoveryFragmentGeneratorApi {
       ToolOptions.createOption(
           String.class,
           "auth_instructions",
-          "A comma-delimited map of language to auth instructions URL: lang:URL,lang:URL",
+          "A comma-delimited map of language to auth instructions URL: lang:URL,lang:URL,...",
           "");
 
   private final ToolOptions options;
@@ -129,7 +129,7 @@ public class DiscoveryFragmentGeneratorApi {
   /*
    * Parses strings of the format
    * "go:https://www.hello.com,java:https://www.world.com" and returns the value
-   * of the key corresponding to id.
+   * of the key corresponding to the key id.
    */
   private String parseAuthInstructionsUrl(String authInstructionsUrl, String id) {
     if (Strings.isNullOrEmpty(authInstructionsUrl)) {
--- a/src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorApi.java
+++ b/src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorApi.java
@@ -73,8 +73,8 @@ public class DiscoveryFragmentGeneratorApi {
   public static final Option<String> AUTH_INSTRUCTIONS_URL =
       ToolOptions.createOption(
           String.class,
-          "auth_url",
-          "A comma delimited map of language to auth instructions URL.",
+          "auth_instructions",
+          "A comma-delimited map of language to auth instructions URL: lang:URL,lang:URL",
           "");
 
   private final ToolOptions options;
@@ -111,9 +111,9 @@ public class DiscoveryFragmentGeneratorApi {
     String factory = generator.getFactory();
     String id = generator.getId();
 
-    String authUrl = options.get(AUTH_INSTRUCTIONS_URL);
+    String authInstructions = options.get(AUTH_INSTRUCTIONS_URL);
     ApiaryConfig apiaryConfig = discovery.getConfig();
-    apiaryConfig.setAuthInstructionsUrl(parseAuthUrl(authUrl, id));
+    apiaryConfig.setAuthInstructionsUrl(parseAuthInstructionsUrl(authInstructions, id));
 
     DiscoveryProviderFactory providerFactory = createProviderFactory(factory);
     DiscoveryProvider provider = providerFactory.create(discovery.getService(), apiaryConfig, id);
@@ -131,12 +131,12 @@ public class DiscoveryFragmentGeneratorApi {
    * "go:https://www.hello.com,java:https://www.world.com" and returns the value
    * of the key corresponding to id.
    */
-  private String parseAuthUrl(String authUrl, String id) {
-    if (Strings.isNullOrEmpty(authUrl)) {
+  private String parseAuthInstructionsUrl(String authInstructionsUrl, String id) {
+    if (Strings.isNullOrEmpty(authInstructionsUrl)) {
       return "";
     }
     // Split on ','
-    Iterable<String> it = Splitter.on(',').split(authUrl);
+    Iterable<String> it = Splitter.on(',').split(authInstructionsUrl);
     for (String item : it) {
       // Split on only the first ':' and return if there are two values and the
       // language matches.
--- a/src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorApi.java
+++ b/src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorApi.java
@@ -14,6 +14,18 @@
  */
 package com.google.api.codegen;
 
+import java.io.File;
+import java.nio.charset.Charset;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+
+import javax.annotation.Nullable;
+
 import com.google.api.codegen.discovery.DiscoveryProvider;
 import com.google.api.codegen.discovery.DiscoveryProviderFactory;
 import com.google.api.codegen.util.ClassInstantiator;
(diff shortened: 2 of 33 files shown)


saicheems:
I've added the `auth_url` flag, but I don't know of a way to test it...
That being said, PTAL, @vchudnov-g

saicheems:
Note: This pull is against a new branch I've made, `disco_head`, that is from right before the consistency merge.

src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorApi.java
```diff
@@ -69,6 +70,13 @@
           "The list of YAML configuration files for the fragment generator.",
           ImmutableList.<String>of());
 
+  public static final Option<String> AUTH_INSTRUCTIONS_URL =
+      ToolOptions.createOption(
+          String.class,
+          "auth_url",
```
vchudnov-g: Wonder whether it would be more correct to call this `auth_instructions`. `auth_url` sounds like a functional URL that you go to to perform a step in the auth process.

src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorApi.java
```diff
@@ -69,6 +70,13 @@
...
+          "A comma delimited map of language to auth instructions URL.",
```
vchudnov-g: hyphenate "comma-delimited"

src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorApi.java
```diff
@@ -115,6 +126,28 @@ protected void process() throws Exception {
     }
   }
 
+  /*
+   * Parses strings of the format
+   * "go:https://www.hello.com,java:https://www.world.com" and returns the value
+   * of the key corresponding to id.
+   */
+  private String parseAuthUrl(String authUrl, String id) {
```
vchudnov-g: Would it be conceptually simpler to parse this once into a language:url map, and then just get values from the map?

src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorTool.java
```diff
@@ -57,6 +57,13 @@ public static void main(String[] args) throws Exception {
             .hasArg()
             .argName("OUTPUT-DIRECTORY")
             .build());
+    options.addOption(
```
vchudnov-g: This is totally fine, but consider whether in the future a YAML input format would be better. If so, we can plan for it.

src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorTool.java
```diff
@@ -57,6 +57,13 @@ public static void main(String[] args) throws Exception {
...
+        Option.builder()
+            .longOpt("auth_url")
+            .desc("A comma delimited map of language to auth instructions URL.")
```
vchudnov-g: hyphenate "comma-delimited"

suggest ending with: `... instructions URL: lang:URL,lang:URL,...`

src/main/resources/com/google/api/codegen/java/discovery_fragment.snip
```diff
@@ -41,18 +41,26 @@
...
+            {@importGoogleCredential()}{@adcBoilerplate()}
+          @default
+            // {@TODO()} Implement this function to get authentication credentials.
+            @let authInstructionsUrl = apiaryConfig.getAuthInstructionsUrl
+              @if authInstructionsUrl
+                // See {@authInstructionsUrl}
+              @end
+            @end
+            @if apiaryConfig.hasAuthScopes(method.getName)
+              // Authorize using the following scopes in order to use this method:
```
vchudnov-g: "using one of the following scopes" would be more accurate, no?

src/test/java/com/google/api/codegen/testdata/discoveries/java/java_adexchangebuyer.v1.4.json.baseline
```diff
@@ -30,18 +30,10 @@ import java.util.Collections;
...
-
-    // The createScopedRequired method returns true when running on GAE or a local developer
-    // machine. In that case, the desired scopes must be passed in manually. When the code is
-    // running in GCE, GKE or a Managed VM, the scopes are pulled from the GCE metadata server.
-    // For more information, see
-    // https://developers.google.com/identity/protocols/application-default-credentials
-    if (credential.createScopedRequired()) {
-      credential = credential.createScoped(Collections.singletonList("https://www.googleapis.com/auth/cloud-platform"));
-    }
+    // TODO: Implement this function to get authentication credentials.
```
vchudnov-g: Would having a blank line above the TODO make it stand out more? If it doesn't violate style guides and is consistent with our coding elsewhere, I would do it.

vchudnov-g:
Looking really good!

re testing the URL flag: do we have any end-to-end tests that go from flags to snippets? That would be one possible test.

Something else we could do is a unit test where we populate the URL value internally in Java and make sure it gets propagated through the template into the generated snippet.

src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorTool.java
saicheems: Yeah, I agree

I admittedly did it this way to save time... but I'll be sure to do file based configuration in the refactoring

src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorApi.java
saicheems: So I had originally gone about implementing with a map... but since the generator operates on a language by language basis, it didn't seem worth storing the other language pairs for any reason. I figured returning when we've hit the right key was sufficient.

I hope that with the comments the function's behavior is pretty clear.

src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorTool.java
vchudnov-g: OK, just be aware when you do, you'll potentially need to support this flag as well for a little bit until I change the scripts to run `discoGen` to actually send out the info as yaml files rather than this long file.

With any  luck, I can have that coded earlier, though, and it'll just be a flag flip from the script side.

saicheems:
Ok, I've populated the flag manually in `DiscoveryGeneratorTestBase.java`

It now carries through to the end of the tests, PTAL

src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorApi.java
```diff
@@ -69,6 +70,13 @@
           "The list of YAML configuration files for the fragment generator.",
           ImmutableList.<String>of());
 
+  public static final Option<String> AUTH_INSTRUCTIONS_URL =
+      ToolOptions.createOption(
+          String.class,
+          "auth_instructions",
+          "A comma-delimited map of language to auth instructions URL: lang:URL,lang:URL",
```
vchudnov-g: add a literal `,...`at the end to indicate the map can be more than two entries long

src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorTool.java
```diff
@@ -67,18 +74,24 @@ public static void main(String[] args) throws Exception {
...
+      String outputDirectory,
+      String authInstructions)
+      throws Exception {
 
     ToolOptions options = ToolOptions.create();
     options.set(DiscoveryFragmentGeneratorApi.DISCOVERY_DOC, discoveryDoc);
     options.set(DiscoveryFragmentGeneratorApi.OUTPUT_FILE, outputDirectory);
     options.set(
         DiscoveryFragmentGeneratorApi.GENERATOR_CONFIG_FILES, Arrays.asList(generatorConfigs));
+    options.set(DiscoveryFragmentGeneratorApi.AUTH_INSTRUCTIONS_URL, authInstructions);
```
vchudnov-g: Just curious: would using `Arrays.asList` as in the line above simplify your parsing?

src/main/resources/com/google/api/codegen/java/discovery_fragment.snip
```diff
@@ -41,18 +41,27 @@
...
-
-        // The createScopedRequired method returns true when running on GAE or a local developer
-        // machine. In that case, the desired scopes must be passed in manually. When the code is
-        // running in GCE, GKE or a Managed VM, the scopes are pulled from the GCE metadata server.
-        // For more information, see
-        // https://developers.google.com/identity/protocols/application-default-credentials
-        if (credential.createScopedRequired()) {
-          credential = credential.createScoped(Collections.singletonList("https://www.googleapis.com/auth/cloud-platform"));
-        }
+
```
vchudnov-g: Does the blank line need to be moved into the `@default` case to not screw up the formatting (and baseline files!) for the ADC case? Or did you intend to insert a blank line there too (which is fine)?

(oh, I see you already changed the baselines below, so that seems intentional. cool)

src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorApi.java
```diff
@@ -115,6 +126,28 @@ protected void process() throws Exception {
     }
   }
 
+  /*
+   * Parses strings of the format
+   * "go:https://www.hello.com,java:https://www.world.com" and returns the value
+   * of the key corresponding to id.
```
vchudnov-g: ...`the value corresponding to the key id.`

vchudnov-g:
Could you change the comma-delimited to semicolon delimiters? We pass arguments through gradlew as a comma-separated list already, so having the same delimiter here would collide.

src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorTool.java
saicheems: I tried... I couldn't get it to work...

saicheems: Are you able to pass a list of config files? I can't get that array option to separate lists

saicheems:
Done

src/main/java/com/google/api/codegen/ApiaryConfig.java
```diff
@@ -102,15 +102,64 @@
...
   /*
-   * The service version string
+   * The service version string.
    */
   private String serviceVersion;
 
+  /*
+   * The auth instructions URL.
+   */
+  private String authInstructionsUrl;
```
vchudnov-g: s/Url/Urls/, and pluralize the comment above, expanding it to clarify that it's a flattened list of per-language URLs.

src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorApi.java
```diff
@@ -69,6 +70,13 @@
           "The list of YAML configuration files for the fragment generator.",
           ImmutableList.<String>of());
 
+  public static final Option<String> AUTH_INSTRUCTIONS_URL =
+      ToolOptions.createOption(
+          String.class,
+          "auth_instructions",
+          "An @-delimited map of language to auth instructions URL: lang:URL@lang:URL@...",
```
vchudnov-g: It just occurred to me that `lang@URL:lang@URL:...` might weirdly be more humanly-parseable. WDYT? I guess we can go with what we have, though.

src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorApi.java
```diff
@@ -115,6 +126,28 @@ protected void process() throws Exception {
     }
   }
 
+  /*
+   * Parses strings of the format
+   * "go:https://www.hello.com@java:https://www.world.com" and returns the value
+   * of the key corresponding to the key id.
+   */
+  private String parseAuthInstructionsUrl(String authInstructionsUrl, String id) {
```
vchudnov-g: I still think parsing once into a map and then retrieving from the map is cleaner, but we can defer this.

src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorTool.java
vchudnov-g: That's interesting. We should ask the original author of that line what the thinking was.

In our current invocations, we're only invoking this one language at a time, so we're not exercising the multiple config file path.

We need to increase our test coverage!

src/main/resources/com/google/api/codegen/java/discovery_fragment.snip
```diff
@@ -226,11 +258,18 @@
   @end
 @end
 
+@private importCredential()
+  {@context.addImport("com.google.api.client.auth.oauth2.Credential")}
+@end
+
+@private importGoogleCredential()
+  {@context.addImport("com.google.api.client.googleapis.auth.oauth2.GoogleCredential")}
```
vchudnov-g: does this blank line matter? remove it for consistency?

src/test/java/com/google/api/codegen/DiscoveryGeneratorTestBase.java
```diff
@@ -81,6 +81,10 @@ protected Object run() {
     String factory = generator.getFactory();
     String id = generator.getId();
 
+    // This field is manually populated to test the end-to-end behavior of the
+    // "auth_instructions" flag.
+    discoveryImporter.getConfig().setAuthInstructionsUrl("https://foo.com/bar");
```
vchudnov-g: rm blank line?

vchudnov-g:
LGTM, but happy to TAL if you make the non-urgent non-trivial changes I mentioned.

src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorApi.java
saicheems: It's the same amount of work... I don't understand the benefit? The map is populated in the same for loop, all this does is not bother constructing a map because only one language matters.

Either way, this is temporary for this branch, I can change it on the other.

src/main/resources/com/google/api/codegen/java/discovery_fragment.snip
saicheems: What blank line? I think everything is right here...

src/main/java/com/google/api/codegen/ApiaryConfig.java
saicheems: This isn't a delimited list - it only contains one URL from the language it's for. The list is split apart in `DiscoveryFragmentGeneratorApi.java`

src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorApi.java
saicheems: I think `:` makes more sense to indicate mapping... I'm going to leave it as is and this whole thing will disappear on the other branch.

src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorApi.java
vchudnov-g: If we did invoke this with multiple languages, then IIUC the code is parsings the serialized map once per language, rather than parsing it once and storing the results in the map. Not a biggie, but not elegant.

Since our current invocation is just one language at a time, it doesn't really matter much right now. But if the code says  we support multiple languages in one invocation, we should make sure it does.

But yeah, totally fine deferring this until we're back on master.

PR #447, opened 2016-09-01
Remove the generate_samples flag altogether.

bjwatson:
We have already removed them from the GAPIC config files, but
Python was still looking for it.


garrettjonesgoogle:
LGTM

PR #448, opened 2016-09-01
Delay loading grpc Ruby.

### LINKED ISSUE #446: Ruby: allow delay loading of grpc module
Current structure of generated file always load dependency at the toplevel of the file. Therefore, when a ruby script has a statement like
```ruby
require "google/cloud/pubsub/v1/publisher_api"
```
this implicitly loads `publisher_service_pb` (grpc-generated file) which implicitly loads `grpc` which initializes the module.

According to https://github.com/grpc/grpc/issues/7951,
> Currently, forking after initializing grpc is not supported

So this will be problematic. We should delay loading required module as far as possible.

jmuk:
Fixes #446 with https://github.com/googleapis/gax-ruby/pull/41.

### DIFF
--- a/src/main/resources/com/google/api/codegen/ruby/main.snip
+++ b/src/main/resources/com/google/api/codegen/ruby/main.snip
@@ -45,7 +45,6 @@
   require "pathname"
 
   require "google/gax"
-  require "{@context.getGrpcFilename(service)}"
 @end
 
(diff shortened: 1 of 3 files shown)


blowmage:
This would be a terrific change for google-cloud-ruby.

Landrito:
LGTM

PR #453, opened 2016-09-02
Add Go 3LO and API key auth support

saicheems:


src/test/java/com/google/api/codegen/testdata/discoveries/go/go_adexchangebuyer.v1.4.json.baseline
```diff
@@ -6,23 +6,22 @@ package main
...
 )
 
 func main() {
   ctx := context.Background()
-  hc, err := google.DefaultClient(ctx, adexchangebuyer.AdexchangeBuyerScope)
+  // TODO: Implement this function to get authentication credentials.
+  // See https://foo.com/bar
+  // Authorize using one of the following scopes in order to use this method:
+  //   https://www.googleapis.com/auth/adexchange.buyer
+  hc, err := getClient(ctx)
```
vchudnov-g: define a stub here, as in Python
```
hc, err := func(_ *context.foo) (*bar, error) { return nil, nil}(ctx)
```

saicheems: Already implemented a function, PTAL at `getClient`

vchudnov-g:
LGTM

saicheems:
FYI, I added a small rider to this pull that crudely fixes the semicolon comment bug where:

`... // eg. "us-central-1";`

would show up.

PR #456, opened 2016-09-02
Add PHP 3LO and API key auth support

saicheems:


vchudnov-g:
LGTM

saicheems:
Most APIs passed in compilecheck. Some failed, and I didn't scrutinize the error messages, but I believe they were due to type name mismatches.

vchudnov-g:
Were these mismatches caused by the auth stub or by issues with the client libraries trailing Discovery?

PR #457, opened 2016-09-02
Add Node.js 3LO and API key auth support

saicheems:


saicheems:
This one might need a bit more scrutiny, I don't know if I've done it properly.

src/test/java/com/google/api/codegen/testdata/discoveries/nodejs/nodejs_adexchangebuyer.v1.4.json.baseline
```diff
@@ -4,24 +4,24 @@
 // 1. If not already done, enable the Ad Exchange Buyer API
 //    and check the quota for your project at
 //    https://console.developers.google.com/apis/api/adexchangebuyer
-// 2. This sample uses Application Default Credentials for authentication.
-//    If not already done, install the gcloud CLI from
-//    https://cloud.google.com/sdk/ and run
-//    'gcloud beta auth application-default login'
-// 3. Install the Node.js client library and Application Default Credentials
+// 2. Install the Node.js client library and Application Default Credentials
```
vchudnov-g: This still refers to ADC. It may be simple enough to just remove the reference in the comments.

One of us should double-check everything works by running the code.

src/main/resources/com/google/api/codegen/nodejs/discovery_fragment.snip
```diff
@@ -13,37 +13,94 @@
...
+          authClient = authClient.createScoped(scopes);
+        }
+
+        var {@requestName()} = {
+          {@initialize(method)}
+          // Auth client
+          auth: authClient
+        };
+
+        @if context.isPageStreaming(method)
```
vchudnov-g: I think you can factor this `@if` out of the `@switch` without any loss in clarity.

(I think you could also factor out stuff in creating the request right above, but there I think we would lose some clarity, so it's probably better not to.)

src/test/java/com/google/api/codegen/testdata/discoveries/nodejs/nodejs_adexchangebuyer.v1.4.json.baseline
saicheems: I did test logging.entries.list some time ago with this change and it worked fine, though I haven't tested the full range of possibilities.
I'm reasonably confident other configurations will work, but if you'd like me to be more comprehensive, let me know.

vchudnov-g: Do remove the ADC reference. If you could run the compilechecks to make sure we're not generating errors, that would be great. I am working on generalizing/automating the compile check, but it will be a little bit yet before that's ready.

saicheems: Will remove, haven't gotten to nodejs yet :)

saicheems: Done

PR #459, opened 2016-09-05
Delete unnecessary imports

pongad:


garrettjonesgoogle:
LGTM

sidenote, it would be nice to have a utility that could do this (instead of having to do it through Eclipse), because we need a way to clean generated imports from our generated code.

pongad:
@garrettjonesgoogle I'm actually working on this as a weekend project 😄

Issue #343, opened 2016-08-04
Support gRPC streaming methods


jmuk:
We want to support gRPC-streaming methods in Gapic codegen eventually, this issue is tracking the effort for it.

https://github.com/googleapis/toolkit/issues/322 will be the first step -- removing them at all. Then, we can start supporting them language-by-language effort.

jmuk:
Here will be the list of tasks.

- [ ] reveal response-streaming methods, as-is (i.e. interface is same as gRPC-gemerated ones)
- [ ] reveal request-streaming and bidi-streaming methods, as-is
- [ ] bring parameter-flattening / request-parameters for response-streaming method
- [ ] in case response stream actually iterates over chunks (i.e. chunk has a repeated field for actual resources), expand them as a stream of resources, as page-streaming does.

For request-streaming and bidi-streaming, it would be great to offer quite language idiomatic way to handle the input stream. For the time being, however, we will use the same interface as gRPC generates.

garrettjonesgoogle:
@jmuk It's probably worth circling back here and determining what work remains still.

jmuk:
Um. I feel like, basically I can close this, gRPC streaming is supported and works already (Sorry I've forgotten to close this).

There might be some chance to improve the existing surface, for example the one @chrisdunelm and you are discussing, but I think it's better to file an issue for each of such improvements.

PR #460, opened 2016-09-07
gRPC streaming support in Python.

jmuk:
This is an effort of #343, for Python:

- allow parameter flattening and retrying for server-side streaming
- the same interface as grpc-generated one for
  client-side streaming and bidi-streaming
- something like page-streaming might be helpful for some server-side
  streaming methods -- that's not yet implemented.

### DIFF
--- a/src/main/java/com/google/api/codegen/GapicContext.java
+++ b/src/main/java/com/google/api/codegen/GapicContext.java
@@ -112,21 +112,21 @@ public class GapicContext extends CodegenContext {
   }
 
   /**
-   * Returns true when the method should appear in the result of codegen. Currently only non
-   * stremaing methods are allowed by defualt, because majority of language clients do not support
-   * streaming APIs. Subclass may override this method to allow them.
+   * Returns true when the method is supported by the current codegen context. By default, only non
+   * stremaing methods are supported unless subclass explicitly allows.
+   * TODO: remove this method when all languages support gRPC streaming.
    */
-  protected boolean shouldMethodAppear(Method method) {
+  protected boolean isSupported(Method method) {
     return !method.getRequestStreaming() && !method.getResponseStreaming();
   }
 
   /**
    * Returns a list of RPC methods supported by the context.
    */
-  public List<Method> getMethods(Interface service) {
+  public List<Method> getSupportedMethods(Interface service) {
     List<Method> simples = new ArrayList<>(service.getMethods().size());
     for (Method method : service.getMethods()) {
-      if (shouldMethodAppear(method)) {
+      if (isSupported(method)) {
         simples.add(method);
       }
     }
@@ -136,9 +136,9 @@ public class GapicContext extends CodegenContext {
   /**
    * Returns a list of RPC methods supported by the context, taking into account GRPC interface
    * rerouting.
-   * TODO replace getMethods with this when all languages are migrated.
+   * TODO replace getSupportedMethods with this when all languages are migrated.
    */
-  public List<Method> getMethodsV2(Interface service) {
+  public List<Method> getSupportedMethodsV2(Interface service) {
     InterfaceConfig interfaceConfig = getApiConfig().getInterfaceConfig(service);
     if (interfaceConfig == null) {
       throw new IllegalStateException(
@@ -147,7 +147,7 @@ public class GapicContext extends CodegenContext {
     List<Method> methods = new ArrayList<>(interfaceConfig.getMethodConfigs().size());
     for (MethodConfig methodConfig : interfaceConfig.getMethodConfigs()) {
       Method method = methodConfig.getMethod();
-      if (shouldMethodAppear(method)) {
+      if (isSupported(method)) {
         methods.add(method);
       }
     }
--- a/src/main/java/com/google/api/codegen/GapicContext.java
+++ b/src/main/java/com/google/api/codegen/GapicContext.java
@@ -112,12 +112,21 @@ public class GapicContext extends CodegenContext {
   }
 
   /**
-   * Returns a list of simple RPC methods.
+   * Returns true when the method should appear in the result of codegen. Currently only non
+   * stremaing methods are allowed by defualt, because majority of language clients do not support
+   * streaming APIs. Subclass may override this method to allow them.
    */
-  public List<Method> getNonStreamingMethods(Interface service) {
+  protected boolean shouldMethodAppear(Method method) {
+    return !method.getRequestStreaming() && !method.getResponseStreaming();
+  }
+
+  /**
+   * Returns a list of RPC methods supported by the context.
+   */
+  public List<Method> getMethods(Interface service) {
     List<Method> simples = new ArrayList<>(service.getMethods().size());
     for (Method method : service.getMethods()) {
-      if (!method.getRequestStreaming() && !method.getResponseStreaming()) {
+      if (shouldMethodAppear(method)) {
         simples.add(method);
       }
     }
@@ -125,10 +134,11 @@ public class GapicContext extends CodegenContext {
   }
 
   /**
-   * Returns a list of simple RPC methods, taking into account GRPC interface rerouting.
-   * TODO replace getNonStreamingMethods with this when all languages are migrated.
+   * Returns a list of RPC methods supported by the context, taking into account GRPC interface
+   * rerouting.
+   * TODO replace getMethods with this when all languages are migrated.
    */
-  public List<Method> getNonStreamingMethodsV2(Interface service) {
+  public List<Method> getMethodsV2(Interface service) {
     InterfaceConfig interfaceConfig = getApiConfig().getInterfaceConfig(service);
     if (interfaceConfig == null) {
       throw new IllegalStateException(
@@ -137,7 +147,7 @@ public class GapicContext extends CodegenContext {
     List<Method> methods = new ArrayList<>(interfaceConfig.getMethodConfigs().size());
     for (MethodConfig methodConfig : interfaceConfig.getMethodConfigs()) {
       Method method = methodConfig.getMethod();
-      if (!method.getRequestStreaming() && !method.getResponseStreaming()) {
+      if (shouldMethodAppear(method)) {
         methods.add(method);
       }
     }
--- a/src/main/java/com/google/api/codegen/PageStreamingConfig.java
+++ b/src/main/java/com/google/api/codegen/PageStreamingConfig.java
@@ -94,7 +94,9 @@ public class PageStreamingConfig {
               resourcesFieldName));
     }
 
-    if (requestTokenField == null && responseTokenField == null && resourcesField == null) {
+    if ((!method.getResponseStreaming()
+            && (requestTokenField == null || responseTokenField == null))
+        || resourcesField == null) {
       return null;
     }
     return new PageStreamingConfig(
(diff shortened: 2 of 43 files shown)


src/main/java/com/google/api/codegen/PageStreamingConfig.java
```diff
@@ -94,7 +94,7 @@ public static PageStreamingConfig createPageStreaming(
               resourcesFieldName));
     }
 
-    if (requestTokenField == null || responseTokenField == null || resourcesField == null) {
+    if (requestTokenField == null && responseTokenField == null && resourcesField == null) {
```
geigerj: Why does gRPC streaming affect PageStreamingConfig? And what does this change mean? For regular page streaming, all three of these fields must not be null.

jmuk: Ah, right. Updated the condition.

I wanted to allow token fields as optional for grpc-streaming methods (and resources field is still required), so that condition wasn't correct.

The reason for optional token fields is that some grpc streaming methods are paged but no token fields exist (see https://github.com/googleapis/googleapis/blob/62ebd10737b35065a6995cfd25eeca81f1433f0f/google/genomics/v1/reads.proto#L458 for example).

Similarly but not quite same -- it's not apparent how to represent the page tokens for Bigtable's ReadRows request/responses (see https://github.com/googleapis/googleapis/blob/72ebf10ad5b2355df038e275a1543418ea8ac109/google/bigtable/v2/bigtable.proto#L96 but I can explain the details).

jmuk:
rebased and fixed the build failure.

src/main/java/com/google/api/codegen/py/PythonGapicContext.java
```diff
@@ -250,23 +256,30 @@ private String returnTypeComment(
...
-                "The maximum number of resources contained in the\n"
-                    + "underlying API response. If page streaming is performed per-\n"
-                    + "resource, this parameter does not affect the return value. If page\n"
-                    + "streaming is performed per-page, this determines the maximum number\n"
-                    + "of resources in a page."));
-      } else {
-        contentBuilder.append(fieldComment(name, field, importHandler, null));
+    if (method.getRequestStreaming()) {
+      contentBuilder.append(
+          "  requests (iterator of "
```
geigerj: "iterator of xxx" --> "Iterator[xxx]"

(https://www.python.org/dev/peps/pep-0484/)

src/main/java/com/google/api/codegen/csharp/CSharpGapicContext.java
```diff
@@ -596,7 +596,7 @@ private PageStreamerInfo getPageStreamerInfo(InterfaceConfig interfaceConfig, Me
     final InterfaceConfig interfaceConfig = getApiConfig().getInterfaceConfig(service);
     // TODO: Change back to .from(service.getMethods()) once streaming is implemented.
```
geigerj: Is it OK to remove this TODO now?

src/main/java/com/google/api/codegen/py/PythonDocConfig.java
```diff
@@ -54,14 +54,28 @@ public String getMethodName() {
...
-  public boolean isIterableResponse() {
+  public boolean isPageStreaming() {
     Method method = getMethod();
     Interface service = (Interface) method.getParent();
     MethodConfig methodConfig = getApiConfig().getInterfaceConfig(service).getMethodConfig(method);
     return methodConfig.isPageStreaming();
   }
 
   /**
+   * Does this method is a gRPC-response streaming?
```
geigerj: "Does this method is" --> "Is this method" (here and below)

geigerj:
Suggestion: make a new `grpcStreaming` config section that is independent of page streaming. Then you don't need conditions like `!method.getResponseStreaming() && config.getMethodConfig(method).isPageStreaming()` or changes to `PageStreamingConfig`. I think this makes sense because changes to GAX to support "chunk" stream flattening will likely need to be separate from the GAX code that handles conventional page streaming.

src/test/java/com/google/api/codegen/testdata/python_doc_main_library.baseline
```diff
@@ -879,6 +888,97 @@ class LibraryServiceApi(object):
...
+    def stream_shelves(
+            self,
+            options=None):
+        """
+        Test server streaming
+        gRPC streaming methods don't have HTTP equivalent and need not have google.api.http option.
+
+        Example:
+          >>> from library.library_service_api import LibraryServiceApi
+          >>> from google.gax import CallOptions, INITIAL_PAGE
```
geigerj: Shouldn't import `INITIAL_PAGE` here

src/test/java/com/google/api/codegen/testdata/python_doc_main_library.baseline
```diff
@@ -879,6 +888,97 @@ class LibraryServiceApi(object):
...
+        Test bidi-streaming.
+        gRPC streaming methods don't have HTTP equivalent and need not have google.api.http option.
+
+        EXPERIMENTAL: This method interface might change in the future.
+
+        Example:
+          >>> from library.library_service_api import LibraryServiceApi
+          >>> from google.example.library.v1 import library_pb2
+          >>> api = LibraryServiceApi()
+          >>> request = tagger_pb2.DiscussBookRequest()
```
geigerj: Where does `tagger_pb2` come from? It is not imported.

IMO, better to do `request = [tagger_pb2.DiscussBookRequest()]` -- this shows that the request is itself an iterator.

src/main/java/com/google/api/codegen/GapicContext.java
```diff
@@ -112,23 +112,33 @@ public CollectionConfig getCollectionConfig(Interface service, String entityName
...
    */
-  public List<Method> getNonStreamingMethods(Interface service) {
+  protected boolean shouldMethodAppear(Method method) {
+    return !method.getRequestStreaming() && !method.getResponseStreaming();
+  }
+
+  /**
+   * Returns a list of RPC methods supported by the context.
+   */
+  public List<Method> getMethods(Interface service) {
```
garrettjonesgoogle: maybe getSupportedMethods instead of just getMethods?

src/main/java/com/google/api/codegen/csharp/CSharpGapicContext.java
jmuk: No, not yet. It's saying to use `service.getMethods()` -- I think it is getting clearer since I've modified the method name to `getSupportedMethods` thanks to garrett's suggestion.

src/main/java/com/google/api/codegen/py/PythonDocConfig.java
jmuk: Done

src/test/java/com/google/api/codegen/testdata/python_doc_main_library.baseline
jmuk: It's gone

src/test/java/com/google/api/codegen/testdata/python_doc_main_library.baseline
jmuk: That is actually not a part of this, but there is a bug in PythonImportHandler which causes this and others. I've modified that.

Regarding `[tagger_pb2.DiscussBookRequest()]`: I get your point, but the things are a bit more complicated, and similar style is seen in `add_comments` example, so I think this is good enough.

src/main/java/com/google/api/codegen/GapicContext.java
jmuk: Done, thanks!

jmuk:
I've also modified the proto structures, so now it's `grpc_streaming` config instead of `page_streaming` (the internal structures are same currently though). PTAL.

src/main/java/com/google/api/codegen/py/PythonGapicContext.java
jmuk: Oh, I've forgotten to fix this. Done.

jmuk:
@geigerj, please take another look.

src/test/java/com/google/api/codegen/testdata/python_doc_main_library.baseline
```diff
@@ -879,6 +888,99 @@ class LibraryServiceApi(object):
...
+            self,
+            options=None):
+        """
+        Test server streaming
+        gRPC streaming methods don't have HTTP equivalent and need not have google.api.http option.
+
+        Example:
+          >>> from library.library_service_api import LibraryServiceApi
+          >>> api = LibraryServiceApi()
+          >>> requests = [request]
```
geigerj: This line doesn't work (here and below).

src/main/java/com/google/api/codegen/PageStreamingConfig.java
```diff
@@ -101,6 +101,80 @@ public static PageStreamingConfig createPageStreaming(
...
+            Diag.error(
+                SimpleLocation.TOPLEVEL,
+                "Request field missing for page streaming: method = %s, message type = %s, field = %s",
+                method.getFullName(),
+                method.getInputType().getMessageType().getFullName(),
+                requestTokenFieldName));
+      }
+    }
+
+    String pageSizeFieldName = pageStreaming.getRequest().getPageSizeField();
```
geigerj: Is page size actually needed/used for gRPC streaming?

src/test/java/com/google/api/codegen/testdata/python_doc_main_library.baseline
jmuk: Changed the template not to happen this (this line unexpectedly appeared)

src/main/java/com/google/api/codegen/PageStreamingConfig.java
jmuk: Not used -- removed this code.

src/main/java/com/google/api/codegen/PageStreamingConfig.java
```diff
@@ -101,6 +101,68 @@ public static PageStreamingConfig createPageStreaming(
...
+      DiagCollector diagCollector, PageStreamingConfigProto pageStreaming, Method method) {
+    String requestTokenFieldName = pageStreaming.getRequest().getTokenField();
+    Field requestTokenField = null;
+    if (!Strings.isNullOrEmpty(requestTokenFieldName)) {
+      requestTokenField = method.getInputType().getMessageType().lookupField(requestTokenFieldName);
+      if (requestTokenField == null) {
+        diagCollector.addDiag(
+            Diag.error(
+                SimpleLocation.TOPLEVEL,
+                "Request field missing for page streaming: method = %s, message type = %s, field = %s",
```
geigerj: s/page streaming/grpc streaming (here and below)

src/main/java/com/google/api/codegen/py/PythonGapicContext.java
```diff
@@ -209,15 +214,16 @@ private String returnTypeComment(
     String path = importHandler.elementPath(returnMessageType, true);
     String classInfo = ":class:`" + path + "` instance";
 
-    if (config.isPageStreaming()) {
+    if (method.getResponseStreaming()) {
+      return "Returns:\n" + "  An iterator which yields " + classInfo + "s.";
```
geigerj: "An iterator which yields ..." --> "iterator[...]"

jmuk:
Rebased. It still fails due to codecov reports, but I feel like this is not helpful (mostly not-covering error reports).

jmuk:
Okay, now the merge conflicts are resolved. PTAL.

src/main/java/com/google/api/codegen/PageStreamingConfig.java
jmuk: done

src/main/java/com/google/api/codegen/py/PythonGapicContext.java
jmuk: done

PR #466, opened 2016-09-08
Go: Only import "time" and "grpc/codes" if retrying

pongad:
Go previously imports "time" and "google.golang.org/grpc/codes"
packages unconditionally.
This causes build to fail if the file has no retry specified,
since retry is the only thing using these packages.
This patch makes us imports these packages only if at least one
retry is specified.

### DIFF
--- a/src/main/java/com/google/api/codegen/go/GoGapicContext.java
+++ b/src/main/java/com/google/api/codegen/go/GoGapicContext.java
@@ -39,6 +39,8 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import com.google.protobuf.DescriptorProtos.FieldDescriptorProto.Type;
 
+import io.grpc.Status;
+
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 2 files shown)


saicheems:
"This patch makes us imports these" -> "This patch makes us import these" in commit message

lgtm

src/main/java/com/google/api/codegen/go/GoGapicContext.java
```diff
@@ -494,19 +500,22 @@ private GoImport createMessageImport(MessageType messageType) {
...
     thirdParty.add(GoImport.create("golang.org/x/net/context"));
     thirdParty.add(GoImport.create("google.golang.org/grpc"));
-    thirdParty.add(GoImport.create("google.golang.org/grpc/codes"));
     thirdParty.add(GoImport.create("google.golang.org/grpc/metadata"));
     thirdParty.add(GoImport.create(GAX_PACKAGE_BASE, "gax"));
     thirdParty.add(GoImport.create("google.golang.org/api/option"));
     thirdParty.add(GoImport.create("google.golang.org/api/transport"));
 
+    if (standard.contains(GoImport.create("time"))) {
+      thirdParty.add(GoImport.create("google.golang.org/grpc/codes"));
```
vchudnov-g: Wouldn't it be better to key this include directly and independently off of whether retry is being used, rather than assuming that whatever uses "time" also uses "codes"? If we ever require "time" for any other reason, it will bring in "codes" too, potentially unnecessarily.

Maybe simply check once for retry and set a bool, and check the bool in both import setting locations?

pongad: This is a good point. I think the easiest solution is to inline the standard imports function. The function used to make sense separately, but with this kind of interaction, I think it's cleaner to bring the logic in one place.

pongad:
@vchudnov-g @saicheems PTAL.

vchudnov-g:
lg, but is there a test for this?

pongad:
@vchudnov-g I seem to have forgotten to commit the test file. PTAL.

vchudnov-g:
LGTM

PR #469, opened 2016-09-09
Infer Go import path from proto package

pongad:
We finally agreed on what to do with Go import paths!

The rule is that in a directory, either all proto files should
specify a go_package or none of them do.
If go_package is specified, all proto files that have the same
go_package belons in the same Go package.
Otherwise, all proto files in the directory belongs in the same
Go package.
This rule will be enforced from within Google;
toolkit doesn't have to enforce it.

If a file specify a go_package, we honor it by using the string
as the import path.
If not, we infer the import path from the proto's package.

The number of proto packages that have to specify their go_package's
should be small and limited to google/api
(and perhaps a couple other places).

This commit should alleviate the problem of having to specify
the go_package attribute in every single proto file.
We should be able to infer most paths correctly.


pongad:
NOTE TO REVIEWERS: Some specific details of createProtoImport might need to change depending on the syntax of go_package we settle on. Any comments on the structure of the change will be valuable nevertheless.

vchudnov-g:
Q: Please LMK offline where we capture this policy internally so I can refer to it.
Q: If the proto files in a directory specify go_package, do they all have to specify the same package or could the specify multiple packages as long as they each specify _some_ package?

vchudnov-g:
Looking good. Are there tests for the new functionality?

(Also, the "specific details" you mention in your previous comments will be part of this PR or will be handled separately?)

pongad:
@vchudnov-g I'll update this PR once the details are determined.

PR #470, opened 2016-09-09
Reroute_to_grpc_interface support for Python

### LINKED ISSUE #430: Implement reroute_to_grpc_interface for all languages (Pub/Sub IAM)
This functionality enables IAM for Pub/Sub.

Java/PHP were implemented in https://github.com/googleapis/toolkit/commit/b92d62397e895df0880fa982071a80f117c066c3 .
- [x] Node.js
- [x] Ruby
- [x] Python
- [x] Go
- [x] C#

@jmuk @bjwatson @chrisdunelm @saicheems @pongad

Landrito:
Resolves #430 for Python.


src/test/java/com/google/api/codegen/testdata/python_doc_main_library.baseline
```diff
@@ -36,8 +36,9 @@ import google.gax
 
 from google.example.library.v1 import field_mask_pb2 as v1_field_mask_pb2
 from google.example.library.v1 import library_pb2
-from google.example.library.v1 import tagger_pb2
+from google.example.library.v1 import tagger_pb2 as library_v1_tagger_pb2
```
geigerj: Per offline discussion, I don't think this logic is correct -- there is no such file.

Landrito: Fixed. Thanks for looking into this!

Landrito:
@geigerj This is ready for another pass!

src/main/java/com/google/api/codegen/py/PythonImportHandler.java
```diff
@@ -59,14 +61,15 @@ public PythonImportHandler(Interface service) {
...
+    for (MethodConfig methodConfig : apiConfig.getInterfaceConfig(service).getMethodConfigs()) {
+      Method method = methodConfig.getMethod();
       addImport(
           method.getFile(),
           PythonImport.create(
               ImportType.APP,
               method.getFile().getProto().getPackage(),
-              PythonProtoElements.getPbFileName(method.getInputMessage())));
-      for (Field field : method.getInputType().getMessageType().getMessageFields()) {
+              PythonProtoElements.getPbFileName(method)));
```
geigerj: What is this change for?

Consider an API which defines all of its RPCs (e.g., `GetFoo`) in a file `foo_service.proto` and all of its request messages (e.g., `GetFooRequest`) in a separate file `foo_service_messages.proto`.

IIUC, previously we would import the file containing the request message type, `foo_service_messages_pb2`, but now we import the file containing the RPC definition, `foo_service_pb2`.

That seems wrong to me, since in the codegen we construct the request message type but don't actually reference the RPC definition.

Landrito: Ah I see. This makes sense. I'll change this to create the input from the inputMessages protofile.

PR #471, opened 2016-09-09
Add getClient function stub in Go discovery gen

saicheems:
Adds a getClient stub function for non-ADC APIs so compilecheck
succeeds.


saicheems:
For Go

vchudnov-g:
LGTM

Amend PR name to clarify this is for Discovery

PR #476, opened 2016-09-12
Refactor Java discovery snippetgen for MVVM 

saicheems:
Wow, this change was an immense amount of work.

- Add a new SampleConfig class that contains an easy to traverse
  heirarchy of discovery types.
- Add a JSON based override mechanism that can be used to override any
  field within the SampleConfig or it's children.
- Support name collisions by storing computed names in a symbol table as
  they are generated.
- Refactor the Java discovery snippet to handle exceptions, and split
  out service generation code to facilitate alternative auth
  boilerplate.


src/main/java/com/google/api/codegen/discovery/MainDiscoveryProviderFactory.java
```diff
@@ -70,13 +72,19 @@ public static DiscoveryProvider defaultCreate(
...
+          .setSnippetSetRunner(new CommonSnippetSetRunner(new CommonRenderingUtil()))
+          .setMethodToViewTransformer(new JavaMethodToViewTransformer())
+          .setOverrides(sampleConfigOverrides)
+          .setOutputRoot(
+              "autogenerated/"
+                  + apiaryConfig.getApiName()
+                  + "/"
+                  + apiaryConfig.getApiVersion()
+                  + "/"
+                  + service.getDocumentation().getOverview())
```
garrettjonesgoogle: Isn't this output root the same pattern for all languages? It seems like it could be defined more centrally than this.

src/main/java/com/google/api/codegen/discovery/transformer/SampleTypeNameConverter.java
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Maps Type instances to TypeName instances.
+ */
+public interface SampleTypeNameConverter {
+
+  /**
+   * Provides a TypeName for the service.
+   */
+  TypeName getServiceTypeName(SampleConfig sampleConfig);
```
garrettjonesgoogle: This seems slightly out of place. It's not converting one type to another. My first instinct is to suggest putting it in SampleNamer.

src/main/java/com/google/api/codegen/util/SymbolTable.java
```diff
@@ -17,23 +17,59 @@
...
+      return desiredName;
+    }
+    return desiredName.join(suffix);
+  }
+
+  /**
+   * Returns a unique name, with a numeric suffix in case of conflicts.
+   *
+   * Not guaranteed to work as expected if used in combination with {@link
+   * #getNewSymbol(String)}.
```
garrettjonesgoogle: I think you mean `#getNewSymbol(Name)`.

src/main/java/com/google/api/codegen/util/SymbolTable.java
```diff
@@ -17,23 +17,59 @@
...
+    String suffix = getSuffix(desiredName);
+    return desiredName + suffix;
+  }
+
+  /**
+   *  Returns the next numeric suffix that makes desiredName unique.
+   *
+   *  Stores the joined desiredName/suffix in an internal map.
+   */
+  private String getSuffix(String desiredName) {
```
garrettjonesgoogle: maybe call this getAndSaveSuffix

src/main/java/com/google/api/codegen/discovery/transformer/java/JavaSampleTypeNameConverter.java
```diff
@@ -0,0 +1,205 @@
...
+    } else {
+      this.apiTypeName = Name.lowerCamel(sampleConfig.apiName()).toUpperCamel();
+    }
+    this.apiNameVersion = sampleConfig.apiNameVersion();
+    List<String> copy = new ArrayList<>(sampleConfig.methods().get(methodName).nameComponents());
+    // Convert the method name components to upper camel case.
+    for (int i = 0; i < copy.size(); i++) {
+      copy.set(i, Name.lowerCamel(copy.get(i)).toUpperCamel());
+    }
+    this.methodNameComponents = copy;
```
garrettjonesgoogle: All of the things in this constructor are specific to a particular type (the api type name), whereas other logic in this class is generic type logic (protobuf typing -> TypeName). I think a better place for the logic around particular types would be SampleNamer.

src/main/java/com/google/api/codegen/discovery/transformer/java/JavaSampleTypeNameConverter.java
```diff
@@ -0,0 +1,205 @@
...
+
+  /**
+   * Returns a TypeName from typeInfo.
+   */
+  @Override
+  public TypeName getTypeName(TypeInfo typeInfo) {
+    String typeName = "";
+    if (typeInfo.isMessage()) {
+      typeName =
+          String.join(
```
garrettjonesgoogle: I think this is only >= Java 8

src/main/java/com/google/api/codegen/discovery/transformer/java/JavaSampleTypeNameConverter.java
```diff
@@ -0,0 +1,205 @@
...
+    for (int i = 0; i < copy.size(); i++) {
+      copy.set(i, Name.lowerCamel(copy.get(i)).toUpperCamel());
+    }
+    this.methodNameComponents = copy;
+  }
+
+  @Override
+  public TypeName getServiceTypeName(SampleConfig sampleConfig) {
+    return typeNameConverter.getTypeName(
+        String.join(".", "com.google.api.services", apiNameVersion, apiTypeName));
```
garrettjonesgoogle: this pattern is repeated in multiple places; I recommend having it in only one place, and probably not in JavaSampleTypeNameConverter (it seems more specific)

src/main/java/com/google/api/codegen/discovery/transformer/java/JavaMethodToViewTransformer.java
```diff
@@ -0,0 +1,164 @@
...
+    sampleBodyView.requestBodyVarName("");
+    sampleBodyView.requestBodyTypeName("");
+    sampleBodyView.resourceGetterName("");
+    sampleBodyView.resourceTypeName("");
+    sampleBodyView.isResourceMap(false);
+
+    if (methodInfo.isPageStreaming()) {
+      FieldInfo fieldInfo = methodInfo.pageStreamingResourceField();
+      // TODO(garrettjones): Should I form this name this way?
+      sampleBodyView.resourceGetterName(Name.lowerCamel("get", fieldInfo.name()).toLowerCamel());
```
garrettjonesgoogle: no

garrettjonesgoogle: The logic should be in SampleNamer. See for example: https://github.com/googleapis/toolkit/blob/master/src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java#L153

src/main/java/com/google/api/codegen/discovery/transformer/java/JavaSampleTypeNameConverter.java
saicheems: Oh, yeah 😞 
Will fix

src/main/java/com/google/api/codegen/discovery/transformer/java/JavaSampleTypeNameConverter.java
saicheems: It's a bit tricky... these values are used in generating all message typenames and the api typename.

Since the api type and request type in java start with
`com.google.api.services.{apiname}.{apiversion}.{apitypename}[.{components}]`

And all other messages start with
`com.google.api.services.{apiname}.{apiversion}.model.{messagetypename}`

I thought it made sense to store all those values in the constructor, since they're important to generating fully qualified paths.

I think you're talking about moving the naming logic (lower -> upper) into samplenamer, which makes sense. I'll do that.

src/main/java/com/google/api/codegen/discovery/transformer/SampleTypeNameConverter.java
saicheems: Well, it does return a fully qualified path for a type... I didn't think it belonged in SampleNamer

src/main/java/com/google/api/codegen/discovery/transformer/java/JavaSampleTypeNameConverter.java
saicheems: It should be only these two methods - the reason it's repeated is that the semantics of import names between service types/request messages is different than other messages in the java apiary libraries

Regardless, I'll see if I can reduce the stuff in common.

src/main/java/com/google/api/codegen/discovery/transformer/java/JavaSampleTypeNameConverter.java
garrettjonesgoogle: In GAPIC, the package is bundled with the type that we parse from the proto. Maybe you could do the same thing in spirit, and have the package string bundled in TypeInfo.

src/main/java/com/google/api/codegen/discovery/MainDiscoveryProviderFactory.java
saicheems: It's a bit tricky, because of that `service.getDocumentation` bit. I'd like to leave it as is, until all 6 languages are complete. Then we can move it more easily.

saicheems:
Okay, @garrettjonesgoogle I've moved all type name generation logic into the `ApiaryConfigToSampleConfigConverter` where type name generation is facilitated by `TypeNameGenerator` and `JavaTypeNameGenerator`

saicheems:
I'm doing some last passes to clean up sections of code here and there and adding more elaboration to docstrings.

Things to do:
- Include a test that makes use of the overrides functionality
- Clean up
Things done:
- Everything else

Otherwise, please begin the review - If we can start with structural comments first, and nitpicks in phase 2 that will help me keep things straight :)

src/main/java/com/google/api/codegen/discovery/config/ApiaryConfigToSampleConfigConverter.java
```diff
@@ -219,7 +243,7 @@ private static boolean isPageStreaming(Method method, ApiaryConfig apiaryConfig)
   /**
    * Returns the resource field of a page streaming response type.
    */
-  private static Field getPageStreamingResouceField(Type type) {
+  private Field getPageStreamingResouceField(Type type) {
```
garrettjonesgoogle: spelling (Resouce)

src/main/java/com/google/api/codegen/discovery/transformer/java/JavaSampleTypeNameConverter.java
```diff
@@ -14,22 +14,19 @@
...
-import com.google.api.codegen.util.Name;
 import com.google.api.codegen.util.TypeName;
 import com.google.api.codegen.util.TypeNameConverter;
 import com.google.api.codegen.util.TypedValue;
 import com.google.api.codegen.util.java.JavaTypeTable;
-import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableMap;
 import com.google.protobuf.Field;
 
+import autovalue.shaded.com.google.common.common.base.Joiner;
```
garrettjonesgoogle: Don't use shaded types

src/main/java/com/google/api/codegen/discovery/transformer/java/JavaSampleTypeNameConverter.java
```diff
@@ -67,84 +64,39 @@
...
     if (typeInfo.isMessage()) {
-      typeName =
-          String.join(
-              ".",
-              "com.google.api.services",
-              apiNameVersion,
-              "model",
-              typeInfo.message().typeName());
+      return typeNameConverter.getTypeName(
+          Joiner.on('.').join(packagePrefix, "model", typeInfo.message().typeName()));
```
garrettjonesgoogle: Maybe make a subpackage attribute in TypeInfo, and set it to "model" earlier in the process, so that it doesn't have to be encoded here.

src/test/java/com/google/api/codegen/DiscoveryGeneratorTestBase.java
```diff
@@ -81,9 +81,10 @@ protected Object run() {
     String factory = generator.getFactory();
     String id = generator.getId();
 
+    // TODO(saicheems): Fix this!
```
garrettjonesgoogle: Be more specific about the "fixing" that is necessary

src/main/resources/com/google/api/codegen/java/discovery_fragment.snip
```diff
@@ -1,274 +1,114 @@
...
-  {@context.addImport("com.google.api.client.googleapis.auth.oauth2.GoogleCredential")}
-  {@context.addImport("com.google.api.client.googleapis.javanet.GoogleNetHttpTransport")}
-  {@context.addImport("com.google.api.client.http.HttpTransport")}
-  {@context.addImport("com.google.api.client.json.JsonFactory")}
-  {@context.addImport("com.google.api.client.json.jackson2.JacksonFactory")}
-@end
-
-@snippet generateMethodSampleCode(docConfig)
-@end
+@end
```
garrettjonesgoogle: add newline

src/main/java/com/google/api/codegen/util/SymbolTable.java
```diff
@@ -17,23 +17,59 @@
 import java.util.HashSet;
 import java.util.Set;
 
+import com.google.api.client.util.Strings;
```
garrettjonesgoogle: use com.google.common.base.Strings

src/main/java/com/google/api/codegen/discovery/viewmodel/SampleView.java
```diff
@@ -0,0 +1,76 @@
...
+  @Override
+  public abstract String templateFileName();
+
+  @Override
+  public abstract String outputPath();
+
+  @Override
+  public String resourceRoot() {
+    // TODO(saicheems): Needed?
+    return SnippetSetRunner.SNIPPET_RESOURCE_ROOT;
```
garrettjonesgoogle: We might be able to nuke this

src/main/java/com/google/api/codegen/discovery/transformer/java/JavaMethodToViewTransformer.java
```diff
@@ -0,0 +1,168 @@
...
+        throw new IllegalArgumentException(
+            "method is page streaming, but the page streaming resource field is null.");
+      }
+      sampleBodyView.resourceGetterName(sampleNamer.getResourceGetterName(fieldInfo.name()));
+      String nickname = sampleTypeTable.getAndSaveNicknameFor(fieldInfo.type());
+      sampleBodyView.resourceTypeName(nickname);
+      // Rename the type from Map<K, V> to Map.Entry<K, V> to match what we
+      // expect in the for-loop.
+      if (fieldInfo.type().isMap()) {
+        sampleBodyView.resourceTypeName(sampleNamer.getMapEntryTypeFromMapType(nickname));
```
garrettjonesgoogle: It seems like getMapEntryTypeFromMapType should live in JavaSampleTypeNameConverter instead of SurfaceNamer, and it should take fieldInfo.type() as input.

src/main/java/com/google/api/codegen/discovery/transformer/MethodToViewTransformer.java
```diff
@@ -0,0 +1,27 @@
...
+package com.google.api.codegen.discovery.transformer;
+
+import com.google.api.codegen.discovery.config.SampleConfig;
+import com.google.api.codegen.viewmodel.ViewModel;
+import com.google.protobuf.Method;
+
+/*
+ * Transforms a Method into a ViewModel instance that can be rendered by a template engine.
+ */
+public interface MethodToViewTransformer {
```
garrettjonesgoogle: Maybe call SampleMethodToViewTransformer

src/main/java/com/google/api/codegen/discovery/config/ApiaryConfigToSampleConfigConverter.java
```diff
@@ -0,0 +1,255 @@
...
+      List<Method> methods, ApiaryConfig apiaryConfig, TypeNameGenerator typeNameGenerator) {
+    this.methods = methods;
+    this.apiaryConfig = apiaryConfig;
+    this.typeNameGenerator = typeNameGenerator;
+
+    methodNameComponents = new HashMap<String, List<String>>();
+    // Pull all the methodNameComponents out.
+    for (Method method : methods) {
+      String methodName = method.getName();
+      LinkedList<String> nameComponents = new LinkedList<>(Arrays.asList(methodName.split("\\.")));
```
garrettjonesgoogle: you could simplify this with Lists.newLinkedList(methodName...)

src/main/java/com/google/api/codegen/discovery/config/ApiaryConfigToSampleConfigConverter.java
```diff
@@ -0,0 +1,255 @@
...
+        && !responseTypeUrl.equals(EMPTY_TYPE_URL)) {
+      responseType = createTypeInfo(method, false);
+    }
+
+    boolean isPageStreaming = isPageStreaming(method, apiaryConfig);
+    FieldInfo pageStreamingResourceField = null;
+    if (isPageStreaming) {
+      Field field = getPageStreamingResouceField(apiaryConfig.getType(responseTypeUrl));
+      // If field is null, then the page streaming resource field is not
+      // repeated. We allow null to be stored, and leave it to the user to
```
garrettjonesgoogle: who is the user?

src/main/java/com/google/api/codegen/discovery/ViewModelProvider.java
```diff
@@ -0,0 +1,172 @@
...
+    if (sampleConfigOverrides != null) {
+      ObjectMapper mapper = new ObjectMapper();
+      JsonNode tree = mapper.valueToTree(sampleConfig);
+      merge((ObjectNode) tree, (ObjectNode) sampleConfigOverrides);
+      try {
+        sampleConfig = mapper.treeToValue(tree, SampleConfig.class);
+      } catch (Exception e) {
+        throw new RuntimeException("failed to parse config to node, this should never happen");
+      }
+    }
```
garrettjonesgoogle: Maybe to achieve a "balance of complexity" in this method, move the code to create the SampleConfig into a helper method; that way, this method just calls out to three things: 1. generate SampleConfig, 2. transform, 3. snippetSetRunner.generate

src/main/java/com/google/api/codegen/discovery/config/ApiaryConfigToSampleConfigConverter.java
saicheems: Clarified

src/main/java/com/google/api/codegen/discovery/ViewModelProvider.java
saicheems: Done

src/main/java/com/google/api/codegen/discovery/transformer/MethodToViewTransformer.java
saicheems: Done

src/main/java/com/google/api/codegen/discovery/viewmodel/SampleView.java
saicheems: How?

src/main/java/com/google/api/codegen/util/SymbolTable.java
saicheems: Done
This is what I get for using Eclipse's auto importer :/

src/main/resources/com/google/api/codegen/java/discovery_fragment.snip
saicheems: Done

src/test/java/com/google/api/codegen/DiscoveryGeneratorTestBase.java
saicheems: Clarified, will be resolved before this is merged

src/main/java/com/google/api/codegen/discovery/config/ApiaryConfigToSampleConfigConverter.java
```diff
@@ -20,8 +20,10 @@
 import java.util.List;
 import java.util.Map;
 
+import com.beust.jcommander.internal.Lists;
```
garrettjonesgoogle: use Guava Lists

src/main/java/com/google/api/codegen/discovery/config/ApiaryConfigToSampleConfigConverter.java
```diff
@@ -47,24 +49,25 @@ public ApiaryConfigToSampleConfigConverter(
     this.typeNameGenerator = typeNameGenerator;
 
     methodNameComponents = new HashMap<String, List<String>>();
-    // Pull all the methodNameComponents out.
+    // Since methodNameComponents are used to generate the request type name, we
+    // produce them here for ease-of-access.
     for (Method method : methods) {
       String methodName = method.getName();
-      LinkedList<String> nameComponents = new LinkedList<>(Arrays.asList(methodName.split("\\.")));
+      LinkedList<String> nameComponents = new LinkedList<String>(Arrays.asList(methodName.split("\\.")));
```
garrettjonesgoogle: Did Lists.newLinkedList not work?

src/main/java/com/google/api/codegen/discovery/config/ApiaryConfigToSampleConfigConverter.java
```diff
@@ -216,24 +224,27 @@ private MessageTypeInfo createMessageTypeInfo(
     Map<String, FieldInfo> fields = new HashMap<>();
     if (deep) {
       for (Field field2 : type.getFieldsList()) {
-        fields.put(field2.getName(), createFieldInfo(field2, method, apiaryConfig));
+        fields.put(field2.getName(), createFieldInfo(field2, method));
       }
     }
-    return MessageTypeInfo.newBuilder().typeName(typeName).fields(fields).build();
+    return MessageTypeInfo.newBuilder().typeName(typeName).subpackage(typeNameGenerator.getSubpackage(false)).fields(fields).build();
```
garrettjonesgoogle: Line is too long (shouldn't this fail the formatting verification?)

src/main/java/com/google/api/codegen/discovery/config/ApiaryConfigToSampleConfigConverter.java
saicheems: Errr... I tried but it still required that `Arrays.asList...` part, and it wasn't really any shorter.

garrettjonesgoogle:
The PR-based approval isn't an LGTM yet - I'm still getting used to this new flow.

src/main/java/com/google/api/codegen/discovery/transformer/java/JavaMethodToViewTransformer.java
saicheems: Done - I've put `getTypeNameForElementType` back.

src/main/java/com/google/api/codegen/discovery/config/ApiaryConfigToSampleConfigConverter.java
saicheems: Done

src/main/java/com/google/api/codegen/discovery/config/ApiaryConfigToSampleConfigConverter.java
saicheems: My mistake, didn't run the formatter before the last push

garrettjonesgoogle: My recommendation: Always run "./gradlew googleJavaFormat build"

src/main/java/com/google/api/codegen/discovery/transformer/java/JavaSampleTypeNameConverter.java
```diff
@@ -91,11 +102,29 @@ public TypeName getTypeName(TypeInfo typeInfo) {
       if (!Strings.isNullOrEmpty(messageInfo.subpackage())) {
         sb.append('.').append(messageInfo.subpackage());
       }
-      return typeNameConverter.getTypeName(sb.append('.').append(messageInfo.typeName()).toString());
+      return typeNameConverter.getTypeName(
+          sb.append('.').append(messageInfo.typeName()).toString());
```
garrettjonesgoogle: Couldn't you use the packagePrefix(String) helper in this method?

src/main/java/com/google/api/codegen/discovery/config/java/JavaTypeNameGenerator.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+package com.google.api.codegen.discovery.config.java;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import com.google.api.codegen.discovery.config.TypeNameGenerator;
+import com.google.api.codegen.util.Name;
+
+import autovalue.shaded.com.google.common.common.base.Joiner;
```
garrettjonesgoogle: Don't use the shaded type

src/main/java/com/google/api/codegen/discovery/transformer/SampleNamer.java
```diff
@@ -0,0 +1,49 @@
...
+
+  public String getFieldVarName(String fieldName) {
+    return varName(Name.lowerCamel(fieldName));
+  }
+
+  public String getResourceGetterName(String fieldName) {
+    return methodName(Name.lowerCamel("get", fieldName));
+  }
+
+  public String getMapEntryTypeFromMapType(String nickname) {
```
garrettjonesgoogle: Remove this

saicheems: Done

src/main/java/com/google/api/codegen/discovery/config/java/JavaTypeNameGenerator.java
saicheems: Done

src/main/java/com/google/api/codegen/discovery/transformer/java/JavaSampleTypeNameConverter.java
saicheems: Done (good call)

garrettjonesgoogle:
LGTM (assuming tests pass)

src/main/java/com/google/api/codegen/util/SymbolTable.java
saicheems: Fixed

src/test/java/com/google/api/codegen/DiscoveryGeneratorTestBase.java
```diff
@@ -53,6 +60,23 @@ public DiscoveryGeneratorTestBase(
     this.name = name;
     this.discoveryDocFileName = discoveryDocFileName;
     this.gapicConfigFileNames = gapicConfigFileNames;
+
+    // Look for a overrides json based on the filename of the discovery doc.
```
saicheems: This is the mechanism I've decided on to test overrides:

It looks to see if there's a file with an overrides extension matching the discovery doc and applies that to the generator.

vchudnov-g:
Can you suggest a logical order to review these files? (Esp for someone who hasn't coded this MVVM.)

src/main/java/com/google/api/codegen/discovery/viewmodel/SampleView.java
vchudnov-g: What does this signify?

src/main/java/com/google/api/codegen/transformer/java/JavaModelTypeNameConverter.java
```diff
@@ -64,7 +64,7 @@
           .build();
 
   /**
-   * A map from primitive types in proto to zero value in Java
+   * A map from primitive types in proto to zero value in Java.
```
vchudnov-g: nit: s/value/values/

src/main/java/com/google/api/codegen/util/CommonRenderingUtil.java
```diff
@@ -24,13 +23,62 @@
...
       result.add(line);
     }
     return result;
   }
+
+  /**
+   * Returns the input text split on newlines and maxWidth.
+   */
+  public static List<String> getDocLines(String text, int maxWidth) {
+    maxWidth = maxWidth - 1;
```
vchudnov-g: Given this subtraction, please clarify the semantics of maxWidth.

src/main/java/com/google/api/codegen/util/SymbolTable.java
```diff
@@ -17,23 +17,59 @@
...
+   *  Returns the next numeric suffix that makes desiredName unique.
+   *
+   *  Stores the joined desiredName/suffix in an internal map.
+   */
+  private String getAndSaveSuffix(String desiredName) {
+    if (!symbolTable.contains(desiredName)) {
+      symbolTable.add(desiredName);
+      return "";
+    }
     int i = 2;
```
vchudnov-g: Why does this start at 2?

src/test/java/com/google/api/codegen/DiscoveryGeneratorTestBase.java
vchudnov-g: sg.

src/test/java/com/google/api/codegen/DiscoveryGeneratorTestBase.java
```diff
@@ -78,12 +99,13 @@ protected void setupDiscovery() {
   protected Object run() {
     GeneratorProto generator = config.getGenerator();
 
-    String factory = generator.getFactory();
     String id = generator.getId();
 
+    // TODO(saicheems): Fix this!
+    // Specifically, create a test that utilizes the overrides functionality.
```
vchudnov-g: Could you do this as part of this PR?

src/test/java/com/google/api/codegen/DiscoveryGeneratorTestBase.java
```diff
@@ -53,6 +58,22 @@ public DiscoveryGeneratorTestBase(
     this.name = name;
     this.discoveryDocFileName = discoveryDocFileName;
     this.gapicConfigFileNames = gapicConfigFileNames;
+
```
vchudnov-g: Does this file automatically generate the snippets from the checked-in discoveries and compare the output to the baselines?

src/main/java/com/google/api/codegen/discovery/viewmodel/SampleView.java
garrettjonesgoogle: This is the package path that the snippet files (*.snip) are located relative to. For MVVM, it is always 'com/google/api/codegen'. We could remove the resourceRoot() method from ViewModel and all derived classes, and in CommonSnippetSetRunner, pass SnippetSetRunner.SNIPPET_RESOURCE_ROOT instead of input.resourceRoot().

src/main/java/com/google/api/codegen/discovery/ViewModelProvider.java
```diff
@@ -0,0 +1,191 @@
...
+   *
+   * Values of tree are modified, and values of overrideTree are not modified.
+   * The merge process loops through the fields of tree and replaces non-object
+   * values with the corresponding value from overrideTree if present. Object
+   * values are traversed recursively to replace sub-properties.
+   *
+   * @param tree the original JsonNode to modify.
+   * @param overrideTree the JsonNode with values to overwrite tree with.
+   */
+  private static void merge(ObjectNode tree, ObjectNode overrideTree) {
```
saicheems: So the overrides file should match the format of the sample config itself.
The general properties you list are to root level properties in SampleConfig.
```
{
  "apiTitle": "Logging API",
  "apiName": "logging",
  "apiVersion": "v2",
  "apiTypeName": "Logging",
  "packagePrefix": "foo.google.bar.logging.v2",
  "methods" {
    "logging.list": {
      "nameComponents": ["foo", "bar", "list"]
      "requestType": {
        "message": {
          "typeName": "ListRequest"
          "subpackage": "model"
        }
      }
    }
  }
```
Above is an example of an overrides file. The root level properties are applied to all type names in the `SampleMethodToViewTransformer`, so the type name of `ListRequest` will be constituted as:
`{packagePrefix}.{subpackage}.{requestTypeName}`
`foo.google.bar.logging.v2.model.ListRequest`

The service type name is constituted as:
`{packagePrefix}.{apiTypeName}`
`foo.google.bar.logging.v2.Logging`

src/main/java/com/google/api/codegen/discovery/ViewModelProvider.java
```diff
@@ -0,0 +1,191 @@
...
+    Iterator<String> fieldNames = overrideTree.fieldNames();
+    while (fieldNames.hasNext()) {
+      String fieldName = fieldNames.next();
+      JsonNode primaryValue = tree.get(fieldName);
+      if (primaryValue == null) {
```
saicheems: It could be an error - a null node here means that the field in the overrides tree wasn't found in the primary tree. I didn't want the override functionality to introduce new nodes in to the primary tree that didn't already exist. I don't think it particularly changes anything to do so, but the function is only intended to replace values within JSONObjects, and not anything else. If we stick to that functionality only, we can't produce an invalid `SampleConfig` tree if a config contains invalid nodes.

src/test/java/com/google/api/codegen/DiscoveryGeneratorTestBase.java
saicheems: Yes

src/test/java/com/google/api/codegen/DiscoveryGeneratorTestBase.java
saicheems: It's ready, I forgot to remove this TODO, will remove

src/main/java/com/google/api/codegen/util/SymbolTable.java
saicheems: Idea is:
`getSymbol("foo")` -> `"foo"`
`getSymbol("foo")` -> `"foo2"`

I'll make a comment to explain

src/main/java/com/google/api/codegen/discovery/ViewModelProvider.java
saicheems: So if we change the `apiTypeName` or `packagePrefix`, that will carry to all types when they're combined in `SampleMethodToViewTransformer`

src/main/java/com/google/api/codegen/util/CommonRenderingUtil.java
saicheems: I copied this code from context common I think...
I suspect it's a simple trick to make sure lines break on char 79 instead of char 80.

I'll verify and comment.

src/main/java/com/google/api/codegen/util/SymbolTable.java
saicheems: Done

src/main/java/com/google/api/codegen/util/CommonRenderingUtil.java
saicheems: Done

src/main/java/com/google/api/codegen/transformer/java/JavaModelTypeNameConverter.java
saicheems: Done

src/main/java/com/google/api/codegen/discovery/config/ApiaryConfigToSampleConfigConverter.java
```diff
@@ -0,0 +1,270 @@
...
+   * Creates a method.
+   */
+  private MethodInfo createMethod(Method method) {
+    Map<String, FieldInfo> fields = new HashMap<>();
+    TypeInfo requestBodyType = null;
+    for (String fieldName : apiaryConfig.getMethodParams(method.getName())) {
+      Field field =
+          apiaryConfig.getField(apiaryConfig.getType(method.getRequestTypeUrl()), fieldName);
+      // If one of the method arguments is a Message, we parse that separately
+      // as the request body.
```
saicheems: Done

src/main/java/com/google/api/codegen/discovery/config/ApiaryConfigToSampleConfigConverter.java
```diff
@@ -0,0 +1,270 @@
...
+
+  public ApiaryConfigToSampleConfigConverter(
+      List<Method> methods, ApiaryConfig apiaryConfig, TypeNameGenerator typeNameGenerator) {
+    this.methods = methods;
+    this.apiaryConfig = apiaryConfig;
+    this.typeNameGenerator = typeNameGenerator;
+
+    methodNameComponents = new HashMap<String, List<String>>();
+    // Since methodNameComponents are used to generate the request type name, we
+    // produce them here for ease-of-access.
```
saicheems: Done

src/test/java/com/google/api/codegen/DiscoveryGeneratorTestBase.java
saicheems: Done

saicheems:
I've added two files:
`test/java/com/google/api/codegen/testdata/discoveries/adexchangebuyer.v1.4.json.overrides`
`test/java/com/google/api/codegen/testdata/discoveries/java/java_adexchangebuyer.v1.4.json`

This demonstrates the overrides file, it's test, and also the format of the client.

src/main/java/com/google/api/codegen/discovery/MainDiscoveryProviderFactory.java
saicheems: sgtm

src/main/java/com/google/api/codegen/discovery/ViewModelProvider.java
vchudnov-g: sg. So, 
(TL;DR)^2: each component that we expect could possibly be overridden only appears once in the overrides file/`SampleConfig`. The actual identifiers that we construct in the samples are built from these components. Correct?

It would be good to capture the explanation you provided somewhere obvious. Maybe as part of the class comment of the Discovery `SampleConfig`?

src/main/java/com/google/api/codegen/discovery/ViewModelProvider.java
vchudnov-g: Right, that's my point. Why not make this an error (exception), to explicitly point out that the user is trying to override a nonexistent node? What's the thinking behind ignoring and silently continuing?

src/main/java/com/google/api/codegen/util/CommonRenderingUtil.java
vchudnov-g: Could you add the comment (eg, "`maxWidth` includes the ending newline)?

src/main/java/com/google/api/codegen/util/SymbolTable.java
vchudnov-g: Ah, makes sense. But I don't see the comment. Did you upload a new commit?

src/main/java/com/google/api/codegen/util/CommonRenderingUtil.java
saicheems: I must've overridden the commit with my last comment changes :(
Will update shortly.

src/main/java/com/google/api/codegen/discovery/ViewModelProvider.java
saicheems: Yes, correct.
Sure, will do

src/main/java/com/google/api/codegen/discovery/ViewModelProvider.java
saicheems: One reason we might want to ignore is because the SampleConfig contains only one method:

generated sampleconfig:
```
  ...
  "methods" {
    // Only one method
  }
```

overrides:
```
  ...
  "methods": {
    // Many methods
  }
```

We'd want to ignore those extra methods (fields) that overrides has

src/main/java/com/google/api/codegen/util/SymbolTable.java
saicheems: Done

src/main/java/com/google/api/codegen/discovery/ViewModelProvider.java
saicheems: Done

src/main/java/com/google/api/codegen/util/CommonRenderingUtil.java
saicheems: Done

saicheems:
Okay, commits added, sorry about that

src/main/java/com/google/api/codegen/discovery/ViewModelProvider.java
vchudnov-g: Makes sense. Could you do the following?:

- add a brief comment saying that `SampleConfig` currently contains only one method at a time while the overrides file could have overrides for many methods
- file an issue so we remember to modify `SampleConfig` so it can contain multiple methods at once.

PR #477, opened 2016-09-12
NodeJS: Remove CallOptions as a class, allow Objects.

jmuk:
Fixes https://github.com/googleapis/gax-nodejs/issues/35.


jmuk:
See also: https://github.com/googleapis/gax-nodejs/pull/37

src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
```diff
@@ -332,9 +346,15 @@ private String returnTypeComment(Method method, MethodConfig config) {
     for (Field field : config.getRequiredFields()) {
       paramTypesBuilder.append(fieldParamComment(field, null, false));
     }
+    paramTypesBuilder.append(
+        "@param {Object=} options\n"
+            + "  Optional parameters. You can override the default settings for this call, e.g, timeout,\n"
+            + "  retries, paginations, etc. See [gax.CallOptions]{@link "
```
geigerj: This link does not work for me. (It pulls up a page, but there is no `#CallOptions` section.)

src/test/java/com/google/api/codegen/testdata/nodejs_doc_main_library.baseline
```diff
@@ -292,16 +291,18 @@ LibraryServiceApi.prototype.matchBookFromArchivedBookName =
...
-  var args = arguejs({
-    shelf: Object,
-    options: [gax.CallOptions],
-    callback: [Function]
-  }, arguments);
+LibraryServiceApi.prototype.createShelf = function createShelf(
+    shelf,
+    options,
+    callback) {
+  if (options instanceof Function && callback === undefined) {
```
geigerj: This looks strange to my eyes, and it produces a lot of repeated code. Is this more JS-idiomatic than just forcing people to pass an empty object as `options` if they want to specify `callback` but not `options`?

src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
```diff
@@ -332,9 +346,15 @@ private String returnTypeComment(Method method, MethodConfig config) {
     for (Field field : config.getRequiredFields()) {
       paramTypesBuilder.append(fieldParamComment(field, null, false));
     }
+    paramTypesBuilder.append(
+        "@param {Object=} options\n"
+            + "  Optional parameters. You can override the default settings for this call, e.g, timeout,\n"
```
geigerj: Is it possible to make it more clear that `foo` is a possible property of `options` iff `foo` is a property of `CallOptions`, as opposed to there being a property of `options` which is an object of type `CallOptions`?

jmuk: Added to the list of keyword -- so it can be distinguished.

jmuk: Added to GAX.

PR #478, opened 2016-09-13
Fix Ruby package file path name

swcloud:
Ruby package file path name should be lowercase original instead of lower underscore.

For example, 

package_name: Google::Cloud::ErrorReporting::V1beta1

should map to

google/cloud/errorreporting/v1beta1

instead of

google/cloud/error_reporting/v1beta1

### DIFF
--- a/src/main/java/com/google/api/codegen/util/ruby/RubyNameFormatter.java
+++ b/src/main/java/com/google/api/codegen/util/ruby/RubyNameFormatter.java
@@ -65,7 +65,7 @@ public class RubyNameFormatter implements NameFormatter {
 
   @Override
   public String packageFilePathPiece(Name name) {
-    return name.toLowerUnderscore();
+    return name.toOriginal().toLowerCase();
   }
 
   @Override


src/main/java/com/google/api/codegen/util/ruby/RubyNameFormatter.java
```diff
@@ -65,7 +65,7 @@ public String qualifiedName(NamePath namePath) {
 
   @Override
   public String packageFilePathPiece(Name name) {
-    return name.toLowerUnderscore();
+    return name.toOriginal().toLowerCase();
```
jmuk: Sorry, probably that's because I'm not familiar with this class, what does the `toOriginal` method do?

jmuk:
Hmm, wait a minute.
According to the "require statement" in http://guides.rubygems.org/name-your-gem/, the file path is better to be `error_reporting`, with underscore. Am I wrong?

swcloud:
Somehow I thought only the gem name should be lower underscore for multiple words, but it looks like the module name should also be the case. So in our delivery, the directory structure should look like

google/cloud/error_reporting/v1beta1

I'll close this PR then.

PR #479, opened 2016-09-13
Google::Gax::VERSION was being called incorrectly in Ruby.

Landrito:
[VERSION](https://github.com/googleapis/gax-ruby/blob/master/lib/google/gax/version.rb#L32) is all capitals in gax-ruby.


jmuk:
Whoa, thanks for doing this. LGTM.

Landrito:
Np!

PR #480, opened 2016-09-14
Fixed python imports for sample-gen

### LINKED ISSUE #473: Python sample-gen import statements working incorrectly for imported messages.
**What:**

The import statement for Python sample-gen is incorrect for transitive protobuf message(s). 

**Why:**

This seems to be because the method_sample.snip is using the import handler for the main snippet file to generate the import as opposed to using one that is specific to the method.

Landrito:
Resolves #473.

### DIFF
--- a/src/main/java/com/google/api/codegen/py/PythonImportHandler.java
+++ b/src/main/java/com/google/api/codegen/py/PythonImportHandler.java
@@ -94,13 +94,19 @@ public class PythonImportHandler {
     }
   }
 
-  public PythonImportHandler(Method method) {
-    addImport(
-        method.getFile(),
-        PythonImport.create(
-            ImportType.APP,
-            method.getFile().getProto().getPackage(),
-            PythonProtoElements.getPbFileName(method.getInputMessage())));
+  public PythonImportHandler(Iterable<Field> requiredFields) {
+    for (Field field : requiredFields) {
+      if (!field.getType().isMessage()) {
+        continue;
+      }
+      MessageType messageType = field.getType().getMessageType();
+      addImport(
+          messageType.getFile(),
+          PythonImport.create(
+              ImportType.APP,
+              messageType.getFile().getProto().getPackage(),
+              PythonProtoElements.getPbFileName(messageType)));
+    }
   }
 
   // Independent import handler to support fragment generation from discovery sources
--- a/src/main/java/com/google/api/codegen/py/PythonImportHandler.java
+++ b/src/main/java/com/google/api/codegen/py/PythonImportHandler.java
@@ -45,6 +45,7 @@ public class PythonImportHandler {
    */
   private final BiMap<ProtoFile, String> fileImports = HashBiMap.create();
 
+  /** This constructor is for the main imports of a generated service file */
   public PythonImportHandler(Interface service) {
     // Add non-service-specific imports.
     addImportStandard("json");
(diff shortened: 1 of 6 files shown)


geigerj:
Nice! Out of curiosity, do you know where the `tagger_pb2` in the old code was coming from?

src/main/java/com/google/api/codegen/py/PythonImportHandler.java
```diff
@@ -94,13 +94,19 @@ public PythonImportHandler(ProtoFile file) {
...
   }
 
-  public PythonImportHandler(Method method) {
-    addImport(
-        method.getFile(),
-        PythonImport.create(
-            ImportType.APP,
-            method.getFile().getProto().getPackage(),
-            PythonProtoElements.getPbFileName(method.getInputMessage())));
+  public PythonImportHandler(Iterable<Field> requiredFields) {
```
geigerj: Optional: maybe we should add some comments explaining the intended use case for each of these constructors. For example, this constructor is currently build for sample-gen; the `ProtoFile` constructor is for doc messages; and the `Interface` constructor is for the real imports.

(At first glance I was concerned that this would affect the import handler used to determine the real imports and cause it to add in unnecessary imports, but now I realize it doesn't because it's built through a different constructor.)

Landrito:
@geigerj My thoughts right now are it is caused by a couple of things:
 - The fileImports map in the importHandler gets a bit muddy when proto-files are referenced transitively since two pb_file names will be mapped to the same protofile. I looked at the bimap for the testcase and saw that there were multiple of the same protofile used as keys so the map not be finding a collision correctly. Furthermore, even if it was finding the collision, we are not handling the case since we are assuming there shouldn't be collisions. Bug can also be seen in the source code at this [line](https://github.com/googleapis/toolkit/blob/b92d62397e895df0880fa982071a80f117c066c3/src/test/java/com/google/api/codegen/testdata/python_doc_main_library.baseline#L383). This would lead me to believe that if transitive messages are used then the import handler is broken.
 - The method_sample was using the import_handler for main.snip getting the muddyness from the main import_handler.

Since import handling seems broken when import handler is using transitive messages, we probably need a better way to get element paths. I proposed a solution in my iam python pr but it is a little rough may need a little more iterations to get clean.

src/main/java/com/google/api/codegen/py/PythonImportHandler.java
Landrito: I agree. It should be updated in my next commit.

Landrito:
PTAL. I fixed the situation where the a collision was occurring for the protofiles. Now getting an elements name should be covered for all cases except when a protofile is loaded both directly and transitively (like it will be for the only for the test case when grpc-reroute-to-interface is used)

src/main/java/com/google/api/codegen/py/PythonImportHandler.java
```diff
@@ -238,11 +228,13 @@ public String addImportLocal(String moduleName, String attributeName) {
...
-  public String fileToModule(ProtoFile file) {
-    if (fileImports.containsKey(file)) {
-      return fileImports.get(file);
-    } else {
-      return "";
+  /** Returns the name for a module for a module and attribute stored in the input handler. */
+  public String getShortName(String module, String attribute) {
+    for (PythonImport pyImport : stringImports.inverse().keySet()) {
+      if (pyImport.attributeName().equals(attribute) && pyImport.moduleName().equals(module)) {
+        return pyImport.shortName();
```
Landrito: This can be changed to be a constant time lookup if the local name is removed from PythonImport since it is essentially stored as the shortname in the stringImports map.

Thoughts?

geigerj:
LGTM

PR #482, opened 2016-09-15
Fixes importHandler to not rely on a file mapping since the file mapping will not work for transitively loaded files.

Landrito:
I fixed the situation where the a collision was occurring for the protofiles. Now getting an elements name should be covered for all cases except when a protofile is loaded both directly and transitively (like it will be for the only for the test case when grpc-reroute-to-interface is used)


src/main/java/com/google/api/codegen/py/PythonImportHandler.java
```diff
@@ -238,11 +218,13 @@ public String addImportLocal(String moduleName, String attributeName) {
...
-  public String fileToModule(ProtoFile file) {
-    if (fileImports.containsKey(file)) {
-      return fileImports.get(file);
-    } else {
-      return "";
+  /** Returns the name for a module for a module and attribute stored in the input handler. */
+  public String getShortName(String module, String attribute) {
+    for (PythonImport pyImport : stringImports.inverse().keySet()) {
+      if (pyImport.attributeName().equals(attribute) && pyImport.moduleName().equals(module)) {
+        return pyImport.shortName();
```
Landrito: This can be changed to be a constant time lookup if the local name is removed from PythonImport since it is essentially stored as the shortname in the stringImports map.

Thoughts?

Landrito:
Looks like some codecov is lost on this CL. Hold off on review until I get some test cases.

geigerj:
To be fair, it seems from the codecov delta summary that it is complaining because your change removes 6 lines of previously-tested code, not because it adds any untested code. IMO, just deleting unneeded code should not cause the check to fail, since the "spirit" of the check is to prevent merging untested code in. 

/cc @bjwatson

bjwatson:
@Landrito How is this related to https://github.com/googleapis/toolkit/pull/480, which makes similar test case changes?

@geigerj I agree with the spirit of the check here, since it's forcing us to advance the needle on test coverage. However, I'm okay with ignoring a scenario like this (just removing previously tested code) in the interest of time.

Landrito:
@bjwatson @geigerj I think this issue was fixed by my most recent commit on #470. Closing for now.

PR #483, opened 2016-09-15
Remove gRPC beta types from Python GAPIC

geigerj:
This will allow the codegen artifacts to interact with google-gax
0.13.0, which also drops gRPC beta.


bjwatson:
LGTM

Issue #484, opened 2016-09-15
Allow CommonGapicCodePathMapper to choose path name format for different languages.


swcloud:
getOutputPath should return lower_underscore path name for Ruby.

garrettjonesgoogle:
Instead of creating a subclass, I would prefer to parameterize CommonGapicCodePathMapper with the desired casing treatment of the package and use that for Ruby.

eoogbe:
This issue may no longer be relevant due to its age. Feel free to re-open.

PR #485, opened 2016-09-15
Add a NameFormatter to CommonGapicCodePathMapper.builder

swcloud:
#484 
getOutputPath  should return lower_underscore path name for Ruby. For example, "ErrorReporting" should be "error_reporting".

Adding a nameFormatter allows CommonGapicCodePathMapper to choose different language specific name formatting rule.


garrettjonesgoogle:
LGTM

PR #487, opened 2016-09-19
Merge disco_head into master

tcoffee-google:
Note: Although some commits from disco_head do not appear in the ancestry for this branch, nevertheless ALL changes from disco_head as of this time have been manually merged into this branch.

This commit has NOT been verified to resolve remaining compile-check errors, some of which will require subsequent changes to compile check code outside this repo.


vchudnov-g:
Could you tell us (by separate cover, in a doc) the compilecheck issues that still remain with this?

@saicheems : Could you take a look at this, since I'll be OOO for a few days? Given the size of this, it might be best to pull into a local branch for review (see the "command line instructions" link on Github)

vchudnov-g:
BTW, even after you check this in, wait to delete the disco_head branch. I want to make sure we've really reached parity (which will be harder to verify without the compilecheck succeeding).

saicheems:
Sorry for the delay on this review - I figured there would be some (trivial) conflicts with the PR I was preparing

I plan to spend a few hours on this tomorrow - for all of the conflicts above, the fixes should be as simple as:
- accept all the changes to the java baselines made by my commit
- remove the autoscaler.v1beta2 baseline (it doesn't play well with compilecheck for library download reasons, will fix later)

I see there are light conflicts in `ApiaryConfig.java` and some other base Discovery files, I hope the changes required aren't too intense. I don't think they should be

tcoffee-google:
@saicheems @vchudnov-g I will take a look at these conflicts. For future could you please make sure I'm notified on PR reviews affecting the sample output --- I did not even realize that https://github.com/googleapis/toolkit/pull/476 was in the pipeline, and I don't see a way to make Github do this automatically.

saicheems:
(I admittedly mostly skimmed through this CL, the great majority was baseline changes)

I have general feedback from the last round of comments:
Backquotes were suggested to be replaced with something else. I had chosen to use no quotes at all - I thought the meaning was carried across sufficiently by the context.

The Go reviewers didn't like the long TODO comments. For:
- Change placeholders below to desired parameter values for the `{@methodName}` method:
- Assign values to desired fields of the `{@requestBodyName()}` object.
- Change code below to process each `{@itemName()}` entry:
They preferred:
- Change placeholders below.
- Assign values to the desired fields.
- Process each entry.
I was a bit over-zealous with that feedback, and applied it to the rest of the languages. I guess this comes down to personal opinion, but I thought it was fairly obvious what each comment was alluding to, and that including field/method names in the comments was redundant.

The reviewers also didn't like the way we organized the parameters list - they felt it wasn't obvious enough what had to be changed. The suggestion was to include `TODO: Change placeholder value(s).` after each parameter line, and remove the header comment stating `Change placeholders below for the ...`.

saicheems:
That being said, I'm totally fine with merging this and having a style discussion tomorrow. I think having everything compile is more important than anything else.

tcoffee-google:
> applied it to the rest of the languages

@saicheems Sorry this made me think it was an omission rather than a deliberate change. I'm fine with the placeholder comment changes in Go (will update), since gofmt prohibits the use of spacing that we make in the other languages; but will await empirical evidence that everything else is more obvious to Go developers.

What's the argument against backquotes? I find plenty of instances where particular variable names make undistinguished identifiers within text seem ambiguous to me.

saicheems:
I looked back through the old pull - it seems like Chris was against it in particular... I thought one of the Python folks didn't like it as well, but I didn't see that immediately.
Unfortunately I don't remember the whole conversation, I thought the consensus was either use the quotation specific to the language (single quotes in python, double in java, etc.) or use no quotes as the context is sufficient.

src/test/java/com/google/api/codegen/testdata/discoveries/go/go_adexchangebuyer.v1.4.json.baseline
```diff
@@ -141,10 +140,9 @@ func main() {
     log.Fatal(err)
   }
 
-  // TODO: Change placeholders below to desired parameter values for the `Patch` call:
-
   // The account id
   id := int64(0)
+  // TODO: Change placeholder value above.
```
saicheems: The feedback was:
`id := int64(0) // TODO: Change placeholder value(s).`

tcoffee-google:
> use the quotation specific to the language

@saicheems I think you're referring to string quotation, which doesn't make sense as a way of offsetting identifier names in text. Python Sphinx (reST) uses backticks for identifier markup, as does Markdown. GoDoc does not have an analogous markup standard, but we use backticks for identifiers in comments in our internal code, and a similar convention is suggested to support future tooling on this thread: https://groups.google.com/forum/#!topic/golang-nuts/GRkQMRM-KqE

src/test/java/com/google/api/codegen/testdata/discoveries/go/go_adexchangebuyer.v1.4.json.baseline
tcoffee-google: This does not work in general. Some comments associated with certain placeholder types (e.g., dates/times) cannot be pleasantly combined with that formatting. I think the syntax highlighting on API Explorer will make this at least as visually clear, and it's even easier to edit.

saicheems: It was preferred to alternative approaches, including having the TODO above the line. We didn't consider having it below, but I suspect there's no difference.

saicheems:
That thread is from more than three years ago - I don't think that any such support has been added or suggested in the meanwhile. At the moment, Go standard libraries don't use any quotation standard.

I don't really mind using backquotes anywhere else, I've lost the context of that discussion and am not certain anymore of the decision. Sphinx processing is unrelated to line comments and performs no markdown there. Even in some Google provided [samples](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/appengine/standard/endpoints-frameworks-v2/quickstart/main.py#L51) we don't surround references with backquotes.

saicheems:
(what I mean is we can defer any decision on Python and other languages until consulting with a language expert)
However for Go I think there's already an established convention we should follow

tcoffee-google:
> for Go I think there's already an established convention we should follow

@saicheems Which is different from what we use for our own internal Go source code? And it is ...?

If you mean no markup at all, that may be fine for some code, but this code is autogenerated, and as such is not hand-edited to make wording unambiguous in the face of easily confused identifier names.

saicheems:
Well looks like this bit is moot anyway, I think all markdown is gone from the Go baselines with the shorter comments.

tcoffee-google:
@saicheems There remain `resp` objects and iterated resource variables.

tcoffee-google:
As suggested, submitting so we can more easily move forward with compile check debugging. Any further nits can be resolved in a follow-up.

PR #488, opened 2016-09-19
Implement path template field init rule.

shinfan:
- Implemented a new field spec rule in GAPIC yaml: `field_name%entity_name=value` which allows value assignments of path template variables.
- Refactor FieldStructureParser by creating a new class which stores config data
- Handles the case that the init values may have conflicts.
- A few renames
  - dottedPathStrings -> initFieldConfigStrings
  - hasInitialValue -> hasSimpleInitialValue

Tests:
 - Unit tests
 - Baseline tests

### DIFF
--- a/src/main/java/com/google/api/codegen/SmokeTestConfig.java
+++ b/src/main/java/com/google/api/codegen/SmokeTestConfig.java
@@ -19,15 +19,19 @@ import com.google.api.tools.framework.model.DiagCollector;
 import com.google.api.tools.framework.model.Interface;
 import com.google.api.tools.framework.model.Method;
 import com.google.api.tools.framework.model.SimpleLocation;
+
+import java.util.ArrayList;
 import java.util.List;
 
 /** SmokeTestConfig represents the smoke test configuration for a method. */
 public class SmokeTestConfig {
   private final Method method;
-  private final List<String> initFields;
+  private final List<String> initFieldNames;
+  private final List<String> initFieldSpecs;
 
-  private SmokeTestConfig(Method method, List<String> initFields) {
-    this.initFields = initFields;
+  private SmokeTestConfig(Method method, List<String> initFieldSpecs, List<String> initFieldNames) {
+    this.initFieldSpecs = initFieldSpecs;
+    this.initFieldNames = initFieldNames;
     this.method = method;
   }
 
@@ -42,7 +46,13 @@ public class SmokeTestConfig {
     }
 
     if (testedMethod != null) {
-      return new SmokeTestConfig(testedMethod, smokeTestConfigProto.getInitFieldsList());
+      List<String> initFieldSpecs = smokeTestConfigProto.getInitFieldsList();
+      ArrayList<String> initFieldNames = new ArrayList<>();
+      for (String fieldSpec : initFieldSpecs) {
+        String[] fieldSpecParts = fieldSpec.split("[<=>]");
+        initFieldNames.add(fieldSpecParts[0]);
+      }
+      return new SmokeTestConfig(testedMethod, initFieldSpecs, initFieldNames);
     } else {
       diagCollector.addDiag(
           Diag.error(SimpleLocation.TOPLEVEL, "The configured smoke test method does not exist."));
@@ -50,9 +60,14 @@ public class SmokeTestConfig {
     }
   }
 
+  /** Returns a list of initialized field names. */
+  public List<String> getInitFieldNames() {
+    return initFieldNames;
+  }
+
   /** Returns a list of initialized fields configuration. */
-  public List<String> getInitFields() {
-    return initFields;
+  public List<String> getInitFieldSpecs() {
+    return initFieldSpecs;
   }
 
   /** Returns the method that is used in the smoke test. */
--- a/src/main/java/com/google/api/codegen/SmokeTestConfig.java
+++ b/src/main/java/com/google/api/codegen/SmokeTestConfig.java
@@ -19,19 +19,15 @@ import com.google.api.tools.framework.model.DiagCollector;
 import com.google.api.tools.framework.model.Interface;
 import com.google.api.tools.framework.model.Method;
 import com.google.api.tools.framework.model.SimpleLocation;
-import java.util.ArrayList;
 import java.util.List;
 
 /** SmokeTestConfig represents the smoke test configuration for a method. */
 public class SmokeTestConfig {
   private final Method method;
-  private final List<String> initFieldNames;
   private final List<String> initFieldConfigStrings;
 
-  private SmokeTestConfig(
-      Method method, List<String> initFieldConfigStrings, List<String> initFieldNames) {
+  private SmokeTestConfig(Method method, List<String> initFieldConfigStrings) {
     this.method = method;
-    this.initFieldNames = initFieldNames;
     this.initFieldConfigStrings = initFieldConfigStrings;
   }
 
@@ -44,15 +40,8 @@ public class SmokeTestConfig {
         break;
       }
     }
-
     if (testedMethod != null) {
-      List<String> initFieldSpecs = smokeTestConfigProto.getInitFieldsList();
-      ArrayList<String> initFieldNames = new ArrayList<>();
-      for (String fieldSpec : initFieldSpecs) {
-        String[] fieldSpecParts = fieldSpec.split("[<=>]");
-        initFieldNames.add(fieldSpecParts[0]);
-      }
-      return new SmokeTestConfig(testedMethod, initFieldSpecs, initFieldNames);
+      return new SmokeTestConfig(testedMethod, smokeTestConfigProto.getInitFieldsList());
     } else {
       diagCollector.addDiag(
           Diag.error(SimpleLocation.TOPLEVEL, "The configured smoke test method does not exist."));
--- a/src/main/java/com/google/api/codegen/SmokeTestConfig.java
+++ b/src/main/java/com/google/api/codegen/SmokeTestConfig.java
@@ -19,7 +19,6 @@ import com.google.api.tools.framework.model.DiagCollector;
 import com.google.api.tools.framework.model.Interface;
 import com.google.api.tools.framework.model.Method;
 import com.google.api.tools.framework.model.SimpleLocation;
-
 import java.util.ArrayList;
 import java.util.List;
 
@@ -27,12 +26,13 @@ import java.util.List;
 public class SmokeTestConfig {
   private final Method method;
   private final List<String> initFieldNames;
-  private final List<String> initFieldSpecs;
+  private final List<String> initFieldConfigStrings;
 
-  private SmokeTestConfig(Method method, List<String> initFieldSpecs, List<String> initFieldNames) {
-    this.initFieldSpecs = initFieldSpecs;
-    this.initFieldNames = initFieldNames;
+  private SmokeTestConfig(
+      Method method, List<String> initFieldConfigStrings, List<String> initFieldNames) {
     this.method = method;
+    this.initFieldNames = initFieldNames;
+    this.initFieldConfigStrings = initFieldConfigStrings;
   }
 
   public static SmokeTestConfig createSmokeTestConfig(
@@ -60,14 +60,9 @@ public class SmokeTestConfig {
     }
   }
 
-  /** Returns a list of initialized field names. */
-  public List<String> getInitFieldNames() {
-    return initFieldNames;
-  }
-
   /** Returns a list of initialized fields configuration. */
-  public List<String> getInitFieldSpecs() {
-    return initFieldSpecs;
+  public List<String> getInitFieldConfigStrings() {
+    return initFieldConfigStrings;
   }
 
   /** Returns the method that is used in the smoke test. */
--- a/src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
+++ b/src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
@@ -63,11 +63,14 @@ public class FieldStructureParser {
 
   private static InitValueConfig createInitValueConfig(
       InitFieldConfig fieldConfig, Map<String, InitValueConfig> initValueConfigMap) {
+    if (fieldConfig.isFormattedConfig()
+        && !initValueConfigMap.containsKey(fieldConfig.fieldPath())) {
+      // If the field is a formatted field, it must exist in the value config map.
+      throw new IllegalArgumentException("The field name is not found in the collection map.");
+    }
+
     InitValueConfig valueConfig = null;
     if (fieldConfig.hasFormattedInitValue()) {
-      if (!initValueConfigMap.containsKey(fieldConfig.fieldPath())) {
-        throw new IllegalArgumentException("The field name is not found in the collection map.");
-      }
       valueConfig =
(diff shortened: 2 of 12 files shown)


src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -129,8 +128,8 @@ public InitCodeView generateSmokeTestInitCode(
             InitTreeParserContext.newBuilder()
                 .table(table)
                 .rootObjectType(testConfig.getMethod().getInputType())
-                .initValueConfigMap(createInitValueMap(context))
-                .dottedPathStrings(testConfig.getInitFields())
+                .initValueConfigMap(createCollectionMap(context))
+                .initFieldConfigStrings(testConfig.getInitFieldSpecs())
```
michaelbausor: Why rename with Specs? It seems like we have removed the Specs naming convention from everywhere else

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
```diff
@@ -72,6 +73,13 @@ public InitValueConfig getInitValueConfig() {
   }
 
   /*
+   * Updates the InitValueConfig.
+   */
+  public void updateInitValueConfig(InitValueConfig initValueConfig) {
```
michaelbausor: Is this used anywhere outside the mergeChild() method? If not, it can be private, or can just be replaced by directly setting the field in mergeChild()

src/main/java/com/google/api/codegen/SmokeTestConfig.java
```diff
@@ -42,17 +46,28 @@ public static SmokeTestConfig createSmokeTestConfig(
...
+      return new SmokeTestConfig(testedMethod, initFieldSpecs, initFieldNames);
     } else {
       diagCollector.addDiag(
           Diag.error(SimpleLocation.TOPLEVEL, "The configured smoke test method does not exist."));
       return null;
     }
   }
 
+  /** Returns a list of initialized field names. */
+  public List<String> getInitFieldNames() {
```
michaelbausor: What is this used for?

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
bjwatson: Specs actually indicate tests in some frameworks.

src/test/java/com/google/api/codegen/testdata/library_gapic.yaml
```diff
@@ -43,9 +43,10 @@ interfaces:
   - name_pattern: archives/{archive_path=**}/books/{book}
     entity_name: archived_book
   smoke_test:
-    method: CreateShelf
+    method: GetBook
     init_fields:
-      - shelf
+      - name%shelf="testShelf"
+      - name%book="testBook"
```
bjwatson: This looks nice!

bjwatson:
Sorry, @garrettjonesgoogle. Me [hyper-corrected](https://en.wikipedia.org/wiki/Hypercorrection) again. :(

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
```diff
@@ -50,26 +74,61 @@ static Pattern getFieldMapPattern() {
...
+      valueConfig = createInitValueConfig(fieldConfig, initValueConfigMap);
+    } else if (initValueConfigMap.containsKey(fieldConfig.fieldPath())) {
+      valueConfig = initValueConfigMap.get(fieldConfig.fieldPath());
+    }
+
+    return parsePartialDottedPathToInitCodeNode(
+        fieldConfig.fieldPath(), InitCodeLineType.Unknown, valueConfig, null);
+  }
+
+  private static InitFieldConfig createInitFieldConfig(String initFieldConfigString) {
```
shinfan: I think the code would be more readable if we parse the string data into InitFieldConfig class first and reuse that object . In this way it avoids using variables with poor readability such as equalParts[0] and equalParts[1].

src/main/java/com/google/api/codegen/SmokeTestConfig.java
```diff
@@ -19,15 +19,19 @@
...
 public class SmokeTestConfig {
   private final Method method;
-  private final List<String> initFields;
+  private final List<String> initFieldNames;
+  private final List<String> initFieldSpecs;
 
-  private SmokeTestConfig(Method method, List<String> initFields) {
-    this.initFields = initFields;
+  private SmokeTestConfig(Method method, List<String> initFieldSpecs, List<String> initFieldNames) {
+    this.initFieldSpecs = initFieldSpecs;
```
shinfan: Done

src/main/java/com/google/api/codegen/SmokeTestConfig.java
shinfan: Removed

src/main/java/com/google/api/codegen/metacode/InitValueConfig.java
```diff
@@ -46,11 +54,24 @@ public static InitValueConfig create(String apiWrapperName, CollectionConfig col
...
-  /**
-   * Creates an updated InitValueConfig with the provided value.
-   */
+  @Nullable
+  public abstract Map<String, String> getCollectionValues();
+
+  /** Creates an updated InitValueConfig with the provided value. */
   public InitValueConfig withInitialValue(String string) {
-    return new AutoValue_InitValueConfig(getApiWrapperName(), getCollectionConfig(), string);
+    return new AutoValue_InitValueConfig(getApiWrapperName(), getCollectionConfig(), string, null);
```
shinfan: Removed.

src/main/java/com/google/api/codegen/metacode/InitValueConfig.java
```diff
@@ -61,7 +82,11 @@ public boolean hasFormattingConfig() {
     return getCollectionConfig() != null;
   }
 
-  public boolean hasInitialValue() {
+  public boolean hasFormattedInitialValue() {
```
shinfan: Done

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
shinfan: Done

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
shinfan: Changed to private

src/test/java/com/google/api/codegen/testdata/library_gapic.yaml
shinfan: Thanks!

shinfan:
PTAL

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
```diff
@@ -173,16 +184,34 @@ private void mergeChild(InitCodeNode newChild) {
     }
   }
 
+  private static InitValueConfig mergeInitValueConfig(
+      InitValueConfig oldConfig, InitValueConfig newConfig) {
+    HashMap<String, String> collectionValues = new HashMap<>();
+    if (oldConfig.hasSimpleInitialValue()
+        && newConfig.hasSimpleInitialValue()
+        && oldConfig.getInitialValue().equals(newConfig.getInitialValue())) {
```
michaelbausor: Should this be checking for not equals?

src/test/java/com/google/api/codegen/metacode/SampleInitCodeTest.java
```diff
@@ -309,7 +328,24 @@ public void testMultipleMapEntries() throws Exception {
         Arrays.asList("key1", "key2", "stringmap", "123", "456", "intmap", "root");
 
     InitCodeNode rootNode =
-        InitCodeNode.createTree(getContextBuilder().dottedPathStrings(fieldSpecs).build());
+        InitCodeNode.createTree(getContextBuilder().initFieldConfigStrings(fieldSpecs).build());
+    List<String> actualKeyList = new ArrayList<>();
+    for (InitCodeNode node : rootNode.listInInitializationOrder()) {
+      actualKeyList.add(node.getKey());
+    }
+    Truth.assertThat(actualKeyList.equals(expectedKeyList));
```
michaelbausor: Append .isTrue

src/test/java/com/google/api/codegen/metacode/SampleInitCodeTest.java
```diff
@@ -309,7 +328,24 @@ public void testMultipleMapEntries() throws Exception {
...
+  }
+
+  @Test
+  public void testMultipleFormattedEntries() throws Exception {
+    List<String> fieldSpecs =
+        Arrays.asList(
+            "formatted_field%entity1", "formatted_field%entity2", "formatted_field%entity3");
```
michaelbausor: You could also add an =value for each entry, and then check that the initValueConfig created has those values.

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
michaelbausor: I definitely agree about equalParts[], it is good to use the fieldName, entityName and value variables. But could they, and the splitting code, be part of the parse method, and be parsed as separate args to createInitValueConfig?

src/main/java/com/google/api/codegen/SmokeTestConfig.java
```diff
@@ -19,16 +19,20 @@
...
 import com.google.api.tools.framework.model.Method;
 import com.google.api.tools.framework.model.SimpleLocation;
+import java.util.ArrayList;
 import java.util.List;
 
 /** SmokeTestConfig represents the smoke test configuration for a method. */
 public class SmokeTestConfig {
   private final Method method;
-  private final List<String> initFields;
+  private final List<String> initFieldNames;
```
michaelbausor: What is this field used for? It is inaccessible now

src/test/java/com/google/api/codegen/metacode/SampleInitCodeTest.java
shinfan: Done. Added a separate test.

src/test/java/com/google/api/codegen/metacode/SampleInitCodeTest.java
shinfan: Done

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
shinfan: Done

src/main/java/com/google/api/codegen/SmokeTestConfig.java
shinfan: Removed

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
```diff
@@ -78,52 +55,32 @@ public static InitCodeNode parse(String initFieldConfigString) {
...
-    InitValueConfig valueConfig = null;
-    if (fieldConfig.hasInitValue()) {
-      valueConfig = createInitValueConfig(fieldConfig, initValueConfigMap);
-    } else if (initValueConfigMap.containsKey(fieldConfig.fieldPath())) {
-      valueConfig = initValueConfigMap.get(fieldConfig.fieldPath());
-    }
-
+    InitFieldConfig fieldConfig = InitFieldConfig.of(initFieldConfigString);
+    InitValueConfig valueConfig =
+        createInitValueConfig(InitFieldConfig.of(initFieldConfigString), initValueConfigMap);
```
michaelbausor: Pass fieldConfig here instead of re-parsing

src/main/java/com/google/api/codegen/metacode/InitFieldConfig.java
```diff
@@ -0,0 +1,66 @@
...
+  @Nullable
+  public abstract String entityName();
+
+  @Nullable
+  public abstract String value();
+
+  /*
+   * Parses the given config string and returns the corresponding object.
+   */
+  public static InitFieldConfig of(String initFieldConfigString) {
```
michaelbausor: I think "from" or "parse" instead of "of"

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
```diff
@@ -78,52 +55,32 @@ public static InitCodeNode parse(String initFieldConfigString) {
...
 
   private static InitValueConfig createInitValueConfig(
       InitFieldConfig fieldConfig, Map<String, InitValueConfig> initValueConfigMap) {
-    if (fieldConfig.isFormattedField()) {
-      if (initValueConfigMap.containsKey(fieldConfig.fieldPath())) {
-        InitValueConfig valueConfig = initValueConfigMap.get(fieldConfig.fieldPath());
-        return valueConfig.withInitialCollectionValue(
-            fieldConfig.entityName(), fieldConfig.value());
+    InitValueConfig valueConfig = null;
+    if (fieldConfig.hasInitValue()) {
```
michaelbausor: Do we want to explicitly handle the case where hasInitValue() is false, but isFormattedField() is true? In that case, if fieldConfig.fieldPath() is missing from initValueConfigMap it should be an error, but at the moment it will return null.

Possibly it might be good to make the checking functions on InitFieldConfig mirror the checking functions on InitValueConfig. So for example, replace fieldConfig.hasInitValue() with fieldConfig.hasSimpleInitValue(), which would check (value()!=null && entityName()==null). I think this would simplify/clarify the login in createInitValueConfig, but I am not 100% sure, let me know what you think.

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
shinfan: Done

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
shinfan: Yes, I think that makes sense. Switched to use hasSimpleInitValue() and hasFormattedInitValue()

src/main/java/com/google/api/codegen/metacode/InitFieldConfig.java
shinfan: Picked "from"

shinfan:
PTAL

src/test/java/com/google/api/codegen/metacode/SampleInitCodeTest.java
```diff
@@ -215,57 +229,63 @@ public void testEmbeddedFieldList() throws Exception {
   }
 
   @Test(expected = IllegalArgumentException.class)
+  public void testFormattedFieldBadField() throws Exception {
+    String fieldSpec = "name%entity=test";
```
michaelbausor: The changes to the conditionals look good, but I think there is still an unhandled error case. This test should still throw an exception if you remove the "=test" component, but at the moment I don't think it will.

shinfan: Fixed and added another test case. Good catch.

PR #492, opened 2016-09-20
Mangle reserved identifiers in method samples

### LINKED ISSUE #298: Sample-gen should respect Python parameter mangling
After https://github.com/googleapis/toolkit/pull/297, sample-gen still uses the blacklisted word `options` in the generated code, and presumably does the same for Python built-ins and keywords. We should mangle the names using the blacklist.

cc: @bjwatson

geigerj:
Fixes #298


bjwatson:
LGTM

PR #500, opened 2016-09-21
C# MVVM migration

### LINKED ISSUE #462: Remove ToClient extension method
Mixins will make ToClient very odd, as we'd need to accept the "other" API client (which we'd expect to have the same endpoint) as a parameter.

If we change to make the FooClientImpl constructor accept as many clients as it needs, that means callers *can* still use it, but it's slightly less jarring when there are two clients.

(We'd still expect most callers to use the `FooClient.Create` extension method, which will create both gRPC clients with the same channel.)

### LINKED ISSUE #463: C# s_channelPool incorrect
Fix s_channelPool instantiation to pass scopes.

chrisdunelm:
Fixes #462.
Fixes #463.


src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
```diff
@@ -202,7 +207,7 @@ private void setFlattenedMethodFields(
     for (Field field : fields) {
       params.add(generateRequestObjectParam(context, field));
     }
-    methodViewBuilder.methodParams(params);
+    methodViewBuilder.methodParams(params).methodParamsWithExtras(params);
```
garrettjonesgoogle: This doesn't follow the pattern of one setting per line - I was actually a bit confused about this for a little while.

src/main/java/com/google/api/codegen/viewmodel/RetryCodesDefinitionView.java
```diff
@@ -18,13 +18,21 @@
...
 public abstract class RetryCodesDefinitionView {
   public abstract String key();
 
+  @Nullable // Used in C#
+  public abstract String name();
+
   public abstract ImmutableSet<Code> codes();
 
+  @Nullable // Used in C#
+  public abstract List<String> codeNames();
```
garrettjonesgoogle: Looking at this case, I might just prefer to populate them (name/codeNames) for all languages and avoid the special logic for C# to open up the view class, shove these in, and rebuild it.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSurfaceTransformer.java
```diff
@@ -0,0 +1,329 @@
...
+            .defaultValue(defaultValue)
+            .build()).toList())
+        .doc(method.doc().toBuilder().paramDocs(FluentIterable.from(method.doc().paramDocs()).append(SimpleParamDocView.newBuilder()
+            .paramName(name)
+            .typeName(typeName)
+            .lines(docList)
+            .firstLine(doc[0])
+            .remainingLines(docList.subList(1, docList.size()))
+            .build()).toList()).build())
+        .build();
```
garrettjonesgoogle: The embedding is a little too deep here.

src/main/resources/com/google/api/codegen/csharp/gapic_client.snip
```diff
@@ -0,0 +1,293 @@
...
+    @end
+    }
+@end
+
+@private flattenedMethodAsyncCancellationToken(method, isImpl)
+    @if isImpl
+    @else
+        public virtual {@method.responseTypeName} {@method.name}(
+            {@objectParams(method.methodParamsWithExtras)}) => {@method.name}(
+                {@objectArgs(method.methodParams)},
```
garrettjonesgoogle: I might prefer to use methodParams for the params showing up in the signature of our surface, and use a new attribute 'forwardingMethodParams' (or the like) for the params of the method we are forwarding the call to.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSurfaceTransformer.java
```diff
@@ -0,0 +1,329 @@
...
+        break;
+      case FlattenedMethodAsyncCancellationToken:
+        method = AddParam(method, "CancellationToken", "cancellationToken", null, "A <see cref=\"CancellationToken\"/> to use for this RPC.");
+        method = MakeAsync(method);
+        method = AddReturnDoc(method, "A Task containing the RPC response.");
+        break;
+      default:
+        throw new RuntimeException("Unexpected method type: '" + method.type() + "'");
+    }
+    return method;
```
garrettjonesgoogle: I am uneasy with this pattern of unsealing/rebuilding view model objects. It makes it more difficult to trace the inputs and outputs of the transformation, because the set of values seen on a final view model object would be provided at multiple places in the code. This is almost equivalent to just having mutable view model classes. I would much prefer to add parameters into generateFlattenedMethod so that it could do all of the necessary work.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpSurfaceNamer.java
```diff
@@ -34,4 +37,25 @@ public String getStaticLangReturnTypeName(Method method, MethodConfig methodConf
...
   }
 
+  @Override
+  public String getApiWrapperClassName(Interface interfaze) {
+    return className(Name.upperCamel(interfaze.getSimpleName(), "Client"));
+  }
+
+  @Override
+  public String getCallableName(Method method) {
+    return "_" + varName(Name.upperCamel("Call", method.getSimpleName()));
```
garrettjonesgoogle: What does this underscore mean?

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
```diff
@@ -408,6 +420,7 @@ private RequestObjectParamView generateRequestObjectParam(
     SurfaceNamer namer = context.getNamer();
     RequestObjectParamView.Builder param = RequestObjectParamView.newBuilder();
     param.name(namer.getVariableName(field));
+    param.nameAsMethodName(namer.methodName(Name.lowerCamel(namer.getVariableName(field))));
```
garrettjonesgoogle: Regarding the error:

`ruby/main.snip:375: exception when invoking method 'getMethodView': java.lang.IllegalArgumentException: Name: identifier not in lower camel: 'string_builder'.`

This is where it is happening. The problem is basically that Name.lowerCamel enforces that its input is *actually* lower-camel, which is not true for variable names in Ruby. This logic should all be done in SurfaceNamer; as I mentioned elsewhere (I know you haven't been able to look over the feedback yet, so this will be repetitive when you see it), all case-awareness should be confined to SurfaceNamer. I would recommend adding a method in SurfaceNamer named something like "getFieldAsMethodName". 

In the future, to narrow down the line that is throwing an exception when you're using the snippet language, I recommend temporarily adding a try/catch in the context method that is throwing it, and print the stack trace there to std err. Note - this becomes unnecessary for any language using MVVM, since calls into Java are pure get calls and can't fail (except for null pointer exceptions due to data not being set).

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -68,11 +68,12 @@ public TypeNameConverter getTypeNameConverter() {
 
   public String getNotImplementedString(String feature) {
     return "$ NOT IMPLEMENTED: " + feature + " $";
+    //throw new RuntimeException(feature);
   }
 
   /** The full path to the source file  */
   public String getSourceFilePath(String path, String className) {
-    return getNotImplementedString("SurfaceNamer.getSourceFilePath");
+    return getNotImplementedString("SurfaceNamer:getSourceFilePath");
```
chrisdunelm: Changed back to original periods.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSurfaceTransformer.java
```diff
@@ -0,0 +1,329 @@
...
+  }
+
+  private StaticLangApiMethodView AddParam(StaticLangApiMethodView method, String typeName, String name, String defaultValue, String... doc) {
```
chrisdunelm: Done

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSurfaceTransformer.java
chrisdunelm: Done

src/main/java/com/google/api/codegen/transformer/csharp/CSharpApiCallableTransformer.java
```diff
@@ -0,0 +1,50 @@
...
+import com.google.common.base.Function;
+import com.google.common.collect.FluentIterable;
+import com.google.common.collect.Maps;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public class CSharpApiCallableTransformer extends ApiCallableTransformer {
+
+  public List<ApiCallSettingsView> generateCallSettings(SurfaceTransformerContext context,
```
chrisdunelm: Done

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
```diff
@@ -118,14 +118,19 @@ public StaticLangApiMethodView generateUnpagedListCallableMethod(
 
   public StaticLangApiMethodView generateFlattenedMethod(
       MethodTransformerContext context, ImmutableList<Field> fields) {
+    return generateFlattenedMethod(context, fields, ApiMethodType.FlattenedMethod);
+  }
+  
+  public StaticLangApiMethodView generateFlattenedMethod(
+      MethodTransformerContext context, ImmutableList<Field> fields, ApiMethodType type) {
```
chrisdunelm: Done

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
chrisdunelm: Fixed

src/main/java/com/google/api/codegen/transformer/csharp/CSharpSurfaceNamer.java
chrisdunelm: It's a convention in C# to prefix fields with an underscore.
But placing the underscore within the argument to Name.upperCamel causes it to complain that the argument isn't in upperCamel format.

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
chrisdunelm: Thanks for the explanation. This is fixed by moving the casing logic to SurfaceNamer, as you suggest.

src/main/java/com/google/api/codegen/viewmodel/ApiMethodType.java
```diff
@@ -25,5 +25,8 @@
   CallableMethod,
   // PHP
   OptionalArrayMethod,
-  PagedOptionalArrayMethod
+  PagedOptionalArrayMethod,
+  // C#
+  FlattenedMethodAsyncCallSettings,
+  FlattenedMethodAsyncCancellationToken,
```
chrisdunelm: Done

src/main/java/com/google/api/codegen/util/csharp/CSharpTypeTable.java
```diff
@@ -0,0 +1,92 @@
...
+import com.google.common.base.Functions;
+import com.google.common.base.Predicates;
+import com.google.common.collect.FluentIterable;
+import java.util.HashMap;
+import java.util.Map;
+
+public class CSharpTypeTable implements TypeTable {
+
+  private final String implicitPackageName;
+  // Fullname to shortname map
```
chrisdunelm: Done

src/main/java/com/google/api/codegen/util/csharp/CSharpTypeTable.java
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  @Override
+  public NamePath getNamePath(String fullName) {
+    return NamePath.dotted(fullName);
+  }
+
+  @Override
+  public TypeName getContainerTypeName(String containerFullName, String... elementFullNames) {
+    throw new RuntimeException();
```
chrisdunelm: This is now implemented properly :)

src/main/java/com/google/api/codegen/util/csharp/CSharpTypeTable.java
```diff
@@ -0,0 +1,92 @@
...
+          int index = fullName.lastIndexOf('.');
+          if (index < 0) {
+            return null;
+          } else {
+            return fullName.substring(0, index);
+          }
+        }
+      })
+      .filter(Predicates.notNull())
+      .toMap(Functions.<String>identity());
```
chrisdunelm: Changed, and outputs a SortedMap for correct import ordering

src/main/java/com/google/api/codegen/util/Name.java
```diff
@@ -69,6 +69,15 @@ public static Name upperCamel(String... pieces) {
...
   }
 
+  public static Name upperUnderscore(String... pieces) {
+    List<NamePiece> namePieces = new ArrayList<>();
+    for (String piece : pieces) {
+      validateUpperUnderscore(piece);
+      namePieces.add(new NamePiece(piece, CaseFormat.UPPER_UNDERSCORE));
+    }
+    return new Name(namePieces);
+  }
```
chrisdunelm: I've not looked at this part of @pongad 's Go MVVM PR ;)

src/main/java/com/google/api/codegen/transformer/csharp/CSharpModelTypeNameConverter.java
```diff
@@ -0,0 +1,95 @@
...
+          .put(Type.TYPE_FIXED32, "int")
+          .put(Type.TYPE_SFIXED32, "int")
+          .put(Type.TYPE_STRING, "string")
+          .put(Type.TYPE_BYTES, "ByteString") // TODO: Full path
+          .build();
+
+  private TypeNameConverter typeNameConverter;
+
+  public CSharpModelTypeNameConverter(String implicitPackageName) {
+    this.typeNameConverter = new CSharpTypeTable(implicitPackageName); // TODO: Is it really OK to create a new instance here?
```
chrisdunelm: OK, thanks

src/main/java/com/google/api/codegen/viewmodel/RetryCodesDefinitionView.java
chrisdunelm: This is already generated specifically for Java in JavaGapicSurfaceTransformer, so I'll leave it here for now.

src/main/resources/com/google/api/codegen/csharp/gapic_client.snip
```diff
@@ -0,0 +1,293 @@
...
+        @join method : api.apiMethods
+            {@xmlDoc(method.doc)}
+            {@method(method, FALSE)}
+
+        @end
+    }
+@end
+
+@private implClient(api)
+    public sealed partial class {@api.name}Impl : {@api.name}
```
chrisdunelm: Done

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSurfaceTransformer.java
chrisdunelm: OK, done, but I'm not at all convinced this is better. If there's a nicer way of doing it in this way, please let me know.

In my (admitedly functional!) opinion the pattern I had used of transforming immutable objects, although somewhat ugly in Java, is nicer as it allows the common parts to be generated by common code, but then the language-specific transforms are applied afterwards. Yes, values seen in the final view model object come from two places, but it does mean the common generation code isn't confused (as it is a bit now) with language-specific concerns.
(This is a well-used pattern in more functional languages. OK, OK, I know Java is functional however much I try to convince it to be ;)

src/main/java/com/google/api/codegen/viewmodel/RetryCodesDefinitionView.java
garrettjonesgoogle: Could you move the retry param code from JavaGapicSurfaceTransformer into a separate transformer class that Java and C# could share?

chrisdunelm:
PTAL - I'll try to sort out the conflict in a few hours.

src/main/resources/com/google/api/codegen/csharp/gapic_client.snip
```diff
@@ -0,0 +1,465 @@
...
+               /// <list type="bullet">
+               @join code : retry.codeNames
+                   /// <item><description><see cref="StatusCode.{@code}"/></description></item>
+               @end
+               /// </list>
+            @else
+                /// There are no RPC <see cref="StatusCode"/>s eligible for retry for "{@retry.name}" RPC methods.
+            @end
+            /// </remarks>
+            public static Predicate<RpcException> {@retry.name}RetryFilter { get; } =
```
garrettjonesgoogle: Don't compose new identifiers inside the templates - compose them in the SurfaceNamer and pass them through the view. (Same for `{@call.methodName}Settings`, `{@var.name}Id`, and possibly other examples)

src/main/java/com/google/api/codegen/transformer/csharp/CSharpSurfaceNamer.java
```diff
@@ -0,0 +1,117 @@
...
+  @Override
+  public String transformTypeToAsync(String name) {
+    if (name.equals("void")) {
+      return "Task";
+    } else {
+      return "Task<" + name + ">";
+    }
+  }
+
+  public String getPageStreamingDescriptorConstName(Method method) {
```
garrettjonesgoogle: add `@Override`

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSurfaceTransformer.java
```diff
@@ -0,0 +1,496 @@
...
+    String name = namer.getApiWrapperClassName(context.getInterface());
+    //String grpcName = namer.className(Name.upperCamel(context.getInterface().getSimpleName()));
+    xapiClass.name(name);
+    xapiClass.implName(namer.getApiWrapperClassImplName(context.getInterface()));
+    xapiClass.grpcServiceName(namer.getGrpcContainerTypeName(context.getInterface())); //(grpcName);
+    xapiClass.grpcTypeName(
+        namer.getGrpcServiceClassName(context.getInterface())); //(grpcName + "." + name);
+    xapiClass.settingsClassName(context.getNamer().getApiSettingsClassName(context.getInterface()));
+    List<ApiCallableView> callables =
+        FluentIterable.from(apiCallableTransformer.generateStaticLangApiCallables(context, true))
```
garrettjonesgoogle: I can't tell what the "true" means from this context.

In a separate PR, I'm having @michaelbausor implement a FeatureConfig class which could control things like this. (In his case, he is turning on/off the resource name usage feature.) Could you co-implement FeatureConfig? (in the sense that you create it at the same location, and then we get the two implementations spliced together later). Put it under codegen/transformer/FeatureConfig.java. It should work like SurfaceNamer, where there is a default value for each feature (set to "false"), and then each language turns the features on as they get implemented. For your case:

public boolean mixinsAreOn() {
  return false;
}

You would have a CSharpFeatureConfig; the FeatureConfig instance would be passed to SurfaceTransformerContext when it is created, and then accessed through the context in the code that needs to know.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSurfaceTransformer.java
```diff
@@ -0,0 +1,496 @@
...
+    SurfaceNamer namer = context.getNamer();
+    StaticLangXApiView.Builder xapiClass = StaticLangXApiView.newBuilder();
+    List<StaticLangApiMethodView> methods = generateApiMethods(context);
+
+    xapiClass.doc(serviceTransformer.generateServiceDoc(context, null));
+
+    xapiClass.templateFileName(XAPI_TEMPLATE_FILENAME);
+    xapiClass.packageName(context.getApiConfig().getPackageName());
+    String name = namer.getApiWrapperClassName(context.getInterface());
+    //String grpcName = namer.className(Name.upperCamel(context.getInterface().getSimpleName()));
```
garrettjonesgoogle: delete?

src/main/resources/com/google/api/codegen/csharp/gapic_client.snip
```diff
@@ -0,0 +1,465 @@
...
+            {@objectParams(method.methodParams)}) => {@method.name}(
+                {@objectArgs(method.forwardingMethodParams)},
+                new CallSettings { CancellationToken = cancellationToken });
+    @end
+@end
+
+@private pagedFlattenedMethodSync(method, isImpl)
+    @if isImpl
+        public override IPagedEnumerable<{@method.pagedView.responseTypeName}, {@method.pagedView.resourceTypeName}> {@method.name}(
+        {@""}    {@objectParams(method.methodParams)}) => new PagedEnumerable<{@method.pagedView.requestTypeName}, {@method.pagedView.responseTypeName}, {@method.pagedView.resourceTypeName}>(
```
garrettjonesgoogle: You shouldn't need the `{@""}` here. The only time you need it in a place like this is when you want to force an indentation on the first line of a snippet that has no raw text lines before it.

src/main/java/com/google/api/codegen/viewmodel/RequestObjectParamView.java
```diff
@@ -30,6 +34,14 @@
 
   public abstract boolean isArray();
 
+  @Nullable // Used in C#
+  public abstract String defaultValue();
+
+  // Work around a limitation in the snippet language
+  public String defaultValue(String template) {
+    return defaultValue() != null ? String.format(template, defaultValue()) : "";
+  }
```
garrettjonesgoogle: I'd prefer a more direct approach. Add a method "hasDefaultValue" and use an `@if` in the snippet to switch the output based on the value.

src/main/java/com/google/api/codegen/util/csharp/CSharpNameFormatter.java
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  @Override
+  public String inittedConstantName(Name name) {
+    return name.toUpperCamel();
+  }
+
+  @Override
+  public String keyName(Name name) {
+    throw new RuntimeException();
```
garrettjonesgoogle: In C#, is it more common for keys to be lower camel, lower underscore, or something else?

src/main/java/com/google/api/codegen/util/csharp/CSharpNameFormatter.java
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  @Override
+  public String packageFilePathPiece(Name name) {
+    return name.toUpperCamel();
+  }
+
+  @Override
+  public String classFileNameBase(Name name) {
+    throw new RuntimeException();
```
garrettjonesgoogle: use:

`return className(name);`

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSurfaceTransformer.java
```diff
@@ -0,0 +1,496 @@
...
+    settingsDoc.serviceAddress(serviceConfig.getServiceAddress(context.getInterface()));
+    settingsDoc.servicePort(serviceConfig.getServicePort());
+    settingsDoc.exampleApiMethodName(""); // Unused in C#
+    settingsDoc.exampleApiMethodSettingsGetter(""); // Unused in C#
+    settingsDoc.apiClassName(namer.getApiWrapperClassName(context.getInterface()));
+    settingsDoc.settingsVarName(namer.getApiSettingsVariableName(context.getInterface()));
+    settingsDoc.settingsClassName(namer.getApiSettingsClassName(context.getInterface()));
+    settingsDoc.settingsBuilderVarName(namer.getApiSettingsBuilderVarName(context.getInterface()));
+    return settingsDoc.build();
+  }
```
garrettjonesgoogle: Java and C# could share a single implementation of this

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
```diff
@@ -53,15 +56,28 @@ public ApiMethodTransformer() {
...
   public StaticLangApiMethodView generatePagedFlattenedMethod(
-      MethodTransformerContext context, ImmutableList<Field> fields) {
+      MethodTransformerContext context,
+      ImmutableList<Field> fields,
+      ApiMethodType type,
+      List<RequestObjectParamView> additionalParams,
+      List<ParamDocView> additionalParamDocs,
+      List<String> returnsDocLines,
+      boolean isAsync,
+      Map<String, PageStreamingDescriptorClassView> pagedByName) {
```
garrettjonesgoogle: I see what you're saying about how this method turned up. I will get back to you on a suggestion with a better approach.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpSurfaceNamer.java
chrisdunelm: Done

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSurfaceTransformer.java
chrisdunelm: Done

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
```diff
@@ -53,15 +56,28 @@ public ApiMethodTransformer() {
...
+    if (pagedByName != null) {
+      methodViewBuilder.pagedView(pagedByName.get(name));
+    }
     setListMethodFields(context, methodViewBuilder);
-    setFlattenedMethodFields(context, fields, methodViewBuilder);
+    setFlattenedMethodFields(
+        context, fields, additionalParams, additionalParamDocs, returnsDocLines, methodViewBuilder);
 
-    return methodViewBuilder.type(ApiMethodType.PagedFlattenedMethod).build();
+    return methodViewBuilder.type(type).build();
```
garrettjonesgoogle: So the general approach I followed for Java was to have a distinct generate method in this class for each api method variant being generated. Then, I use private helper methods for any duplicate work between them. Your change here doesn't match that; it re-uses generate methods for multiple api method variants. This method in particular (generatePagedFlattenedMethod) should just be used to generate ApiMethodType.PagedFlattenedMethod. 

Specific notes:

* PageStreamingDescriptorClassView isn't the appropriate view to use for request/response/resource type information; that view is supposed to be used for the implementation of a page streaming descriptor. The setListMethodFields function should be able to provide you the data you need to generate.
* returnsDocLines: You could acquire the strings for this from SurfaceNamer similar to the throwsDocLines case, except you will need multiple functions and a couple of them will need to know the resource type name. 
* Instead of two lists containing information on the extra parameters (one for the parameter, one for the doc), create a new class (just used in transformer-land) that mimics a Field and that can provide the information in a non-redundant way. Then, within the individual generateXMethod functions, you can create the list of those params and pass them off to the other functions in this file. There are a couple other places in this file that I can refactor later to use the new class you create. 

Doing all of these things would remove the need for any of the new parameters you added, and also would make it follow the same pattern as the existing code.

src/main/java/com/google/api/codegen/viewmodel/RequestObjectParamView.java
chrisdunelm: Done

src/main/java/com/google/api/codegen/util/csharp/CSharpNameFormatter.java
chrisdunelm: I'm not quite sure what a 'key' means here?

src/main/java/com/google/api/codegen/util/csharp/CSharpNameFormatter.java
chrisdunelm: Done

src/main/resources/com/google/api/codegen/csharp/gapic_client.snip
chrisdunelm: Done

src/main/resources/com/google/api/codegen/csharp/gapic_client.snip
chrisdunelm: Done

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -547,4 +560,24 @@ public String getNamespace(Interface service) {
...
+    return methodName(Name.upperUnderscore(code.toString()));
+  }
+
+  /* The name of a retry definition */
+  public String getRetryDefinitionName(String retryDefinitionKey) {
+    return methodName(Name.from(retryDefinitionKey));
+  }
+
+  /** Transform a sync method name to async (this is a bit hacky) */
+  public String transformMethodNameToAsync(String name) {
```
chrisdunelm: Done

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -547,4 +560,24 @@ public String getNamespace(Interface service) {
...
+    return getNotImplementedString("SurfaceNamer.transformNameToAsync");
+  }
+
+  /** Transform a sync type to an async type (this is a bit hacky) */
+  public String transformTypeToAsync(String name) {
```
chrisdunelm: Done

src/main/java/com/google/api/codegen/util/csharp/CSharpNameFormatter.java
chrisdunelm: Well, there isn't really any standard that I'm aware of.
I'll just say it's UpperCamel I think ;)
Done.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSurfaceTransformer.java
```diff
@@ -0,0 +1,329 @@
...
+            .lines(docList)
+            .firstLine(doc[0])
+            .remainingLines(docList.subList(1, docList.size()))
+            .build()).toList()).build())
+        .build();
+  }
+
+  private StaticLangApiMethodView MakeAsync(StaticLangApiMethodView method) {
+    return method.toBuilder()
+        .name(method.name() + "Async")
```
chrisdunelm: Done (well, method removed)

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSurfaceTransformer.java
chrisdunelm: Done

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
chrisdunelm: Done

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
chrisdunelm: Done

src/main/resources/com/google/api/codegen/csharp/gapic_client.snip
chrisdunelm: I don't really agree with this desire that *nothing whatsoever* is composed in the templates. These examples are unlikely to be used in any other language from C#, what advantage is there in adding these 4 or 5 identifiers, and making the template more opaque?
And if there is a problem later, it's trivial to move them into the model.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSurfaceTransformer.java
chrisdunelm: Maybe later?

src/main/java/com/google/api/codegen/transformer/csharp/CSharpSurfaceNamer.java
chrisdunelm: I'll wait until the Go PR is merged, then update this.

src/main/java/com/google/api/codegen/viewmodel/RetryCodesDefinitionView.java
chrisdunelm: In a later PR?

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSurfaceTransformer.java
```diff
@@ -0,0 +1,329 @@
...
+      SurfaceTransformerContext context) {
+    List<RetryCodesDefinitionView> definitions = new ArrayList<>();
+
+    final SurfaceNamer namer = context.getNamer();
+    for (Entry<String, ImmutableSet<Code>> retryCodesDef :
+        context.getInterfaceConfig().getRetryCodesDefinition().entrySet()) {
+      List<String> codeNames = FluentIterable.from(retryCodesDef.getValue())
+          .transform(new Function<Code, String>() {
+            @Override public String apply(Code code) {
+              return namer.methodName(Name.upperUnderscore(code.toString()));
```
chrisdunelm: Done

src/main/resources/com/google/api/codegen/csharp/gapic_client.snip
garrettjonesgoogle: I would argue that composed identifiers are already opaque. I don't know what the form of retry.name is here; I can only tell that RetryFilter is being added onto it. 

I'm trying to push for composing the strings in one place so that it is easier to predict where to look for something - either it is a direct string (find it in the template), or it is composed (find it in the surface namer).

src/main/java/com/google/api/codegen/viewmodel/RetryCodesDefinitionView.java
garrettjonesgoogle: Add a TODO then.

src/main/resources/com/google/api/codegen/csharp/gapic_client.snip
chrisdunelm: But every language will be adding their own fields to the model, which is going to cause a fair amount of model complication. From a single-langueg view-point (which is where myself, and I expect most people will see it) it won't be obvious why quite a few of the fields are there or what they're for, or if they're important.

src/main/java/com/google/api/codegen/viewmodel/RetryCodesDefinitionView.java
chrisdunelm: Done

src/main/java/com/google/api/codegen/transformer/csharp/CSharpSurfaceNamer.java
chrisdunelm: TODO added

chrisdunelm:
PTAL

src/main/resources/com/google/api/codegen/csharp/gapic_client.snip
garrettjonesgoogle: This new approach definitely goes in a direction away from the single-language point of view. It turns out that most people on the vkit team aren't working on just one language - they often deal with 2 or 3 regularly, and occasionally all of the rest (with certain features that we do for all languages). That was actually one of the (multiple) motivations for orienting the design more in this direction. Our four dynamic languages have quite a bit of pieces in common, and three of them were actually bootstrapped off of the same starting point (Python). 

Now maybe it's true that banning *all* identifier composition in the templates might end up being extreme. I think it's worked out ok so far; Java and PHP were written this way, and Go is about to be merged following this pattern. In any case, I kind of want to follow this particular experiment to its conclusion, despite being a little bit non-optimal in certain local cases. 

My hypothesis is that this approach would make it easier (and safer) for non-C# developers to make certain changes or additions, because they could depend on the system to make it safe for them. One particular example is enforcement of input data casing; we had some weird naming problems earlier because we assumed that certain input data was lower-underscore when it was actually lower-camel, producing really weird names for Python. When we go through the Name class, all input name casing is enforced, so we catch it at generation time. In your case, what happens if you want to form a variable name out of retry.name? You have to be aware that it is currently in upper camel, and then you would have to convert it to lower camel in the template, and then you're back into template casing hell again. With the new casing system, our snippet templates are cut down in size by about 30-50% (depending on the template). The short of it is that I'm currently wary of touching the current C# generation, but I would feel much less concern about touching it under the MVVM refactoring with no composition in the templates. I don't want the C# generation to be siloed away from the rest; it's under the ownership of the larger toolkit umbrella. 

Another reason I'm pushing for zero composition in the templates is that I'm concerned with a broken windows situation. It's usually a smaller initial effort compose in the templates, and if there are already cases where it's done in a template, it's not a big leap to add even more. 

Now all that said, if the experiment doesn't work out, I'm happy to back off on the requirement.

src/main/java/com/google/api/codegen/gapic/MainGapicProviderFactory.java
```diff
@@ -86,17 +89,20 @@
...
+          CommonGapicCodePathMapper.newBuilder()
+              .setPrefix("")
+              .setShouldAppendPackage(true)
+              .setPackageFilePathNameFormatter(new CSharpNameFormatter())
+              .build();
+      GapicProvider<? extends Object> mainProvider =
+          ViewModelGapicProvider.newBuilder()
               .setModel(model)
-              .setView(new InterfaceView())
-              .setContext(new CSharpGapicContext(model, apiConfig))
```
garrettjonesgoogle: Could you delete CSharpGapicContext and other orphaned classes as part of this PR?

src/main/java/com/google/api/codegen/transformer/ApiCallableTransformer.java
```diff
@@ -53,7 +58,7 @@ public ApiCallableTransformer() {
     return settingsMembers;
   }
 
-  private List<ApiCallableView> generateStaticLangApiCallables(MethodTransformerContext context) {
+  protected List<ApiCallableView> generateStaticLangApiCallables(MethodTransformerContext context) {
```
garrettjonesgoogle: Can this be switched back to private?

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
```diff
@@ -116,14 +132,33 @@ public StaticLangApiMethodView generateUnpagedListCallableMethod(
...
+      List<ParamWithSimpleDoc> additionalParams) {
     StaticLangApiMethodView.Builder methodViewBuilder = StaticLangApiMethodView.newBuilder();
 
     setCommonFields(context, methodViewBuilder);
     methodViewBuilder.name(context.getNamer().getApiMethodName(context.getMethod()));
-    setFlattenedMethodFields(context, fields, methodViewBuilder);
-    setStaticLangSyncReturnFields(context, methodViewBuilder);
+    methodViewBuilder.callableName(context.getNamer().getCallableName(context.getMethod()));
+    setFlattenedMethodFields(context, fields, additionalParams, false, methodViewBuilder);
+    setStaticLangReturnFields(context, false, methodViewBuilder);
```
garrettjonesgoogle: I can't tell what the "false" values are for here. Could you make a boolean that captures the notion of Sync/Async?

src/main/java/com/google/api/codegen/transformer/ParamWithSimpleDoc.java
```diff
@@ -0,0 +1,123 @@
...
+  public static List<ParamDocView> asParamDocViews(List<ParamWithSimpleDoc> ps) {
+    return FluentIterable.from(ps)
+        .transform(
+            new Function<ParamWithSimpleDoc, ParamDocView>() {
+              @Override
+              public ParamDocView apply(ParamWithSimpleDoc p) {
+                return p.asParamDocView();
+              }
+            })
+        .toList();
```
garrettjonesgoogle: I know this is a pain, but could you convert your remaining FluentIterable usages to plain loops? Particularly in simple cases like this, the FluentIterable approach is considerably more verbose (twice the lines) as compared to plain loops:

```
List<ParamDocView> views = new ArrayList<>();
for (ParamWithSimpleDoc param : ps) {
  views.add(param.asParamDocView());
}
return views;
```

I will be quite happy when we can use the much more concise Java 8 style.

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -68,6 +68,7 @@ public TypeNameConverter getTypeNameConverter() {
 
   public String getNotImplementedString(String feature) {
     return "$ NOT IMPLEMENTED: " + feature + " $";
+    //throw new RuntimeException(feature);
```
garrettjonesgoogle: remove

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -322,6 +328,11 @@ public String getApiMethodName(Method method) {
     return methodName(Name.upperCamel(method.getSimpleName()));
   }
 
+  /** The name of the surface async surface method which can call the given API method. */
```
garrettjonesgoogle: redundant "surface"

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -359,6 +377,11 @@ public boolean shouldImportRequestObjectParamElementType(Field field) {
     return new ArrayList<>();
   }
 
+  public List<String> getReturnDocLines(
```
garrettjonesgoogle: add a doc

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSurfaceTransformer.java
```diff
@@ -0,0 +1,434 @@
...
+    xsettingsClass.bundlingDescriptors(bundlingTransformer.generateDescriptorClasses(context));
+    xsettingsClass.retryCodesDefinitions(retryCodesDefs);
+    xsettingsClass.retryParamsDefinitions(retryParamsDefs);
+
+    // must be done as the last step to catch all imports
+    xsettingsClass.imports(
+        importTypeTransformer.generateImports(context.getTypeTable().getImports()));
+
+    String outputPath = pathMapper.getOutputPath(context.getInterface(), context.getApiConfig());
+    xsettingsClass.outputPath(outputPath + "/" + name + ".java");
```
garrettjonesgoogle: Don't you want ".cs"?

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
```diff
@@ -53,17 +53,33 @@ public ApiMethodTransformer() {
   }
 
   public StaticLangApiMethodView generatePagedFlattenedMethod(
-      MethodTransformerContext context, ImmutableList<Field> fields) {
+      MethodTransformerContext context,
+      ImmutableList<Field> fields,
+      List<ParamWithSimpleDoc> additionalParams) {
```
garrettjonesgoogle: I think I would prefer to have an overload that only took context & fields and passed an empty array list to this function. That would avoid the requirement of passing null in a lot of places, and you wouldn't need the if condition in setFlattenedMethodFields (an empty additionalParams list would be a no-op).

src/main/java/com/google/api/codegen/viewmodel/RequestObjectParamView.java
```diff
@@ -30,6 +34,17 @@
...
+  @Nullable // Used in C#
+  public abstract String defaultValue();
+
+  public boolean hasDefaultValue() {
+    return defaultValue() != null;
+  }
+  // Work around a limitation in the snippet language
+  public String defaultValue(String template) {
+    return defaultValue() != null ? String.format(template, defaultValue()) : "";
+  }
```
garrettjonesgoogle: You can remove this now, right?

src/main/java/com/google/api/codegen/viewmodel/StaticLangApiMethodView.java
```diff
@@ -65,6 +71,11 @@
   @Nullable
   public abstract RequestObjectMethodDetailView requestObjectMethod();
 
+  //@Nullable
+  //public abstract PageStreamingDescriptorClassView pagedView();
```
garrettjonesgoogle: delete the dead code

src/main/java/com/google/api/codegen/viewmodel/StaticLangXCommonView.java
```diff
@@ -0,0 +1,56 @@
...
+import com.google.auto.value.AutoValue;
+import javax.annotation.Nullable;
+
+// View of both api and settings for static language.
+// ViewModel members delegate to api.
+@AutoValue
+public abstract class StaticLangXCommonView implements ViewModel {
+  public abstract StaticLangXApiView api();
+
+  @Nullable // TODO: Remove
```
garrettjonesgoogle: still TODO?

src/main/resources/com/google/api/codegen/csharp/gapic_client.snip
```diff
@@ -0,0 +1,465 @@
...
+        {@pagedFlattenedMethodAsync(method, isImpl)}
+    @default
+        *** ERROR: Cannot handle method type '{@method.type.toString}' ***
+    @end
+@end
+
+@private flattenedMethodSync(method, isImpl)
+    @if isImpl
+        public override {@method.responseTypeName} {@method.name}(
+        {@""}    {@objectParams(method.methodParams)}) => {@method.callableName}.Sync(
```
garrettjonesgoogle: Looks like there are plenty more instances of `{@""}` that can be removed from this file.

src/main/java/com/google/api/codegen/viewmodel/StaticLangApiMethodView.java
chrisdunelm: Done

src/main/java/com/google/api/codegen/viewmodel/StaticLangXCommonView.java
chrisdunelm: Removed

src/main/resources/com/google/api/codegen/csharp/gapic_client.snip
chrisdunelm: Done

src/main/java/com/google/api/codegen/viewmodel/RequestObjectParamView.java
chrisdunelm: Done

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSurfaceTransformer.java
chrisdunelm: This isn't used in C#, so I've changed it to an empty string with a comment.

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
chrisdunelm: Done

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
chrisdunelm: Done

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
chrisdunelm: Done

src/main/java/com/google/api/codegen/transformer/ParamWithSimpleDoc.java
chrisdunelm: Done

src/main/java/com/google/api/codegen/viewmodel/RetryCodesDefinitionView.java
chrisdunelm: This is now done. New RetryDefinitionsTransformer class.

src/main/java/com/google/api/codegen/gapic/MainGapicProviderFactory.java
chrisdunelm: Done (I think)

src/main/java/com/google/api/codegen/transformer/ApiCallableTransformer.java
chrisdunelm: Done

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
chrisdunelm: Done

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
chrisdunelm: Done

src/main/resources/com/google/api/codegen/csharp/gapic_client.snip
chrisdunelm: Done

chrisdunelm:
All done. PTAL.

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
```diff
@@ -350,7 +434,7 @@ private ApiMethodDocView generateOptionalArrayMethodDoc(MethodTransformerContext
 
     docBuilder.mainDocLines(context.getNamer().getDocLines(context.getMethod()));
     List<ParamDocView> paramDocs =
-        getMethodParamDocs(context, context.getMethodConfig().getRequiredFields());
+        getMethodParamDocs(context, context.getMethodConfig().getRequiredFields(), null);
```
garrettjonesgoogle: I'd also prefer this to either pass an empty list or use an overload instead of using null.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpSurfaceNamer.java
```diff
@@ -0,0 +1,174 @@
...
+  @Override
+  public String retryBackoffMethodName(String key) {
+    return methodName(Name.from("get").join(key).join("retry").join("backoff"));
+  }
+
+  /** The method name of the timeout backoff for the given key */
+  @Override
+  public String timeoutBackoffMethodName(String key) {
+    return methodName(Name.from("get").join(key).join("timeout").join("backoff"));
+  }
```
garrettjonesgoogle: My preference is for functions that purely provide names to have a default implementation in SurfaceNamer instead of just using the not-implemented string. That way, if someone is generating for a similar scenario, they could re-use the names. I have tried to use the not-implemented string when there is not a sensible default. Your default names for the methods above seem sensible to me for SurfaceNamer.

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
chrisdunelm: Done

src/main/java/com/google/api/codegen/transformer/csharp/CSharpSurfaceNamer.java
chrisdunelm: Moved the three retry names into SurfaceNamer.

PR #501, opened 2016-09-21
Acronym table

Landrito:


src/main/java/com/google/api/codegen/util/CommonAcronyms.java
```diff
@@ -0,0 +1,38 @@
...
+          .put("IAM", "Iam")
+          .put("HTTP", "Http")
+          .put("XML", "Xml")
+          .put("API", "Api")
+          .build();
+
+  public static String replaceAcronyms(String str) {
+    for (String acronym : ACRONYMS.keySet()) {
+      str = str.replace(acronym, ACRONYMS.get(acronym));
+    }
```
garrettjonesgoogle: Let's hope no one uses the acronym "APIAM"

src/main/java/com/google/api/codegen/util/Name.java
```diff
@@ -223,4 +225,16 @@ private NamePiece(String identifier, CaseFormat caseFormat) {
       this.caseFormat = caseFormat;
     }
   }
+
+  private NamePiece replaceAcronyms(NamePiece namePiece) {
+    if (namePiece.caseFormat != CaseFormat.LOWER_CAMEL
+        && namePiece.caseFormat != CaseFormat.UPPER_CAMEL) {
+      return namePiece;
+    }
```
garrettjonesgoogle: It might only make sense to apply to upper camel.

src/main/java/com/google/api/codegen/util/CommonAcronyms.java
```diff
@@ -0,0 +1,38 @@
...
+  private static final ImmutableMap<String, String> ACRONYMS =
+      ImmutableMap.<String, String>builder()
+          .put("IAM", "Iam")
+          .put("HTTP", "Http")
+          .put("XML", "Xml")
+          .put("API", "Api")
+          .build();
+
+  public static String replaceAcronyms(String str) {
+    for (String acronym : ACRONYMS.keySet()) {
```
garrettjonesgoogle: iterate over entrySet so you don't have to do the get() below

src/main/java/com/google/api/codegen/util/Name.java
```diff
@@ -136,6 +136,7 @@ public String toLowerUnderscore() {
   private String toUnderscore(CaseFormat caseFormat) {
     List<String> newPieces = new ArrayList<>();
     for (NamePiece namePiece : namePieces) {
+      namePiece = replaceAcronyms(namePiece);
```
garrettjonesgoogle: I think it might make more sense to do this in Name.upperCamel; it is the entry point for upper camel strings.

src/main/java/com/google/api/codegen/util/CommonAcronyms.java
Landrito: We still need the key since we are replacing the key by its entry. Maybe there is another way that I am missing?

src/main/java/com/google/api/codegen/util/CommonAcronyms.java
Landrito: Agreed. I may want to add some documentation clearly stating that this case is not covered and will probably cause an error.

src/main/java/com/google/api/codegen/util/CommonAcronyms.java
garrettjonesgoogle:
```
  for (Map.Entry<String, String> mapping : ACRONYMS.entrySet()) {
    str = str.replace(mapping.getKey(), mapping.getValue());
  }
```

Landrito: Oh I see. I confused entrySet() with values().

src/main/java/com/google/api/codegen/util/Name.java
Landrito: By replacing the acronyms in Name.upperCamel this method is no longer needed.

src/main/java/com/google/api/codegen/util/CommonAcronyms.java
Landrito: I wrote a check to make sure this is not happening to avoid future headache. PTAL

src/main/java/com/google/api/codegen/util/CommonAcronyms.java
```diff
@@ -0,0 +1,80 @@
...
+      for (AcronymIndex acronym2 : acronymIndices) {
+        if (acronym1 == acronym2) continue;
+        if (acronym1.index >= acronym2.index
+            && acronym1.index < acronym2.index + acronym2.acronym.length()) {
+          return true;
+        }
+      }
+    }
+    return false;
+  }
```
garrettjonesgoogle: So it's neat that we can detect ambiguous replacements, but I'm concerned about the performance impact because this will be called at a relatively low level, e.g. for every upper camel string that we process. It includes an n^2 algorithm in it. I'm thinking that we could simplify this for now and handle more complex cases later. The simple way is to just search for "APIAM" (the only current possibility); we could add an alert in the list of acronyms to consider new combinations when new acronyms are added.

src/main/java/com/google/api/codegen/LanguageUtil.java
```diff
@@ -64,14 +68,17 @@ public static String lowerUnderscoreToLowerCamel(String name) {
...
     return CaseFormat.LOWER_CAMEL.to(CaseFormat.UPPER_CAMEL, name);
   }
 
   public static String lowerCamelToLowerUnderscore(String name) {
+    name = CommonAcronyms.replaceAcronyms(name);
     return CaseFormat.LOWER_CAMEL.to(CaseFormat.LOWER_UNDERSCORE, name);
   }
 
   public static String lowerCamelToUpperUnderscore(String name) {
+    name = CommonAcronyms.replaceAcronyms(name);
```
garrettjonesgoogle: Be consistent with the Name class and only apply replacements for upper camel.

src/main/java/com/google/api/codegen/util/CommonAcronyms.java
Landrito: Agreed.

garrettjonesgoogle:
LGTM

PR #503, opened 2016-09-22
Fixes an unresolved nit from review of PR #405

geigerj:


src/main/java/com/google/api/codegen/py/PythonGapicContext.java
```diff
@@ -253,14 +253,17 @@ private String returnTypeComment(
...
+  private String getTrimmedDocs(ProtoElement elt) {
     String description = "";
-    if (method.hasAttribute(ElementDocumentationAttribute.KEY)) {
-      String sphinxified = getSphinxifiedScopedDescription(method);
-      description = sphinxified.replaceAll("\\s*\\n\\s*", "\n");
+    if (elt.hasAttribute(ElementDocumentationAttribute.KEY)) {
+      description = getSphinxifiedScopedDescription(elt);
     }
-    return splitToLines(description);
+    return description.replaceAll("\\s*\\n\\s*", "\n");
```
bjwatson: This changes the logic a bit to move this outside the `if` test. Is that a deliberate bug fix?

geigerj: Whoops, no, that change was unintentional. Changed back to how it was.

bjwatson:
LGTM

PR #506, opened 2016-09-22
Conversion for api name should be from uppercamel not lowercamel

Landrito:
This makes a difference since acronym replacing only occurs on upper camel causing toolkit to not WAI.  An baseline test case needs to be made to cover acronym replacements.


geigerj:
Could you do new baseline test case as part of this PR? You can just change the API name on the existing `no_path_templates` test(s).

Landrito:
Will do

Landrito:
Ready for another review. Thanks for pointing out no path as a place to put the baseline test. I was a little curious as to the best way to add an acronym to a service name.

src/main/java/com/google/api/codegen/util/CommonAcronyms.java
```diff
@@ -30,6 +30,7 @@
           .put("HTTP", "Http")
           .put("XML", "Xml")
           .put("API", "Api")
+          .put("TEST", "Test")
```
geigerj: Hm, would rather not modify the real class just to add a test acronym. Maybe just call the service `NoTemplatesAPIService` instead?

Landrito:
Fixed.

Landrito:
Thanks!

PR #507, opened 2016-09-22
Add media upload placeholders in discovery sample gen.

tcoffee-google:
Should be later replaced by more elaborate code illustrating media
upload funcionality in client libraries. It remains to determine where
media upload is required for method calls to succeed.

Addresses b/30637043.

Also removes unused baseline test files corresponding to removed
outdated discovery docs.


saicheems:
lgtm

PR #508, opened 2016-09-22
Clarify placeholder TODO for singleton parameters in discovery samples.

tcoffee-google:
Add feature omitted for non-Java languages in earlier PR
https://github.com/googleapis/toolkit/pull/487.

Not needed for Go due to alternative placeholder comment
formatting (motivated by gofmt prohibition of multiple consecutive blank
lines).

### DIFF
--- a/src/main/resources/com/google/api/codegen/csharp/discovery_fragment.snip
+++ b/src/main/resources/com/google/api/codegen/csharp/discovery_fragment.snip
@@ -117,7 +117,11 @@
     {@BREAK}
     @if sample.params
 
-        // {@TODO()} Change placeholders below to desired parameter values for the `{@sample.methodName}` method:
+        @if context.isSingleton(sample.params)
+            // {@TODO()} Change placeholder below to desired parameter value for the `{@sample.methodName}` method:
+        @else
+            // {@TODO()} Change placeholders below to desired parameter values for the `{@sample.methodName}` method:
+        @end
 
         @join param : sample.params
             {@description(param.description)}
--- a/src/main/resources/com/google/api/codegen/nodejs/discovery_fragment.snip
+++ b/src/main/resources/com/google/api/codegen/nodejs/discovery_fragment.snip
@@ -99,7 +99,11 @@
        params = context.getFlatMethodParams(method)
     @if params
 
-      // {@TODO()} Change placeholders below to desired parameter values for the `{@methodName}` method:
+      @if context.isSingleton(params)
+        // {@TODO()} Change placeholder below to desired parameter value for the `{@methodName}` method:
+      @else
+        // {@TODO()} Change placeholders below to desired parameter values for the `{@methodName}` method:
+      @end
 
       @join param : params
         @let paramField = context.getField(signatureType, param), \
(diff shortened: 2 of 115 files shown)


saicheems:
So I asked Les, the DPE for Java what his opinion was between these two formats:

```
========
Option 1
========

// TODO: Change placeholders below to desired parameter value for the `Delete` method:

// The id of the project.
String id = "";

// The number of items to return.
int count = 0;

Request request = Service.Delete(id, count);
Response response = request.Execute();
 
// TODO: Change code below to process the `response` object:
System.out.println(response);
```

```
========
Option 2
========

// The id of the project.
String id = ""; // TODO: Update placeholder value(s).

// The number of items to return.
int count = 0; // TODO: Update placeholder value(s).

Request request = Service.Delete(id, count);
Response response = request.Execute();
System.out.println(response);
```

Of the two, he preferred option 2 - stating that most IDEs are configured to highlight TODOs, and it was more obvious what to do.
Specifically, he said that he preferred the `TODO` to the right of the fields.

I really think we should adopt that format.

tcoffee-google:
@saicheems On a point of detail, did you show Les the samples in their entirety? I consider the surrounding line-breaking essential to making the instructions clear, which is why I agreed to the change for Go. I feel that IDE TODO highlighting is a poor substitute to making things visually apparent.

As I've said, there are two main reasons I haven't adopted this format: (1) in cases like the ones shown, it obscures where the TODOs and the placeholder values are in the code, and makes it more difficult to edit the code into a satisfactory state; and (2) in cases where we provide further information about appropriate values, it clashes with the (sometimes lengthy) inline comments we use for that purpose.

In short, I'm still not convinced we have a problem, and if we do, there are some other problems that need to be addressed by any proposed solution.

saicheems:
I did also show Les a whole sample in it's current format, but I didn't ask for further opinion with the 3-line breaks around the parameters.

(1) I disagree, having TODOs to the right of the statement highlights exactly what a user has to do and leaves no ambiguity. I also think the additional work needed to edit a field with a TODO attached to it is negligible and not really an issue in any way (nobody will drop a sample because editing was too time consuming).

(2) I have an idea to eliminate the inline comment issue - we can use the override mechanism to edit the comment above a field to include additional instructions, or insert them automatically to the end of field comments in `ApiaryConfigToSampleConfigConverter.java` (on a new line). There's no reason we *have* to include clarification at the end of the TODO.

tcoffee-google:
Even if we ultimately decide to change the format, this change is an improvement over the current state, so if there are no immediate issues, let's merge it.

saicheems:
Ok, no objections

tcoffee-google:
@saicheems Re: (1), I meant 'obscures', not 'muddles' --- it makes it harder to see that the TODOs and missing values are there. Re: (2), note that many of the field descriptions are quite lengthy and unlikely to be fully read, so having potentially helpful hints appended to the end will probably often be useless, unless there is a way to make them stand out visually. I could be convinced that the current solution for Go is better in all cases, but I would want fresh eyes to compare entire samples with the current formatting (particularly as shown by API Explorer).

PR #510, opened 2016-09-22
Fix style consistency for the Java discovery sample

saicheems:


saicheems:
compilecheck passes for all discovery docs.

tcoffee-google:
@saicheems So we just need to update the template so that the baseline files don't change (that's why they're there). My approach would probably be to start with the current discovery_fragment.snip and edit to use the references from sample.snip. Should also incorporate the fix in https://github.com/googleapis/toolkit/pull/508. I'm happy to help if needed.

saicheems:
I wanted to change the format - the differences from the baselines are intentional.
While refactoring, I thought that this new format was more extensible, and cleaner than the old one. It has a separate `main`, `createService` and `run`, which I think follows Java style more nicely (main creates a new instance of the class to call run on), and having `createService` in it's own function makes supporting new auth mechanisms much easier. It also handles exceptions a bit nicer by logging them.

I asked Victor some time ago on what his opinion was between this format and the old one, and he also preferred this one. If you could provide feedback on the new format and comment on things you'd prefer to be changed, that'd be helpful.

(I also think that #508 isn't the right move for Java - I really prefer TODOs to the right of statements, and from my conversations with the toolkit Java folks, and with the Java DPE, that's preferred for this language)

tcoffee-google:
Oh I see ... I would have preferred to make the refactoring change separate from a template change, but I'll take a closer look.

garrettjonesgoogle:
+1 : in general, it's much easier to review refactoring changes and output changes separately from each other.

saicheems:
Right - I hand't split it out because there was no content change from the original review, I'll do that and tag everyone again.

saicheems:
(This commit was just to put back the already reviewed surface changes in #476 - I've reset it now to include only the logging baseline - please comment if there are any changes you'd like me to make)

saicheems:
PTAL

src/test/java/com/google/api/codegen/testdata/discoveries/datastore.v1beta3.json.overrides
```diff
@@ -0,0 +1,5 @@
+{
```
vchudnov-g: Are these all the overrides we've needed thus far for Java?

src/test/java/com/google/api/codegen/testdata/discoveries/java/java_logging.v2beta1.json.baseline
```diff
@@ -1,871 +1,962 @@
...
         continue;
-
-      for (LogEntry logEntry : response.getEntries()) {
-        // TODO: Change code below to process each `logEntry` resource:
-        System.out.println(logEntry);
       }
-
-      requestBody.setPageToken(response.getNextPageToken());
+      for (LogEntry item : response.getEntries()) {
+        System.out.println(item);
```
vchudnov-g: We should keep the TODO here to highlight this `println` is just for sample purposes.

src/test/java/com/google/api/codegen/testdata/discoveries/datastore.v1beta3.json.overrides
saicheems: There are many more that are already checked in. This one is included because the old datastore version was removed.

src/test/java/com/google/api/codegen/testdata/discoveries/java/java_logging.v2beta1.json.baseline
saicheems: Done

src/test/java/com/google/api/codegen/testdata/discoveries/java/java_logging.v2beta1.json.baseline
```diff
@@ -1,871 +1,966 @@
...
-    // https://developers.google.com/identity/protocols/application-default-credentials
+  public void run() throws IOException, GeneralSecurityException {
+    // TODO: Assign values to the desired fields of `requestBody`.
+    WriteLogEntriesRequest requestBody = new WriteLogEntriesRequest();
+
+    Logging service = createService();
+    Write request = service.entries().write(requestBody);
+
+    WriteLogEntriesResponse response = request.execute();
+    System.out.println(response);
```
vchudnov-g: Just like we do for loops, let's have a consistent message along the lines of  "TODO: process response" above this `println`.

Are there any other places where get a response and should do this?

saicheems: Discussed in person - sticking to the way it is.

saicheems:
If there are no other comments, I'd like to merge this.

vchudnov-g:
Update the PR title to reflect the current content.

saicheems:
Done

vchudnov-g:
LGTM.

Thanks!

PR #512, opened 2016-09-23
Resolve name clashes between Python enums

### LINKED ISSUE #511: Python Enum aliases need disambiguation logic
### What

We should automatically disambiguate if there are multiple enums of the same simple name in an API.

### Why

This has cropped up in the Language API (multiple enums of simple name `Type`), Vision (`Type`),  and Datastore (`Operator`)

### How

We can rely on the proto namespace. Typically this will mean prefixing the enum with the name of its parent message or service. In the case of Vision, this should mean that [(1)](https://github.com/googleapis/googleapis/blob/master/google/cloud/vision/v1/image_annotator.proto#L48) becomes `FeatureType`, [(2)](https://github.com/googleapis/googleapis/blob/master/google/cloud/vision/v1/image_annotator.proto#L114) becomes `LandmarkType`.

geigerj:
Introduces a symbol table for enums that handles disambiguation logic.
Previously, the enum simple name was always used. If an API contained
multiple enums with a common simple name, this could lead to name
clashes.

Fixes #511

### DIFF
--- a/src/main/java/com/google/api/codegen/py/PythonImportHandler.java
+++ b/src/main/java/com/google/api/codegen/py/PythonImportHandler.java
@@ -47,8 +47,13 @@ public class PythonImportHandler {
    */
   private final BiMap<ProtoFile, String> fileImports = HashBiMap.create();
 
+  private final PythonEnumSymbolTable enumTable;
+
   /** This constructor is for the main imports of a generated service file */
-  public PythonImportHandler(Interface service, ApiConfig apiConfig) {
+  public PythonImportHandler(
+      Interface service, ApiConfig apiConfig, PythonEnumSymbolTable enumTable) {
+    this.enumTable = enumTable;
+
     // Add non-service-specific imports.
     addImportStandard("json");
     addImportStandard("os");
@@ -61,7 +66,7 @@ public class PythonImportHandler {
     addImportExternal("google.gax", "path_template");
 
     // only if add enum import if there are enums
-    if (!Iterables.isEmpty(new PythonContextCommon().getEnumTypes(service.getModel()))) {
+    if (!Iterables.isEmpty(enumTable.getEnums())) {
       addImportLocal(apiConfig.getPackageName(), "enums");
     }
 
@@ -87,7 +92,8 @@ public class PythonImportHandler {
   }
 
   /** This constructor is used for doc messages. */
-  public PythonImportHandler(ProtoFile file) {
+  public PythonImportHandler(ProtoFile file, PythonEnumSymbolTable enumTable) {
+    this.enumTable = enumTable;
     for (MessageType message : file.getMessages()) {
       for (Field field : message.getMessageFields()) {
         MessageType messageType = field.getType().getMessageType();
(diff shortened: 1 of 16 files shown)


src/main/java/com/google/api/codegen/py/PythonImportHandler.java
```diff
@@ -61,7 +66,7 @@ public PythonImportHandler(Interface service, ApiConfig apiConfig) {
     addImportExternal("google.gax", "path_template");
 
     // only if add enum import if there are enums
-    if (!Iterables.isEmpty(new PythonContextCommon().getEnumTypes(service.getModel()))) {
+    if (!Iterables.isEmpty(enumTable.getEnums())) {
```
Landrito: I don't think that the enumTable should be in the import handler since it does not add any extra functionality to the class. Maybe make it "global" to the snippets?

geigerj: Acknowledged. You're right that PythonEnumSymbolTable doesn't make too much sense being owned by PythonImportHandler. I will make this change on Monday.

geigerj: Done.

src/test/java/com/google/api/codegen/testdata/nodejs_doc_message_library.baseline
```diff
@@ -756,21 +756,69 @@ var Book = {
  * @property {Object} field2
  *   This object should have the same structure as [SomeMessage2]{@link SomeMessage2}
  *
+ * @property {number} alignment
+ *   The number should be among the values of [Alignment]{@link Alignment}
```
geigerj: @jmuk for context, this PR is connected to a bug in Python where there are multiple enums of the same simple name in an API. (For example, the Vision API has two different enums called `Type`.) Does NodeJS handle this properly currently, i.e., between `[Alignment]{@link Alignment}` on this line, and on line 792, which represent two different enums?

jmuk: In Node, enum values will be children of the parent namespace/message. So in the case like this, they would be accessed through `SomeMessage.Alignment.GOOD` or `SomeMessage2.Alignment.CENTERED`.

So that would be a bug. That would be `[SomeMessage.Alignment]{@link SomeMessage.Alignment}` -- so then jsdoc can take care of the link address.

You can leave it as-is (but file a bug in toolkit and assign me) or fix it by yourself around `linkForMessage` in `NodeJSGapicContext`.

src/test/java/com/google/api/codegen/testdata/python_enum_library.baseline
```diff
@@ -55,6 +55,36 @@ class Rating(object):
...
+    Tests service with two enums of the same simple name
+
+    Attributes:
+      GOOD (int)
+      NEUTRAL (int)
+      EVIL (int)
+    """
+    GOOD = 0
+    NEUTRAL = 1
+    EVIL = 2
```
bjwatson: Hilarious! Remind me of D&D when I was a kid. 👍

src/test/java/com/google/api/codegen/testdata/nodejs_doc_message_library.baseline
bjwatson: @jmuk What exactly is the bug for NodeJS? It sounds like from your description that the message type namespacing keeps the different enum types cleanly separated.

bjwatson: Is it a documentation bug above, not a surface bug?

geigerj: Right, just a documentation bug. I just wanted to check with Jun because I was suspicious when the test baseline contained the same link for the two different enums.

bjwatson:
LGTM

PR #514, opened 2016-09-26
C# gapic snippet generator

chrisdunelm:
Mostly done. Just a few things still to do.
Known problems:
* duplicate snippet method names
* initialization of some maps and lists missing generic parameters
* Incorrect enum initialization name
* probably more...


src/test/java/com/google/api/codegen/testdata/csharp_gapic_snippets_library.baseline
```diff
@@ -0,0 +1,696 @@
...
+            // End snippet
+        }
+
+        public async Task GetShelfAsync()
+        {
+            // Snippet: GetShelfAsync(string,SomeMessage,CallSettings)
+            // Additional: GetShelfAsync(string,SomeMessage,CancellationToken)
+            // Create client
+            LibraryServiceClient libraryServiceClient = LibraryServiceClient.Create();
+            // Initialize request argument(s)
```
jskeet: Nit for the future... we know how many arguments there are, so we don't need the (s).

src/test/java/com/google/api/codegen/testdata/csharp_gapic_snippets_library.baseline
```diff
@@ -0,0 +1,696 @@
...
+        public async Task ListShelvesAsync()
+        {
+            // Snippet: ListShelvesAsync(string,int?,CallSettings)
+            // Additional: ListShelvesAsync(string,int?,CancellationToken)
+            // Create client
+            LibraryServiceClient libraryServiceClient = LibraryServiceClient.Create();
+            // Initialize request argument(s)
+
+            // Make the request
+            IPagedAsyncEnumerable<ListShelvesResponse,Shelf> response = await libraryServiceClient.ListShelvesAsync();
```
jskeet: We might want to consider adding a line break after the `=` here, partly because docfx has such narrow docs.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_snippets_library.baseline
```diff
@@ -0,0 +1,696 @@
...
+                Console.Writeline(item);
+            }
+
+            // Or iterate over fixed-sized pages, lazily performing RPCs as required
+            int pageSize = 10;
+            foreach (FixedSizePage<Shelf> page in response.AsPages().WithFixedSize(pageSize))
+            {
+                // Do something with each page of items
+                Console.Writeline("A page of results:");
+                foreach (Shelf item in page) {
```
jskeet: Nit: `{` on new line.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_snippets_library.baseline
```diff
@@ -0,0 +1,696 @@
...
+        }
+
+        public void PublishSeries()
+        {
+            // Snippet: PublishSeries(Shelf,IEnumerable<Book>,uint,CallSettings)
+            // Create client
+            LibraryServiceClient libraryServiceClient = LibraryServiceClient.Create();
+            // Initialize request argument(s)
+            Shelf shelf = new Shelf();
+            IEnumerable<Book> books = new List<???>();
```
jskeet: Is this one of the missing generic bits? I haven't looked at the code yet, but we seem to know the types...

src/test/java/com/google/api/codegen/testdata/csharp_gapic_snippets_library.baseline
```diff
@@ -0,0 +1,696 @@
...
+
+        public async Task UpdateBookAsync()
+        {
+            // Snippet: UpdateBookAsync(string,Book,CallSettings)
+            // Additional: UpdateBookAsync(string,Book,CancellationToken)
+            // Create client
+            LibraryServiceClient libraryServiceClient = LibraryServiceClient.Create();
+            // Initialize request argument(s)
+            string formattedName = LibraryServiceClient.FormatBookName("[SHELF]", "[BOOK]");
+            Book book = new Book();
```
jskeet: It's a shame we don't have a better update story here... maybe this would be a pattern to spot in the future - get the book, modify it, call update...

src/test/java/com/google/api/codegen/testdata/csharp_gapic_snippets_library.baseline
```diff
@@ -0,0 +1,696 @@
...
+
+        public async Task AddCommentsAsync()
+        {
+            // Snippet: AddCommentsAsync(string,IEnumerable<Comment>,CallSettings)
+            // Additional: AddCommentsAsync(string,IEnumerable<Comment>,CancellationToken)
+            // Create client
+            LibraryServiceClient libraryServiceClient = LibraryServiceClient.Create();
+            // Initialize request argument(s)
+            string formattedName = LibraryServiceClient.FormatBookName("[SHELF]", "[BOOK]");
+            IEnumerable<Comment> comments = new []
```
jskeet: Can we remove the space between the `new` and the `[]`?

src/test/java/com/google/api/codegen/testdata/csharp_gapic_snippets_library.baseline
```diff
@@ -0,0 +1,696 @@
...
+            {
+                new Comment
+                {
+                    Comment = ByteString.CopyFromUtf8(""),
+                    Stage = Stage.UNSET,
```
jskeet: This is the enum name bit that's broken, yes? That should be reasonably easy to fix - the C# enumification is painstakingly documented in protoc :) Let me know if you want me to find it.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_snippets_library.baseline
```diff
@@ -0,0 +1,696 @@
...
+            // Initialize request argument(s)
+            string formattedName = LibraryServiceClient.FormatArchivedBookName("[ARCHIVE_PATH]", "[BOOK]");
+            // Make the request
+            Book response = libraryServiceClient.GetBookFromArchive(formattedName);
+            // End snippet
+        }
+
+        public async Task UpdateBookIndexAsync()
+        {
+            // Snippet: UpdateBookIndexAsync(string,string,IDictionary<string, string>,CallSettings)
```
jskeet: Hmm. I've no idea whether the snippet generator works with generics at the moment. We'll have to see :)

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSurfaceClientTransformer.java
```diff
@@ -0,0 +1,46 @@
...
+ * limitations under the License.
+ */
+package com.google.api.codegen.transformer.csharp;
+
+import com.google.api.codegen.ApiConfig;
+import com.google.api.codegen.gapic.GapicCodePathMapper;
+import com.google.api.codegen.transformer.SurfaceTransformerContext;
+import java.io.File;
+
+public class CSharpGapicSurfaceClientTransformer extends CSharpGapicSurfaceCommonTransformer {
```
garrettjonesgoogle: I would very strongly prefer the existing CSharpGapicSurfaceTransformer to be parameterized over making subclasses of a common variant.

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -252,26 +254,27 @@ private InitCodeView buildInitCodeView(
...
-        default:
-          throw new RuntimeException("unhandled line type: " + item.getLineType());
-      }
+      surfaceLines.add(generateSurfaceInitCodeLine(context, item));
     }
     return surfaceLines;
   }
 
+  private InitCodeLineView generateSurfaceInitCodeLine(
+      MethodTransformerContext context, InitCodeNode SpecItemNode) {
```
garrettjonesgoogle: the variable SpecItemNode should be lower camel; it looks like there was a previous mistake that put SpecItemNodes in upper camel - it should be lower camel too.

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -314,10 +317,13 @@ private InitCodeLineView generateListInitCodeLine(
     surfaceLine.elementTypeName(
         typeTable.getAndSaveNicknameForElementType(item.getType().makeOptional()));
     List<String> entries = new ArrayList<>();
+    List<InitCodeLineView> views = new ArrayList<>();
     for (InitCodeNode child : item.getChildren().values()) {
       entries.add(namer.varName(child.getIdentifier()));
+      views.add(generateSurfaceInitCodeLine(context, child));
     }
     surfaceLine.elementIdentifiers(entries);
+    surfaceLine.elementViews(views);
```
garrettjonesgoogle: I prefer not to use "view" for fields holding the views in other views; that should be obvious. This should just be "elements". Also name the local variable here accordingly.

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -343,6 +349,7 @@ private InitCodeLineView generateMapInitCodeLine(
           typeTable.renderPrimitiveValue(
               item.getType().getMapKeyField().getType(), entry.getKey()));
       mapEntry.value(context.getNamer().getVariableName(entry.getValue().getIdentifier(), null));
+      mapEntry.valueView(generateSurfaceInitCodeLine(context, entry.getValue()));
```
garrettjonesgoogle: I would prefer the existing one to be called "valueString" and this one be "value"

src/main/resources/com/google/api/codegen/csharp/gapic_snippets.snip
```diff
@@ -0,0 +1,229 @@
...
+    //
+    // Unless required by applicable law or agreed to in writing, software
+    // distributed under the License is distributed on an "AS IS" BASIS,
+    // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    // See the License for the specific language governing permissions and
+    // limitations under the License.
+
+    // Generated code. DO NOT EDIT!
+
+    @join import : common.api.imports
```
garrettjonesgoogle: I would much prefer this snippets file to receive a different object from the one that the main surface file uses. This snippet file uses a very small subset of the data, and thus it is a little confusing that it is used here. You might want something like SnippetsFileView or something like that. You could then put in exactly the api methods that you want to generate snippets for, and you wouldn't need the if condition below to skip FlattenedAsyncCancellationTokenMethod.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSurfaceClientTransformer.java
chrisdunelm: Done

src/main/resources/com/google/api/codegen/csharp/gapic_snippets.snip
chrisdunelm: Done

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
chrisdunelm: Done

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
chrisdunelm: Done

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
chrisdunelm: Done

src/test/java/com/google/api/codegen/testdata/csharp_gapic_snippets_library.baseline
chrisdunelm: Done

src/test/java/com/google/api/codegen/testdata/csharp_gapic_snippets_library.baseline
chrisdunelm: Ack

src/test/java/com/google/api/codegen/testdata/csharp_gapic_snippets_library.baseline
chrisdunelm: Done.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_snippets_library.baseline
chrisdunelm: Ack :)

src/test/java/com/google/api/codegen/testdata/csharp_gapic_snippets_library.baseline
chrisdunelm: Done.
Yes, this was one of the missing generics.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_snippets_library.baseline
chrisdunelm: Ack

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSurfaceClientTransformer.java
```diff
@@ -54,7 +54,7 @@
 import java.util.Arrays;
 import java.util.List;
 
-public class CSharpGapicSurfaceTransformer implements ModelToViewTransformer {
+public class CSharpGapicSurfaceClientTransformer implements ModelToViewTransformer {
```
garrettjonesgoogle: So I consider "Surface" and "Client" to be redundant here. I would recommend the original name, "CSharpGapicSurfaceTransformer".

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSurfaceSnippetsTransformer.java
```diff
@@ -0,0 +1,190 @@
...
+import com.google.api.tools.framework.model.Model;
+import com.google.common.collect.ImmutableList;
+import java.io.File;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public class CSharpGapicSurfaceSnippetsTransformer implements ModelToViewTransformer {
+
+  private static final String XAPI_TEMPLATE_FILENAME = "csharp/gapic_snippets.snip";
```
garrettjonesgoogle: A better name would be SNIPPETS_TEMPLATE_FILENAME

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSurfaceSnippetsTransformer.java
```diff
@@ -0,0 +1,190 @@
...
+    return Arrays.asList(XAPI_TEMPLATE_FILENAME);
+  }
+
+  private ModelTypeTable createTypeTable(String implicitPackageName) {
+    return new ModelTypeTable(
+        new CSharpTypeTable(implicitPackageName),
+        new CSharpModelTypeNameConverter(implicitPackageName));
+  }
+
+  private void addCommonImports(SurfaceTransformerContext context) {
```
garrettjonesgoogle: You could make a CSharpCommonTransformer (or CSharpSharedTransformer) class which does work that is the same between the transformers, and could include:

* This function (addCommonImports)
* The definitions of pagedMethodAdditionalParams and other additional params
* makeParam

src/main/resources/com/google/api/codegen/csharp/gapic_snippets.snip
```diff
@@ -0,0 +1,229 @@
+@extends "csharp/common.snip"
+
+@snippet generate(api)
```
garrettjonesgoogle: I think "snippets" would be a more natural name for the object here. To resolve the naming conflict with the snippet function named "snippets", I recommend renaming that to "renderSnippets".

src/main/resources/com/google/api/codegen/csharp/gapic_snippets.snip
```diff
@@ -0,0 +1,229 @@
...
+    @case "FlattenedMethod"
+        {@flattenedMethodSync(api, method)}
+    @case "FlattenedAsyncCallSettingsMethod"
+        {@flattenedMethodAsyncCallSettings(api, method)}
+    @case "PagedFlattenedMethod"
+        {@pagedFlattenedMethodSync(api, method)}
+    @case "PagedFlattenedAsyncMethod" 
+        {@pagedFlattenedMethodAsync(api, method)}
+    @default
+        *** ERROR: Cannot handle method type '{@method.type.toString}' ***
```
garrettjonesgoogle: I'd prefer to have the unhandled cases consistent, so we can do an automated check on them appearing in the output. The existing pattern is the one you used in renderInitValue below:

`$unhandledCase: {@initValue.type}$`

src/main/resources/com/google/api/codegen/csharp/gapic_snippets.snip
```diff
@@ -0,0 +1,229 @@
...
+
+@private sampleMethodCallArgList(fieldSettings)
+    @join fieldSetting : fieldSettings on ", "
+        {@fieldSetting.identifier}
+    @end
+@end
+
+@private topLevelInitCode(initCodeSpec)
+    @join line : initCodeSpec.topLevelLines
+        {@initCode(initCodeType(line).add(" ").add(line.identifier).add(" = "), line)};
```
garrettjonesgoogle: Could you figure out a different way of doing this? I'd rather have duplication in the initCode snippet, because it's hard to tell what's going on here.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_snippets_library.baseline
chrisdunelm: Yes, this is a broken enum name.
This is now fixed, by porting the enum renaming code from https://github.com/google/protobuf/blob/master/src/google/protobuf/compiler/csharp/csharp_helpers.cc

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSurfaceClientTransformer.java
chrisdunelm: Renamed to "CSharpGapicClientTransformer", with the snippets transformer named "CSharpGapicSnippetsTransformer".
This also matches the naming of the .snip files.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSurfaceSnippetsTransformer.java
chrisdunelm: Done

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSurfaceSnippetsTransformer.java
chrisdunelm: Done

src/main/resources/com/google/api/codegen/csharp/gapic_snippets.snip
chrisdunelm: Done.
It actually does allow identical function and parameter names, but I've changed the function as you suggest for clarity.

src/main/resources/com/google/api/codegen/csharp/gapic_snippets.snip
chrisdunelm: Done

src/main/resources/com/google/api/codegen/csharp/gapic_snippets.snip
chrisdunelm: Done

chrisdunelm:
All done. PTAL

src/main/java/com/google/api/codegen/transformer/csharp/CSharpModelTypeNameConverter.java
```diff
@@ -107,7 +132,29 @@ public TypeName getTypeName(ProtoElement elem) {
...
   @Override
   public TypedValue getZeroValue(TypeRef type) {
-    return TypedValue.create(getTypeName(type), "ZERO_VALUE"); // TODO: Generate correctly
+    if (type.isMap()) {
+      TypeName keyTypeName = getTypeName(type.getMapKeyField());
+      TypeName valueTypeName = getTypeName(type.getMapValueField());
+      String pattern =
+          "new %s<" + keyTypeName.getNickname() + ", " + valueTypeName.getNickname() + ">()";
+      return TypedValue.create(
+          typeNameConverter.getTypeName("System.Collections.Generic.Dictionary"), pattern);
```
garrettjonesgoogle: This implementation loses the ability to defer name clash resolution because it gets the nicknames without consulting the type table. I think you need two layers of templating here (and below too). You need almost what is in getTypeName(TypeRef) to get the TypeName to pass to TypedValue, except you would want Dictionary instead of IDictionary as the type name. So something like:

TypeName implTypeName = getImplTypeName(type); // uses Dictionary instead of IDictionary
return TypedValue.create(implTypeName, "new %s()";

src/main/java/com/google/api/codegen/transformer/csharp/CSharpSurfaceNamer.java
```diff
@@ -30,16 +30,78 @@
...
+import com.google.common.base.Function;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import java.util.List;
+import java.util.Map;
 
 public class CSharpSurfaceNamer extends SurfaceNamer {
 
+  public static final class MethodFields {
+    public MethodFields(Method method, List<Field> fields) {
```
garrettjonesgoogle: Will AutoValue not work for this class? (I'm suspecting that using method.getFullName() for hashCode/equals would get in the way of that)

(entire class becomes obsolete with my suggestion about redoing generateMethods)

src/test/java/com/google/api/codegen/testdata/csharp_gapic_snippets_library.baseline
```diff
@@ -0,0 +1,783 @@
...
+            // Create client
+            LibraryServiceClient libraryServiceClient = LibraryServiceClient.Create();
+            // Initialize request argument(s)
+            Shelf shelf = new Shelf();
+            // Make the request
+            Shelf response = libraryServiceClient.CreateShelf(shelf);
+            // End snippet
+        }
+
+        public async Task GetShelf1Async()
```
garrettjonesgoogle: It is very odd to see a number in the middle of an identifier like this instead of at the end...

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSnippetsTransformer.java
```diff
@@ -0,0 +1,181 @@
...
+    snippetsBuilder.snippetMethods(generateMethods(context));
+
+    // must be done as the last step to catch all imports
+    snippetsBuilder.imports(
+        importTypeTransformer.generateImports(context.getTypeTable().getImports()));
+
+    return snippetsBuilder.build();
+  }
+
+  private List<StaticLangApiMethodView> generateMethods(SurfaceTransformerContext context) {
```
garrettjonesgoogle: *Everyone stand back, I have an idea!*

There are a couple issues in this PR that I would like to address with one solution. The first is that the whole pathway of `generateMethodNameMap` -> `Map<MethodFields, Function<SurfaceNamer, String>>` -> `CSharpSurfaceNamer` is somewhat hard to follow. The second is that in the snippet, the "snippets" object is passed everywhere. 

My suggestion is to create a StaticLangApiMethodSnippetView class which would hold a StaticLangApiMethodView plus whatever extra information is necessary to render each snippet beyond what is already in StaticLangApiMethodView. It would include:

* snippetMethodName
* apiClassName (instead of clientTypeName, to be consistent)
* apiVariableName
* callerResponseTypeName

Now to form snippetMethodName, just create a SymbolTable object within this method (`generateMethods`) and use it to generate unique method names based off of the result of getApiMethodName / getAsyncApiMethodName.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpCommonTransformer.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.List;
+
+public class CSharpCommonTransformer {
+
+  public void addCommonImports(SurfaceTransformerContext context, boolean includeByteString) {
+    ModelTypeTable typeTable = context.getTypeTable();
+    // Common imports, only one class per required namespace is needed.
+    if (includeByteString) {
+      typeTable.saveNicknameFor("Google.Protobuf.Bytestring");
+    }
```
chrisdunelm: Done

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSnippetsTransformer.java
```diff
@@ -0,0 +1,181 @@
...
+  private Map<MethodFields, Function<SurfaceNamer, String>> generateMethodNameMap(
+      Model model, ApiConfig apiConfig) {
+    Map<MethodFields, Function<SurfaceNamer, String>> nameMap = new HashMap<>();
+    // Multiple interfaces with the same mixin will create method name collisions.
+    // This doesn't matter as the names will be the same in all interfaces.
+    for (Interface interfaze : new InterfaceView().getElementIterable(model)) {
+      for (MethodConfig methodConfig : apiConfig.getInterfaceConfig(interfaze).getMethodConfigs()) {
+        if (methodConfig.isFlattening()) {
+          final Method method = methodConfig.getMethod();
+          List<ImmutableList<Field>> fieldss = methodConfig.getFlattening().getFlatteningGroups();
```
chrisdunelm: This was deliberate, it's a list of lists. Changed to fieldGroups for clarity.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_snippets_library.baseline
chrisdunelm: This is desired. "Async" is really meant to be a suffix.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpModelTypeNameConverter.java
chrisdunelm: Thanks, done.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpSurfaceNamer.java
chrisdunelm: It uses a deep equality of the field list, so requires a custom hashcode and equals. I expect this is overridable in AutoValue, but it didn't really seem worth it.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSnippetsTransformer.java
chrisdunelm: Yes, I agree this is an improvement.
However, I'd rather not do it in this PR, as it will involve changing non-C# code and snippets, increasing the scope of this PR.
I've added a todo in the C# code, and I'll do a follow-up PR.

The SymbolTable class doesn't quite do what's required - it returns the first version of a name with no suffix, whereas I'd like a suffix of "1" in that case.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_snippets_library.baseline
chrisdunelm: Changed to have the numerical suffix last

src/main/java/com/google/api/codegen/transformer/csharp/CSharpSurfaceNamer.java
chrisdunelm: Class now removed

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSnippetsTransformer.java
chrisdunelm: Done

chrisdunelm:
PTAL

src/main/java/com/google/api/codegen/viewmodel/SnippetsFileView.java
```diff
@@ -0,0 +1,67 @@
...
+  @Override
+  public abstract String outputPath();
+
+  public abstract List<ImportTypeView> imports();
+
+  public abstract String packageName();
+
+  public abstract String name();
+
+  public abstract String clientTypeName();
```
garrettjonesgoogle: call this `apiClassName` for consistency

chrisdunelm: Removed this, it's no longer used.

chrisdunelm:
Thanks. Will merge on green

PR #517, opened 2016-09-26
Fix nodejs codegen.

jmuk:
- the code is wrong actually when both callback and options are
  not specified. Add if (options === undefined) ... clause because
  of this.
- the grpc stub name is wrong for bundled method.
- remove global timeout parameter.

### DIFF
--- a/src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
+++ b/src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
@@ -89,7 +89,8 @@ public class NodeJSGapicContext extends GapicContext implements NodeJSContext {
   }
 
   private String getStubNameFor(Interface service, Method method) {
-    String jsMethodName = upperCamelToLowerCamel(method.getSimpleName());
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
```diff
@@ -88,6 +88,19 @@ public ApiMethodView getApiMethodView(Interface service, Method method) {
     return grpcStubTransformer.generateGrpcStubs(context);
   }
 
+  private String getStubNameFor(Interface service, Method method) {
+    String jsMethodName = upperCamelToLowerCamel(method.getSimpleName());
```
geigerj: Optional: this seems strange -- does this duplicate existing logic to transform a `Method` instance into its NodeJS name? (Since clearly this transformation has already happened elsewhere in order for the method name to be attached to the `GrpcStubView`...)

jmuk: Done

PR #519, opened 2016-09-27
Made the getByteLengthFunction use MVVM pattern in a more MVVM way. 

Landrito:
Transformation methods and objects should only be accessed by the transformers. The surfaces should only be able to handle view objects.


src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
```diff
@@ -498,7 +484,7 @@ public String getByteLengthFunction(Interface service, Method method, TypeRef ty
     switch (typeRef.getKind()) {
       case TYPE_MESSAGE:
         return "gax.createByteLengthFunction(grpcClients."
-            + getStubNameFor(service, method)
+            + getApiMethodView(service, method).stubName()
```
jmuk: This was a typo -- it turned out to be `grpcClientVariableName()`. Are there good name for that?

Landrito:
Let me look into this later tonight. I might make a separate view for this
since its pretty inefficient create a whole api method view for a single
name.

src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
Landrito: Fixed.

Landrito:
PTAL. Fixed to use grpcClientVariableName.

src/main/java/com/google/api/codegen/transformer/GrpcStubTransformer.java
```diff
@@ -62,4 +64,23 @@
...
+    stub.fullyQualifiedType(namer.getFullyQualifiedStubType(targetInterface));
+    stub.createStubFunctionName(namer.getCreateStubFunctionName(targetInterface));
+    String grpcClientTypeName = namer.getGrpcClientTypeName(targetInterface);
+    stub.grpcClientTypeName(context.getTypeTable().getAndSaveNicknameFor(grpcClientTypeName));
+    stub.grpcClientVariableName(namer.getGrpcClientVariableName(targetInterface));
+    stub.grpcClientImportName(namer.getGrpcClientImportName(targetInterface));
+    stub.methodNames(Collections.singletonList(namer.getApiMethodName(method)));
+    stub.stubMethodsArrayName(namer.getStubMethodsArrayName(targetInterface));
+    stub.namespace(namer.getNamespace(targetInterface));
+    stub.protoFileName(targetInterface.getFile().getSimpleName());
```
garrettjonesgoogle: This is duplicating the code in the function above.

garrettjonesgoogle:
LGTM

src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
```diff
@@ -46,6 +46,9 @@
 import com.google.common.collect.ImmutableSet;
 import com.google.protobuf.DescriptorProtos.FieldDescriptorProto.Type;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
```
jmuk: nit: strip away those imports? seems not used in the new code.

Landrito: Just pushed an amendment. Good catch.

PR #520, opened 2016-09-27
Update naming of grpc fields.

jmuk:
Updates googleapis/gax-nodejs#43


jmuk:
@jmdobry

jmuk:
@geigerj, please review this too.

src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
```diff
@@ -250,7 +250,7 @@ private String fieldParamComment(Field field, String paramComment, boolean isOpt
    */
   public List<String> fieldPropertyComment(Field field) {
     String commentType = fieldTypeCardinalityComment(field);
-    String fieldName = wrapIfKeywordOrBuiltIn(field.getSimpleName());
+    String fieldName = lowerUnderscoreToLowerCamel(field.getSimpleName());
```
geigerj: Don't we still want to check it it's a keyword?

src/main/resources/com/google/api/codegen/nodejs/main.snip
```diff
@@ -393,13 +393,13 @@
       }
       var req = {
         @join field : requiredParams on ",".add(BREAK)
-          {@field.getSimpleName}: {@fieldParameterName(field)}
+          {@context.lowerUnderscoreToLowerCamel(field.getSimpleName)}: {@fieldParameterName(field)}
```
geigerj: Let's add a helper method in the context instead of duplicating this.

src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
jmuk: unnecessary -- actually existing `wrapIf...` isn't necessary, because this is a field type. In JS, something like `{if: 'foo'}` is syntactically valid.

src/main/resources/com/google/api/codegen/nodejs/main.snip
jmuk: Done

geigerj:
Please rebase and confirm that build passes before merging.

PR #521, opened 2016-09-27
Go merges metadata instead of overwriting

### LINKED ISSUE #474: Don't overwrite user's metadata
Currently, we do this at the beginning of every method:
```
ctx = metadata.NewContext(ctx, c.metadata)
```
where `c.metadata` contains the version of the client library.

However, this will nuke any metadata that already exists in the context. We should be adding to it instead.

cc @jba @garye

pongad:
Fixes #474.


jba:
Couldn't metadata.Join(nil, md) be defined to return md? Then this code would be a one-liner (since metadata.FromContext should return nil with false).

vchudnov-g:
If we can make this a one-liner like jba suggests, that would be ideal.

pongad:
@vchudnov-g @jba 
When I was writing `metadata.Join`, I thought it would be least confusing if it always creates a new map. I mainly wanted to avoid cases like
```
ar := []metadata.MD{foo, bar}
md := metadata.Join(ar...)
md["x"] = []string{"y"} // changes neither foo nor bar

ar := []metadata.MD{foo}
md := metadata.Join(ar...)
md["x"] = []string{"y"} // changes foo
```

So, if we do something like
```
md, _ := metadata.FromContext(ctx)
ctx = metadata.NewContext(md, c.metadata)
```
(not quite one-liner, but close), we'd be copying `c.metadata` each time even if we don't need to.

The performance difference should most likely not matter, but it sounds like something we should keep in mind.

vchudnov-g:
@pongad 

1. But the calling code you currently have in this PR already has the discrepancy you mention: if you retrieve `md`, then you join by making a copy and operations on the result do not affect `c.metadata`, whereas if you don't retrieve `md`, you don't join and so can affect `c.metadata` directly
2. Given that, `Join` with one argument could just return the argument and you'd be behaving the same way at this callsite. Or you could indeed always copy to have consistent semantics.

pongad:
@vchudnov-g Hmm! I was only considering that no one can change `c.metadata` since it's not exported. I never considered that someone else might change it. We are only passing these contexts to grpc clients though, and they shouldn't be modifying it.

That said, I think copying makes sense and reduces potential confusion. It sounds like a decision we can reverse anyway.

pongad:
@vchudnov-g @jba PTAL

vchudnov-g:
LGTM

PR #522, opened 2016-09-27
Go: Make sure clientNamePrefix doesn't break on new Interfaces

pongad:
Go has a logic of dropping service names when they are not necessary.
Eg, instead of "library.NewLibraryClient",
we just say "library.NewClient".

Previously we drop the service name in two situations
- When the service is the only one defined in the package
- When the name of the service is the same as the name of the package

This commit removes the first condition,
as it might break client libraries when new services are added
to the package later on.


src/main/java/com/google/api/codegen/transformer/go/GoSurfaceNamer.java
```diff
@@ -160,12 +161,12 @@ public String getExamplePackageName() {
...
 
-  private Name clientNamePrefix(Interface service) {
+  @VisibleForTesting
+  Name clientNamePrefix(Interface service) {
     Name name = getReducedServiceName(service);
-    // If there's only one service, or the service name matches the package name, don't prefix with
-    // the service name.
-    if (model.getSymbolTable().getInterfaces().size() == 1
-        || Name.from(getLocalPackageName()).equals(name)) {
+    // If the service name matches the package name, don't prefix with the service name.
```
vchudnov-g: "don't include the service name in the prefix"

src/test/java/com/google/api/codegen/transformer/go/GoSurfaceNamerTest.java
```diff
@@ -0,0 +1,64 @@
...
+      throw new IllegalStateException(model.getDiagCollector().getDiags().toString());
+    }
+  }
+
+  @Test
+  public void testClientNamePrefix() {
+    GoSurfaceNamer namer = new GoSurfaceNamer(model, apiConfig.getPackageName());
+    List<Interface> services = model.getSymbolTable().getInterfaces().asList();
+    Truth.assertThat(namer.clientNamePrefix(services.get(0))).isEqualTo(Name.from("my"));
+    Truth.assertThat(namer.clientNamePrefix(services.get(1))).isEqualTo(Name.from("my", "proto"));
```
vchudnov-g: Is this testing when there are multiple services? How about also testing when there's only one service, since that's what motivated the PR?

src/main/java/com/google/api/codegen/transformer/go/GoSurfaceNamer.java

src/test/java/com/google/api/codegen/transformer/go/GoSurfaceNamerTest.java
pongad: I'm a silly person. Done.

pongad:
@vchudnov-g PTAL

src/test/java/com/google/api/codegen/transformer/go/GoSurfaceNamerTest.java
```diff
@@ -33,32 +33,52 @@
...
 
-    apiConfig = ApiConfig.createApiConfig(model, configProto);
+    ApiConfig apiConfig = ApiConfig.createApiConfig(model, configProto);
 
     if (model.getDiagCollector().hasErrors()) {
       throw new IllegalStateException(model.getDiagCollector().getDiags().toString());
     }
+    GoSurfaceNamer namer = new GoSurfaceNamer(apiConfig.getPackageName());
+    List<Interface> services = model.getSymbolTable().getInterfaces().asList();
+    Truth.assertThat(namer.clientNamePrefix(services.get(0))).isEqualTo(Name.from());
```
vchudnov-g: Because this test depends on several files, I suggest making sure we document clearly what's going on. I think it would be good to 

* assert or check explicitly that getPackageName() returns "gopher" (purely for the benefit of humans reading the code; this check isn't the point of the test)
* and then comment above this line that since the first service name in the proto is "gopher" as well, the prefix should be empty

pongad: Done.

PR #528, opened 2016-09-28
Go: Fix iterator

pongad:
This commit fixes a few iterator issues.
- There was duplicate iterators if more than one method paginates over
  the same resource types. Iterators are now deduped.
- The iterator over strings mistakenly tried to return nil.
  Zero values now work properly.
- Iterators declared return type "err" instead of "error".
  This is now fixed.


pongad:
@jba Thank you (and the Go team in general) for raising the bar.

PR #529, opened 2016-09-28
Fix unused import in Go discovery sample gen

tcoffee-google:
Change to template means "fmt" is no longer used when response is
empty.


saicheems:
lgtm

vchudnov-g:
LGTM

PR #530, opened 2016-09-28
Remove unreliable print dependency in C# discovery sample gen

tcoffee-google:
It appears some versions of .NET do not include System.Web.Script
namespace by default:
http://stackoverflow.com/questions/13035192/error-cs0234-the-type-or-namespace-name-script-does-not-exist-in-the-namespac?noredirect=1

### DIFF
--- a/src/main/resources/com/google/api/codegen/csharp/discovery_fragment.snip
+++ b/src/main/resources/com/google/api/codegen/csharp/discovery_fragment.snip
@@ -219,16 +219,10 @@
   @end
 @end
 
-# Unlike placeholders for other languages, this placeholder does not pretty-print output.
-# This seems to require a third-party library like Json.NET: https://www.nuget.org/packages/newtonsoft.json/
 @private print(element)
-    @let JavaScriptSerializer = context.using("System.Web.Script.Serialization.JavaScriptSerializer")
-        Console.WriteLine(new {@JavaScriptSerializer}().Serialize({@element}));
-    @end
+    Console.WriteLine({@element});
 @end
 
 @private printEntry(entry)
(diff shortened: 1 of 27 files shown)


saicheems:
lgtm

src/main/resources/com/google/api/codegen/csharp/discovery_fragment.snip
```diff
@@ -219,16 +219,10 @@
...
-        Console.WriteLine(new {@JavaScriptSerializer}().Serialize({@element}));
-    @end
+    Console.WriteLine({@element});
 @end
 
 @private printEntry(entry)
-    @let JavaScriptSerializer = context.using("System.Web.Script.Serialization.JavaScriptSerializer")
-        Console.WriteLine("[" + {@entry}.Key + "] = " + new {@JavaScriptSerializer}().Serialize({@entry}.Value));
-    @end
+    Console.WriteLine("[" + {@entry}.Key + "] = " + {@entry}.Value);
```
vchudnov-g: Do we not have a baseline that exercises this codepath? If we don't, could you please start and share a document with baseline features we need to test and include "map response" in it? That way when we do the synthetic (manual or autogen) API/baseline, we make sure this is among the cases covered.

vchudnov-g:
LGTM

src/main/resources/com/google/api/codegen/csharp/discovery_fragment.snip
tcoffee-google: The Compute baseline exercises this, but it was too big to show up in the file listing on Github.

Issue #494, opened 2016-09-21
DiscoveryImporter investigation items


saicheems:
DiscoveryImporter doesn't parse the method parameters of storagetransfer.getGoogleServiceAccount

Should have 1 parameter: `projectId`
Actually has none.

tcoffee-google:
Current workaround is override in src/test/java/com/google/api/codegen/testdata/discoveries/storagetransfer.v1.json.overrides ... can be removed if resolved.

saicheems:
Also investigate why `DiscoveryImporter` produces `DatasetList.Datasets` instead of `Dataset` for the page streaming resource of the `bigquery.datasets.list` method.

saicheems:
(And is there a rule we can find that allows us to produce DatasetList.Datasets in Java/C# and Dataset in other languages)

saicheems:
Obsolete, guess it was an out of date library issue.

PR #531, opened 2016-09-28
Support placeholder strings in samplegen

saicheems:


saicheems:
I've included one baseline to see the changes

saicheems:
Rebased, PTAL

saicheems:
I've changed it to not generate placeholders if the pattern is empty or null, which keeps path template patterns, but removes stuff like `String zone = "{MY-ZONE}";` Michael was ok with this.

the sample comments `eg. us-central-1` will come in a follow up

src/main/java/com/google/api/codegen/discovery/config/FieldInfo.java
```diff
@@ -38,6 +38,12 @@
   public abstract TypeInfo type();
 
   /**
+   * Returns the placeholder value of the field, and empty string if none.
```
vchudnov-g: s/and/or/

src/test/java/com/google/api/codegen/testdata/discoveries/storagetransfer.v1.json.overrides
```diff
@@ -13,6 +13,7 @@
               "isArray": false,
               "isMessage": false
             },
+            "placeholder": "",
```
vchudnov-g: So this is because we would generate the wrong placeholder otherwise, until #494 is fixed?

saicheems: Well it's really here because this field isn't populated by the `DiscoveryImporter`, and so it's entire definition is specified in the overrides. We have to include it because it's a non-nullable field and the `SampleConfig` can't be built unless it's specified.

saicheems:
There is always a TODO after each field:
`String zone = ""; // Update placeholder value(s).`

vchudnov-g:
@saicheems: WIth the word `TODO` in them I assume? OK.

Please respond to the other two comments, but otherwise LGTM. Thanks!

saicheems:
Oh, yeah it's `String zone = ""; // TODO: Update placeholder value(s).

PR #532, opened 2016-09-28
Add MVVM baselines for Java

saicheems:


saicheems:
Note: These will be deleted once we switch flows.

vchudnov-g:
We currently have tests that run samplegen on the checked-in Discoveries and compare the output against the (previous) baselines, correct?

Could you please add similar tests that exercise the MVVM samplegen on the checked-in Discoveries and compare the output against these new baselines? Hopefully it's as easy as calling a common test harness with two new parameters: which languages to use MVVM for, and where the MVVM baselines are located.

saicheems:
Reduced to 3 files

vchudnov-g:
OK, talked offline. We will skip adding separate tests for these baselines. When we switch over to Java MVVM, we will overwrite the current Java baselines with the MVVM java baselines, and the test will exercise that. If we need to revert, reverting will revert both the codepath and the baselines, so test would go back to the old baselines.

@saicheems tells me that locally he's been overwriting the codepath and the baselines with MVVM versions to ensure things pass the compilecheck, and they do.

vchudnov-g:
LGTM

PR #533, opened 2016-09-28
Change Python enum structure

geigerj:
Use a new nested structure for Python enums. This ensures that adding
new enum types to an API is not a breaking change for the client
library. Previously, a new type could cause an old type to be
disambiguated, which was a breaking change.

Note: the new generated `enum.py` contains some inconsistent, spurious whitespace (in particular, 3 newlines between some top-level declarations instead of 2). I couldn't figure out a way to get rid of the extra newline without running into a template language bug that causes inconsistent indentation. Since the extra newlines are purely of stylistic concern and will probably be removed during the post-generation formatting task in `artman` anyway, I decided to proceed as-is.


src/test/java/com/google/api/codegen/testdata/ruby_doc_main_library.baseline
```diff
@@ -810,6 +810,7 @@ module Library
       # @example
       #   require "library/v1/library_service_api"
       #
+      #   Alignment = Google::Example::Library::V1::SomeMessage2::SomeMessage3::Alignment
```
bjwatson: @Landrito @swcloud Not a comment on this PR per se, but this could be fragile if multiple types with the same simple name (e.g. `Alignment`) are needed, correct?

Landrito: Yep. Some form of disambiguation will probably be needed. I don't think this is breaking though. The RubyTypeTable in MVVM accounts for scenarios where [shortnames](https://github.com/googleapis/toolkit/blob/master/src/main/java/com/google/api/codegen/util/ruby/RubyTypeTable.java#L87) are already used.

bjwatson: Right, I don't think future changes to improve samples would be considered breaking

geigerj:
@garrettjonesgoogle Do you mind taking a look at the updated Java baselines and making sure that the output is expected given the change to `library.proto`?

bjwatson:
LGTM, but I think @shinfan should review the Java and PHP baseline updates, and @swcloud or @Landrito the Ruby & NodeJS updates. Specifically, is gRPC generating the enum classes for these other four languages, and we're only having to do it explicitly in the GAPIC layer for Python?

shinfan:
Java and PHP baseline changes LGTM

bjwatson:
@jmuk sez Ruby and NodeJS changes look good

bjwatson:
@geigerj LGTM

bjwatson:
@geigerj Can you create a P3 issue to fix the whitespace problems later?

geigerj:
Done as https://github.com/googleapis/toolkit/issues/535

bjwatson:
Thanks Jacob!

PR #534, opened 2016-09-28
Java gen: Fixing Codacy warnings, fixing acronym handling

garrettjonesgoogle:


src/test/java/com/google/api/codegen/util/NameTest.java
```diff
@@ -139,4 +139,22 @@ public void ambiguousAcronym() {
   public void ambiguousRepeatedAcronym() {
     System.out.println(Name.upperCamel("APIDogAPIAMName").toLowerUnderscore());
   }
+
+  @Test
+  public void upperCamelUpperAcronymsSeparate() {
+    Name name = Name.upperCamelKeepUpperAcronyms("IAM", "HTTP", "XML", "Dog");
+    Truth.assertThat(name.toLowerUnderscore()).isEqualTo("iam_http_xml_dog");
+    Truth.assertThat(name.toUpperUnderscore()).isEqualTo("IAM_HTTP_XML_DOG");
+    Truth.assertThat(name.toLowerCamel()).isEqualTo("iamHTTPXMLDog");
```
michaelbausor: Is this definitely what we want when calling toLowerCamel on a string that starts with an acronym?

src/main/java/com/google/api/codegen/util/CommonAcronyms.java
```diff
@@ -15,44 +15,165 @@
...
 public class CommonAcronyms {
-  private static final ImmutableMap<String, String> ACRONYMS =
-      ImmutableMap.<String, String>builder()
-          .put("IAM", "Iam")
-          .put("HTTP", "Http")
-          .put("XML", "Xml")
-          .put("API", "Api")
-          .build();
+  private static final ImmutableSet<String> ACRONYMS =
+      ImmutableSet.<String>builder().add("IAM").add("HTTP").add("XML").add("API").build();
```
michaelbausor: Removing the map means we can't have an arbitrary capitalization scheme for a given acronym. Will we regret that later? I can't think of any cases now

src/main/java/com/google/api/codegen/LanguageUtil.java
```diff
@@ -37,17 +37,17 @@ public static String getTODO() {
   }
 
   public static String upperCamelToUpperUnderscore(String name) {
-    name = CommonAcronyms.replaceAcronyms(name);
+    name = CommonAcronyms.camelizeUpperAcronyms(name);
```
garrettjonesgoogle: LOL

src/test/java/com/google/api/codegen/util/NameTest.java
garrettjonesgoogle: It's the only reasonable thing I can think of; `iAM` looks really horrible.

src/main/java/com/google/api/codegen/util/CommonAcronyms.java
garrettjonesgoogle: I don't think so - the point of my implementation here is to actually vary the capitalization of the output. If we need to special-case something like that, it will have to be handled when we get there.

garrettjonesgoogle:
I needed to make one small tweak - I want to use upper acronyms for mock grpc classes, so that the naming correlates better.

PR #536, opened 2016-09-28
Add temporary hack to resolve Go import path in google/api

pongad:
Types in protobuf package google.api is laid out differently in Go.
The proto files defining these types will need `go_package` option
so that we can generate imports properly.

This commit introduce a hack so that we can generate services while
waiting for this option to be added.


vchudnov-g:
LGTM

PR #537, opened 2016-09-28
Go: Iterators appends to buffer

pongad:
Previously the "fetch" function on  iterators overwrite the buffer.
This is incorrect since fetch might need to be called more than once to
fill the buffer up to a certain page size.

This commit makes fetch appends to the buffer instead.


vchudnov-g:
LGTM

PR #539, opened 2016-09-29
Add media upload placeholder and misc. style changes

saicheems:
- Add patch specific comment to the request body object.
- Revert comments above page streaming resources in loops.

### DIFF
--- a/src/main/java/com/google/api/codegen/discovery/transformer/java/JavaSampleMethodToViewTransformer.java
+++ b/src/main/java/com/google/api/codegen/discovery/transformer/java/JavaSampleMethodToViewTransformer.java
@@ -88,6 +88,7 @@ public class JavaSampleMethodToViewTransformer implements SampleMethodToViewTran
     sampleBodyView.serviceVarName(symbolTable.getNewSymbol(sampleNamer.getServiceVarName()));
     sampleBodyView.serviceTypeName(
         sampleTypeTable.getAndSaveNicknameForServiceType(sampleConfig.apiTypeName()));
+    sampleBodyView.methodVerb(methodInfo.verb());
     sampleBodyView.methodNameComponents(methodInfo.nameComponents());
     sampleBodyView.requestVarName(symbolTable.getNewSymbol(sampleNamer.getRequestVarName()));
     sampleBodyView.requestTypeName(
@@ -97,6 +98,7 @@ public class JavaSampleMethodToViewTransformer implements SampleMethodToViewTran
     sampleBodyView.requestBodyVarName("");
     sampleBodyView.requestBodyTypeName("");
     sampleBodyView.resourceGetterName("");
+    sampleBodyView.resourceVarName("");
     sampleBodyView.resourceTypeName("");
     sampleBodyView.isResourceMap(false);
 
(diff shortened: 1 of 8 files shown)


src/main/java/com/google/api/codegen/discovery/transformer/java/JavaSampleMethodToViewTransformer.java
```diff
@@ -107,8 +109,11 @@ public SampleBodyView createSampleBody(SampleTransformerContext context) {
             "method is page streaming, but the page streaming resource field is null.");
       }
       sampleBodyView.resourceGetterName(sampleNamer.getResourceGetterName(fieldInfo.name()));
-      sampleBodyView.resourceTypeName(
-          sampleTypeTable.getAndSaveNickNameForElementType(fieldInfo.type()));
+      String resourceTypeName = sampleTypeTable.getAndSaveNickNameForElementType(fieldInfo.type());
+      String resourceVarName =
+          sampleNamer.getResourceVarName(fieldInfo.type().isMessage() ? resourceTypeName : "");
+      sampleBodyView.resourceVarName(symbolTable.getNewSymbol(resourceVarName));
+      sampleBodyView.resourceTypeName(resourceTypeName);
```
vchudnov-g: nit: since this only depends on `resourceTypeName`, move to just after the `String resourceTypeName = ...` line.

src/test/java/com/google/api/codegen/testdata/discoveries/mvvm_java/java_adexchangebuyer.v1.4.json.baseline
```diff
@@ -1650,6 +1652,7 @@ import java.security.GeneralSecurityException;
 public class AdExchangeBuyerExample {
   public void run() throws IOException, GeneralSecurityException {
     // TODO: Assign values to the desired fields of `requestBody`.
+    // Only the assigned fields will be changed.
```
vchudnov-g: Just curious: do you know whether every time that the HTTP verb is `PATCH`, the client library method is also `.patch(..)`? Looking at these baselines, it seems to be, which is good.

saicheems: I'm not sure, but I figure if the patch verb is used then the request should follow patch semantics, and we should be right regardless.
The alternative is that the method is `.patch()` but the verb is not patch... which sounds like an API design error.

src/main/java/com/google/api/codegen/discovery/transformer/java/JavaSampleMethodToViewTransformer.java
saicheems: done

src/test/java/com/google/api/codegen/testdata/discoveries/mvvm_java/java_adexchangebuyer.v1.4.json.baseline
vchudnov-g: I was more worried about HTTP `patch` being matched to a non-`patch` library method, where the comment might be puzzling.

OK.

vchudnov-g:
Once #542 lands, you should create an analogous PR for MVVM Java, but don't block this one on that. 

(For future languages, obviously you can do both together.)

src/main/resources/com/google/api/codegen/java/sample.snip
```diff
@@ -39,6 +39,9 @@
   public void run() throws IOException, GeneralSecurityException {
     @if sampleBody.hasInputRequest
       // TODO: Assign values to the desired fields of `{@sampleBody.requestBodyVarName}`.
+      @if sampleBody.methodVerb == "PATCH"
```
tcoffee-google: Change comments to headline style for consistency with others ("Assign values to desired fields", "Only assigned fields will be changed").

src/test/java/com/google/api/codegen/testdata/discoveries/mvvm_java/java_adexchangebuyer.v1.4.json.baseline
tcoffee-google: Sai is correct for all APIs seen thus far.

src/main/resources/com/google/api/codegen/java/sample.snip
saicheems: Changed to
```
TODO: Assign values to desired fields of `{@sampleBody.requestBodyVarName}`.
// Only assigned fields will be changed
```

PR #538, opened 2016-09-29
Update media upload for Node.js discovery sample gen

tcoffee-google:
Initialize fields required by Node.js client libraries for media
upload, also needed to pass compile check.

PR #542, opened 2016-09-29
Add media download placeholders in discovery sample gen

tcoffee-google:
Should be later replaced by more elaborate code illustrating media
download funcionality in client libraries.

Changes to Java baselines other than Cloud Storage due to minor
formatting correction.

### DIFF
--- a/src/main/java/com/google/api/codegen/DiscoveryContext.java
+++ b/src/main/java/com/google/api/codegen/DiscoveryContext.java
@@ -240,6 +240,12 @@ public abstract class DiscoveryContext extends CodegenContext {
   }
 
   public boolean hasMediaDownload(Method method) {
+    // ignore media download for methods supporting media upload, as Apiary cannot combine both in
+    // single request, and no sensible use cases are known for download with a method supporting
+    // upload
+    if (hasMediaUpload(method)) {
+      return false;
+    }
     return apiaryConfig.getMediaDownload().contains(method.getName());
   }
 
--- a/src/main/java/com/google/api/codegen/DiscoveryContext.java
+++ b/src/main/java/com/google/api/codegen/DiscoveryContext.java
@@ -239,6 +239,10 @@ public abstract class DiscoveryContext extends CodegenContext {
     return apiaryConfig.getMediaUpload().contains(method.getName());
   }
 
+  public boolean hasMediaDownload(Method method) {
+    return apiaryConfig.getMediaDownload().contains(method.getName());
+  }
+
   public boolean hasAuthScopes(Method method) {
     return apiaryConfig.hasAuthScopes(method.getName());
   }
--- a/src/main/resources/com/google/api/codegen/java/discovery_fragment.snip
+++ b/src/main/resources/com/google/api/codegen/java/discovery_fragment.snip
@@ -66,11 +66,18 @@
         JsonFactory jsonFactory = JacksonFactory.getDefaultInstance();
 
         {@ApiName} {@serviceName()} = new {@ApiName}.Builder(httpTransport, jsonFactory, credential)
-          .setApplicationName("Google Cloud Platform Sample")
-          .build();
+            .setApplicationName("Google Cloud Platform Sample")
+            .build();
 
         {@initialize(method)}
         {@methodCall(method)}
+        @if context.hasMediaDownload(method)
+
+          // {@TODO()} Download desired media content. For more information, see:
+          // https://developers.google.com/api-client-library/java/google-api-java-client/media-download
+
+
+        @end
         @if context.isPageStreaming(method)
           {@iterableCall(method)}
         @else
--- a/src/main/resources/com/google/api/codegen/java/discovery_fragment.snip
+++ b/src/main/resources/com/google/api/codegen/java/discovery_fragment.snip
@@ -73,7 +73,7 @@
         {@methodCall(method)}
         @if context.hasMediaDownload(method)
 
-          // {@TODO()} Download desired media content. For more information, see:
+          // {@TODO()} Download media content if desired. For more information, see:
           // https://developers.google.com/api-client-library/java/google-api-java-client/media-download
 
 
--- a/src/main/resources/com/google/api/codegen/php/discovery_fragment.snip
+++ b/src/main/resources/com/google/api/codegen/php/discovery_fragment.snip
@@ -148,7 +148,6 @@
 
       {@mediaDownload()}
 
-
     @else
       @if context.hasMediaUpload(method)
         {@mediaUpload()}
@@ -211,7 +210,7 @@
 # TODO(tcoffee): refer to documentation if extant and valid on option for HTTP download with URL:
 # http://stackoverflow.com/questions/17855705/downloading-files-using-google-api-php-client
 @private mediaDownload()
-  // {@TODO()} Use `'alt' => 'media'` to download media content.
+  // {@TODO()} To download media content, use:
   //
   // ${@optParamsName()}['alt'] = 'media';
 @end
--- a/src/main/resources/com/google/api/codegen/php/discovery_fragment.snip
+++ b/src/main/resources/com/google/api/codegen/php/discovery_fragment.snip
@@ -126,17 +126,36 @@
       @end
 
     @end
-  @end
-  @if context.hasRequestField(method)
+    @if context.hasRequestField(method)
 
-    ${@requestBodyName()} = new {@context.getRequestClassName(method)}($client);
-    @if context.isPatch(method)
-      // {@TODO()} Assign values to desired properties of the `{@requestBodyName()}` object to be changed.
-    @else
-      // {@TODO()} Assign values to desired properties of the `{@requestBodyName()}` object.
+      ${@requestBodyName()} = new {@context.getRequestClassName(method)}($client);
+      @if context.isPatch(method)
+        // {@TODO()} Assign values to desired properties of the `{@requestBodyName()}` object to be changed.
+      @else
+        // {@TODO()} Assign values to desired properties of the `{@requestBodyName()}` object.
+      @end
+
+
     @end
 
+    @if context.hasMediaDownload(method)
+      ${@optParamsName()} = array();
+      @if context.hasMediaUpload(method)
 
+        {@mediaUpload()}
+
+      @end
+
+      {@mediaDownload()}
+
+
+    @else
+      @if context.hasMediaUpload(method)
+        {@mediaUpload()}
+
+
+      @end
+    @end
   @end
 @end
 
@@ -144,9 +163,8 @@
   @let responseTypeRef = context.getApiaryConfig.getType(method.getResponseTypeUrl), \
        resourceField = context.getFirstRepeatedField(responseTypeRef), \
        resources = resourceField.getName
-    ${@optParamsName()} = array();
-    @if context.hasMediaUpload(method)
-      {@mediaUpload()}
+    @if not(context.hasMediaDownload(method))
+      ${@optParamsName()} = array();
 
     @end
     do {
@@ -175,12 +193,6 @@
 @end
 
 @private compactCall(method)
-  @if context.hasMediaUpload(method)
-    {@mediaUpload()}
-
-
-
-  @end
   @if context.isResponseEmpty(method)
     {@methodCall(method)};
   @else
@@ -196,6 +208,14 @@
   // https://developers.google.com/api-client-library/php/guide/media_upload
 @end
 
+# TODO(tcoffee): refer to documentation if extant and valid on option for HTTP download with URL:
+# http://stackoverflow.com/questions/17855705/downloading-files-using-google-api-php-client
+@private mediaDownload()
+  // {@TODO()} Use `'alt' => 'media'` to download media content.
+  //
+  // ${@optParamsName()}['alt'] = 'media';
+@end
+
 @private methodCall(method)
   @let methodName = context.getMethodName(method), \
        resourcesPath = context.getApiaryConfig.getResources(method.getName)
@@ -205,7 +225,7 @@
 
 @private arguments(method)
   @let pathArgs = pathArguments(method)
-    @if context.isPageStreaming(method)
+    @if or(context.isPageStreaming(method), context.hasMediaDownload(method))
       @if pathArgs
         {@pathArgs}, ${@optParamsName()}
       @else
--- a/src/main/resources/com/google/api/codegen/php/discovery_fragment.snip
+++ b/src/main/resources/com/google/api/codegen/php/discovery_fragment.snip
@@ -140,20 +140,14 @@
 
     @if context.hasMediaDownload(method)
       ${@optParamsName()} = array();
-      @if context.hasMediaUpload(method)
-
-        {@mediaUpload()}
-
-      @end
 
       {@mediaDownload()}
 
-    @else
-      @if context.hasMediaUpload(method)
-        {@mediaUpload()}
+    @end
+    @if context.hasMediaUpload(method)
+      {@mediaUpload()}
 
 
-      @end
     @end
   @end
 @end
--- a/src/test/java/com/google/api/codegen/testdata/discoveries/go/go_storage.v1.json.baseline
+++ b/src/test/java/com/google/api/codegen/testdata/discoveries/go/go_storage.v1.json.baseline
@@ -1469,6 +1469,8 @@ func main() {
     // TODO: Add desired fields of the request body.
   }
 
+  // TODO: Use `Download` in place of `Do` below to download media content. For more information, see:
+  // https://godoc.org/google.golang.org/api/storage/v1
   resp, err := storageService.Objects.Compose(destinationBucket, destinationObject, &rb).Context(ctx).Do()
   if err != nil {
     log.Fatal(err)
@@ -1541,6 +1543,8 @@ func main() {
     // TODO: Add desired fields of the request body.
   }
 
+  // TODO: Use `Download` in place of `Do` below to download media content. For more information, see:
+  // https://godoc.org/google.golang.org/api/storage/v1
   resp, err := storageService.Objects.Copy(sourceBucket, sourceObject, destinationBucket, destinationObject, &rb).Context(ctx).Do()
   if err != nil {
     log.Fatal(err)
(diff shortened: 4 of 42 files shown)


src/main/resources/com/google/api/codegen/java/discovery_fragment.snip
```diff
@@ -66,11 +66,18 @@
...
-          .setApplicationName("Google Cloud Platform Sample")
-          .build();
+            .setApplicationName("Google Cloud Platform Sample")
+            .build();
 
         {@initialize(method)}
         {@methodCall(method)}
+        @if context.hasMediaDownload(method)
+
+          // {@TODO()} Download desired media content. For more information, see:
```
saicheems: Should we just say `Download media content. For more information, see:`
I'm not sure if there can be multiple media objects that are returned, or if there's only one
The documentation suggests there's one output stream

tcoffee-google: Downloading media content is not the only use case for such methods, as their generic form is used to download metadata. So the user only need add code to download media content if desired. I'm open to alternative suggestions on wording.

tcoffee-google: Perhaps "Download media content if desired."

saicheems: Ok, sgtm

saicheems: Yeah I like that wording, `Download media content if desired`

src/test/java/com/google/api/codegen/testdata/discoveries/nodejs/nodejs_storage.v1.json.baseline
```diff
@@ -1571,10 +1592,19 @@ google.auth.getApplicationDefault(function(err, authClient) {
       // TODO: Add desired properties of the request body.
     },
 
-
-    // TODO: Add desired media content for upload. For more information, see:
+    // TODO: Use `media` to add desired media content for upload. For more information, see:
```
vchudnov-g: Similarly here. At the very least, I suggest a common structure with upload and download:
eg
```
// TODO: To do foo,.... use:
//   blah blah
// For xxxx more information, see
// http://
```
as above. Or having a different order like you have here. But I think both should have the same order.

src/test/java/com/google/api/codegen/testdata/discoveries/nodejs/nodejs_storage.v1.json.baseline
```diff
@@ -1353,6 +1353,13 @@ google.auth.getApplicationDefault(function(err, authClient) {
       // TODO: Add desired properties of the request body.
     },
 
+    // TODO: Use `alt: 'media'` to download media content.
```
vchudnov-g: For the upload section, you have a URL that seems to be specific to uploads (`https://github.com/google/google-api-nodejs-client#media-uploads`). Is there a similar specific URL for downloads, or does the one below cover it? (Your wording suggests it only covers a subset of download situations).

src/test/java/com/google/api/codegen/testdata/discoveries/php/php_storage.v1.json.baseline
```diff
@@ -1126,7 +1126,14 @@ $requestBody = new Google_Service_Storage_ComposeRequest($client);
 // TODO: Assign values to desired properties of the `requestBody` object.
 
 
-$response = $service->objects->compose($destinationBucket, $destinationObject, $requestBody);
+$optParams = array();
+
+// TODO: Use `'alt' => 'media'` to download media content.
+//
+// $optParams['alt'] = 'media';
```
vchudnov-g: As in the node comments above, if there' s commented -out code the user might want to use, I slightly prefer being more concise by removing the blank line above and indenting the commented-out code.

src/test/java/com/google/api/codegen/testdata/discoveries/php/php_storage.v1.json.baseline
```diff
@@ -1126,7 +1126,14 @@ $requestBody = new Google_Service_Storage_ComposeRequest($client);
...
+
+
```
vchudnov-g: I'd prefer a single blank line here, which is more compact and seems consistent with the spacing between placeholders below.

src/test/java/com/google/api/codegen/testdata/discoveries/ruby/ruby_storage.v1.json.baseline
```diff
@@ -1045,6 +1045,11 @@ request_body = Google::Apis::StorageV1::ComposeRequest.new
 # TODO: Assign values to desired members of the `request_body` object.
 
 
+# TODO: Specify destination to download media content. For more information, see:
```
vchudnov-g: is there commented-out code we can use here showing how we'd specify it?

src/test/java/com/google/api/codegen/testdata/discoveries/go/go_storage.v1.json.baseline
```diff
@@ -1474,6 +1474,8 @@ func main() {
     // TODO: Add desired fields of the request body.
   }
 
+  // TODO: Use `Download` in place of `Do` below to download media content. For more information, see:
+  // https://godoc.org/google.golang.org/api/storage/v1
```
vchudnov-g: Is there a more specific media download URL we could use, or is this the best we can do for Go? The other languages where you specify URLs appear to have it.

Maybe a link to the method, eg `https://godoc.org/google.golang.org/api/storage/v1#ObjectsComposeCall.Download`?

src/main/resources/com/google/api/codegen/java/discovery_fragment.snip
tcoffee-google: Sorry, what did you mean by "non-media content metadata"?

vchudnov-g: You mentioned previously that "Downloading media content is not the only use case for such methods"

src/main/resources/com/google/api/codegen/php/discovery_fragment.snip
```diff
@@ -126,27 +126,45 @@
...
-    ${@requestBodyName()} = new {@context.getRequestClassName(method)}($client);
-    @if context.isPatch(method)
-      // {@TODO()} Assign values to desired properties of the `{@requestBodyName()}` object to be changed.
-    @else
-      // {@TODO()} Assign values to desired properties of the `{@requestBodyName()}` object.
     @end
 
+    @if context.hasMediaDownload(method)
+      ${@optParamsName()} = array();
+      @if context.hasMediaUpload(method)
```
tcoffee-google: Media upload normally passes additional keyword parameters to the method, which is done in a labeled array in PHP. If there is already page-streaming or media download that refers to a provided `optParams` variable, then the user will need to add media upload parameters to this variable; otherwise, the user can just add them to the method signature as an array literal if desired (which is how it appears in the docs).

src/test/java/com/google/api/codegen/testdata/discoveries/go/go_storage.v1.json.baseline
tcoffee-google: Go programmers don't appear to like expository documentation ;-). I thought about a link to the method: are we comfortable assuming these URL anchors will remain stable?

src/test/java/com/google/api/codegen/testdata/discoveries/nodejs/nodejs_storage.v1.json.baseline
tcoffee-google: Wording change seems fine to me, but I suggest keeping my indentation, which is what the code would be at if used. Changed.

tcoffee-google: For Node.js, there is not a corresponding doc for media download that I could find. IIRC, the closest thing is instructions specific to the Google Drive API. The included URL above is specific to CORS, and does not address media download per se.

src/test/java/com/google/api/codegen/testdata/discoveries/nodejs/nodejs_storage.v1.json.baseline
tcoffee-google: Will make this update in https://github.com/googleapis/toolkit/pull/538.

src/test/java/com/google/api/codegen/testdata/discoveries/php/php_storage.v1.json.baseline
tcoffee-google: As in comment above, I dislike this because then the code is not at the proper indentation level if uncommented, and also stands out less from the corresponding comment.

src/test/java/com/google/api/codegen/testdata/discoveries/ruby/ruby_storage.v1.json.baseline
tcoffee-google: It involves adding a parameter to the method signature; without that, it seems misleading to provide a variable declaration.

src/test/java/com/google/api/codegen/testdata/discoveries/php/php_storage.v1.json.baseline
tcoffee-google: Oops --- done.

src/main/resources/com/google/api/codegen/java/discovery_fragment.snip
tcoffee-google: The other use case is downloading metadata.

src/test/java/com/google/api/codegen/testdata/discoveries/nodejs/nodejs_storage.v1.json.baseline
tcoffee-google: On second look, this update no longer makes sense in https://github.com/googleapis/toolkit/pull/538, since the code is not commented out.

src/main/resources/com/google/api/codegen/java/discovery_fragment.snip
vchudnov-g: Right, so my comment is that the `Download media content if desired.` (either wording) does not address that use case. Is there a way to distinguish the two cases? If so, we should tailor the wording. If not, let's file an issue to follow up on this.

src/main/resources/com/google/api/codegen/php/discovery_fragment.snip
vchudnov-g: Ah, ok. Thanks for clarifying!

src/test/java/com/google/api/codegen/testdata/discoveries/nodejs/nodejs_storage.v1.json.baseline
vchudnov-g: OK, thanks for checking.

src/test/java/com/google/api/codegen/testdata/discoveries/nodejs/nodejs_storage.v1.json.baseline
vchudnov-g: Oh, I see. The nodsjs changes in #538 supersede these, correct?

But the structure for both upload and download is now this, correct?:
```
// TODO: blah blah
// For more information, see xxx
someCode = goes.Here
```

src/test/java/com/google/api/codegen/testdata/discoveries/ruby/ruby_storage.v1.json.baseline
vchudnov-g: Two thoughts:
(a) Can we hint at it inline?
```
// service.compose_object(..., mediaDestination=newMediaDestination().setLocation())
```
(b) would it make sense to always show the destination variable, but simply have the value be an empty placeholder in the snippet? (And, if anything, say in the destination variable declaration `TODO: fill in the desired destination. If you don't want to download media content, you may omit this parameter entirely in the function call below`.) Thoughts?

If neither of these are easy or you disagree, I am fine deferring this discussion until later.

src/test/java/com/google/api/codegen/testdata/discoveries/go/go_storage.v1.json.baseline
vchudnov-g: Since this is autogen doc, I would expect them to be deterministic and hence stable as long as the method does not disappear. Moreover, IIRC a non-existent anchor still loads the page, so we'd be no worse off.

@saicheems Thoughts on the link stability?

src/test/java/com/google/api/codegen/testdata/discoveries/nodejs/nodejs_storage.v1.json.baseline
vchudnov-g: See my response below.

src/main/resources/com/google/api/codegen/java/discovery_fragment.snip
tcoffee-google: My general approach has been that the sample comments should clarify the mechanics of the client library, but need not substitute for the API documentation. For instance, https://cloud.google.com/storage/docs/json_api/v1/objects/get#response says:

"By default, this responds with an object resource in the response body. If you provide the URL parameter alt=media, then it will respond with the object data in the response body."

From this I expect the user to understand that there's a switch between getting just metadata vs. getting the object data, and the purpose of the sample is just to indicate where and how that switch gets accomplished in client-library calling code. (Granted, the API docs could be clearer on this point.)

vchudnov-g:
Address remaining comments, including making issues for the deferred topics.

Otherwise, LGTM.

src/test/java/com/google/api/codegen/testdata/discoveries/nodejs/nodejs_storage.v1.json.baseline
tcoffee-google: The changes in #538 address only upload, whose parameters are marked required in client library docs. The download has been changed to your suggested format.

src/test/java/com/google/api/codegen/testdata/discoveries/php/php_storage.v1.json.baseline
tcoffee-google: For at least one other aesthetic issue where Sai and I could not agree, we suggested asking for fresh-look A/B comparisons from other engineers based on full samples once we are converged on the other formatting issues.

src/test/java/com/google/api/codegen/testdata/discoveries/ruby/ruby_storage.v1.json.baseline
tcoffee-google: Two thoughts:
(a) I don't see that this hint adds value over the documentation that we link to currently.
(b) I think a better approach long-term will be to have the sample include two different cases, as many of the hand-written samples for media download already do. But this goes beyond the placeholder triage of present. There's something to be said for having, by default, the most minimal sample that can get something working.

src/test/java/com/google/api/codegen/testdata/discoveries/go/go_storage.v1.json.baseline
saicheems: IMO it's a good idea to do it - GoDoc has a predictable format based on the names of types, and so it should always link correctly as long as we're up to date.

src/main/resources/com/google/api/codegen/java/discovery_fragment.snip
vchudnov-g: OK, I see what you mean. The uncommented code gives you the metadata, the commented code gives you the object data. Correct?

src/test/java/com/google/api/codegen/testdata/discoveries/php/php_storage.v1.json.baseline
vchudnov-g: Yeah, I agree looking on a clean slate would be easier. Please record someplace that this is one of the things we want to get feedback on.

src/test/java/com/google/api/codegen/testdata/discoveries/ruby/ruby_storage.v1.json.baseline
vchudnov-g: I'm of mixed mind on (b), but OK, let's defer discussion.

src/test/java/com/google/api/codegen/testdata/discoveries/nodejs/nodejs_storage.v1.json.baseline
tcoffee-google: On closer look, while CORS is documented as supported with the Javascript client library, it is not with the Node.js client library, so I'm going to leave this part out unless we have some positive indication for it.

src/main/resources/com/google/api/codegen/java/discovery_fragment.snip
tcoffee-google: Correct.

vchudnov-g:
ping

src/test/java/com/google/api/codegen/testdata/discoveries/go/go_storage.v1.json.baseline
tcoffee-google: Done

src/main/java/com/google/api/codegen/DiscoveryContext.java
```diff
@@ -240,6 +240,12 @@ public boolean hasMediaUpload(Method method) {
   }
 
   public boolean hasMediaDownload(Method method) {
+    // ignore media download for methods supporting media upload, as Apiary cannot combine both in
```
vchudnov-g: Capitalize and end with a period.

vchudnov-g: Include a pointer to the external doc stating this:
https://developers.google.com/discovery/v1/using#discovery-doc-methods

src/test/java/com/google/api/codegen/testdata/discoveries/ruby/ruby_storage.v1.json.baseline
tcoffee-google: Updated https://github.com/googleapis/toolkit/issues/10#issuecomment-250868101

src/test/java/com/google/api/codegen/testdata/discoveries/php/php_storage.v1.json.baseline
tcoffee-google: Added https://github.com/googleapis/toolkit/issues/552

PR #543, opened 2016-09-29
Fix iterator names in Go discovery sample gen

tcoffee-google:


src/main/java/com/google/api/codegen/go/GoDiscoveryContext.java
```diff
@@ -58,7 +58,7 @@ protected String mapTypeName(String keyName, String valueName) {
 
   @Override
   protected String objectTypeName(String typeName) {
-    return upperCamelToLowerCamel(typeName);
+    return upperCamelToLowerCamel(getSimpleName(typeName));
```
vchudnov-g: Does this change only affect the iterators? Could this break any other uses?

vchudnov-g:
If you're confident this won't have untoward side effects, LGTM.

tcoffee-google:
This function is only used for generating variable names in Go, even though it's statically typed (this unique situation is probably what led to the bug). The only potential side effect of this fix I can think of in general is introducing a name collision, but the baseline tests confirm this does not occur in Cloud APIs, and IIUC the pending MVVM work will solve that issue in general.

PR #544, opened 2016-09-30
Stage Go GRPC streaming support

pongad:
There are a couple of missing things

- Baseline for client-side streaming
  This will affect baseline files in other languages.
  I'll do this in a separate PR to avoid create a large one.
- "Iterator-like" server-side streaming
  This feature is discussed with the Go team back in July.
  If the streaming method is used to supply the client with a bounded
  list of values, it can be treated as an iterator that uses streaming,
  and toolkit should provide an iterator with .Next() method instead
  of GRPC's Recv().
  It's an open question how this feature should interact with
  "grpc-streaming with page-streaming".
  If each response has an array of objects, should we "flatten" the
  stream? If so, what do we do with other fields in the response? etc.
  I will leave this feature unimplemented for now, until we have good
  answers to these questions.


src/main/java/com/google/api/codegen/transformer/FeatureConfig.java
```diff
@@ -36,4 +36,9 @@ public boolean useResourceNameFormatOption(Field field) {
   public boolean enableMixins() {
     return false;
   }
+
+  /** Returns true if mixin APIs are supported. */
+  public boolean enableStreaming() {
```
garrettjonesgoogle: Fix the comment

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -443,6 +443,14 @@ public String getAsyncApiMethodExampleName(Method method) {
     return getNotImplementedString("SurfaceNamer.getAsyncApiMethodExampleName");
   }
 
+  public String getGrpcStreamingApiMethodName(Method method) {
+    return getApiMethodName(method);
+  }
+
+  public String getGrpcStreamingApiMethodExampleName(Interface interfaze, Method method) {
+    return getGrpcStreamingApiMethodName(method);
+  }
```
garrettjonesgoogle: Add docs to the new methods here

src/main/java/com/google/api/codegen/transformer/SurfaceTransformerContext.java
```diff
@@ -100,7 +100,10 @@ public MethodTransformerContext asMethodContext(Method method) {
    * TODO: integrate with GapicContext for this method.
    */
   private boolean isSupported(Method method) {
-    return !method.getResponseStreaming() && !method.getRequestStreaming();
+    if (!getFeatureConfig().enableStreaming()) {
+      return !method.getResponseStreaming() && !method.getRequestStreaming();
+    }
+    return true;
```
garrettjonesgoogle: You could cut out one negation by starting with `if(getFeatureConfig().enableStreaming()) {return true;}`

src/main/java/com/google/api/codegen/transformer/go/GoSurfaceNamer.java
```diff
@@ -213,4 +213,14 @@ public String getStubName(Interface service) {
...
   }
+
+  @Override
+  public String getStaticLangStreamingReturnTypeName(Method method, MethodConfig methodConfig) {
+    // Unsafe string manipulation: The name looks like "LibraryService_StreamShelvesClient",
+    // neither camel or underscore.
+    return String.format(
+        "%s_%sClient",
+        converter.getTypeName(method.getParent()).getNickname(),
+        method.getSimpleName());
```
garrettjonesgoogle: I think you should compose the %sClient part using the Name class, and then do the remaining composition with plain concatenation.

Side question: why does the return type name look like it is a client type name?

src/main/java/com/google/api/codegen/viewmodel/StaticLangXCombinedSurfaceView.java
```diff
@@ -73,6 +73,10 @@ public String resourceRoot() {
 
   public abstract List<PageStreamingDescriptorClassView> pageStreamingDescriptorClasses();
 
+  public boolean getTrue() {
+    return true;
+  }
```
garrettjonesgoogle: this looks very weird

src/main/resources/com/google/api/codegen/go/example.snip
```diff
@@ -22,24 +22,110 @@
...
-        @else
-            @if method.hasReturnValue
-                {@simpleMethod(view, method)}
-            @else
-                {@emptyReturnMethod(view, method)}
-            @end
+        @case method.hasReturnValue
+            {@simpleMethod(view, method)}
+        @default
+            {@emptyReturnMethod(view, method)}
```
garrettjonesgoogle: Please use an enum instead of this very odd-looking pattern.

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
```diff
@@ -255,6 +255,8 @@ private void setCommonFields(
     methodViewBuilder.stubName(namer.getStubName(context.getTargetInterface()));
     methodViewBuilder.settingsGetterName(namer.getSettingsFunctionName(context.getMethod()));
     methodViewBuilder.callableName(context.getNamer().getCallableName(context.getMethod()));
+    methodViewBuilder.isRequestStreaming(context.getMethod().getRequestStreaming());
+    methodViewBuilder.isResponseStreaming(context.getMethod().getResponseStreaming());
```
garrettjonesgoogle: If you use an enum for the streaming type, you wouldn't need to add each boolean as a separate attribute.

src/main/resources/com/google/api/codegen/go/example.snip
pongad: I thought it looked odd but found it strangely familiar, then I realized: https://tour.golang.org/flowcontrol/11

src/main/java/com/google/api/codegen/transformer/FeatureConfig.java
pongad: Fixed by https://github.com/googleapis/toolkit/pull/566

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
pongad: Done

src/main/java/com/google/api/codegen/transformer/SurfaceTransformerContext.java
pongad: Done

src/main/java/com/google/api/codegen/transformer/go/GoSurfaceNamer.java
pongad: Done.

In Go, streaming is done with two stream objects, one sending and one receiving (if bidirectional, then both can send *and* receive). One of these objects is owned by the server and another is owned by the client. So the word "client" here doesn't mean "client library" but "the client half of the streaming object pair".

src/main/java/com/google/api/codegen/viewmodel/StaticLangXCombinedSurfaceView.java
pongad: Done

pongad:
@garrettjonesgoogle PTAL

jba:
This all looks OK to me, except that if we follow the iterator pattern for some streaming RPCs then some of these methods will change.

On the other hand, maybe using the iterator pattern for just one kind of streaming doesn't make sense, since there will be an inconsistency between bidi streams and receive-only streams.

So LGTM, but we should discuss streams vs. iterators a bit more.

src/main/resources/com/google/api/codegen/go/example.snip
```diff
@@ -22,13 +22,22 @@
     }
 
     @join method : view.apiMethods
-        @if method.isPageStreaming
-            {@pageStreamingMethod(view, method)}
-        @else
-            @if method.hasReturnValue
-                {@simpleMethod(view, method)}
+        @switch method.grpcStreamingType.toString
```
garrettjonesgoogle: you don't need the `.toString` part (which I didn't originally realize) (same in main.snip)

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -540,6 +550,10 @@ public String getStaticLangAsyncReturnTypeName(Method method, MethodConfig metho
     return getNotImplementedString("SurfaceNamer.getStaticLangAsyncReturnTypeName");
   }
 
+  public String getStaticLangStreamingReturnTypeName(Method method, MethodConfig methodConfig) {
```
garrettjonesgoogle: one more method needing a doc

pongad: Sorry I missed this. Done

src/main/resources/com/google/api/codegen/go/example.snip
pongad: Good point, done

pongad:
I'll hold off merging this until https://github.com/googleapis/api-client-staging/pull/75 lands.

That PR is itself waiting for #575 and #576 to land.

PR #545, opened 2016-09-30
Additional style changes for Java discovery sample

saicheems:
Misc. style changes:
- Move request body declaration below the field list.
- Add logic to determine plurality of the placeholder TODO.


saicheems:
To explain the slightly large CL size:

I've included two new baselines diffed against the ones on master:
`java_compute.v1.json.baseline` contains `Map<K, V>` page streaming resources.
`java_storage.v1.json.baseline` contains media upload boilerplate.

Please point out any remaining issues you'd like me to fix, and once that's done I think we're good to update all baselines and merge.

src/main/java/com/google/api/codegen/discovery/config/ApiaryConfigToSampleConfigConverter.java
```diff
@@ -141,18 +143,23 @@ private MethodInfo createMethod(Method method) {
...
+    boolean isPlaceholderSingular = true;
     TypeInfo typeInfo = createTypeInfo(field, method);
     if (typeInfo.kind() == Field.Kind.TYPE_STRING && !typeInfo.isArray() && !typeInfo.isMap()) {
       String pattern = apiaryConfig.getFieldPattern().get(containerType.getName(), field.getName());
       if (!Strings.isNullOrEmpty(pattern)) {
-        placeholder = DefaultString.getPlaceholder(field.getName(), pattern);
+        placeholder = DefaultString.getNonTrivialPlaceholder(pattern);
+        // String placeholders are always contained within braces, so counting
+        // the number of open braces is an easy test for singularity.
+        isPlaceholderSingular = StringUtils.countMatches(placeholder, '{') < 2;
```
vchudnov-g: Could there be  a case where the count could be zero, and we print a TODO when there's nothing to replace? Or would this never happen because there wouldn't be a pattern in that case?

saicheems: Can't happen as far as I know... there shouldn't be a pattern where there's nothing to replace, at least I haven't seen such a field.

Also, the DefaultString algorithm only operates on a specific pattern type, so if there's nothing to replace in the pattern it does no work

So the scenario you describe should never happen

saicheems:
If the changes look good, I'll rebase to only merge the style changes and follow up with a merge CL.

vchudnov-g:
LGTM

PR #547, opened 2016-09-30
Additional style changes for Java discovery samplegen

saicheems:
- Qualify service var name.
- Revert sample title.


tcoffee-google:
LGTM for now.

I had meant to ask @vchudnov-g if we still want to use "Google Cloud Platform Sample" when we move beyond Cloud APIs.

PR #551, opened 2016-09-30
Add missing import in C# discovery sample gen

tcoffee-google:


vchudnov-g:
OK, just ran compilecheck and this seems to work.

LGTM.

PR #555, opened 2016-10-01
Manually update Ruby names file for compile check

tcoffee-google:
Based on latest discovery docs, also updated across all test APIs.


tcoffee-google:
Hang on, left out commit updating baselines based on changed discoveries.

tcoffee-google:
Should now be consistent.

vchudnov-g:
Did  you verify that Ruby compilecheck passes? If so, LGTM.

tcoffee-google:
As discussed offline, submitting despite current compile-check failures so we can update status.

Issue #550, opened 2016-09-30
Don't generate identifiers that are reserved keywords


vchudnov-g:
When creating the Go samples for the drive API, we generate this gem:

```
// The A1 notation of the values to update.
  range := "{MY-RANGE}"
// TODO: Change placeholder value above.
```

which does not compile because `range` is a keyword. We should ensure none of our generated identifiers are keywords.

Solution
1. Short-term: tcoffee@ please add an exception for this particular case in the Go generator, with a TODO to build a more scalable solution
2. Short-term: saicheems@ port this exception when doing the Go MVVM refactoring
3. saicheems@, tcoffee@: Please design and implement a longer term solution (on the MVVM-based code) where we check all generated identifiers against that language's keywords, and modify them if there's a collision.

saicheems:
The long term solution for this (done for Java in #556) is to seed the symbol table with reserved identifiers.

vchudnov-g:
@saicheems Sweet! That will take care of #2 and #3. @tcoffee-google , we still need to do #1 in the meantime.

tcoffee-google:
#1 should be resolved by https://github.com/googleapis/toolkit/pull/567, but please check since Cloud baselines are unchanged.

tcoffee-google:
Additional cases handled in https://github.com/googleapis/toolkit/pull/588.

PR #556, opened 2016-10-01
Even more style changes to Java discovery samplegen

saicheems:
- Remove the main function.
- Rename createService to create{apiTypeName}service and make it static.
- Add missing nextPageToken assignment.
- Change to two spaces before placeholder TODOs.
- Put back response TODO.
- Fully qualify request types.
- Use double asterisk in instructions doc comment.
- Be consistent in auth link TODO formatting.
- Seed symbol table with reserved identifiers.
- Add override option for where the page streaming resource setter is in the
  request body instead of the request.
- Override isPageStreaming in sql.users.list, the Java client doesn't have the
  setPageToken function on the SQLAdmin.Users.List request.

### DIFF
--- a/src/main/java/com/google/api/codegen/discovery/config/SampleConfig.java
+++ b/src/main/java/com/google/api/codegen/discovery/config/SampleConfig.java
@@ -88,17 +88,6 @@ public abstract class SampleConfig {
   public abstract String apiTypeName();
 
   /**
-   * Returns the lower-camel formatted API type name.
-   *
-   * This is mostly for utility for overrides where apiTypeName is used to
-   * generate a variable name, and the automatically generated lower-camel
-   * format is wrong.
-   * For example: "SQLAdmin" to "sQLAdmin"
-   */
-  @JsonProperty("lowerCamelApiTypeName")
-  public abstract String lowerCamelApiTypeName();
-
-  /**
    * Returns the language specific package prefix for API types.
    */
   @JsonProperty("packagePrefix")
@@ -140,9 +129,6 @@ public abstract class SampleConfig {
 
     @JsonProperty("apiTypeName")
     public abstract Builder apiTypeName(String val);
-
-    @JsonProperty("lowerCamelApiTypeName")
-    public abstract Builder lowerCamelApiTypeName(String val);
 
     @JsonProperty("packagePrefix")
     public abstract Builder packagePrefix(String val);
--- a/src/main/java/com/google/api/codegen/discovery/config/SampleConfig.java
+++ b/src/main/java/com/google/api/codegen/discovery/config/SampleConfig.java
@@ -88,6 +88,17 @@ public abstract class SampleConfig {
   public abstract String apiTypeName();
 
   /**
+   * Returns the lower-camel formatted API type name.
+   *
+   * This is mostly for utility for overrides where apiTypeName is used to
+   * generate a variable name, and the automatically generated lower-camel
+   * format is wrong.
+   * For example: "SQLAdmin" to "sQLAdmin"
+   */
+  @JsonProperty("lowerCamelApiTypeName")
+  public abstract String lowerCamelApiTypeName();
+
+  /**
    * Returns the language specific package prefix for API types.
    */
   @JsonProperty("packagePrefix")
@@ -129,6 +140,9 @@ public abstract class SampleConfig {
 
     @JsonProperty("apiTypeName")
     public abstract Builder apiTypeName(String val);
+
+    @JsonProperty("lowerCamelApiTypeName")
+    public abstract Builder lowerCamelApiTypeName(String val);
 
     @JsonProperty("packagePrefix")
     public abstract Builder packagePrefix(String val);
--- a/src/main/java/com/google/api/codegen/discovery/config/java/JavaTypeNameGenerator.java
+++ b/src/main/java/com/google/api/codegen/discovery/config/java/JavaTypeNameGenerator.java
@@ -17,10 +17,11 @@ package com.google.api.codegen.discovery.config.java;
 import java.util.ArrayList;
 import java.util.List;
 
+import com.google.api.client.util.Strings;
+import com.google.api.codegen.discovery.DefaultString;
 import com.google.api.codegen.discovery.config.TypeNameGenerator;
 import com.google.api.codegen.util.Name;
 import com.google.common.base.Joiner;
-import com.google.protobuf.Field;
 
 public class JavaTypeNameGenerator implements TypeNameGenerator {
 
@@ -61,12 +62,23 @@ public class JavaTypeNameGenerator implements TypeNameGenerator {
   }
 
   @Override
-  public String formatValue(String value, Field.Kind kind) {
-    switch (kind) {
-      case TYPE_STRING:
-        return "\"" + value + "\"";
-      default:
-        throw new IllegalArgumentException("unsupported kind: " + kind.toString());
+  public String getExample(String format) {
+    if (Strings.isNullOrEmpty(format)) {
+      return "";
     }
+    switch (format) {
+      case "byte":
+        return "Base64-encoded string of bytes: see http://tools.ietf.org/html/rfc4648";
+      case "date":
+        return "\"YYYY-MM-DD\": see java.text.SimpleDateFormat";
+      case "date-time":
+        return "\"YYYY-MM-DDThh:mm:ss.fffZ\": see com.google.api.client.util.DateTime.toStringRfc3339()";
+    }
+    // Returns "" if format is invalid.
+    String def = DefaultString.getNonTrivialPlaceholder(format);
+    if (Strings.isNullOrEmpty(def)) {
+      return "";
+    }
+    return String.format("ex: \"%s\"", def);
   }
 }
--- a/src/main/java/com/google/api/codegen/discovery/config/java/JavaTypeNameGenerator.java
+++ b/src/main/java/com/google/api/codegen/discovery/config/java/JavaTypeNameGenerator.java
@@ -62,7 +62,7 @@ public class JavaTypeNameGenerator implements TypeNameGenerator {
   }
 
   @Override
-  public String getExample(String format) {
+  public String getStringFormatExample(String format) {
     if (Strings.isNullOrEmpty(format)) {
       return "";
     }
@@ -73,12 +73,17 @@ public class JavaTypeNameGenerator implements TypeNameGenerator {
         return "\"YYYY-MM-DD\": see java.text.SimpleDateFormat";
       case "date-time":
         return "\"YYYY-MM-DDThh:mm:ss.fffZ\": see com.google.api.client.util.DateTime.toStringRfc3339()";
+      default:
+        return "";
     }
-    // Returns "" if format is invalid.
-    String def = DefaultString.getNonTrivialPlaceholder(format);
+  }
+
+  @Override
+  public String getFieldPatternExample(String pattern) {
+    String def = DefaultString.getNonTrivialPlaceholder(pattern);
     if (Strings.isNullOrEmpty(def)) {
       return "";
     }
-    return String.format("ex: \"%s\"", def);
+    return String.format("\"%s\"", def);
   }
 }
--- a/src/main/java/com/google/api/codegen/discovery/transformer/SampleNamer.java
+++ b/src/main/java/com/google/api/codegen/discovery/transformer/SampleNamer.java
@@ -38,9 +38,8 @@ public class SampleNamer extends NameFormatterDelegator {
   /**
    * Returns the variable name of the service.
    */
-  public String getServiceVarName(String lowerCamelApiTypeName) {
-    return localVarName(
-        Name.lowerCamel(Name.lowerCamel(lowerCamelApiTypeName).toLowerCamel(), "service"));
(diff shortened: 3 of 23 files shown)


src/main/java/com/google/api/codegen/discovery/transformer/SampleNamer.java
```diff
@@ -38,8 +38,9 @@ public String getSampleClassName(String apiTypeName) {
   /**
    * Returns the variable name of the service.
    */
-  public String getServiceVarName(String apiTypeName) {
-    return localVarName(Name.lowerCamel(Name.upperCamel(apiTypeName).toLowerCamel(), "service"));
+  public String getServiceVarName(String lowerCamelApiTypeName) {
+    return localVarName(
+        Name.lowerCamel(Name.lowerCamel(lowerCamelApiTypeName).toLowerCamel(), "service"));
```
garrettjonesgoogle: There is a simpler way to do this:

1. Don't bother with the whole lowerCamelApiTypeName in SampleConfig
2. Use this code:

`return localVarName(Name.upperCamel(apiTypeName, "Service"));`

Basically, each of the static creation methods of Name just need to have their arguments all be in the same casing, so you would just pass the upper camel string "Service" here.

src/main/java/com/google/api/codegen/discovery/transformer/SampleNamer.java
```diff
@@ -88,4 +89,12 @@ public String getRequestBodyVarName() {
   public String getResponseVarName() {
     return localVarName(Name.lowerCamel("response"));
   }
+
+  /**
+   * Returns the name of the createService function.
+   */
+  public String createServiceFuncName(String apiTypeName) {
+    return publicMethodName(
+        Name.lowerCamel("create", Name.upperCamel(apiTypeName).toLowerCamel(), "service"));
```
garrettjonesgoogle: Simpler suggestion:

`return publicMethodName(Name.upperCamel("Create", apiTypeName, "Service"));`

src/main/java/com/google/api/codegen/discovery/transformer/SampleNamer.java
saicheems: The reason I included lowerCamelApiTypeName was for cases where the capitalizations were wrong:

`SQLAdmin` -> `sQLAdmin`

I don't think that that can be fixed without some manual override, but this line can be shortened still, so I'll use the code you provided. Thanks!

src/test/java/com/google/api/codegen/testdata/discoveries/mvvm_java/java_adexchangebuyer.v1.4.json.baseline
```diff
@@ -16,29 +16,30 @@ import com.google.api.client.http.HttpTransport;
...
   }
 
-  public AdExchangeBuyer createService() throws IOException, GeneralSecurityException {
+  public static AdExchangeBuyer createAdExchangeBuyerService() throws IOException, GeneralSecurityException {
     HttpTransport httpTransport = GoogleNetHttpTransport.newTrustedTransport();
     JsonFactory jsonFactory = JacksonFactory.getDefaultInstance();
 
-    // TODO: Change placeholder below to generate authentication credentials.
-    // See: https://foo.com/bar
+    // TODO: Change placeholder below to generate authentication credentials, see:
```
tcoffee-google: Need either a semicolon (`credentials; see:`) or a period (`credentials. See:`). Since there's already a colon after the `TODO`, I prefer the latter.

src/main/java/com/google/api/codegen/discovery/transformer/SampleNamer.java
garrettjonesgoogle: Add "SQL" to the acronym list on CommonAcronyms.

src/main/resources/com/google/api/codegen/java/sample.snip
```diff
@@ -14,136 +14,133 @@
...
+      @if body.methodVerb == "PATCH"
         // Only assigned fields will be changed.
       @end
-      {@sampleBody.requestBodyTypeName} {@sampleBody.requestBodyVarName} = new {@sampleBody.requestBodyTypeName}();
+      {@body.requestBodyTypeName} {@body.requestBodyVarName} = new {@body.requestBodyTypeName}();
 
     @end
-    @if sampleBody.hasMediaUpload
+    @if body.hasMediaUpload
       // TODO: Add desired media content for upload. For more information, see:
```
tcoffee-google: To match less verbose style elsewhere, may want to remove `For more information`.

src/main/resources/com/google/api/codegen/java/sample.snip
```diff
@@ -14,136 +14,133 @@
...
       // https://developers.google.com/api-client-library/java/google-api-java-client/media-upload
 
     @end
-    {@sampleBody.serviceTypeName} {@sampleBody.serviceVarName} = createService();
-    {@sampleBody.requestTypeName} {@sampleBody.requestVarName} = {@sampleBody.serviceVarName}.{@callChain(sampleBody.methodNameComponents)}({@paramList(sampleBody.fieldVarNames)});
+    {@body.serviceTypeName} {@body.serviceVarName} = {@body.createServiceFuncName}();
+    {@body.requestTypeName} {@body.requestVarName} = {@body.serviceVarName}.{@callChain(body.methodNameComponents)}({@paramList(body.fieldVarNames)});
+    @if {@body.hasMediaDownload()}
+
+      // TODO: Download media content if desired. For more information, see:
```
tcoffee-google: As above.

src/main/java/com/google/api/codegen/discovery/transformer/SampleNamer.java
saicheems: Oh I didn't know that existed. Sweet, that saves a lot of trouble

src/main/java/com/google/api/codegen/util/java/JavaTypeTable.java
```diff
@@ -174,4 +175,62 @@ private boolean isImplicitImport(String name) {
     implicitImports.put(name, yes);
     return yes;
   }
+
+  public static final ImmutableSet<String> RESERVED_IDENTIFIER_SET =
```
tcoffee-google: Why was this not already there for GAPIC? (Just curious.)

src/main/java/com/google/api/codegen/discovery/config/SampleConfig.java
```diff
@@ -88,6 +88,17 @@
   public abstract String apiTypeName();
 
   /**
+   * Returns the lower-camel formatted API type name.
+   *
+   * This is mostly for utility for overrides where apiTypeName is used to
```
vchudnov-g: `"...mostly to allow for overrides when..."`

src/main/java/com/google/api/codegen/util/SymbolTable.java
```diff
@@ -30,6 +30,23 @@
   private final Set<String> symbolTable = new HashSet<>();
 
   /**
+   * Returns a new SymbolTable seeded with all the words in seed.
+   *
+   * For example, if seed is {"int"}, a subsequent call to {@link
+   * #getNewSymbol(String)} for "int" will return "int2".
```
vchudnov-g: Something to think about: it might be nice to be able to pass in a hint parameter for when renaming is needed. For example, this could be the resource name. Thus, `int` -> `resourceInt`. Or, more realistically for Go (prompted by #550), `range` -> `listRange`. These would read better and be less confusing than `int2` and `range2`.

I'm OK deferring this, but if so, please file a tracking issue.

src/main/java/com/google/api/codegen/discovery/transformer/SampleNamer.java
saicheems: done

src/main/java/com/google/api/codegen/discovery/transformer/SampleNamer.java
saicheems: done

src/test/java/com/google/api/codegen/testdata/discoveries/mvvm_java/java_adexchangebuyer.v1.4.json.baseline
saicheems: Used period

src/main/resources/com/google/api/codegen/java/sample.snip
saicheems: Done

src/main/resources/com/google/api/codegen/java/sample.snip
saicheems: Done

src/main/java/com/google/api/codegen/util/java/JavaTypeTable.java
saicheems: I'm not sure, I guess they don't have the same problems with conflicts?
@garrettjonesgoogle

src/main/java/com/google/api/codegen/discovery/config/SampleConfig.java
saicheems: Fixed by deletion - Garrett suggested a much cleaner fix to this problem that doesn't require a function in `SampleConfig`

src/main/java/com/google/api/codegen/util/java/JavaTypeTable.java
garrettjonesgoogle: It is very possible for this to happen; we just haven't encountered it yet.

src/main/java/com/google/api/codegen/util/SymbolTable.java
saicheems: #564

src/main/java/com/google/api/codegen/discovery/config/java/JavaTypeNameGenerator.java
```diff
@@ -61,12 +62,23 @@ public String getSubpackage(boolean isRequest) {
   }
 
   @Override
-  public String formatValue(String value, Field.Kind kind) {
-    switch (kind) {
-      case TYPE_STRING:
-        return "\"" + value + "\"";
-      default:
-        throw new IllegalArgumentException("unsupported kind: " + kind.toString());
+  public String getExample(String format) {
```
saicheems: @tcoffee-google 

This generates an example or explanatory comment:
```
// Lorem ipsum dolor...
bytes myField = '';
// Base64-encoded string of bytes...
```

```
// Lorem ipsum dolor...
String myLog = '';
// ex: "projects/my-project/log/my-log"
```

tcoffee-google: Seems to be in the spirit of MVVM that only the content of these special-case comments should be in language-specific code ... would you agree?

src/test/java/com/google/api/codegen/testdata/discoveries/mvvm_java/java_logging.v2beta1.json.baseline
```diff
@@ -429,7 +433,8 @@ public class LoggingExample {
   public static void main(String args[]) throws IOException, GeneralSecurityException {
     // Required. The resource name of the project containing the metrics. Example:
     // `"projects/my-project-id"`.
-    String projectName = "projects/{MY-PROJECT}";  // TODO: Update placeholder value.
+    String projectName = "";  // TODO: Update placeholder value.
+    // ex: "projects/my-project"
```
tcoffee-google: Cases like this do look a bit odd, but I'm inclined not to worry about it unless you have any immediately good ideas.

src/main/java/com/google/api/codegen/discovery/config/java/JavaTypeNameGenerator.java
saicheems: Right, this function is implemented in each language-specific `TypeNameGenerator`, is that what you mean?

Once we get to Node.js, this whole function will have to be reimplemented under `NodeJSTypeNameGenerator`, this class is Java specific.

src/test/java/com/google/api/codegen/testdata/discoveries/mvvm_java/java_logging.v2beta1.json.baseline
saicheems: I'm fine with it - in logging we might just want to override the examples to be empty, seeing as they're part of the description.

src/main/java/com/google/api/codegen/discovery/config/java/JavaTypeNameGenerator.java
saicheems: Got it, I've split this into two functions:
`getStringFormatExample` and `getFieldPatternExample`
Those functions only return the language specific string/description with the appropriate quotes.

`"ex: "` is prefixed in the `createFieldInfo` function in `ApiaryConfigToSampleConfigConverter` if there's a field pattern, otherwise there is no prefix.

tcoffee-google:
LGTM

vchudnov-g:
If compilecheck passes, LGTM.

saicheems:
Compilecheck indeed passes, woohoo!

PR #559, opened 2016-10-03
Moving config classes to own java package

garrettjonesgoogle:
Simultaneously moving code currently under config to configgen.

### DIFF
--- a/src/main/java/com/google/api/codegen/config/BundlingConfig.java
+++ b/src/main/java/com/google/api/codegen/config/BundlingConfig.java
@@ -12,8 +12,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.google.api.codegen;
+package com.google.api.codegen.config;
 
+import com.google.api.codegen.ConfigProto;
+import com.google.api.codegen.BundlingConfigProto;
+import com.google.api.codegen.BundlingDescriptorProto;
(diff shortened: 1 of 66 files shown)


src/main/java/com/google/api/codegen/config/BundlingConfig.java
```diff
@@ -12,8 +12,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.google.api.codegen;
+package com.google.api.codegen.config;
 
+import com.google.api.codegen.ConfigProto;
```
michaelbausor: Should ConfigProto, BundlingConfigProto, etc. also be moved into codegen.config package?

garrettjonesgoogle: I want to do that, but it's a breaking change for all GAPIC config out there, so I'd prefer to defer until another day.

michaelbausor: Fair enough, maybe create an issue for this?

michaelbausor:
LGTM

src/main/java/com/google/api/codegen/config/BundlingConfig.java
garrettjonesgoogle: https://github.com/googleapis/toolkit/issues/563

PR #565, opened 2016-10-03
Upgrading to google-java-format 1.1

garrettjonesgoogle:
Also reformatting everything.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -73,7 +73,7 @@ ext {
     jacksonGuava: 'com.fasterxml.jackson.datatype:jackson-datatype-guava:2.7.0',
 
     // Formatter
-    javaFomatter: 'com.google.googlejavaformat:google-java-format:0.1-alpha'
+    javaFomatter: 'com.google.googlejavaformat:google-java-format:1.1'
   ]
 }
 
@@ -193,7 +193,7 @@ task setupProtocEnvironment << {
 test.dependsOn setupProtocEnvironment
 
 task showRuntimeClassPath << {
-  println sourceSets.main.runtimeClasspath.asPath
+  println 'output: ' + sourceSets.main.runtimeClasspath.asPath
 }
 
 // Test Logging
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -14,7 +14,7 @@ buildscript {
   }
   dependencies {
     classpath "com.google.protobuf:protobuf-gradle-plugin:0.7.3"
-    classpath "gradle.plugin.com.github.sherter.google-java-format:google-java-format-gradle-plugin:0.2"
+    classpath "gradle.plugin.com.github.sherter.google-java-format:google-java-format-gradle-plugin:0.3.2"
```
pongad: I don't think this actually changes the version of the formatter we use to format our code.

We use formatter in two places. One to format and verify Java code in toolkit itself and another to format the java files we render. If we look at the [build log](https://travis-ci.org/googleapis/toolkit/jobs/164732499#L661), we can see that the format verification is still done with version 1.0. I have no idea why the formatting in the comment changed. It's possible that the gradle plugin adds a new option to the formatter.

To change the version of the formatter we use to format toolkit, I think we need to change [this line](https://github.com/garrettjonesgoogle/toolkit/blob/7a75b5aa164d18026de8acc08ebf4366b0598ff5/build.gradle#L446)

build.gradle
```diff
@@ -73,7 +73,7 @@ ext {
     jacksonGuava: 'com.fasterxml.jackson.datatype:jackson-datatype-guava:2.7.0',
 
     // Formatter
-    javaFomatter: 'com.google.googlejavaformat:google-java-format:0.1-alpha'
+    javaFomatter: 'com.google.googlejavaformat:google-java-format:1.1'
```
pongad: I think this only changes the formatter we format our rendered code with.

build.gradle
garrettjonesgoogle: Ahh, I didn't realize the two were separate.

garrettjonesgoogle:
The main thing I wanted to update was the version that I format the GAPIC output with. I have undone the  change to the version of google-java-format-gradle-plugin and consequently all of the changes to toolkit code, so now this is a pretty trivial PR.

garrettjonesgoogle:
(Note - I recognize that I didn't follow my own advice on the PR reset here, but it was going to be more difficult than it was worth to reverse all of the formatting changes)

garrettjonesgoogle:
(PTAL)

pongad:
@garrettjonesgoogle Do we want to change the version of the formatter we format toolkit with eventually?

garrettjonesgoogle:
@pongad yes, but I found that the gradle plugin doesn't work with the latest google-java-format (a class has been removed that the plugin references). We'll need to wait until the plugin is updated.

PR #566, opened 2016-10-03
Implement gRPC streaming support for Java

shinfan:
- Implement surface changes corresponding to the GAX PR: https://github.com/googleapis/gax-java/pull/113
- Add Java code sample support for gRPC streaming
- Separate PageStreamingConfig with GrpcStreamingConfig
- Update baseline tests

### DIFF
--- a/src/main/java/com/google/api/codegen/config/GrpcStreamingConfig.java
+++ b/src/main/java/com/google/api/codegen/config/GrpcStreamingConfig.java
@@ -1,0 +1,108 @@
+/* Copyright 2016 Google Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen.config;
+
+import com.google.api.codegen.PageStreamingConfigProto;
+import com.google.api.tools.framework.model.Diag;
+import com.google.api.tools.framework.model.DiagCollector;
+import com.google.api.tools.framework.model.Field;
+import com.google.api.tools.framework.model.Method;
+import com.google.api.tools.framework.model.SimpleLocation;
+
+import javax.annotation.Nullable;
+
+/** GrpcStreamingConfig represents the gRPC streaming configuration for a method. */
+public class GrpcStreamingConfig {
+
+  /** Grpc streaming types */
+  public enum StreamingType {
+    NonStreaming,
+    ClientStreaming,
+    ServerStreaming,
+    BidiStreaming
+  }
+
+  private final Field resourcesField;
+  private final StreamingType type;
+
+  /**
+   * Creates an instance of GrpcStreamingConfig for gRPC response streaming, based on
+   * PageStreamingConfigProto, linking it up with the provided method. On errors, null will be
+   * returned, and diagnostics are reported to the diag collector.
+   */
+  @Nullable
+  public static GrpcStreamingConfig createGrpcStreaming(
+      DiagCollector diagCollector, PageStreamingConfigProto pageStreaming, Method method) {
+    String resourcesFieldName = pageStreaming.getResponse().getResourcesField();
+    Field resourcesField = method.getOutputType().getMessageType().lookupField(resourcesFieldName);
+    StreamingType type = getStreamingType(diagCollector, method);
+    if (type == null) {
+      return null;
+    }
+    return new GrpcStreamingConfig(resourcesField, type);
+  }
+
+  /**
+   * Creates an instance of GrpcStreamingConfig for gRPC response streaming from the given Grpc
+   * method. On errors, null will be returned, and diagnostics are reported to the diag collector.
+   */
+  @Nullable
+  public static GrpcStreamingConfig createGrpcStreaming(
+      DiagCollector diagCollector, Method method) {
+    StreamingType type = getStreamingType(diagCollector, method);
+    if (type == null) {
+      return null;
+    }
+    return new GrpcStreamingConfig(null, type);
+  }
+
+  private static StreamingType getStreamingType(DiagCollector diagCollector, Method method) {
+    StreamingType type = null;
+    if (method.getRequestStreaming() && method.getResponseStreaming()) {
+      type = StreamingType.BidiStreaming;
+    } else if (method.getResponseStreaming()) {
+      type = StreamingType.ServerStreaming;
+    } else if (method.getRequestStreaming()) {
+      type = StreamingType.ClientStreaming;
+    } else {
+      diagCollector.addDiag(
+          Diag.error(
+              SimpleLocation.TOPLEVEL,
+              "The given grpc method does not support streaming: %s",
+              method.getFullName()));
+    }
+    return type;
+  }
+
+  private GrpcStreamingConfig(Field resourcesField, StreamingType type) {
+    this.resourcesField = resourcesField;
+    this.type = type;
+  }
+
+  /** Returns true if the resource field is set. */
+  public boolean hasResourceField() {
+    return resourcesField != null;
+  }
+
+  /** Returns the field used in the response to hold the resource being returned. */
+  public Field getResourcesField() {
+    return resourcesField;
+  }
+
+  /** Returns the streaming type. */
+  public StreamingType getType() {
+    return type;
+  }
+}
--- a/src/main/java/com/google/api/codegen/config/GrpcStreamingConfig.java
+++ b/src/main/java/com/google/api/codegen/config/GrpcStreamingConfig.java
@@ -20,14 +20,13 @@ import com.google.api.tools.framework.model.DiagCollector;
 import com.google.api.tools.framework.model.Field;
 import com.google.api.tools.framework.model.Method;
 import com.google.api.tools.framework.model.SimpleLocation;
-
 import javax.annotation.Nullable;
 
 /** GrpcStreamingConfig represents the gRPC streaming configuration for a method. */
 public class GrpcStreamingConfig {
 
   /** Grpc streaming types */
-  public enum StreamingType {
+  public enum GrpcStreamingType {
     NonStreaming,
     ClientStreaming,
     ServerStreaming,
@@ -35,19 +34,21 @@ public class GrpcStreamingConfig {
   }
 
   private final Field resourcesField;
-  private final StreamingType type;
+  private final GrpcStreamingType type;
 
   /**
    * Creates an instance of GrpcStreamingConfig for gRPC response streaming, based on
    * PageStreamingConfigProto, linking it up with the provided method. On errors, null will be
    * returned, and diagnostics are reported to the diag collector.
+   *
+   * <p>TODO: Create a separate GrpcStreamingConfigProto
    */
   @Nullable
   public static GrpcStreamingConfig createGrpcStreaming(
       DiagCollector diagCollector, PageStreamingConfigProto pageStreaming, Method method) {
     String resourcesFieldName = pageStreaming.getResponse().getResourcesField();
     Field resourcesField = method.getOutputType().getMessageType().lookupField(resourcesFieldName);
-    StreamingType type = getStreamingType(diagCollector, method);
+    GrpcStreamingType type = getGrpcStreamingType(diagCollector, method);
     if (type == null) {
       return null;
     }
@@ -61,21 +62,22 @@ public class GrpcStreamingConfig {
   @Nullable
   public static GrpcStreamingConfig createGrpcStreaming(
       DiagCollector diagCollector, Method method) {
-    StreamingType type = getStreamingType(diagCollector, method);
+    GrpcStreamingType type = getGrpcStreamingType(diagCollector, method);
     if (type == null) {
       return null;
     }
     return new GrpcStreamingConfig(null, type);
   }
 
-  private static StreamingType getStreamingType(DiagCollector diagCollector, Method method) {
-    StreamingType type = null;
+  private static GrpcStreamingType getGrpcStreamingType(
+      DiagCollector diagCollector, Method method) {
+    GrpcStreamingType type = null;
     if (method.getRequestStreaming() && method.getResponseStreaming()) {
-      type = StreamingType.BidiStreaming;
+      type = GrpcStreamingType.BidiStreaming;
     } else if (method.getResponseStreaming()) {
-      type = StreamingType.ServerStreaming;
+      type = GrpcStreamingType.ServerStreaming;
     } else if (method.getRequestStreaming()) {
-      type = StreamingType.ClientStreaming;
+      type = GrpcStreamingType.ClientStreaming;
     } else {
       diagCollector.addDiag(
           Diag.error(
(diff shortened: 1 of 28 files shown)


shinfan:
@jmuk I separate the grpc-streaming config from the page-streaming config since I feel using one class for both purposes can be confusing.

src/main/java/com/google/api/codegen/transformer/SurfaceTransformerContext.java
```diff
@@ -117,6 +107,13 @@ private boolean isSupported(Method method) {
     return methods;
   }
 
+  private boolean isSupported(Method method) {
+    if (!getFeatureConfig().enableStreaming()) {
+      return !method.getRequestStreaming() && !method.getResponseStreaming();
+    }
+    return true;
+  }
```
jmuk: Not sure how worthwhile doing (and/or how does that fit with the design principle of MVVM), but can we just move `isSupported` into the FeatureConfig -- with receiving `Method` as an argument?

src/main/java/com/google/api/codegen/config/GrpcStreamingConfig.java
```diff
@@ -0,0 +1,108 @@
...
+  private final StreamingType type;
+
+  /**
+   * Creates an instance of GrpcStreamingConfig for gRPC response streaming, based on
+   * PageStreamingConfigProto, linking it up with the provided method. On errors, null will be
+   * returned, and diagnostics are reported to the diag collector.
+   */
+  @Nullable
+  public static GrpcStreamingConfig createGrpcStreaming(
+      DiagCollector diagCollector, PageStreamingConfigProto pageStreaming, Method method) {
```
garrettjonesgoogle: Why does grpc streaming dake PageStreamingConfigProto as input?

src/main/java/com/google/api/codegen/config/MethodConfig.java
```diff
@@ -334,10 +329,14 @@ public boolean hasRequestObjectMethod() {
...
-   * Returns the Interface that should be used for the GRPC call in place of the interface
-   * in which this method appears.
+   * Returns the Interface that should be used for the GRPC call in place of the interface in which
+   * this method appears.
    */
   public String getRerouteToGrpcInterface() {
     return rerouteToGrpcInterface;
   }
+
+  static private boolean isStreamingMethod(Method method) {
```
garrettjonesgoogle: why is this package-private?

src/main/java/com/google/api/codegen/transformer/ApiCallableTransformer.java
```diff
@@ -79,7 +80,11 @@ public ApiCallableTransformer() {
     apiCallableBuilder.memberName(context.getNamer().getSettingsMemberName(method));
     apiCallableBuilder.settingsFunctionName(context.getNamer().getSettingsFunctionName(method));
 
-    if (methodConfig.isBundling()) {
+    if (methodConfig.isGrpcStreaming()) {
+      apiCallableBuilder.type(ApiCallableType.StreamingApiCallable);
+      apiCallableBuilder.streamingType(methodConfig.getGrpcStreaming().getType());
+      typeTable.saveNicknameFor("io.grpc.stub.StreamObserver");
```
garrettjonesgoogle: This is Java-specific, so it should either be made more generic (go through SurfaceNamer) or should be moved somewhere else. (Likewise for the same import further down)

src/main/java/com/google/api/codegen/transformer/SurfaceTransformerContext.java
shinfan: By looking up the history, I followed the pattern provided by @chrisdunelm here. I am okay with adding `isSupported` into the feature config but I prefer to make the change in a different PR.

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
```diff
@@ -238,6 +240,12 @@ public StaticLangApiMethodView generateCallableMethod(MethodTransformerContext c
         !ServiceMessages.s_isEmptyType(context.getMethod().getOutputType()));
     methodViewBuilder.isPageStreaming(false);
 
+    MethodConfig methodConfig = context.getMethodConfig();
+    if (methodConfig.isGrpcStreaming()) {
+      GrpcStreamingConfig grpcStreaming = methodConfig.getGrpcStreaming();
+      methodViewBuilder.streamingType(grpcStreaming.getType());
+    }
```
garrettjonesgoogle: Maybe it would be easier to have this if condition elsewhere, so that we could just always set streamingType (and set it to None if there is no streaming)

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTransformer.java
```diff
@@ -249,7 +243,8 @@ private void addXSettingsImports(SurfaceTransformerContext context) {
     ModelTypeTable typeTable = context.getTypeTable();
     typeTable.saveNicknameFor("com.google.api.gax.core.ConnectionSettings");
     typeTable.saveNicknameFor("com.google.api.gax.core.RetrySettings");
-    typeTable.saveNicknameFor("com.google.api.gax.grpc.ApiCallSettings");
+    typeTable.saveNicknameFor("com.google.api.gax.grpc.UnaryApiCallSettings");
+    typeTable.saveNicknameFor("com.google.api.gax.grpc.StreamingCallSettings");
```
garrettjonesgoogle: The StreamingCallSettings won't always be used. The existing pattern is to add the import only in the branches that generate code depending on the type.

src/main/java/com/google/api/codegen/viewmodel/StaticLangXSettingsView.java
```diff
@@ -37,6 +39,16 @@
...
 
+  public List<ApiCallSettingsView> unaryCallSettings() {
+    ArrayList<ApiCallSettingsView> unaryCallSettings = new ArrayList<>();
+    for (ApiCallSettingsView settingsView : callSettings()) {
+      if (!settingsView.isStreaming()) {
+        unaryCallSettings.add(settingsView);
+      }
+    }
+    return unaryCallSettings;
+  }
```
garrettjonesgoogle: Don't put logic like this in view models

src/test/java/com/google/api/codegen/testdata/java_main_library.baseline
```diff
@@ -1643,12 +1655,124 @@ public class LibraryServiceApi implements AutoCloseable {
...
    */
-  public final ApiCallable<UpdateBookIndexRequest, Empty> updateBookIndexCallable() {
+  public final UnaryApiCallable<UpdateBookIndexRequest, Empty> updateBookIndexCallable() {
     return updateBookIndexCallable;
   }
 
   // AUTO-GENERATED DOCUMENTATION AND METHOD
   /**
+   * Test server streaming
+   * gRPC streaming methods don't have HTTP equivalent and need not have google.api.http option.
```
garrettjonesgoogle: -> "gRPC streaming methods don't have an HTTP equivalent and don't need to have the google.api.http option."

src/test/java/com/google/api/codegen/testdata/java_settings_library.baseline
```diff
@@ -263,6 +273,27 @@ public class LibraryServiceSettings extends ServiceApiSettings {
   }
 
   /**
+   * Returns the object with the settings used for calls to streamShelves.
+   */
+  public StreamingCallSettings<StreamShelvesRequest, StreamShelvesResponse> streamShelvesSettings() {
+    return streamShelvesSettings;
+   }
```
garrettjonesgoogle: Why is the indentation of this bracket wrong?

src/test/java/com/google/api/codegen/testdata/java_settings_library.baseline
```diff
@@ -572,6 +606,9 @@ public class LibraryServiceSettings extends ServiceApiSettings {
     private final SimpleCallSettings.Builder<AddCommentsRequest, Empty> addCommentsSettings;
     private final SimpleCallSettings.Builder<GetBookFromArchiveRequest, Book> getBookFromArchiveSettings;
     private final SimpleCallSettings.Builder<UpdateBookIndexRequest, Empty> updateBookIndexSettings;
+    private StreamingCallSettings.Builder<StreamShelvesRequest, StreamShelvesResponse> streamShelvesSettings;
+    private StreamingCallSettings.Builder<StreamBooksRequest, Book> streamBooksSettings;
+    private StreamingCallSettings.Builder<DiscussBookRequest, Comment> discussBookSettings;
```
garrettjonesgoogle: make final

src/test/java/com/google/api/codegen/testdata/java_settings_library.baseline
```diff
@@ -658,7 +701,7 @@ public class LibraryServiceSettings extends ServiceApiSettings {
 
       addLabelSettings = SimpleCallSettings.newBuilder(LabelerGrpc.METHOD_ADD_LABEL);
 
-      methodSettingsBuilders = ImmutableList.<ApiCallSettings.Builder>of(
+      unaryMethodSettingsBuilders = ImmutableList.<ApiCallSettings.Builder>of(
```
garrettjonesgoogle: Why no streamingMethodSettingsBuilders? I see that as a consequence, you don't support applyToAllApiMethods for streaming methods.

src/main/java/com/google/api/codegen/config/GrpcStreamingConfig.java
shinfan: @jmuk This follows Jun's grpc streaming design and implementation of Python. Both page-streaming and grpc-streaming shares the same the config proto for now. But I would prefer to have a separate proto for grpc-streaming. With that said, if we agree that we should use a separate proto, I can add a TODO here and make the change in a separate PR.

src/test/java/com/google/api/codegen/testdata/java_settings_library.baseline
shinfan: Right, applyToAllApiMethods does not support streaming methods since right now we do not have any settings for streaming methods yet.

src/main/java/com/google/api/codegen/config/MethodConfig.java
shinfan: It was only used within this class but I found there are other places can reuse this method. Now I make it public. Done.

src/main/java/com/google/api/codegen/transformer/ApiCallableTransformer.java
shinfan: This is actually not used...removed.

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
shinfan: I am not sure if I understand this comment. I moved this if block into the `MethodConfig` class.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTransformer.java
shinfan: Right, I was thinking we now have the import cleaner so I just did not implement the import logic. The issue is that the branch that generates the code is not language specific. (e.g. StreamingCallSettings are generated in ApiCallableTransformer). 

My fix is to add an extra helper method to determine we need to add such imports.

src/main/java/com/google/api/codegen/viewmodel/StaticLangXSettingsView.java
shinfan: Done.

src/test/java/com/google/api/codegen/testdata/java_main_library.baseline
shinfan: Done.

src/test/java/com/google/api/codegen/testdata/java_settings_library.baseline
shinfan: That's the typo in the snip file. Fixed.

src/test/java/com/google/api/codegen/testdata/java_settings_library.baseline
shinfan: Done

shinfan:
PTAL

src/main/java/com/google/api/codegen/transformer/ApiCallableTransformer.java
```diff
@@ -79,7 +80,10 @@ public ApiCallableTransformer() {
     apiCallableBuilder.memberName(context.getNamer().getSettingsMemberName(method));
     apiCallableBuilder.settingsFunctionName(context.getNamer().getSettingsFunctionName(method));
 
-    if (methodConfig.isBundling()) {
+    if (methodConfig.isGrpcStreaming()) {
+      apiCallableBuilder.type(ApiCallableType.StreamingApiCallable);
+      apiCallableBuilder.streamingType(methodConfig.getGrpcStreaming().getType());
```
garrettjonesgoogle: If this is just gRPC streaming, I'd prefer the view's attribute name to be called `grpcStreamingType` (same for the other places where `streamingType` occurs).

src/main/java/com/google/api/codegen/transformer/FeatureConfig.java
```diff
@@ -36,4 +36,9 @@ public boolean useResourceNameFormatOption(Field field) {
   public boolean enableMixins() {
     return false;
   }
+
+  /** Returns true if streaming APIs are supported. */
+  public boolean enableStreaming() {
```
garrettjonesgoogle: I would prefer "enableGrpcStreaming".

src/test/java/com/google/api/codegen/testdata/java_main_library.baseline
garrettjonesgoogle: Doesn't look done.

src/main/java/com/google/api/codegen/config/GrpcStreamingConfig.java
garrettjonesgoogle: Yeah, add a TODO. I want either of the following:

* Rename the config proto to StreamingConfigProto
* Create a separate GrpcStreamingConfigProto

It's really confusing to have one name specifically for one case but used for both.

src/main/java/com/google/api/codegen/config/MethodConfig.java
garrettjonesgoogle: Ahh, I misread this - you had the `static` before the `private`; it should have been the other order.

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
garrettjonesgoogle: That's what I was going for :-)

src/main/java/com/google/api/codegen/transformer/FeatureConfig.java
shinfan: Done

src/main/java/com/google/api/codegen/transformer/ApiCallableTransformer.java
shinfan: Done

src/test/java/com/google/api/codegen/testdata/java_main_library.baseline
shinfan: Oh sorry somehow I thought it was fixed. Fixed now.

src/main/java/com/google/api/codegen/config/GrpcStreamingConfig.java
shinfan: Picked `Create a separate GrpcStreamingConfigProto`. TODO added.

garrettjonesgoogle:
LGTM (after tests pass)

PR #567, opened 2016-10-03
Work around keyword name collisions in Go discovery sample gen

tcoffee-google:
Temporary fix pending general collision handling in MVVM refactoring.

### DIFF
--- a/src/main/java/com/google/api/codegen/go/GoDiscoveryContext.java
+++ b/src/main/java/com/google/api/codegen/go/GoDiscoveryContext.java
@@ -48,7 +48,7 @@ public class GoDiscoveryContext extends DiscoveryContext implements GoContext {
   }
 
   public boolean isKeyword(String name) {
-    return GoTypeTable.KEYWORD_BUILT_IN_SET.contains(name);
+    return GoTypeTable.RESERVED_IDENTIFIER_SET.contains(name);
   }
 
   @Override
--- a/src/main/java/com/google/api/codegen/go/GoDiscoveryContext.java
+++ b/src/main/java/com/google/api/codegen/go/GoDiscoveryContext.java
@@ -14,6 +14,7 @@
  */
 package com.google.api.codegen.go;
 
+import com.google.api.codegen.util.go.GoTypeTable;
 import com.google.api.codegen.ApiaryConfig;
 import com.google.api.codegen.DiscoveryContext;
 import com.google.api.codegen.DiscoveryImporter;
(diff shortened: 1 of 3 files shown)


tcoffee-google:
@saicheems note this provides the reserved word set for Go we will need for MVVM refactoring.

src/main/java/com/google/api/codegen/util/go/GoTypeTable.java
```diff
@@ -104,4 +105,38 @@ private static boolean isStandardImport(String importPath) {
     // we might have to special case them.
     return !importPath.contains("/");
   }
+
+  /**
+   * A set of Go reserved words. See
+   * https://golang.org/ref/spec#Keywords
+   */
+  public static final ImmutableSet<String> KEYWORD_BUILT_IN_SET =
```
saicheems: Any chance we can call this `RESERVED_KEYWORD_SET`?
I know `KEYWORD_BUILT_IN_SET` is the same naming as in `NodeJSTypeTable`, but I think the above name is a bit more descriptive for this set.

I wonder if it's having another set or including non-reserved identifiers like type names. While we technically don't have to, we could run into some hairy situations like:
```
int := 3
```

saicheems: Or maybe `RESERVED_KEYWORD_SET`

saicheems: Though `RESERVED_IDENTIFIER_SET` is a more accurate name if we do decide to include the names in builtin:

https://golang.org/pkg/builtin/

All of these can be overridden by declaring a var of the same name

saicheems:
The mechanism lgtm

src/main/java/com/google/api/codegen/go/GoDiscoveryContext.java
```diff
@@ -46,6 +47,10 @@ public String getMethodName(Method method) {
     return lowerCamelToUpperCamel(super.getMethodName(method));
   }
 
+  public boolean isKeyword(String name) {
```
saicheems: isReserved may be a better function name

saicheems:
LGTM

vchudnov-g:
I just ran this against the sheets API, and the Go compilecheck passes.

LGTM.

src/main/java/com/google/api/codegen/go/GoDiscoveryContext.java
tcoffee-google: Done

PR #568, opened 2016-10-03
Fix auth scope list bug in Go discovery sample gen

tcoffee-google:
Resolves https://github.com/googleapis/toolkit/issues/558.


saicheems:
Oh, interesting, that's what was going on

LGTM!

PR #570, opened 2016-10-04
Update Java MVVM discovery baselines + minor style fix

saicheems:


tcoffee-google:
LGTM

src/test/java/com/google/api/codegen/testdata/discoveries/mvvm_java/java_logging.v2beta1.json.baseline
```diff
@@ -2,7 +2,60 @@
...
+import com.google.api.services.logging.v2beta1.Logging;
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+import java.util.Arrays;
+
+public class LoggingExample {
+  public static void main(String args[]) throws IOException, GeneralSecurityException {
+    // Required. The resource name of the log to delete.  Example:
+    // `"projects/my-project/logs/syslog"`.
+    String logName = "";  // TODO: Update placeholder value.
```
vchudnov-g: Seems weird to refer to a placeholder when there is none. Maybe it's more appropriate to say "TODO: Fill in value". Sorry it just hit me now!

src/test/java/com/google/api/codegen/testdata/discoveries/mvvm_java/java_logging.v2beta1.json.baseline
```diff
@@ -2,7 +2,60 @@
...
+    // ex: "billingAccounts/my-billingAccount/logs/my-log"
```
vchudnov-g: Let's be consistent with the use (or not) of backquotes and periods here and above. I support the format here: no backquotes, no period.

saicheems: Those aren't controlled by us, that's the description provided by the API owner.

src/test/java/com/google/api/codegen/testdata/discoveries/mvvm_java/java_logging.v2beta1.json.baseline
```diff
@@ -2,7 +2,60 @@
...
+import com.google.api.client.json.JsonFactory;
+import com.google.api.client.json.jackson2.JacksonFactory;
+import com.google.api.services.logging.v2beta1.Logging;
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+import java.util.Arrays;
+
+public class LoggingExample {
+  public static void main(String args[]) throws IOException, GeneralSecurityException {
+    // Required. The resource name of the log to delete.  Example:
```
saicheems: The below version was what Thomas and I decided on... the above is the description provided in the discovery doc. I think the ideal move is probably to override away the below examples for the logging API.

src/test/java/com/google/api/codegen/testdata/discoveries/mvvm_java/java_logging.v2beta1.json.baseline
saicheems: Sure, any objections @tcoffee-google ?

src/test/java/com/google/api/codegen/testdata/discoveries/mvvm_java/java_logging.v2beta1.json.baseline
vchudnov-g: Ideally, we would do this by detecting whether the proto doc contains an example, but that starts getting dicey if the examples are not provided in a consistent format. If we could reliably detect them, we could either delete them from the proto doc we include, or delete our own example. If we do the latter, we should try to be consistent in case parameters further down don't come with examples, and we generate inconsistent ones (I'm thinking of the discrepancy here between `projects/my-project/logs/syslog` and `billingAccounts/my-billingAccount/logs/my-log`)

I'll track this. In the meantime, unless the above is trivial to do (I expect not), having an exception for logging is fine for the purposes of this PR.

vchudnov-g:
I don't see the review changes. Did they get lost?

vchudnov-g:
@saicheems and I talked offline. They way this is coming about is rather roundabout: a single delete request message is being bound to multiple methods, so  for all but one, the sample string they provide is incorrect. From our POV, the correct solution would be to remove the sample string entirely, but then that leaves a delete request message undocumented in the proto, which is probably not good. We could have more specific one-off workarounds, but given that this API will be manually  wrapped, and that we've not seen this in other APIs, let's defer this.

PR #571, opened 2016-10-04
Discovery ruby names update

tcoffee-google:


tcoffee-google:
We're still getting an error in datastore that appears not due to discovery updates. Investigating ...

src/main/resources/com/google/api/codegen/ruby/apiary_names.yaml.old
```diff
@@ -2765,6 +2765,7 @@
 "/cloudresourcemanager:v1beta1/key": key
 "/cloudresourcemanager:v1beta1/quotaUser": quota_user
 "/cloudresourcemanager:v1beta1/cloudresourcemanager.projects.create": create_project
+"/cloudresourcemanager:v1beta1/cloudresourcemanager.projects.create/useLegacyStack": use_legacy_stack
```
vchudnov-g: Why is this file have an `.old` suffix?

src/main/resources/com/google/api/codegen/ruby/apiary_names_update.sh
```diff
@@ -1,17 +0,0 @@
-#!/usr/bin/env bash
```
vchudnov-g: Has this script moved somewhere? Is the process info copied somewhere?

tcoffee-google:
I believe datastore discovery got fixed. Our generator should now be consistent with Ruby client library pending https://github.com/google/google-api-ruby-client/pull/478.

src/main/resources/com/google/api/codegen/ruby/apiary_names_update.sh
tcoffee-google: This script is no longer relevant; it was previously used to generate a Ruby names file from our baseline discovery docs to allow our generator tests to pass. (The process logic came directly from the generate script in the client library.) But we've now copied over the full names file from the updated client library, which covers all the APIs.

src/main/resources/com/google/api/codegen/ruby/apiary_names.yaml.old
tcoffee-google: This was a backup made by the apiary_names_update.sh script for comparison purposes. I'll remove it.

tcoffee-google:
@vchudnov-g PTAL

src/main/java/com/google/api/codegen/DiscoveryImporter.java
```diff
@@ -41,7 +41,7 @@
  * DiscoveryImporter parses a Discovery Doc into a {@link com.google.api.Service} object and an
  * additional {@link ApiaryConfig} object for data not accommodated by Service.
  *
- * The current implementation also provides a main method used for manual sanity checking.
+ * <p>The current implementation also provides a main method used for manual sanity checking.
```
vchudnov-g: We're getting these <p> tags again....

src/main/resources/com/google/api/codegen/ruby/apiary_names.yaml.old
vchudnov-g: OK

src/main/resources/com/google/api/codegen/ruby/apiary_names_update.sh
vchudnov-g: OK

src/main/java/com/google/api/codegen/DiscoveryImporter.java
vchudnov-g: Your editor settings, maybe?

vchudnov-g:
LGTM

src/main/java/com/google/api/codegen/DiscoveryImporter.java
tcoffee-google: I think it happens when I run google-java-format. It suspect it applies changes that are not mandated by the verifier.

PR #572, opened 2016-10-04
Don't support media download if media upload is enabled

saicheems:


tcoffee-google:
Confirmed offline with @saicheems that Cloud Storage baseline will be updated.

LGTM.

Issue #573, opened 2016-10-04
Node: do not name the anonymous function set on the prototype of service objects.


landrito:
### What
Naming the anonymous function set on the prototype of a service object is not idiomatic.

### Example
https://github.com/googleapis/api-client-staging/blob/master/generated/nodejs/google-cloud-node/packages/monitoring/src/v3/metric_service_api.js#L292

### Expected
```js
// Change this...
Service.prototype.methodName = function methodName() {

// into this.
Service.prototype.methodName = function() {
```

landrito:
cc @jmuk

jmuk:
I inherited this design from packman, and I've heard that the benefit of this design is that a stack trace can record a name of the function with the current style while the new style can't.

However, as I tried locally (with node v4.2.3), it seems no change on the stack trace of errors. Let's do this.

landrito:
Interesting! Thanks for checking the stack trace! :)

PR #574, opened 2016-10-04
Node: Idiomatic fixes found when merging monitoring.

landrito:
Removed name from anonymous methods.
Fixed path_template variables.
#573

### DIFF
--- a/src/main/resources/com/google/api/codegen/nodejs/main.snip
+++ b/src/main/resources/com/google/api/codegen/nodejs/main.snip
@@ -310,7 +310,7 @@
        *   A fully-qualified path representing a {@collectionConfig.getEntityName} resources.
        * @@returns {String} - A string representing the {@subField}.
        */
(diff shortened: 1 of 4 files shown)


src/main/resources/com/google/api/codegen/nodejs/main.snip
```diff
@@ -311,7 +311,7 @@
        * @@returns {String} - A string representing the {@subField}.
        */
       {@context.getApiWrapperName(service)}.prototype.match{@subFieldUpper}From{@fieldPath} =
-          function match{@subFieldUpper}From{@fieldPath}({@fieldParam}) {
+          function({@fieldParam}) {
```
jmuk: nit: you can bring this back to the previous line (`...From{@filedPath} = function({@fieldParam}) {`), since the line would be much shorter.

landrito: Done.

PR #576, opened 2016-10-05
Go: Only import request messages in example

pongad:
Previously we import packages for both request and responses.
In examples, it is not necessary to import packages for responses,
since we only use short declaration for response variables.


src/main/java/com/google/api/codegen/transformer/go/GoGapicSurfaceTransformer.java
```diff
@@ -314,10 +314,19 @@ private void addXApiImports(SurfaceTransformerContext context) {
...
-  private void addXExampleImports(SurfaceTransformerContext context) {
+  @VisibleForTesting
+  void addXExampleImports(SurfaceTransformerContext context) {
+    // Examples are different from the API. In particular, we use short declaration
+    // and so we omit most type names. We only need
+    //   context, to initialize the client
+    //   the VKit generated library, that's what the sample is for
+    //   the input types of the methods, to initialize the requests
     ModelTypeTable typeTable = context.getTypeTable();
+    typeTable.getImports().clear();
```
garrettjonesgoogle: I think it would be clearer for the imports to be reset in generateExample, where the imports are actually used. Then this method would purely "add" (thus performing an action that matches its name).

src/test/java/com/google/api/codegen/transformer/go/testdata/myproto.proto
```diff
@@ -2,10 +2,14 @@ syntax = "proto3";
...
 package google.example.myproto.v1;
 
+import "singleservice.proto";
+
 service Gopher {
   rpc SimpleMethod(SimpleRequest) returns (SimpleResponse);
   rpc RetryMethod(SimpleRequest) returns (SimpleResponse);
   rpc PageStreamMethod(PageStreamRequest) returns (PageStreamResponse);
+
+  rpc MonkeyFood(SimpleRequest) returns (google.example.odd.v1.SpicyBanana);
```
garrettjonesgoogle: I think naming the method as a command would be more natural, e.g. "FeedMonkey".

src/main/java/com/google/api/codegen/transformer/go/GoGapicSurfaceTransformer.java
pongad: Done

src/test/java/com/google/api/codegen/transformer/go/testdata/myproto.proto
pongad: Done

pongad:
@garrettjonesgoogle PTAL.

src/main/java/com/google/api/codegen/transformer/go/GoGapicSurfaceTransformer.java
```diff
@@ -188,6 +188,13 @@ private StaticLangXExampleView generateExample(SurfaceTransformerContext context
         generateApiMethods(
             context, context.getSupportedMethods(), Collections.<String>emptyList()));
 
+    // Examples are different from the API. In particular, we use short declaration
+    // and so we omit most type names. We only need
+    //   context, to initialize the client
+    //   the VKit generated library, that's what the sample is for
+    //   the input types of the methods, to initialize the requests
```
garrettjonesgoogle: Put numbers on these comments and capitalize the first word of each line - as it is now, it looks like a weirdly formatted run-on sentence.

garrettjonesgoogle:
LGTM after the comment is fixed.

src/main/java/com/google/api/codegen/transformer/go/GoGapicSurfaceTransformer.java
pongad: Makes sense. Done.

PR #577, opened 2016-10-05
Fix a typo in ApiCallSettings snip

shinfan:


src/test/java/com/google/api/codegen/testdata/java_settings_library.baseline
```diff
@@ -717,9 +717,6 @@ public class LibraryServiceSettings extends ServiceApiSettings {
           addCommentsSettings,
           getBookFromArchiveSettings,
           updateBookIndexSettings,
-          streamShelvesSettings,
```
michaelbausor: Why are these disappearing as part of this change?

shinfan: Those settings are streaming settings which should not be a part of unaryMethodSettingsBuilders array.

michaelbausor: Yes, but what change in toolkit is making them disappear? So, how is xsettingsClass.unaryCallSettings changing?

shinfan: ummm before the template was referring  xsettingsClass.callSettings which contains all settings but I corrected it to xsettingsClass.unaryCallSettings which only contains unary settings. Does it answer you question?

garrettjonesgoogle:
LGTM

PR #580, opened 2016-10-05
Rename some package versions in Go discovery sample gen

tcoffee-google:
Resolves related compile-check errors.

### DIFF
--- a/src/main/java/com/google/api/codegen/go/GoDiscoveryContext.java
+++ b/src/main/java/com/google/api/codegen/go/GoDiscoveryContext.java
@@ -26,6 +26,9 @@ import com.google.protobuf.Field;
 import com.google.protobuf.Method;
 import com.google.protobuf.Type;
 
+import java.util.regex.Pattern;
+import java.util.regex.Matcher;
+
 public class GoDiscoveryContext extends DiscoveryContext implements GoContext {
   public GoDiscoveryContext(Service service, ApiaryConfig apiaryConfig) {
     super(service, apiaryConfig);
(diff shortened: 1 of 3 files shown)


tcoffee-google:
This will also need to be adapted to Go MVVM.

src/main/java/com/google/api/codegen/go/GoDiscoveryContext.java
```diff
@@ -159,21 +162,21 @@ public boolean isPageStreaming(Method method) {
...
   public String getApiVersion() {
-    String rename = API_VERSION_RENAME.get(getApi().getName(), getApi().getVersion());
-    return rename == null ? getApi().getVersion() : rename;
+    String version = getApi().getVersion();
+    if (version.equals("alpha") || version.equals("beta")) {
+      return "v0." + version;
+    }
+    Matcher subVersion = SUB_VERSION.matcher(version);
+    if (subVersion.matches()) {
+      return subVersion.group(1) + "/" + subVersion.group(2);
```
vchudnov-g: Where is this slashed output used?

vchudnov-g:
Fore reference here (and when doing MVVM, @saicheems):

See https://github.com/google/google-api-go-client/issues/170

The alpha and beta versions as well as api names of the form "apiname_v999" are remapped here: https://github.com/google/google-api-go-client/blob/master/google-api-go-generator/gen.go#L320

src/main/java/com/google/api/codegen/go/GoDiscoveryContext.java
tcoffee-google: IIRC, this case only arises in admin APIs, and not in our baselines.

vchudnov-g: And the extra path component works fine in the imports? OK....

PR #584, opened 2016-10-06
Add overrides for Node.js

saicheems:


saicheems:
The TYPE_STRING overrides are actually pretty tricky - what's happening is in the discovery doc is, for some types in adexchangebuyer and other APIs, type is listed as "string", but the format is listed as "int64". For Java, the correct choice in terms of the client libraries is to use "int64", and for Node.js the correct choice is to use "string".

The "resource" to "resource_" overrides are straightforward. Theoretically you could do this in the transformer, but I don't think that's in the spirit of MVVM.

tcoffee-google:
Surely we don't want the handling of TYPE_STRING to be done as overrides ...? Or perhaps I misunderstand.

saicheems:
See comment above, Java and Node.js do the opposite thing in relation to these strings - since we derive the kind from the `Field` object, one of these languages is doing this wrong.

tcoffee-google:
It's not clear to me that one approach is correct; the origin of the confusion comes from JSON's lack of certain types, requiring them to be encoded as strings. Let's generalize MVVM if need be to handle different options, equivalently to how it was done in *DiscoveryContext.java.

src/main/java/com/google/api/codegen/discovery/transformer/nodejs/NodeJSSampleTypeNameConverter.java
```diff
@@ -46,9 +46,9 @@
       ImmutableMap.<Field.Kind, String>builder()
           .put(Field.Kind.TYPE_BOOL, "false")
           .put(Field.Kind.TYPE_INT32, "0")
-          .put(Field.Kind.TYPE_INT64, "0")
+          .put(Field.Kind.TYPE_INT64, "\'\'")
```
tcoffee-google: You don't need to escape these quotes.

saicheems: Done

tcoffee-google:
LGTM

PR #585, opened 2016-10-07
Add streaming support to Java automated tests.

shinfan:
- Add grpc streaming support to unit tests and grpc mock classes
  - Add MockStreamObserver to GAX (https://github.com/googleapis/gax-java/pull/119/)
  - Support generate callable method tests for grpc streaming
- Add grpc reroute to SurfaceTransformerContext

Tests:
- Baseline tests

### DIFF
--- a/src/main/java/com/google/api/codegen/config/MethodConfig.java
+++ b/src/main/java/com/google/api/codegen/config/MethodConfig.java
@@ -29,11 +29,14 @@ import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
+
+import org.joda.time.Duration;
+
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
+
 import javax.annotation.Nullable;
-import org.joda.time.Duration;
 
 // TODO(garrettjones) consider using AutoValue in this class and related classes.
 /**
@@ -280,9 +283,9 @@ public class MethodConfig {
     }
   }
 
-  /** Returns true if this method has flattening configured. */
+  /** Returns true if this method supports parameter flattening. */
   public boolean isFlattening() {
-    return flattening != null;
+    return !isGrpcStreaming() && flattening != null;
   }
 
   /** Returns the flattening configuration of the method. */
--- a/src/main/java/com/google/api/codegen/config/MethodConfig.java
+++ b/src/main/java/com/google/api/codegen/config/MethodConfig.java
@@ -283,9 +283,9 @@ public class MethodConfig {
     }
   }
 
-  /** Returns true if this method supports parameter flattening. */
+  /** Returns true if this method has flattening configured. */
   public boolean isFlattening() {
-    return !isGrpcStreaming() && flattening != null;
+    return flattening != null;
   }
 
   /** Returns the flattening configuration of the method. */
--- a/src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
@@ -42,7 +42,6 @@ import com.google.api.codegen.viewmodel.testing.GapicSurfaceTestAssertView;
 import com.google.api.tools.framework.model.Field;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
-
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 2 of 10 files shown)


garrettjonesgoogle:
I saw several commits by e-mail, but it looks like you erased them my squashing commits...

src/test/java/com/google/api/codegen/testdata/java_test_library.baseline
```diff
@@ -650,6 +657,105 @@ public class LibraryServiceTest {
...
+    MockStreamObserver<StreamingRecognizeResponse> responseObserver = new MockStreamObserver<>();
+
+    StreamingApiCallable<StreamShelvesRequest, StreamShelvesResponse> callable =
+        api.streamShelvesCallable();
+    callable.serverStreamingCall(request, responseObserver);
+
+    StreamShelvesRequest request =
+        StreamShelvesRequest.newBuilder().build();
+    requestObserver.onNext(request);
+    requestObserver.onCompleted();
```
garrettjonesgoogle: Have you actually tried running these tests yet? I can see this won't compile...

* `requestObserver` isn't defined
* `request` is passed to serverStreamingCall, but defined after that line

shinfan: Sorry I was running the tests against the speech API which does not cover this server streaming case. I have tested the code against the BigTable API and fixed these issues. Consider it done.

shinfan:
@garrettjonesgoogle Right, I thought the commits history was a bit messy so I squashed them into one. I thought doing it would be okay if no one started reviewing yet.

PTAL

src/main/java/com/google/api/codegen/config/MethodConfig.java
```diff
@@ -280,9 +283,9 @@ public GrpcStreamingType getGrpcStreamingType() {
     }
   }
 
-  /** Returns true if this method has flattening configured. */
+  /** Returns true if this method supports parameter flattening. */
   public boolean isFlattening() {
-    return flattening != null;
+    return !isGrpcStreaming() && flattening != null;
```
garrettjonesgoogle: Wait, so will this silently ignore flattening configuration if it is a grpc streaming method? If we don't want to allow it, we should emit an error somewhere instead of hiding it.

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -95,7 +97,11 @@ public InitCodeView generateTestMethodInitCode(
                 .initFields(fields)
                 .suggestedName(Name.from("request"))
                 .build());
-    return buildInitCodeViewFlattened(context, rootNode);
+    if (isFlattened) {
+      return buildInitCodeViewFlattened(context, rootNode);
+    } else {
+      return buildInitCodeViewRequestObject(context, rootNode);
```
garrettjonesgoogle: All of our other entry points for flattening vs request object are differentiated as separate entry points, instead of using a boolean flag. I would prefer to be consistent.

src/test/java/com/google/api/codegen/testdata/java_test_library.baseline
```diff
@@ -652,6 +659,102 @@ public class LibraryServiceTest {
...
+    callable.serverStreamingCall(request, responseObserver);
+
+    try {
+      List<StreamShelvesResponse> actualResponses = responseObserver.future.get();
+      Assert.assertEquals(1, actualResponses.size());
+      Assert.assertEquals(expectedResponse, actualResponses.get(0));
+      Assert.assertEquals(0, responseObserver.errors.size());
+    } catch (Exception e) {
+      Assert.fail();
+    }
```
garrettjonesgoogle: Why do we need a try/catch here? Won't an exception cause the test to fail anyway?

src/main/java/com/google/api/codegen/config/MethodConfig.java
shinfan: Reverted.

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
Now we have 6 different entry points to generate init code for different cases. I would simplified this in a separate PR.

src/test/java/com/google/api/codegen/testdata/java_test_library.baseline
shinfan: try/catch removed.

garrettjonesgoogle:
LGTM

garrettjonesgoogle:
LGTM2

PR #589, opened 2016-10-07
Fix some PHP compilecheck bugs

saicheems:


src/main/resources/com/google/api/codegen/php/discovery_fragment.snip
```diff
@@ -248,8 +248,8 @@
   optParams
 @end
 
-@private requestBodyName()
-  requestBody
+@private postBodyName()
+  postBody
```
tcoffee-google: Are there actual errors caused by this variable name? I looked everywhere for documentation of such and found nothing.

saicheems: Yeah... I think the compilecheck script is doing 1:1 comparisons of the parameter names

tcoffee-google: Per offline discussion, let's keep our original name and rewrite it in the compile checker like the one for Python.

saicheems: Done

vchudnov-g:
If/when the compilecheck passes, LGTM

saicheems:
(compile check does not succeed because the client libraries are lagging the discovery doc)

vchudnov-g:
Is that the only issue we seem to be having? In that case, go ahead and submit. We'll get the PHP libraries updated.

vchudnov-g:
(BTW, if you have a list handy of the APIs that fail due to the libraries being out of date, please send it my way.)

PR #590, opened 2016-10-07
Switch Node.js discovery samplegen to MVVM

saicheems:


saicheems:
ping

src/test/java/com/google/api/codegen/testdata/discoveries/nodejs/nodejs_translate.v2.json.baseline
```diff
@@ -37,13 +35,14 @@ translate.detections.list(request, function(err, response) {
 //    https://console.developers.google.com/apis/api/translate
 // 2. Install the Node.js client library by running
 //    `npm install googleapis --save`
+
 var google = require('googleapis');
 var translate = google.translate('v2');
 
-var request = {
+var apiKey = ''; // TODO: Update placeholder with desired API key.
```
tcoffee-google: Let's have two spaces before comment for consistency.

saicheems:
Yes

src/test/java/com/google/api/codegen/testdata/discoveries/nodejs/nodejs_translate.v2.json.baseline
saicheems: Done

saicheems:
Rebased, PTAL

tcoffee-google:
LGTM

src/main/java/com/google/api/codegen/discovery/MainDiscoveryProviderFactory.java
```diff
@@ -99,11 +98,14 @@ public static DiscoveryProvider defaultCreate(
           .setOutputRoot(outputRoot)
           .build();
     } else if (id.equals(NODEJS)) {
-      return CommonDiscoveryProvider.newBuilder()
-          .setContext(new NodeJSDiscoveryContext(service, apiaryConfig))
-          .setSnippetSetRunner(
-              new NodeJSSnippetSetRunner<Method>(SnippetSetRunner.SNIPPET_RESOURCE_ROOT))
-          .setSnippetFileName(id + "/" + DEFAULT_SNIPPET_FILE)
+      return ViewModelProvider.newBuilder()
```
saicheems: Did it slightly differently, PTAL

vchudnov-g: I think this is better than the original, but it still feels repetitive. You could have two hash map literals of the form
```
Map<String,Class> test = 
    new HashMap<String, Class>{
      NODEJS: NodeJSSampleMethodToViewTransformer.getClass(),
      ...}
```

and then have a single builder/return statement rather than the if chain (or switch. why wasn't the code using a switch to begin with?).

These hash maps could easily be static variables, BTW.

(And there might be ways to get away without even getClass, maybe?)

The advantage of this is that all the builder methods are called in one place, and the parameters that differ are very easy to visually inspect. And it's more concise.

src/main/java/com/google/api/codegen/discovery/MainDiscoveryProviderFactory.java
```diff
@@ -129,10 +120,36 @@ public static DiscoveryProvider defaultCreate(
...
 
-    } else {
+    // Below is the MVVM pathway.
+    SampleMethodToViewTransformer sampleMethodToViewTransformer = null;
+    TypeNameGenerator typeNameGenerator = null;
+    try {
+      sampleMethodToViewTransformer = SAMPLE_METHOD_TO_VIEW_TRANSFORMER_MAP.get(id).newInstance();
+      typeNameGenerator = TYPE_NAME_GENERATOR_MAP.get(id).newInstance();
+    } catch (Exception e) {
+    }
```
garrettjonesgoogle: why are you hiding the exception?

saicheems: It didn't seem particularly useful to do anything with it, on the next line I check if `sampleMethodToViewTransformer` or `typeNameGenerator` are null, in which case I raise another exception.

The only reason it should fail is if `id` isn't handled - otherwise if `newInstance` is called where the constructors take arguments but none are provided.

saicheems: Should I print the stack-trace, or throw a `RuntimeException`?

saicheems: Opted to throw a `RuntimeException`

src/main/java/com/google/api/codegen/discovery/MainDiscoveryProviderFactory.java
```diff
@@ -54,6 +59,16 @@
 
   private static final String DEFAULT_SNIPPET_FILE = "discovery_fragment.snip";
 
+  private static final Map<String, Class<? extends SampleMethodToViewTransformer>>
+      SAMPLE_METHOD_TO_VIEW_TRANSFORMER_MAP =
+          ImmutableMap.of(
+              JAVA,
```
vchudnov-g: nit: if we can keep keys and values in the same line, it will be easier for humans to parse

src/main/java/com/google/api/codegen/discovery/MainDiscoveryProviderFactory.java
```diff
@@ -54,6 +59,16 @@
...
+              JavaSampleMethodToViewTransformer.class,
+              NODEJS,
+              NodeJSSampleMethodToViewTransformer.class);
+  private static final Map<String, Class<? extends TypeNameGenerator>> TYPE_NAME_GENERATOR_MAP =
+      ImmutableMap.of(JAVA, JavaTypeNameGenerator.class, NODEJS, NodeJSTypeNameGenerator.class);
```
vchudnov-g: nit: if we can keep one language per line (key/value pair), it will be easier for humans to parse

src/main/java/com/google/api/codegen/discovery/MainDiscoveryProviderFactory.java
```diff
@@ -129,10 +120,37 @@ public static DiscoveryProvider defaultCreate(
...
+      sampleMethodToViewTransformer = SAMPLE_METHOD_TO_VIEW_TRANSFORMER_MAP.get(id).newInstance();
+      typeNameGenerator = TYPE_NAME_GENERATOR_MAP.get(id).newInstance();
+    } catch (InstantiationException | IllegalAccessException e) {
+      throw new RuntimeException(e);
+    }
+    if (sampleMethodToViewTransformer == null || typeNameGenerator == null) {
       throw new NotImplementedException("MainDiscoveryProviderFactory: invalid id \"" + id + "\"");
     }
+
+    ViewModelProvider.Builder viewModelProvider = ViewModelProvider.newBuilder();
```
vchudnov-g: Is it more conventional to have this variable definition in a separate line or just inline it in the one place it's used? I'm fine either way, just wondering whether there's a reason you chose this.

(I'm referring to using `return ViewModelProvider.newBuilder().setMethods(...).setApiary(...)....build();`)

src/main/java/com/google/api/codegen/discovery/MainDiscoveryProviderFactory.java
saicheems: I'll see if I can make it happen, this is something that the formatter seems to enforce... It'll probably be the case that once another language is added, the line will be long enough that the formatter will break it into the way you want.

src/main/java/com/google/api/codegen/discovery/MainDiscoveryProviderFactory.java
saicheems: Yeah, I tried to do it, the formatter is strongly opinionated and forces it back :/

src/main/java/com/google/api/codegen/discovery/MainDiscoveryProviderFactory.java
saicheems: Done, I could've sworn the formatter was doing weird stuff before doing it the other way, but looks like that's not the case anymore haha

vchudnov-g:
LGTM

PR #594, opened 2016-10-10
Add override for CloudResourceManager API in discovery sample gen

tcoffee-google:


src/test/java/com/google/api/codegen/testdata/discoveries/cloudresourcemanager.v1.json.overrides
```diff
@@ -0,0 +1,56 @@
+{
+  "java": {
+    "apiTypeName": "CloudResourceManager",
+    "methods": {
+      "cloudresourcemanager.organizations.search": {
+        "isPageStreamingResourceSetterInRequestBody": true
```
saicheems: Wow I'm actually glad this weird flag came in handy twice :)

saicheems:
lgtm

saicheems:
if compilecheck succeeds lgtm

src/test/java/com/google/api/codegen/testdata/discoveries/nodejs/nodejs_cloudresourcemanager.v1.json.baseline
```diff
@@ -0,0 +1,767 @@
...
+    authClient = authClient.createScoped(scopes);
+  }
+
+  var request = {
+
+    // TODO: Change placeholder below to desired parameter value for the `setIamPolicy` method:
+
+    // REQUIRED: The resource for which the policy is being specified. `resource` is usually specified as a
+    // path. For example, a Project resource is specified as `projects/{project}`.
+    resource_: 'organizations/{MY-ORGANIZATION}',
```
vchudnov-g: To clarify: the NodeJS client library request (and only that)  actually expects both `resource` and `resource_`???

tcoffee-google: This is copied from cloudresourcemanager v1beta1, and there's a similar collision in v1 ... @saicheems any insight?

saicheems: All request bodies in Node.js are called `resource`, but Cloud Resource Manager also has a field called `resource`. The library owners decided to resolve the conflict by renaming the field to `resource`.

This override is applied in NodeJSDiscoveryContext, so once the Node.js MVVM flow gets merged, it will be applied from the overrides file.

tcoffee-google:
I have been unsuccessful in getting useful output from the C# compile check, and the Java compile check is stuck on other errors. So I currently can't confirm that compile check errors are resolved.

src/test/java/com/google/api/codegen/testdata/discoveries/nodejs/nodejs_cloudresourcemanager.v1.json.baseline
vchudnov-g: Unfortunate collision....

tcoffee-google:
Java compile check no longer shows these errors. Java and Python show new errors because CloudResourceManager discovery doc is out of sync with client libraries (and API documentation ... I can't even tell which is supposed to be right). Since @vchudnov-g will be running a full compile check again soon, I'll submit this for now so we can verify C#.

PR #597, opened 2016-10-11
Fix Python oneof default parameter generation

### LINKED ISSUE #591: Multiple oneof fields are being set
### What

`oneof` is intended to represent a single field, yet in the Python generated code appears as multiple fields, only one of which may be set at any time. In the current Python codegen, we are setting multiple of these. For example, there might be a proto request:

```
message ListFooRequest {
  ...
  oneof msg {
    string oneof1 = 17;
    string oneof2 = 18;
  }
}
```

which would generate

```
def list_foo(..., oneof1='', oneof2=''):
  request = ListFooRequest(..., oneof1=oneof1, oneof2=oneof2)
```

This is incorrect.

### Why

Setting a oneof to its proto3 default value is *not* the same as not setting it. Instead, we should only set oneof fields in the generated proto class if they are set in method parameters:

```
def list_foo(..., oneof1=None, oneof2=None):
  if oneof1 is not None:
    request = ListFooRequest(..., oneof1=oneof1)
  if oneof2 is not None:
    request = ListFooRequest(..., oneof2=oneof1)
```

This gets messy if there are multiple oneofs in a message (# lines of code is exponential in the number oneofs), but we can clean this up by letting GAX do the work in a utility function that constructs the oneof args:

```
def list_foo(..., oneof1=None, oneof2=None):
    oneof_args = gax.oneof_args(ListFooRequest, oneof1=oneof1, oneof2=oneof2)
    request = ListFooRequest(..., **oneof_args)
```

cc: @bjwatson

geigerj:
Previously multiple oneof fields were being set.

Fixes #591

### DIFF
--- a/src/test/java/com/google/api/codegen/testdata/library_config.baseline
+++ b/src/test/java/com/google/api/codegen/testdata/library_config.baseline
@@ -131,10 +131,12 @@ interfaces:
         - shelf
         - books
         - edition
+        - review_copy
     required_fields:
     - shelf
     - books
     - edition
+    - review_copy
     request_object_method: true
     retry_codes_name: non_idempotent
     retry_params_name: default
--- a/src/main/java/com/google/api/codegen/GapicContext.java
+++ b/src/main/java/com/google/api/codegen/GapicContext.java
@@ -26,7 +26,6 @@ import com.google.api.tools.framework.model.Method;
 import com.google.api.tools.framework.model.Model;
 import com.google.api.tools.framework.model.ProtoElement;
 import com.google.common.base.Preconditions;
-
 import java.util.ArrayList;
 import java.util.List;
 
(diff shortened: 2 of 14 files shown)


src/test/java/com/google/api/codegen/testdata/library_config.baseline
```diff
@@ -131,10 +131,12 @@ interfaces:
         - shelf
         - books
         - edition
+        - review_copy
```
bjwatson: A similar update has not been done to `src/test/java/com/google/api/codegen/testdata/library_gapic.yaml`. Also, isn't it a problem for oneof fields to be required, since only one of them can be populated?

geigerj: Yes, we should update the config generator not to mark `oneof` fields as required. I will do that as a separate PR, because it's orthogonal to the default oneof value in Python.

bjwatson: Thanks. Please also warn in the GAPIC codegen if a oneof field is marked required or flattened that we are ignoring required/flattened for that field.

PR #599, opened 2016-10-11
Go specially handles IAM resources

pongad:
This commit makes use of https://code-review.googlesource.com/#/c/8130/


src/test/java/com/google/api/codegen/testdata/library_gapic.yaml
```diff
@@ -49,6 +49,12 @@ interfaces:
     init_fields:
       - name%shelf="testShelf"
       - name%book="testBook"
+  iam_resources:
+  - type: google.example.library.v1.Book
+    field: name
+  iam_replace_methods:
+    - AddTag
+    - AddLabel
```
garrettjonesgoogle: This is not the right level of abstraction for GAPIC config; our config should not be aware of domain-specific concerns (of which IAM is an example). A more appropriate level of abstraction is mix-ins. For example, I added the `reroute_to_grpc_interface` to handle the pubsub mix-in hack. It looks like for Go, you want to ignore that rerouting and use a different mechanism. We should be careful about "settings clusters" where some settings are used by some languages but a mutually-exclusive set of settings are used by others. We have this for flattening vs required fields; in that case, the connection is well-enough understood, and they sit side-by-side. In your case, there is no apparent connection. 

If we are supporting a domain-modeled approach here, we should define the settings in terms of that concept. Potentially we should be explicit in config that this only applies to Go. 

* `iam_replace_methods`: potentially merge with `reroute_to_grpc_interface` to create a new abstraction. Example:

```
  - name: AddLabel
    ...
    surface_treatments:
    - exclude_languages: go
      reroute_to_grpc_interface: google.tagger.v1.Labeler
    - include_languages: go
      visibility: disabled
```

* Potentially bootstrap config related to generating accessors to handles like you need:
```
interfaces:
- name: google.example.library.v1.LibraryService
  ...
  handles:
  - include_languages: go
    input:
    - name: book
      type: google.example.library.v1.Book
    passthrough:
    // This should be understood in a language-independent way; "self" refers to the api wrapper object.
    - name: self.connection
    // dotted accessor path
    - name: book.name
    output:
    - type: google.cloud.iam.Handle
```

src/test/java/com/google/api/codegen/testdata/go_main_library.baseline
```diff
@@ -219,6 +213,10 @@ func LibraryArchivedBookPath(archivePath, book string) string {
 }
 
 
+func (c *Client) BookIAM(book *librarypb.Book) *iam.Handle {
+    return iam.InternalNewHandle(c.Connection(), book.Name)
+}
```
garrettjonesgoogle: This is self-inconsistent. If you are removing AddTag/AddLabel, your replacement accessor function should match them semantically; they should be on the topic of tagging/labeling instead of IAM.

src/test/java/com/google/api/codegen/testdata/library_gapic.yaml
pongad: Done as discussed.

pongad:
@garrettjonesgoogle PTAL

src/main/java/com/google/api/codegen/transformer/go/GoSurfaceNamer.java
```diff
@@ -214,4 +209,25 @@ public String getStaticLangStreamingReturnTypeName(Method method, MethodConfig m
...
+        .join(Name.upperCamelKeepUpperAcronyms("IAM"))
+        .toUpperCamel();
+  }
+
+  @Override
+  public String getIamResourceGetterFunctionExampleName(Interface service, Field field) {
+    return exampleFunction(service, getIamResourceGetterFunctionName(field));
+  }
+
+  private String exampleFunction(Interface service, String func) {
```
garrettjonesgoogle: rename `func` to `functionName`

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -314,3 +318,36 @@ message BundlingDescriptorProto {
...
+
+// ExperimentalFeatures contain special-case features that,
+// for lack of similar problems, we don't quite know how to properly
+// expose as a config.
+// Configs defined here should eventually be replaced with more general ones.
+message ExperimentalFeatures {
+  repeated IamResourceProto iam_resources = 1;
+}
+
+message IamResourceProto {
```
garrettjonesgoogle: add a comment saying that this is experimental - even though it is already in ExperimentalFeatures, it might get moved around / separated later.

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -314,3 +318,36 @@ message BundlingDescriptorProto {
...
+  // Fully qualified name of resource types. Eg, google.example.library.v1.Book.
+  string type = 1;
+
+  // The field containing the resource name.
+  string field = 2;
+}
+
+message SurfaceTreatmentProto {
+  repeated string include_languages = 1;
+  repeated TreatmentProto treatments = 2;
```
garrettjonesgoogle: add comments on these

src/test/java/com/google/api/codegen/testdata/go_main_library.baseline
garrettjonesgoogle: ok.

src/test/java/com/google/api/codegen/testdata/library_gapic.yaml
```diff
@@ -382,6 +386,11 @@ interfaces:
     field_name_patterns:
       resource: book
     timeout_millis: 60000
+    surface_treatments:
+    - include_languages:
+      - go
+      treatments:
+      - visibility: DISABLED
```
garrettjonesgoogle: what is the motivation for using oneof? It seems like it could be slightly simpler this way:

```
  surface_treatments:
  - include_languages:
    - go
    visibility: DISABLED
```

src/main/java/com/google/api/codegen/transformer/go/GoSurfaceNamer.java

src/main/proto/com/google/api/codegen/config.proto
pongad: Done.

src/main/proto/com/google/api/codegen/config.proto
pongad: Done

src/test/java/com/google/api/codegen/testdata/library_gapic.yaml
pongad: You're right, I was overthinking it.

pongad:
@garrettjonesgoogle PTAL

garrettjonesgoogle:
LGTM

PR #600, opened 2016-10-11
Go: Refer to hand-written client when one exists

pongad:


src/test/java/com/google/api/codegen/testdata/go_doc_library.baseline
```diff
@@ -19,6 +19,8 @@
 // library API.
 //
 // A simple Google Example Library API.
+//
+// Use the handwritten client (cloud.google.com/go/library) in preference to this.
```
jba: I think we shouldn't use the word "handwritten". 

Instead of trying to find the perfect word, let's just say "Use the client at {path} in preference to this."

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -84,6 +84,9 @@ message ConfigProto {
 message LanguageSettingsProto {
   // Package name used for the language.
   string package_name = 1;
+
+  // Location of hand-written client, if any.
+  string hand_written_location = 2;
```
garrettjonesgoogle: I think I would prefer something more along the lines of what I suggested elsewhere: `domain_layer_location`. The wording in the Go docs can be settled separately.

pongad:
@jba @garrettjonesgoogle PTAL

src/main/java/com/google/api/codegen/config/ApiConfig.java
```diff
@@ -40,10 +40,14 @@
   /**
    * Returns the package name.
    */
-  @Nullable
   public abstract String getPackageName();
 
   /**
+   * Returns the location of the hand-written client, if any.
+   */
+  public abstract String getDomainLayerLocation();
```
garrettjonesgoogle: Update the doc to reflect the new name.

garrettjonesgoogle:
LGTM after you address the small doc change I requested.

src/main/java/com/google/api/codegen/config/ApiConfig.java
pongad: Done.

PR #606, opened 2016-10-12
Refactor Go discovery snippetgen for MVVM

saicheems:


src/main/java/com/google/api/codegen/discovery/transformer/go/GoSampleTypeNameConverter.java
```diff
@@ -0,0 +1,114 @@
...
+    if (primitiveTypeName != null) {
+      return new TypeName(primitiveTypeName);
+    }
+    if (typeInfo.kind() != Field.Kind.TYPE_MESSAGE) {
+      throw new IllegalArgumentException("unsupported type kind: " + typeInfo.kind());
+    }
+    if (typeInfo.isMap()) {
+      throw new IllegalArgumentException("map types are unsupported");
+    }
+    // TODO(garrettjones): Is this okay?
```
saicheems: @garrettjonesgoogle Is this okay? I'm using a static method in the sample namer to pull apart the Go package. Mostly I just want this functionality in one place, since I use it twice.

src/main/java/com/google/api/codegen/discovery/transformer/go/GoSampleTypeNameConverter.java
```diff
@@ -0,0 +1,114 @@
...
+  @Override
+  public TypeName getTypeNameForElementType(TypeInfo typeInfo) {
+    String primitiveTypeName = PRIMITIVE_TYPE_MAP.get(typeInfo.kind());
+    if (primitiveTypeName != null) {
+      return new TypeName(primitiveTypeName);
+    }
+    if (typeInfo.kind() != Field.Kind.TYPE_MESSAGE) {
+      throw new IllegalArgumentException("unsupported type kind: " + typeInfo.kind());
+    }
+    if (typeInfo.isMap()) {
```
tcoffee-google: Why is this not supported?

src/main/java/com/google/api/codegen/discovery/transformer/java/JavaSampleTypeNameConverter.java
```diff
@@ -168,9 +168,6 @@ private TypeName getTypeNameForElementType(TypeInfo typeInfo, boolean shouldBoxP
    */
   @Override
   public TypedValue getZeroValue(TypeInfo typeInfo) {
-    if (typeInfo.isMap()) {
```
tcoffee-google: Why is this gone?

src/main/java/com/google/api/codegen/discovery/transformer/nodejs/NodeJSSampleTypeNameConverter.java
```diff
@@ -70,14 +70,13 @@ public TypeName getRequestTypeName(String apiTypeName, TypeInfo typeInfo) {
...
   public TypeName getTypeName(TypeInfo typeInfo) {
-    if (typeInfo.isMap()) {
-      return new TypeName("Object");
-    } else if (typeInfo.isArray()) {
+    if (typeInfo.isArray()) {
       TypeName elementTypeName = getTypeNameForElementType(typeInfo);
       return new TypeName("", "", "%i[]", elementTypeName);
-    } else {
-      return getTypeNameForElementType(typeInfo);
+    } else if (typeInfo.isMap()) {
```
tcoffee-google: And here?

src/main/java/com/google/api/codegen/discovery/transformer/nodejs/NodeJSSampleTypeNameConverter.java
```diff
@@ -92,18 +91,12 @@ public TypeName getTypeNameForElementType(TypeInfo typeInfo) {
   @Override
   public TypedValue getZeroValue(TypeInfo typeInfo) {
     // Don't call getTypeName; we don't need to import these.
-    if (typeInfo.isMap()) {
```
tcoffee-google: ?

src/main/java/com/google/api/codegen/discovery/transformer/nodejs/NodeJSSampleTypeNameConverter.java
```diff
@@ -92,18 +91,12 @@ public TypeName getTypeNameForElementType(TypeInfo typeInfo) {
...
-      return TypedValue.create(new TypeName("Object"), "{}");
-    }
     if (typeInfo.isArray()) {
       return TypedValue.create(new TypeName("Array"), "[]");
     }
     if (PRIMITIVE_ZERO_VALUE.containsKey(typeInfo.kind())) {
       return TypedValue.create(getTypeName(typeInfo), PRIMITIVE_ZERO_VALUE.get(typeInfo.kind()));
     }
-    if (typeInfo.isMessage()) {
```
tcoffee-google: ?

src/main/resources/com/google/api/codegen/java/sample.snip
```diff
@@ -67,12 +67,12 @@
     @end
     {@class.serviceTypeName} {@class.serviceVarName} = {@class.createServiceFuncName}();
     {@class.requestTypeName} {@class.requestVarName} = {@class.serviceVarName}.{@callChain(class.methodNameComponents)}({@paramList(class.fieldVarNames)});
-    @if {@class.hasMediaDownload()}
 
+    @if {@class.hasMediaDownload()}
```
tcoffee-google: This change doesn't affect any baselines, correct?

src/main/java/com/google/api/codegen/discovery/transformer/nodejs/NodeJSSampleTypeNameConverter.java
saicheems: So all these things that I removed shouldn't happen, I removed them straight up because I want an exception raised if we encounter these conditions:

If I'm not mistaken, fields can be objects and arrays but they cannot be maps. Also, if a method has a request body, the field cannot be a message. I opted to fail entirely if we do happen to see that since these code paths don't get tested at all, and probably cannot be tested.

src/main/java/com/google/api/codegen/discovery/transformer/go/GoSampleMethodToViewTransformer.java
```diff
@@ -0,0 +1,211 @@
...
+import com.google.api.codegen.discovery.viewmodel.SampleFieldView;
+import com.google.api.codegen.discovery.viewmodel.SamplePageStreamingView;
+import com.google.api.codegen.discovery.viewmodel.SampleView;
+import com.google.api.codegen.util.Name;
+import com.google.api.codegen.util.SymbolTable;
+import com.google.api.codegen.util.go.GoTypeTable;
+import com.google.api.codegen.viewmodel.ViewModel;
+import com.google.protobuf.Method;
+
+public class GoSampleMethodToViewTransformer implements SampleMethodToViewTransformer {
```
saicheems: What do you mean?

src/main/java/com/google/api/codegen/discovery/transformer/go/GoSampleMethodToViewTransformer.java
```diff
@@ -0,0 +1,211 @@
...
+    addStaticImports(context);
+    SampleConfig config = context.getSampleConfig();
+    MethodInfo methodInfo = config.methods().get(context.getMethodName());
+    SampleNamer namer = context.getSampleNamer();
+    SampleTypeTable typeTable = context.getSampleTypeTable();
+    SymbolTable symbolTable = SymbolTable.fromSeed(GoTypeTable.RESERVED_IDENTIFIER_SET);
+
+    SampleView.Builder builder = SampleView.newBuilder();
+
+    String clientVarName = symbolTable.getNewSymbol(namer.localVarName(Name.lowerCamel("c")));
```
saicheems: Yeah, that's probably worth doing, I'll make a note

src/main/java/com/google/api/codegen/discovery/transformer/go/GoSampleTypeNameConverter.java
saicheems: See below

src/main/java/com/google/api/codegen/discovery/transformer/java/JavaSampleTypeNameConverter.java
saicheems: See below

src/main/java/com/google/api/codegen/discovery/transformer/nodejs/NodeJSSampleTypeNameConverter.java
saicheems: Below

src/main/java/com/google/api/codegen/discovery/transformer/nodejs/NodeJSSampleTypeNameConverter.java
saicheems: Still below

src/main/java/com/google/api/codegen/discovery/transformer/nodejs/NodeJSSampleTypeNameConverter.java
saicheems: Also below

saicheems: Discussed offline, these pathways are not tested by any doc we currently have, but at the risk of failing for customers in the future we'll include handling of maps and objects.

src/main/java/com/google/api/codegen/discovery/transformer/go/GoSampleMethodToViewTransformer.java
garrettjonesgoogle: If there are any functions in here that are mostly the same across languages, create a cross-language transformer and call it from the language-specific transformers.

src/main/java/com/google/api/codegen/discovery/transformer/go/GoSampleTypeNameConverter.java
garrettjonesgoogle: I'm perfectly fine with Go-specific utility classes calling static Go-related methods. My concern would actually be the other calling site, in the Go sample transformer; having a static call to Go logic makes it more difficult to move the transformer code to a cross-language transformer.

src/main/java/com/google/api/codegen/discovery/config/go/GoTypeNameGenerator.java
```diff
@@ -0,0 +1,71 @@
...
+package com.google.api.codegen.discovery.config.go;
+
+import java.util.List;
+
+import com.google.api.codegen.discovery.DefaultString;
+import com.google.api.codegen.discovery.config.TypeNameGenerator;
+import com.google.api.codegen.util.Name;
+import com.google.common.base.Strings;
+
+public class GoTypeNameGenerator implements TypeNameGenerator {
```
saicheems: Yeah, I noticed that as well... the scope of this class expanded a bit since it was first conceptualized.

I think the right move is just to rename it - it's a utility for generating language-specific content in the SampleConfig.

I've made note of it and will refactor once the rate of MVVM changes settles down.

src/test/java/com/google/api/codegen/testdata/discoveries/go/go_storage.v1.json.baseline
```diff
@@ -257,19 +227,18 @@ func main() {
...
   // group-emailAddress, allUsers, or allAuthenticatedUsers.
-  entity := "{MY-ENTITY}"
-  // TODO: Change placeholder value above.
+  entity := ""  // TODO: Update placeholder value.
 
-  rb := storage.BucketAccessControl{
-    // TODO: Add desired fields of the request body to be changed.
+  rb := &storage.BucketAccessControl{
+    // TODO: Add desired fields of the request body to be changed. Only assigned
+    // fields will be changed:
```
saicheems: Done

src/main/java/com/google/api/codegen/discovery/transformer/go/GoSampleTypeNameConverter.java
saicheems: Alright if I leave it for now and make a note?

saicheems:
Reverted changes and added support for maps for Go fields @tcoffee-google 
PTAL

src/main/java/com/google/api/codegen/discovery/transformer/go/GoSampleTypeNameConverter.java
garrettjonesgoogle: Remove the TODO here, and make a note in the transformer code about it.

saicheems:
OK lots of changes ensure that the only overrides required are disabling page streaming on some set of functions. I'll follow up with that in a future CL, but this one is looking good now. PTAL

src/test/java/com/google/api/codegen/testdata/discoveries/go/go_storage.v1.json.baseline
```diff
@@ -149,14 +131,13 @@ func main() {
   }
 
   // Name of a bucket.
-  bucket := "{MY-BUCKET}"
-  // TODO: Change placeholder value above.
+  bucket := ""  // TODO: Update placeholder value.
 
-  rb := storage.BucketAccessControl{
+  rb := &storage.BucketAccessControl{
```
saicheems: I would like to leave this as is - I really don't think that where we take the reference of the object impacts readability or improves any highlighting. Specifically, no Go developer will be impacted.

Two reasons:
- it requires another line of logic in the template
- zero values for messages, should they ever happen to exist will be defined as the reference to a new object: &Empty{}

src/main/java/com/google/api/codegen/discovery/transformer/go/GoSampleMethodToViewTransformer.java
```diff
@@ -0,0 +1,216 @@
...
+    String serviceVarName = symbolTable.getNewSymbol(namer.getServiceVarName(servicePackageName));
+    List<String> methodNameComponents = new ArrayList<String>();
+    for (String nameComponent : methodInfo.nameComponents()) {
+      methodNameComponents.add(namer.publicFieldName(Name.lowerCamel(nameComponent)));
+    }
+    String requestVarName = symbolTable.getNewSymbol(namer.localVarName(Name.lowerCamel("req")));
+    // For this and other type name assignments, we don't go through the symbol
+    // table for some minor conveniences in the sample template. Since all types
+    // we care about are imported with the package in addStaticImports we don't
+    // have to worry about missing imports.
```
vchudnov-g: I'm not clear what this comment means:

* what are the "minor conveniences" you refer to?
* is the second sentence related to the first?

src/main/java/com/google/api/codegen/discovery/transformer/go/GoSampleMethodToViewTransformer.java
```diff
@@ -0,0 +1,216 @@
...
+        .build();
+  }
+
+  private SampleAuthView createSampleAuthView(SampleTransformerContext context) {
+    SampleConfig config = context.getSampleConfig();
+    MethodInfo methodInfo = config.methods().get(context.getMethodName());
+    SampleNamer namer = context.getSampleNamer();
+
+    List<String> scopeConsts = new ArrayList<>();
+    // Pull the last scope and reconstruct it into a Go constant.
```
vchudnov-g: nit: s/into/as/

src/main/java/com/google/api/codegen/discovery/transformer/go/GoSampleMethodToViewTransformer.java
```diff
@@ -0,0 +1,216 @@
...
+        .build();
+  }
+
+  private void addStaticImports(SampleTransformerContext context, SymbolTable symbolTable) {
+    SampleConfig config = context.getSampleConfig();
+    SampleTypeTable typeTable = context.getSampleTypeTable();
+
+    // Since nearly any identifier can be shadowed in Go, we add every package
+    // name to the symbol table.
+    typeTable.saveNicknameFor("log;;;");
```
vchudnov-g: What does the `;;;` mean?

src/test/java/com/google/api/codegen/testdata/discoveries/go/go_storage.v1.json.baseline
vchudnov-g: initializing with references to structs is pretty standard in Go.

src/test/java/com/google/api/codegen/testdata/discoveries/go/go_storage.v1.json.baseline
```diff
@@ -18,16 +18,12 @@ import (
...
-
   "google.golang.org/api/storage/v1"
 )
 
 func main() {
   ctx := context.Background()
 
-  // Use Application Default Credentials for authentication when running locally.
-  // For more information, see:
-  // https://developers.google.com/identity/protocols/application-default-credentials
```
vchudnov-g: Huh, we're removing this link explaining ADC? I wonder whether we should move it to the end of comment "2". We can discuss separately from this PR and do uniformly across languages.

tcoffee-google: I overlooked this thinking it had already been done (moving to the end of comment 2). Yes, let's do that (for all languages).

src/main/java/com/google/api/codegen/discovery/transformer/go/GoSampleMethodToViewTransformer.java
saicheems: Wow, yeah this really reads like a word salad. I rewrote the paragraph, should make more sense now.

What I was trying to say was:
We're not adding the types we generate into the GoTypeTable because the type names returned are fully qualified ("logging.LogEntry"). Sometimes we only need the type name without the local name prefix. In the template we use `{class.packageName}.{class.requestTypeName}`.
We can't end up with a missing import because everything is added in `getStaticImports`

src/main/java/com/google/api/codegen/discovery/transformer/go/GoSampleMethodToViewTransformer.java
saicheems: Done

src/main/java/com/google/api/codegen/discovery/transformer/go/GoSampleMethodToViewTransformer.java
saicheems: Added an explanation comment:

```
// GoTypeTable expects imports to be of the format:
// "{packagePath};{localName};{typeName};{pointer}"
```

saicheems:
For the ADC stuff - I was planning to follow-up with auth like I did for Java and Node.js. Do you want me to include it in this PR?

src/main/java/com/google/api/codegen/discovery/transformer/SampleNamer.java
```diff
@@ -95,4 +95,20 @@ public String getResponseVarName() {
   public String createServiceFuncName(String apiTypeName) {
     return publicMethodName(Name.upperCamel("Create", apiTypeName, "Service"));
   }
+
+  /**
+   * Returns the constant form of given an authScope URL.
+   *
+   * Only intended for use in Go.
```
saicheems: Yes, I made it a static method in `GoSampleNamer` instead.

src/main/java/com/google/api/codegen/discovery/config/java/JavaTypeNameGenerator.java
```diff
@@ -29,6 +30,11 @@
   private static final String NON_REQUEST_SUBPACKAGE = "model";
 
   @Override
+  public String getApiVersion(String apiVersion) {
```
saicheems: Nothing changes anywhere else - this is just a convenience for Go since there is a common regex replacement for all API version names.

src/test/java/com/google/api/codegen/testdata/discoveries/go/go_storage.v1.json.baseline
saicheems: #634

tcoffee-google:
LGTM

vchudnov-g:
Non-ADC: doing in a follow-up PR is fine, as long as it happens before the PR to do the control flow switch to the Go MVVM path.

If compilecheck passes, LGTM.

PR #609, opened 2016-10-12
Fix resource rename in C# Bigquery discovery client library

tcoffee-google:


src/main/java/com/google/api/codegen/csharp/CSharpDiscoveryContext.java
```diff
@@ -526,4 +531,16 @@ public String typeName(Type parentType, Field field, String requestTypeName) {
     }
     throw new IllegalArgumentException("Unknown type kind: " + kind);
   }
+
+  // Handlers for Exceptional Inconsistencies
+  // ========================================
+
+  // used to handle inconsistency in page-streaming methods for Bigquery API
```
saicheems: s/used/Used & period at the end

saicheems:
lgtm

vchudnov-g:
Were you able to run compilecheck on this, or are you still having issues with C#?

Address Sai's comments, and then lgtm.

PR #611, opened 2016-10-12
Update Ruby names and discovery docs

tcoffee-google:


saicheems:
Hmm... will this discovery doc update now break other languages?

vchudnov-g:
It might break the baseline tests, though it shouldn't of itself break compilecheck in production. It looks like @tcoffee-google ran the baseline tests and updated as needed, so IIUC, this should work.

LGTM

tcoffee-google:
Yes, the baselines have been updated.

tcoffee-google:
Note this will not resolve Ruby compile check issues until https://github.com/google/google-api-ruby-client/pull/482 is submitted, assuming no discovery docs have been updated in the meantime.

PR #612, opened 2016-10-13
Made ruby generated files load grpc files once service is required

### LINKED ISSUE #610: gapic generated code should add 'require xxx_pb.rb'
Due to a restriction in grpc module loading (they need to be delayed until after the gapic module is loaded). This will cause the reference to the protobuf classes (such as those in pubsub_pb) is not available until the corresponding API service classes (such as publisher_api) are initialized. As a result, we have some order restriction in terms of how the API is called. The solution is that we should load the protobuf classes from the API service classes directly instead of counting on protobuf service classes (such as pubsub_services_pb)

landrito:
Resolves #610.

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/ImportTypeTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/ImportTypeTransformer.java
@@ -16,11 +16,19 @@ package com.google.api.codegen.transformer;
 
 import com.google.api.codegen.util.TypeAlias;
 import com.google.api.codegen.viewmodel.ImportTypeView;
+import com.google.api.tools.framework.model.Field;
 import com.google.api.tools.framework.model.Interface;
+import com.google.api.tools.framework.model.MessageType;
 import com.google.api.tools.framework.model.Method;
+import com.google.api.tools.framework.model.Oneof;
+import com.google.api.tools.framework.model.ProtoElement;
+
 import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Queue;
 import java.util.Set;
 import java.util.TreeSet;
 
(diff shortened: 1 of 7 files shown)


landrito:
cc @bjwatson

src/test/java/com/google/api/codegen/testdata/ruby_main_library.baseline
```diff
@@ -27,6 +27,8 @@ require "json"
 require "pathname"
 
 require "google/gax"
+require "google/example/library/v1/library_pb"
+require "google/tagger/v1/tagger_pb"
```
swcloud: Why on line 220 and 221, there are no full path?

src/main/java/com/google/api/codegen/transformer/ImportTypeTransformer.java
```diff
@@ -51,6 +59,39 @@
       fullNames.add(namer.getProtoFileImportFromService(targetInterface));
     }
 
+    return createImportFilesFromNames(fullNames);
+  }
+
+  public List<ImportTypeView> generateProtoFileImports(SurfaceTransformerContext context) {
```
swcloud: This seems to be way more complicated than what I thought. Why can't we simply use

service.getFile().getSimpleName().replace(".proto", "")

to get the required proto classes? I thought xxx_services_pb.rb has a 1:1 corresponding xxx_pb.rb

swcloud: Will the bfsHandler load all the proto classes in a BFS traversal? If so, it might not be necessary, since xxx_pb.rb (such as pubsub_pb.rb will load the proto classes it depends on)

jmuk: Right, BFS traversal is not necessary, the same list is fine. The code would be quite similar to `getServiceImports`.

swcloud: I had a PR #613 which is simpler.

PR #613, opened 2016-10-13
Fix ruby proto file import

swcloud:
Duplicated effort with #612 . But push anyway as reference.

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/ImportTypeTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/ImportTypeTransformer.java
@@ -25,6 +25,10 @@ import java.util.Set;
 import java.util.TreeSet;
 
 public class ImportTypeTransformer {
+  private enum ImportFileType {
+    SERVICE_FILE,
+    PROTO_FILE
+  }
 
   public List<ImportTypeView> generateImports(Map<String, TypeAlias> imports) {
     List<ImportTypeView> generatedImports = new ArrayList<>();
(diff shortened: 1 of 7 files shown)


swcloud:
Please review #612 first.

landrito:
LGTM

jmuk:
LGTM too, but I'd like to invite @garrettjonesgoogle for the sanity check.

jmuk:
Garrett: we want for Ruby to distinguish grpc service definition file (`foo_services_pb.rb`) and proto file (`foo_pb.rb`) and import them differently. I am wondering if Ruby is the first case here and want to make sure this design is the right approach.

src/main/java/com/google/api/codegen/transformer/ImportTypeTransformer.java
```diff
@@ -61,4 +74,11 @@
 
     return imports;
   }
+
+  private String getFileImports(
```
garrettjonesgoogle: This name is misleading because it is plural, but it only returns one import. Maybe call it `getFileImportFromService`.

garrettjonesgoogle:
I feel that it's a little hard to judge within the context of non-MVVM generation. I don't see anything inherently wrong with this approach, but I reserve the right to change my mind when this is converted to MVVM :-)

src/main/java/com/google/api/codegen/transformer/ImportTypeTransformer.java
swcloud: I like being rigorous. :)

Fixed.

garrettjonesgoogle:
LGTM

PR #614, opened 2016-10-13
Go: Properly handle new go_package format

pongad:
I'm currently working to add go_package option in appropriate places.
The format of the option is in form
  "path/to/pkg;localname"
which is different than the one we are currently using.

This commit lets us handle both.
The logic handling the old version can be deleted after the go_package
change lands in googleapis.


src/test/java/com/google/api/codegen/transformer/go/GoModelTypeNameConverterTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+public class GoModelTypeNameConverterTest {
+
+  private static final GoModelTypeNameConverter converter = new GoModelTypeNameConverter();
+
+  @Test
+  public void testGetTypeName() {
+    // Not in curated proto. Don't guess anything.
+    Truth.assertThat(
+            converter.getTypeName("github.com/someone/repo", "foo.bar", "Baz", false).getNickname())
```
garrettjonesgoogle: to make the test more readable (w.r.t. the booleans), could you make two "helper" variables, isPointerTrue and isPointerFalse, and pass them to getTypeName instead of true/false?

pongad:
@garrettjonesgoogle PTAl

garrettjonesgoogle:
LGTM

PR #619, opened 2016-10-13
Allow code generator to output all generated files into a jar file.

ethanbao:
Allow code generator to output all generated files into a jar file. The ToolUtil currently only allow writing output files to either a directory or a jar file.

Fix package path in runConfigGen task.


ethanbao:
I will squash the 3 commits into 1 during merging.

shinfan:
LGTM. Please let @bjwatson approve as well.

bjwatson:
LGTM, although I don't love the regression in test coverage. Seems like we need to refactor the `process()` method into smaller, testable pieces (unless it's going away after we finish MVVM refactoring).

garrettjonesgoogle:
@bjwatson -  `process()` isn't going away.

ethanbao:
Right. I notice the regression but fail to find a good way to add test coverage for this.

bjwatson:
@ethanbao Decomposing `process()` is the answer. For example, you can have a method that just builds `outputFiles` and you can test that. You can mock `ToolUtil.writeJar` and `ToolUtil.writeFiles`, and test the correct one gets called.

ethanbao:
PTAL. codecov/patch is still not happy. In order to hit the 81% target, I need to add more tests for CodeGeneratorApi, which is not the goal of this PR.

garrettjonesgoogle:
This PR is for bootstrapping/glue code, which in general I'm not very concerned about w.r.t. coverage. I think we should worry more about coverage in logic classes.

bjwatson:
LGTM.

The codecov target could be reach by also breaking out the `outputFiles` initialization code into its own method and unit testing it.

I agree that testing logic classes is most important. As a second order concern, testing the glue code is also important. The problem here is that `process()` is a big chunk of monolithic imperative code. It needs to be decomposed to be more testable, which I agree is outside the scope of this PR.

PR #621, opened 2016-10-14
Fix the pattern of param name conflicting with language keywords.

### LINKED ISSUE #618: wrap language keywords for resource name functions
Got a report. With Cloud Functions API (see https://github.com/googleapis/googleapis/blob/master/google/cloud/functions/v1beta2/functions_gapic.yaml#L22), it has `function` variable name.

The generated code looks like:

``` javascript
CloudFunctionsServiceApi.prototype.functionPath = function(project, location, function) {
   ...
```

`function` is a keyword in JS, the third argument needs to be renamed something else (like `function_`).

jmuk:
Fixes #618.

As is seen in the bug, a name pattern for cloud functions
contain 'function' name which becomes to be a parameter of
a function and therefore causes syntax error on JS.

Potentially the same problem could happen in many other languages.
This patch includes fixes on those potential problems.


src/main/java/com/google/api/codegen/util/php/PhpTypeTable.java
```diff
@@ -105,7 +105,7 @@ public String getAndSaveNicknameForInnerType(
   /**
    * A set of PHP keywords and built-ins. keywords: http://php.net/manual/en/reserved.keywords.php
    */
-  private static final ImmutableSet<String> KEYWORD_BUILT_IN_SET =
+  public static final ImmutableSet<String> KEYWORD_BUILT_IN_SET =
```
garrettjonesgoogle: Could you move all of the `KEYWORD_BUILT_IN_SET` constants to the corresponding `{Lang}NameFormatter` classes? It is a more appropriate location, and I don't think any of them are referenced within the corresponding `{Lang}TypeTable` classes.

src/main/java/com/google/api/codegen/util/php/PhpNameFormatter.java
```diff
@@ -50,22 +58,22 @@ public String varReference(Name name) {
 
   @Override
   public String publicMethodName(Name name) {
-    return name.toLowerCamel();
+    return wrapIfKeywordOrBuiltIn(name.toLowerCamel());
```
garrettjonesgoogle: Why didn't you do all of the methods in this class?

src/main/java/com/google/api/codegen/util/java/JavaNameFormatter.java
```diff
@@ -50,17 +58,17 @@ public String varReference(Name name) {
```
garrettjonesgoogle: Handling varReference is necessary too...

src/main/java/com/google/api/codegen/util/php/PhpTypeTable.java
jmuk: Done.

geigerj:
Python component LGTM

src/main/java/com/google/api/codegen/util/php/PhpNameFormatter.java
garrettjonesgoogle: @michaelbausor could you answer for PHP?

src/main/java/com/google/api/codegen/util/java/JavaNameFormatter.java
garrettjonesgoogle: Ahh, right.

src/main/java/com/google/api/codegen/util/php/PhpNameFormatter.java
```diff
@@ -17,15 +17,24 @@
...
 import com.google.api.codegen.util.NamePath;
+import com.google.common.collect.ImmutableSet;
 
 /**
  * The NameFormatter for PHP.
  */
 public class PhpNameFormatter implements NameFormatter {
 
+  private String wrapIfKeywordOrBuiltIn(String name) {
+    if (KEYWORD_BUILT_IN_SET.contains(name)) {
```
garrettjonesgoogle: So PHP is weird in that it treats any casing of a keyword as using that keyword. Examples: `empty`, `Empty`, `EMPTY`. So, you need to do a `toLower()` on the input `name`. Also, there are more functions in this class that need `wrapIfKeywordOrBuiltIn` applied.

src/main/java/com/google/api/codegen/util/nodejs/NodeJSNameFormatter.java
```diff
@@ -17,20 +17,29 @@
...
     return name.toUpperCamel();
   }
 
   @Override
   public String localVarName(Name name) {
-    return name.toLowerCamel();
+    return wrapIfKeywordOrBuiltIn(name.toLowerCamel());
   }
 
   @Override
```
landrito: Why not wrap `*FieldName`s, `*MethodName`s, or `StaticFunctionName`?

src/main/java/com/google/api/codegen/util/ruby/RubyNameFormatter.java
```diff
@@ -17,6 +17,7 @@
 import com.google.api.codegen.util.Name;
 import com.google.api.codegen.util.NameFormatter;
 import com.google.api.codegen.util.NamePath;
+import com.google.common.collect.ImmutableSet;
 
 /**
  * The NameFormatter for Ruby.
```
landrito: Why are we not wrapping here?

src/main/java/com/google/api/codegen/util/nodejs/NodeJSNameFormatter.java
jmuk: In NodeJS, the following code is syntactically correct.
```javascript
var foo = {};
foo.function = 'foo';
this.class = 'bar';
foo.prototype.return = function() {};
```

Therefore, field name, method name, etc wouldn't need that.
Static functions might need them.

src/main/java/com/google/api/codegen/util/ruby/RubyNameFormatter.java
jmuk: I was just skipping this file. Added (note that skipping private fields intentionally because those are `@foo` style IIRC).

jmuk:
What is the status of the review?
It seems the remaining thing is about PHP and waiting for the review from @michaelbausor ?

jmuk:
ping?

src/main/java/com/google/api/codegen/util/php/PhpNameFormatter.java
michaelbausor: This is correct - variable names and field names do not cause conflicts.

src/test/java/com/google/api/codegen/testdata/java_main_library.baseline
```diff
@@ -233,21 +247,45 @@ public class LibraryServiceApi implements AutoCloseable {
 
   /**
    * Parses the archive_path from the given fully-qualified path which
-   * represents a archivedBook resource.
+   * represents a archived_book resource.
```
michaelbausor: Nit: Why is this changing?

jmuk: This happens for the consistency with existing path templates -- it's "archived_book" in `formatArchivedBookName` already. Because it's in a comment, it would usually indicate something in "archived_book", not a variable name, so I think this would be okay.

michaelbausor: SGTM

PR #622, opened 2016-10-14
Concept art for awesome samplegen testing

saicheems:


saicheems:
@vchudnov-g 
This is what I was talking about - I've made a trivial discovery doc with a get method, used toolkit to generate a sample in Python, and made a mock library that mimics the structure of a real client library.

Now, we can run a sample and verify that it behaves as we expect. That means that we can tailor mocks for this made up API to assert types, check that functions are called in the order we expect, any number of things.

You can verify by checking this commit out and calling `python sample.py` in the `.../tmp` directory.

vchudnov-g:
I like the idea; thanks for prototyping. I do still think we should create the library for our synthetic API from real client library generators rather than using a mock. I have updated thoughts on how to do this cleanly. I'm capturing them so we can all discuss on Monday.

tcoffee-google:
If we want syntax and type verification for the entirety of all our samples in the dynamic languages, I've always felt the proper way to do it is to generate mock methods for the actual API calls, which can be spliced into the sample generator to create checkers corresponding to each sample. This will probably turn out to be even simpler than the existing "compile" check, to say nothing of completeness and maintainability.

The same approach can be applied to a fake API for testing purposes, but the real work here is in capturing the combinatorial space of discovery features that determine sample output. There are enough interacting components that this probably requires a fair degree of automation.

PR #624, opened 2016-10-14
Add oneof logic/validation for GAPIC config

### LINKED ISSUE #620: Language-independent one-of fixes
### What
- Do not mark oneofs as required or flattened in the GAPIC config generator.
- Warn in GAPIC codegen if a oneof is marked as required.
- Warn in GAPIC if multiple values of a oneof are flattened in a single output method.
### Why
- Any of the above scenarios can lead to invalid requests if multiple possible values of the oneof get set.

geigerj:
This change adds a few behaviors:
- The config generator will no longer mark oneof fields as required
  or flattened
- Flattened oneof fields generate a warning in codegen
- Specifying multiple oneof fields in a single flattening generates
  an error in codegen
- Required oneof fields generate an error in codegen

Fixes #620

### DIFF
--- a/src/main/java/com/google/api/codegen/config/FlatteningConfig.java
+++ b/src/main/java/com/google/api/codegen/config/FlatteningConfig.java
@@ -14,21 +14,20 @@
  */
 package com.google.api.codegen.config;
 
-import com.google.api.codegen.ConfigProto;
 import com.google.api.codegen.FlatteningConfigProto;
 import com.google.api.codegen.FlatteningGroupProto;
 import com.google.api.tools.framework.model.Diag;
 import com.google.api.tools.framework.model.DiagCollector;
 import com.google.api.tools.framework.model.Field;
 import com.google.api.tools.framework.model.Method;
+import com.google.api.tools.framework.model.Oneof;
 import com.google.api.tools.framework.model.SimpleLocation;
 import com.google.common.collect.ImmutableList;
-
+import java.util.HashSet;
+import java.util.Set;
 import javax.annotation.Nullable;
 
-/**
- * FlatteningConfig represents the flattening configuration for a method.
- */
+/** FlatteningConfig represents the flattening configuration for a method. */
 public class FlatteningConfig {
   private final ImmutableList<ImmutableList<Field>> flatteningGroups;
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/api/codegen/config/FlatteningConfig.java
```diff
@@ -41,12 +40,37 @@ public static FlatteningConfig createFlattening(
...
+                      "Multiple oneof fields flattened: method = %s, message type = %s, "
+                          + "oneof = %s, field = %s",
+                      method.getFullName(),
+                      method.getInputMessage().getFullName(),
+                      oneof.getName(),
+                      parameter));
+            } else {
+              seenOneofs.add(oneof);
+            }
+          }
```
bjwatson: This is a big chunk of code added to an already large method. It should be broken out into its own package private or protected method (so it can be tested).

src/main/java/com/google/api/codegen/config/FlatteningConfig.java
```diff
@@ -41,12 +40,37 @@ public static FlatteningConfig createFlattening(
...
+          Oneof oneof = parameterField.getOneof();
+          if (oneof != null) {
+            diagCollector.addDiag(
+                Diag.warning(
+                    SimpleLocation.TOPLEVEL,
+                    "Oneof field flattened: method = %s, message type = %s, oneof = %s, field = %s",
+                    method.getFullName(),
+                    method.getInputMessage().getFullName(),
+                    oneof.getName(),
+                    parameter));
```
bjwatson: Why be so permissive about the single flattened oneof parameter, if it's always an error for required parameters? This could be an opportunity to simplify the logic.

geigerj: There is difference between flattening and required/optional: it is a possible approach in languages that support overriding (and therefore flattening) to provide n methods if there are n possible values of the oneof; that is, if you have a request

```
FooRequest {
  oneof {
    A a = 1;
    B b = 2;
    C c = 3;
  }
}
```

You could potentially multiple overloaded methods

```
function foo(A a) { ... }
function foo(B b) { ... }
function foo(C c) { ... }
```

In fact, this is currently how `RunQuery` for datastore is currently configured: https://github.com/googleapis/googleapis/blob/master/google/datastore/v1/datastore_gapic.yaml#L52

I think this is a bad idea because it only works if no two oneof values have the same type. But it's not universally illegal as it would be to do this in a language that doesn't support overloading (and therefore uses the required parameters configuration).

geigerj:
I will investigate how we can test validation code in `toolkit`. I am not aware of any such tests currently existing, but I'm sure the functionality is possible by examination of the state of the `DiagCollector` in the test code.

src/main/java/com/google/api/codegen/config/FlatteningConfig.java
geigerj: Done, although I don't see this as a significant improvement -- the "chunk of code" is actually only four statements (line wrapping makes it look like more) and is highly dependent on the state of the parent method.

bjwatson:
@geigerj Let's bring this up to date in January and get this through review.

I've added @lukesneeringer to get his input. For context, oneof sets of parameters can be defined in protobuf, and this PR is adding some enforcement to our generated Python code to ensure that only one member of the oneof set is defined.

lukesneeringer:
@geigerj Is it safe to assume that this is dead? I think we did a separate implementation last month.

geigerj:
I think this should be redone from scratch; there are too many merge conflicts at this point. The separate implementation you refer to was in the Python GAPIC output. OTOH, this was about changing the tooling not to generate configuration -- and not to accept configuration -- that mark smultiple oneof fields as required or flattened.

PR #633, opened 2016-10-18
Change google-java-import-gradle-plugin to 0.4.1

pongad:
Many imports are rewritten accord to [Google style rule](https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing).


bjwatson:
@pongad This PR is going to be fragile with active development on the toolkit master branch. We should coordinate a time when everyone quiesces their activity, and then you can update with the latest from master and I can review.

IIUC, this is not urgent, correct?

pongad:
@bjwatson I thought it would be. I mostly wanted to float this out in case I miss anything major.
A possibility is Sydney Monday, US Sunday. We kind of get the automatic quiet(er) period.

pongad:
@bjwatson I just realized that I don't have your LGTM on this. If it looks good to you, we might consider merging it MTV morning. If not (or it would cause too much churn) we can do it next week as well.

bjwatson:
@pongad LGTM

PR #635, opened 2016-10-18
Add page-streaming method support to api sample.

shinfan:


garrettjonesgoogle:
I think the template might assume only flattened methods - try generating to check.

shinfan:
Update the template and fixed the format (which caused the travis fail).

Tested against the adwords service and output looks good.

garrettjonesgoogle:
LGTM

PR #637, opened 2016-10-19
Extend VisibilityConfig to private methods

pongad:
This commit extends VisibilityConfig to include private methods. The main motivation of these private methods is to let us create hand-written methods in their places. The private methods can be used by the hand-written ones to simplify work.

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/SurfaceTransformerContext.java
+++ b/src/main/java/com/google/api/codegen/transformer/SurfaceTransformerContext.java
@@ -149,10 +149,11 @@ public abstract class SurfaceTransformerContext {
     return methods;
   }
 
-  /** Returns a list of methods with samples, similar to getSupportedMethods,
-   *  but also filter out private methods.
+  /**
+   * Returns a list of methods with samples, similar to getSupportedMethods,
+   * but also filter out private methods.
    */
-  public List<Method> getMethodsWithSamples() {
+  public List<Method> getPublicMethods() {
     List<Method> methods = new ArrayList<>(getInterfaceConfig().getMethodConfigs().size());
     for (MethodConfig methodConfig : getInterfaceConfig().getMethodConfigs()) {
       Method method = methodConfig.getMethod();
--- a/src/main/java/com/google/api/codegen/transformer/SurfaceTransformerContext.java
+++ b/src/main/java/com/google/api/codegen/transformer/SurfaceTransformerContext.java
@@ -137,12 +137,27 @@ public abstract class SurfaceTransformerContext {
         getFeatureConfig());
   }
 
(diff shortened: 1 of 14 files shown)


src/test/java/com/google/api/codegen/testdata/go_main_library.baseline
```diff
@@ -606,6 +611,22 @@ func (c *Client) FindRelatedBooks(ctx context.Context, req *librarypb.FindRelate
...
+// addLabel adds a label to the entity.
+func (c *Client) addLabel(ctx context.Context, req *taggerpb.AddLabelRequest) (*taggerpb.AddLabelResponse, error) {
+    md, _ := metadata.FromContext(ctx)
+    ctx = metadata.NewContext(ctx, metadata.Join(md, c.metadata))
+    var resp *taggerpb.AddLabelResponse
+    err := gax.Invoke(ctx, func(ctx context.Context) error {
+        var err error
+        resp, err = c.labelerClient.addLabel(ctx, req)
+        return err
+    }, c.CallOptions.AddLabel...)
```
jba: I guess we have to make the CallOptions field public even though the method is private. Fine for now. We may have to revisit someday.

src/main/java/com/google/api/codegen/transformer/SurfaceTransformerContext.java
```diff
@@ -149,6 +149,21 @@ public MethodTransformerContext asMethodContext(Method method) {
     return methods;
   }
 
+  /** Returns a list of methods with samples, similar to getSupportedMethods,
```
garrettjonesgoogle: If you have a multi-line comment, I don't think the text is supposed to start on the first line

src/main/java/com/google/api/codegen/transformer/SurfaceTransformerContext.java
```diff
@@ -149,6 +149,21 @@ public MethodTransformerContext asMethodContext(Method method) {
...
+   *  but also filter out private methods.
+   */
+  public List<Method> getMethodsWithSamples() {
```
garrettjonesgoogle: It seems like "getPublicMethods" would be better - there is no check in the logic for having samples.

src/main/java/com/google/api/codegen/transformer/SurfaceTransformerContext.java
pongad: Done.

src/main/java/com/google/api/codegen/transformer/SurfaceTransformerContext.java
pongad: Good idea, done.

PR #638, opened 2016-10-19
Go: Use LRO client

pongad:
This change only affects Go. There are some baseline changes in other languages due to the added `GetBigBook` method in `library.proto`.

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/go/GoGapicSurfaceTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/go/GoGapicSurfaceTransformer.java
@@ -15,6 +15,7 @@
 package com.google.api.codegen.transformer.go;
 
 import com.google.api.codegen.config.ApiConfig;
+import com.google.api.codegen.config.InterfaceConfig;
 import com.google.api.codegen.config.MethodConfig;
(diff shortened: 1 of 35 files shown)


src/main/java/com/google/api/codegen/transformer/go/GoGapicSurfaceTransformer.java
```diff
@@ -247,9 +255,13 @@ static ModelTypeTable createTypeTable() {
 
   @VisibleForTesting
   List<StaticLangApiMethodView> generateApiMethods(
-      SurfaceTransformerContext context, List<Method> methods, List<String> pageStreamImports) {
+      SurfaceTransformerContext context,
+      List<Method> methods,
+      List<String> pageStreamImports,
```
jba: Maybe a more general map from enum to List<String>? I suspect there will be more of these.

pongad: This is a good point. I'll look into this.

pongad:
@jba @garrettjonesgoogle PTAL

jba:
LGTM.

garrettjonesgoogle:
LGTM

PR #642, opened 2016-10-19
Add support of `$RANDOM` token in smoke test config.

shinfan:
- Fixed an issue where the smoke test data was not init correctly.

### DIFF
--- a/src/main/java/com/google/api/codegen/metacode/InitFieldConfig.java
+++ b/src/main/java/com/google/api/codegen/metacode/InitFieldConfig.java
@@ -23,6 +23,8 @@ import javax.annotation.Nullable;
  */
 @AutoValue
 public abstract class InitFieldConfig {
+  private static String randomValueToken = "$RANDOM";
+
   public abstract String fieldPath();
 
   @Nullable
@@ -43,7 +45,7 @@ public abstract class InitFieldConfig {
     if (equalsParts.length > 2) {
       throw new IllegalArgumentException("Inconsistent: found multiple '=' characters");
     } else if (equalsParts.length == 2) {
-      value = equalsParts[1];
+      value = parseValueString(equalsParts[1]);
     }
 
     String[] fieldSpecs = equalsParts[0].split("[%]");
@@ -66,5 +68,10 @@ public abstract class InitFieldConfig {
 
   public boolean hasFormattedInitValue() {
     return entityName() != null && value() != null;
+  }
+
+  private static String parseValueString(String valueString) {
+    String randomValue = Integer.toString(Math.abs(valueString.hashCode()));
+    return valueString.replace(randomValueToken, randomValue);
   }
 }
--- a/src/main/java/com/google/api/codegen/metacode/InitFieldConfig.java
+++ b/src/main/java/com/google/api/codegen/metacode/InitFieldConfig.java
@@ -23,7 +23,7 @@ import javax.annotation.Nullable;
  */
 @AutoValue
 public abstract class InitFieldConfig {
-  private static String randomValueToken = "$RANDOM";
+  private static final String randomValueToken = "$RANDOM";
 
   public abstract String fieldPath();
 
@@ -45,7 +45,7 @@ public abstract class InitFieldConfig {
     if (equalsParts.length > 2) {
       throw new IllegalArgumentException("Inconsistent: found multiple '=' characters");
     } else if (equalsParts.length == 2) {
-      value = parseValueString(equalsParts[1]);
+      value = parseValueString(equalsParts[1], equalsParts[0]);
     }
 
     String[] fieldSpecs = equalsParts[0].split("[%]");
@@ -70,8 +70,8 @@ public abstract class InitFieldConfig {
     return entityName() != null && value() != null;
   }
 
-  private static String parseValueString(String valueString) {
-    String randomValue = Integer.toString(Math.abs(valueString.hashCode()));
+  private static String parseValueString(String valueString, String stringToHash) {
+    String randomValue = Integer.toString(Math.abs(stringToHash.hashCode()));
     return valueString.replace(randomValueToken, randomValue);
   }
 }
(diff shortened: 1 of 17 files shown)


src/main/java/com/google/api/codegen/metacode/InitFieldConfig.java
```diff
@@ -23,6 +23,8 @@
  */
 @AutoValue
 public abstract class InitFieldConfig {
+  private static String randomValueToken = "$RANDOM";
```
michaelbausor: Should this be final?

src/main/java/com/google/api/codegen/metacode/InitFieldConfig.java
```diff
@@ -67,4 +69,9 @@ public boolean isFormattedConfig() {
   public boolean hasFormattedInitValue() {
     return entityName() != null && value() != null;
   }
+
+  private static String parseValueString(String valueString) {
+    String randomValue = Integer.toString(Math.abs(valueString.hashCode()));
```
michaelbausor: This approach will produce the same "random" value for the same string. Is that what we want? The most common case will be just the string "$RANDOM", which is likely to occur often and will always produce the same value.

Is this a problem? Or is it more important for the value to be reproducible?

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -329,15 +317,27 @@ private InitValueView getInitValue(MethodTransformerContext context, InitCodeNod
...
-      List<String> varList, InitValueConfig initValueConfig) {
+      MethodTransformerContext context, Field field, InitValueConfig initValueConfig) {
     List<String> formatFunctionArgs = new ArrayList<>();
-    for (String entityName : varList) {
+
+    Set<String> entityNames;
+    boolean shouldUseCollectionConfigEntityNames =
+        !context.getFeatureConfig().useResourceNameFormatOption(field)
+            || initValueConfig.hasFormattingConfigInitialValues();
+    if (shouldUseCollectionConfigEntityNames) {
```
michaelbausor: I found where the "_id" suffix is coming from:
https://github.com/googleapis/toolkit/blob/master/src/test/java/com/google/api/codegen/testdata/library.proto#L30

This is a problem that will be fixed once the resource name configuration moves out of the proto and into the gapic config, although it was a mistake to allow this difference in the first place - my original PR should have checked for this inconsistency.

In this PR, which do you think is preferable - the changes you have here, or switching shelf_id->shelf in the proto, or both?

shinfan: As discussed offline, I will change the proto and revert the fix in the java code.

src/main/java/com/google/api/codegen/metacode/InitFieldConfig.java
shinfan: Good point. I changed the hash string so the $RANDOM value would be more unique. It is not a strict requirements that each $RANDOM value needs to be unique. But it should be unique enough so the data can be validated by tests.

shinfan:
@michaelbausor  PTAL

src/test/java/com/google/api/codegen/testdata/nodejs_doc_main_library.baseline
```diff
@@ -180,78 +180,78 @@ function LibraryServiceApi(gaxGrpc, grpcClients, opts) {
...
  * Returns a fully-qualified shelf resource name string.
- * @param {String} shelf
+ * @param {String} shelf_id
  * @returns {String}
  */
-LibraryServiceApi.prototype.shelfPath = function(shelf) {
+LibraryServiceApi.prototype.shelfPath = function(shelfId) {
   return SHELF_PATH_TEMPLATE.render({
-    shelf: shelf
+    shelf_id: shelfId
```
garrettjonesgoogle: It looks like there is a casing bug in node js... the doc doesn't match the real parameter.

src/main/java/com/google/api/codegen/metacode/InitFieldConfig.java
```diff
@@ -67,4 +69,9 @@ public boolean isFormattedConfig() {
   public boolean hasFormattedInitValue() {
     return entityName() != null && value() != null;
   }
+
+  private static String parseValueString(String valueString, String stringToHash) {
+    String randomValue = Integer.toString(Math.abs(stringToHash.hashCode()));
+    return valueString.replace(randomValueToken, randomValue);
```
garrettjonesgoogle: We are generating a random number every time, even if we hardly ever do the replacement. Maybe we should do a find first?

src/test/java/com/google/api/codegen/testdata/csharp_gapic_client_library.baseline
```diff
@@ -704,57 +704,57 @@ namespace Google.Example.Library.V1
...
         /// </list>
         /// </summary>
-        public static PathTemplate ShelfTemplate { get; } = new PathTemplate("shelves/{shelf}");
+        public static PathTemplate ShelfTemplate { get; } = new PathTemplate("shelves/{shelf_id}");
 
         /// <summary>
         /// Creates a shelf resource name from its component IDs.
         /// </summary>
-        /// <param name="shelfId">The shelf ID.</param>
+        /// <param name="shelfIdId">The shelfId ID.</param>
```
garrettjonesgoogle: CSharp is acting weird here.

shinfan: Issue filed: https://github.com/googleapis/toolkit/issues/644

src/main/java/com/google/api/codegen/metacode/InitFieldConfig.java
shinfan: Good point. Done.

garrettjonesgoogle:
LGTM

PR #645, opened 2016-10-20
Fix download links in Go discovery sample gen

tcoffee-google:

PR #649, opened 2016-10-20
Add dfareporting baseline test for discovery sample gen

tcoffee-google:
Covers Go multipart resource download link case in templates (#645).


vchudnov-g:
If all tests pass, LGTM.

PR #650, opened 2016-10-20
Add support of gRPC stremaing methods to NodeJS codegen.

jmuk:
updates #343


src/test/java/com/google/api/codegen/testdata/nodejs_doc_main_library.baseline
```diff
@@ -1030,6 +1041,89 @@ LibraryServiceApi.prototype.updateBookIndex = function(request, options, callbac
...
+ *
+ * @example
+ *
+ * var api = libraryV1.libraryServiceApi();
+ *
+ * var stream = api.discussBook().on('data', function(response) {
+ *     // doThingsWith(response);
+ * });
+ * // Write request objects.
+ * stream.write({});
```
landrito: This is just my personal preference; I think it would be more readable to our users to have request messages for writable streams be instantiated after the method call to make it clearer that the stream is a writable stream.

```js
var api = libraryV1.libraryServiceApi();

var stream = api.discussBook().on('data', function(response) {
     // doThingsWith(response);
 });

// Declare the request messages after the method is called to
// make it stand out that this stream is writable.
var comment = "";
 
// Write request objects.
stream.write({comment: comment});
```

src/main/resources/com/google/api/codegen/nodejs/method_sample.snip
```diff
@@ -21,6 +35,35 @@
     @end
 @end
 
+@private methodCallReturningStream(apiMethod, elementVarName)
+    {@methodCallSampleCode(apiMethod)}.on('data', function({@elementVarName}) {
```
landrito: nit: s/`elementVarName`/`elementName`

`var` seems unnecessary.

src/main/resources/com/google/api/codegen/nodejs/method_sample.snip
```diff
@@ -11,6 +11,20 @@
         {@initializeRequestObject(apiMethod)}
 
     @end
+    @switch apiMethod.grpcStreamingType.toString
```
swcloud: I think this should be under optionalArrayMethodSampleCode, based on the logic in toolkit, where streaming type is assigned as part of OptionalArrayMethodView

src/test/java/com/google/api/codegen/testdata/nodejs_doc_main_library.baseline
jmuk: done

PR #652, opened 2016-10-20
Add support of `enabled_artifacts` flag

shinfan:
- Add` GapicGeneratorConfig` class which stores generator related config
    - generator ID
    - if surface/test generator is enabled

Usage:
- Generate surface only:
`./gradlew runVGen -Pclargs={YOUR ORIGINAL ARGS},--enabled_artifacts=surface`
- Generate test only:
`./gradlew runVGen -Pclargs={YOUR ORIGINAL ARGS},--enabled_artifacts=test`

This PR fixes https://github.com/googleapis/toolkit/issues/646


src/main/java/com/google/api/codegen/CodeGeneratorApi.java
```diff
@@ -61,9 +59,14 @@
...
           ImmutableList.<String>of());
 
-  /**
-   * Constructs a code generator api based on given options.
-   */
+  public static final Option<Boolean> DISABLE_TEST_GEN =
+      ToolOptions.createOption(
+          Boolean.class,
+          "disable_test_gen",
+          "Disables test suit generation of the code generator.",
```
garrettjonesgoogle: s/suit/suite/

src/main/java/com/google/api/codegen/CodeGeneratorApi.java
```diff
@@ -61,9 +59,14 @@
           "The list of YAML configuration files for the code generator.",
           ImmutableList.<String>of());
 
-  /**
-   * Constructs a code generator api based on given options.
-   */
+  public static final Option<Boolean> DISABLE_TEST_GEN =
```
garrettjonesgoogle: I much prefer a positively-worded term. Maybe instead we could do "filter_artifacts", and support the values "surface" and "test"; if nothing is given, generate everything; for users who only want the surface, they pass in "--filter_artifacts=surface".

src/main/java/com/google/api/codegen/CodeGeneratorApi.java
shinfan: Obsolete.

src/main/java/com/google/api/codegen/CodeGeneratorApi.java
shinfan: I feel filter_artifacts can be ambiguous since I don't know if the listed artifacts are going to be enabled or disabled. I picked `enabled_artifacts` instead.

shinfan:
@garrettjonesgoogle PTAL

src/main/java/com/google/api/codegen/CodeGeneratorTool.java
```diff
@@ -66,6 +64,16 @@ public static void main(String[] args) throws Exception {
             .hasArg()
             .argName("OUTPUT-DIRECTORY")
             .build());
+    options.addOption(
+        Option.builder()
+            .longOpt("enabled_artifacts")
+            .desc(
+                "Optional. Artifacts enabled for the generator. "
+                    + "Currently support 'surface' and 'test'.")
```
garrettjonesgoogle: s/support/supports/

PR #657, opened 2016-10-20
WIP: Resource name support

michaelbausor:
In this PR:
- Add resource_name_treatment field to config proto
- Add FieldConfig object, and use it instead of Field in many places. This is because it is not sufficient to know which field you have - it is necessary to know whether it has a resource name and if resource names are enabled, which is per flattening group
- Changed some config classes to AutoValue

This PR is still missing:
- Documentation
- Support for list methods
- Support for resource names on non-request messages


michaelbausor:
Replaced by https://github.com/googleapis/toolkit/pull/664

PR #664, opened 2016-10-21
Add field config

michaelbausor:
This is a modified PR which should be able to be merged without requiring more changes or churning of baseline files.

Supersedes  https://github.com/googleapis/toolkit/pull/657

Changes:
- Add FieldConfig, and use it in place of Field in many places
- Add FlatteningConfig field to MethodTransformerContext, so that a list of fields does not need to be passed around between many methods
- Add flattening parameters to smoke test config, so that a particular flattening config may be specified


src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -132,6 +132,9 @@ message SmokeTestConfigProto {
 
   // A list describing the name and the value of the init fields.
   repeated string init_fields = 2;
+
+  // A list of the flattened parameters in the method to test.
+  repeated string flattened_parameters = 3;
```
garrettjonesgoogle: This seems like it's redundant with init_fields.

michaelbausor:
The high level goal is to prepare for introducing resource names. The main change preparing for this is to introduce the FieldConfig class, and to use a FieldConfig in place of a field in most places. This is necessary because once we introduce the power to specify whether an individual field should use a resource name type, a path template validation, or nothing, it is no longer sufficient to have simply the Field object.

This change also introduces the ResourceNameTreatment enum, and places it in the FieldConfig. However, at this stage the value in inferred from the resource name and/or field_name_pattern.

Uses of the ResourceNameUtil class have been refactored into the config construction, so that the proto extension can be replaced with GAPIC config.

A parallel refactoring is to add a flattening config field to the MethodContextTransformer. This means that instead of passing around a list of fields or field configs, those field configs are accessible from the context.

src/main/proto/com/google/api/codegen/config.proto
michaelbausor: The init_fields are used to specify the initialization, using a dotted path. At the moment, they also implicitly define which flattened method to use, because they are used as the arguments in the method call. If a flattened parameter were to not be initialized, the params in the flattened method would be incorrect.

The goal of the flattened_parameters was to specify a list which must match a flattening config that exists in the GAPIC config. This list can be used to identify the flattened method and its config.

It would be possible to infer this list from the init_fields, assuming that all fields will always be initialized - I thought it would be clearer to have a separate list.

garrettjonesgoogle: As discussed offline:

* In-language (e.g. Java), methods are matched up by parameter types, not parameter names.
* People may be surprised about seeing this either name-based or type-based matching specified in config.
* It will probably be easier for people to work with optionally giving names to flattening groups, and referencing that name here.

michaelbausor: Updated to use optional flattening name config instead of parameter list

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
```diff
@@ -280,25 +266,29 @@ private void setListMethodFields(
...
+    FieldConfig resourceFieldConfig = pageStreaming.getResourcesFieldConfig();
+    Field resourceField = resourceFieldConfig.getField();
 
     String resourceTypeName;
 
-    if (context.getFeatureConfig().useResourceNameFormatOption(resourceField)) {
-      String resourceShortName = ResourceNameUtil.getResourceName(resourceField);
+    if (context.getFeatureConfig().useResourceNameFormatOption(resourceFieldConfig)) {
+      String resourceClassName =
+          namer.getResourceTypeClassName(resourceFieldConfig.getEntityName());
```
garrettjonesgoogle: `Type` and `Class` are redundant - I would recommend `getResourceTypeName`.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
```diff
@@ -314,14 +354,15 @@ private GapicSurfaceTestCaseView createTestCaseView(
...
+    if (methodContext.getFeatureConfig().useResourceNameFormatOption(resourcesFieldConfig)) {
+      String resourceClassName =
+          namer.getResourceTypeClassName(resourcesFieldConfig.getEntityName());
+      Name resourceName = namer.getResourceTypeName(resourcesFieldConfig.getEntityName());
       resourceTypeName =
           methodContext
               .getTypeTable()
               .getAndSaveNicknameForTypedResourceName(
-                  resourcesField, resourcesField.getType().makeOptional(), resourceName);
+                  resourcesField, resourcesField.getType().makeOptional(), resourceClassName);
```
garrettjonesgoogle: I think I would prefer to just pass the entity into this method so that we don't need the dual getResourceTypeName/getResourceTypeClassName functions. I would prefer to have only getResourceTypeName which returns the class name.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
```diff
@@ -396,7 +437,7 @@ private InitCodeContext createResponseInitCodeContext(
     if (context.getMethodConfig().isPageStreaming()) {
       // Initialize one resource element if it is page-streaming.
       PageStreamingConfig config = context.getMethodConfig().getPageStreaming();
-      String resourceFieldName = config.getResourcesField().getSimpleName();
+      String resourceFieldName = config.getResourcesFieldConfig().getField().getSimpleName();
```
garrettjonesgoogle: maybe add a `getResourcesFieldName()` method to PageStreamingConfig to make this simpler (I have seen this in a couple places now)

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTransformer.java
```diff
@@ -429,29 +433,33 @@ public SettingsDocView generateSettingsDoc(
...
 
       if (methodConfig.isPageStreaming()) {
         if (methodConfig.isFlattening()) {
-          for (ImmutableList<Field> fields : methodConfig.getFlattening().getFlatteningGroups()) {
-            apiMethods.add(
-                apiMethodTransformer.generatePagedFlattenedMethod(methodContext, fields));
+          for (FlatteningConfig flatteningGroup : methodConfig.getFlatteningConfigs()) {
+            MethodTransformerContext methodContext =
+                context.asFlattenedMethodContext(method, flatteningGroup);
+            apiMethods.add(apiMethodTransformer.generatePagedFlattenedMethod(methodContext));
```
garrettjonesgoogle: Since you have `requestMethodContext` in this function, I think using `flattenedMethodContext` would be clearer here than `methodContext`.

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -132,6 +132,9 @@ message SmokeTestConfigProto {
 
   // A list describing the name and the value of the init fields.
   repeated string init_fields = 2;
+
+  // The name of the flattening config to use.
+  string flattening_name = 3;
```
garrettjonesgoogle: maybe `flattening_group_name` would be clearer here.

src/test/java/com/google/api/codegen/testdata/library_gapic.yaml
```diff
@@ -177,6 +178,7 @@ interfaces:
       groups:
       - parameters:
         - name
+        flattening_name: get_book
```
garrettjonesgoogle: having `flattening_name` first might be more natural. Also, calling this group `get_book_1` might be a more sustainable pattern.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
michaelbausor: Kept getResourceTypeName which returns the Name object, removed getResourceTypeClassName and added a getAndSave method to Surfacenamer that accepts the entityName.

I think we need to keep getResourceTypeName returning the Name object, not the result of className(...), because after calling className() we don't know the casing.

src/main/proto/com/google/api/codegen/config.proto
michaelbausor: Done

src/test/java/com/google/api/codegen/testdata/library_gapic.yaml
michaelbausor: Done

src/main/java/com/google/api/codegen/config/MethodConfig.java
```diff
@@ -204,37 +201,26 @@ public static MethodConfig createMethodConfig(
...
+            FieldConfig.createFieldConfig(
+                field,
+                ResourceNameTreatment.VALIDATE,
+                fieldNamePatterns.get(field.getSimpleName()));
+      } else {
+        fieldConfig = FieldConfig.createDefaultFieldConfig(field);
+      }
+      optionalFieldConfigsBuilder.add(fieldConfig);
+    }
+    List<FieldConfig> optionalFieldConfigs = optionalFieldConfigsBuilder.build();
```
garrettjonesgoogle: pull the new code above into its own function.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTransformer.java
```diff
@@ -438,9 +438,10 @@ public SettingsDocView generateSettingsDoc(
       if (methodConfig.isPageStreaming()) {
         if (methodConfig.isFlattening()) {
           for (FlatteningConfig flatteningGroup : methodConfig.getFlatteningConfigs()) {
-            MethodTransformerContext methodContext =
+            MethodTransformerContext flattenedMethodContext =
```
garrettjonesgoogle: There is another case like this below.

src/main/java/com/google/api/codegen/config/MethodConfig.java
michaelbausor: Done

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTransformer.java
michaelbausor: Done

michaelbausor:
PTAL

garrettjonesgoogle:
LGTM

PR #665, opened 2016-10-21
Fix request body constructors in PHP discovery sample gen

tcoffee-google:


vchudnov-g:
This wasn't working before? When did this breakage happen? It seems like something basic.

The change looks fine to me, but let @shinfan and @saicheems weigh in.

tcoffee-google:
It's been there since request bodies were added to PHP sample gen: 
https://github.com/googleapis/toolkit/commit/ab244b982b5e9af14114d90f9b51859c7bc4dc5b%23diff-92164f3905edf4c2776a8110e7b9d31d&sa=D&ust=1477070683861000&usg=AFQjCNHd1eACWJKKK2RB2RbQjAaZzWMRXQ

Seems to me like another example of the incompleteness of our current compile-check approach.

shinfan:
LGTM

Thanks for fixing this. I think one of the limitations of our current compile-check (at lease for PHP) is that it only checks for our expected issues. And in PHP this did not cause an error since you can always [pass more parameters than the function expects](http://stackoverflow.com/questions/6051309/can-you-pass-more-parameters-than-a-function-expects) (those extra parameters will be ignored).

saicheems:
lgtm

PR #667, opened 2016-10-23
Implement grpc streaming for Ruby

swcloud:
#343


src/main/resources/com/google/api/codegen/ruby/main.snip
```diff
@@ -378,44 +378,57 @@
...
+              @if optionalParams
+                {@requiredParameterNames(requiredParams)},
+                {@optionalParameterValues(optionalParams)},
+              @else
+                {@requiredParameterNames(requiredParams)},
+              @end
             @else
-              {@requiredParameterNames(requiredParams)},
+              {@optionalParameterValues(optionalParams)},
             @end
```
landrito: I don't think you edited this but it might be good to swap this logic to be a little bit cleaner.

ie 
Swap
```snip
      def {@methodName} @\
          @if requiredParams
            @if optionalParams
              {@requiredParameterNames(requiredParams)},
              {@optionalParameterValues(optionalParams)},
            @else
              {@requiredParameterNames(requiredParams)},
            @end
          @else
            {@optionalParameterValues(optionalParams)},
          @end
          options: nil
```
for this.
```snip
      def {@methodName} @\
          @if requiredParams
            {@requiredParameterNames(requiredParams)},
          @end
          @if optionalParams
            {@optionalParameterValues(optionalParams)},
          @end
          options: nil
```

src/main/resources/com/google/api/codegen/ruby/main.snip
```diff
@@ -378,44 +378,57 @@
        sample = comments(util.getDocLines(sampleCode(methodView)))
     @if documentation
       {@documentation}
+      @if method.getRequestStreaming()
+        @#
+        @# EXPERIMENTAL: This method interface might change in the future.
```
swcloud: Done

src/main/resources/com/google/api/codegen/ruby/main.snip
swcloud: Done

src/test/java/com/google/api/codegen/testdata/nodejs_doc_main_library.baseline
```diff
@@ -1159,8 +1159,11 @@ LibraryServiceApi.prototype.streamBooks = function(request, options) {
  *     // doThingsWith(response);
  * });
  * var name = '';
+ * var request = {
+ *     name : name
+ * };
  * // Write request objects.
- * stream.write({name: name});
+ * stream.write({request: request});
```
jmuk: This does not look correct. Please update nodejs method_sample.snip or somewhere else to keep the previous result.

src/main/resources/com/google/api/codegen/nodejs/method_sample.snip
```diff
@@ -45,7 +45,7 @@
 
 @private sampleWriteStreamingRequest(apiMethod)
     // Write request objects.
-    stream.write({@sampleMethodCallArgList(apiMethod)});
+    stream.write(request);
```
jmuk: Is there any way to obtain the name of this from `apiMethod` instead of having `request` always?

jmuk: It seems this is fine for now, at least this `request` variable name already appears twice in `nodejs/method_sample.snip`.  I'm wondering if this could conflict with other names in the sample potentially, but that might be a different topic.

PR #668, opened 2016-10-24
Go: Properly capitalize GRPC client's method names

pongad:
When generating private methods, the generated client's method names
are lower-cased.
However, we mistakenly also lower-case the GRPC client's method names.
This commit fixes this problem.


garrettjonesgoogle:
LGTM

PR #672, opened 2016-10-24
Add Go overrides files

saicheems:
Also update pubsub overrides for Node.js

### DIFF
--- a/src/test/java/com/google/api/codegen/testdata/discoveries/nodejs/nodejs_pubsub.v1.json.baseline
+++ b/src/test/java/com/google/api/codegen/testdata/discoveries/nodejs/nodejs_pubsub.v1.json.baseline
@@ -18,7 +18,7 @@ authorize(function(authClient) {
   var request = {
     // REQUIRED: The resource for which the policy is being requested. `resource` is usually specified as
     // a path. For example, a Project resource is specified as `projects/{project}`.
-    resource: '',  // TODO: Update placeholder value.
+    resource_: '',  // TODO: Update placeholder value.
     // ex: 'projects/my-project/snapshots/my-snapshot'
 
     auth: authClient
@@ -67,7 +67,7 @@ authorize(function(authClient) {
   var request = {
     // REQUIRED: The resource for which the policy is being specified. `resource` is usually specified as
     // a path. For example, a Project resource is specified as `projects/{project}`.
-    resource: '',  // TODO: Update placeholder value.
+    resource_: '',  // TODO: Update placeholder value.
     // ex: 'projects/my-project/snapshots/my-snapshot'
 
     resource: {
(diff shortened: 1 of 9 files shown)


saicheems:
This is actually wrong - gimme a bit to fix it

saicheems:
Updated

Btw, it strikes me that the Go client library is wrong for these cases - these methods are page streaming, but the response doesn't contain the field `nextPageToken`. Java seems to handle these APIs correctly, but since there's not `nextPageToken` field (only `pageToken`), Go doesn't generate the necessary boilerplate for page streaming. I wonder if this is something that we should file against `google-api-go-client`.

saicheems:
ping

src/test/java/com/google/api/codegen/testdata/discoveries/nodejs/nodejs_pubsub.v1.json.baseline
```diff
@@ -67,7 +67,7 @@ authorize(function(authClient) {
   var request = {
     // REQUIRED: The resource for which the policy is being specified. `resource` is usually specified as
     // a path. For example, a Project resource is specified as `projects/{project}`.
-    resource: '',  // TODO: Update placeholder value.
+    resource_: '',  // TODO: Update placeholder value.
```
saicheems: This is what the client library expects, cloudresourcemanager has a field `resource`, and the Node.js library maps requestBody to the `resource` parameter. To resolve this conflict, the library owners have decided to rename any parameters called `resource` to `resource_`.

So, this is a deliberate design choice in the client libraries, so this is correct.

src/test/java/com/google/api/codegen/testdata/discoveries/nodejs/nodejs_pubsub.v1.json.baseline
```diff
@@ -18,7 +18,7 @@ authorize(function(authClient) {
   var request = {
     // REQUIRED: The resource for which the policy is being requested. `resource` is usually specified as
     // a path. For example, a Project resource is specified as `projects/{project}`.
-    resource: '',  // TODO: Update placeholder value.
+    resource_: '',  // TODO: Update placeholder value.
```
saicheems: it does:
https://github.com/google/google-api-nodejs-client/blob/master/apis/pubsub/v1.js#L1762

src/test/java/com/google/api/codegen/testdata/discoveries/nodejs/nodejs_pubsub.v1.json.baseline
saicheems: https://github.com/google/google-api-nodejs-client/blob/master/apis/pubsub/v1.js#L1726

vchudnov-g: OK

vchudnov-g:
If compile checks pass, LGTM.

Do file that issue and maybe link to it in this PR as well.

saicheems:
https://github.com/google/google-api-go-client/issues/176

PR #674, opened 2016-10-25
Unify imports using ImportTypeView

pongad:
Previously Go GAPIC and sample gen use plain String to render imports.
However, many transformers uses ImportTypeView.
Consequently, using ImportTypeView everywhere should reduce our code
maintenance, even if doing so may at first feel more complex.


garrettjonesgoogle:
I was going to support Go in https://github.com/googleapis/toolkit/pull/676 until I realized that I need the change in this PR. Let's wait until 676 is merged, then rework this PR on top of it.

pongad:
@garrettjonesgoogle PTAL

garrettjonesgoogle:
LGTM

PR #675, opened 2016-10-25
Use hash with delete_if for all request object constructor

swcloud:
Resolve [(https://github.com/googleapis/gax-ruby/issues/45)]

### DIFF
--- a/src/main/resources/com/google/api/codegen/ruby/main.snip
+++ b/src/main/resources/com/google/api/codegen/ruby/main.snip
@@ -397,18 +397,16 @@
             {@optionalParameterValues(optionalParams)},
           @end
           options: nil
-        @if requiredParams
-          req = {@inputTypeName}.new(
-            {@namedParameters(requiredParams)}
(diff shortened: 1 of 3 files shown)


src/main/resources/com/google/api/codegen/ruby/main.snip
```diff
@@ -397,18 +397,16 @@
...
-        @else
-          req = {@inputTypeName}.new
-        @end
-        @join field : optionalParams
-          @let paramName = context.wrapIfKeywordOrBuiltIn(field.getSimpleName())
-            req.{@field.getSimpleName()} = {@paramName} unless {@paramName}.nil?
-          @end
-        @end
+        req = {@inputTypeName}.new(
+          {
```
jmuk: nit: personally prefer this to be in the previous line. i.e.
```ruby
req = {@inputTypeName}).new({
  ...
}.delete_if { |_, v| v.nil? })
```

PR #676, opened 2016-10-25
Loading license data from files in Java and C#

garrettjonesgoogle:
Also:

* Having SurfaceNamer store packageName
* Restructuring top-level views for Java and C# so that they are better
  normalized (file-level information is separate from class information)

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicClientTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicClientTransformer.java
@@ -24,7 +24,7 @@ import com.google.api.codegen.gapic.GapicCodePathMapper;
 import com.google.api.codegen.transformer.ApiCallableTransformer;
 import com.google.api.codegen.transformer.ApiMethodTransformer;
 import com.google.api.codegen.transformer.BundlingTransformer;
-import com.google.api.codegen.transformer.ImportTypeTransformer;
+import com.google.api.codegen.transformer.FileHeaderTransformer;
 import com.google.api.codegen.transformer.MethodTransformerContext;
 import com.google.api.codegen.transformer.ModelToViewTransformer;
 import com.google.api.codegen.transformer.ModelTypeTable;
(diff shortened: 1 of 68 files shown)


src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTransformer.java
```diff
@@ -68,7 +72,7 @@
   private ApiMethodTransformer apiMethodTransformer;
   private PageStreamingTransformer pageStreamingTransformer;
   private BundlingTransformer bundlingTransformer;
-  private ImportTypeTransformer importTypeTransformer;
+  private FileHeaderTransformer fileHeaderTransformer;
   private RetryDefinitionsTransformer retryDefinitionsTransformer;
```
pongad: A general question: Should these be turned into static finals to help signify that they don't hold state?

src/main/java/com/google/api/codegen/transformer/go/GoGapicSurfaceTransformer.java
```diff
@@ -212,17 +214,22 @@ private PackageInfoView generatePackageInfo(
...
-        .importPath(apiConfig.getPackageName())
-        .packageName(namer.getLocalPackageName())
-        .serviceDocs(Collections.<ServiceDocView>emptyList())
-        // TODO(pongad): GoContextCommon should be deleted after we convert Go discovery to MVMM.
-        .packageDoc(new GoContextCommon().getWrappedCommentLines(doc))
-        .domainLayerLocation(apiConfig.getDomainLayerLocation())
-        .build();
+    PackageInfoView.Builder packageInfo = PackageInfoView.newBuilder();
+
+    packageInfo.templateFileName(DOC_TEMPLATE_FILENAME);
```
pongad: Nothing wrong with this, but wondering why you're against using the chained calls?

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTransformer.java
garrettjonesgoogle: Even static final objects can have state. Making them at least `final` makes sense though. Maybe in a separate PR?

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicClientTransformer.java
```diff
@@ -155,23 +170,15 @@ private StaticLangXApiView generateXApi(SurfaceTransformerContext context) {
...
   private boolean methodTypeHasImpl(ApiMethodType type) {
     return type != ApiMethodType.FlattenedAsyncCancellationTokenMethod;
   }
 
-  private StaticLangXSettingsView generateXSettings(SurfaceTransformerContext context) {
-    StaticLangXSettingsView.Builder xsettingsClass = StaticLangXSettingsView.newBuilder();
-    xsettingsClass.templateFileName(""); // Unused in C#
-    xsettingsClass.packageName(context.getApiConfig().getPackageName());
+  private StaticLangSettingsView generateSettingsClass(SurfaceTransformerContext context) {
+    StaticLangSettingsView.Builder xsettingsClass = StaticLangSettingsView.newBuilder();
```
chrisdunelm: Remove x from variable name?

michaelbausor:
LGTM

PR #677, opened 2016-10-25
Add gapic resource name support

michaelbausor:
This PR:
- Removes resource_name_format.proto
- Adds GAPIC configuration for resource names and resource name treatment

This PR does not:
- Support Oneofs
- Support "invalid" resource name types


michaelbausor:
Updated to test parameter_resource_name_treatment setting, and increase code coverage.

A question regarding generation of samples - if the method-level resource name treatment setting is set to NONE, VALIDATE, or unset, but a resource name has been specified and generated for a field, what behaviour would we like:
1. Don't use the resource name in the sample (current behaviour)
2. Use the resource name, and pass it to the flattened method via the toString() method

src/test/java/com/google/api/codegen/testdata/library_gapic.yaml
```diff
@@ -421,8 +439,58 @@ interfaces:
...
+    resource_name_treatment: STATIC_TYPES
     retry_codes_name: non_idempotent
     retry_params_name: default
     field_name_patterns:
       name: book
     timeout_millis: 60000
+resource_name_generation:
+- message_name: Book
+  field_resource_types:
+    name: book
```
michaelbausor: Done

garrettjonesgoogle:
I like the idea of (2), in order to encourage usage of the resource name types.

garrettjonesgoogle:
LGTM.

I feel like things are getting very complicated. That is a discussion for another day though, I suppose.

PR #679, opened 2016-10-26
Fix the missing comma between required and optional parameters

swcloud:
When using hash with the grpc message object constructor, there was a missing ',' between required and optional parameters.


swcloud:
I guess I can try to join required and optional params, but that does not seem to save much.

PR #681, opened 2016-10-26
Support for IAM in pubsub

### LINKED ISSUE #680: Snippet generation for list methods includes cancellation token overloads
We don't include a cancellation token overload in our generated *actual* code for list methods, but the generated snippet refers to this phantom method. Examples of errors from snippet consumption:

```text
PublisherClientSnippets.g.cs:125: Member ID 'ListTopicsAsync(string,string,int?,CancellationToken)' matches no members.
PublisherClientSnippets.g.cs:191: Member ID 'ListTopicSubscriptionsAsync(string,string,int?,CancellationToken)' matches no members.
SubscriberClientSnippets.g.cs:92: Member ID 'ListSubscriptionsAsync(string,string,int?,CancellationToken)' matches no members.
```

chrisdunelm:
And fixes #680.
And corrects some missing imports.
And prefixes the snippet class/file name with 'Generated'.

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicClientTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicClientTransformer.java
@@ -39,6 +39,7 @@ import com.google.api.codegen.viewmodel.ApiCallSettingsView;
 import com.google.api.codegen.viewmodel.ApiCallableType;
 import com.google.api.codegen.viewmodel.ApiCallableView;
 import com.google.api.codegen.viewmodel.ApiMethodType;
+import com.google.api.codegen.viewmodel.ReroutedGrpcView;
 import com.google.api.codegen.viewmodel.SettingsDocView;
 import com.google.api.codegen.viewmodel.StaticLangApiMethodView;
 import com.google.api.codegen.viewmodel.StaticLangXApiView;
@@ -53,7 +54,9 @@ import com.google.common.collect.ImmutableList;
 import java.io.File;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 public class CSharpGapicClientTransformer implements ModelToViewTransformer {
 
@@ -154,6 +157,7 @@ public class CSharpGapicClientTransformer implements ModelToViewTransformer {
     }
     xapiClass.apiMethodsImpl(methodsImpl);
     xapiClass.hasDefaultInstance(context.getInterfaceConfig().hasDefaultInstance());
+    xapiClass.reroutedGrpcClients(generateReroutedGrpcView(context));
 
     // must be done as the last step to catch all imports
     xapiClass.imports(importTypeTransformer.generateImports(context.getTypeTable().getImports()));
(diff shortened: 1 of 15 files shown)


src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicClientTransformer.java
```diff
@@ -218,6 +222,26 @@ private StaticLangXSettingsView generateXSettings(SurfaceTransformerContext cont
     return settingsMembers;
   }
 
+  private List<ReroutedGrpcView> generateReroutedGrpcView(SurfaceTransformerContext context) {
+    SurfaceNamer namer = context.getNamer();
+    ModelTypeTable typeTable = context.getTypeTable();
+    Set<ReroutedGrpcView> rerouteds = new HashSet<>();
```
jskeet: reroutedViews?

jskeet: And perhaps use a LinkedHashSet so the iteration order is predictable? We wouldn't want generating the code twice to give different results. (I'm guessing the order is used later - I haven't checked yet.)

src/main/java/com/google/api/codegen/transformer/csharp/CSharpModelTypeNameConverter.java
```diff
@@ -127,6 +130,14 @@ public TypeName getTypeName(ProtoElement elem) {
       prefix = parentEl.getSimpleName() + ".Types." + prefix;
       parentEl = parentEl.getParent();
     }
+    if (parentEl instanceof ProtoFile) {
```
jskeet: Is this part of the IAM stuff, or fixing imports?

src/test/java/com/google/api/codegen/testdata/csharp_gapic_snippets_library.baseline
```diff
@@ -1,4 +1,4 @@
-============== file: Google/Example/Library/V1/LibraryServiceClientSnippets.g.cs ==============
+============== file: Google/Example/Library/V1/GeneratedLibraryServiceClientSnippets.g.cs ==============
```
jskeet: Can we keep the filename as LibraryServiceClientSnippets.g.cs, but change the class name?
That's the way I've got it now - it's a little odd, but it gives prominence to the manual code, while still making the generated code available easily. Not a problem if it's hard though.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicClientTransformer.java
chrisdunelm: Good point, changed to LinkedHashSet. And renamed the local.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpModelTypeNameConverter.java
chrisdunelm: Fixing imports. I've added a line to the PR description.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_snippets_library.baseline
chrisdunelm: Done

chrisdunelm:
PTAL

chrisdunelm:
Really must learn to run tests before committing...

PR #689, opened 2016-10-27
Add collection oneof

michaelbausor:
This PR:
- Adds support for oneof collections
- Adds a collection config setting to the top level of config proto (ApiConfig), while still supporting the interface-level collection configs
- Adds a ResourceNameType enum to distinguish Single, Oneof, Invalid, Any cases of resource name configuration
- Add test for oneof


src/test/java/com/google/api/codegen/testdata/library.proto
```diff
@@ -109,7 +109,12 @@ service LibraryService {
   }
 
   // Gets a book from an archive.
-  rpc GetBookFromArchive(GetBookFromArchiveRequest) returns (Book) {
+  rpc GetBookFromArchive(GetBookFromArchiveRequest) returns (BookFromArchive) {
+    option (google.api.http) = { get: "/v1/{name=archives/**/books/*}"};
```
garrettjonesgoogle: Sot turns out that ** is only supposed to be valid at the end of a resource name pattern.

src/test/java/com/google/api/codegen/testdata/java_test_library.baseline
```diff
@@ -641,6 +645,36 @@ public class LibraryServiceTest {
 
   @Test
   @SuppressWarnings("all")
+  public void getBookFromAnywhereTest() {
+    BooNameOneof name2 = BooNameOneof.from(BookName.create("[SHELF_ID]", "[BOOK_ID]"));
```
garrettjonesgoogle: Why is the type "Boo..." instead of "Book..."?

src/main/java/com/google/api/codegen/config/ApiConfig.java
```diff
@@ -199,8 +219,72 @@ public static ApiConfig createDummyApiConfig(
...
+            CollectionConfig.createCollection(diagCollector, collectionConfigProto);
+        if (collectionConfig == null) {
+          continue;
+        }
+        if (collectionConfigsMap.containsKey(collectionConfig.getEntityName())) {
+          CollectionConfig otherConfig = collectionConfigsMap.get(collectionConfig.getEntityName());
+          if (!collectionConfig.getNamePattern().equals(otherConfig.getNamePattern())) {
+            throw new IllegalArgumentException(
+                "Inconsistent collection configs across interfaces. Entity name: "
+                    + collectionConfig.getEntityName());
```
garrettjonesgoogle: add a diag error and return null instead, in order to accumulate more errors for reporting to the user.

src/main/java/com/google/api/codegen/config/ResourceNameType.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen.config;
+
+public enum ResourceNameType {
+  NONE,
+  SINGLE,
+  INVALID,
```
garrettjonesgoogle: Is INVALID ever populated?

src/main/java/com/google/api/codegen/config/ResourceNameType.java
```diff
@@ -0,0 +1,23 @@
...
+  ONEOF,
+  ANY
+}
```
garrettjonesgoogle: I would recommend documenting each of these enums.

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
```diff
@@ -277,15 +280,17 @@ private void setListMethodFields(
...
     } else {
       resourceTypeName = typeTable.getAndSaveNicknameForElementType(resourceField.getType());
     }
 
     String iterateMethodName =
         context
             .getNamer()
-            .getPagedResponseIterateMethod(context.getFeatureConfig(), resourceFieldConfig);
+            .getPagedResponseIterateMethod(
+                context.getFeatureConfig(), resourceFieldConfig, resourceNameType);
```
garrettjonesgoogle: I haven't thought enough about this to have an opinion, but would it make sense for resource name type to be an attribute of the field config?

src/main/java/com/google/api/codegen/transformer/MethodTransformerContext.java
```diff
@@ -79,11 +80,31 @@ public InterfaceConfig getInterfaceConfig() {
...
+
+  public boolean entityNameIsCollectionConfig(String entityName) {
+    return getApiConfig().getCollectionConfig(entityName) != null;
+  }
+
+  public boolean entityNameIsCollectionOneofConfig(String entityName) {
+    return getApiConfig().getCollectionOneofConfigs().get(entityName) != null;
+  }
+
+  public CollectionConfig getFirstCollectionConfig(String entityName) {
```
garrettjonesgoogle: Can you explain why we need the notion of "first" collection config?

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -104,18 +105,40 @@ public String getPagedResponseWrappersClassName() {
   }
 
   /** The name of the generated resource type from the entity name. */
-  public Name getResourceTypeName(String entityName) {
-    return Name.from(entityName).join("name");
+  public Name getResourceTypeName(String entityName, ResourceNameType resourceNameType) {
```
garrettjonesgoogle: None of the other SurfaceNamer methods return a Name - they return a resolved string. This string will only be used as a className. Maybe make this method protected and make a public one that returns the resolved String using className.

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -83,6 +83,13 @@ message ConfigProto {
   // A list of API interface configurations.
   repeated InterfaceConfigProto interfaces = 10;
 
+  // A list of resource name collection configs. This must be consistent with
+  // collections defined in the interfaces.
+  repeated CollectionConfigProto collections = 15;
```
garrettjonesgoogle: Does this PR deprecate the collection config at the interface level, or will they both be supported for the moment?

src/main/resources/com/google/api/codegen/java/initcode.snip
```diff
@@ -48,8 +48,14 @@
...
+  @case "ResourceNameOneofInitValueView"
+    {@initValue.resourceOneofTypeName}.from({@renderResourceName(initValue.specificResourceNameView)})
   @default
     $unhandledCase: {@initValue.type}$
   @end
 @end
+
+@private renderResourceName(initValue)
+  {@initValue.resourceTypeName}.create({@argList(initValue.formatArgs)})
+@end
```
garrettjonesgoogle: add newline

src/test/java/com/google/api/codegen/testdata/library_config.baseline
```diff
@@ -19,6 +19,8 @@ interfaces:
   collections:
   - name_pattern: archives/{archive_path=**}/books/{book}
     entity_name: book
+  - name_pattern: archives/{archive}/books/{book_path=**}
```
garrettjonesgoogle: the name_pattern for the book entity above needs to be fixed so that the ** isn't in the middle of the pattern - maybe file an issue.

src/main/java/com/google/api/codegen/config/ResourceNameType.java
michaelbausor: Not in this PR - once this is merged I will follow up with support for invalid resource names

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
michaelbausor: Yes, that does make a lot of sense. The reason I haven't done that is related to the design choice between having the Config classes reflect the config proto or not.

So having the resourceNameType be a property of the field config would be more require more complex construction and be less reflective of the proto, but would reduce the complexity of the transformers.

If we are happy to break the connection between config classes and config proto, I think this is a sensible change.

src/main/java/com/google/api/codegen/transformer/MethodTransformerContext.java
michaelbausor: This is a "helper" method that allows a collection config to always be retrieved, even if the field is set to use a collection oneof. In that case, the first collection config used in the oneof is returned.

src/main/java/com/google/api/codegen/config/ApiConfig.java
michaelbausor: Done

src/main/java/com/google/api/codegen/config/ResourceNameType.java
michaelbausor: Done

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
michaelbausor: It is used outside the surface namer here:
https://github.com/googleapis/toolkit/blob/master/src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java#L363

But this is only for a local var name. And because it is in the Java test transformer, the upper camel format is known, so have updated as you suggest.

src/main/resources/com/google/api/codegen/java/initcode.snip
michaelbausor: Done

src/test/java/com/google/api/codegen/testdata/library_config.baseline
michaelbausor: This was caused by a wrong pattern in library.proto - have updated.

src/test/java/com/google/api/codegen/testdata/library_config.baseline
```diff
@@ -17,14 +17,12 @@ language_settings:
...
   collections:
-  - name_pattern: archives/{archive_path=**}/books/{book}
-    entity_name: book
   - name_pattern: archives/{archive}/books/{book_path=**}
     entity_name: book_path
   - name_pattern: bookShelves/{book_shelf}
     entity_name: book_shelf
   - name_pattern: bookShelves/{book_shelf}/books/{book}
-    entity_name: book_2
+    entity_name: book
```
garrettjonesgoogle: We're now losing our testing of multiple resources having the same last path component

src/main/java/com/google/api/codegen/config/ResourceNameType.java
```diff
@@ -15,9 +15,18 @@
 package com.google.api.codegen.config;
 
 public enum ResourceNameType {
+  /** No resource type is available. */
   NONE,
+
+  /** A single concrete resource type. */
   SINGLE,
+
+  /** A single "invalid value" resource type. */
```
garrettjonesgoogle: I think this would benefit from an example to make it clearer. The thing that is potentially confusing is what level this is "invalid" at - it's not invalid at the toolkit level; it's a possible value of resource names returned from the surface, but invalid to use for actually creating or referencing a resource.

src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
garrettjonesgoogle: I'm ok with breaking the connection here. I'm starting to be of the opinion that if our config classes just mirror the proto exactly, we aren't gaining much, and might as well use the generated proto classes directly. It's more valuable to provide a nicer surface to work on for transformation.

src/main/java/com/google/api/codegen/transformer/MethodTransformerContext.java
garrettjonesgoogle: That actually does't address the "why" though - what makes it so that the "first" collection config found is the correct one for the use case?

src/main/proto/com/google/api/codegen/config.proto
garrettjonesgoogle: Agreed.

src/main/java/com/google/api/codegen/transformer/MethodTransformerContext.java
michaelbausor: This method exists to support method sample generation, which requires any collection config to use as an example. It is only used in one place - maybe it would be clearer to remove the method, and put the logic to get the first collection config in the init transformer.

src/main/java/com/google/api/codegen/config/ResourceNameType.java
michaelbausor: Done

src/test/java/com/google/api/codegen/testdata/library_config.baseline
michaelbausor: Fixed

michaelbausor:
PTAL

garrettjonesgoogle:
LGTM

PR #693, opened 2016-10-28
Change sample application names for discovery sample gen

tcoffee-google:
Previous sample application names were specific to Google Cloud Platform, and did not generalize appropriately to non-Cloud APIs. Moreover, they did not have the appropriate format for building the User-Agent header, for which purpose they are used by some client libraries.

Also simplify unnecessary naming overrides.


saicheems:
Btw, why do we set a sample name in Java? It doesn't seem to be done in any other language - does tagging with the service name and version allow us to gather statistics on sample usage?

saicheems:
Well yeah it does actually add the application name to the user agent:
https://github.com/google/google-api-java-client/blob/4fc8c099d9db5646770868cc1bc9a33c9225b3c7/google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClientRequest.java#L117

And if you don't set the application name it logs a warning at you.

src/main/java/com/google/api/codegen/discovery/config/SampleConfig.java
```diff
@@ -87,6 +87,10 @@
   @JsonProperty("packagePrefix")
   public abstract String packagePrefix();
 
+  /** Returns a sample application name. */
+  @JsonProperty("appName")
+  public abstract String appName();
```
tcoffee-google: Yes, more consistent that way, I was just trying to avoid copying more code. We can do this for now, but I'd like to de-dupe the SampleMethodToViewTransformers once they're done.

src/main/java/com/google/api/codegen/discovery/viewmodel/SampleView.java
```diff
@@ -41,6 +41,8 @@ public String resourceRoot() {
 
   public abstract String apiVersion();
 
+  public abstract String appName();
```
saicheems: If you set `@Nullable` you don't have to provide it in any other language transformer.

vchudnov-g:
Could you add a few words to the description as to why we need this change?

src/main/java/com/google/api/codegen/discovery/viewmodel/SampleView.java
tcoffee-google: Good point. Done.

tcoffee-google:
> add a few words to the description

Done.

saicheems:
I suspect your test is failing due to the same issue Michael and I were observing...

LGTM - try to get it to pass before submitting

tcoffee-google:
Tests pass locally. Submitting.

PR #694, opened 2016-10-28
Delete old Java, Go, Ruby, PHP files

pongad:
These files are no longer needed after MVVM migration.

Adding engineers who worked on Discovery here to make sure I don't delete anything important.


saicheems:
I'm cool with it! Probably wait for another lgtm though

garrettjonesgoogle:
LGTM - let someone from discovery sign off too, as you say.

vchudnov-g:
I was defaulting to a plan where after we switch all languages to MVVM and have them working successfully, we would delete the old files en masse then, just to make it easier to refer back if we found a problem or something that hadn't been ported.

We are working and switching incrementally, though, (although we have not externally published anything created by the MVVM code yet) so if we're pretty confident all the switching work for these languages is done, we could go ahead and remove the old files incrementally. Thoughts?

garrettjonesgoogle:
If we want to save code for reference, I would recommend creating a reference branch instead of keeping the code around in master.

vchudnov-g:
That sounds good. Let's move the pre-MVVM code to a reference branch. As we continue to deprecate more languages' pre-MVVM code, let's also move it to the same (rebased) branch. Our final state should be all the MVVM code in that branch, and it should build. Soon after, once we can verify we're happy with the MVVM code paths, we'll delete the branch entirely.

pongad:
@vchudnov-g PTAL, are we good with this?

vchudnov-g:
LGTM. Update the issue title before submitting.

PR #695, opened 2016-10-28
Go uses FileHeaderView

pongad:


src/main/java/com/google/api/codegen/transformer/ImportTypeTransformer.java
```diff
@@ -24,12 +24,13 @@
 import java.util.Set;
 import java.util.TreeSet;
 
-public class ImportTypeTransformer {
+public class ImportTypeTransformer implements ImportGenerator {
```
garrettjonesgoogle: I don't think we should have separate "transformer" and "generator" terms. I would suggest renaming this existing class `ImportTypeTransformer` to `StandardImportTypeTransformer` and renaming the interface `ImportGenerator` to `ImportTypeTransformer`.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicClientTransformer.java
```diff
@@ -133,7 +127,7 @@ private StaticLangApiAndSettingsFileView generateApiAndSettingsView(
 
     // must be done as the last step to catch all imports
     csharpCommonTransformer.addCommonImports(context);
-    fileView.fileHeader(fileHeaderTransformer.generateFileHeader(context));
+    fileView.fileHeader(fileHeaderTransformer.generateFileHeader(context, importTypeTransformer));
```
garrettjonesgoogle: I would prefer the fileHeaderTransformer to take importTypeTransformer in its constructor.

src/main/java/com/google/api/codegen/transformer/ImportTypeTransformer.java
pongad: Done

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicClientTransformer.java
pongad: Done

pongad:
@garrettjonesgoogle PTAL

garrettjonesgoogle:
LGTM

PR #699, opened 2016-10-31
Add the mock half of Go unit test

pongad:
This commit implements mocks to be used by Go unit tests.
The tests themselves will be added in a later commit.


src/test/java/com/google/api/codegen/testdata/go_mock_library.baseline
```diff
@@ -0,0 +1,281 @@
...
+
+import (
+    "io"
+
+    "golang.org/x/net/context"
+)
+
+var _ = io.EOF
+
+type MockLibraryServiceImpl struct {
```
jba: We don't want to export this. Also, I don't think we need "Impl". So: mockLibraryService.

src/test/java/com/google/api/codegen/testdata/go_mock_library.baseline
```diff
@@ -0,0 +1,281 @@
...
+}
+
+var _ librarypb.LibraryServiceServer = &MockLibraryServiceImpl{}
+
+func (s *MockLibraryServiceImpl) CreateShelf(_ context.Context, req *librarypb.CreateShelfRequest) (*librarypb.Shelf, error) {
+    s.reqs = append(s.reqs, req)
+    if s.err != nil {
+        return nil, s.err
+    }
+    return s.resps[0].(*librarypb.Shelf), nil
```
jba: Why do you always return the first response? When do the others get used?

src/test/java/com/google/api/codegen/testdata/go_mock_library.baseline
```diff
@@ -0,0 +1,281 @@
...
+    err error
+
+    // responses to return if err == nil
+    resps []interface{}
+}
+
+var _ librarypb.LibraryServiceServer = &MockLibraryServiceImpl{}
+
+func (s *MockLibraryServiceImpl) CreateShelf(_ context.Context, req *librarypb.CreateShelfRequest) (*librarypb.Shelf, error) {
+    s.reqs = append(s.reqs, req)
```
jba: Since the only thing different is the cast, I think you can factor this out to a common function.

src/test/java/com/google/api/codegen/testdata/go_mock_library.baseline
jba: Oh, I think I see -- for streaming RPCs. OK.

src/test/java/com/google/api/codegen/testdata/go_mock_library.baseline
pongad: You're right. Wasn't thinking when reusing java code. Done.

src/test/java/com/google/api/codegen/testdata/go_mock_library.baseline
pongad: Done.

pongad:
@jba @garrettjonesgoogle PTAL

garrettjonesgoogle:
I'll have @shinfan take a look (since he's the test-gen person)

jba:
LGTM after one small change.

src/main/java/com/google/api/codegen/viewmodel/testing/MockGrpcMethodView.java
```diff
@@ -25,6 +25,8 @@
 
   public abstract String responseTypeName();
 
+  public abstract String streamHandle();
```
shinfan: streamHandle sounds vague to me. Can you be more specific?

pongad:
I'll "dismiss" @jba's review, since he already LGTMed.

src/main/java/com/google/api/codegen/transformer/MockServiceTransformer.java
```diff
@@ -0,0 +1,76 @@
...
+import com.google.api.tools.framework.model.Method;
+import com.google.api.tools.framework.model.Model;
+import com.google.common.base.Strings;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+public class MockServiceTransformer {
```
pongad: Done.

src/main/java/com/google/api/codegen/transformer/go/GoGapicSurfaceTransformer.java
```diff
@@ -230,6 +239,38 @@ private PackageInfoView generatePackageInfo(
     return packageInfo.build();
   }
 
+  private MockCombinedView generateMockServiceView(
```
pongad: Ah that's why. Done

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -483,13 +483,25 @@ public String getServicePhraseName(Interface service) {
   }
 
   /**
+   * The type name of the Grpc server class. This needs to match what Grpc generates for the
+   * particular language.
+   */
+  public String getGrpcServerTypeName(Interface service) {
```
pongad: This is mostly copied from `getGrpcClientTypeName` and changed "Client" to "Server". I think this is just as general. If we want to turn it into `getNotImplementedString`, we should probably change both.

src/main/java/com/google/api/codegen/viewmodel/testing/MockGrpcMethodView.java
pongad: I left it rather vague because I didn't know what name to generalize it with. In Go, this is the type name of the "stream object", which you can use to send and receive data. An analog in Java would probably be `StreamObserver` though Go has both send and receive on the same object.

streamHandleTypeName?

src/test/java/com/google/api/codegen/testdata/go_mock_library.baseline
```diff
@@ -42,138 +42,150 @@ type MockLibraryServiceImpl struct {
     resps []interface{}
 }
 
-var _ librarypb.LibraryServiceServer = &MockLibraryServiceImpl{}
+var _ librarypb.LibraryServiceServer = &mockLibraryService{}
 
-func (s *MockLibraryServiceImpl) CreateShelf(_ context.Context, req *librarypb.CreateShelfRequest) (*librarypb.Shelf, error) {
+func (s *mockLibraryService) recordReq(req interface{}) {
     s.reqs = append(s.reqs, req)
```
pongad: No problem! Reverted.

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
shinfan: Oh yes, getGrpcClient/ServerTypeName sounds language specific to me so I think in the case we probably should change both.

src/main/java/com/google/api/codegen/gapic/MainGapicProviderFactory.java
```diff
@@ -125,14 +126,29 @@
                 .build();
         providers.add(provider);
       }
+      if (generatorConfig.enableTestGenerator()) {
+        GapicCodePathMapper javaTestPathMapper =
+            CommonGapicCodePathMapper.newBuilder()
+                .setPrefix("src/test/java")
```
garrettjonesgoogle: this prefix is wrong for Go, right?

src/main/java/com/google/api/codegen/transformer/MockServiceTransformer.java
```diff
@@ -0,0 +1,77 @@
...
+import com.google.common.base.Strings;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+/** MockServiceTransformer contains helper methods useful for creating mock views. */
+public class MockServiceTransformer {
+  public Collection<Interface> getGrpcInterfacesToMock(Model model, ApiConfig apiConfig) {
```
garrettjonesgoogle: Why not a List<Interface>? We are using an implementation where order matters, and ordering is important for deterministic baselines.

src/main/java/com/google/api/codegen/transformer/go/GoGapicSurfaceTestTransformer.java
```diff
@@ -0,0 +1,94 @@
...
+          SurfaceTransformerContext.create(
+              service,
+              apiConfig,
+              GoGapicSurfaceTransformer.createTypeTable(),
+              namer,
+              featureConfig);
+      impls.add(
+          MockServiceImplView.newBuilder()
+              .outputPath("")
+              .templateFileName("")
```
garrettjonesgoogle: I think we should do something similar to what I did for separate API / Settings classes vs combined, where the file information is in a view model containing the view model with class-only information. That way we don't get weird things like this where `outputPath` and `templateFileName` are empty.

src/main/java/com/google/api/codegen/transformer/go/GoGapicSurfaceTransformer.java
```diff
@@ -69,9 +70,9 @@
 
 public class GoGapicSurfaceTransformer implements ModelToViewTransformer {
 
-  private static final String XAPI_TEMPLATE_FILENAME = "go/main.snip";
-  private static final String SAMPLE_TEMPLATE_FILENAME = "go/example.snip";
-  private static final String DOC_TEMPLATE_FILENAME = "go/doc.snip";
+  private static final String XAPI_TEMPLATE_FILE = "go/main.snip";
```
garrettjonesgoogle: Remove the 'X', since that convention is gone now. Also, why are you removing the 'NAME' part? Using only 'FILE' implies the `File` class type.

src/main/resources/com/google/api/codegen/go/mock.snip
```diff
@@ -0,0 +1,129 @@
+@extends "go/header.snip"
+
+@snippet generate(view)
+    {@headerComment(view.fileHeader)}
+
+    package {@view.fileHeader.localPackageName}
+
+    import (
+        {@imports(view.fileHeader.imports)}
```
garrettjonesgoogle: Why don't you have a snippet in header.snip which does all of the header comment, package, and imports?

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
pongad: Done.

src/main/java/com/google/api/codegen/gapic/MainGapicProviderFactory.java
pongad: Heh. We don't even use the variable. I must have just forgotten to delete it.

src/main/java/com/google/api/codegen/transformer/MockServiceTransformer.java
pongad: This was copied straight from the Java implementation, but you're right, a List is more appropriate.

shinfan:
LGTM after s/streamHandle/streamHandleTypeName.

src/main/java/com/google/api/codegen/viewmodel/testing/MockGrpcMethodView.java
pongad: Done

src/main/java/com/google/api/codegen/transformer/go/GoGapicSurfaceTestTransformer.java
pongad: I agree it would make this less weird. With some experience from `StaticLangClientFileView` though, I think this increases the cost of code maintenance. For example, if someone added a method for completely creating a `MockServiceImplView`, Go would not be able to make sure of it. As more fields get added, Go would also need to manually copy them where they're needed. I think this also benefits engineers as take move between languages. Allowing for these weird fields, I think, is offset by less code duplication and more uniformity. Wdyt?

garrettjonesgoogle: Maybe you're misunderstanding my recommendation:

* Go: `MockCombinedView` contains `FileHeaderView` and list of `MockServiceImplView` (maybe rename `MockCombinedView` to `MockMultipleServiceImplFileView`)
* Java: `MockServiceImplFileView` contains `FileHeaderView` and a single `MockServiceImplView`; `fileHeader` is removed from `MockServiceImplView`

There would be a method for creating a `MockServiceImplView` which is usable by both Java and Go. The differing aspect would be that Go creates a `MockCombinedView` containing it, whereas Java creates a `MockServiceImplFileView` containing it. The difference from the current state of your PR is mostly a change for Java.

src/main/java/com/google/api/codegen/viewmodel/testing/MockCombinedView.java
```diff
@@ -0,0 +1,62 @@
...
+
+import com.google.api.codegen.SnippetSetRunner;
+import com.google.api.codegen.viewmodel.FileHeaderView;
+import com.google.api.codegen.viewmodel.ViewModel;
+import com.google.auto.value.AutoValue;
+import java.util.List;
+
+/**
+ * Sort of a union for mock-related classes. Used in languages that idiomatically put mocks and
+ * tests in the same file.
```
garrettjonesgoogle: Be more concrete than "sort of". I don't see any test-related stuff here; is this just a collection of multiple mocks in a single file?

src/main/resources/com/google/api/codegen/go/mock.snip
garrettjonesgoogle: Ok that makes sense and sounds familiar, I think I encountered the same thing when prototyping Go header generation.

src/main/java/com/google/api/codegen/viewmodel/testing/MockCombinedView.java
pongad: Reworded. The tests are "written" as a TODO for the time being. I will land them in a separate PR to keep the PR small.

src/main/java/com/google/api/codegen/viewmodel/testing/MockCombinedView.java
```diff
@@ -0,0 +1,62 @@
...
+package com.google.api.codegen.viewmodel.testing;
+
+import com.google.api.codegen.SnippetSetRunner;
+import com.google.api.codegen.viewmodel.FileHeaderView;
+import com.google.api.codegen.viewmodel.ViewModel;
+import com.google.auto.value.AutoValue;
+import java.util.List;
+
+/**
+ * MockCombinedView acts gathers unit-test-related classes. Used in languages that idiomatically put
```
garrettjonesgoogle: Wording problem on this line

pongad: OTL. Removed "acts".

PR #701, opened 2016-10-31
Ruby: Add LRO Wrapper.

landrito:


src/main/java/com/google/api/codegen/transformer/ApiMethodTransformer.java
```diff
@@ -535,6 +535,10 @@ public OptionalArrayMethodView generateDynamicLangApiMethod(MethodTransformerCon
         generateRequestObjectParams(context, filteredFieldConfigs));
     apiMethod.grpcStreamingType(context.getMethodConfig().getGrpcStreamingType());
 
+    ServiceMessages messages = new ServiceMessages();
+    apiMethod.isLongRunning(
+        messages.isLongRunningOperationType(context.getMethod().getOutputType()));
+
```
swcloud: This can be simplified as 

apiMethod.isLongRunning(context.isLongRunning(context.getMethod()));

swcloud: I see.

landrito:
@swcloud @jmuk 

This should be ready for another look.

I added a baseline test for checking whether the OperationsApi is affected by the change since the LRO design depends on the OperationsApi returning raw Google::Longrunning::Operations.

landrito:
Hold off on reviewing. There was some changes to how lro is checked that I need to confirm this PR to.

landrito:
This is ready for another look!

src/main/java/com/google/api/codegen/ruby/RubyGapicContext.java
```diff
@@ -382,6 +384,29 @@ private SurfaceTransformerContext getSurfaceTransformerContextFromService(Interf
...
+  public boolean isLongrunning(Method method, Interface service) {
+    return getApiConfig()
+        .getInterfaceConfig(service)
+        .getMethodConfig(method)
+        .isLongRunningOperation();
+  }
+
+  public boolean hasLongrunningMethod(Interface service) {
+    for (Method method : getSupportedMethodsV2(service)) {
+      if (getLongrunningConfig(method, service) != null) {
```
swcloud: I think calling 'isLongrunning(method, service)' is  a bit better here since it's more self-explaining and isLongRunningOperation is checking 'getLongRunningConfig() != null'

Also please use 'LongRunning' instead of 'Longrunning' in order to be consistent.

src/main/resources/com/google/api/codegen/ruby/main.snip
```diff
@@ -435,10 +454,22 @@
   {@context.upperCamelToLowerUnderscore(method.getSimpleName())}
 @end
 
-@private makeApiCall(method, request, isReturnEmptyMessage)
-  @@{@getMethodName(method)}.call({@request}, options)
-  @if isReturnEmptyMessage
-    nil
+@private makeApiCall(method, service, request, isReturnEmptyMessage)
```
swcloud: I feel that we can pass 'methodConfig' together with 'method' and 'request' instead.

We can use methodConfig.isLongRunningOperation(), methodConfig.getLongrunningConfig(), methodConfig.isReturnEmptyMessage inside 'makeApiCall'. In this way, we don't need to depend on context, which will make migrating to MVVM easier.

src/main/java/com/google/api/codegen/ruby/RubyGapicContext.java
landrito: Done.

src/main/resources/com/google/api/codegen/ruby/main.snip
```diff
@@ -435,10 +453,32 @@
   {@context.upperCamelToLowerUnderscore(method.getSimpleName())}
 @end
 
-@private makeApiCall(method, request, isReturnEmptyMessage)
-  @@{@getMethodName(method)}.call({@request}, options)
-  @if isReturnEmptyMessage
-    nil
+@private makeApiCall(methodConfig, service)
```
swcloud: 'service' can be removed now.

src/main/resources/com/google/api/codegen/ruby/main.snip
```diff
@@ -435,10 +453,32 @@
...
+  @let request = requestParameterName(methodConfig)
```
swcloud: You may add 'method = methodConfig.getMethod' here and just use 'method' in this function.

src/main/resources/com/google/api/codegen/ruby/main.snip

src/main/resources/com/google/api/codegen/ruby/main.snip

PR #702, opened 2016-10-31
Support $PROJECT in smoke test gen.

shinfan:
- Support `$PROJECT_ID` token in the smoke test config
- Add `InitValue` class which adds support of variable names in the init fields

### DIFF
--- a/src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
+++ b/src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
@@ -79,7 +79,7 @@ public class FieldStructureParser {
       InitValue initValue = fieldConfig.value();
       valueConfig =
           InitValueConfig.createWithValue(
-              new InitValue(stripQuotes(initValue.getValue()), initValue.getType()));
+              InitValue.create(stripQuotes(initValue.getValue()), initValue.getType()));
     } else if (initValueConfigMap.containsKey(fieldConfig.fieldPath())) {
       valueConfig = initValueConfigMap.get(fieldConfig.fieldPath());
     }
--- a/src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
+++ b/src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
@@ -76,7 +76,10 @@ public class FieldStructureParser {
               .get(fieldConfig.fieldPath())
               .withInitialCollectionValue(fieldConfig.entityName(), fieldConfig.value());
     } else if (fieldConfig.hasSimpleInitValue()) {
-      valueConfig = InitValueConfig.createWithValue(stripQuotes(fieldConfig.value()));
+      InitValue initValue = fieldConfig.value();
+      valueConfig =
+          InitValueConfig.createWithValue(
+              new InitValue(stripQuotes(initValue.getValue()), initValue.getType()));
     } else if (initValueConfigMap.containsKey(fieldConfig.fieldPath())) {
       valueConfig = initValueConfigMap.get(fieldConfig.fieldPath());
     }
--- a/src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
+++ b/src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
@@ -15,7 +15,6 @@
 package com.google.api.codegen.metacode;
 
 import com.google.api.codegen.config.FieldConfig;
-import com.google.api.codegen.metacode.InitValue.InitValueType;
 import com.google.api.codegen.util.Name;
 import com.google.api.codegen.util.SymbolTable;
 import com.google.api.codegen.util.testing.TestValueGenerator;
@@ -278,8 +277,7 @@ public class InitCodeNode {
           && !type.isRepeated()
           && valueGenerator != null) {
         String newValue = valueGenerator.getAndStoreValue(type, identifier);
-        initValueConfig =
-            InitValueConfig.createWithValue(new InitValue(newValue, InitValueType.Literal));
+        initValueConfig = InitValueConfig.createWithValue(InitValue.createLiteral(newValue));
       }
     }
   }
--- a/src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
+++ b/src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
@@ -15,6 +15,7 @@
 package com.google.api.codegen.metacode;
 
 import com.google.api.codegen.config.FieldConfig;
+import com.google.api.codegen.metacode.InitValue.InitValueType;
 import com.google.api.codegen.util.Name;
 import com.google.api.codegen.util.SymbolTable;
 import com.google.api.codegen.util.testing.TestValueGenerator;
@@ -191,7 +192,7 @@ public class InitCodeNode {
 
   private static InitValueConfig mergeInitValueConfig(
       InitValueConfig oldConfig, InitValueConfig newConfig) {
-    HashMap<String, String> collectionValues = new HashMap<>();
+    HashMap<String, InitValue> collectionValues = new HashMap<>();
     if (oldConfig.hasSimpleInitialValue()
         && newConfig.hasSimpleInitialValue()
         && !oldConfig.getInitialValue().equals(newConfig.getInitialValue())) {
@@ -271,13 +272,14 @@ public class InitCodeNode {
 
       // Validate initValueConfig, or generate random value
       if (initValueConfig.hasSimpleInitialValue()) {
-        validateValue(type, initValueConfig.getInitialValue());
+        validateValue(type, initValueConfig.getInitialValue().getValue());
       } else if (initValueConfig.isEmpty()
           && type.isPrimitive()
           && !type.isRepeated()
           && valueGenerator != null) {
         String newValue = valueGenerator.getAndStoreValue(type, identifier);
-        initValueConfig = InitValueConfig.createWithValue(newValue);
+        initValueConfig =
+            InitValueConfig.createWithValue(new InitValue(newValue, InitValueType.Literal));
       }
     }
   }
--- a/src/main/java/com/google/api/codegen/metacode/InitFieldConfig.java
+++ b/src/main/java/com/google/api/codegen/metacode/InitFieldConfig.java
@@ -22,10 +22,10 @@ import javax.annotation.Nullable;
  */
 @AutoValue
 public abstract class InitFieldConfig {
-  public static final String projectIdVariableName = "project_id";
+  public static final String PROJECT_ID_VARIABLE_NAME = "project_id";
 
-  private static final String randomValueToken = "$RANDOM";
-  private static final String projectIdToken = "$PROJECT_ID";
+  private static final String RANDOM_TOKEN = "$RANDOM";
+  private static final String PROJECT_ID_TOKEN = "$PROJECT_ID";
 
   public abstract String fieldPath();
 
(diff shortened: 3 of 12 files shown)


garrettjonesgoogle:
@michaelbausor could you do the first pass?

src/test/java/com/google/api/codegen/testdata/java_smoke_test_library.baseline
```diff
@@ -31,18 +36,34 @@ public class LibraryServiceSmokeTest {
...
-      System.out.println("ok");
+      Options options = new Options();
+      options.addOption("h", "help", false, "show usage");
+      options.addOption(
+          Option.builder()
+              .longOpt("project_id")
+              .desc("Project id")
+              .hasArg()
+              .argName("PROJECT-ID")
+              .required(true)
```
michaelbausor: Do we want this to be required? Or can it be optional, with a default? Especially in the case that nothing in the smoke test config contains $PROJECT_ID, it is odd for it to be required.

src/main/java/com/google/api/codegen/metacode/InitFieldConfig.java
```diff
@@ -69,11 +72,16 @@ public boolean hasFormattedInitValue() {
...
 
-  private static String parseValueString(String valueString, String stringToHash) {
+  private static InitValue parseValueString(String valueString, String stringToHash) {
+    InitValue initValue = new InitValue(valueString, InitValueType.Literal);
     if (valueString.contains(randomValueToken)) {
       String randomValue = Integer.toString(Math.abs(stringToHash.hashCode()));
       valueString = valueString.replace(randomValueToken, randomValue);
+      initValue = new InitValue(valueString, InitValueType.Literal);
+    } else if (valueString.contains(projectIdToken)) {
+      valueString = projectIdVariableName;
```
michaelbausor: This replaces the whole string, which is different from the randomValue case. Is it possible to allow the project id to be specified alongside other data? If not, it should be an error to specify a value which contains projectIdToken, but is not equal to projectIdToken, because the extra is lost.

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -342,13 +345,18 @@ private InitValueView getInitValue(MethodTransformerContext context, InitCodeNod
...
-      List<String> varList, InitValueConfig initValueConfig) {
+      MethodTransformerContext context, List<String> varList, InitValueConfig initValueConfig) {
     List<String> formatFunctionArgs = new ArrayList<>();
     for (String entityName : varList) {
       String entityValue = "\"[" + Name.from(entityName).toUpperUnderscore() + "]\"";
       if (initValueConfig.hasFormattingConfigInitialValues()
           && initValueConfig.getCollectionValues().containsKey(entityName)) {
-        entityValue = initValueConfig.getCollectionValues().get(entityName);
+        InitValue initValue = initValueConfig.getCollectionValues().get(entityName);
+        if (initValue.getType() == InitValueType.Variable) {
```
michaelbausor: Use a switch statement

src/main/java/com/google/api/codegen/metacode/InitValue.java
```diff
@@ -0,0 +1,53 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen.metacode;
+
+/** A class which represents the initialized value of a field. */
+public class InitValue {
```
michaelbausor: I haven't looked through whether this works, but is it possible for this class to be rolled into InitFieldConfig?

src/main/java/com/google/api/codegen/metacode/InitFieldConfig.java
shinfan: Done. We should not allow users to specify project value as the random value case.

src/test/java/com/google/api/codegen/testdata/java_smoke_test_library.baseline
shinfan: Done. Changed to optional.

michaelbausor: What happens when it isn't set? Does it take a default value, or null? Having a default value seems the most sensible. Or perhaps if projectId is used but is null, there is an error.

shinfan:
@michaelbausor PTAL

src/main/java/com/google/api/codegen/metacode/InitValue.java
michaelbausor: Fair enough.

Why not use AutoValue for this class?

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
```diff
@@ -271,13 +272,14 @@ private void resolveNamesAndTypes(
...
-        validateValue(type, initValueConfig.getInitialValue());
+        validateValue(type, initValueConfig.getInitialValue().getValue());
       } else if (initValueConfig.isEmpty()
           && type.isPrimitive()
           && !type.isRepeated()
           && valueGenerator != null) {
         String newValue = valueGenerator.getAndStoreValue(type, identifier);
-        initValueConfig = InitValueConfig.createWithValue(newValue);
+        initValueConfig =
+            InitValueConfig.createWithValue(new InitValue(newValue, InitValueType.Literal));
```
michaelbausor: There are lots of uses of "new InitValue(...)". What do you think about static factory methods, such as InitValue.createLiteral(String value) and InitValue.createVariable(String variableName), or something similar?

src/test/java/com/google/api/codegen/testdata/java_smoke_test_library.baseline
michaelbausor: Can we test this logic in the baseline?

src/main/java/com/google/api/codegen/metacode/InitValue.java
shinfan: Done

src/test/java/com/google/api/codegen/testdata/java_smoke_test_library.baseline
shinfan: We only support one smoke test per interface so we can only test one of two logic in the baseline. I tested the first logic since output of the second logic is mostly a subset of the first one.  The alternative is to add one more interface to the test library but that would double the number of the baseline files so it is probably not worth doing.

src/main/java/com/google/api/codegen/metacode/InitValue.java
```diff
@@ -14,29 +14,32 @@
...
+  public static InitValue createVariable(String variableName) {
+    return new AutoValue_InitValue(variableName, InitValueType.Variable);
   }
 
+  public abstract String getValue();
+
+  public abstract InitValueType getType();
+
   @Override
   public boolean equals(Object o) {
```
michaelbausor: Is it necessary to override equals and hashcode?

shinfan: Yes it is used in the test code:
https://github.com/shinfan/toolkit/blob/8a04b2e9dbaba5e81c3ff23ba516d45a52f9bb23/src/test/java/com/google/api/codegen/metacode/SampleInitCodeTest.java#L171

michaelbausor: Yes, but AutoValue provides an implementation. Do we need to override that?

src/test/java/com/google/api/codegen/testdata/java_smoke_test_library.baseline
michaelbausor: OK yes, I agree lets not add another interface.

Not suggesting in this PR, but is there a reason we only support one smoke test? One seems enough for most cases, but maybe there will be some edge case where we want to smoke test multiple methods? Or maybe not...

src/main/java/com/google/api/codegen/metacode/InitValue.java
shinfan: Ah you are right. I forgot that the class is switched to auto value. Removed.

shinfan:
Rebased and addressed feedback. PTAL.

michaelbausor:
LGTM - let @garrettjonesgoogle take a look

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
```diff
@@ -76,7 +76,10 @@ private static InitValueConfig createInitValueConfig(
               .get(fieldConfig.fieldPath())
               .withInitialCollectionValue(fieldConfig.entityName(), fieldConfig.value());
     } else if (fieldConfig.hasSimpleInitValue()) {
-      valueConfig = InitValueConfig.createWithValue(stripQuotes(fieldConfig.value()));
+      InitValue initValue = fieldConfig.value();
+      valueConfig =
+          InitValueConfig.createWithValue(
+              InitValue.create(stripQuotes(initValue.getValue()), initValue.getType()));
```
garrettjonesgoogle: would it make sense for InitValue to do the quote stripping?

src/main/java/com/google/api/codegen/metacode/InitFieldConfig.java
```diff
@@ -22,23 +22,26 @@
  */
 @AutoValue
 public abstract class InitFieldConfig {
+  public static final String projectIdVariableName = "project_id";
```
garrettjonesgoogle: make all of these `static final` fields UPPER_SNAKE - they are constants.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
```diff
@@ -139,12 +146,15 @@ private SmokeTestClassView createSmokeTestClassView(SurfaceTransformerContext co
...
+    TestMethodView testMethodView = createSmokeTestMethodView(methodContext);
+
     testClass.apiSettingsClassName(namer.getApiSettingsClassName(service));
     testClass.apiClassName(namer.getApiWrapperClassName(service));
     testClass.name(name);
     testClass.outputPath(namer.getSourceFilePath(outputPath, name));
     testClass.templateFileName(SMOKE_TEST_TEMPLATE_FILE);
-    testClass.method(createSmokeTestMethodView(methodContext));
+    testClass.method(testMethodView);
+    testClass.requireProjectID(requireProjectId(testMethodView.initCode(), context.getNamer()));
```
garrettjonesgoogle: We should be consistent with the casing - I think per Java convention, we should have "Id" (not "ID")

src/main/resources/com/google/api/codegen/java/smoke_test.snip
```diff
@@ -10,21 +10,55 @@
     public static void main(String args[]) {
       Logger.getLogger("").setLevel(Level.WARNING);
       try {
-        executeNoCatch(args);
-        System.out.println("ok");
+        {@testBody(smokeTest)}
+        System.out.println("OK");
```
garrettjonesgoogle: You didn't like my lower case "ok"? :-(

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
shinfan: The quote stripping is used at multiple places in FieldStructureParser so I would prefer to keep it to this class.

src/main/resources/com/google/api/codegen/java/smoke_test.snip
shinfan: Ummm in my opinion upper case "OK" is more eye-catching in the output so it would be more readable. But I am OK with both :8ball:

src/main/java/com/google/api/codegen/metacode/InitFieldConfig.java
shinfan: Done

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
shinfan: Done

garrettjonesgoogle:
LGTM

PR #709, opened 2016-11-01
Reduce # of forked JVMs and initial heap size

saicheems:
Should close the case of the mysterious flaky builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 sudo: false
 language: java
+env:
+  global:
+    - GRADLE_OPTS="-Xms128m"
 jdk:
   - oraclejdk8
   - oraclejdk7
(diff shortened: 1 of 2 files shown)


saicheems:
Prior art:
https://discuss.gradle.org/t/travis-ci-org-gradle-launcher-daemon-client-daemondisappearedexception-gradle-build-daemon-disappeared-unexpectedly-it-may-have-been-killed-or-may-have-crashed/13106/2

saicheems:
Feeling pretty good about this one, ptal

garrettjonesgoogle:
I like it, LGTM

vchudnov-g:
LGTM. Can we be more aggressive with the parallelization or is this the best we can do and not run into problems?

Thanks for solving this!

saicheems:
Luckily the number of parallel jobs is a property of Travis, so we can increase that number in the UI. I'll mess with it a bit in future PRs to see if it does anything.

PR #714, opened 2016-11-02
Refactor PHP discovery snippetgen for MVVM

saicheems:


src/main/java/com/google/api/codegen/discovery/transformer/php/PhpSampleNamer.java
```diff
@@ -0,0 +1,37 @@
...
+
+public class PhpSampleNamer extends SampleNamer {
+
+  public PhpSampleNamer() {
+    super(new PhpNameFormatter());
+  }
+
+  @Override
+  public String getSampleApplicationName(String apiTypeName) {
+    // Unfortunate replacement that's necessary because the API type name is of a special format.
```
tcoffee-google: Seems like we've abused apiTypeName to also serve as what should be, say, apiCanonicalName. I'd suggest introducing another parameter.

src/main/java/com/google/api/codegen/discovery/transformer/php/PhpSampleTypeNameConverter.java
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  @Override
+  public TypeName getTypeName(TypeInfo typeInfo) {
+    // This method is only ever called for the request body type, which should
+    // always be a message.
+    if (typeInfo.isMessage()) {
+      return new TypeName(apiTypeName + "_" + typeInfo.message().typeName());
+    }
+    return getTypeNameForElementType(typeInfo);
```
tcoffee-google: Does the comment imply this line should be unreachable?

src/test/java/com/google/api/codegen/testdata/discoveries/php/php_translate.v2.json.baseline
```diff
@@ -15,18 +15,12 @@ require_once __DIR__ . '/vendor/autoload.php';
 
 $client = new Google_Client();
 $client->setApplicationName('Google-TranslateSample/0.1');
-
-// TODO: Change placeholder below to desired API key:
-$client->setDeveloperKey('{MY-API-KEY}');
+$client->setDeveloperKey(''); // TODO: Update placeholder with desired API key.
```
tcoffee-google: Two spaces before comment for consistency.

src/test/java/com/google/api/codegen/testdata/discoveries/php/php_translate.v2.json.baseline
```diff
@@ -15,18 +15,12 @@ require_once __DIR__ . '/vendor/autoload.php';
...
 
 $service = new Google_Service_Translate($client);
 
-
-// TODO: Change placeholder below to desired parameter value for the `listDetections` method:
-
 // The text to detect
-$q = [];
-
+$q = array();  // TODO: Update placeholder value.
```
tcoffee-google: Don't you think previous syntax was preferable?

src/main/java/com/google/api/codegen/discovery/transformer/php/PhpSampleTypeNameConverter.java
saicheems: Yes, and that method just returns an exception

tcoffee-google: Is there a reason it makes sense to retain the method?

src/test/java/com/google/api/codegen/testdata/discoveries/php/php_translate.v2.json.baseline
tcoffee-google: Hmm. I guess I had written `array()` for `optParams` since it was being used with just named indices, but now that you mention it, maybe it would be more idiomatic to use `[]` for both. Certainly for the placeholder; what do you think about `optParams`?

src/main/java/com/google/api/codegen/discovery/transformer/php/PhpSampleTypeNameConverter.java
tcoffee-google: I'm just wondering if there's any logic to having getTypeName delegate to getTypeNameForElementType, or if it should throw its own exception and leave getTypeNameForElementType returning null.

saicheems: Oh, I've just been doing it to save writing the exception out twice. In other languages `getTypeName` does call `getTypeNameForElementType`, so it seemed natural. In the unexpected case that the transformer does happen to call the element function it could be useful to get an stacktrace instead of null in snippet

src/test/java/com/google/api/codegen/testdata/discoveries/php/php_translate.v2.json.baseline
saicheems: Let's do `[]` for both? Based on my limited reading, `[]` is valid for PHP >= 5.4, and `array` is compatible with all versions
If we expect to target 5.4 and up then let's do `[]` for both

src/main/java/com/google/api/codegen/discovery/transformer/php/PhpSampleTypeNameConverter.java
tcoffee-google: Still seems odd to me, but I don't feel strongly about it.

src/test/java/com/google/api/codegen/testdata/discoveries/php/php_translate.v2.json.baseline
tcoffee-google: SGTM.

src/main/java/com/google/api/codegen/discovery/transformer/php/PhpSampleNamer.java
saicheems: Done

src/test/java/com/google/api/codegen/testdata/discoveries/php/php_sqladmin.v1beta4.json.baseline
```diff
@@ -1692,39 +1497,41 @@ echo '<pre>', var_export($response, true), '</pre>', "\n";
...
 // Project ID of the project that contains the instance.
-$project = '{MY-PROJECT}';
+$project = '';  // TODO: Update placeholder value.
 
 // Database instance ID. This does not include the project ID.
-$instance = '{MY-INSTANCE}';
+$instance = '';  // TODO: Update placeholder value.
 
+$optParams = [];
+do {
```
tcoffee-google: Need override here: this method is not page-streaming. Do we have this in the other languages?

src/test/java/com/google/api/codegen/testdata/discoveries/php/php_storage.v1.json.baseline
```diff
@@ -1172,51 +1034,44 @@ echo '<pre>', var_export($response, true), '</pre>', "\n";
...
 
+// TODO: Assign values to desired properties of `requestBody`:
 $requestBody = new Google_Service_Storage_StorageObject();
-// TODO: Assign values to desired properties of the `requestBody` object.
-
-
-$optParams = array();
 
+$optParams = [];
 // TODO: To download media content, use:
```
tcoffee-google: Should we have a blank line here?

src/test/java/com/google/api/codegen/testdata/discoveries/java/java_storagetransfer.v1.json.baseline
```diff
@@ -27,12 +27,8 @@ import java.util.Arrays;
 
 public class StoragetransferExample {
   public static void main(String args[]) throws IOException, GeneralSecurityException {
-    // The ID of the Google Developers Console project that the Google service account is associated with.
-    // Required.
-    String projectId = "";  // TODO: Update placeholder value.
-
     Storagetransfer storagetransferService = createStoragetransferService();
-    Storagetransfer.GoogleServiceAccounts.Get request = storagetransferService.googleServiceAccounts().get(projectId);
+    Storagetransfer.V1.GetGoogleServiceAccount request = storagetransferService.v1().getGoogleServiceAccount();
```
saicheems: Yes, I actually had the override wrong before, I only realized while fixing it in PHP

There are two calls, one is
`V1.GetGoogleServiceAccount`

and the other is
`GoogleServiceAccounts.Get`

saicheems: So this one fixes the v1.getGoogleServiceAccount path

src/test/java/com/google/api/codegen/testdata/discoveries/php/php_bigquery.v2.json.baseline
```diff
@@ -135,39 +119,34 @@ echo '<pre>', var_export($response, true), '</pre>', "\n";
...
-
-$optParams = array();
+$optParams = [];
 do {
   $response = $service->datasets->listDatasets($projectId, $optParams);
 
-  foreach ($response['datasets'] as $dataset_list._datasets) {
-    // TODO: Change code below to process each `dataset_list._datasets` resource:
-    echo '<pre>', var_export($dataset_list._datasets, true), '</pre>', "\n";
+  foreach ($response['datasets'] as $datasets) {
```
saicheems: I believe so, `$response['datasets']` is an array of `Dataset`, so this should be correct.

src/test/java/com/google/api/codegen/testdata/discoveries/php/php_storage.v1.json.baseline
saicheems: I don't mind, it seemed reasonable to not use one there. I mostly chose to drop it because there's no newline in
```
$optParams = [];
do {
  ...
```
Do you want me to keep the newline before the media content case?

src/test/java/com/google/api/codegen/testdata/discoveries/php/php_sqladmin.v1beta4.json.baseline
saicheems: Good call, done, I had it in Go and Java but not in Ruby

src/test/java/com/google/api/codegen/testdata/discoveries/php/php_storage.v1.json.baseline
saicheems: Doen

Issue #331, opened 2016-07-28
Convert proto markdown to language-specific doc format in all languages


garrettjonesgoogle:
- [ ] Java
- [ ] PHP
- [ ] Node.js
- [ ] Ruby
- [ ] Python
- [ ] Go
- [ ] C#

bjwatson:
Related PRs for Python:
- https://github.com/googleapis/toolkit/issues/222
- https://github.com/googleapis/toolkit/issues/402

Per @geigerj:

See also

https://github.com/googleapis/toolkit/blob/master/src/main/java/com/google/api/codegen/CommentPatterns.java

and 

https://github.com/googleapis/toolkit/blob/master/src/main/java/com/google/api/codegen/py/PythonSphinxCommentFixer.java

for what Python does here. As discussed offline it would eventually be good to get a proper Markdown processor hooked up to this, like what we have internally.

pongad:
According to an internal wiki, all docs should conform to https://github.com/sirthias/pegdown (Apache licensed). Might be possible to just use it.

jskeet:
As I've noted internally, I would *much* prefer that a lot of this work is done in protoc. Without that, all the generated Javadoc for Java protos and ditto the XML documentation for C# will end up still being done in Markdown.

chrisdunelm:
Not necessary in C#. The C# documentation processors handle markdown fine.

jskeet:
@chrisdunelm: Well, sort of. docfx does, but Visual Studio doesn't (for Intellisense) - and we don't resolve all the inter-proto references etc. It would definitely be *better* if we could generate idiomatic XML documentation, IMO.

chrisdunelm:
OK, marking this as not done for C# again ;)

miraleung:
Looks like this is fixed for Java, and most other languages (save PHP) have migrated to microgenerators. Closing as obsolete.

PR #716, opened 2016-11-03
Fix datastore grpc doc class generation for python and ruby

swcloud:
#716 This is due to datastore API's GqlQuery message's comment contains '$' so we have to escape it, otherwise it will cause java's regex failure (throwing IAE exception)

In https://github.com/googleapis/googleapis/blob/master/google/datastore/v1/query.proto#L232,

-------------------------

Key must match regex `[A-Za-z_$][A-Za-z_$0-9]*`

-------------------------

The original CommentPatterns.PROTO_LINK_PATTERN causes the above [A-Za-z_$][A-Za-z_$0-9] to be parsed as a proto link, using Ruby as an example (similar with Python and NodeJS)

In RDocCommentFixer.java:
```
  private static String rdocifyProtoMarkdownLinks(String comment) {
    StringBuffer sb = new StringBuffer();
    Matcher m = CommentPatterns.PROTO_LINK_PATTERN.matcher(comment);
    if (!m.find()) {
      return comment;
    }
    do {
      m.appendReplacement(sb, String.format("%s", protoToRubyDoc(m.group(1))));
    } while (m.find());
    m.appendTail(sb);
    return sb.toString();
  }
```

When m.appendReplacement is called, the second argument contains a '$' in it (because m.group(1) is A-Za-z_$, so that causes appendReplacement to throw IAE.

### DIFF
--- a/src/main/java/com/google/api/codegen/CommentPatterns.java
+++ b/src/main/java/com/google/api/codegen/CommentPatterns.java
@@ -27,6 +27,6 @@ public final class CommentPatterns {
   public static final Pattern CLOUD_LINK_PATTERN =
       Pattern.compile("\\[([^\\]]+)\\]\\(((?!\\p{Alpha}+:)[^\\)]+)\\)");
   public static final Pattern PROTO_LINK_PATTERN =
-      Pattern.compile("\\[([^\\]]+)\\]\\[([A-Za-z_][A-Za-z_.0-9]*)*\\]");
+      Pattern.compile("\\[([^\\]]+)\\]\\[([A-Za-z_][A-Za-z_.0-9]*)?\\]");
   public static final Pattern HEADLINE_PATTERN = Pattern.compile("^#+", Pattern.MULTILINE);
 }
--- a/src/main/java/com/google/api/codegen/CommentPatterns.java
+++ b/src/main/java/com/google/api/codegen/CommentPatterns.java
@@ -26,6 +26,7 @@ public final class CommentPatterns {
       Pattern.compile("\\[([^\\]]+)\\]\\((\\p{Alpha}+:[^\\)]+)\\)");
   public static final Pattern CLOUD_LINK_PATTERN =
       Pattern.compile("\\[([^\\]]+)\\]\\(((?!\\p{Alpha}+:)[^\\)]+)\\)");
-  public static final Pattern PROTO_LINK_PATTERN = Pattern.compile("\\[([^\\]]+)\\]\\[[^\\]]*\\]");
+  public static final Pattern PROTO_LINK_PATTERN =
+      Pattern.compile("\\[([^\\]]+)\\]\\[([A-Za-z_][A-Za-z_.0-9]*)*\\]");
   public static final Pattern HEADLINE_PATTERN = Pattern.compile("^#+", Pattern.MULTILINE);
 }
--- a/src/main/java/com/google/api/codegen/py/PythonSphinxCommentFixer.java
+++ b/src/main/java/com/google/api/codegen/py/PythonSphinxCommentFixer.java
@@ -37,7 +37,8 @@ public class PythonSphinxCommentFixer {
       return comment;
     }
     do {
-      m.appendReplacement(sb, String.format("``%s``", m.group(1)));
+      // proto display name may contain '$' which needs to be escaped using Matcher.quoteReplacement
+      m.appendReplacement(sb, Matcher.quoteReplacement(String.format("``%s``", m.group(1))));
     } while (m.find());
     m.appendTail(sb);
     return sb.toString();
--- a/src/main/java/com/google/api/codegen/py/PythonSphinxCommentFixer.java
+++ b/src/main/java/com/google/api/codegen/py/PythonSphinxCommentFixer.java
@@ -22,8 +22,6 @@ public class PythonSphinxCommentFixer {
 
   /** Returns a Sphinx-formatted comment string. */
   public static String sphinxify(String comment) {
-    // escape '$' in the comment first
-    comment = comment.replaceAll("\\$", "\\\\\\$");
     comment = CommentPatterns.BACK_QUOTE_PATTERN.matcher(comment).replaceAll("``");
     comment = comment.replace("\"", "\\\"");
     comment = sphinxifyProtoMarkdownLinks(comment);
@@ -53,7 +51,9 @@ public class PythonSphinxCommentFixer {
       return comment;
     }
     do {
-      m.appendReplacement(sb, String.format("`%s <%s>`_", m.group(1), m.group(2)));
+      // absolute markdown links may contain '$' which needs to be escaped using Matcher.quoteReplacement
+      m.appendReplacement(
+          sb, Matcher.quoteReplacement(String.format("`%s <%s>`_", m.group(1), m.group(2))));
     } while (m.find());
     m.appendTail(sb);
     return sb.toString();
@@ -67,8 +67,11 @@ public class PythonSphinxCommentFixer {
       return comment;
     }
     do {
+      // cloud markdown links may contain '$' which needs to be escaped using Matcher.quoteReplacement
       m.appendReplacement(
-          sb, String.format("`%s <https://cloud.google.com%s>`_", m.group(1), m.group(2)));
+          sb,
+          Matcher.quoteReplacement(
+              String.format("`%s <https://cloud.google.com%s>`_", m.group(1), m.group(2))));
     } while (m.find());
     m.appendTail(sb);
     return sb.toString();
--- a/src/main/java/com/google/api/codegen/py/PythonSphinxCommentFixer.java
+++ b/src/main/java/com/google/api/codegen/py/PythonSphinxCommentFixer.java
@@ -22,6 +22,8 @@ public class PythonSphinxCommentFixer {
 
   /** Returns a Sphinx-formatted comment string. */
   public static String sphinxify(String comment) {
+    // escape '$' in the comment first
+    comment = comment.replaceAll("\\$", "\\\\\\$");
     comment = CommentPatterns.BACK_QUOTE_PATTERN.matcher(comment).replaceAll("``");
     comment = comment.replace("\"", "\\\"");
     comment = sphinxifyProtoMarkdownLinks(comment);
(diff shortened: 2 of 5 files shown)


src/main/java/com/google/api/codegen/py/PythonSphinxCommentFixer.java
```diff
@@ -22,6 +22,8 @@
 
   /** Returns a Sphinx-formatted comment string. */
   public static String sphinxify(String comment) {
+    // escape '$' in the comment first
+    comment = comment.replaceAll("\\$", "\\\\\\$");
```
bjwatson: Why so many backslashes?

swcloud: For each \, you need to two extra \\, any better way?

jmuk: http://docs.oracle.com/javase/7/docs/api/java/lang/String.html#replace(java.lang.CharSequence,%20java.lang.CharSequence) This can be used?

src/main/java/com/google/api/codegen/py/PythonSphinxCommentFixer.java
```diff
@@ -22,6 +22,8 @@
 
   /** Returns a Sphinx-formatted comment string. */
   public static String sphinxify(String comment) {
+    // escape '$' in the comment first
```
jmuk: It is unclear to me why '$' needs to be replaced. Can you bring a URL to explain what '$' sign means in Sphinx?

src/main/java/com/google/api/codegen/ruby/RDocCommentFixer.java
```diff
@@ -21,8 +21,10 @@
 /** Utility class for formatting source comments to follow RDoc style. */
 public class RDocCommentFixer {
 
-  /** Returns a Sphinx-formatted comment string. */
+  /** Returns a RDoc-formatted comment string. */
   public static String rdocify(String comment) {
+    // escape '$' in the comment first
```
jmuk: ditto for RDoc or YARD link

src/main/java/com/google/api/codegen/py/PythonSphinxCommentFixer.java
geigerj: Do we need to undo the extra backslashes at the end so that the generated comment doesn't contain the escape characters?

src/main/java/com/google/api/codegen/py/PythonSphinxCommentFixer.java
swcloud: This has nothing to do with Sphinix. This is java regex API's requirement, for example, https://github.com/googleapis/toolkit/blob/master/src/main/java/com/google/api/codegen/ruby/RDocCommentFixer.java#L64, The second argument of 'appendReplacement' cannot contain a non-escaped '$' in it.

src/main/java/com/google/api/codegen/py/PythonSphinxCommentFixer.java
bjwatson: 👍 to @jmuk's suggestion. We're not doing any abstract pattern matching here. Literals are fine. (i.e. `fgrep` rather than `egrep`)

swcloud: @jmuk Do you mean we should use 'replace' instead of 'replaceAll'?

src/main/java/com/google/api/codegen/CommentPatterns.java
```diff
@@ -26,6 +26,7 @@
       Pattern.compile("\\[([^\\]]+)\\]\\((\\p{Alpha}+:[^\\)]+)\\)");
   public static final Pattern CLOUD_LINK_PATTERN =
       Pattern.compile("\\[([^\\]]+)\\]\\(((?!\\p{Alpha}+:)[^\\)]+)\\)");
-  public static final Pattern PROTO_LINK_PATTERN = Pattern.compile("\\[([^\\]]+)\\]\\[[^\\]]*\\]");
+  public static final Pattern PROTO_LINK_PATTERN =
+      Pattern.compile("\\[([^\\]]+)\\]\\[([A-Za-z_][A-Za-z_.0-9]*)*\\]");
```
bjwatson: This doesn't look right. I think what you want here is `\\[([^\\]]+)\\]\\[[A-Za-z_][A-Za-z_.0-9]*]*\\]`.

jmuk: I'm not sure why an asterisk appears after `([A-Za-z_][A-Za-z_.0-9]*)`. Why that?

bjwatson:
@jmuk Like fuzzing?

jmuk:
oh, not saying that random :)

I'd like to recommend to copy&paste some comments which causes this problem from the production proto files (or write something by ourselves).

bjwatson:
Got it. SGTM

src/main/java/com/google/api/codegen/CommentPatterns.java
jmuk: `?` is better if so. `([A-Za-z_][A-Za-z_.0-9]*)?`

geigerj: Actually, I think I like the one we use [in packman](https://github.com/googleapis/packman/blob/master/lib/api_repo.js#L951) best:`([A-Za-z_]+(\\.[A-Za-z_0-9]+)*)`

jmuk: @geigerj I like that, but I'm afraid that it's not strictly correct, it's going to be `([A-Za-z_][A-Za-z_0-9]*(\\.[A-Za-z_][A-Za-z_0-9]*)*)` in that way.

swcloud: @geigerj But that would not allow the second group to be empty

swcloud: Actually there still be problem with the current approach. For example, [A-Za-z_$][a.b.c.d] is considered to be valid proto link, since we allow '$' within the first pair of [] , so a.b.c.d will be removed. So as long as comment string contains [<anything>][a.b.c.d], it will be parsed as a valid proto link, which will have the problem of removing the content within the second pair of bracket. As @jmuk mentioned, #331 may be the right approach. But in the mean time, we will have to live with is imperfect solution.

swcloud:
@jmuk @bjwatson @geigerj I added unit test cases to test how Python/Ruby/NodeJS CommentFixer should handle '$' to avoid throw IAE. These are not comprehensive test cases for the CommentFixers, because I think this PR should focus on related problems. I filed https://github.com/googleapis/toolkit/issues/722 to add more comprehensive tests. I think @geigerj understands best for these markdown link requirements, so I assigned it to him.

swcloud:
@jmuk I updated the description to add more details how this problem is caused.

src/main/java/com/google/api/codegen/CommentPatterns.java
geigerj: > But that would not allow the second group to be empty such as [book][]

As @jmuk mentioned, you can use the `?` operator to achieve that effect.

> For example, [A-Za-z_$][a.b.c.d] is considered to be valid proto link, since we allow '$' within the first pair of []

I think we should treat `[A-Za-z_$][a.b.c.d]` as a valid proto link for the time being. Without any knowledge of the proto namespace we are working with there is no way to distinguish this programmatically from a link to a proto called `a.b.c.d` with display text `A-Za-z_$`.

swcloud:
@jmuk @bjwatson @geigerj PTAL

src/test/java/com/google/api/codegen/ProtoDocumentLinkTest.java
```diff
@@ -0,0 +1,99 @@
...
+    Matcher m;
+    m = CommentPatterns.PROTO_LINK_PATTERN.matcher("[Shelf][google.example.library.v1.Shelf]");
+    Truth.assertThat(m.find()).isTrue();
+    m = CommentPatterns.PROTO_LINK_PATTERN.matcher("[Shelf][]");
+    Truth.assertThat(m.find()).isTrue();
+    m = CommentPatterns.PROTO_LINK_PATTERN.matcher("[$Shelf][]");
+    Truth.assertThat(m.find()).isTrue();
+    m = CommentPatterns.PROTO_LINK_PATTERN.matcher("[][abc]");
+    Truth.assertThat(m.find()).isFalse();
+    m = CommentPatterns.PROTO_LINK_PATTERN.matcher("[A-Za-z_$][A-Za-z_$0-9]");
```
bjwatson: It's kind of confusing to use something that looks like a regex here. `[$Shelf][$abc]` would fit better with the other examples.

geigerj: It's common for proto comments to contain regular expressions that describe the format of string fields, and those can almost look like markdown links if they have two character classes in a row. (That's what triggered this PR originally...) I think it's OK to retain this as a sort of regression test; Song could also add the example you gave.

swcloud: The reason why [A-Za-z_$][A-Za-z_$0-9] is added as a test case is that it is a corner case causing the issue.

bjwatson:
OK, makes sense. Comments will help make this clear.

jmuk:
I was thinking of adding comments to library.proto (so that you can make sure everything is fine in the baseline files on every language).

It is actually not safe as unit tests you wrote, but it could reveal errors on other languages too.

swcloud:
If we want to comprehensive unit test cases and comments, it should be done in #722 separately.

src/test/java/com/google/api/codegen/ProtoDocumentLinkTest.java
```diff
@@ -23,19 +23,25 @@
...
+    // Display name may contain special character '$'
     m = CommentPatterns.PROTO_LINK_PATTERN.matcher("[$Shelf][]");
     Truth.assertThat(m.find()).isTrue();
+    // Display name may NOT be empty
     m = CommentPatterns.PROTO_LINK_PATTERN.matcher("[][abc]");
     Truth.assertThat(m.find()).isFalse();
+    // Fully qualified name may NOT contain special character '$'
     m = CommentPatterns.PROTO_LINK_PATTERN.matcher("[A-Za-z_$][A-Za-z_$0-9]");
     Truth.assertThat(m.find()).isFalse();
+    // Fully qualified name many NOT be numbers only
```
bjwatson: Nit: s/many/may/

swcloud: Thanks!

PR #718, opened 2016-11-03
Support Invalid and Any resource names

michaelbausor:


src/test/java/com/google/api/codegen/testdata/library_gapic.yaml
```diff
@@ -26,11 +26,15 @@ license_header:
 collections:
   - name_pattern: shelves/{shelf_id}
     entity_name: shelf
+invalid_collections:
```
garrettjonesgoogle: I think I'd like to avoid the use of 'invalid' because it can be misinterpreted.

src/main/java/com/google/api/codegen/config/UnformattedResourceNameConfig.java
```diff
@@ -0,0 +1,58 @@
...
+import com.google.api.tools.framework.model.SimpleLocation;
+import com.google.auto.value.AutoValue;
+import javax.annotation.Nullable;
+
+/**
+ * UnformattedResourceNameConfig represents a resource name configuration that accepts a particular
+ * unformatted value.
+ */
+@AutoValue
+public abstract class UnformattedResourceNameConfig implements ResourceNameConfig {
```
garrettjonesgoogle: I'm not sure that "Unformatted" conveys quite what we want. I am currently favoring "Fixed".

src/test/java/com/google/api/codegen/testdata/library_gapic.yaml
michaelbausor: Removed

src/main/java/com/google/api/codegen/config/UnformattedResourceNameConfig.java
michaelbausor: I agree - fixed is better, updated.

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -168,19 +170,27 @@ message CollectionConfigProto {
...
-  // Name to be used as a basis for generated methods.
+  // Name to be used as a basis for generated methods and classes.
   string oneof_name = 1;
 
   // A list of the entity names of the CollectionConfigs to be included in the
   // oneof collection.
   repeated string collection_names = 2;
 }
 
+message FixedCollectionProto {
```
garrettjonesgoogle: This isn't really a "collection". Maybe call this `FixedResourceNameValueProto`. I know you're going to fix the field names and proto names containing "collection" later, but it is an especially bad fit for this case, so we might as well start this one right.

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -90,6 +90,8 @@ message ConfigProto {
   // A list of resource name oneof configs.
   repeated CollectionOneofProto collection_oneofs = 16;
 
+  repeated FixedCollectionProto fixed_collections = 17;
```
garrettjonesgoogle: Maybe call the field `fixed_resource_name_values`.

michaelbausor: Done

src/main/proto/com/google/api/codegen/config.proto
michaelbausor: Done

michaelbausor:
PTAL

garrettjonesgoogle:
LGTM

PR #719, opened 2016-11-03
Java: GAX settings refactoring

garrettjonesgoogle:
GAX refactoring: https://github.com/googleapis/gax-java/pull/132

### DIFF
--- a/src/main/resources/com/google/api/codegen/java/settings.snip
+++ b/src/main/resources/com/google/api/codegen/java/settings.snip
@@ -77,6 +77,14 @@
 @private members(xsettingsClass)
   {@methodMembers(xsettingsClass)}
   {@methodGetters(xsettingsClass)}
+
+  /**
+   * Returns a builder for the default ExecutorProvider for this service.
+   */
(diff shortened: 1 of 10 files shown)


src/main/resources/com/google/api/codegen/java/settings.snip
```diff
@@ -102,6 +110,30 @@
...
+
+  /**
+   * Returns a builder for the default credentials for this service.
+   */
+  public static GoogleCredentialsProvider.Builder defaultCredentialsProviderBuilder() {
+    return GoogleCredentialsProvider.newBuilder()
+        @if xsettingsClass.hasDefaultServiceScopes
+            .setScopesToApply(DEFAULT_SERVICE_SCOPES)
+        @end
+        ;
```
michaelbausor: Nit: I *think* this formatting problem could be fixed by putting the if statement into its own function - not sure if it is worth it...

garrettjonesgoogle: I'm inclined to not fight with snippet formatting since GoogleJavaFormat fixes it anyway.

PR #720, opened 2016-11-03
Fix remaining C# codegen bugs

### LINKED ISSUE #697: C# generation errors
- The using directive for Google.Api.Gax is often missing
- v1beta1 is being capitalized as V1Beta1, contrary to the proto capitalization
- When used as an identifier, event needs to be @event
- Where we used nested types, we need using static instead of using
- Logging tries to import WriteLogEntriesRequest.Types, but that doesn't exist. (Not sure what's going on there.)
- Dictionary parameters need to use collection initializers

To be fixed.

chrisdunelm:
Fixes #697


jskeet:
Looks good - have you tried running the generateapis.sh script with the new code?

chrisdunelm:
Good idea. I've just run generateapis.sh and it all looks good. build.sh builds everything successfully, and I checked Datastore and Logging build OK.

PR #724, opened 2016-11-07
Go: Add InternalFetch

### LINKED ISSUE #706: Make it easier for handwritten clients to wrap Iterator
The motivating example is https://code-review.googlesource.com/#/c/8997/ .
It is much too complicated for handwritten clients to wrap the generated Iterators and maintain idiomatic surface.

pongad:
InternalFetch is a low-level method for accessing a page in paginating
RPCs.
They are only exported when internal_fetch GAPIC option is set to true.

Fixes #706.


src/test/java/com/google/api/codegen/testdata/go_main_library.baseline
```diff
@@ -686,6 +670,13 @@ type BookIterator struct {
     items    []*librarypb.Book
     pageInfo *iterator.PageInfo
     nextFunc func() error
+
+    // internalFetch is for use by the Google Cloud Libraries only.
+    //
+    // internalFetch returns results from a single call to the underlying RPC.
+    // The number of results is no greater than pageSize.
+    // If there are no more results, nextPageToken is empty and err is nil.
+    internalFetch func(pageSize int, pageToken string) (results []*librarypb.Book, nextPageToken string, err error)
```
jba: You have to export it so the handwritten client can access it.

garrettjonesgoogle:
Why do we want to hide this when there is now handwritten client?

src/test/java/com/google/api/codegen/testdata/go_main_library.baseline
pongad: I made the change to export it on a config but forgot to change the config. Done.

pongad:
@garrettjonesgoogle It's the opposite. We only want to expose this when there's a handwritten client. The functionality is low-level and the user shouldn't have to bother. On the other hand, handwritten client might want to wrap our iterators, this makes wrapping easier.

garrettjonesgoogle:
I mistyped, I meant "no", not "now." 

I know the user shouldn't "have to" bother, but is it necessary to hide it? What if someone else wants to make their own iterator in their own client for some reason?

pongad:
@garrettjonesgoogle The [iterator](https://godoc.org/google.golang.org/api/iterator) is not stable. We promise that we won't break the methods on the iterator, but we explicitly state that the "constructor" of the iterator might change in the future. 

InternalFetch is design specifically to help with the current constructor, so we want to discourage users from using it directly, hence the comment "for use by the Google Cloud Libraries only".

garrettjonesgoogle:
But then we still have the problem of exposing something unstable when the handwritten client is present. I would prefer to make sure we have a good strategy for marking code that is known to be unstable, and just be consistent in the gapic clients across the board. @jba do you have an opinion on this point?

jba:
In handwritten Go code we usually just write something like "This method is not part of the stable interface of this package" or something along those lines. That's fine. Not ideal, but fine.

pongad:
@garrettjonesgoogle @jba Ping. I think I have addressed all concerns now.

garrettjonesgoogle:
@pongad , since we appear to have a strategy for marking unstable code (the verbiage by @jba ), I would like to expose this internal fetch functionality in all cases, just with the right comments on it.

pongad:
@garrettjonesgoogle That would probably make generation easier and more consistent. I'm on it.

garrettjonesgoogle:
Since this no longer touches MVVM code, I wash my hands :-)

pongad:
Since @garrettjonesgoogle washed his hands and @jba approved. I'll go ahead and merge.

PR #726, opened 2016-11-07
Multiple fixes to resource name generation

michaelbausor:
- Update surface namer to import specific resource names used in
  construction of oneof resource names
- Update namespaces
- Update get/set functions to include resource name type
- Update baselines


src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -255,21 +251,24 @@ public String getFieldSetFunctionName(TypeRef type, Name identifier) {
     }
   }
 
-  public String getResourceNameFieldSetFunctionName(TypeRef type, Name identifier) {
+  public String getResourceNameFieldSetFunctionName(FieldConfig fieldConfig, TypeRef type) {
```
garrettjonesgoogle: You could get by without the type parameter, and instead use `fieldConfig.getField().getType()`. (It looks like this pairing occurs several times in `SurfaceNamer`)

src/main/java/com/google/api/codegen/transformer/java/JavaModelTypeNameConverter.java
```diff
@@ -227,6 +231,21 @@ public TypeName getTypeNameForTypedResourceName(
     }
   }
 
+  private static String getResourceNamePackage(ProtoFile file, ResourceNameType resourceNameType) {
```
garrettjonesgoogle: My preference would be for the type (resourceNameType) to know what package itself is in (i.e. all of the logic here); that bundles the concept together more coherently. Basically I think the "implicit package" analog in this case is actually inside `ResourceNameType`.

michaelbausor: This runs into the problem that packaging is language-dependent. So it will be necessary to have the JavaModelTypeNameConverter, or some other Java-specific class, involved somewhere in the package construction.

Or alternatively, the ResourceNameConfig could provide a getJavaPackage() method, and other languages would add a (for example) getCSharpPackage() method as required. This seems neater that what we have right now, but a bit of a break with the existing patterns.

michaelbausor:
PTAL

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -39,6 +38,7 @@
 import com.google.api.codegen.viewmodel.SimpleInitValueView;
 import com.google.api.codegen.viewmodel.StructureInitCodeLineView;
 import com.google.api.codegen.viewmodel.testing.GapicSurfaceTestAssertView;
+import com.google.api.tools.framework.model.ProtoFile;
```
garrettjonesgoogle: dead import?

PR #727, opened 2016-11-08
Fix java unit test gen with latest gax.

shinfan:
Tested against api-client-staging repo.


garrettjonesgoogle:
LGTM

shinfan:
I will submit once the travis is ok

PR #728, opened 2016-11-08
Fix C# enum naming

### LINKED ISSUE #725: C# incorrect enum name
See https://github.com/GoogleCloudPlatform/google-cloud-dotnet/commit/6fd7a38cf662e41c53a5fe9532debbedea8bc2fa for manual fix. Needs codegen fix.

"The C# protoc plugin removes a prefix of the enum name from enum values. We need to do the same in the toolkit generator."

chrisdunelm:
Fixes #725


jskeet:
Ah, I think I see now - the code already there, but it was receiving the wrong value, so wasn't trimming properly. If that's the case, LGTM :)

chrisdunelm:
Correct, it wasn't trimming because the enum name was "CommitRequest+Types+Mode", rather than just "Mode"

PR #729, opened 2016-11-08
Minor fixes for resource name list methods

michaelbausor:


garrettjonesgoogle:
LGTM after travis passes

PR #730, opened 2016-11-08
Java gapic: Fixing settings builder with no default instance

garrettjonesgoogle:


src/main/resources/com/google/api/codegen/java/settings.snip
```diff
@@ -457,6 +457,8 @@
   private Builder() {
     @if xsettingsClass.hasDefaultInstance
       super(defaultChannelProviderBuilder().build());
+    @else
+      super((InstantiatingChannelProvider) null);
```
michaelbausor: Is it necessary to cast null here?

garrettjonesgoogle: Yes, because otherwise the call is ambiguous (there is another single-arg constructor that takes ServiceApiSettings).

michaelbausor:
LGTM

PR #734, opened 2016-11-11
Fix issues in smoke test generator

shinfan:
- Fixed the snippet where `sampleMethodCallArgList` was private
- Fixed an issue that `InitCodeTransformer` treated variable values as quoted string
- Fixed the `requireProjectId` method which did not return correct values in some cases.

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
@@ -51,10 +51,12 @@ import com.google.api.codegen.util.testing.TestValueGenerator;
 import com.google.api.codegen.viewmodel.ApiMethodType;
 import com.google.api.codegen.viewmodel.FieldSettingView;
 import com.google.api.codegen.viewmodel.FileHeaderView;
+import com.google.api.codegen.viewmodel.FormattedInitValueView;
 import com.google.api.codegen.viewmodel.InitCodeLineView;
 import com.google.api.codegen.viewmodel.InitCodeView;
 import com.google.api.codegen.viewmodel.ResourceNameInitValueView;
 import com.google.api.codegen.viewmodel.SimpleInitCodeLineView;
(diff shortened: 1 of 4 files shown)


shinfan:
@garrettjonesgoogle  friendly ping

garrettjonesgoogle:
Sorry I missed this. I don't see any updates to baselines - is it hard to come up with a baseline for it?

shinfan:
@garrettjonesgoogle  Yes, it may require some extra changes if we want the baseline to cover all cases. @michaelbausor and I had this dicussion in a previous PR. Currently we only support one smoke test method per interface which means we can only test for one of many cases.

garrettjonesgoogle:
Which real API exercises this functionality?

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
```diff
@@ -185,11 +187,17 @@ private boolean requireProjectId(InitCodeView initCodeView, SurfaceNamer namer)
...
         SimpleInitCodeLineView simpleLine = (SimpleInitCodeLineView) line;
+        String projectVarName =
+            namer.localVarName(Name.from(InitFieldConfig.PROJECT_ID_VARIABLE_NAME));
         if (simpleLine.initValue() instanceof ResourceNameInitValueView) {
           ResourceNameInitValueView initValue = (ResourceNameInitValueView) simpleLine.initValue();
-          return initValue
-              .formatArgs()
-              .contains(namer.localVarName(Name.from(InitFieldConfig.PROJECT_ID_VARIABLE_NAME)));
+          return initValue.formatArgs().contains(projectVarName);
+        } else if (simpleLine.initValue() instanceof SimpleInitValueView) {
```
shinfan: Logging goes into this case

shinfan:
@garrettjonesgoogle  I have marked them inline.

garrettjonesgoogle:
Ok. LGTM

PR #735, opened 2016-11-11
Generalize discovery API overrides to support regex fields

tcoffee-google:


src/test/java/com/google/api/codegen/testdata/discoveries/cloudmonitoring.v2beta2.json.overrides
```diff
@@ -1,13 +1,7 @@
...
   "java": {
     "methods": {
-      "cloudmonitoring.metricDescriptors.list": {
-        "requestBodyType": null
-      },
-      "cloudmonitoring.timeseries.list": {
-        "requestBodyType": null
-      },
-      "cloudmonitoring.timeseriesDescriptors.list": {
+      ".*list": {
```
saicheems: While unlikely, it's possible that they add another list method that is actually correct.
Should we do `cloudmonitoring.(metricDescriptors|timeseries|timeseriesDescriptors).list" instead?

src/test/java/com/google/api/codegen/testdata/discoveries/translate.v2.json.overrides
```diff
@@ -1,17 +1,7 @@
...
-          "q": {
-            "type": {
-              "kind": "TYPE_STRING",
-              "isArray": false
-            }
-          }
-        }
-      },
-      "language.translations.list": {
+      "language.(detections|translations).list": {
```
saicheems: Do the periods here correspond to any character or do they need to be escaped?

src/test/java/com/google/api/codegen/testdata/discoveries/cloudmonitoring.v2beta2.json.overrides
tcoffee-google: I considered it, but thought it more likely such a hypothetical method would follow their existing pattern. Seems like compile check would catch it either way.

src/test/java/com/google/api/codegen/testdata/discoveries/translate.v2.json.overrides
tcoffee-google: They're any character, but the method ID format means they must be periods.

saicheems:
lgtm

PR #737, opened 2016-11-14
Format parse methods

michaelbausor:
This fixes an issue introduced by adding support for collection configs to the top level of the gapic config. Format and parse methods will now be generated in each XApi class based on the collection config for that interface only, instead of based on all collection configs in the gapic yaml.

Have moved the archive collection config to the top level to test that format and parse methods are not generated.


garrettjonesgoogle:
LGTM

PR #738, opened 2016-11-15
Update smoke test and minor fixes.

shinfan:
- Deprecate PageAccessor.
- Fixed an issue that the variable name "projectId" is already taken by
  the `executeNoCatch()` method.
- Fixed an issue of default flattened parameters were initialized correctly.

- Tests passed against api-client-staging

After this PR I will add the generated smoke tests into api-client-staging repo.

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
@@ -236,14 +236,8 @@ public class JavaGapicSurfaceTestTransformer implements ModelToViewTransformer {
             .initFieldConfigStrings(testConfig.getInitFieldConfigStrings())
             .symbolTable(table)
             .fieldConfigMap(fieldConfigMap);
-
     if (context.getMethodConfig().isFlattening()) {
-      ArrayList<Field> initFields = new ArrayList<>();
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
```diff
@@ -217,14 +221,31 @@ private InitCodeContext createSmokeTestInitContext(MethodTransformerContext cont
...
+            .symbolTable(table)
+            .fieldConfigMap(fieldConfigMap);
+
+    if (context.getMethodConfig().isFlattening()) {
+      ArrayList<Field> initFields = new ArrayList<>();
+      for (FieldConfig fieldConfig :
+          context.getFlatteningConfig().getFlattenedFieldConfigs().values()) {
+        initFields.add(fieldConfig.getField());
+      }
+      contextBuilder.initFields(initFields);
```
michaelbausor: You can use context.getFlatteningConfig().getFlattenedFields() instead of creating initFields

shinfan: Good point. Done

shinfan:
@michaelbausor  Travis is failing but it looks like flaky. I will try to upload and trigger it one more time.

PR #740, opened 2016-11-15
C# resource names

chrisdunelm:


chrisdunelm:
Known issues:
* Single resource names and not converted to a one-of resource name; as required in a flattened method that accepts the single, but the request type requires a one-of.
* Snippets for methods that page-stream a resource-name have the wrong type for the local variable.
* Not all resource-name initialization code for snippets is generated.

chrisdunelm:
Jon, Garrett, Michael, please can you take a preliminary look over this.
I'm sure there's loads of things I'm not quite doing right! Thanks.

src/main/resources/com/google/api/codegen/csharp/gapic_resourcenames.snip
```diff
@@ -0,0 +1,252 @@
+@extends "csharp/common.snip"
+
+@snippet generate(resourceNames)
+    {@renderFileHeader(resourceNames.fileHeader)}
+
+    namespace {@resourceNames.fileHeader.packageName}
+    {
+        public interface IResourceName{}// TODO: This will be in GAX
+        public interface ISingleResourceName:IResourceName{} // TODO: Is this required?
```
michaelbausor: We don't have this in Java. It may be useful, I am not sure.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
```diff
@@ -0,0 +1,685 @@
...
+
+namespace Google.Example.Library.V1
+{
+    public interface IResourceName{}// TODO: This will be in GAX
+    public interface ISingleResourceName:IResourceName{} // TODO: Is this required?
+
+    /// <summary>
+    /// Resource name for a shelf.
+    /// </summary>
+    public sealed partial class ShelfName : ISingleResourceName
```
michaelbausor: Were you planning to add a getType() method to these classes?

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicClientTransformer.java
```diff
@@ -63,6 +64,7 @@
 public class CSharpGapicClientTransformer implements ModelToViewTransformer {
```
michaelbausor: Possibly it is worth considering adding a new transformer for the resource name generation, I am not sure.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
```diff
@@ -0,0 +1,685 @@
...
+        /// <summary>
+        /// Constructs a shelf name from its component parts.
+        /// </summary>
+        /// <param name="shelfIdId">The shelfId ID. Must not be <c>null</c>.</param>
+        public ShelfName(string shelfIdId)
+        {
+            ShelfIdId = GaxPreconditions.CheckNotNull(shelfIdId, nameof(shelfIdId));
+        }
+
+        /// <summary>
```
jskeet: Fun option: when C# 7 is out, we may want to add `Deconstruct` methods which would allow 

```csharp
var (projectId, topicId) = topicName;
```

We can add that now, of course, but it won't be very useful yet. Worth doing eventually.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
```diff
@@ -0,0 +1,685 @@
...
+
+        /// <inheritdoc />
+        public override string ToString() => s_template.Expand(ShelfId, BookId, ReturnId);
+    }
+
+    public sealed partial class BookNameOneof : IResourceName
+    {
+        public enum Type
+        {
+            BOOK_NAME = 1,
```
jskeet: The enum values should be idiomatic.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
```diff
@@ -0,0 +1,685 @@
...
+        /// <summary>
+        /// The return ID. Never <c>null</c>.
+        /// </summary>
+        public string ReturnId { get; }
+
+        /// <inheritdoc />
+        public override string ToString() => s_template.Expand(ShelfId, BookId, ReturnId);
+    }
+
+    public sealed partial class BookNameOneof : IResourceName
```
jskeet: Is the name of this controlled in gapic? I think we'll want to experiment with what feels natural here.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
```diff
@@ -0,0 +1,685 @@
...
+        {
+            BookNameOneof result;
+            if (TryParse(name, out result))
+            {
+                return result;
+            }
+            throw new ArgumentException("Invalid name", "name");
+        }
+
+        public static bool TryParse(string name, out BookNameOneof result)
```
jskeet: I thought "fixed sets" would always accept a general resource name as well, so that the sets can be expanded.
Is that just still work to be done?

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
jskeet: XML doc generation TBD for all of this, of course.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
```diff
@@ -0,0 +1,685 @@
...
+            return false;
+        }
+
+        public static BookNameOneof From(BookName bookName) => new BookNameOneof(Type.BOOK_NAME, bookName);
+
+        public static BookNameOneof From(ArchivedBookName archivedBookName) => new BookNameOneof(Type.ARCHIVED_BOOK_NAME, archivedBookName);
+
+        public static BookNameOneof From(DeletedBookNameFixed deletedBookNameFixed) => new BookNameOneof(Type.DELETED_BOOK_NAME_FIXED, deletedBookNameFixed);
+
+        public BookNameOneof(Type type, IResourceName value)
```
jskeet: Validate that `type` is a member of the enum?
Ideally, validate that the type of the value corresponds to the type, as well... we could do that via reflection, having an attribute on each enum value. Definitely a later enhancement.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
```diff
@@ -0,0 +1,685 @@
...
+        public BookName BookName => CheckAndReturn<BookName>(Type.BOOK_NAME);
+
+        public ArchivedBookName ArchivedBookName => CheckAndReturn<ArchivedBookName>(Type.ARCHIVED_BOOK_NAME);
+
+        public DeletedBookNameFixed DeletedBookNameFixed => CheckAndReturn<DeletedBookNameFixed>(Type.DELETED_BOOK_NAME_FIXED);
+
+        public override string ToString() => _value.ToString();
+    }
+
+    public sealed partial class DeletedBookNameFixed : IResourceName
```
jskeet: Would a base type be useful here? It would avoid code duplication - not sure how important that is. (A separate helper class could do that too, I suspect.)

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
```diff
@@ -0,0 +1,685 @@
...
+        /// <summary>
+        /// A <see cref="IEnumerable<BookName>"/>-typed view over the <see cref="Names"/> resource-name property.
+        /// </summary>
+        public IEnumerable<BookName> NamesAsResources
+        {
+            get
+            {
+                return Names.Select(x => BookName.Parse(x));
+            }
+            set
```
jskeet: The introduction of a setter here (when there isn't one for the collection itself) feels a little unnatural.

It might be better to expose an `IList<T>` here instead, which just delegates to the underlying collection. We could create a new one each time the getter is called, or add a field (at the cost of memory) and use that all the time. Let's talk about this offline.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
```diff
@@ -0,0 +1,685 @@
...
+        public override string ToString() => FixedValue;
+    }
+
+
+    public partial class FindRelatedBooksResponse
+    {
+        /// <summary>
+        /// A <see cref="IEnumerable<BookName>"/>-typed view over the <see cref="Names"/> resource-name property.
+        /// </summary>
+        public IEnumerable<BookName> NamesAsResources
```
jskeet: I'm not keen on "AsResource[s]" as a suffix yet. Let's toss around a few ideas for that. In particular, it's the name as a resource *name*, not as an actual resource.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
chrisdunelm: Agreed this is better, I've implemented creating a new one each time. Can discuss further as required.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicClientTransformer.java
chrisdunelm: I'll leave it as-is for now.

chrisdunelm:
PTAL

src/main/java/com/google/api/codegen/transformer/PathTemplateTransformer.java
```diff
@@ -43,6 +63,150 @@
...
+      index += 1;
+    }
+    return resourceNames;
+  }
+
+  private ResourceNameSingleView generateResourceNameSingle(
+      SurfaceTransformerContext context, int index, SingleResourceNameConfig config) {
+    SurfaceNamer namer = context.getNamer();
+    ResourceNameSingleView.Builder builder = ResourceNameSingleView.newBuilder();
+    builder.typeName(namer.getResourceTypeName(config));
```
garrettjonesgoogle: Indeed there is... I have been using the non-fluent api in general, since many builders have quite a few parameters. I personally haven't been forcing anyone either way.

src/main/java/com/google/api/codegen/transformer/PathTemplateTransformer.java
```diff
@@ -46,6 +66,155 @@
...
+      Collection<Field> fields = new ArrayList<Field>(entry.getValue());
+      ResourceProtoView.Builder protoBuilder = ResourceProtoView.newBuilder();
+      protoBuilder.protoClassName(namer.getTypeNameConverter().getTypeName(msgName).getNickname());
+      List<ResourceProtoFieldView> fieldViews = new ArrayList<>();
+      for (Field field : fields) {
+        String fieldName = field.getSimpleName();
+        FieldConfig fieldConfig = FieldConfig.createDefaultFieldConfig(field);
+        String fieldResourceName = resourceConfigs.getFieldResourceName(field);
+        String fieldTypeSimpleName;
+        if (fieldResourceName.equals("*")) {
```
michaelbausor: Maybe use AnyResourceNameConfig.GAPIC_CONFIG_ANY_VALUE here?

michaelbausor: Maybe use AnyResourceNameConfig.GAPIC_CONFIG_ANY_VALUE here?

src/test/java/com/google/api/codegen/testdata/csharp_gapic_client_library.baseline
```diff
@@ -3983,22 +3983,22 @@ namespace Google.Example.Library.V1
         /// <returns>
         /// A pageable asynchronous sequence of <see cref="string"/> resources.
         /// </returns>
-        public override IPagedAsyncEnumerable<FindRelatedBooksResponse, string> FindRelatedBooksAsync(
-            IEnumerable<string> names,
-            IEnumerable<string> shelves,
+        public override IPagedAsyncEnumerable<FindRelatedBooksResponse, BookName> FindRelatedBooksAsync(
```
michaelbausor: Was the decision for csharp to always iterate over the resource name in this case? In Java we still default to string, but the return object has an iterateAllAsBookName() method.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
```diff
@@ -0,0 +1,719 @@
...
+        {
+            Value = GaxPreconditions.CheckNotNull(value, nameof(value));
+        }
+
+        public string Value { get; }
+
+        public override string ToString() => Value;
+    }
+
+    public sealed class ResourceNameList<TResource> : IList<TResource> where TResource : IResourceName // TODO: This will be in GAX
```
jskeet: This is in GAX so not as relevant, but let's call it TName rather than TResource. I'd like to maintain a clear separation between resources and resource names wherever possible.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
jskeet: Not quite sure what's meant, but I can't see anything here at the moment...

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
```diff
@@ -0,0 +1,719 @@
...
+        /// </summary>
+        /// <param name="shelfName">The shelf resource name in string form. Must not be <c>null</c>.</param>
+        public static ShelfName Parse(string shelfName)
+        {
+            GaxPreconditions.CheckNotNull(shelfName, nameof(shelfName));
+            ResourceName resourceName = s_template.ParseName(shelfName);
+            return new ShelfName(resourceName[0]);
+        }
+
+        public static bool TryParse(string shelfName, out ShelfName result)
```
jskeet: XML doc

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
```diff
@@ -0,0 +1,719 @@
...
+        /// The shelfId ID. Never <c>null</c>.
+        /// </summary>
+        public string ShelfIdId { get; }
+
+        /// <inheritdoc />
+        public override string ToString() => s_template.Expand(ShelfIdId);
+    }
+
+    /// <summary>
+    /// Resource name for a archived_book.
```
jskeet: a archived_book?

- A/an (annoying to fix; can we reword to avoid it being a problem?)
- Is archived_book meant to refer to something the user would know about? Maybe we need a configurable documentation name

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
```diff
@@ -0,0 +1,719 @@
...
+        public static ShelfName Parse(string shelfName)
+        {
+            GaxPreconditions.CheckNotNull(shelfName, nameof(shelfName));
+            ResourceName resourceName = s_template.ParseName(shelfName);
+            return new ShelfName(resourceName[0]);
+        }
+
+        public static bool TryParse(string shelfName, out ShelfName result)
+        {
+            GaxPreconditions.CheckNotNull(shelfName, nameof(shelfName));
```
jskeet: All the redundancy here feels annoying. I suspect we can get this down to

```csharp
public static bool TryParse(string shelfName, out ShelfName result) =>
    s_template.TryParseName(shelfName, nameof(shelfName), out result, p0 => new ShelfName(p0));
```

... and likewise for up to (say) 4 parameters. Can definitely do that later though.

Ditto for `Parse`:

```csharp
public static ShelfName Parse(string shelfName) =>
    s_template.ParseName(shelfName, nameof(shelfName), p0 => new ShelfName(p0));
```


(With GAX changes, of course.)

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
```diff
@@ -0,0 +1,719 @@
...
+        }
+        public IEnumerator<TResource> GetEnumerator() => _underlyingList.Select(_stringToName).GetEnumerator();
+        System.Collections.IEnumerator System.Collections.IEnumerable.GetEnumerator() => GetEnumerator();
+        public int IndexOf(TResource value) => _underlyingList.IndexOf(_nameToString(value));
+        public void Insert(int index, TResource item) => _underlyingList.Insert(index, _nameToString(item));
+        public bool Remove(TResource item) => _underlyingList.Remove(_nameToString(item));
+        public void RemoveAt(int index) => _underlyingList.RemoveAt(index);
+    }
+
+    /// <summary>
```
jskeet: There's no qualified version here. I can't remember what we were going to do about those.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
jskeet: Ping on this one.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
```diff
@@ -0,0 +1,719 @@
...
+
+        public static BookNameOneof From(BookName bookName) => new BookNameOneof(OneofType.BookName, bookName);
+
+        public static BookNameOneof From(ArchivedBookName archivedBookName) => new BookNameOneof(OneofType.ArchivedBookName, archivedBookName);
+
+        public static BookNameOneof From(DeletedBookNameFixed deletedBookNameFixed) => new BookNameOneof(OneofType.DeletedBookNameFixed, deletedBookNameFixed);
+
+        public BookNameOneof(OneofType type, IResourceName value)
+        {
+            Type = GaxPreconditions.CheckEnumValue<OneofType>(type, nameof(value));
```
jskeet: nameof(type)

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
```diff
@@ -0,0 +1,719 @@
...
+            return false;
+        }
+
+        public static BookNameOneof From(BookName bookName) => new BookNameOneof(OneofType.BookName, bookName);
+
+        public static BookNameOneof From(ArchivedBookName archivedBookName) => new BookNameOneof(OneofType.ArchivedBookName, archivedBookName);
+
+        public static BookNameOneof From(DeletedBookNameFixed deletedBookNameFixed) => new BookNameOneof(OneofType.DeletedBookNameFixed, deletedBookNameFixed);
+
+        public BookNameOneof(OneofType type, IResourceName value)
```
jskeet: This doesn't validate that the type we give is actually sane, of course. I may have mentioned elsewhere that we could check that.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
```diff
@@ -0,0 +1,719 @@
...
+        {
+            Type = GaxPreconditions.CheckEnumValue<OneofType>(type, nameof(value));
+            Value = GaxPreconditions.CheckNotNull(value, nameof(value));
+        }
+
+        public OneofType Type { get; }
```
jskeet: In protos I think we use "case" rather than "type" here. It's only a little inconsistency, so I don't think it's awful, but worth considering.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
```diff
@@ -0,0 +1,719 @@
...
+        {
+            DeletedBookNameFixed result;
+            if (!TryParse(deletedBookNameFixed, out result))
+            {
+                throw new ArgumentException($"Invalid resource name, must be \"{FixedValue}\"", nameof(deletedBookNameFixed));
+            }
+            return result;
+        }
+
+        public static bool TryParse(string deletedBookNameFixed, out DeletedBookNameFixed result)
```
jskeet: This could definitely be a helper method in GAX.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
```diff
@@ -0,0 +1,719 @@
...
+    public partial class FindRelatedBooksResponse
+    {
+        /// <summary>
+        /// A <see cref="IList<BookName>"/>-typed view over the <see cref="Names"/> resource-name property.
+        /// </summary>
+        public IList<BookName> NamesAsResources =>
+            new ResourceNameList<BookName>(Names,
+                name => name.ToString(),
+                str => BookName.Parse(str));
+
```
jskeet: How easy is it to remove the extra line at the end of each of these partials?

src/main/resources/com/google/api/codegen/csharp/gapic_resourcenames.snip
```diff
@@ -0,0 +1,307 @@
...
+
+        public sealed class UnknownResourceName : IResourceName // TODO: This will be in GAX
```
jskeet: Hmm... this doesn't guarantee that it's a valid resource name at all, of course, in terms of reserved characters etc. We can tighten that aspect up later, of course - and make sure that the XML documentation only guarantees that TryParse will return true when it's a valid resource name, etc.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_client_library.baseline
```diff
@@ -4024,22 +4024,22 @@ namespace Google.Example.Library.V1
...
             string pageToken = null,
             int? pageSize = null,
             CallSettings callSettings = null)
         {
             FindRelatedBooksRequest request = new FindRelatedBooksRequest
             {
-                Names = { names },
-                Shelves = { shelves },
+                BookNames = { names },
+                ShelvesAsShelfName = { shelves },
```
jskeet: This should be ShelvesAsShelfNames

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
```diff
@@ -0,0 +1,965 @@
...
+        /// <returns>The parsed <see cref="ShelfName"/> if successful.</returns>
+        public static ShelfName Parse(string shelfName)
+        {
+            GaxPreconditions.CheckNotNull(shelfName, nameof(shelfName));
+            TemplatedResourceName resourceName = s_template.ParseName(shelfName);
+            return new ShelfName(resourceName[0]);
+        }
+
+        /// <summary>
+        /// Tries to parse shelf resource name in string form into a <see cref="ShelfName"/>/
```
jskeet: Would be nicer as "Tries to parse a" but then we need to work out a/an.
Mind you, that exists for the "into a" as well.
Non-blocking, certainly.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
```diff
@@ -0,0 +1,965 @@
...
+            }
+            else
+            {
+                result = null;
+                return false;
+            }
+        }
+
+        /// <summary>
+        /// Constructs a archived_book name from its component parts.
```
jskeet: Eww. Again, non-blocking but let's revisit the exact wording later to try to avoid needing a/an.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
```diff
@@ -0,0 +1,965 @@
...
+            BookIdId = GaxPreconditions.CheckNotNull(bookIdId, nameof(bookIdId));
+        }
+
+        /// <summary>
+        /// The archivePath ID. Never <c>null</c>.
+        /// </summary>
+        public string ArchivePathId { get; }
+
+        /// <summary>
+        /// The bookId ID. Never <c>null</c>.
```
jskeet: We could potentially be clever and omit the " ID" part if the name ends with "Id"

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
```diff
@@ -0,0 +1,965 @@
...
+
+            /// <summary>
+            /// A resource of type 'archived_book'.
+            /// </summary>
+            ArchivedBookName = 2,
+
+            /// <summary>
+            /// A resource of type 'deleted_book'.
+            /// </summary>
+            DeletedBookNameFixed = 3,
```
jskeet: Does this need the Fixed suffix here? Don't mind much either way.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
```diff
@@ -0,0 +1,965 @@
...
+        /// <param name="deletedBookNameFixed">The <see cref="DeletedBookNameFixed"/> to be contained within
+        /// the returned <see cref="BookNameOneof"/>. Must not be <c>null</c>.</param>
+        /// <returns>A new <see cref="BookNameOneof"/>, containing <paramref name="deletedBookNameFixed"/>.</returns>
+        public static BookNameOneof From(DeletedBookNameFixed deletedBookNameFixed) => new BookNameOneof(OneofType.DeletedBookNameFixed, deletedBookNameFixed);
+
+        private static bool IsValid(OneofType type, IResourceName name)
+        {
+            switch (type)
+            {
+            case OneofType.Unknown: return true; // Anything goes with Unknown.
```
jskeet: Nit: Indent more.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
```diff
@@ -0,0 +1,965 @@
...
+        /// A <see cref="IList<BookName>"/>-typed view over the <see cref="Names"/> resource-name property.
+        /// </summary>
+        public IList<BookName> BookNames =>
+            new ResourceNameList<BookName>(Names,
+                str => BookName.Parse(str));
+
+        /// <summary>
+        /// A <see cref="IList<ShelfName>"/>-typed view over the <see cref="Shelves"/> resource-name property.
+        /// </summary>
+        public IList<ShelfName> ShelvesAsShelfName =>
```
jskeet: As before, this should be ShelvesAsShelfNames.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
```diff
@@ -0,0 +1,965 @@
...
+        public IList<BookName> BookNames =>
+            new ResourceNameList<BookName>(Names,
+                str => BookName.Parse(str));
+
+    }
+
+    public partial class GetBookFromAnywhereRequest
+    {
+        /// <summary>
+        /// A <see cref="BookNameOneof"/>-typed view over the <see cref="Name"/> resource-name property.
```
jskeet: I see lingering resource-name...

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
```diff
@@ -0,0 +1,965 @@
...
+        }
+
+    }
+
+    public partial class Shelf
+    {
+        /// <summary>
+        /// A <see cref="ShelfName"/>-typed view over the <see cref="Name"/> resource-name property.
+        /// </summary>
+        public ShelfName ShelfName
```
jskeet: I really like this - much better than NameAsShelfName.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
chrisdunelm: Bug filed: #784

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
chrisdunelm: Bug filed: #784

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
chrisdunelm: It's more complex to remove it. I think I'll leave it for now.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
chrisdunelm: Ack :)

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
chrisdunelm: Fixed

src/test/java/com/google/api/codegen/testdata/csharp_gapic_client_library.baseline
chrisdunelm: Done

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
chrisdunelm: Done

PR #741, opened 2016-11-15
Support runtime random values in smoke tests

shinfan:


src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceNamer.java
```diff
@@ -144,4 +147,25 @@ public String getPagedResponseTypeInnerName(
   public String getFullyQualifiedApiWrapperClassName(Interface service) {
     return getPackageName() + "." + getApiWrapperClassName(service);
   }
+
+  @Override
+  public String getRandomStringValue(String randomString) {
+    String randomValueGen = "System.currentTimeMillis()";
+    String delimiter = ",";
+    String[] splitted =
```
michaelbausor: Nit: name can be just "split"

src/main/java/com/google/api/codegen/metacode/InitValue.java
```diff
@@ -20,8 +20,9 @@
 @AutoValue
 public abstract class InitValue {
   public enum InitValueType {
-    Literal,
-    Variable,
+    Literal, // Literal values such as strings and numbers
```
michaelbausor: Nit: use doc style:

/** Comment */
EnumValue

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceNamer.java
```diff
@@ -144,4 +147,25 @@ public String getPagedResponseTypeInnerName(
...
+        FieldStructureParser.stripQuotes(randomString)
```
michaelbausor: Can you strip quotes earlier, perhaps in the InitFieldConfig, so you don't need to reference the FieldStructureParser from the transformer?

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceNamer.java
```diff
@@ -33,6 +33,7 @@
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import org.apache.commons.lang3.StringUtils;
```
michaelbausor: Maybe use com.google.common.base.Joiner to avoid the apache commons reference?

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceNamer.java
```diff
@@ -144,4 +147,25 @@ public String getPagedResponseTypeInnerName(
...
+    for (String token : splitted) {
+      if (token.length() > 0) {
+        if (token.equals(randomValueGen)) {
+          stringParts.add(token);
+        } else {
+          stringParts.add("\"" + token + "\"");
+        }
+      }
+    }
+    return String.join(" + ", stringParts);
```
garrettjonesgoogle: I think I'd be more comfortable with `randomValueGen` being injected later than you do - i.e. in your `replace(...)` call, use RANDOM_TOKEN instead of `randomValueGen`, and then when you are adding to `stringParts`, use `"System.currentTimeMillis()"` there.

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceNamer.java
shinfan: This is general inject method in `JavaSurfaceNamer` so it would be risky to make such an assumption on the input that it is not quoted.  I would prefer to keep the quote stripping here for robustness.

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceNamer.java
shinfan: Done

src/main/java/com/google/api/codegen/metacode/InitValue.java
shinfan: Done

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceNamer.java
shinfan: Done

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceNamer.java
```diff
@@ -144,4 +147,25 @@ public String getPagedResponseTypeInnerName(
   public String getFullyQualifiedApiWrapperClassName(Interface service) {
     return getPackageName() + "." + getApiWrapperClassName(service);
   }
+
+  @Override
+  public String getRandomStringValue(String randomString) {
```
shinfan: I will take this name. Done.

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceNamer.java
```diff
@@ -150,23 +150,24 @@ public String getFullyQualifiedApiWrapperClassName(Interface service) {
...
+                InitFieldConfig.RANDOM_TOKEN, delimiter + InitFieldConfig.RANDOM_TOKEN + delimiter)
             .split(delimiter);
     ArrayList<String> stringParts = new ArrayList<>();
-    for (String token : splitted) {
+    for (String token : split) {
       if (token.length() > 0) {
-        if (token.equals(randomValueGen)) {
-          stringParts.add(token);
+        if (token.equals(InitFieldConfig.RANDOM_TOKEN)) {
+          stringParts.add(randomValueGen);
```
garrettjonesgoogle: I think it's better to have the hardcoded string here over making a one-use variable higher up.

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceNamer.java
garrettjonesgoogle: The problem is more with the dependency on `FieldStructureParser` from `JavaSurfaceNamer`. The `stripQuotes` function should probably be moved to a more general place.

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceNamer.java
shinfan: Done

garrettjonesgoogle:
LGTM once we either fix the Travis failure or determine it's a non-issue.

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceNamer.java
michaelbausor: I don't think we want to avoid apache commons - just that we should try to stick to a single commons library for consistency unless we really need both

garrettjonesgoogle: +1 - Generally, a lot of apache commons functionality is also available in Google Guava, so it's silly to split our usage across both.

shinfan:
@saicheems The Travis still seems very flaky. Is there any update on this?

saicheems:
The custom baseline is one of the next things on my timeline, sorry for the issues :/

PR #742, opened 2016-11-15
Fix Go methodNameComponent capitalization

saicheems:


tcoffee-google:
Will there be baseline changes? LGTM in the abstract.

saicheems:
Actually... this is totally wrong... What's happening is dns/v2beta1 is generating

`type Service {
    ...
    Dnskeys ...
}`

but no other API fails like this
I'm going to assume that the client library is out of date actually, I'll double check and reopen this if there's actually an issue

PR #743, opened 2016-11-16
Import common dependencies

shinfan:
**Proposal**

Since google-java-formatter now removes unused imports. It would be
convenient if we just import common dependencies such as
java.util.* and remove unused ones in the post-process.


src/test/java/com/google/api/codegen/testdata/java_mock_service_impl_library.baseline
```diff
@@ -23,6 +23,7 @@ import com.google.tagger.v1.AddLabelResponse;
 import com.google.tagger.v1.LabelerGrpc.LabelerImplBase;
 import io.grpc.stub.StreamObserver;
 import java.util.ArrayList;
+import java.util.Arrays;
```
michaelbausor: Can we apply google-java-format to these baselines?

shinfan: Unfortunately no, the formatter only supports java files. This is the drawback of this approach that the baseline files will contain some unnecessary imports.

shinfan:
This is a proposal. I think the benefit is to take advantage of the formatter and make the import management in java a bit easier. But the drawback is to have some unused imports in the baseline.

src/test/java/com/google/api/codegen/testdata/java_mock_service_impl_library.baseline
michaelbausor: I just had a look at the google java formatter readme (https://github.com/google/google-java-format) and I think the formatter can be used as a library in java code. So we could apply the format to the code generated by toolkit. This would let us:
- not have extra import in the baseline
- optionally, we could remove the formatting step from artman. This means artman has fewer non-python dependencies, and our toolkit-generated code doesn't get changed by anything (at least in Java)

michaelbausor: On the downside, this will be different than all the other languages

garrettjonesgoogle:
Here's a scenario where this breaks down, and why I haven't done this already:

1. Some API has a Queue message in it (for this example, let's say the package is com.google.spiffyapi.v1).
2. Since java.util.Queue is already imported, Queue gets fully qualified (com.google.spiffyapi.v1.Queue).
3. The unused import (java.util.Queue) gets cleaned out
4. The only thing left is an unnecessary fully-qualified reference: com.google.spiffyapi.v1.Queue

shinfan:
Thanks for the feedback. @garrettjonesgoogle @michaelbausor  I think maybe keep what we have right is the better approach. I will close this PR.

Issue #270, opened 2016-06-27
Use the service YAML `title:` as the API name in generated artifacts


bjwatson:
What:
The current README auto-generates the service title, which doesn't always look right. For example, "Google Pubsub API" in http://pythonhosted.org/gax-google-pubsub-v1/ should be "Google Cloud Pub/Sub API", per https://github.com/googleapis/googleapis/blob/master/google/pubsub/v1/pubsub.yaml#L7.

We can also see this in package repositories:
- [Python](https://pypi.python.org/pypi/gapic-google-cloud-logging-v2): now correct
- [Java](https://search.maven.org/#artifactdetails%7Ccom.google.cloud%7Cgoogle-cloud-logging%7C1.3.1%7Cjar): needs to be fixed

Why:
Use the correct name for our APIs. Even more importantly, this impacts our branding for the [Stackdriver family of APIs](https://cloud.google.com/stackdriver/docs/).
- [x] C#
- [ ] Go
- [ ] Java
- [ ] NodeJS
- [ ] PHP
- [x] Python
- [x] Ruby

geigerj:
As part of this, let's ensure that the source and usage of the API name/title is consistent across all of our tools. I think `artman`, `packman`, and `toolkit` all have some sort of concept of API name -- we should ensure these do not vary without justification, and eliminate any duplicate concepts (for example, maybe the `api_name` field of the Artman config/`pipeline_args`).

bjwatson:
@geigerj Has this been partially done?

geigerj:
@bjwatson This is done for Python. (Just checked the box.)

bjwatson:
Thanks @geigerj!

eoogbe:
@bjwatson Is this still relevant?

bjwatson:
@eoogbe This needs to be evaluated by the language leads for the remaining languages.

@geigerj Why do http://pythonhosted.org/gax-google-pubsub-v1/ and http://pythonhosted.org/gax-google-logging-v2/ show the wrong names? They should be "Google Cloud Pub/Sub API" and "Stackdriver Logging API", respectively.

eoogbe:
cc/ @jskeet @pongad @garrettjonesgoogle @lukesneeringer @michaelbausor @geigerj Language leads for remaining languages

geigerj:
[...]

PR #744, opened 2016-11-16
Start porting Packman to Toolkit as PackageMetadataGenerator

geigerj:
This change brings the functionality of Packman for Python gRPC package metadata generation to Toolkit.

It consists of a new tool, PackageMetadataGenerator, that runs two phases. In the first phase, raw gRPC source code is copied to the final destination. The tool may operate on the source code or compute information based on its contents during this phase. The results of the first phase are passed to the second phase, which consists of standard snippet template rendering to generate top-level package metadata.

PackageMetadataGenerator preserves the Packman configuration format and template structure to simplify porting additional languages and package types.


src/main/java/com/google/api/codegen/metadatagen/PackageMetadataContext.java
```diff
@@ -0,0 +1,125 @@
...
+import java.nio.file.Paths;
+import java.util.HashMap;
+import java.util.Map;
+import org.yaml.snakeyaml.Yaml;
+
+/*
+ * A Context object to be passed into the snippet template rendering phase of package metadata
+ * generation.
+ */
+public class PackageMetadataContext extends CodegenContext {
```
garrettjonesgoogle: I recommend against re-using CodegenContext. It is the root of the unmanageable pre-MVVM context hierarchy.

geigerj: Done.

geigerj:
Note: Travis CI failures are for the test `JavaNoPathTemplatesCodeGeneratorTest`. @garrettjonesgoogle do you know if this could be connected to this change? I would suspect not, since this PR is purely additive.

src/main/resources/com/google/api/codegen/metadatagen/py/MANIFEST.in.snip
```diff
@@ -0,0 +1,8 @@
+@snippet generateFilename()
+    MANIFEST.in
+@end
+
+@snippet generateBody(model)
+    include README.rst LICENSE
+
+@end
```
garrettjonesgoogle: add newline

src/main/java/com/google/api/codegen/metadatagen/py/PythonPackageCopier.java
```diff
@@ -0,0 +1,120 @@
...
+    public PythonPackageFileVisitor(Path inputPath, String apiVersion) {
+      this.inputPath = inputPath;
+      this.apiVersion = apiVersion;
+    }
+
+    @Override
+    public FileVisitResult visitFile(Path file, BasicFileAttributes attr) throws IOException {
+      String outFile = inputPath.relativize(file).toString();
+      docBuilder.put(
+          outFile, Doc.text(new String(Files.readAllBytes(file), StandardCharsets.UTF_8)));
```
garrettjonesgoogle: I'm concerned about the fact that it looks like this loads the entire contents of all the files in the input directory tree into memory and the entirety of this content is passed onto the next step.

src/main/java/com/google/api/codegen/metadatagen/PackageMetadataSnippetSetRunner.java
```diff
@@ -0,0 +1,36 @@
...
+  public GeneratedResult generate(
+      Model model, String snippetFileName, PackageMetadataContext context) {
+    PackageMetadataSnippetSet snippets =
+        SnippetSet.createSnippetInterface(
+            PackageMetadataSnippetSet.class,
+            SnippetSetRunner.SNIPPET_RESOURCE_ROOT + "/metadatagen",
+            snippetFileName,
+            ImmutableMap.<String, Object>of("context", context));
+    return GeneratedResult.create(
+        snippets.generateBody(model), snippets.generateFilename().prettyPrint());
```
garrettjonesgoogle: Looking over your snippets, I see that none of them reference `model`. You could convert this to use the same snippet-running classes as MVVM without even changing the content of your snippets (only the scaffolding). 

* Have PackageMetadataContext implement ViewModel
* Use CommonSnippetSetRunner
* Change the snippets to have a single entry point, `@snippet generate(context)` (where `context` is the `PackageMetadataContext` object)

src/main/java/com/google/api/codegen/metadatagen/py/PythonPackageCopier.java
```diff
@@ -0,0 +1,120 @@
...
+    List<String> pythonNamespacePackages = visitor.getNamespacePackages();
+    ImmutableMap.Builder<String, Doc> docBuilder = visitor.getDocBuilder();
+
+    StringBuilder namespacePackages = new StringBuilder("[");
+    for (String namespacePackage : pythonNamespacePackages) {
+      namespacePackages.append("'" + namespacePackage + "', ");
+    }
+    namespacePackages.append("]");
+
+    docBuilder.put(copierMetadataKey, Doc.text(namespacePackages.toString()));
```
garrettjonesgoogle: Why not just create a simple value type that contains `Map<String,Doc>` and the namespace packages? Heterogeneous key/value bags are extremely difficult to maintain because it's really hard to trace how things are injected.

src/main/java/com/google/api/codegen/metadatagen/ApiNameInfo.java
```diff
@@ -0,0 +1,38 @@
...
+/** Represents naming information connected to a particular API. */
+@AutoValue
+public abstract class ApiNameInfo {
+  static AutoValue_ApiNameInfo create(
+      String fullName, String shortName, String majorVersion, String path) {
+    return new AutoValue_ApiNameInfo(fullName, shortName, majorVersion, path);
+  }
+
+  /** The full name of the API, including branding. E.g., "Stackdriver Logging". */
+  abstract String fullName();
```
garrettjonesgoogle: make the methods here public

src/main/java/com/google/api/codegen/metadatagen/PackageMetadataContext.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+    return (String) current.get(keys[keys.length - 1]);
+  }
+
+  public ApiNameInfo getApiNameInfo() {
+    return apiNameInfo;
+  }
+
+  // TODO (geigerj): accessing apiNameInfo directly in the snippets leads to errors, hence these
+  // aliases in the context. Investigate whether these are truly needed.
```
garrettjonesgoogle: What kind of errors? Is it due to the fact that the methods in ApiNameInfo are package-private?

src/main/resources/com/google/api/codegen/metadatagen/py/setup.py.snip
```diff
@@ -0,0 +1,50 @@
...
+    https://packaging.python.org/en/latest/distributing.html
+    https://github.com/pypa/sampleproject
+    """
+
+    import setuptools
+
+    from setuptools import setup, find_packages
+
+    install_requires = [
+      'oauth2client>={@context.getDependencyValue("auth:python:version")}, <{@context.getDependencyValue("auth:python:next_version")}',
```
garrettjonesgoogle: I feel like it would be more natural to use a dotted path instead of a coloned path.

src/main/java/com/google/api/codegen/metadatagen/PackageMetadataContext.java
```diff
@@ -0,0 +1,124 @@
...
+    return getUncheckedMapValue(defaultsMap, keySpecifier);
+  }
+
+  @SuppressWarnings("unchecked")
+  // Note: Snippets doesn't seem to support variadic args, so we use ":" as a separator.
+  //
+  // TODO (geigerj): this is a hack to port the dynamic Javascript configuration from Packman to
+  // Java. To hew closer to Java style, investigate defining configuration information in a
+  // well-defined format that can be converted properly to Java types.
+  private String getUncheckedMapValue(Map<String, Object> map, String keySpecifier) {
```
garrettjonesgoogle: This could potentially have a better name that reflects the recursive or path nature of the key.

src/main/java/com/google/api/codegen/metadatagen/ApiNameInfo.java
```diff
@@ -0,0 +1,38 @@
...
+
+  /** A single-word short name of the API. E.g., "logging". */
+  abstract String shortName();
+
+  /** The major version of the API, as used in the package name. E.g., "v1". */
+  abstract String majorVersion();
+
+  /** The path to the API protos in the googleapis repo. */
+  abstract String path();
```
garrettjonesgoogle: maybe call this `protoPath`?

src/main/java/com/google/api/codegen/metadatagen/ApiNameInfo.java
geigerj: Done.

src/main/java/com/google/api/codegen/metadatagen/ApiNameInfo.java
geigerj: Done.

src/main/java/com/google/api/codegen/metadatagen/PackageMetadataContext.java
geigerj: Renamed to "nestedGet".

src/main/java/com/google/api/codegen/metadatagen/py/PythonPackageCopier.java
geigerj: Originally I thought this would be nice because in `PackageMetadataGenerator`, the `Map<String, Doc>` of results would contain the full output of the tool, which is analogous to many of the other API Compiler-based tools; it makes the PackageCopier phase "functional" rather the having the side-effect of actually altering the filesystem. I can see that this actually would be pretty bad if you pointed `inputPath` to the wrong directory though. I'll change this to do the copying here.

src/main/java/com/google/api/codegen/metadatagen/py/PythonPackageCopier.java
geigerj: Agreed. Will make this change in a subsequent commit.

src/main/resources/com/google/api/codegen/metadatagen/py/MANIFEST.in.snip
geigerj: Done.

src/main/resources/com/google/api/codegen/metadatagen/py/setup.py.snip
geigerj: Done.

src/main/java/com/google/api/codegen/metadatagen/PackageMetadataSnippetSetRunner.java
garrettjonesgoogle: MVVM is very modular. The only thing I'm having you use here is the pattern for invoking snippets. Your context class (`PackageMetadataContext`) can still hold a `model` as a variable and indirectly invoke it.

geigerj: Done.

src/main/java/com/google/api/codegen/metadatagen/py/PythonPackageCopier.java
garrettjonesgoogle: A subsequent commit on this PR or a separate PR? If a separate PR, then add a TODO.

geigerj: Sorry,  meant on this PR. Taking a look at a few other things this afternoon, then will come back to this.

geigerj: Done.

src/main/java/com/google/api/codegen/metadatagen/py/PythonPackageCopier.java
geigerj: Done.

src/main/java/com/google/api/codegen/metadatagen/PackageMetadataSnippetSet.java
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen.metadatagen;
+
+import com.google.api.tools.framework.model.Model;
+import com.google.api.tools.framework.snippet.Doc;
+
+public interface PackageMetadataSnippetSet {
```
garrettjonesgoogle: You can delete this now

src/main/java/com/google/api/codegen/metadatagen/PackageMetadataSnippetSetRunner.java
```diff
@@ -0,0 +1,36 @@
...
+package com.google.api.codegen.metadatagen;
+
+import com.google.api.codegen.GeneratedResult;
+import com.google.api.codegen.SnippetSetRunner;
+import com.google.api.tools.framework.model.Model;
+import com.google.api.tools.framework.snippet.SnippetSet;
+import com.google.common.collect.ImmutableMap;
+
+/** A SnippetSetRunner for package metadata template rendering. */
+public class PackageMetadataSnippetSetRunner {
```
garrettjonesgoogle: You can delete this class too

src/main/resources/com/google/api/codegen/metadatagen/py/setup.py.snip
```diff
@@ -0,0 +1,46 @@
...
+        'Programming Language :: Python :: 3.4',
+        'Programming Language :: Python :: 3.5',
+        'Programming Language :: Python :: Implementation :: CPython',
+      ],
+      description='GRPC library for the {@context.getApiNameInfo.fullName} service',
+      long_description=open('README.rst').read(),
+      install_requires=install_requires,
+      license='{@context.getDefaultsValue("license")}',
+      packages=find_packages(),
+      namespace_packages={@context.getCopierResults},
```
garrettjonesgoogle: Both the copier and the template know that "copier results" contains namespace packages - why not name it that way? Otherwise, "copier results" becomes an unknown blob that's harder to trace.

src/test/java/com/google/api/codegen/metadatagen/PackageMetadataGeneratorTest.java
```diff
@@ -0,0 +1,109 @@
...
+    options.set(PackageMetadataGenerator.LONG_API_NAME, "Google Library Example");
+    options.set(PackageMetadataGenerator.API_VERSION, "v1");
+    options.set(PackageMetadataGenerator.API_PATH, "google/example/library");
+    Map<String, Doc> generatedDocs =
+        new PackageMetadataGenerator(
+                options,
+                PackageMetadataGeneratorTool.getSnippets(language),
+                PackageMetadataGeneratorTool.getCopier(language))
+            .generateDocs(model);
+    OutputCollector collector = new OutputCollector(Paths.get(outFile));
```
garrettjonesgoogle: This class is missing from the PR

src/main/java/com/google/api/codegen/metadatagen/PackageMetadataContext.java
```diff
@@ -0,0 +1,114 @@
...
+public class PackageMetadataContext implements ViewModel {
+  private final Map<String, Object> dependenciesMap;
+
+  private final Map<String, Object> defaultsMap;
+
+  private String copierResults;
+
+  private ApiNameInfo apiNameInfo;
+
+  private String templateFileName;
```
garrettjonesgoogle: Can't all of these member variables be `final`?

src/main/resources/com/google/api/codegen/metadatagen/py/setup.py.snip
geigerj: For Ruby and NodeJS, the copierResults will not be namespace packages -- this just happens to be the data returned for Python. I'm not even sure having a `String` copier result will be sufficient for those languages -- we might need a more structured object. Let's revisit this when porting those languages.

src/test/java/com/google/api/codegen/metadatagen/PackageMetadataGeneratorTest.java
geigerj: It was previously declared on line 49 of this file. I've moved it to the top where it may be more visible.

src/main/java/com/google/api/codegen/metadatagen/PackageMetadataSnippetSetRunner.java
geigerj: Done.

src/main/java/com/google/api/codegen/metadatagen/PackageMetadataSnippetSet.java
geigerj: Done.

src/main/java/com/google/api/codegen/metadatagen/PackageMetadataContext.java
geigerj: Done.

src/test/java/com/google/api/codegen/metadatagen/PackageMetadataGeneratorTest.java
garrettjonesgoogle: Oh whoops, I'll take a look :-)

src/main/resources/com/google/api/codegen/metadatagen/py/setup.py.snip
garrettjonesgoogle: I'm suggesting that we use a "union" strategy, where we add individual `@Nullable` attributes into PackageCopierResult that are used only by some languages. We can put a more normalized structure around it later. Maybe it would actually make sense to make the namespacePackages attribute hold a List<String> instead of a String, and do a `@join` in the template. 

With this strategy we don't really have to think about what future languages will use - we just know we'll add the attributes that they need.

src/main/java/com/google/api/codegen/metadatagen/PackageMetadataGenerator.java
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  @SuppressWarnings("unchecked")
+  protected Map<String, Doc> generateDocs(Model model) throws IOException {
+    // Used to create a "special" Doc that can then be fed into the main generation to propagate
+    // the results of the copy phase.
+    String copierMetadataKey = "_PACKAGE_COPIER_RESULTS_METADATA";
+
+    // Copy gRPC package and add non-top-level files
+    PackageCopierResult copierResults = packageCopier.run(options, copierMetadataKey);
```
garrettjonesgoogle: It looks like `copierMetadataKey` can be removed everywhere.

geigerj: Done.

src/main/java/com/google/api/codegen/metadatagen/PackageMetadataGeneratorTool.java
```diff
@@ -0,0 +1,173 @@
...
+    options.addOption(
+        Option.builder("d")
+            .longOpt("dependencies_config")
+            .desc("The dependencies configuration file.")
+            .hasArg()
+            .argName("DEPENDENCIES-FILE")
+            .required(true)
+            .build());
+    options.addOption(
+        Option.builder("d")
```
jmuk: short option "d" is already declared at line 84. conflicting?

src/main/java/com/google/api/codegen/metadatagen/PackageMetadataGeneratorTool.java
```diff
@@ -0,0 +1,173 @@
...
+            .longOpt("defaults_config")
+            .desc("The defaults configuration file.")
+            .hasArg()
+            .argName("DEFAULTS-FILE")
+            .required(true)
+            .build());
+    options.addOption(
+        Option.builder()
+            .longOpt("python_pkg_config")
```
jmuk: Why it's Python specific?  Is there anything special on Python?

I'm a bit concerned about this because it's unclear what else parameters will be added when supporting more languages. Adding more language specific parameters would lead to a combination problem.

src/main/java/com/google/api/codegen/metadatagen/PackageMetadataGeneratorTool.java
```diff
@@ -0,0 +1,173 @@
...
+      new ImmutableMap.Builder<String, PackageCopier>()
+          .put("python", new PythonPackageCopier())
+          .build();
+
+  public static void main(String[] args) throws Exception {
+    Options options = new Options();
+    options.addOption("h", "help", false, "show usage");
+    options.addOption(
+        Option.builder()
+            .longOpt("descriptor_set")
```
jmuk: why descriptor_set exists? Who uses this?

src/main/java/com/google/api/codegen/metadatagen/PackageMetadataGeneratorTool.java
geigerj: This is ported from Packman: https://github.com/googleapis/packman/blob/master/config/python_pkg.yml

I think we should consider deleting this file, but I do not want to introduce behavior changes from packman in this initial PR. We can address this as a follow-up.

src/main/java/com/google/api/codegen/metadatagen/PackageMetadataGeneratorTool.java
geigerj: We use this to build the model. This is described in the internal design doc, but the short answer is that eventually we will use the Model built from the protos to ensure greater consistency with the rest of toolkit (e.g., to fix #270). It is not yet used because I did not want to alter behavior on this PR.

geigerj:
@jmuk What parts are you concerned about not being extensible? As discussed previously in the code review, `PackageCopierResult.Metadata` is currently intended to be a language-specific union type specifically designed to be extensible to the needs of other languages (e.g., NodeJS may use a different constructor to compute the index.js file, similar for Ruby, ...).

geigerj:
@jmuk To clarify, to add a new gRPC language, you should create the appropriate snippet templates (which involves simple translation from the packman mustache templates), update the language maps in `PackageMetadataGeneratorTool.java`, and create a custom implementation of `PackageCopier.java` that performs any necessary computation on the gRPC input.

I am leaning towards not using this tool for GAPIC packaging; instead we can just generate metadata directly in GAPIC if needed, similar to how we do docs. However, if we want to re-use this, the process would be similar to adding a new gRPC language.

src/main/java/com/google/api/codegen/metadatagen/PackageMetadataGeneratorTool.java
jmuk: Hm, I see. But, for now it's not used yet, right?  It's passed to the options object as `PYTHON_PACKAGE_FILE` in line 149, but I can't find any other references to use it.

src/main/java/com/google/api/codegen/metadatagen/PackageMetadataGeneratorTool.java
jmuk: In that case then let's remove this for now. You can add when it's necessary.

jmuk:
Yes, my concern is the python referecence in `CopierResults.Metadata`, and that feels like a bad design.

I've seen some patterns where the base class is empty but individual subclass has language-specific information. For example, `ApiMethodView` is empty, whereas subclasses like `OptionalArrayApiMethodView` has concrete information. I'm wondering the same pattern can be applied here.

geigerj:
@jmuk True, we could have an empty interface `CopierResultsMetadata`, and each language could extend it as it needs, but I'm not sure I understand why that is better or different from collecting them in a union type.

Doing this makes sense to me in the `ApiMethodView` case because the implementations of that class are very complex and it would be a mess to combine them into a single Java file. But each `Metadata` is probably very simple. AFAIK, the types of metadata we might want are:

- a list of namespace packages (for Python)
- a list of .js files for the index.js (NodeJS)
- a list of source files for the `v1.rb` file (if we do Ruby GAPIC)

Do we actually want each of these in a separate Java file?

jmuk:
Note that index.js files will be ported into GAPIC codegen -- @landrito will work on that. v1.rb will be same, I think.

I'm hesitant for this because I'm feeling like `Metadata` should be language agnostic, and language-specific parts should be separated into a different directory. Another idea is make `Metadata` as a `Map`, so that individual `copierResults` implementation can put any data as they like.

garrettjonesgoogle:
@jmuk , I think it's easier to start with the simple structure that we have here, and then move towards a more separated design as it grows bigger. The initial design is not the permanent design. As @geigerj noted, this isn't fully using the MVVM patterns yet - it's mostly mirroring the packman patterns for easy porting. When it's fully ported, I feel like refactoring to a more normalized design will be easier.

src/main/java/com/google/api/codegen/metadatagen/PackageMetadataGeneratorTool.java
geigerj: There's a TODO in that file to finish the implementation, which I intend to do in a follow-up PR. However, I'll remove this option from this PR to separate the work out more clearly.

src/main/java/com/google/api/codegen/metadatagen/PackageMetadataGeneratorTool.java
geigerj: This is still needed to build on top of API Compiler, even if we don't yet use the `Model` that the API Compiler framework provides yet. It is easier to include the scaffolding in this initial PR than to have to refactor the tool in a follow-up, so I would like to leave this in.

PR #748, opened 2016-11-16
Fix Python bytes default value

### LINKED ISSUE #747: Fix proto default value for "bytes" fields
### What

Python proto fields of type are currently initialized with `''`. They should be initialized with `b''`.

### Why

The old way works in Python 2 but not 3.

geigerj:
Fixes #747

### DIFF
--- a/src/test/java/com/google/api/codegen/testdata/php_main_library.baseline
+++ b/src/test/java/com/google/api/codegen/testdata/php_main_library.baseline
@@ -742,6 +742,8 @@ class LibraryServiceApi
      *          The edition of the series
      *     @type bool $reviewCopy
      *          If the book is in a pre-publish state
+     *     @type string $seriesUuid
+     *          Uniquely identifies the series to the publishing house.
      *     @type Google\GAX\RetrySettings $retrySettings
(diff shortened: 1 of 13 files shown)


src/test/java/com/google/api/codegen/testdata/php_main_library.baseline
```diff
@@ -742,6 +742,8 @@ class LibraryServiceApi
      *          The edition of the series
      *     @type bool $reviewCopy
      *          If the book is in a pre-publish state
+     *     @type string $seriesUuid
+     *          Uniquely identifies the series to the publishing house.
```
bjwatson: @michaelbausor Why is a required field treated as optional in PHP?

geigerj: @bjwatson This field is optional in all languages -- the config generator marks it as required, but the config that is passed into the tests has it as optional.

bjwatson: Oh, got it. I was misreading the tests. Thanks, Jacob.

PR #749, opened 2016-11-16
Limit max heap size in the Travis environment

saicheems:


saicheems:
I've lit the appropriate ceremonial candles for this build to pass, ptal

saicheems:
In serious:

I've switched the flags to specify maximum heap size... If nothing else it at least forces the Java 7 JVMs to garbage collect before travis kills the Gradle process. I have no clue if this will continue to work, but it's something.

bjwatson:
Java has a default maximum heap size, so there already is a GC forcing function. See [Oracle docs](https://docs.oracle.com/cd/E13150_01/jrockit_jvm/jrockit/jrdocs/refman/optionX.html#wp999528) for more details.

This change won't do anything, unless one or more tests are having trouble fitting in memory (or some combinations of tests if they are allowed to run concurrently).

saicheems:
Thanks for the doc link! It looks like the maximum defaults to a value based on the memory available on the machine.

The information I do have indicates that this is a memory issue, the instance is killing the Gradle process before it can complete (exit code 137). It could be the case that the JVM is defaulting to a maximum heap size that is greater than the memory available on the instance because the heuristic sets it too high. Plus Travis instances run several other potentially memory intensive services.

The Travis environment has 3gb of memory, so the minimum setting for the default max heap size is 1.5gb. Limiting it could help, if a lot of memory is allocated to other services then the machine might kill the Gradle process before it hits its default max heap size setting.

I've run the build on this PR twice, and both have succeeded. Given the number of failures on master , it may be statistically significant :)

garrettjonesgoogle:
LGTM, let's give it a whirl with the team and circle back if things keep failing.

bjwatson:
Two is coincidence, three is a trend. ;)

Nevertheless, my fingers are crossed that this is all we need to stabilize the builds.

PR #753, opened 2016-11-17
Fix API returning empty proto message for ruby

### LINKED ISSUE #752: GAPIC API should return void value when the return message is empty
For Ruby, when the API returns google.protobuf.empty, instead of returning Google::Protobuf::Empty, we should just return 'nil'.

swcloud:
Fix #752

### DIFF
--- a/src/main/java/com/google/api/codegen/config/MethodConfig.java
+++ b/src/main/java/com/google/api/codegen/config/MethodConfig.java
@@ -26,6 +26,7 @@ import com.google.api.codegen.config.GrpcStreamingConfig.GrpcStreamingType;
 import com.google.api.tools.framework.model.Diag;
 import com.google.api.tools.framework.model.DiagCollector;
 import com.google.api.tools.framework.model.Field;
+import com.google.api.tools.framework.model.MessageType;
 import com.google.api.tools.framework.model.Method;
 import com.google.api.tools.framework.model.SimpleLocation;
 import com.google.auto.value.AutoValue;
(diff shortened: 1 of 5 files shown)


src/test/java/com/google/api/codegen/testdata/ruby_main_library.baseline
```diff
@@ -1133,7 +1137,6 @@ module Library
       # @param options [Google::Gax::CallOptions]
       #   Overrides the default settings for this call, e.g, timeout,
       #   retries, etc.
-      # @return [Google::Tagger::V1::AddTagResponse]
```
swcloud: I think we should not just handle 'google.protobuf.empty', in the case here 'AddTagResponse' is also defined as an empty message here, so we should return nil too. When later, if AddTagResponse becomes non-empty, since we need to re-generate the gapic, then the AddTagResponse will be added back as return type.

src/main/java/com/google/api/codegen/config/MethodConfig.java
```diff
@@ -333,6 +334,12 @@ public static boolean isGrpcStreamingMethod(Method method) {
     return method.getRequestStreaming() || method.getResponseStreaming();
   }
 
+  /** Returns true if the method returns empty message */
+  public static boolean isReturnEmptyMessageMethod(Method method) {
+    MessageType returnMessageType = method.getOutputMessage();
+    return returnMessageType.getFields().size() == 0;
```
jmuk: You don't have to check like this -- we should only special-handle the `google.protobuf.Empty` type.  For example, Java returns `AddLabelResponse`. See https://github.com/googleapis/toolkit/blob/master/src/test/java/com/google/api/codegen/testdata/java_main_library.baseline#L2118

PR #755, opened 2016-11-17
Java gen: supporting long running operations

garrettjonesgoogle:
Also:

* New GAPIC config option for configuring the return type of an operation
* Some renames of fields/types starting with 'api' to either 'client'
or 'service' so that their intent is clearer
* Java/PHP: Moving most conditional importing into the top-level
transformer classes - using SurfaceNamer wasn't sustainable


src/main/java/com/google/api/codegen/config/LongRunningConfig.java
```diff
@@ -0,0 +1,52 @@
...
+  /** Returns the message type returned from a completed operation. */
+  public abstract TypeRef getReturnType();
+
+  /** Creates an instance of LongRunningConfig based on LongRunningConfigProto. */
+  @Nullable
+  public static LongRunningConfig createLongRunningConfig(
+      Model model, DiagCollector diagCollector, LongRunningConfigProto longRunningConfigProto) {
+
+    TypeRef returnType = model.getSymbolTable().lookupType(longRunningConfigProto.getReturnType());
+    // TODO switch to diag message
```
michaelbausor: Were you planning to do this before merging?

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -556,17 +527,17 @@ public String getApiMethodName(Method method, VisibilityConfig visibility) {
...
   }
 
   /** The name of the example for the method. */
   public String getApiMethodExampleName(Interface interfaze, Method method) {
     return getApiMethodName(method, VisibilityConfig.PUBLIC);
   }
 
-  public String getAsyncApiMethodExampleName(Method method) {
-    return getNotImplementedString("SurfaceNamer.getAsyncApiMethodExampleName");
+  public String getAsyncApiMethodExampleName(Interface interfaze, Method method) {
```
michaelbausor: Why does this need the Interface? And in getOperationCallableMethodExampleName

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -690,9 +666,9 @@ public String getStreamingServerName(Method method) {
     return getNotImplementedString("SurfaceNamer.getStreamingServerName");
   }
 
-  /** The GRPC streaming client type name for a given method. */
-  public String getStreamingClientName(Method method) {
-    return getNotImplementedString("SurfaceNamer.getStreamingClientName");
+  public String getGrpcStreamingApiReturnTypeName(Method method) {
```
michaelbausor: Nit: add comment

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -680,6 +651,11 @@ public String getStaticLangAsyncReturnTypeName(Method method, MethodConfig metho
     return getNotImplementedString("SurfaceNamer.getStaticLangAsyncReturnTypeName");
   }
 
+  public String getAndSaveOperationResponseTypeName(
```
michaelbausor: Nit: add comment

src/main/resources/com/google/api/codegen/java/main.snip
```diff
@@ -274,7 +294,18 @@
     @case "RequestObjectMethod"
       {@requestObjectMethod(apiMethod)}
     @case "CallableMethod"
-      {@callableMethod(apiMethod)}
+    # TODO create a GrpcStreaming ApiMethodType
```
michaelbausor: Nit: create an issue?

src/test/java/com/google/api/codegen/testdata/java_main_library.baseline
```diff
@@ -370,6 +392,7 @@ public class LibraryServiceApi implements AutoCloseable {
    * @throws com.google.api.gax.grpc.ApiException if the remote call fails
    */
   public final Shelf createShelf(Shelf shelf) {
+
```
michaelbausor: Nit: why extra lines? But no point fighting the snippet language if there isn't a good reason...

src/main/java/com/google/api/codegen/config/LongRunningConfig.java
garrettjonesgoogle: Ahh yes I intended on doing that - I will update the code.

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
garrettjonesgoogle: The thought was that since `getApiMethodExampleName` in Go needs Interface, this should be consistent in case it needs to be overridden.

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
garrettjonesgoogle: I'll add missing comments for this file.

src/main/resources/com/google/api/codegen/java/main.snip
garrettjonesgoogle: Nit: Is that really a nit?

I view this as a more "opportunistic" TODO - if someone is modifying this code, then take care of it; otherwise, I don't know that it's important enough for an issue to be filed.

src/test/java/com/google/api/codegen/testdata/java_main_library.baseline
garrettjonesgoogle: It's a line between the validation calls and the request construction code - if there are no validations, then this is what you see. I was fighting another issue where multiple lines were getting squashed onto one line - this is a more palatable outcome.

garrettjonesgoogle:
Addressed feedback, PTAL

src/main/resources/com/google/api/codegen/java/main.snip
```diff
@@ -380,27 +405,64 @@
...
+  public final UnaryCallable<{@apiMethod.serviceRequestTypeName}, \
+      {@apiMethod.responseTypeName}> {@apiMethod.name}() {
+    return {@apiMethod.callableMethod.callableName};
+  }
 @end
 
-@private nonStreamingCallableMethod(apiMethod)
-  public final UnaryCallable<{@apiMethod.apiRequestTypeName}, \
+@private streamingCallableMethod(apiMethod)
+  @let coreSampleCode = streamingCallableMethodSampleCode(apiMethod)
```
michaelbausor: We have stopped using let statements in most places - is it necessary here, for formatting or parsing reasons? If not, would a line break also work?

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -696,11 +705,12 @@ public String getCallableMethodExampleName(Interface interfaze, Method method) {
     return getCallableMethodName(method);
   }
 
-  /** The name of the plain callable variant of the given method. */
+  /** The name of the operation callable variant of the given method. */
   public String getOperationCallableMethodName(Method method) {
     return publicMethodName(Name.upperCamel(method.getSimpleName(), "OperationCallable"));
   }
 
+  /** The ame of the example for the operation callable variant of the given method. */
```
garrettjonesgoogle: oops will fix

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
```diff
@@ -637,6 +659,9 @@ private void addUnitTestImports(SurfaceTransformerContext context) {
     typeTable.saveNicknameFor("com.google.protobuf.GeneratedMessageV3");
     typeTable.saveNicknameFor("io.grpc.Status");
     typeTable.saveNicknameFor("io.grpc.StatusRuntimeException");
+    if (context.getInterfaceConfig().hasLongRunningOperations()) {
```
garrettjonesgoogle: Oh that's really easy to work in, I'll include it.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTransformer.java
```diff
@@ -474,6 +493,24 @@ public SettingsDocView generateSettingsDoc(
       } else if (methodConfig.isGrpcStreaming()) {
         context.getTypeTable().saveNicknameFor("com.google.api.gax.grpc.StreamingCallable");
         apiMethods.add(apiMethodTransformer.generateCallableMethod(requestMethodContext));
+      } else if (methodConfig.isLongRunningOperation()) {
+        context.getTypeTable().saveNicknameFor("com.google.api.gax.grpc.OperationCallable");
+        if (methodConfig.isFlattening()) {
+          if (methodConfig.isFlattening()) {
```
garrettjonesgoogle: WAT

src/main/resources/com/google/api/codegen/java/main.snip
garrettjonesgoogle: I've used a `@let` for this particular scenario from the beginning. In this scenario, we need to fully render a sub-snippet and then add stars to every line of it; using a `@let` is the only way to do it.

garrettjonesgoogle:
Comments addressed, PTAL

michaelbausor:
LGTM

PR #757, opened 2016-11-17
Fix the pattern of enums.

### LINKED ISSUE #705: Enum object is not emitted in Node.js code
See GoogleCloudPlatform/google-cloud-node#1761 for reference.

@jmuk PTAL

jmuk:
The enum values should be tracked as packageName.Path.To.Enum.VALUE
in NodeJS.

In order to refer to packageName in the architecture, I found
that the current package_name structure for nodejs gapic config
does not fit well into the current MVVM architecture.

So I changed the expected format for the 'package_name' field
in the gapic config as 'apiname.version'. This makes the name
being not a package name of NPM, but probably okay for now,
and it also indicates how the module will be accessed in the
code.

For simplicity, I put the scope part `@google-cloud/` into
`domain_layer_location`.

Fixes #705.


jmuk:
@landrito, please take a look.

src/main/java/com/google/api/codegen/nodejs/NodeJSGapicContext.java
```diff
@@ -157,23 +157,47 @@ public String getNamespace(Interface service) {
...
    */
-  public String getModuleName(Interface service) {
-    List<String> names = Splitter.on(".").splitToList(service.getFullName());
-    return names.get(names.size() - 3) + lowerUnderscoreToUpperCamel(names.get(names.size() - 2));
+  public String getModuleName() {
+    List<String> names = Splitter.on(".").splitToList(getApiConfig().getPackageName());
+    if (names.size() < 2) {
+      return getApiConfig().getPackageName();
+    }
+    return names.get(0) + lowerUnderscoreToUpperCamel(names.get(names.size() - 1));
```
jmuk: Right, it's going to be `apinameV1beta2`, small `b`. However, I don't think this is problematic, both are equally understandable and acceptable. So I'd like to take simpler implementation.

src/main/java/com/google/api/codegen/nodejs/NodeJSUtils.java
```diff
@@ -20,10 +20,10 @@
...
   /**
    * Returns true if the current API is a part of gcloud (i.e. cloud API). This can be known if the
-   * package name configuration is in the pattern of "@google-cloud/(API_NAME)".
+   * domain_layer_location is "google-cloud".
    */
   public static boolean isGcloud(ApiConfig config) {
-    String packageName = config.getPackageName();
-    return !Strings.isNullOrEmpty(packageName) && packageName.startsWith("@google-cloud/");
+    String domainName = config.getDomainLayerLocation();
+    return !Strings.isNullOrEmpty(domainName) && domainName.equals("google-cloud");
```
jmuk: Done

PR #762, opened 2016-11-18
Change suffix from Api to Client for Ruby

### LINKED ISSUE #670: Use "Client" as GAPIC generated class name
Currently we use `Api` suffix for class name. So gRPC service name is `Library`, the generated class is `LibraryApi`.

Internally we got feedback to rename this. Probably `Client` is fine, so `LibraryClient`?

- [x] C#
- [x] Go
- [x] Java
- [x] NodeJS
- [x] PHP
- [x] Python
- [x] Ruby

swcloud:
This fixes #670 for Ruby.

I could make changes in super classes GapicContext.java and SurfaceNamer.java such that other languages are fixed too, but I found that triggers massive changes, which seems to be quite risky.

It's safer to wait until other language fixes this issue too, then we can do refactoring to pull up the changes to the super classes.

### DIFF
--- a/src/main/java/com/google/api/codegen/ruby/RubyGapicContext.java
+++ b/src/main/java/com/google/api/codegen/ruby/RubyGapicContext.java
@@ -65,6 +65,12 @@ public class RubyGapicContext extends GapicContext implements RubyContext {
     return true;
   }
 
+  @Override
+  /** Return the name of the class which is the GAPIC wrapper for this service interface. */
+  public String getApiWrapperName(Interface service) {
+    return service.getSimpleName() + "Client";
+  }
+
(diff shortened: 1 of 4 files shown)


jmuk:
Why don't we simply update the common `GapicContext` class and `SurfaceNamer` class?
Anyways we want to update the suffix for all languages, right?

jmuk:
cc: @garrettjonesgoogle

garrettjonesgoogle:
@jmuk , I think I agree with @swcloud 's approach of one language at a time - we want to control when the renaming is pushed out with respect to other changes in each language. This is a pretty disruptive renaming, so some care is necessary.

PR #763, opened 2016-11-21
improve SymbolTable

pongad:
In the previous implementation,
adding the same name N times to the table requires
linear space and quadratic time.

The new implementation requires constant space
and linear time.


garrettjonesgoogle:
Why did you close this with no comment?

pongad:
I forgot to comment. The algorithm is incorrect. Suppose we add three strings "x", "x", and "x2"

The first will return "x", the second "x2", but the third will also return "x2" since it gets counted separately. The only way I can see to make it work right is to have a custom made prefix tree. I think that's a lot more problem than it's worth as the number of collisions we have is probably small.

PR #764, opened 2016-11-21
go: make concrete type LRO

pongad:
This commit creates a specialized type for LRO, so that
they can return concrete-typed values.

A couple surface questions remain:
- Do we want to move longrunning into a support package?
- Do we want to change to name of Operation.Wait? Maybe WaitProto?
  If we do, we should change the name of Poll to match.


jba:
This seems like a good change. Can you generate a specialized Poll as well? Also, there should be codegen options for generating a specialized metadata function, when we know the type.

Re your questions: What does it mean to move longrunning into a support package? And what would we gain from renaming the methods?

pongad:
@jba Good point about specializing Poll and Metadata. I originally thought to leave them out to reduce the weight of the surface; same reason we reduced the number of methods on iterators. If you think they are worth adding, I don't have any objection.

My thoughts behind renaming the methods is this. Embedding Operation sounds like a good idea so we don't have to redefine a bunch of methods. The docs are probably also cleaner this way. However, imagine we publish an API. Maybe by mistake, we forget to configure concrete type for Wait or Metadata. `bookop.Wait` now refers to the Wait method of Operation. If we later configure the concrete type, `bookop.Wait` becomes the version defined on the BookOperation itself. The method changes signature, breaking users. Changing Operation.Wait to Operation.WaitProto solves this problem even though the name sounds a little funny.

@rakyll brought up the point of maybe moving longrunning package into internal. It feels a little weird to me to embed an internal type. I'm not sure if that feeling is warranted or not though.

If you think this discussion might drag on, we could consider bring the design discussion back to a mailinglist first.

jba:
I think forgetting to make a custom operation would just be a bug that we'd fix. Actually, I don't see any reason to not make a custom type for every resource that has a long-running operation associated with it -- maybe that should be the default, and you have to turn it off manually.

The longrunning package will probably still have a place, so I don't want to make its method names too ugly. Anyway, XXXProto won't work because I think we decided to change the arg type to interface{}.

pongad:
I think we still need a config to turn the concrete type on, not off. We can't tell from the proto definition what the concrete type of the LRO is. Currently, the concrete type is in the config file.

jba:
OK, makes sense.

jba:
@pongad, I don't think we should make longrunning internal, because we may need to expose it for some APIs. I'm okay with moving it into a support subdir, if @rakyll and you feel strongly about that. 

I'll do the CL described in https://github.com/GoogleCloudPlatform/google-cloud-go/issues/424, and maybe that will clarify things a bit.

jba:
I'm moving longrunning to cloud.google.com/go/support/longrunning, but otherwise not changing it.

The problem with embedding it is, what if we know that the operation has no metadata? We could not avoid surfacing the Metadata method. So maybe it's better to just store the operation inside a struct and re-implement the necessary methods.

Also, since we provide the operation name, there should be a way to re-create the operation from the name, to support using the operation in a different process from the one that created it.

pongad:
@garrettjonesgoogle @jba PTAL. It feels a little funny to generate so many methods. It's probably not a big deal. If we want to generate fewer methods and don't want to embed, one way out is to expose the `lro` member. I'm not sure if I'm overthinking this.

src/main/java/com/google/api/codegen/viewmodel/LongRunningOperationDetailView.java
```diff
@@ -21,6 +21,8 @@
 
   public abstract String operationReturnType();
 
+  public abstract String operationResponseType();
```
pongad: When I was working on this, my thoughts were that
- "return type" -> The type returned by the client's method (`OperationFuture<Foo>` in Java and `FooOperation` in Go). I was tempted to rename this `handleType` but that sound just as vague.
- "response type" -> The concrete type of the message in [Operation.result.response field](https://github.com/googleapis/googleapis/blob/master/google/longrunning/operations.proto#L112).

src/main/resources/com/google/api/codegen/go/main.snip
```diff
@@ -324,17 +324,89 @@
...
         }
+        if !op.Done() {
+            return nil, nil
+        }
         return &resp, nil
     }
+
+    // Metadata returns metadata associated with the long-running operation.
+    // Latest metadata can be fetched by Poll.
+    // If the metadata is not availabe, the returned metadata and error are both nil.
```
jba: available

src/test/java/com/google/api/codegen/testdata/go_main_library.baseline
```diff
@@ -790,14 +790,86 @@ func (it *StringIterator) takeBuf() interface{} {
...
+        return nil, err
+    }
+    if !op.Done() {
+        return nil, nil
+    }
+    return &resp, nil
+}
+
+// Metadata returns metadata associated with the long-running operation.
+// Latest metadata can be fetched by Poll.
```
jba: I don't think this is correct. Poll returns a Book, not metadata.

src/main/resources/com/google/api/codegen/go/main.snip
pongad: Done

src/test/java/com/google/api/codegen/testdata/go_main_library.baseline
pongad: It doesn't. I was trying to document something like
> Metadata itself doesn't contact the server, but Poll does.
> So, if you need the latest metadata, call Poll first to fetch it into the object,
> then call Metadata to get it out.

I've moved this section of the doc into Poll. Please let me know if it reads better to you.

pongad:
@jba PTAL

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -664,6 +664,14 @@ public String getAndSaveOperationResponseTypeName(
     return getNotImplementedString("SurfaceNamer.getAndSaveOperationResponseTypeName");
   }
 
+  /**
+   * In languages with pointers, strip the pointer, leaving only the base type. Eg, in C, "int*"
+   * would become "int".
+   */
+  public String plainType(String type) {
```
garrettjonesgoogle: maybe `valueType`?

src/main/java/com/google/api/codegen/viewmodel/LongRunningOperationDetailView.java
```diff
@@ -19,15 +19,27 @@
 @AutoValue
 public abstract class LongRunningOperationDetailView {
 
-  public abstract String operationReturnType();
+  public abstract String constructorName();
+
+  public abstract String returnType();
```
garrettjonesgoogle: I would suggest `clientReturnTypeName`

src/main/java/com/google/api/codegen/viewmodel/LongRunningOperationDetailView.java
```diff
@@ -19,15 +19,27 @@
...
+
+  public abstract String responseType();
```
garrettjonesgoogle: I would suggest `operationPayloadTypeName`

src/main/java/com/google/api/codegen/viewmodel/LongRunningOperationDetailView.java
```diff
@@ -19,15 +19,27 @@
...
+
+  public abstract String metadataType();
```
garrettjonesgoogle: I would suggest `metadataTypeName`

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
pongad: done

src/main/java/com/google/api/codegen/viewmodel/LongRunningOperationDetailView.java
pongad: Done

src/main/java/com/google/api/codegen/viewmodel/LongRunningOperationDetailView.java
pongad: Done

src/test/java/com/google/api/codegen/testdata/go_main_library.baseline
```diff
@@ -787,3 +789,87 @@ func (it *StringIterator) takeBuf() interface{} {
...
 }
 
+
+// BookOperation manages a long-running operation yielding librarypb.Book.
+type BookOperation struct {
+    lro *longrunning.Operation
+}
+
+// BookOperation returns a new BookOperation from a given name.
+// The name must refer to a long-running operation created by the same service.
```
jba: How about: "The name must be that of a previously created BookOperation, possibly from a different process."

src/test/java/com/google/api/codegen/testdata/go_main_library.baseline
```diff
@@ -787,3 +789,87 @@ func (it *StringIterator) takeBuf() interface{} {
...
+
+// Name returns the name of the long-running operation.
+// The name is assigned by the server and is unique within the service from which the operation is created.
+func (op *BookOperation) Name() string {
+    return op.lro.Name()
+}
+
+// Cancel starts asynchronous cancellation on a long-running operation.
+// The server makes a best effort to cancel the operation, but success is not guaranteed.
+// If the server doesn't support this method, it returns grpc.Code(error) == codes.Unimplemented.
```
jba: Maybe we can figure this out and not generate the method?

src/test/java/com/google/api/codegen/testdata/go_main_library.baseline
jba: We definitely need some doc on Metadata. I would put it this way:

> Metadata itself doesn't contact the server, but Poll does. Call this method after a successful call to Poll.

src/main/java/com/google/api/codegen/config/LongRunningConfig.java
```diff
@@ -41,7 +41,7 @@ public static LongRunningConfig createLongRunningConfig(
     boolean error = false;
 
     TypeRef returnType = model.getSymbolTable().lookupType(longRunningConfigProto.getReturnType());
-    TypeRef metadataType =
+    TypeRef metadataTypeName =
```
garrettjonesgoogle: This isn't a type name, it's a type - rename back to `metadataType`

pongad: Sed was overzealous. Done.

src/test/java/com/google/api/codegen/testdata/go_main_library.baseline
pongad: I like it. Done.

pongad:
@garrettjonesgoogle @jba PTAL.

Per PR request, two fields are added to LRO config, controlling whether the service implements cancel and delete. As far as I can tell, speech implements neither.

PR #769, opened 2016-11-22
Add PHP rendering util, update baselines

michaelbausor:
Update formatting of PHP docs to prevent comment blocks from being escaped by */ chars in proto comments


garrettjonesgoogle:
LGTM

PR #771, opened 2016-11-22
convert ClientTestCaseView to TestMethodView

pongad:
This commit converts all uses of ClientTestCaseView to TestMethodView.

The generated code is affected in only one place. There's an extra import in Java smoke test, since the shared logic imports request types, which are not used by the rest of the file. Because we use a formatter to throw out unused imports, I don't think this is a serious concern. Please let me know if it is.


src/main/java/com/google/api/codegen/transformer/go/GoGapicSurfaceTestTransformer.java
```diff
@@ -125,73 +120,27 @@ private MockCombinedView generateMockServiceView(
...
+        clientMethodType = ClientMethodType.PagedRequestObjectMethod;
+      }
       testCaseViews.add(
-          createTestCaseView(
-              context.asRequestMethodContext(method),
+          mockServiceTransformer.createTestMethodView(
+              methodContext,
               testNameTable,
-              Collections.<FieldConfig>emptyList()));
+              initCodeTransformer.createRequestInitCodeContext(
```
shinfan: I would prefer to make this a local var.

src/test/java/com/google/api/codegen/testdata/java_smoke_test_library.baseline
```diff
@@ -21,6 +21,7 @@ import com.google.common.collect.Lists;
 import com.google.example.library.v1.Book;
 import com.google.example.library.v1.Book.Rating;
 import com.google.example.library.v1.BookName;
+import com.google.example.library.v1.UpdateBookRequest;
```
shinfan: Any reason that this import is added but not used in the code?

src/main/java/com/google/api/codegen/transformer/go/GoGapicSurfaceTestTransformer.java
shinfan: I would prefer to make this a local var (also other similar occurrences)

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
```diff
@@ -278,26 +258,48 @@ private ClientTestFileView createUnitTestFileView(SurfaceTransformerContext cont
...
+                testNameTable,
+                initCodeTransformer.createRequestInitCodeContext(
+                    methodContext,
+                    new SymbolTable(),
+                    methodConfig.getRequiredFieldConfigs(),
+                    InitCodeOutputType.SingleObject,
+                    valueGenerator),
+                ClientMethodType.CallableMethod));
       } else if (methodConfig.isFlattening()) {
+        ClientMethodType clientMethodType;
```
shinfan: Maybe worth moving this logic into methodConfig so we can just call methodConfig.getMethodType() here.

src/main/java/com/google/api/codegen/transformer/go/GoGapicSurfaceTestTransformer.java
shinfan: Oh I saw the description. No as long as we know the reason I don't think it is a big concern.

pongad: Done

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
pongad: I don't think this works. Different languages set this to different values. Maybe a way out is to put this in a table specific to Java. This logic is only used once though, so I think the benefit is rather small.

pongad:
@shinfan PTAL

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
shinfan: Ummm, fair enough. I thought the client method type is shared across languages.

shinfan:
LGTM. Let @garrettjonesgoogle  take a look too.

src/main/java/com/google/api/codegen/viewmodel/testing/TestMethodView.java
```diff
@@ -14,40 +14,78 @@
...
+import com.google.api.codegen.config.GrpcStreamingConfig.GrpcStreamingType;
 import com.google.api.codegen.viewmodel.ClientMethodType;
 import com.google.api.codegen.viewmodel.InitCodeView;
 import com.google.auto.value.AutoValue;
+import java.util.List;
 
 @AutoValue
 public abstract class TestMethodView {
 
+  public abstract String surfaceMethodName();
```
garrettjonesgoogle: Use `clientMethodName`

src/main/java/com/google/api/codegen/transformer/go/GoGapicSurfaceTestTransformer.java
```diff
@@ -125,73 +121,26 @@ private MockCombinedView generateMockServiceView(
         .build();
   }
 
-  private List<ClientTestCaseView> createTestCaseViews(SurfaceTransformerContext context) {
-    ArrayList<ClientTestCaseView> testCaseViews = new ArrayList<>();
+  private List<TestMethodView> createTestCaseViews(SurfaceTransformerContext context) {
```
garrettjonesgoogle: We should be consistent between `TestMethod` and `TestCase`. I personally prefer `TestCase`. So, I would prefer the view to be named `TestCaseView` (or the original `ClientTestCaseView`)

src/main/java/com/google/api/codegen/transformer/MockServiceTransformer.java
```diff
@@ -103,6 +106,57 @@ public MockServiceTransformer(ValueProducer valueProducer) {
...
+  public TestMethodView createTestMethodView(
+      MethodTransformerContext methodContext,
+      SymbolTable testNameTable,
+      InitCodeContext initCodeContext,
+      ClientMethodType clientMethodType) {
+    Method method = methodContext.getMethod();
+    MethodConfig methodConfig = methodContext.getMethodConfig();
+    SurfaceNamer namer = methodContext.getNamer();
+
+    String surfaceMethodName;
```
pongad: Done

src/main/java/com/google/api/codegen/viewmodel/testing/TestMethodView.java
pongad: Done

pongad:
@garrettjonesgoogle PTAL

garrettjonesgoogle:
LGTM

pongad:
YAY! Thank you @garrettjonesgoogle

PR #772, opened 2016-11-22
Added metadata to longrunning config.

### LINKED ISSUE #766: [LRO] Add metadata_type to longrunning field in gapic yaml.
### What
[google.longrunning.Operation](https://github.com/googleapis/googleapis/blob/master/google/longrunning/operations.proto#L80) objects hold a metadata google.protobuf.Any message. A type is needed in order for LRO wrappers to be able to unpack the metadata from the message. There is currently specification for the [response type](https://github.com/googleapis/toolkit/blob/master/src/test/java/com/google/api/codegen/testdata/library_gapic.yaml#L502) to unpack the response of an Operation but not the metadata. This needs to be added.

### Expected
```yaml
  - name: GetBigBook
    ...
    long_running:
      return_type: google.example.library.v1.Book
      metadata_type: google.example.library.v1.GetBigBookMetadata
```

landrito:
Resolves #766.

### DIFF
--- a/src/main/java/com/google/api/codegen/config/LongRunningConfig.java
+++ b/src/main/java/com/google/api/codegen/config/LongRunningConfig.java
@@ -30,6 +30,9 @@ public abstract class LongRunningConfig {
   /** Returns the message type returned from a completed operation. */
   public abstract TypeRef getReturnType();
 
+  /** Returns the message type returned from a completed operation. */
+  public abstract TypeRef getMetdataType();
+
   /** Creates an instance of LongRunningConfig based on LongRunningConfigProto. */
   @Nullable
   public static LongRunningConfig createLongRunningConfig(
@@ -38,6 +41,8 @@ public abstract class LongRunningConfig {
     boolean error = false;
 
     TypeRef returnType = model.getSymbolTable().lookupType(longRunningConfigProto.getReturnType());
+    TypeRef metadataType =
+        model.getSymbolTable().lookupType(longRunningConfigProto.getMetadataType());
 
     if (returnType == null) {
       diagCollector.addDiag(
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/api/codegen/config/LongRunningConfig.java
```diff
@@ -30,6 +30,9 @@
   /** Returns the message type returned from a completed operation. */
   public abstract TypeRef getReturnType();
 
+  /** Returns the message type returned from a completed operation. */
+  public abstract TypeRef getMetdataType();
```
garrettjonesgoogle: fix spelling

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -453,4 +453,7 @@ message LongRunningConfigProto {
   // The fully-qualified type that is returned from an Operation when it is
   // complete.
   string return_type = 1;
+
+  // The fully-qualified type  of the metadata of an Operation.
```
garrettjonesgoogle: convert the double space into a single space in the middle of the doc line here

src/main/java/com/google/api/codegen/config/LongRunningConfig.java
landrito: done.

src/main/proto/com/google/api/codegen/config.proto
landrito: done.

src/main/java/com/google/api/codegen/config/LongRunningConfig.java
landrito: fixed!

PR #774, opened 2016-11-22
Fix sslCreds and credentialsLoader bugs

michaelbausor:


garrettjonesgoogle:
Do you have any references for the bugs (threads, issues filed, etc)?

michaelbausor:
Only the email thread from earlier today, I cc'd you. Should I create issues for each of these?

garrettjonesgoogle:
Ahh ok. LGTM

PR #777, opened 2016-11-23
Fix name conflict with grpc stub

michaelbausor:
Use fully qualified name for grpc stub to avoid conflicting with the generated class name


garrettjonesgoogle:
LGTM

michaelbausor:
Updated to add leading backslash and remove unused function, PTAL

michaelbausor:
Updated to remove backlashes from use statements, PTAL

garrettjonesgoogle:
LGTM

PR #778, opened 2016-11-23
Add some more options to the MetadadataGen CLI

geigerj:
I forgot to add these to the CLI, though they were already present in the ToolDriver.


garrettjonesgoogle:
LGTM. Do we intend on supporting the full set of command line options indefinitely, or will we eventually reduce them by using config?

geigerj:
My intention was to add these directly to the artman config and pass them to the tool via these CLI options. The exception is the API full name, which we want to derive from the service config, not the artman config (#270).

garrettjonesgoogle:
I wonder if it would be more maintainable to have the tool understand the same config that artman does. As it is now, for every new config line we put in artman config, we have to add a new command line option to the tool. I doubt anyone is going to handcraft the full command line for this tool.

geigerj:
Continue discussion on an issue so this doesn't get lost? https://github.com/googleapis/toolkit/issues/780

PR #782, opened 2016-11-23
Support alias in PHP use statement

michaelbausor:
Support alias in PHP use statement instead of relying on the fully qualified type to support the grpc stub.

Fixes issue https://github.com/googleapis/toolkit/issues/779


garrettjonesgoogle:
cool! LGTM

PR #790, opened 2016-11-28
Remove superfluous resource-name methods

### LINKED ISSUE #788: Remove FormatXyzName and templates from C# client code
For example, `SubscriberClient` still has:

```csharp
/// <summary>
/// Path template for a project resource. Parameters:
/// <list type="bullet">
/// <item><description>project</description></item>
/// </list>
/// </summary>
public static PathTemplate ProjectTemplate { get; } = new PathTemplate("projects/{project}");

/// <summary>
/// Creates a project resource name from its component IDs.
/// </summary>
/// <param name="projectId">The project ID.</param>
/// <returns>
/// The full project resource name.
/// </returns>
public static string FormatProjectName(string projectId) => ProjectTemplate.Expand(projectId);
```

These are unnecessary now we have resource names.

chrisdunelm:
Fixes #788

### DIFF
--- a/src/main/resources/com/google/api/codegen/csharp/gapic_client.snip
+++ b/src/main/resources/com/google/api/codegen/csharp/gapic_client.snip
@@ -184,29 +184,6 @@
 
         private static readonly ChannelPool s_channelPool = new ChannelPool(DefaultScopes);
 
-        @join format : api.formatResourceFunctions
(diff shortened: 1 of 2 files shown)


jskeet:
LGTM. Where we do still want the string value, we can just create an instance and call ToString.

PR #793, opened 2016-11-28
Initial implementation of Ruby testgen (unit test)

jcanizales:
The test passes, but implementation is hacky.

It doesn’t do streaming yet.

Please focus on the generated code first, so that we're on the same page. There are a couple of big issues I want to discuss with you.

Then I'll add Shin and Garret too to help review the implementation (I'm not very happy with it).


jcanizales:
Also I need to solve the merge conflict :)

jcanizales:
(CC: @bjwatson )

jcanizales:
Ah, duh! This is what the generated test looks like: https://github.com/GoogleCloudPlatform/google-cloud-ruby/compare/master...jcanizales:test-testgen

bjwatson:
@jcanizales I added @shinfan and @garrettjonesgoogle so that we can zero in on a good implementation as soon as possible.

Can you update the Ruby library baseline file, so that we can see what the unit tests look like?

src/main/java/com/google/api/codegen/transformer/ruby/RubyTestsTransformer.java
```diff
@@ -0,0 +1,252 @@
...
+import com.google.api.tools.framework.model.Method;
+import com.google.api.tools.framework.model.Model;
+import com.google.common.collect.ImmutableMap;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+
+/** A subclass of ModelToViewTransformer which translates model into API tests in Ruby. */
+public class RubyTestsTransformer implements ModelToViewTransformer {
```
garrettjonesgoogle: Don't pluralize words in an identifier that aren't the last word - so this should be `RubyTestTransformer` instead. But actually, be consistent with the existing test transformers - `RubyGapicClientTestTransformer` would be best.

src/main/resources/com/google/api/codegen/ruby/test.snip
```diff
@@ -0,0 +1,166 @@
...
+        # LanguageService::Stub.stub(:new, mock_client) { yield(mock_client) }
+        {@gRPCClass}::Stub.stub(:new, mock_client) { yield(mock_client) }
+      end
+
+      @join testCase : testClass.testCases
+        @let methodName = testCase.surfaceMethodName
+          # TODO(jcanizales): Undo this join-if hack.
+          @join method : util.getSupportedMethodsV2(testClass.service)
+            @if util.upperCamelToLowerUnderscore(method.getSimpleName()).equals(methodName)
+              @let methodView = util.getMethodView(testClass.service, method)
```
garrettjonesgoogle: This needs to be converted into the proper MVVM pattern.

src/main/java/com/google/api/codegen/viewmodel/testing/GapicSurfaceTestClassView.java
```diff
@@ -18,12 +18,15 @@
...
 import com.google.api.codegen.viewmodel.ViewModel;
 import com.google.api.codegen.viewmodel.testing.MockServiceImplView.Builder;
+import com.google.api.tools.framework.model.Interface;
 import com.google.auto.value.AutoValue;
 import java.util.List;
 
 @AutoValue
 public abstract class GapicSurfaceTestClassView implements ViewModel {
 
+  public abstract Interface service();
```
garrettjonesgoogle: ViewModel classes should not expose any classes from framework.model.

src/main/java/com/google/api/codegen/transformer/ruby/RubyTestsTransformer.java
```diff
@@ -0,0 +1,252 @@
...
+
+    GapicSurfaceTestClassView.Builder testClass = GapicSurfaceTestClassView.newBuilder()
+        // We want something like language_service_api_test.rb (i.e. append _test to the name of
+        // the generated VKit client).
+        .outputPath("tests/" + namer.getServiceFileName(service) + "_test.rb")
+        .templateFileName(UNIT_TEST_TEMPLATE_FILE)
+        .service(service)
+        .name(namer.getUnitTestClassName(service))
+        .apiClassName(namer.getApiWrapperClassName(service))
+        .apiSettingsClassName(namer.getApiSettingsClassName(service)) // What is this?
```
garrettjonesgoogle: This isn't used in Ruby.

src/main/java/com/google/api/codegen/transformer/ruby/RubyTestsTransformer.java
```diff
@@ -0,0 +1,252 @@
...
+/** A subclass of ModelToViewTransformer which translates model into API tests in Ruby. */
+public class RubyTestsTransformer implements ModelToViewTransformer {
+  // Template files
+  private static String UNIT_TEST_TEMPLATE_FILE = "ruby/test.snip";
+
+  private final FileHeaderTransformer fileHeaderTransformer = new FileHeaderTransformer();
+  private final InitCodeTransformer initCodeTransformer = new InitCodeTransformer();
+  // It looks like this isn't producing anything that's specific to Java?
+  // TODO(jcanizales): Use same value generator as sample gen!
+  private final TestValueGenerator valueGenerator = new TestValueGenerator(new JavaValueProducer());
```
shinfan: Rename JavaValueProducer -> JavaAndRubyValueProducer

src/main/java/com/google/api/codegen/transformer/ruby/RubyTestsTransformer.java
```diff
@@ -0,0 +1,252 @@
...
+        .mockServices(Collections.<MockServiceUsageView>emptyList())
+        .testCases(createTestCaseViews(context));
+
+    // Imports must be done as the last step to catch all imports.
+    testClass.fileHeader(fileHeaderTransformer.generateFileHeader(context));
+
+    return testClass.build();
+  }
+
+  private List<GapicSurfaceTestCaseView> createTestCaseViews(SurfaceTransformerContext context) {
```
shinfan: Use TestCaseTransformer to transform test cases

src/main/resources/com/google/api/codegen/ruby/test.snip
```diff
@@ -0,0 +1,166 @@
...
+                      # expected_request = AnnotateTextRequest.new(            
+                      expected_request = {@testCase.requestTypeName}.new(
+                        {@initRequest(methodView.initCode)}
+                      )
+
+                      # mocked_response = AnnotateTextResponse.new
+                      mocked_response = {@testCase.responseTypeName}.new
+                      # TODO(jcanizales): testCase.mockResponse.initCode
+
+                      mock_client.expect(
```
shinfan: This only tests for the happy cases. Probably add a TODO to add test for exceptions.

src/main/java/com/google/api/codegen/transformer/ruby/RubyTestsTransformer.java
garrettjonesgoogle: I think it's much preferable to come up with a general name that makes room for other languages, so that it doesn't have to be renamed for each language it is used for.

jcanizales:
The snippet now doesn't use any logic, nor does it access the model.

Pending refactorings:
* The ViewModel classes introduced are only accidentally Ruby-specific. Remove that language dependence. Then move them to some shared package, or merge them with the appropriate language-agnostic ViewModel if it already exists.
* s/test.snip/unit_test.snip/

Pending features, for the next PR:
* Test exceptions
* Adjust test for LRO methods (and tweak sample gen to make the sample testable).
* Test streaming correctly.

src/main/java/com/google/api/codegen/gapic/MainGapicProviderFactory.java
```diff
@@ -323,8 +329,16 @@
                 .setSnippetFileNames(Arrays.asList("clientconfig/json.snip"))
                 .setCodePathMapper(rubyPathMapper)
                 .build();
+        GapicProvider<?> testsProvider =
```
jcanizales: Done.

src/main/java/com/google/api/codegen/viewmodel/testing/GapicSurfaceTestClassView.java
jcanizales: Done.

src/main/resources/com/google/api/codegen/ruby/test.snip
jcanizales: Done.

src/main/java/com/google/api/codegen/transformer/ruby/RubyTestsTransformer.java
jcanizales: Why not? This handles all unit tests; "test" in singular would be misleading.

`RubyGapicClientTestTransformer` is both noisy and a mouthful. Is "Client" a differentiator from "Server"? Is "Gapic" a differentiator from another project?

src/main/java/com/google/api/codegen/gapic/MainGapicProviderFactory.java
```diff
@@ -421,6 +427,16 @@
           providers.add(messageProvider);
         }
       }
```
swcloud: space

src/main/java/com/google/api/codegen/gapic/MainGapicProviderFactory.java
```diff
@@ -421,6 +427,16 @@
...
+      if (generatorConfig.enableTestGenerator() && id.equals(RUBY)) {
+        GapicProvider<?> testsProvider =
+            ViewModelGapicProvider.newBuilder()
+                .setModel(model)
+                .setApiConfig(apiConfig)
+                .setSnippetSetRunner(new CommonSnippetSetRunner(snippetContext))
+                .setModelToViewTransformer(new RubyTestsTransformer())
```
swcloud: Consider using 'RubyGapicSurfaceTestTransformer' instead of 'RubyTestsTransformer' in order to be consistent with other languages.

src/main/java/com/google/api/codegen/transformer/ruby/RubyTestsTransformer.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/** A subclass of ModelToViewTransformer which translates model into API tests in Ruby. */
+public class RubyTestsTransformer implements ModelToViewTransformer {
+  // Template files
+  private static String UNIT_TEST_TEMPLATE_FILE = "ruby/test.snip";
+
+  @Override
+  public List<String> getTemplateFileNames() {
```
swcloud: You can use one-liner

return  ImmutableList.<String>of(UNIT_TEST_TEMPLATE_FILE);

src/main/java/com/google/api/codegen/transformer/ruby/UnitTestCaseViewModel.java
```diff
@@ -0,0 +1,53 @@
...
+import com.google.api.codegen.transformer.ModelTypeTable;
+import com.google.api.codegen.viewmodel.OptionalArrayMethodView;
+import com.google.api.tools.framework.model.Interface;
+import com.google.api.tools.framework.model.Method;
+import com.google.common.base.Preconditions;
+
+/** ViewModel of each unit test case for Ruby. */
+// TODO(jcanizales): Only RubyGapicContext here is Ruby-specific, and unnecessarily so. Refactor so
+// this can be shared with other languages.
+public class UnitTestCaseViewModel {
```
swcloud: I know we should have used 'xxxViewModel' instead of 'xxxView', but at this point, I suggest keep naming consistency. In the future, we can rename all the 'xxxView' to 'xxxViewModel' all together if that's necessary.

src/main/java/com/google/api/codegen/transformer/ruby/RubySurfaceNamer.java
```diff
@@ -98,4 +99,11 @@ public String getServiceFileImportFromService(Interface service) {
   public String getProtoFileImportFromService(Interface service) {
     return service.getFile().getSimpleName().replace(".proto", "_pb");
   }
+
+  // For some reason this is only being called for tests.
+  @Override
+  public String getSourceFilePath(String path, String className) {
+    // Class name is some Java thing like LanguageServiceTest
+    return path + File.separator + className + "_test.jorge.rb";
```
swcloud: I think the following is better:

  public String getSourceFilePath(String path, String publicClassName) {
    return path + File.separator + Name.upperCamel(publicClassName).toLowerUnderscore() + ".rb";
  }

src/main/java/com/google/api/codegen/transformer/ruby/UnitTestCaseViewModel.java
```diff
@@ -0,0 +1,53 @@
...
+package com.google.api.codegen.transformer.ruby;
+
+import com.google.api.codegen.ruby.RubyGapicContext;
+import com.google.api.codegen.transformer.ModelTypeTable;
+import com.google.api.codegen.viewmodel.OptionalArrayMethodView;
+import com.google.api.tools.framework.model.Interface;
+import com.google.api.tools.framework.model.Method;
+import com.google.common.base.Preconditions;
+
+/** ViewModel of each unit test case for Ruby. */
```
swcloud: What is the reason to add another general UnitTestcaseViewModel given that we already have 'TestCaseView'? Is it because you want to use 'OptionalArrayMethodView'?

I discussed with @shinfan that we should include 'ApiMethodView' in the 'TestCaseView', so that's case, then that should address your need here.

Please discuss with @shinfan

src/main/java/com/google/api/codegen/transformer/ruby/UnitTestsViewModel.java
```diff
@@ -0,0 +1,83 @@
...
+    return templateFile;
+  }
+
+  @Override
+  public String outputPath() {
+    // TODO(jcanizales): This is Ruby-specific. Move it inside the SurfaceNamer.
+
+    // We want something like language_service_api_test.rb (i.e. append _test to the name of
+    // the generated VKit client).
+    return "tests/" + namer.getServiceFileName(service) + "_test.rb";
```
swcloud: Please pass in a pathMapper to the RubyTestTransformer instead of hard-coding the prefix here.

For example:

String outputPath = pathMapper.getOutputPath(service, context.getApiConfig());
return namer.getSourceFilePath(outputPath, namer.getUnitTestClassName);

src/main/java/com/google/api/codegen/transformer/ruby/UnitTestsViewModel.java
```diff
@@ -0,0 +1,83 @@
...
+import com.google.api.tools.framework.model.Method;
+import com.google.common.base.Preconditions;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * ViewModel for the generated unit test suite for a specific library. It represents a generated
+ * tests file.
+ */
+public class UnitTestsViewModel implements ViewModel {
```
swcloud: Again, why this cannot be 'ClientTestFileView'?

@shinfan

src/main/resources/com/google/api/codegen/ruby/test.snip
```diff
@@ -0,0 +1,156 @@
...
+  @# See the License for the specific language governing permissions and
+  @# limitations under the License.
+
+  require "minitest/autorun"
+  # require "google/cloud/language/v1beta1/language_service_api"
+  require "{@testViewModel.libraryUnderTest.outputFile}"
+  # include Google::Cloud::Language::V1beta1
+  include {@testViewModel.libraryUnderTest.namespace}
+
+  # describe LanguageServiceApi do
```
swcloud: Remove 'LanguageServiceApi'

src/main/resources/com/google/api/codegen/ruby/test.snip
```diff
@@ -0,0 +1,156 @@
...
+          
+        @end
+      @end
+    @end    
+  end
+
+@end
+
+# The following were adapted from method_sample.snip.
+# TODO(jcanizales): Is there a way to reduce redundancy with them?
```
swcloud: These will be moved to a separate initcode.snip in my PR for smoke test

src/main/resources/com/google/api/codegen/ruby/test.snip
```diff
@@ -0,0 +1,156 @@
...
+        # the host string and that the 2nd argument isn't
+        # :this_channel_is_insecure.
+        # LanguageService::Stub.stub(:new, mock_client) { yield(mock_client) }
+        {@gRPCClass}::Stub.stub(:new, mock_client) { yield(mock_client) }
+      end
+
+      @join testCase : testViewModel.testCases
+        @let methodView = testCase.methodView, \
+             methodName = methodView.name
+          # describe "#annotate_text" do
```
swcloud: Remove this line

src/main/resources/com/google/api/codegen/ruby/test.snip
```diff
@@ -0,0 +1,156 @@
...
+            it "forwards calls to {@gRPCClass}::Stub" do
+              stub_grpc_client do |mock_client|
+                # TODO(jcanizales): Initialize enums with the second value, strings
+                # with "sample", arrays with 1 element, etc. Do the same with the
+                # mocked response.
+                # TODO(jcanizales): Instead of aliasing all used types,
+                # include Google::Cloud::Language::V1beta1 in the sample.
+                {@sampleCodeAliases(methodView)}
+
+                # expected_request = AnnotateTextRequest.new(
```
swcloud: Remove this line.

src/test/java/com/google/api/codegen/testdata/ruby_test_library.baseline
```diff
@@ -0,0 +1,911 @@
...
+        # Or iterate over results one page at a time.
+        library_service_client.list_shelves.each_page do |page|
+          # Process each page at a time.
+          page.each do |element|
+            # Process element.
+          end
+        end
+
+        # Verification:
+        response.must_equal mocked_response
```
swcloud: Where does 'response' come from in this case?

src/test/java/com/google/api/codegen/testdata/ruby_test_library.baseline
```diff
@@ -0,0 +1,911 @@
...
+          mocked_response,
+          [expected_request, Hash],
+        )
+
+        library_service_client = LibraryServiceClient.new
+        formatted_name = LibraryServiceClient.shelf_path("[SHELF_ID]")
+        library_service_client.delete_shelf(formatted_name)
+
+        # Verification:
+        response.must_equal mocked_response
```
swcloud: Where does 'response' come from in this case?

src/test/java/com/google/api/codegen/testdata/ruby_test_library.baseline
```diff
@@ -0,0 +1,911 @@
...
+        # Or iterate over results one page at a time.
+        library_service_client.list_books(formatted_name).each_page do |page|
+          # Process each page at a time.
+          page.each do |element|
+            # Process element.
+          end
+        end
+
+        # Verification:
+        response.must_equal mocked_response
```
swcloud: Where does 'response' come from in this case?

src/test/java/com/google/api/codegen/testdata/ruby_test_library.baseline
```diff
@@ -0,0 +1,911 @@
...
+          mocked_response,
+          [expected_request, Hash],
+        )
+
+        library_service_client = LibraryServiceClient.new
+        formatted_name = LibraryServiceClient.book_path("[SHELF_ID]", "[BOOK_ID]")
+        library_service_client.delete_book(formatted_name)
+
+        # Verification:
+        response.must_equal mocked_response
```
swcloud: Where does 'response' come from in this case?

src/test/java/com/google/api/codegen/testdata/ruby_test_library.baseline
```diff
@@ -0,0 +1,911 @@
...
+        # Or iterate over results one page at a time.
+        library_service_client.list_strings.each_page do |page|
+          # Process each page at a time.
+          page.each do |element|
+            # Process element.
+          end
+        end
+
+        # Verification:
+        response.must_equal mocked_response
```
swcloud: Where does 'response' come from in this case?

src/test/java/com/google/api/codegen/testdata/ruby_test_library.baseline
```diff
@@ -0,0 +1,911 @@
...
+
+        # Manually reload the operation.
+        operation.reload!
+
+        # Or block until the operation completes, triggering callbacks on
+        # completion.
+        operation.wait_until_done!
+
+        # Verification:
+        response.must_equal mocked_response
```
swcloud: Where does 'response' come from in this case?

src/test/java/com/google/api/codegen/testdata/ruby_test_library.baseline
```diff
@@ -0,0 +1,911 @@
...
+          [expected_request, Hash],
+        )
+
+        library_service_client = LibraryServiceClient.new
+        library_service_client.stream_shelves.each do |element|
+          # Process element.
+        end
+
+        # Verification:
+        response.must_equal mocked_response
```
swcloud: Where does 'response' come from in this case?

src/main/java/com/google/api/codegen/transformer/ruby/RubyTestsTransformer.java
```diff
@@ -0,0 +1,58 @@
...
+import com.google.api.codegen.util.ruby.RubyTypeTable;
+import com.google.api.codegen.viewmodel.ViewModel;
+import com.google.api.tools.framework.model.Interface;
+import com.google.api.tools.framework.model.Model;
+import java.util.ArrayList;
+import java.util.List;
+
+/** A subclass of ModelToViewTransformer which translates model into API tests in Ruby. */
+public class RubyTestsTransformer implements ModelToViewTransformer {
+  // Template files
```
shinfan: Unnecessary comment

src/main/java/com/google/api/codegen/transformer/ruby/RubyTestsTransformer.java
```diff
@@ -0,0 +1,58 @@
...
+  @Override
+  public List<String> getTemplateFileNames() {
+    List<String> fileNames = new ArrayList<>();
+    fileNames.add(UNIT_TEST_TEMPLATE_FILE);
+    return fileNames;
+  }
+
+  @Override
+  public List<ViewModel> transform(Model model, ApiConfig apiConfig) {
+
```
shinfan: remove empty line

src/main/java/com/google/api/codegen/transformer/ruby/RubyTestsTransformer.java
```diff
@@ -0,0 +1,58 @@
...
+    SurfaceNamer namer = new RubySurfaceNamer(apiConfig.getPackageName());
+    RubyGapicContext util = new RubyGapicContext(model, apiConfig);
```
shinfan: s/util/context

src/main/java/com/google/api/codegen/transformer/ruby/UnitTestsViewModel.java
shinfan: Generally we should reuse the views as much as we can. Feel free to add more fields to the existing views if anything is missing for Ruby.

src/main/java/com/google/api/codegen/transformer/ruby/UnitTestCaseViewModel.java
shinfan: Instead of creating a new view, adding 'ApiMethodView' in the 'TestCaseView' should be the better approach here.

src/main/java/com/google/api/codegen/transformer/ruby/RubySurfaceNamer.java
shinfan: Why the suffix of the file is '_test.jorge.rb'?

src/main/java/com/google/api/codegen/transformer/ruby/GeneratedLibraryViewModel.java
```diff
@@ -0,0 +1,55 @@
...
+ * Ruby-specific about it.
+ */
+// TODO(jcanizales): Move it to/merge it with something common to all languages. Reuse it for Ruby
+// MVVM.
+public class GeneratedLibraryViewModel {
+  private final Interface service;
+  private final SurfaceNamer namer;
+  private final ModelTypeTable typeTable;
+
+  public GeneratedLibraryViewModel(
```
shinfan: These view classes do not follow the concept of MVVM, where view models are only data containers. We should not include any namers or type tables in the views, put this kind of logic into transformers instead.

Also this view class should not be necessary if you reuse `GapicSurfaceTestClassView`

src/main/java/com/google/api/codegen/gapic/MainGapicProviderFactory.java
jcanizales: Consistency is good, but what's the `GapicSurface` part in those names for? After renaming `RubyTestsTransformer` should I leave a TODO to remove that part from the names?

src/main/java/com/google/api/codegen/gapic/MainGapicProviderFactory.java
jcanizales: Done

src/main/java/com/google/api/codegen/transformer/ruby/RubyTestsTransformer.java
jcanizales: Done

src/main/java/com/google/api/codegen/transformer/ruby/RubySurfaceNamer.java
jcanizales: Sorry, this method wasn't used anymore but I had forgotten to delete it.

src/main/resources/com/google/api/codegen/ruby/test.snip
jcanizales: Done

src/main/resources/com/google/api/codegen/ruby/test.snip
jcanizales: Done

src/main/resources/com/google/api/codegen/ruby/test.snip
jcanizales: Done

src/main/java/com/google/api/codegen/transformer/ruby/RubyTestsTransformer.java
jcanizales: Removed

src/main/java/com/google/api/codegen/transformer/ruby/RubyTestsTransformer.java
jcanizales: Done

src/main/java/com/google/api/codegen/transformer/ruby/RubyTestsTransformer.java
jcanizales: Done

src/main/java/com/google/api/codegen/transformer/ruby/RubySurfaceNamer.java
jcanizales: This method wasn't used anymore and I had forgotten to delete it, sorry.

bjwatson: When I started my career, I worked on communication satellites. The techs always hung big red flags from anything attached to the satellite that should not be there on launch day. My cyber-equivalent for coding is to put a FIXME comment next to anything I want to purge before merge. As others have pointed out to me, FIXME comments do not conform to the Google standard, which is exactly the point: they're easy to find in a `git grep`, and they should be removed (along with the offending code) before code review is complete.

src/main/java/com/google/api/codegen/transformer/ruby/GeneratedLibraryViewModel.java
jcanizales: > view models are only data containers

The instances are immutable: what good does it make to calculate the values of the accessors on construction, vs on access?

Making the view model class an `AutoValue` has the following drawbacks:
* All values have to be calculated on construction, making that code very complex and hard to share between languages.
* No value can be `null`, even for things that a language doesn't use (e.g. https://github.com/googleapis/toolkit/pull/793#discussion_r91332509 ).
* The class definition ends up with 2x boilerplate, to get a `Builder` for it.
* If one language or feature needs a new field, transformers of all languages have to populate it or the constructor throws.
* The class tends to end up as a giant bag of undocumented fields. With a "regular" class, the implementation tells what an accessor returns, with an `AutoValue`, you have to fish for the places where it's constructed.
* The result is very un-OOP-like.

I tried that approach first for this PR and was burned very quickly. I'm not against `AutoValue` in general. It's very useful for data structures in which having structural equality (instead of referential equality) is necessary. But we don't need that for view models as far as I've seen, and it becomes a big pain.

garrettjonesgoogle: Values can be null if you mark them `@Nullable`.

src/main/java/com/google/api/codegen/transformer/ruby/UnitTestCaseViewModel.java
jcanizales: `ApiMethodView` is an empty interface. What purpose does it serve?

src/main/java/com/google/api/codegen/transformer/ruby/GeneratedLibraryViewModel.java
jcanizales: If a field is `@Nullable`, can it be skipped in the builder, or does the `null` need to be explicitly set?

garrettjonesgoogle: It can be skipped in the builder.

jcanizales:
(I'm still working on the bigger changes).

src/main/java/com/google/api/codegen/gapic/MainGapicProviderFactory.java
swcloud: 'GapicSurfaceTestTransformer' indicates that the test is for surfaces created at the gapic layer.

bjwatson:
@jcanizales Will this still be ready to merge this week? I don't see any updates in the past two weeks, and it has conflicts with master.

jcanizales:
@swcloud, when merging master into my branch to resolve conflicts, a bug showed up:

In the generated sample for `publish_books`, the three fields of the request proto (which are flattened) are initialized like this:

```ruby
shelf = Shelf.new
books = []
series_string = "foobar"
series_uuid = SeriesUuid.new
series_uuid.series_string = series_string
```

Note the intermediate variable `series_string`.

Well, in the generated test, the request proto is also initialized unflattened, for the gRPC mock to compare against the received proto. My initialization view is an adaptation of the corresponding samplegen view, that outputs ":" instead of "=" among other tweaks, to produce:

```ruby
expected_request = PublishSeriesRequest.new(
  shelf: Shelf.new,
  books: [],
  series_string: "foobar",
  series_uuid: SeriesUuid.new(
    series_string: series_string
  ),
)
```

And the intermediate variable now breaks it, because `PublishSeriesRequest` has no `series_string` field :)

The solution is to just reuse the samplegened code as is, and append a proto initializer this way:

```ruby
shelf = Shelf.new
books = []
series_string = "foobar"
series_uuid = SeriesUuid.new
series_uuid.series_string = series_string

expected_request = PublishSeriesRequest.new(
  shelf: shelf,
  books: books,
  series_uuid: series_uuid,
)
```

This will decouple the proto initializer from the init code of the sample gen.

src/main/java/com/google/api/codegen/transformer/ruby/GeneratedLibraryViewModel.java
shinfan: View objects are designed to be language-independent and the transformers for each language are suppose to be language-dependent. Adding transformer logic into view classes conflicts with this concept and prevents the flexibility of the fields used by multiple languages.

For example, the field `grpcClassName` in the view class `MockServiceImplView` is used to store both server type name for NodeJS but client type name for PHP:
Node: https://github.com/googleapis/toolkit/blob/master/src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSGapicSurfaceTestTransformer.java#L98

PHP:https://github.com/googleapis/toolkit/blob/28aeed1615b8a2417e990088f5d24eab457cc7fe/src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTestTransformer.java#L107

If you find many duplicate code during construction time across different languages, the common way is to create a shared transformer, see `TestCaseTransformer`: 
https://github.com/googleapis/toolkit/blob/master/src/main/java/com/google/api/codegen/transformer/TestCaseTransformer.java

swcloud:
@jcanizales I think you can directly reuse 'initCode' defined in ruby/initcode.snip such as

{@initCode(apiMethod.initCode)}

src/main/java/com/google/api/codegen/transformer/ruby/UnitTestCaseViewModel.java
swcloud: ApiMethodView is just an abstraction. The real view that's used by dynamic languages is the implementation 'OptionalArrayMethodView'. I think you need to address this fundamental design issue.

swcloud:
@jcanizales Can you resolve the conflict?

I think some design issues have not been addressed since last review.  I suggest that's the focus now.

src/main/resources/com/google/api/codegen/ruby/test.snip
```diff
@@ -0,0 +1,173 @@
...
+          end
+          
+        @end
+      @end
+    @end    
+  end
+
+@end
+
+# The following were adapted from initcode.snip.
```
swcloud: You can remove these duplicated code and reuse the initcode.snip, right?

src/main/resources/com/google/api/codegen/ruby/method_sample.snip
```diff
@@ -13,20 +13,32 @@
...
+
+@snippet sampleCodeCore(apiMethod)
   {@apiMethod.apiVariableName} = {@apiMethod.apiClassName}.new
   @if apiMethod.initCode.lines
     {@initCode(apiMethod.initCode)}
   @end
+  # The different *SampleCode(apiMethod) snippets include calls to
+  # non-declared methods, like:
+  #
+  # @# Process the response.
```
swcloud: I don't think you should put these 'process_xxx' calls here in 'sampleCodeCore' inside method_sample.snip, because these calls are defined just for the purpose of unit tests. SampleGen should stay as the purpose of generating samples. Plus, it will break smoke tests since they don't know what these calls are. So these calls should be moved out of method_sample.snip.

jcanizales: The fix for smoke test is trivial, I already have it. It's very hard, if at all possible, to put these outside method_sample.snip, because of these samples:

```ruby
library_service_client = LibraryServiceClient.new

# Iterate over all results.
library_service_client.list_shelves.each do |element|
  # Process element.
  process_element element
end

# Or iterate over results one page at a time.
library_service_client.list_shelves.each_page do |page|
# Process each page at a time.
  page.each do |element|
    # Process element.
    process_element element
  end
end
```

```ruby
library_service_client = LibraryServiceClient.new

# Register a callback during the method call.
operation = library_service_client.get_big_book(formatted_name) do |op|
  raise op.results.message if op.error?
  op_results = op.results
  # Process the results.
  process_results op_results

  metadata = op.metadata
  # Process the metadata.
  process_metadata metadata
end

# Or use the return value to register a callback.
operation.on_done do |op|
  raise op.results.message if op.error?
  op_results = op.results
  # Process the results.
  process_results op_results

  metadata = op.metadata
  # Process the metadata.
  process_metadata metadata
end

# Manually reload the operation.
operation.reload!

# Or block until the operation completes, triggering callbacks on
# completion.
operation.wait_until_done!
```

Initially when I was only doing tests of basic unary calls, the verification check was being output by test.snip, just after the sample.

I'll be very happy to have a workaround or alternative, though!

src/main/resources/com/google/api/codegen/ruby/test.snip
jcanizales: Check the difference between what `{@initCode(apiMethod.initCode)}` outputs:

```ruby
shelf = Shelf.new
books = []
series_string = "foobar"
series_uuid = SeriesUuid.new
series_uuid.series_string = series_string
```

and what this outputs:

```ruby
shelf: Shelf.new,
books: [],
series_string: "foobar",
series_uuid: SeriesUuid.new(
  series_string: series_string
),
```

jcanizales: This is problematic anyway and I'm in the process of removing it, as explained here: https://github.com/googleapis/toolkit/pull/793#issuecomment-284502760

jcanizales:
@swcloud Agreed with focus on design feedback and @shinfan's guidance. But I think the first priority is to unbreak the tests so they actually work again (solve https://github.com/googleapis/toolkit/pull/793#issuecomment-284502760).

jcanizales:
Unbroken - now merging master to resolve conflicts again.

src/main/resources/com/google/api/codegen/ruby/smoke_test.snip
```diff
@@ -23,7 +23,15 @@ describe "{@smokeTest.name}" do
       project_id = ENV["SMOKE_TEST_PROJECT"].freeze
     @end
 
-    {@sampleCodeCore(smokeTest.apiMethod)}
+    process_response = ->(response) {
```
swcloud: I still don't think we should put a dummy process_response definition here. This way we're forcing the caller of the samplegen to provide a response handler. Why can't you call 'process_xxx' after {@sampleCodeCall(methodView)} in test.snip just  for the unit test purpose?

src/main/resources/com/google/api/codegen/ruby/smoke_test.snip
```diff
@@ -23,7 +23,15 @@ describe "{@smokeTest.name}" do
...
+      @# Nothing to do.
+    }
+
+    {@sampleCodeAliases(smokeTest.apiMethod)}
```
swcloud: Even if you break the sampleCodeCore into 4 different pieces for the reuse purpose of unit test gen, I think we should still provide a simple call entry 'sampleCodeCore', which can be called by smoke test directly, because smoke test does not care about this finer granularity, at least for now.

src/main/resources/com/google/api/codegen/ruby/smoke_test.snip
jcanizales: That works only for basic unary calls, and it was how I was doing it initially. How do you propose we handle [these other cases](https://github.com/googleapis/toolkit/pull/793#discussion_r105246194) though?

swcloud: I can see the reason why you want to directly re-use samplegen code as a drop-in, but I suspect that may not work, for example, in LRO, the sample code has 'operation.wait_until_done!' which is a block call, in the unit-test case, you might not want that. After all, the unit test here is just sort of compiler check for the API calls. There is no need to test operation.reload! or operation.wait_until_done! So I think you may end up with tweaking a lot of samplegen code. Instead, I would focus on the code to make API calls.

jcanizales: Good catch about LRO. Having the gRPC mock return a done operation when the method is LRO would prevent it from blocking.

In any case, because `operation.wait_until_done!` etc. aren't generated, I'm fine with leaving them out of the generated test if that makes it easier.

But that's not the issue though, rather: methods whose response is async can't be verified by appending lines to the sample code. Because the response is within a callback. E.g. for this method:

```ruby
library_service_client = LibraryServiceClient.new

# Iterate over all results.
library_service_client.list_shelves.each do |element|
  # Process element.
end

# Or iterate over results one page at a time.
library_service_client.list_shelves.each_page do |page|
  # Process each page at a time.
  page.each do |element|
    # Process element.
  end
end
```

We can't just add the verification at the end.

src/main/resources/com/google/api/codegen/ruby/test.snip
```diff
@@ -0,0 +1,124 @@
...
+  @#
+  @# Unless required by applicable law or agreed to in writing, software
+  @# distributed under the License is distributed on an "AS IS" BASIS,
+  @# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  @# See the License for the specific language governing permissions and
+  @# limitations under the License.
+
+  require "minitest/autorun"
+  require "{@testViewModel.libraryUnderTest.outputFile}"
+  include {@testViewModel.libraryUnderTest.namespace}
```
remi: Is there a reason why this `include` is global, rather than in the `describe` block?

This is a bit of a code smell.  Using `include` at all to include members into the current scope for convenience is also a bit non-idiomatic.

If you want to not have to use fully-qualified class names in the test, have you considered using a constant alias rather than include?

```ruby
describe {@testViewModel.libraryUnderTest.apiClassName} do
  {@testViewModel.libraryUnderTest.apiClassName} = {@testViewModel.libraryUnderTest.namespace}
```

If those are the right template properties... inotherwords something that generates like:

```ruby
describe {@testViewModel.libraryUnderTest.apiClassName} do
  Language = Google::Cloud::Language::V1beta1

  # ...
  Language::LanguageService::Stub.stub(:new, mock_client) { yield(mock_client) }
```

This is similar to what we do in code samples using the libraries.

jcanizales: I was trying to avoid what the current samples do of having a big block of aliasing each class. I presumed the `include` would be how people use the generated library, so it made sense to do the same in the test.

IIRC, I got errors when trying to scope the include other than globally.

I hadn't thought of aliasing the namespace itself. As it's just one line, I'd be all on for it.

PR #796, opened 2016-11-28
Python wrappers: return proto Empty type

### LINKED ISSUE #760: [Python] Return empty messages (except protobuf.Empty)
### What:
The following Python methods should return a value (as their samples suggest):

https://github.com/googleapis/toolkit/blob/master/src/test/java/com/google/api/codegen/testdata/python_main_library.baseline#L1200
https://github.com/googleapis/toolkit/blob/master/src/test/java/com/google/api/codegen/testdata/python_main_library.baseline#L1168

### Why:
The API owners might intend to add content to these messages later, and we don't want our client libraries to break users as a result.

### How:
This appears to be the responsible code:
https://github.com/googleapis/toolkit/blob/master/src/main/java/com/google/api/codegen/py/PythonProtoElements.java#L35-L37

geigerj:
Previously returned None.

Fixes #760.

### DIFF
--- a/src/main/java/com/google/api/codegen/py/PythonGapicContext.java
+++ b/src/main/java/com/google/api/codegen/py/PythonGapicContext.java
@@ -225,6 +225,9 @@ public class PythonGapicContext extends GapicContext {
   private String returnTypeComment(
       Method method, MethodConfig config, PythonImportHandler importHandler) {
     MessageType returnMessageType = method.getOutputMessage();
+    if (PythonProtoElements.isEmptyMessage(returnMessageType)) {
+      return null;
+    }
 
     String path = importHandler.elementPath(returnMessageType, true);
     String classInfo = ":class:`" + path + "`";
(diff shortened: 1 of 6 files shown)


geigerj:
Sorry, I realize now that we do not need to return `google.protobuf.Empty`, only *other* empty messages. I need to revise this PR.

geigerj:
@jmuk @swcloud Since this is beta blocking for Python and @bjwatson is OOO, it would be great if one of you has time to do this review.

geigerj:
@jmuk Thanks!

swcloud:
Actually I think it's better use Empty.getDescriptor().getFullName() instead of using the hard-coded string. Also I think you should call MethodConfig.isReturnEmptyMessageMethod.

geigerj:
@swcloud OK, I will make that change on a follow-up PR.

bjwatson:
@geigerj Which PR or issue is this? Does this follow-up PR need to be done for beta?

geigerj:
No, it is a style change. There is no impact on the surface. It was done in #797.

bjwatson:
Got it. Thanks!

PR #799, opened 2016-11-28
Fix Python namespace package separator

geigerj:
Previously Python packages use the file separator (e.g., '/') instead
of the standard dot separator.


src/main/java/com/google/api/codegen/metadatagen/py/PythonPackageCopier.java
```diff
@@ -81,7 +82,8 @@ public FileVisitResult postVisitDirectory(Path dir, IOException exc) {
       } else {
         docBuilder.put(
             outFile, Doc.text("__import__('pkg_resources').declare_namespace(__name__)\n"));
-        pythonNamespacePackages.add(inputPath.relativize(dir).toString());
+        pythonNamespacePackages.add(
+            inputPath.relativize(dir).toString().replace(File.separator, "."));
```
jmuk: `Joiner.on(".").join(inputPath.relativize(dir).iterator())` might work?

geigerj: Ah, that is nicer. Done.

PR #800, opened 2016-11-29
Fix handling of Any resource name

michaelbausor:
- Moves logic for creating FieldConfig objects with resource names out of PageStreamingConfig and PathTemplateTransformer, and into the FieldConfig static factory methods.
- Updates C# to use IResourceName instead of ResourceName
- Fixes some error handling


src/main/java/com/google/api/codegen/transformer/csharp/CSharpSurfaceNamer.java
```diff
@@ -232,8 +233,8 @@ public String getExamplePackageName() {
   }
 
   @Override
-  public String getAnyFieldResourceTypeName() {
-    return "IResourceName";
+  protected Name getAnyResourceTypeName() {
+    return Name.from("i_resource_name");
```
garrettjonesgoogle: I think using `Name.anyCamel("IResourceName")` would be more natural.

Issue #804, opened 2016-11-29
Dummy Python message classes contain extraneous data, imports


geigerj:
### What
A proto message is "transitively referenced" by an API if there is some method of the API whose request or response message contains a field whose type is, or whose type transitively references that proto message. `ProtoFileView` compiles the proto files containing all such transitiviely referenced proto types.

However, the proto files may also contain other messages and imports that are _not_ transitively referenced by the API. We should not generate Python docs or imports for these messages.

### Why
The imports are broken if they point to files that are not contained in the `ProtoFileView` (since we do not generate their corresponding dummy Python module), causing Sphinx doc generation to fail.

The extraneous messages don't break anything, but they're irrelevant and may themselves contain broken links (if they link to files not contained in the `ProtoFileView`), so they should be excluded as well.

geigerj:
@lukesneeringer

eoogbe:
Closing because these classes do not appear to exist any longer.

PR #806, opened 2016-11-29
Updates to dummy class doc generation

geigerj:
- Generate dummy classes for both input and output types (previously,
  only input types were generated)
- In Python, don't generate imports for classes where there is no
  dummy class, since such imports will break during Sphinx doc
  generation.

Updates #804 

cc: @swcloud @jmuk, this PR adds additional classes to the dummy proto classes, since we previously didn't generate dummy classes for output types.

### DIFF
--- a/src/main/java/com/google/api/codegen/py/PythonImportHandler.java
+++ b/src/main/java/com/google/api/codegen/py/PythonImportHandler.java
@@ -33,6 +33,7 @@ import com.google.common.collect.Lists;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Set;
 
 public class PythonImportHandler {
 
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/api/codegen/py/PythonImportHandler.java
```diff
@@ -104,14 +105,15 @@ public PythonImportHandler(Interface service, ApiConfig apiConfig) {
...
 
   /** This constructor is used for doc messages. */
-  public PythonImportHandler(ProtoFile file) {
+  public PythonImportHandler(ProtoFile file, Set<ProtoFile> importableProtoFiles) {
     for (MessageType message : file.getMessages()) {
       for (Field field : message.getMessageFields()) {
         MessageType messageType = field.getType().getMessageType();
         // Don't include imports to messages in the same file.
-        if (!messageType.getFile().equals(file)) {
+        ProtoFile importFile = messageType.getFile();
```
ethanbao: nit: call it messageParentFile.

src/main/java/com/google/api/codegen/py/PythonProtoFileInitializer.java
```diff
@@ -14,14 +14,23 @@
...
 
+import com.google.api.codegen.ProtoFileView;
 import com.google.api.tools.framework.model.ProtoFile;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Sets;
+import java.util.Set;
 
 public class PythonProtoFileInitializer implements PythonSnippetSetInputInitializer<ProtoFile> {
 
+  private Set<ProtoFile> importableProtoFiles = null;
```
ethanbao: Why not turn it into a method-level variable?

geigerj: If we did, we would traverse the entire model every for every proto file we create a dummy of. This way we only do the computation once.

src/main/java/com/google/api/codegen/py/PythonImportHandler.java
geigerj: Done

src/test/java/com/google/api/codegen/testdata/library.proto
```diff
@@ -6,6 +6,7 @@ package google.example.library.v1;
 
 import "field_mask.proto";
 import "google/api/annotations.proto";
+import "google/api/monitored_resource.proto";
```
geigerj: Done

PR #809, opened 2016-11-30
add go happy-case unit testing

pongad:
This commit adds a change to make this easier:

The "zero value" of a message type *T is &T{} instead of nil.

This necessitates a small change in iterator implementation,
though the behavior should stay the same.


src/test/java/com/google/api/codegen/testdata/go_main_library.baseline
```diff
@@ -688,13 +688,13 @@ func (it *BookIterator) PageInfo() *iterator.PageInfo {
 
 // Next returns the next result. Its second return value is iterator.Done if there are no more
 // results. Once Next returns Done, all subsequent calls will return Done.
-func (it *BookIterator) Next() (*librarypb.Book, error) {
-    if err := it.nextFunc(); err != nil {
-        return nil, err
+func (it *BookIterator) Next() (item *librarypb.Book, err error) {
```
jba: Can you revert to the previous implementation? They seem identical in behavior (unless I'm missing something) and the original is more idiomatic Go.

jba: Your change to the generator is fine by me, but you could also write the code this way:

```
func (it *BookIterator) Next() (*librarypb.Book, error) {
    var item *libraryp.Book
    if err := it.nextFunc(); err != nil {
        return item, err
    }
    item = it.items[0]
    it.items = it.items[1:]
    return item, nil
}
```
I would much prefer that.

pongad:
@jba I see what you're saying. PTAL

pongad:
@garrettjonesgoogle Could you take a look since this affects MVVM files?

src/main/resources/com/google/api/codegen/go/mock.snip
```diff
@@ -120,111 +125,181 @@
...
-        return err
+        return nil
     }
 @end
 
-@private testError(test)
-    func {@test.nameWithException}(t *testing.T) {
-        errCode := codes.Internal
-        {@test.mockServiceVarName}.err = grpc.Errorf(errCode, "test error")
+@private testFunc(test)
```
garrettjonesgoogle: this implies a single test function, but there are two - maybe call this `testFunctions`

src/main/resources/com/google/api/codegen/go/initcode.snip
```diff
@@ -0,0 +1,57 @@
...
+    {@renderResourceName(initValue)}
+  @case "ResourceNameOneofInitValueView"
+    {@initValue.resourceOneofTypeName}.from({@renderResourceName(initValue.specificResourceNameView)})
+  @end
+@end
+
+@private argList(args)
+  @join a : args on ", "
+    {@a}
+  @end
```
garrettjonesgoogle: The other Go snippets are indented differently - shouldn't this be indented the same as them?

src/main/resources/com/google/api/codegen/go/mock.snip
pongad: Done

src/main/resources/com/google/api/codegen/go/initcode.snip
pongad: Changed to 4-space. Thank you for noticing!

PR #813, opened 2016-12-01
Nodejs: Use LRO wrapper.

landrito:


src/test/java/com/google/api/codegen/testdata/nodejs_doc_main_library.baseline
```diff
@@ -115,13 +115,42 @@ function LibraryServiceClient(gaxGrpc, grpcClients, opts) {
...
+
+  extend(protoDescriptorPool, gaxGrpc.load([{
+    root: require('google-proto-files')('..'),
+    file: 'library.proto'
+  }]));
+
+  extend(protoDescriptorPool, gaxGrpc.load([{
+    root: require('google-proto-files')('..'),
+    file: 'tagger.proto'
+  }]));
```
jmuk: I'm afraid that this does not work. `extend` overrides existing properties on the new items, and because `gaxGrpc.load` returns `google` as for its top namespace, the entire protoDescriptorTool result will be replaced by the result of `tagger.proto`.

See:

```javascript
> var extend = require('extend');
undefined
> var x = {foo: {bar: {baz: 1}}}
undefined
> var y = {foo: {baz: 2}}
undefined
> extend(extend({}, x), y)
{ foo: { baz: 2 } }
```

You should use `lodash.merge` if you need things like this.

jmuk: Also, this is quite similar to `grpcClients` structure. It would be better to consolidate them into a single construct, so that it does not have to load/parse the proto twice.

src/test/java/com/google/api/codegen/testdata/nodejs_doc_main_library.baseline
```diff
@@ -1691,18 +1720,57 @@ LibraryServiceClient.prototype.addLabel = function(request, options, callback) {
  * @param {function(?Error, ?Object)=} callback
  *   The function which will be called with the result of the API call.
  *
- *   The second parameter to the callback is an object representing [google.longrunning.Operation]{@link external:"google.longrunning.Operation"}.
+ *   The second parameter to the callback is a [gax.Operation]{@link https://googleapis.github.io/gax-nodejs/global.html#Operation} object.
```
jmuk: `Operation` is a class, therefore the URL would be like .../Operation.html, not under global.html (see http://googleapis.github.io/gax-nodejs/PageDescriptor.html for example).

src/test/java/com/google/api/codegen/testdata/nodejs_doc_main_library.baseline
landrito: Oh interesting! I am using lodash.merge now.

src/test/java/com/google/api/codegen/testdata/nodejs_doc_main_library.baseline
```diff
@@ -31,6 +31,7 @@
 var configData = require('./library_service_client_config');
 var extend = require('extend');
 var gax = require('google-gax');
+var _ = require('lodash');
```
jmuk: better to use 'lodash.merge' module to minimize the dependency size: https://www.npmjs.com/package/lodash.merge/

src/test/java/com/google/api/codegen/testdata/nodejs_doc_main_library.baseline
landrito: Done!

landrito:
Thanks for the review!

PR #815, opened 2016-12-01
Use fullName instead of shortName where it makes sense

### LINKED ISSUE #814: Switch Python PackageMetadataGen templates to use FullName consistently
### What

Many of the occurrences of `{@context.getApiNameInfo.shortName}` in the templates would be better served by `{@context.getApiNameInfo.fullName}`. Audit all occurences of `shortName` and replace by `fullName` where appropriate. Example: https://github.com/googleapis/toolkit/blob/master/src/main/resources/com/google/api/codegen/metadatagen/py/setup.py.snip#L2

### Why
`shortName` doesn't pick up branding.

geigerj:
This is an essentially superficial change that improves the descriptions of the gRPC packages by including the full API name.

Fixes #814


bjwatson:
LGTM

PR #818, opened 2016-12-02
Treat google.iam as common proto package

geigerj:
Context is the same as with https://github.com/googleapis/googleapis/pull/209; this is needed to prevent Python codegen from changing the package to `google.cloud.grpc.iam.[...]`.

### DIFF
--- a/src/main/java/com/google/api/codegen/py/PythonImportHandler.java
+++ b/src/main/java/com/google/api/codegen/py/PythonImportHandler.java
@@ -40,6 +40,7 @@ public class PythonImportHandler {
   // TODO (geigerj): Read this from configuration?
   private final List<String> COMMON_PROTOS =
       Lists.newArrayList(
+          "google.iam",
           "google.protobuf",
           "google.api",
           "google.longrunning",


bjwatson:
LGTM, but I want to get buy in from a few other people due to the broad impact of changes to `common-protos`.

FYI, @garrettjonesgoogle, @jba, @jskeet, @jmuk, @jonparrott, @dhermes, @swcloud, @michaelbausor.

jskeet:
We already have Google.Iam (going to be Google.Cloud.Iam soon) as a separate package for C#, just like Google.Longrunning (going to be Google.LongRunning). We don't have either of those within our Google.Api.CommonProtos package. I personally have no objection to inconsistency between languages on this particular point.

bjwatson:
@jskeet Would this change cause your `common-protos` build to include IAM? IIUC, it affects all languages.

Similarly, when I added LRO to this list a couple months ago, I might have affected the `common-protos` build for all languages.

Sounds like you consider IAM to be a Cloud meta-API. @garrettjonesgoogle had a different opinion about this, since it could be used by non-cloud APIs.

jskeet:
This pull request only seems to change the Python imports - I can't see how it would affect anything else. Our common protos are built in the gax-dotnet repo, and the build process is entirely independent of anything else here.

We debated about whether IAM should be Google.Cloud.Iam or Google.Iam for .NET namespaces, and decided that Google.Cloud.Iam would be more appropriate for almost all use cases (given the description in policy.proto "Defines an Identity and Access Management (IAM) policy. It is used to specify access control policies for Cloud Platform resources"). Where it's used for a non-Cloud API, the namespace certainly won't cause any issues.

geigerj:
@bjwatson @jskeet This change and https://github.com/googleapis/artman/pull/122 affects only Python.

@garrettjonesgoogle, @jba, @jskeet, @jmuk, @jonparrott, @dhermes, @swcloud, @michaelbausor I intended to discuss the broader change outside of this PR -- would appreciate if we can get this PR in separately to keep things moving for beta.

bjwatson:
Oh, got it. I missed that the filename is *Python*ImportHandler.

PR #820, opened 2016-12-02
Remove format and parse methods from Java

michaelbausor:
- Remove format and parse methods
- Remove path templates
- Remove VALIDATION setting from resource name treatment


src/main/java/com/google/api/codegen/transformer/FeatureConfig.java
```diff
@@ -40,4 +40,9 @@ public boolean enableMixins() {
   public boolean enableGrpcStreaming() {
     return false;
   }
+
+  /** Returns true is string format functions are supported. */
```
shinfan: s/is/if/

src/test/java/com/google/api/codegen/testdata/java_test_library.baseline
```diff
@@ -171,11 +171,11 @@ public class LibraryServiceTest {
   @Test
   @SuppressWarnings("all")
   public void getShelfTest() {
-    String formattedName2 = LibraryServiceClient.formatShelfName("[SHELF_ID]");
+    String name2 = "";
```
shinfan: This does not look correct. The assigned value should be a randomized string (not empty).

PR #821, opened 2016-12-05
Add C# request methods

chrisdunelm:
And refactor all other methods to call through these new methods

### DIFF
--- a/src/test/java/com/google/api/codegen/testdata/csharp_gapic_snippets_library.baseline
+++ b/src/test/java/com/google/api/codegen/testdata/csharp_gapic_snippets_library.baseline
@@ -60,6 +60,36 @@ namespace Google.Example.Library.V1.Snippets
             // End snippet
         }
 
+        public async Task CreateShelfAsync_RequestObject()
+        {
+            // Snippet: CreateShelfAsync(CreateShelfRequest,CallSettings)
+            // Create client
+            LibraryServiceClient libraryServiceClient = LibraryServiceClient.Create();
+            // Initialize request argument(s)
+            CreateShelfRequest request = new CreateShelfRequest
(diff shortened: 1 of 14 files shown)


jskeet:
Just looked at the C# results so far - I like this a lot. It feels like a much nicer separation of "API surface exposed" and "implementation of API". In particular, I think this approach will work better with users creating fakes - they *only* need to fake out the request-taking methods, and everything should work.

chrisdunelm:
PTAL

src/test/java/com/google/api/codegen/testdata/csharp_gapic_snippets_library.baseline
```diff
@@ -141,6 +171,38 @@ namespace Google.Example.Library.V1.Snippets
...
 
+        public async Task GetShelfAsync_RequestObject()
+        {
+            // Snippet: GetShelfAsync(GetShelfRequest,CallSettings)
+            // Create client
+            LibraryServiceClient libraryServiceClient = LibraryServiceClient.Create();
+            // Initialize request argument(s)
+            GetShelfRequest request = new GetShelfRequest
+            {
+                Name = new ShelfName("[SHELF_ID]").ToString(),
```
jskeet: I think it would be better to generate snippets using the added properties, e.g.

```csharp
ShelfName = new ShelfName(...)
```

but that can *definitely* come as a follow-up

chrisdunelm: Hmm, many are (e.g. line 467).
I suspect this hasn't been defined as a resource-name in the test GAPIC configuration. Very possibly deliberately to test that it works properly.

PR #826, opened 2016-12-07
Fix oneof format samples

michaelbausor:
Return the first SingleResourceNameConfig from a ResourceNameOneofConfig when generating samples


garrettjonesgoogle:
LGTM

PR #828, opened 2016-12-07
Fix partial classes generation for resource-names

chrisdunelm:
Generate partial resource-name classes for proto messages from all services.


jskeet:
Sorry, that should have been "approve", not "request changes". Code review on a phone.

PR #837, opened 2016-12-08
C# Snippet updates

### LINKED ISSUE #835: XyzOneof.GetHashCode constructs string unnecessarily
This:

```csharp
public override int GetHashCode() => ToString().GetHashCode()
```

would be better as:

```csharp
public override int GetHashCode() => Name.GetHashCode()
```

Currently it does the same thing, but resource names could override `GetHashCode` more efficiently (e.g. to call `GetHashCode` on each part and combine them. It does require that every resource name type implements `GetHashCode`, but we should document that as a requirement anyway.

It would be nice to make `Equals` more efficient too - but that's harder, as we'd want an `Unknown` resource name and a "known" resource name with the same string representation to be equal. But if the types are the same, we can call `Equals`, and if they're different but neither is `Unknown`, we can return false.

This is not urgent :)

### LINKED ISSUE #836: C# async snippets use sync service Create() method
They should use CreateAsync() instead.

chrisdunelm:
Fixes #836 (Async snippets not using CreateAsync)

### DIFF
--- a/src/main/resources/com/google/api/codegen/csharp/gapic_snippets.snip
+++ b/src/main/resources/com/google/api/codegen/csharp/gapic_snippets.snip
@@ -47,7 +47,7 @@
     public void {@snippetMethod.snippetMethodName}()
     {
         // Snippet: {@snippetMethod.method.name}({@snippetArgTypes(snippetMethod.method.methodParams)},CallSettings)
-        {@preparations(snippetMethod)}
+        {@preparations(snippetMethod, FALSE)}
         @if snippetMethod.method.hasReturnValue
             {@snippetMethod.callerResponseTypeName} response = {@sampleMethodCall(snippetMethod)};
         @else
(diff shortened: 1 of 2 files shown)


src/main/resources/com/google/api/codegen/csharp/gapic_resourcenames.snip
```diff
@@ -292,13 +292,13 @@
...
         /// <inheritdoc />
-        public override int GetHashCode() => ToString().GetHashCode();
+        public override int GetHashCode() => Name.GetHashCode();
 
         /// <inheritdoc />
         public override bool Equals(object obj) => Equals(obj as {@resource.typeName});
 
         /// <inheritdoc />
-        public bool Equals({@resource.typeName} other) => ToString() == other?.ToString();
+        public bool Equals({@resource.typeName} other) => Name.Equals(other?.Name);
```
jskeet: I don't think we want to do this part yet - because we still want an "unknown" which is actually a ProjectName to match a *real* ProjectName, if they have the same string representation. This will take a bit more work.

chrisdunelm: Following on from offline discussion, the hashcode and equality changes have been reverted.

chrisdunelm:
PTAL

PR #838, opened 2016-12-08
Adding ReleaseLevel concept, supporting in Java

garrettjonesgoogle:
The "unset" case is present in the no_path_templates service.

### DIFF
--- a/src/test/java/com/google/api/codegen/testdata/java_main_library.baseline
+++ b/src/test/java/com/google/api/codegen/testdata/java_main_library.baseline
@@ -153,7 +153,6 @@ import javax.annotation.Generated;
  * </pre>
  */
 @Generated("by GAPIC")
-@ExperimentalApi
 public class LibraryServiceClient implements AutoCloseable {
   private final LibraryServiceSettings settings;
   private final ScheduledExecutorService executor;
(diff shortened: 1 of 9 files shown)


src/test/java/com/google/api/codegen/testdata/java_main_library.baseline
```diff
@@ -153,7 +153,6 @@ import javax.annotation.Generated;
  * </pre>
  */
 @Generated("by GAPIC")
-@ExperimentalApi
```
michaelbausor: Did you intent to test the ALPHA case, when this would still show up?

michaelbausor:
Also, where you intending to set ALPHA on all APIs? Otherwise you might want the unset case to include experimentalApi for Java

src/test/java/com/google/api/codegen/testdata/java_main_library.baseline
garrettjonesgoogle: I didn't want to create two whole new services (for ALPHA and GA), that would be a bit of overhead.

garrettjonesgoogle:
I was going to leave APIs unset for now. The unset case *does* include ExperimentalApi; notice that the no_path_templates service is untouched in this PR.

PR #839, opened 2016-12-08
Handle construction of oneof from null resource name

michaelbausor:
This case is the only null-handling not handled by the extended get/set methods on the proto types


garrettjonesgoogle:
LGTM

Issue #795, opened 2016-11-28
2nd refactor round to reduce the complexity of discovery


saicheems:
- Remove or reduce the work done by `TypeNameGenerator`
- Consolidate as much of the repetition in the language transformers as possible

tcoffee-google:
Noting here opportunity to refactor variable name formatting: https://github.com/googleapis/toolkit/pull/750#discussion_r88981404

PR #840, opened 2016-12-08
Add support for the "version_module" field in discovery

saicheems:
"version_module" determines whether or not an API's version should be
included in the import path of Java discovery client library classes.
This commit adds support for that field and removes the corresponding
overrides.

### DIFF
--- a/src/main/java/com/google/api/codegen/DiscoveryImporter.java
+++ b/src/main/java/com/google/api/codegen/DiscoveryImporter.java
@@ -141,6 +141,17 @@ public class DiscoveryImporter {
     importer.config.setApiName(serv.getApis(0).getName());
     importer.config.setApiVersion(serv.getApis(0).getVersion());
 
+    boolean versionModule = false;
+    JsonNode versionModuleNode = disco.get("version_module");
+    if (versionModuleNode != null) {
+      // Annoyingly, "version_module" seems to be arbitrary defined either as
+      // the string "True" or an actual boolean.
+      versionModule =
+          disco.get("version_module").asText().equals("True")
+              || disco.get("version_module").asBoolean();
+    }
+    importer.config.setVersionModule(versionModule);
+
     importer.config.getTypes().putAll(importer.types);
     for (Type type : importer.types.values()) {
       for (Field field : type.getFieldsList()) {
--- a/src/main/java/com/google/api/codegen/DiscoveryImporter.java
+++ b/src/main/java/com/google/api/codegen/DiscoveryImporter.java
@@ -144,7 +144,7 @@ public class DiscoveryImporter {
     boolean versionModule = false;
     JsonNode versionModuleNode = disco.get("version_module");
     if (versionModuleNode != null) {
-      // Annoyingly, "version_module" seems to be arbitrary defined either as
+      // Annoyingly, "version_module" seems to be arbitrarily defined either as
       // the string "True" or an actual boolean.
       versionModule =
           disco.get("version_module").asText().equals("True")
--- a/src/main/java/com/google/api/codegen/discovery/config/SampleConfig.java
+++ b/src/main/java/com/google/api/codegen/discovery/config/SampleConfig.java
@@ -80,6 +80,10 @@ public abstract class SampleConfig {
   @JsonProperty("apiVersion")
   public abstract String apiVersion();
 
+  /** Returns true if clients should be qualified with the API version. */
+  @JsonProperty("versionModule")
+  public abstract boolean versionModule();
+
   /**
    * Returns the API's type name.
    *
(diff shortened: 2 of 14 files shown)


saicheems:
Gonna try rolling out these CLs in small chunks and start to shift functionality around bit by bit

saicheems:
#795

src/main/java/com/google/api/codegen/discovery/config/SampleConfig.java
```diff
@@ -80,6 +80,10 @@
   @JsonProperty("apiVersion")
   public abstract String apiVersion();
 
+  /** Returns true if clients should be qualified with the API version. */
```
vchudnov-g: nit: Comment format seems inconsistent with, eg, the one-line comment below.

src/main/java/com/google/api/codegen/discovery/transformer/java/JavaSampleTypeNameConverter.java
```diff
@@ -62,29 +61,26 @@
...
-    }
-    return packagePrefix;
-  }
-
   @Override
   public TypeName getRequestTypeName(String apiTypeName, TypeInfo typeInfo) {
     MessageTypeInfo type = typeInfo.message();
     return new TypeName(
-        Joiner.on('.').join(packagePrefix(type.subpackage()), apiTypeName, type.typeName()),
+        Joiner.on('.').join(packagePrefix, "model", apiTypeName, type.typeName()),
```
vchudnov-g: what does this "model" denote? An abstract model? But aren't we already in a Java-output-specific file?

src/main/java/com/google/api/codegen/discovery/config/SampleConfig.java
saicheems: I'm not quite sure what you mean, but I've changed it to be less specific than the `true` case.

saicheems:
Tests pass

src/main/java/com/google/api/codegen/DiscoveryImporter.java
```diff
@@ -141,6 +141,17 @@ public static DiscoveryImporter parse(Reader reader) throws IOException {
     importer.config.setApiName(serv.getApis(0).getName());
     importer.config.setApiVersion(serv.getApis(0).getVersion());
 
+    boolean versionModule = false;
+    JsonNode versionModuleNode = disco.get("version_module");
+    if (versionModuleNode != null) {
+      // Annoyingly, "version_module" seems to be arbitrary defined either as
```
tcoffee-google: s/arbitrary/arbitrarily/

src/main/java/com/google/api/codegen/discovery/transformer/java/JavaSampleTypeNameConverter.java
```diff
@@ -62,29 +61,26 @@
...
-  /** Returns packagePrefix with subpackage appended if not empty. */
-  private String packagePrefix(String subpackage) {
-    if (!Strings.isNullOrEmpty(subpackage)) {
-      return new StringBuilder(packagePrefix).append('.').append(subpackage).toString();
-    }
-    return packagePrefix;
-  }
-
   @Override
   public TypeName getRequestTypeName(String apiTypeName, TypeInfo typeInfo) {
```
tcoffee-google: Can we de-duplicate the redundant construction here?

saicheems: If I understand correctly you mean calling directly into `getTypeName` from `getRequestTypeName` so we can remove the bit of redundancy?

It's tricky because they aren't constructed the same way...

model types don't have `apiTypeName` as one of their parts, and also follow different rules  because the request types must be partially qualified.

src/main/java/com/google/api/codegen/DiscoveryImporter.java
saicheems: Done

src/main/java/com/google/api/codegen/discovery/transformer/java/JavaSampleTypeNameConverter.java
tcoffee-google: I thought maybe each of them could call a common helper capturing the way the packages are structured, but don't worry about it if it makes things worse.

tcoffee-google:
LGTM

vchudnov-g:
lgtm

src/main/java/com/google/api/codegen/discovery/transformer/java/JavaSampleTypeNameConverter.java
saicheems: I'll come back to this, I have half a plan to clean this class - trying to get the low hanging fruit out of the way first

PR #841, opened 2016-12-09
Fix streaming test

michaelbausor:
Replace hardcoded StreamingRecognizeResponse with type name


garrettjonesgoogle:
Well that was silly. LGTM.

PR #844, opened 2016-12-09
Support using resource name types for formatting strings

michaelbausor:
When a format is specified for a method parameter, but no resource name is specified in the resource name generation config, we can use the resource name to create a formatted string instead of simply using an unformatted string.


garrettjonesgoogle:
LGTM

michaelbausor:
Updated to include:
- Properly support resource names for fields in the response object. This affects test generation, and requires knowing about all fields that have a resource name assigned
- Make sure that a resource name is assigned a fixed proto file, which can be used by each language to determine the package for that resource name

src/main/java/com/google/api/codegen/config/ApiConfig.java
```diff
@@ -70,18 +73,23 @@
   public abstract ImmutableMap<String, ResourceNameConfig> getResourceNameConfigs();
 
   /**
+   * Returns a map from field names to FieldConfigs for all fields that have a resource name type
+   * specified.
+   */
+  public abstract ImmutableMap<String, FieldConfig> getDefaultResourceNameFieldConfigMap();
```
garrettjonesgoogle: Why is this a "default"?

src/main/java/com/google/api/codegen/config/ResourceNameMessageConfigs.java
```diff
@@ -28,20 +36,44 @@
 
   abstract ImmutableMap<String, ResourceNameMessageConfig> getResourceTypeConfigMap();
 
+  public abstract ListMultimap<String, Field> getFieldsWithResourceNamesByMessage();
```
garrettjonesgoogle: I can't tell what exactly the key is in this context - a String isn't === a Message.

src/main/java/com/google/api/codegen/config/ApiConfig.java
michaelbausor: This is default in the sense that it is the field config that should apply outside of the context of a particular method.

I will rephrase/rename to make this clearer.

src/main/java/com/google/api/codegen/config/ResourceNameMessageConfigs.java
michaelbausor: It is the message name - I will rephrase/restructure to make this clearer.

src/main/java/com/google/api/codegen/config/ApiConfig.java
michaelbausor: Added comment

src/main/java/com/google/api/codegen/config/ResourceNameMessageConfigs.java
michaelbausor: Added comment

michaelbausor:
PTAL.

I have created an issue for why BookFromAnywhere is now missing from generated Ruby, NodeJS and Python doc classes (https://github.com/googleapis/toolkit/issues/866)

cc @chrisdunelm for changes to add support for generating C# resource name get/set methods for messages that are defined in proto files that don't contain a service definition

garrettjonesgoogle:
LGTM

PR #845, opened 2016-12-09
Support `resource` -> `resource_` rename

saicheems:
Push the rename of fields named `resource` to the transformer to remove
a good deal of overrides. This is required because the Node.js client
libraries call request bodies `resource` by default, so all fields are
disambiguated with an underscore.


vchudnov-g:
LGTM

src/main/java/com/google/api/codegen/discovery/transformer/nodejs/NodeJSSampleMethodToViewTransformer.java
```diff
@@ -73,9 +73,16 @@ private SampleView createSampleView(SampleTransformerContext context) {
 
     List<SampleFieldView> fields = new ArrayList<>();
     for (FieldInfo field : methodInfo.fields().values()) {
+      String name = field.name();
+      // Since the requestBody is named `resource`, all fields named `resource`
+      // are renamed by the Node.js client library generator to `resource_`.
+      if (name.equals("resource")) {
+        name = "resource_";
```
vchudnov-g: Is this captured in a test?

saicheems: It is captured by baseline tests, but because this PR doesn't actually change the output it doesn't show up in the diff.

vchudnov-g:
LGTM

PR #846, opened 2016-12-09
Support underscore suffix for disambiguating names

### LINKED ISSUE #564: Support custom prefix for identifiers that conflict with reserved identifiers in Discovery
`int` -> `resourceInt` instead of `int2`.

The semantics are up for debate - I would fix this by:

```
public SymbolTable(String conflictPrefix) {
  this.conflictPrefix = conflictPrefix;
}

public String getSymbol(String name) {
  if (RESERVED_IDENTIFIER_SET.contains(name)) {
    return conflictPrefix + name + getSuffix(name);
  }
  ...
}
```

Basically, a prefix for names conflicting with reserved identifiers that is decided when the table is constructed. That could be the last methodNameComponent, or anything really.

saicheems:
Python and Ruby use the `UNDERSCORE` disambiguation strategy to resolve
name collisions in the symbol table.

Resolves #564

### DIFF
--- a/src/main/java/com/google/api/codegen/util/SymbolTable.java
+++ b/src/main/java/com/google/api/codegen/util/SymbolTable.java
@@ -28,8 +28,23 @@ public class SymbolTable {
 
   private final Set<String> symbolTable = new HashSet<>();
 
+  public enum DisambiguationStrategy {
+    UNDERSCORE,
+    NUMERIC
+  }
+
+  private final DisambiguationStrategy disambiguationStrategy;
+
+  public SymbolTable() {
+    this.disambiguationStrategy = DisambiguationStrategy.NUMERIC;
+  }
+
+  public SymbolTable(DisambiguationStrategy disambiguationStrategy) {
+    this.disambiguationStrategy = disambiguationStrategy;
+  }
+
   /**
-   * Returns a new SymbolTable seeded with all the words in seed.
+   * Returns the same SymbolTable seeded with all the words in seed.
    *
    * <p>For example, if seed is {"int"}, a subsequent call to {@link #getNewSymbol(String)} for
    * "int" will return "int2".
@@ -37,8 +52,8 @@ public class SymbolTable {
    * <p>The behavior of the returned SymbolTable is guaranteed if used with {@link
    * #getNewSymbol(String)}, but not with {@link #getNewSymbol(Name)}.
    */
-  public static SymbolTable fromSeed(Set<String> seed) {
-    SymbolTable symbolTable = new SymbolTable();
+  public SymbolTable seed(Set<String> seed) {
+    SymbolTable symbolTable = new SymbolTable(disambiguationStrategy);
     for (String s : seed) {
       symbolTable.getNewSymbol(s);
     }
@@ -46,7 +61,7 @@ public class SymbolTable {
   }
 
   /**
-   * Returns a unique name, with a numeric suffix in case of conflicts.
+   * Returns a unique name, with a suffix in case of conflicts.
    *
    * <p>Not guaranteed to work as expected if used in combination with {@link
    * #getNewSymbol(String)}.
@@ -61,7 +76,7 @@ public class SymbolTable {
   }
 
   /**
-   * Returns a unique name, with a numeric suffix in case of conflicts.
+   * Returns a unique name, with a suffix in case of conflicts.
    *
    * <p>Not guaranteed to work as expected if used in combination with {@link #getNewSymbol(Name)}.
    */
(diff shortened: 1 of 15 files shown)


src/main/java/com/google/api/codegen/util/SymbolTable.java
```diff
@@ -28,25 +28,40 @@
...
+
+  public SymbolTable(DisambiguationStrategy disambiguationStrategy) {
+    this.disambiguationStrategy = disambiguationStrategy;
+  }
+
   /**
-   * Returns a new SymbolTable seeded with all the words in seed.
+   * Returns the same SymbolTable seeded with all the words in seed.
    *
    * <p>For example, if seed is {"int"}, a subsequent call to {@link #getNewSymbol(String)} for
```
tcoffee-google: This comment should be updated.

src/main/java/com/google/api/codegen/util/SymbolTable.java
```diff
@@ -28,25 +28,40 @@
...
+    this.disambiguationStrategy = DisambiguationStrategy.NUMERIC;
+  }
+
+  public SymbolTable(DisambiguationStrategy disambiguationStrategy) {
+    this.disambiguationStrategy = disambiguationStrategy;
+  }
+
   /**
-   * Returns a new SymbolTable seeded with all the words in seed.
+   * Returns the same SymbolTable seeded with all the words in seed.
```
vchudnov-g: How about s/Returns the same/Updates and returns/ ?

src/main/java/com/google/api/codegen/util/SymbolTable.java
```diff
@@ -70,17 +85,27 @@ public String getNewSymbol(String desiredName) {
...
-   *
-   * <p>Stores the joined desiredName/suffix in an internal map. For example, if "foo" is passed, ""
-   * is returned. If "foo" is passed again, "2" is returned, and then "3" and so on.
-   */
   private String getAndSaveSuffix(String desiredName) {
     if (!symbolTable.contains(desiredName)) {
       symbolTable.add(desiredName);
       return "";
     }
+    if (disambiguationStrategy == DisambiguationStrategy.NUMERIC) {
```
vchudnov-g: Consider a switch statement for clarity, especially if we grow the number of strategies.

src/main/java/com/google/api/codegen/util/SymbolTable.java
```diff
@@ -89,4 +114,24 @@ private String getAndSaveSuffix(String desiredName) {
...
     return Integer.toString(i);
   }
+
+  /**
+   * Returns the next underscore suffix that makes desiredName unique.
+   *
+   * <p>Stores the joined desiredName/suffix in an internal map. Assumes that desiredName is already
+   * in the symbolTable.
+   *
+   * <p>For example, if "foo" is passed, "_" is returned. If "foo" is passed again, "__" is
```
vchudnov-g: This is acceptable in some languages? Having more than one trailing underscore seems so much more unreadable than just using numbers a la `foo3` or `foo_3`.

src/main/java/com/google/api/codegen/util/SymbolTable.java
```diff
@@ -70,17 +85,27 @@ public String getNewSymbol(String desiredName) {
...
       return "";
     }
+    if (disambiguationStrategy == DisambiguationStrategy.NUMERIC) {
+      return getAndSaveNumericSuffix(desiredName);
+    }
+    return getAndSaveUnderscoreSuffix(desiredName);
+  }
+
+  /**
+   * Returns the next numeric suffix that makes desiredName unique.
```
vchudnov-g: test for this method and the next?

src/main/java/com/google/api/codegen/util/SymbolTable.java
saicheems: Good point, I'll adjust it...

Incidentally, do you think it's useful to support underscores at all? Having written this PR, I'm not sure if underscores are necessarily more idiomatic in Ruby/Node.js

Is there a better strategy besides numbers/underscores you can think of off the top of your head?

saicheems:
Going to come back to this

src/main/java/com/google/api/codegen/util/SymbolTable.java
tcoffee-google: As in my comments https://github.com/googleapis/toolkit/issues/564#issuecomment-257925186, a current negative outcome of the numeric scheme is that in some instances, we get variable `foo` used to iterate over elements of `foo2`, which is quite confusing. Underscores are not great in this instance either, but at least they don't contradict the actual relationship so blatantly. Perhaps this case would be better resolved by handling collisions with iterators explicitly, so that `foo` can be used to iterate over `foo_page` or whatever a more appropriate name might be for the collection.

Issue #834, opened 2016-12-08
Optional parameters with resource names can't be empty


jskeet:
Example: `LoggingClient.WriteLogEntries` takes a `LogNameOneof`.

The parameter is optional in the API, but is in the generated code as a method parameter.

How do we "omit" it?

- If we pass in null, we'll get a NullReferenceException
- We can't create a resource name which is empty

I suspect we want to:

- Have an `EmptyResourceName` singleton
- Allow that as another "kind" in each oneof
- Parse "" to `EmptyResourceName.Instance`

// cc @michaelbausor who may run into the same thing for Java

chrisdunelm:
Could we allow `null` in this case?

jskeet:
I was wondering about that. Maybe. But we don't for string values... so if this is meant to be like the string property, it would be odd. We should probably only do *one* of those - I don't think we should have two ways of representing "an empty resource name".

garrettjonesgoogle:
@michaelbausor and I were just discussing this yesterday! In my understanding, in proto3, native types can't be null, but objects (messages) can be. It's true that the underlying type is String (null), but the resource type is an object. I would probably go for making it null, and just translate between empty string and null resource type; I think that's a little simpler than an empty resource name class.

michaelbausor:
Another reason I think to prefer null is to handle this when the resource name is not a oneof. So for example looking at [Group.parent_name in monitoring](https://github.com/googleapis/googleapis/blob/master/google/monitoring/v3/group.proto#L64), I think it is preferable to have group.getParentNameAsGroupName() return null, rather than have group.getParentNameAsGroupNameOneof() return a GroupNameOneof that supports a single valid format (GroupName) and the EmptyResourceName.

chrisdunelm:
@michaelbausor, @garrettjonesgoogle C# implements this using `null`. If Java has also resolved this issue, then it can be closed.

jskeet:
@garrettjonesgoogle Ping for closing?

garrettjonesgoogle:
This is done in Java.

PR #854, opened 2016-12-13
C# bug fixes

### LINKED ISSUE #852: "Color Color" issue with C# codegen
The C# codegen currently emits this in the partial proto for DeleteEventsRequest (and others):

```csharp
public ProjectName ProjectNameAsProjectName
{
    get { return ProjectName.Parse(ProjectName); }
    set { ProjectName = GaxPreconditions.CheckNotNull(value, nameof(value)).ToString(); }
}
```

This fails as `ProjectName.Parse` binds `ProjectName` to the property, not the type.

The solution is just to fully-qualify the name - it's easiest to do this completely, e.g.

```csharp
public ProjectName ProjectNameAsProjectName
{
    get { return Google.Cloud.ErrorReporting.V1Beta1.ProjectName.Parse(ProjectName); }
    set { ProjectName = GaxPreconditions.CheckNotNull(value, nameof(value)).ToString(); }
}
```

chrisdunelm:
Fixes #852 - Now uses faully qualified name to call Parse() method

Updates #834 - Allows ResourceName properties to take and return null

### DIFF
--- a/src/main/resources/com/google/api/codegen/csharp/gapic_resourcenames.snip
+++ b/src/main/resources/com/google/api/codegen/csharp/gapic_resourcenames.snip
@@ -426,15 +426,15 @@
                     public {@field.typeName} {@field.propertyName}
                     {
                         @if field.isAny
-                            get { return UnknownResourceName.Parse({@field.underlyingPropertyName}); }
+                            get { return string.IsNullOrEmpty({@field.underlyingPropertyName}) ? null : UnknownResourceName.Parse({@field.underlyingPropertyName}); }
                         @else
                             @if field.isOneof
-                                get { return {@field.typeName}.Parse({@field.underlyingPropertyName}, true); }
+                                get { return string.IsNullOrEmpty({@field.underlyingPropertyName}) ? null : {@field.parseMethodTypeName}.Parse({@field.underlyingPropertyName}, true); }
                             @else
-                                get { return {@field.typeName}.Parse({@field.underlyingPropertyName}); }
+                                get { return string.IsNullOrEmpty({@field.underlyingPropertyName}) ? null : {@field.parseMethodTypeName}.Parse({@field.underlyingPropertyName}); }
                             @end
                         @end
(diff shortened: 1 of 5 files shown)


src/main/resources/com/google/api/codegen/csharp/gapic_resourcenames.snip
```diff
@@ -409,6 +409,11 @@
             @join field : proto.fields
                 /// <summary>
                 /// A <see cref="{@field.docTypeName}"/>-typed view over the <see cref="{@field.underlyingPropertyName}"/> resource name property.
+                @if field.isOptional
```
jskeet: Hmm. I'm not sure I'd do it in the proto itself. That suggests that it's *always* either optional or required - whereas in reality it may be required in some cases, and optional in others.

I'd make the proto always accept null, setting the value to "". That's simple and dumb.

Then the *method* calls can validate that required parameters are non-null - assuming it already does so for required parameters of other types. (If it doesn't, let's not *just* do it for that.)

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
```diff
@@ -689,11 +689,12 @@ namespace Google.Example.Library.V1
     {
         /// <summary>
         /// A <see cref="BookName"/>-typed view over the <see cref="Name"/> resource name property.
+        /// May be set to <c>null</c>.
         /// </summary>
         public BookName BookName
         {
-            get { return BookName.Parse(Name); }
-            set { Name = GaxPreconditions.CheckNotNull(value, nameof(value)).ToString(); }
+            get { return string.IsNullOrEmpty(Name) ? null : Google.Example.Library.V1.BookName.Parse(Name); }
```
jskeet: I'm wondering whether Parse should just return null if we pass it "". It would simplify all the code calling it, but might feel a little odd. Let's think about that.

chrisdunelm: As dicussed offline, it doesn't seem right for Parse to handle an empty string by returning null.

chrisdunelm:
PTAL

PR #855, opened 2016-12-13
Node: remove use of .bind

landrito:
Resolves [gax-nodejs#78](https://github.com/googleapis/gax-nodejs/issues/78)


src/main/resources/com/google/api/codegen/nodejs/main.snip
```diff
@@ -309,14 +309,18 @@
               '{@method}'
           @end
         ];
+        var self = this;
```
jmuk: It's nicer to make this assignment happen only once even when mixin exists.

landrito:
Thanks for the review!

PR #858, opened 2016-12-14
make Go handle LRO returning empty properly

pongad:
A new method GetBigNothing is added to library.proto and corresponding
config to test this behavior.

Package naming also changed slightly:
Previously, if a proto file declare a go_package option with
package name, the name is used in imports.
If the proto file does not declare the name,
we infer the name from the package declaration and add "pb" to it.

Now, all proto files in googleapis has been updated with go_package,
following above rules will drop "pb" from the import name.
This is still correct but causes a large amount of useless diff.
The fix is to add the "pb" back.


pongad:
I apologize for the large diff in baseline files.

@jba please jump to the Go baselines.

garrettjonesgoogle:
generator code LGTM

pongad:
I just realized how unclear it is. Thank you for pointing out.

PR #860, opened 2016-12-14
Add ADC link + sync sample headers

### LINKED ISSUE #634: Add ADC link to end of item 2 in discovery sample instructions
```
// 2. This sample uses Application Default Credentials for authentication.
//    If not already done, install the gcloud CLI from
//    https://cloud.google.com/sdk/ and run
//    `gcloud beta auth application-default login`
//    For more information, see:
//    https://developers.google.com/identity/protocols/application-default-credentials
```

saicheems:
With this change all sample headers are in sync.

Resolves #634

### DIFF
--- a/src/main/resources/com/google/api/codegen/go/sample.snip
+++ b/src/main/resources/com/google/api/codegen/go/sample.snip
@@ -14,6 +14,8 @@
     //    If not already done, install the gcloud CLI from
     //    https://cloud.google.com/sdk/ and run
     //    `gcloud beta auth application-default login`
+    //    For more information, see:
+    //    https://developers.google.com/identity/protocols/application-default-credentials
     // 3. Install and update the Go dependencies by running `go get -u` in the
     //    project directory.
   @default
--- a/src/main/resources/com/google/api/codegen/go/sample.snip
+++ b/src/main/resources/com/google/api/codegen/go/sample.snip
@@ -14,7 +14,7 @@
     //    If not already done, install the gcloud CLI from
     //    https://cloud.google.com/sdk/ and run
     //    `gcloud beta auth application-default login`.
-    //    For more information, see:
+    //    For more information, see
     //    https://developers.google.com/identity/protocols/application-default-credentials
     // 3. Install and update the Go dependencies by running `go get -u` in the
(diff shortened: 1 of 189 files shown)


src/main/resources/com/google/api/codegen/go/sample.snip
```diff
@@ -14,6 +14,8 @@
     //    If not already done, install the gcloud CLI from
     //    https://cloud.google.com/sdk/ and run
     //    `gcloud beta auth application-default login`
+    //    For more information, see:
```
vchudnov-g: For readability, I suggest indenting the two lines above this one:
```
If not already done,...
  https:...
  `gcloud ...`
For more information...
```

Additionally, I think it reads better to s/:// on this line.

tcoffee-google: I would prefer adding a period after the second backtick so that we just have sentences with line breaks. Hanging indent doesn't seem right to me for this case.

saicheems: I'm also not much a fan of the hanging indents.
Added the period in the meanwhile, PTAL

vchudnov-g: The reason I suggested it is that there isn't enough visual separation between the two sentences. My preferred solution would be to have a blank line, but that would require more blank lines elsewhere and I was trying to conserve vspace. :-(

I'm OK with keeping as is.

But I still vote for s/:// here.

tcoffee-google: I think we're currently consistent about using colons in forms like 
```
... see:
<URL>
```
so if we decide to change it here, it might be preferable to change elsewhere. I personally prefer the colon, but leaving it out seems okay.

tcoffee-google: It's a fair point --- I think it would be fine to remove all the colons. We use them, for instance, to refer to external URLs for media upload/download and auth instructions.

It would be nice if we could get API explorer to hyperlink URLs appearing in our samples.

saicheems:
ptal

PR #861, opened 2016-12-14
Make scopes comment formatting consistent

### LINKED ISSUE #704: Use 2 space indents in scope lists in Java and Go discovery samples

saicheems:
Resolves #704

### DIFF
--- a/src/test/java/com/google/api/codegen/testdata/discoveries/go/go_adexchangebuyer.v1.4.json.baseline
+++ b/src/test/java/com/google/api/codegen/testdata/discoveries/go/go_adexchangebuyer.v1.4.json.baseline
@@ -49,7 +49,7 @@ func getClient(ctx context.Context) (*http.Client, error) {
   // https://foo.com/bar
   //
   // Authorize using the following scope:
-  //    adexchangebuyer.AdexchangeBuyerScope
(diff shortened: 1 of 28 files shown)


src/test/java/com/google/api/codegen/testdata/discoveries/go/go_adexchangebuyer.v1.4.json.baseline
```diff
@@ -49,7 +49,7 @@ func getClient(ctx context.Context) (*http.Client, error) {
   // https://foo.com/bar
   //
   // Authorize using the following scope:
-  //    adexchangebuyer.AdexchangeBuyerScope
+  //     adexchangebuyer.AdexchangeBuyerScope
```
vchudnov-g: Here the scope is a Go identifier, whereas in Java it is an http URI. I assume the value of the Go identifier is the corresponding URI.

But this makes me wonder whether for consistency we should quote the URIs in the comments in the other languages.

Thoughts?

saicheems: Done, please double-check I didn't miss anything dumb

PR #867, opened 2016-12-15
Auto-generate index.js for NodeJS

shinfan:
- Auto-generate support for src/index.js
- Fixed an issue in paged method tests
- Add `lodash.union` dependency to package.json if the service contains more than one interface
- Fixed service var name

Tests:
- Baselines
- Manual tests passed for cloudtrace and errorreporting 

cc/ @bjwatson

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSPackageMetadataTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSPackageMetadataTransformer.java
@@ -28,7 +28,6 @@ import java.util.List;
 /** Responsible for producing package metadata related views for NodeJS */
 public class NodeJSPackageMetadataTransformer implements ModelToViewTransformer {
   private static final String PACKAGE_FILE = "nodejs/package.snip";
-  private static final String INDEX_FILE = "nodejs/index.snip";
 
   // TODO: Retrieve the following values from static file
   // Github issue: https://github.com/googleapis/toolkit/issues/848
@@ -41,7 +40,6 @@ public class NodeJSPackageMetadataTransformer implements ModelToViewTransformer 
   public List<String> getTemplateFileNames() {
     List<String> fileNames = new ArrayList<>();
     fileNames.add(PACKAGE_FILE);
-    fileNames.add(INDEX_FILE);
     return fileNames;
   }
 
@@ -50,10 +48,10 @@ public class NodeJSPackageMetadataTransformer implements ModelToViewTransformer 
     Iterable<Interface> services = new InterfaceView().getElementIterable(model);
     boolean hasMultipleServices = Iterables.size(services) > 1;
     List<ViewModel> models = new ArrayList<ViewModel>();
-    NodeJSPackageMetadataNamer metadataNamer =
+    NodeJSPackageMetadataNamer namer =
         new NodeJSPackageMetadataNamer(
             apiConfig.getPackageName(), apiConfig.getDomainLayerLocation());
-    models.add(generateMetadataView(metadataNamer, hasMultipleServices));
+    models.add(generateMetadataView(namer, hasMultipleServices));
     return models;
   }
 
--- a/src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSPackageMetadataTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSPackageMetadataTransformer.java
@@ -23,6 +23,7 @@ import com.google.api.tools.framework.model.Interface;
 import com.google.api.tools.framework.model.Model;
 import com.google.common.collect.Iterables;
 import java.util.ArrayList;
+import java.util.Arrays;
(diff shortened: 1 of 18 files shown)


src/main/resources/com/google/api/codegen/nodejs/index.snip
```diff
@@ -0,0 +1,43 @@
...
+
+  {@index.apiVersion}.SERVICE_ADDRESS = {@index.primaryService.clientName}.SERVICE_ADDRESS;
+  @if index.hasMultipleServices
+    {@index.apiVersion}.ALL_SCOPES = union(
+      @join require : index.requireViews on ", ".add(BREAK)
+        {@require.clientName}.ALL_SCOPES
+      @end
+    );
+  @else
+    {@index.apiVersion}.ALL_SCOPES = {@index.primaryService.clientName}.ALL_SCOPES;
```
jmuk: You don't have to do that in this PR, but I'm wondering if it's possible to define the list of all scopes here and remove `ALL_SCOPES` in each service .js file.

src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSPackageMetadataTransformer.java
```diff
@@ -14,18 +14,23 @@
...
 import com.google.api.tools.framework.model.Model;
+import com.google.common.collect.Iterables;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 
 /** Responsible for producing package metadata related views for NodeJS */
 public class NodeJSPackageMetadataTransformer implements ModelToViewTransformer {
   private static final String PACKAGE_FILE = "nodejs/package.snip";
+  private static final String INDEX_FILE = "nodejs/index.snip";
```
jmuk: For historical reason the `index.js` file is generated by packman, but conceptually it is not a package metadata, but just a part of autogen files, and actually it always needs to be generated along with service. Can you make that as a part of gapic codegen?  (if it's hard due to some reasons, like nodejs gapic is not yet MVVM -- that's fine. Let's do that later).

shinfan: Oh okay I see. I thought everything generated by packman was part of the metadata. I can make it a part of gapic codegen.

src/main/resources/com/google/api/codegen/nodejs/index.snip
shinfan: This change would be out of the scope of this PR so I prefer to do it in a separate one. But I suspect it would have a negative impact on the readability of the code. Imagine we have multiple interfaces and we group all scopes in this index file, the scope list could be very long and therefore not that readable.

src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSPackageMetadataTransformer.java
shinfan: Done

shinfan:
@jmuk  PTAL

src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSPackageMetadataTransformer.java
```diff
@@ -36,20 +38,25 @@
 
   @Override
   public List<String> getTemplateFileNames() {
-    return Collections.singletonList(PACKAGE_FILE);
+    List<String> fileNames = new ArrayList<>();
+    fileNames.add(PACKAGE_FILE);
+    return fileNames;
```
jmuk: I prefer `Arrays.asList` (https://docs.oracle.com/javase/7/docs/api/java/util/Arrays.html#asList(T...))

src/main/resources/com/google/api/codegen/nodejs/index.snip
```diff
@@ -0,0 +1,43 @@
...
+  function {@index.apiVersion}(options) {
+    options = extend({
+      scopes: {@index.apiVersion}.ALL_SCOPES
+    }, options);
+    var gaxGrpc = gax.grpc(options);
+    var result = {};
+    @join require : index.requireViews
+      extend(result, {@require.clientName}(gaxGrpc));
+    @end
+    return result;
```
landrito: It would be more idiomatic here to return the client directly if there is only one.

ie:
```
@switch index.requireViews.size.toString
@case "1"
  return {@require.clientName}(gaxGrpc);
@default
  var result = {};
  @join require : index.requireViews
    extend(result, {@require.clientName}(gaxGrpc));
  @end
  return result;
@end
```

shinfan: Done

src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSPackageMetadataTransformer.java
shinfan: Done

Issue #780, opened 2016-11-23
Feature request: MetadataGen config input


geigerj:
Original discussion: https://github.com/googleapis/toolkit/pull/778#issuecomment-262612735

From @garrettjonesgoogle:

> I wonder if it would be more maintainable to have the tool understand the same config that artman does. As it is now, for every new config line we put in artman config, we have to add a new command line option to the tool. I doubt anyone is going to handcraft the full command line for this tool.

I'm not opposed to having MetadataGen read YAML as input (as it already does for the existing packman config files). But what does this mean for how the artman config is consumed? Is it consumed once in whole by `artman`, where it is merged from several YAML files, and once additionally in part by `MetadataGen`, where just the `artman_{API}.yaml` file is read in? That seems to complicate the logic somewhat, and it means that MetadataGen doesn't benefit from `artman`'s built-in config merging and per-language section selection logic.

garrettjonesgoogle:
One option is to have artman perform the merging logic and create a temporary merged yaml file to pass to `MetadataGen`.

geigerj:
Having artman create a temporary merged yaml SGTM.

PR #870, opened 2016-12-19
Port GAPIC metadata generation for Python

### LINKED ISSUE #848: Load metadata info from static file.
Reuse the code in PackageMetadataGen to read in https://github.com/googleapis/googleapis/blob/master/gapic/packaging/api_defaults.yaml.

geigerj:
Move the functionality of generating Python GAPIC package
metadata from Packman to GAPIC. To do so, introduce a new package
metadata configuration format, which lives at the same level as the
GAPIC config in terms of input/test input. This new config file
fixes #848.

Sorry, this PR is big, but a lot of the LOC changed are just test data.

The following follow-up PRs are planned:
- A PR to Artman, introducing a task to generate the
  new package metadata config from the artman config, and removing the
  Python Packman task.
- A PR to Toolkit, replacing all inputs to GrpcMetadataGen with the
  new package metadata config, which addresses #780
- (Potentially) A PR to Toolkit investigating the consolidation of
  gRPC and GAPIC Python package metadata snippet templates. Note that
  Packman currently uses separate templates for gRPC and GAPIC.

cc: @bjwatson, @shinfan


src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
```diff
@@ -0,0 +1,203 @@
...
+import java.nio.file.Path;
+import java.util.Map;
+import javax.annotation.Nullable;
+import org.yaml.snakeyaml.Yaml;
+
+@AutoValue
+public abstract class PackageMetadataConfig {
+
+  @AutoValue
+  public abstract static class VersionBound {
```
garrettjonesgoogle: It seems like this concept could be usable outside of PackageMetadataConfig - maybe lift up to its own file?

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
```diff
@@ -0,0 +1,203 @@
...
+    public abstract String lower();
+
+    @Nullable
+    public abstract String upper();
+  }
+
+  protected abstract Map<String, VersionBound> gaxVersion();
```
garrettjonesgoogle: The name here (`gaxVersion`) implies a single version; maybe `gaxVersionBound` would be clearer. (Applies to the attributes below too)

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
```diff
@@ -0,0 +1,203 @@
...
+        .protoPath("")
+        .author("")
+        .email("")
+        .homepage("")
+        .license("")
+        .build();
+  }
+
+  @SuppressWarnings("unchecked")
+  public static PackageMetadataConfig createFromFile(Path packageConfigPath) throws IOException {
```
garrettjonesgoogle: This seems like transformer logic

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
```diff
@@ -0,0 +1,203 @@
...
+  protected abstract Map<String, VersionBound> protoVersion();
+
+  protected abstract Map<String, VersionBound> commonProtosVersion();
+
+  protected abstract Map<String, VersionBound> packageVersion();
+
+  protected abstract Map<String, String> packageName();
+
+  /** The version of GAX that this package depends on. Configured per language. */
+  public VersionBound gaxVersion(String language) {
```
garrettjonesgoogle: I'd prefer not to pass around language identifiers as plain strings - lots of room for odd variations & mistakes, and it is an "almost" closed set.

src/main/java/com/google/api/codegen/transformer/py/PythonPackageMetadataTransformer.java
```diff
@@ -0,0 +1,182 @@
...
+    List<ViewModel> topLevelMetadata =
+        Lists.transform(
+            getTopLevelTemplateFileNames(),
+            new Function<String, ViewModel>() {
+              @Override
+              @Nullable
+              public ViewModel apply(@Nullable String templateFileName) {
+                return generateMetadataView(model, apiConfig, templateFileName);
+              }
+            });
```
garrettjonesgoogle: This would be much more concise with a simple for loop.

src/main/java/com/google/api/codegen/transformer/py/PythonPackageMetadataTransformer.java
```diff
@@ -0,0 +1,182 @@
...
+    boolean first = true;
+    for (String pkg : Splitter.on(".").split(packageName)) {
+      if (!first) {
+        current.append("." + pkg);
+      } else {
+        current.append(pkg);
+        first = false;
+      }
+      packages.add(current.toString());
+    }
```
garrettjonesgoogle: At first read, it looks like you're splitting by "." and then joining by "."... what is this supposed to do?

src/main/java/com/google/api/codegen/transformer/py/PythonPackageMetadataTransformer.java
```diff
@@ -0,0 +1,182 @@
...
+  private List<String> computeNamespacePackages(String packageName, final String apiVersion) {
+    return Lists.newArrayList(
+        Iterables.filter(
+            computePackages(packageName),
+            new Predicate<String>() {
+              @Override
+              public boolean apply(@Nullable String input) {
+                return isNamespacePackage(input, apiVersion);
+              }
+            }));
```
garrettjonesgoogle: This would be simpler as a for loop.

src/main/java/com/google/api/codegen/transformer/py/PythonPackageMetadataTransformer.java
```diff
@@ -0,0 +1,182 @@
...
+          @Override
+          @Nullable
+          public ViewModel apply(@Nullable final String packageName) {
+            final String template;
+            if (isNamespacePackage(packageName, apiVersion)) {
+              template = "py/namespace__init__.py.snip";
+            } else {
+              template = "py/__init__.py.snip";
+            }
+            return new ViewModel() {
```
garrettjonesgoogle: Just make a simple ViewModel class with AutoBuilder that can be used here. It feels kind of odd to construct an anonymously-typed object for every instance.

src/main/java/com/google/api/codegen/viewmodel/PackageMetadataView.java
```diff
@@ -33,15 +36,36 @@ public String resourceRoot() {
 
   public abstract String identifier();
 
-  public abstract String version();
+  public abstract VersionBound packageVersion();
```
garrettjonesgoogle: Similar to my comment within PackageMetadataConfig: +`Bound`

src/main/resources/com/google/api/codegen/nodejs/package.snip
```diff
@@ -1,23 +1,23 @@
...
-      "google-gax": "{@metadata.gaxVersion}",
+      "google-proto-files": "{@metadata.protoVersion.lower}",
+      "google-gax": "{@metadata.gaxVersion.lower}",
       "extend": "^3.0.0"
     },
     "devDependencies": {
       "mocha": "3.2.0"
     },
-    "license": "Apache-2.0",
+    "license": "{@metadata.license}",
```
garrettjonesgoogle: The name `license` implies the content of the license; maybe this should be `licenseName`.

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
```diff
@@ -0,0 +1,203 @@
...
+  public abstract static class VersionBound {
+    public abstract String lower();
+
+    @Nullable
+    public abstract String upper();
+  }
+
+  protected abstract Map<String, VersionBound> gaxVersion();
+
+  protected abstract Map<String, VersionBound> protoVersion();
```
shinfan: What is the key of this map?

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
shinfan: We usually put such logic into a separate class and this package config class only a container class.

src/main/java/com/google/api/codegen/gapic/GapicProviderFactory.java
```diff
@@ -15,12 +15,16 @@
...
 
 /** A factory for GapicProviders which perform code generation. */
 public interface GapicProviderFactory<ProviderT> {
   /** Create the provider from the given model, apiConfig, and generatorConfig. */
   public List<ProviderT> create(
-      Model model, ApiConfig apiConfig, GapicGeneratorConfig generatorConfig);
+      Model model,
+      ApiConfig apiConfig,
+      GapicGeneratorConfig generatorConfig,
+      PackageMetadataConfig packageConfig);
```
shinfan: Does it make sense to put PackageMetadataConfig under GapicGeneratorConfig? I am a bit cautious of adding more parameters to this interface and conceptually package metadata is a part of GAPIC gen now.

src/main/java/com/google/api/codegen/transformer/py/PythonPackageMetadataTransformer.java
```diff
@@ -0,0 +1,182 @@
...
+import com.google.common.base.Splitter;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
+import java.io.File;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+import javax.annotation.Nullable;
+
+public class PythonPackageMetadataTransformer implements ModelToViewTransformer {
```
shinfan: nit: add comments

src/main/java/com/google/api/codegen/transformer/py/PythonPackageMetadataTransformer.java
```diff
@@ -0,0 +1,182 @@
...
+        .homepage(packageConfig.homepage())
+        .license(packageConfig.license())
+        .fullName(model.getServiceConfig().getTitle())
+        .serviceName(apiConfig.getPackageName())
+        .namespacePackages(
+            computeNamespacePackages(apiConfig.getPackageName(), packageConfig.apiVersion()))
+        .build();
+  }
+
+  private List<String> computePackages(String packageName) {
```
shinfan: Add a comment of the computing logic here and `computeInitFiles`

src/main/resources/com/google/api/codegen/common.snip
```diff
@@ -33,3 +33,9 @@
 @snippet TODO()
   {@context.getTODO}
 @end
+
+@snippet quotedList(list)
+    @join element : list on ", "
+        '{@element}'
+    @end
+@end
```
shinfan: nit: new line

src/test/java/com/google/api/codegen/testdata/library_pkg.yaml
```diff
@@ -0,0 +1,48 @@
...
+major_version: v1
+proto_path: google/library
+author: Google, Inc.
+email: googleapis-packages@google.com
+homepage: https://github.com/googleapis/googleapis
+license: Apache-2.0
+
+# Language-dependent configuration
+package_name:
+  default: google-cloud-library-v1
```
shinfan: Add a language specific package name here for better coverage

src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSPackageMetadataTransformer.java
```diff
@@ -45,19 +45,29 @@
...
-        .gaxVersion(GAX_VERSION)
-        .protoVersion(PROTO_VERSION)
-        .url(PACKAGE_URL)
+        .packageVersion(packageConfig.packageVersion("nodejs"))
+        .protoPath(packageConfig.protoPath())
+        .shortName(packageConfig.shortName())
+        .gaxVersion(packageConfig.gaxVersion("nodejs"))
+        .protoVersion(packageConfig.protoVersion("nodejs"))
+        .commonProtosVersion(packageConfig.commonProtosVersion("nodejs"))
+        .packageName(packageConfig.packageName("nodejs"))
```
geigerj: What would you suggest here? Is it sufficient that this to be changed to an enum, or are you specifically concerned with the repetition? (I think it's probably overkill to create a per-language "view" of the config object.)

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
geigerj: Changed to `TargetLanguage` enum

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
geigerj: Created an enum for this

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
geigerj: Not sure I understand -- are you talking about the `*PackageMetadataTransformer` classes? Why would they be concerned with file I/O?

src/main/java/com/google/api/codegen/transformer/py/PythonPackageMetadataTransformer.java
geigerj: A Python package name `foo.bar.baz` implicitly defines three packages: `foo`, `foo.bar`, and `foo.bar.baz`. Is there a more concise way to express this in Java?

src/main/resources/com/google/api/codegen/common.snip
geigerj: Done

src/main/resources/com/google/api/codegen/nodejs/package.snip
geigerj: Done

src/test/java/com/google/api/codegen/testdata/library_pkg.yaml
geigerj: Respectfully disagree: there's coverage for language-specific configuration elsewhere in this file. This tests the logic for handling the "default" case.

shinfan: Oh..... you are right.  Default case here should be better.

src/main/java/com/google/api/codegen/transformer/py/PythonPackageMetadataTransformer.java
geigerj: Fewer Java lines, maybe, but simpler? This situation is what a filter is *for*...

(but anyhow, done)

src/main/java/com/google/api/codegen/transformer/py/PythonPackageMetadataTransformer.java
geigerj: Done

src/main/java/com/google/api/codegen/transformer/py/PythonPackageMetadataTransformer.java
geigerj: Done.

src/main/java/com/google/api/codegen/viewmodel/PackageMetadataView.java
geigerj: Done

geigerj:
@garrettjonesgoogle @shinfan PTAL -- I addressed most of the comments. Would appreciate more guidance the MVVM stuff regarding resource files vs. templates and whether transformers should be doing I/O.

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
shinfan: I don't think transformers should handle I/O. It is probably okay to keep it here then.

garrettjonesgoogle: I think the actual place where file I/O should be handled is at the top level; transformers accept an in-memory data model as input.

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
```diff
@@ -0,0 +1,205 @@
...
+        .homepage("")
+        .licenseName("")
+        .build();
+  }
+
+  @SuppressWarnings("unchecked")
+  public static PackageMetadataConfig createFromFile(Path packageConfigPath) throws IOException {
+    Yaml yaml = new Yaml();
+    String contents = new String(Files.readAllBytes(packageConfigPath), StandardCharsets.UTF_8);
+    Map<String, Object> configMap = (Map<String, Object>) yaml.load(contents);
```
garrettjonesgoogle: All of our other yaml uses proto as a schema - is it worth converting this to use the same thing?

src/main/java/com/google/api/codegen/transformer/py/PythonPackageMetadataTransformer.java
```diff
@@ -0,0 +1,161 @@
...
+        .author(packageConfig.author())
+        .email(packageConfig.email())
+        .homepage(packageConfig.homepage())
+        .licenseName(packageConfig.licenseName())
+        .fullName(model.getServiceConfig().getTitle())
+        .serviceName(apiConfig.getPackageName())
+        .namespacePackages(
+            computeNamespacePackages(apiConfig.getPackageName(), packageConfig.apiVersion()))
+        .hasMultipleServices(false)
+        .build();
```
garrettjonesgoogle: How much of this is the same between Python & Node? (It's hard to eyeball it)

src/main/java/com/google/api/codegen/transformer/py/PythonPackageMetadataTransformer.java
garrettjonesgoogle: My fault for not reading the comment. I didn't realize the accumulative aspect of the algorithm. 

Here is an alternate approach (not checked for correctness - beware off-by-one errors):

```
List<String> parts = Splitter.on(".").split(packageName);
for (int i = 1; i <= parts.size(); i++) {
  packages.add(Joiner.on(".").join(parts.subList(0, i)));
}
```

src/main/resources/com/google/api/codegen/LICENSE.snip
```diff
@@ -0,0 +1,204 @@
+@snippet generate(context)
+    {@""}                            Apache License
```
garrettjonesgoogle: Hmm if this is accessible from more than one language, then maybe it's fine.

src/main/java/com/google/api/codegen/transformer/py/PythonPackageMetadataTransformer.java
garrettjonesgoogle: Filters should look much nicer in Java 8 when it rolls around... but in Java 7, they are pretty ugly.

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
geigerj: I'm not sure see the benefit, other than for consistency. Creating a proto schema increases the number of places you need to touch if you want to change the schema, and it forces everyone to rebuild the generated Java classes. Given that the schema here is likely to be quite simple, do you think this is worthwhile?

src/main/java/com/google/api/codegen/gapic/GapicProviderFactory.java
shinfan: Fair enough. Could you update the comment o the `ApiConfig` class to point out that it represents the contents of {api}_gapic.yaml. Thanks.

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
garrettjonesgoogle: I see the value as having nice type safety over a configuration model. If you have no schema, your validation is at run time / test time. It might not be worthwhile here if it's simple enough, though.

src/main/java/com/google/api/codegen/gapic/GapicProviderFactory.java
geigerj: Done.

src/main/java/com/google/api/codegen/transformer/py/PythonPackageMetadataTransformer.java
garrettjonesgoogle: If the differences are that small, you could make a common transformer and set slightly more values than any particular language needs.

geigerj: That's true, but I think it's more readable to keep it this way than to have a class that incorporates a `NodeJSPackageMetadataNamer`, some bits of logic about Python namespace packages, and presumably soon other small bits of other language-specific logic. WDYT?

garrettjonesgoogle: This code is kind of a long stretch to copy for every language...

I don't think you need a metadata namer. You could just pass in TargetLanguage as an argument.

Here are some ideas for strategies for setting language-specific bits:

1. Have the transformer take an empty Builder for PackageMetadataView instead of creating its own fully-constructed instance; the caller can then fill in its own bits here & there and call build() itself.
2. Pass in the extra language-specific things as arguments. (This unfortunately is likely to grow)
3. Create language-specific specialty views: PythonExtraMetadata (holds namespacePackages), NodeExtraMetadata (holds hasMultipleServices, identifier)

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
geigerj: Moved I/O out of this method. Do you still think this should be moved to the transformer? Currently, this is the standin for the (hidden) builder, since there's no one that currently needs to build this class explicitly.

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
garrettjonesgoogle: Here is the difference:

direct yaml loading:

1. Validation of config file: validated at runtime when build() runs
2. Validation of code accessing data from the config file: validated at runtime when build() runs

Note - you can't distinguish between 1 & 2. 

proto-based loading:

1. Validation of config file: validated at runtime in MultiYamlReader
2. Validation of code accessing data from the config file: validated at compile time against Java classes generated by the schema

Like I said, this is simple enough that I won't require a proto-based schema for it. I just wanted to make sure we agree on the tradeoffs.

geigerj: Yup, agreed. I see what you're saying -- the tradeoff is in the type safety of the (relatively simple) code in `createFromFile` here as opposed to the (relatively complex) code in, say, `createApiConfig`.

src/main/java/com/google/api/codegen/transformer/py/PythonPackageMetadataTransformer.java
geigerj: Done.

garrettjonesgoogle:
LGTM

geigerj:
@shinfan Do you want to take another look before I merge?

shinfan:
LGTM

PR #875, opened 2016-12-20
Add LRO support for PHP

michaelbausor:
Depends on https://github.com/googleapis/gax-php/pull/42

- Add LRO support
- Rename grpc client instead of using aliased import
- Add optional arg to replace grpc stub in client class


garrettjonesgoogle:
LGTM after you get the tests passing

PR #877, opened 2016-12-21
Clarify overwriting of resource fields in request bodies of methods with HTTP PUT semantics.

tcoffee-google:
Also make language-specific terminology for resource fields consistent.

### DIFF
--- a/src/main/resources/com/google/api/codegen/nodejs/sample.snip
+++ b/src/main/resources/com/google/api/codegen/nodejs/sample.snip
@@ -64,13 +64,13 @@
       resource: {
         @switch class.methodVerb
         @case "PATCH"
-          // TODO: Add desired properties of the request body. Only these properties
+          // TODO: Add desired properties to the request body. Only these properties
           // will be changed.
         @case "PUT"
-          // TODO: Add desired properties of the request body. All existing properties
+          // TODO: Add desired properties to the request body. All existing properties
           // will be replaced.
         @default
-          // TODO: Add desired properties of the request body.
+          // TODO: Add desired properties to the request body.
         @end
       },
 
--- a/src/main/resources/com/google/api/codegen/nodejs/sample.snip
+++ b/src/main/resources/com/google/api/codegen/nodejs/sample.snip
@@ -62,10 +62,14 @@
     @end
     @if class.hasRequestBody
       resource: {
-        @if class.methodVerb == "PATCH"
-          // TODO: Add desired properties of the request body. Only added properties
+        @switch class.methodVerb
+        @case "PATCH"
+          // TODO: Add desired properties of the request body. Only these properties
           // will be changed.
-        @else
+        @case "PUT"
+          // TODO: Add desired properties of the request body. All existing properties
+          // will be replaced.
+        @default
           // TODO: Add desired properties of the request body.
         @end
       },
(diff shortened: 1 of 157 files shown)


saicheems:
lgtm

src/main/resources/com/google/api/codegen/nodejs/sample.snip
```diff
@@ -62,10 +62,14 @@
     @end
     @if class.hasRequestBody
       resource: {
-        @if class.methodVerb == "PATCH"
-          // TODO: Add desired properties of the request body. Only added properties
+        @switch class.methodVerb
+        @case "PATCH"
+          // TODO: Add desired properties of the request body. Only these properties
```
vchudnov-g: Should we change these to "Assign values to desired properties of..." to be have wording more consistent with the other languages'?

src/main/resources/com/google/api/codegen/py/sample.snip
```diff
@@ -71,10 +71,14 @@
...
         {@class.requestBodyVarName} = {
-            @if class.methodVerb == "PATCH"
+            @switch class.methodVerb
+            @case "PATCH"
                 @# TODO: Add desired entries of the request body. Only assigned entries
-                @# will be changed:
-            @else
+                @# will be changed.
+            @case "PUT"
+                @# TODO: Add desired entries of the request body. All existing entries
```
vchudnov-g: Ditto with consistent wording for these.

src/test/java/com/google/api/codegen/testdata/discoveries/go/go_compute.v1.json.baseline
```diff
@@ -3555,7 +3559,8 @@ func main() {
   httpHealthCheck := ""  // TODO: Update placeholder value.
 
   rb := &compute.HttpHealthCheck{
-    // TODO: Add desired fields of the request body.
+    // TODO: Add desired fields of the request body to be changed. All existing
```
vchudnov-g: "to be changed" reads confusingly here to me, since it implies only the fields to be changed should be filled in, but the next sentence implies fields not filled in will be overwritten. How can we clarify?

src/main/resources/com/google/api/codegen/nodejs/sample.snip
tcoffee-google: This inconsistency is deliberate, because in Node.js/Python one is constructing an object literal/dictionary whose collection of properties/entries is not predetermined, hence these are "added" rather than "assigned". I changed the wording here because "added properties" seemed open to misinterpretation (could mean added to the original object rather than added to the request body).

src/test/java/com/google/api/codegen/testdata/discoveries/go/go_compute.v1.json.baseline
tcoffee-google: Agreed, I overlooked that this was inconsistent before and should be changed.

tcoffee-google:
PTAL

src/main/resources/com/google/api/codegen/nodejs/sample.snip
vchudnov-g: Hmmm, I think "Assign values..." would work just as well, but OK.

If keeping as you have it, how about s/of the request body/to the request body/g?

vchudnov-g:
Just one comment about "**to** the request body". Otherwise LGTM.

Issue #692, opened 2016-10-27
Check all languages for cases where `pageToken` is in the request body


saicheems:
Currently wrong in Node.js for `logging.entries.list`

Some APIs have the `pageToken` field in the request body. For a few languages we generate code as if it is in the parameter list instead. Only Go and Java are correct right now.

saicheems:
Go is okay, Java is right.

vchudnov-g:
Could you add a bit more context to this issue? ("Requests should ... but some do ... and so our samples do ..."). Nothing long, just a few words.

saicheems:
Doen

tcoffee-google:
Here is a summary of occurrences for 'pageToken' and 'nextPageToken' in recent live discovery docs:

Method parameters | Request body fields | Response fields | Methods
------------ | ------------- | ------------- | -------------
- | - | - | 3568
pageToken | - | nextPageToken | 663
- | pageToken | nextPageToken | cloudresourcemanager.organizations.search <br/> genomics.annotationSets.search <br/> genomics.annotations.search <br/> genomics.annotationsets.search <br/> genomics.callsets.search <br/> genomics.jobs.search <br/> genomics.readgroupsets.search <br/> genomics.reads.search <br/> genomics.references.search <br/> genomics.referencesets.search <br/> genomics.variants.search <br/> genomics.variantsets.search <br/> logging.entries.list <br/>
pageToken | - | - | adexchangebuyer.performanceReport.list <br/> books.layers.list <br/> calendar.acl.watch <br/> calendar.calendarList.watch <br/> calendar.events.watch <br/> calendar.settings.watch <br/> clouduseraccounts.linux.getLinuxAccountViews <br/> directory.users.watch <br/> drive.changes.watch <br/> ghire.feedbacks.list <br/> reports.activities.watch <br/> storage.objects.watchAll <br/>
- | - | nextPageToken | blogger.posts.search <br/> compute.projects.listXpnHosts <br/> sql.users.list <br/> urlshortener.url.list <br/> youtube.guideCategories.list <br/> youtube.videoCategories.list <br/>
pageToken | - | pageToken | bigquery.jobs.getQueryResults <br/> bigquery.tabledata.list <br/>
- | nextPageToken | nextPageToken | identitytoolkit.relyingparty.downloadAccount
- | - | pageToken | bigquery.jobs.query

tcoffee-google:
[...]

PR #880, opened 2016-12-22
Make page token identification consistent with all APIs

tcoffee-google:
Addresses #692.

### DIFF
--- a/src/main/java/com/google/api/codegen/discovery/config/ApiaryConfigToSampleConfigConverter.java
+++ b/src/main/java/com/google/api/codegen/discovery/config/ApiaryConfigToSampleConfigConverter.java
@@ -17,7 +17,9 @@ package com.google.api.codegen.discovery.config;
 import com.google.api.codegen.ApiaryConfig;
 import com.google.api.codegen.DiscoveryImporter;
 import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Table;
 import com.google.protobuf.Field;
 import com.google.protobuf.Field.Kind;
 import com.google.protobuf.Method;
@@ -32,8 +34,8 @@ public class ApiaryConfigToSampleConfigConverter {
 
   private static final String KEY_FIELD_NAME = "key";
   private static final String VALUE_FIELD_NAME = "value";
-  private static final String PAGE_TOKEN_FIELD_NAME = "pageToken";
-  private static final String NEXT_PAGE_TOKEN_FIELD_NAME = "nextPageToken";
+  private static final ImmutableList<String> PAGE_TOKEN_NAMES =
+      ImmutableList.of("pageToken", "nextPageToken");
 
   private final List<Method> methods;
   private final ApiaryConfig apiaryConfig;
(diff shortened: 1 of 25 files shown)


tcoffee-google:
It looks like there will be some additional steps to fully resolve page-streaming, namely (1) adding cases for page tokens in request bodies to some of the languages, which seems to have been overlooked, and (2) possibly adding these cases to the client library generators for some languages. But I wanted to get these changes reviewed sooner since they have higher conflict potential.

tcoffee-google:
Note that among non-Cloud APIs there should be several more instances fixed by this change. See table in #692.

src/main/java/com/google/api/codegen/discovery/config/ApiaryConfigToSampleConfigConverter.java
```diff
@@ -264,24 +308,44 @@ private MessageTypeInfo createMessageTypeInfo(
...
-   * <p>The heuristic implemented checks if there is some field "nextPageToken" within the method's
-   * response type, and returns true if so.
+   * <p>The heuristic implemented checks if one of the names "pageToken" or "nextPageToken" occurs
+   * both among the fields of the method's response type and among either the method's parameters or
+   * the fields of the method's request body.
    */
-  private boolean isPageStreaming(Method method) {
-    Type type = apiaryConfig.getType(method.getResponseTypeUrl());
-    if (type == null) {
+  private boolean isPageStreaming(
```
vchudnov-g: Is this method still used? The calling line seems to have been removed above.

src/main/java/com/google/api/codegen/discovery/config/ApiaryConfigToSampleConfigConverter.java
```diff
@@ -264,24 +308,44 @@ private MessageTypeInfo createMessageTypeInfo(
...
+    Table<Type, String, Field> fields = apiaryConfig.getFields();
+    for (String tokenName : PAGE_TOKEN_NAMES) {
+      if (fields.contains(requestType, tokenName)) {
         return true;
       }
     }
     return false;
   }
 
+  private boolean hasResponsePageToken(Method method) {
```
vchudnov-g: The logic here seems very similar to that in createMethod. Factor out?

src/main/java/com/google/api/codegen/discovery/transformer/nodejs/NodeJSSampleMethodToViewTransformer.java
```diff
@@ -163,6 +163,8 @@ private SamplePageStreamingView createSamplePageStreamingView(
         symbolTable.getNewSymbol(namer.localVarName(Name.lowerCamel("handlePage"))));
     builder.pageVarName(
         symbolTable.getNewSymbol(namer.localVarName(Name.lowerCamel(fieldInfo.name(), "page"))));
+    builder.pageTokenName(methodInfo.requestPageTokenName());
```
vchudnov-g: Just checking: no need for case manipulation?

src/main/java/com/google/api/codegen/discovery/transformer/php/PhpSampleMethodToViewTransformer.java
```diff
@@ -169,6 +169,8 @@ private SamplePageStreamingView createSamplePageStreamingView(
     builder.isResourceMap(fieldInfo.type().isMap());
     builder.pageVarName(
         symbolTable.getNewSymbol(namer.localVarName(Name.lowerCamel(fieldInfo.name()))));
+    builder.pageTokenName(methodInfo.requestPageTokenName());
```
vchudnov-g: How come we need case manipulation for the nextPageToken and  not for pageToken?

src/test/java/com/google/api/codegen/testdata/discoveries/nodejs/nodejs_sqladmin.v1beta4.json.baseline
```diff
@@ -2067,28 +2067,15 @@ authorize(function(authClient) {
...
-    }
-
-    if (response.nextPageToken) {
-      request.pageToken = response.nextPageToken;
-      sqlAdmin.users.list(request, handlePage);
-    }
-  };
-
-  sqlAdmin.users.list(request, handlePage);
+    // TODO: Change code below to process the `response` object:
```
vchudnov-g: This diff is eye-catching, but I take it it's expected?

src/main/java/com/google/api/codegen/discovery/config/ApiaryConfigToSampleConfigConverter.java
tcoffee-google: Good catch, will remove.

src/main/java/com/google/api/codegen/discovery/config/ApiaryConfigToSampleConfigConverter.java
tcoffee-google: Also to be removed.

src/main/java/com/google/api/codegen/discovery/transformer/nodejs/NodeJSSampleMethodToViewTransformer.java
tcoffee-google: Correct. We access fields directly by name in Node.js.

src/main/java/com/google/api/codegen/discovery/transformer/php/PhpSampleMethodToViewTransformer.java
tcoffee-google: In PHP, the pageToken is assigned to an array by name, but the nextPageToken is accessed from an object by method ("getNextPageToken").

src/test/java/com/google/api/codegen/testdata/discoveries/nodejs/nodejs_sqladmin.v1beta4.json.baseline
tcoffee-google: Yes, the override for SQLAdmin was never added for two languages before, so this fixes it.

src/main/java/com/google/api/codegen/discovery/config/ApiaryConfigToSampleConfigConverter.java
```diff
@@ -86,53 +88,95 @@ public SampleConfig convert() {
...
-      Field field = getPageStreamingResourceField(containerType);
-      pageStreamingResourceField = createFieldInfo(field, containerType, method);
-    }
     boolean isPageStreamingResourceSetterInRequestBody = false;
+    String requestPageTokenName = "";
     if (requestBodyType != null) {
-      isPageStreamingResourceSetterInRequestBody =
-          requestBodyType.message().fields().containsKey(PAGE_TOKEN_FIELD_NAME);
+      Map<String, FieldInfo> requestBodyFields = requestBodyType.message().fields();
+      for (String tokenName : PAGE_TOKEN_NAMES) {
```
saicheems: For this and the loop below, we could simplify a bit by writing a function `String hasField(MessageType messageType, String name...);` that returns the name of the field if found and `null` otherwise

saicheems:
Thanks for doing this! Could you make sure compilecheck is good?

src/main/java/com/google/api/codegen/discovery/config/ApiaryConfigToSampleConfigConverter.java
tcoffee-google: Unfortunately, each of the three loops uses a different sort of lookup, so this style would either require three different functions or fancy functional programming, either of which seems more complex. Unless I'm missing something.

tcoffee-google:
As mentioned above, I'm expecting we'll need some client library generator changes (as with Go) to get compilecheck passing for BigQuery. I'm investigating.

vchudnov-g:
Any updates on this?

tcoffee-google:
Client library generator changes are in progress. This is my first foray into the code of six of the generators, so it's been a bit slow going. I'm thinking we'll be able to use dartman to make a single PR for all.

PR #884, opened 2016-12-27
Improve method samples for longrunning methods

michaelbausor:
Also change namespace for OperationsClient


src/main/resources/com/google/api/codegen/php/method_sample.snip
```diff
@@ -37,6 +41,22 @@
     }
 @end
 
+@private longRunningOptionalArrayMethodSampleCode(apiMethod)
```
garrettjonesgoogle: Since all PHP methods are optional array methods, I think you could remove the "OptionalArray" from the snippet function names.

src/main/java/com/google/api/codegen/viewmodel/OptionalArrayMethodView.java
```diff
@@ -57,6 +57,8 @@
 
   public abstract boolean isLongrunningOperation();
 
+  public abstract boolean hasLongrunningReturnValue();
```
garrettjonesgoogle: `LongRunningOperationDetailView` already has `isEmptyOperation`.

michaelbausor: But LongRunningOperationDetailView is not available in OptionalArrayMethodView. Do you think it is better to construct and add a LongRunningOperationDetailView field, even if only the isEmptyOperation is used?

src/main/resources/com/google/api/codegen/php/method_sample.snip
michaelbausor: Done

src/main/java/com/google/api/codegen/viewmodel/OptionalArrayMethodView.java
garrettjonesgoogle: That's the only one used now, but won't the others be used at some point too?

michaelbausor: It's possible, but I don't think so necessarily. Longrunning operations are now supported without them, so until we find additional features that need to be added, there won't be more changes

michaelbausor:
@garrettjonesgoogle What do you think? I am happy to add LongRunningOperationDetailView if that is preferable.

src/main/java/com/google/api/codegen/viewmodel/OptionalArrayMethodView.java
garrettjonesgoogle: My concern is having multiple places to specify the same thing. It could get confusing for another language implementing LRO features, even if we currently have it clear in the existing languages.

michaelbausor: Done

michaelbausor:
PTAL

garrettjonesgoogle:
LGTM

PR #886, opened 2016-12-28
Update paged method samples and constructor docs

michaelbausor:
- Fix missing iterateAllElements bug in paged method samples
- Add page iteration sample with pageSize parameter
- Add throws declaration to <X>Client constructor when $options arg is
  required

### DIFF
--- a/src/main/resources/com/google/api/codegen/php/main.snip
+++ b/src/main/resources/com/google/api/codegen/php/main.snip
@@ -285,6 +285,10 @@
      *                              A CredentialsLoader object created using the
      *                              Google\Auth library.
      * }
+     @if xapiClass.hasMissingDefaultOptions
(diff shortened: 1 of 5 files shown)


src/main/resources/com/google/api/codegen/php/main.snip
```diff
@@ -285,6 +285,10 @@
      *                              A CredentialsLoader object created using the
      *                              Google\Auth library.
      * }
+     @if xapiClass.hasMissingDefaultOptions
+     * @throws ValidationException Throws a ValidationException if required arguments are missing
+     *                             from the $options array.
+     @end
```
garrettjonesgoogle: The stars should be indented here (for the snippet's indentation - won't affect the output indentation)

PR #887, opened 2016-12-29
Work in progress - Plug Python into MVVM machinery

jcanizales:
Sorry for the delay, I hope it'll help!

This is a commit prior to fixing the snippets to not depend on elements of the
Model; instead getting what they need from ViewModel objects. It’s not
really MVVM yet, but it allows tests to pass while migrating, so it can
be done incrementally (and thus is easier to reason about and review).

This commit itself is work in progress. Things left to do:

  * [ ] Fix the repetitive usage of `CommonGapicCodePathMapper` in `ServiceViewModel` and `EnumViewModel`.
  * [x] Fix failing baseline test: `no_path_templates[python_enum]`.~
  * [ ] Comprehensive javadocs.


src/main/java/com/google/api/codegen/py/PythonInterfaceInitializer.java
```diff
@@ -1,38 +0,0 @@
...
-  }
-
-  @Override
-  public PythonImportHandler getImportHandler(Interface iface) {
-    return new PythonImportHandler(iface, apiConfig);
-  }
-
-  @Override
-  public ImmutableMap<String, Object> getGlobalMap(Interface iface) {
-    return ImmutableMap.of("pyproto", (Object) new PythonProtoElements());
```
jcanizales: This global wasn't used anywhere in the snippets.

src/main/java/com/google/api/codegen/transformer/py/EnumViewModel.java
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen.transformer.py;
+
+import com.google.api.client.repackaged.com.google.common.base.Preconditions;
```
jcanizales: Not sure why Eclipse did this. Are we vendoring Guava within GAPIC, instead of just depending on it?

shinfan: `src/main/java/com/google/api/codegen/util/CommonAcronyms.java` uses com.google.common.base.Preconditions

src/main/java/com/google/api/codegen/transformer/py/ServiceViewModel.java
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen.transformer.py;
+
+import com.google.api.client.repackaged.com.google.common.base.Preconditions; // ???
```
shinfan: Replace with com.google.common.base.Preconditions

src/main/java/com/google/api/codegen/transformer/py/ServiceViewModel.java
```diff
@@ -0,0 +1,78 @@
...
+import com.google.api.codegen.py.PythonGapicContext;
+import com.google.api.codegen.py.PythonImportHandler;
+import com.google.api.codegen.rendering.CommonSnippetSetRunner.TransitionViewModel;
+import com.google.api.codegen.viewmodel.ViewModel;
+import com.google.api.tools.framework.model.Interface;
+import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableMap;
+import java.util.Map;
+
+class ServiceViewModel implements ViewModel, TransitionViewModel {
```
shinfan: View models are supposed to be data-container classes only. Move the extra logic into `PythonGapicSurfaceTransformer`

src/main/java/com/google/api/codegen/transformer/py/EnumViewModel.java
```diff
@@ -0,0 +1,73 @@
...
+import com.google.api.codegen.gapic.CommonGapicCodePathMapper;
+import com.google.api.codegen.py.PythonGapicContext;
+import com.google.api.codegen.rendering.CommonSnippetSetRunner.TransitionViewModel;
+import com.google.api.codegen.viewmodel.ViewModel;
+import com.google.api.tools.framework.model.Model;
+import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableMap;
+import java.util.Map;
+
+public class EnumViewModel implements ViewModel, TransitionViewModel {
```
shinfan: Same comments as `ServiceViewModel`

src/main/java/com/google/api/codegen/transformer/py/EnumViewModel.java
garrettjonesgoogle: When you are using Eclipse and you type in `Preconditions` and auto-complete it, Eclipse chooses from the available imports for you automatically. Unfortunately, it typically makes a bad choice and gives you vendored types from other dependencies instead of the main type that you want.

lukesneeringer:
> Fix failing baseline test: no_path_templates[python_enum]. Expects an empty example/enums.py, gets a different, still empty one (whitespaces?), and fails. Now that I look at it again, I think the difference is my code did generate an empty example/enums.py, whereas the baseline is to not generate any file at all?

The actual issue is that the `.baseline` files have a special header as the first; even empty files therefore are not really empty. I fixed this and pushed. :-)

src/main/java/com/google/api/codegen/transformer/py/EnumViewModel.java
jcanizales: Who's vendoring Guava (declaring `com.google.api.client.repackaged`), though? We control the `com.google.api.client` namespace.

src/main/java/com/google/api/codegen/transformer/py/ServiceViewModel.java
jcanizales: Yeah, the intention here was to be able and do it in increments, instead of sending out a CthulhuPullRequest with several independent changes.

jcanizales: (As I was commenting with @lukesneeringer earlier, I don't even expect this class to survive till the end of the migration).

geigerj:
@jcanizales Would definitely prefer not to general empty `.py` files.

lukesneeringer:
I am going to actually close this PR for now. I have changes that are massive enough (will be pushed hopefully today, maybe tomorrow) that it more or less renders the current questions moot.

jcanizales:
> I have changes that are massive enough

That doesn't sound good :) Can you make your changes manageable to review?

lukesneeringer:
> I am going to actually close this PR for now. I have changes that are massive enough (will be pushed hopefully today, maybe tomorrow) that it more or less renders the current questions moot.

This was a mistake. Re-opening.

I am doing more work against this PR to get rid of `ServiceViewModel`, but I also think it would be fine to merge this so that the incremental changes are getting into trunk.

garrettjonesgoogle:
Why are you marking this PR as approved? It isn't in a state where it can be merged to master.

eoogbe:
Work is continuing in a separate branch

PR #889, opened 2017-01-03
Auto-generate version index file for NodeJS.

shinfan:
cc/ @swcloud @bjwatson

### DIFF
--- a/src/main/resources/com/google/api/codegen/nodejs/version_index.snip
+++ b/src/main/resources/com/google/api/codegen/nodejs/version_index.snip
@@ -1,0 +1,11 @@
+@extends "nodejs/common.snip"
+
+@snippet generate(index)
+  /*
+   {@copyright()}
+   */
+  'use strict';
+
+  module.exports.v1 = require('./{@index.apiVersion}');
+@end
+
--- a/src/main/resources/com/google/api/codegen/nodejs/version_index.snip
+++ b/src/main/resources/com/google/api/codegen/nodejs/version_index.snip
@@ -6,6 +6,6 @@
    */
   'use strict';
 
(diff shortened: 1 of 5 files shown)


garrettjonesgoogle:
The generator changes LGTM.

src/main/resources/com/google/api/codegen/nodejs/version_index.snip
```diff
@@ -0,0 +1,11 @@
+@extends "nodejs/common.snip"
+
+@snippet generate(index)
+  /*
+   {@copyright()}
+   */
+  'use strict';
+
+  module.exports.v1 = require('./{@index.apiVersion}');
```
landrito: This should be:

```
module.exports.{@index.apiVersion} = require('./{@index.apiVersion}');

Comment (shinfan) [inline src/main/resources/com/google/api/codegen/nodejs/version_index.snip]: Oops you are right. This one was not caught by the baseline. Fixed

**Comment by shinfan** [comment]
@landrito PTAL

File: src/test/java/com/google/api/codegen/testdata/nodejs_doc_version_index_library.baseline [hunk 2]
```diff
@@ -0,0 +1,19 @@
+============== file: src/index.js ==============
+/*
+ * Copyright 2016 Google Inc. All rights reserved.
```
landrito: nit: Should we be using 2017 now?

shinfan: Lol you are right. This is a from a shared component so I prefer to fix in a separate PR. https://github.com/googleapis/toolkit/issues/897

PR #891, opened 2017-01-03
Integrate _OperationFuture into Python LRO

eoogbe:
Implements LRO for Python.

Depends on googleapis/gax-python#142 to generate functioning code.

### DIFF
--- a/src/main/java/com/google/api/codegen/py/PythonImportHandler.java
+++ b/src/main/java/com/google/api/codegen/py/PythonImportHandler.java
@@ -87,8 +87,8 @@ public class PythonImportHandler {
     for (MethodConfig methodConfig : apiConfig.getInterfaceConfig(service).getMethodConfigs()) {
       if (methodConfig.isLongRunningOperation()) {
         addImportExternal("google.gapic.longrunning", "operations_client");
-        addImportForType(methodConfig.getLongRunningConfig().getReturnType());
-        addImportForType(methodConfig.getLongRunningConfig().getMetadataType());
+        addImportForMessage(methodConfig.getLongRunningConfig().getReturnType().getMessageType());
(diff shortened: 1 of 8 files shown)


eoogbe:
@landrito Something like this?

```python
          >>> from google.cloud.gapic.example.library.v1 import library_service_client
          >>> api = library_service_client.LibraryServiceClient()
          >>> name = api.book_path('[SHELF_ID]', '[BOOK_ID]')
          >>> response = api.get_big_book(name)
          >>> response.add_done_callback(do_something_with)
```

landrito:
A bit more verbose. [This](https://github.com/googleapis/toolkit/blob/master/src/test/java/com/google/api/codegen/testdata/ruby_doc_main_library.baseline#L1233) is what the example is for Ruby.

For python I'd imagine it to look like.
```python
from google.cloud.gapic.example.library.v1 import library_service_client
api = library_service_client.LibraryServiceClient()
name = api.book_path('[SHELF_ID]', '[BOOK_ID]')
operation_future = api.get_big_book(name)

def callback(operation_future):
    # Handle response.
    result = operation_future.result


response.add_done_callback(callback)

# Handle metadata.
metadata = operation_future.metadata
```

src/main/java/com/google/api/codegen/py/PythonImportHandler.java
```diff
@@ -233,6 +233,17 @@ public String addImportLocal(String moduleName, String attributeName) {
     return addImport(ImportType.APP, moduleName, attributeName).shortName();
   }
 
+  /** Add an import for the proto associated with the given type. */
+  private PythonImport addImportForType(TypeRef type) {
```
geigerj: Make this method `addImportForMessage(MessageType message)`. As-is, this will raise a NullPointerException for any non-message TypeRef.

PR #894, opened 2017-01-04
Add sample generator for the JS discovery client

saicheems:


saicheems:
Hey folks, got this to a state where it's mostly done - I've been manually testing the samples, and am reasonably confident of most functionality.

That being said, there's still some polish to get through tomorrow, so please restrict review to the structure of the sample for now.

saicheems:
Oh and one other bit of cool news is that I've taken the liberty to add a working OAuth2 flow since its setup is very terse with this client. So samples will pop up a consent screen if you include a good `clientId`.

src/main/java/com/google/api/codegen/discovery/MainDiscoveryProviderFactory.java
```diff
@@ -73,6 +76,7 @@
           .put(CSHARP, CSharpTypeNameGenerator.class)
           .put(GO, GoTypeNameGenerator.class)
           .put(JAVA, JavaTypeNameGenerator.class)
+          .put(JS, NodeJSTypeNameGenerator.class) // Intentional mismatch.
```
vchudnov-g: I wonder whether it makes sense to make switch the mismatch so that it's JSTypeNameGenerator.class, since NodeJS is a superset of JS.

src/main/java/com/google/api/codegen/discovery/transformer/js/JSSampleMethodToViewTransformer.java
```diff
@@ -0,0 +1,159 @@
+/* Copyright 2016 Google Inc
```
vchudnov-g: 2017 on new files.

Keep the creation date on old files (cf. https://engdoc.corp.google.com/eng/doc/copyright.html?cl=head)

src/main/java/com/google/api/codegen/discovery/viewmodel/SampleView.java
```diff
@@ -146,6 +148,17 @@ public String resourceRoot() {
   @Nullable
   public abstract List<String> methodParamAssigments();
 
+  // JS specific section...
```
vchudnov-g: Nit: I'd remove the ellipses in each of the language section headers.

src/main/resources/com/google/api/codegen/js/sample.snip
```diff
@@ -0,0 +1,181 @@
...
+  @if class.isPageStreaming
+    function {@pageStreamingFuncSignature(class)} {
+      var {@class.requestVarName} = {@requestFuncSignature(class)};
+      request.then(function({@class.responseVarName}) {
+        var {@class.pageStreaming.pageVarName} = {@class.responseVarName}.result.{@class.pageStreaming.resourceFieldName};
+        @if class.pageStreaming.isResourceRepeated
+          @if class.pageStreaming.isResourceMap
+            Object.keys({@class.pageStreaming.pageVarName}).forEach(function({@class.pageStreaming.resourceKeyVarName}) {
+              // TODO: Change code below to process each `{@class.pageStreaming.resourceKeyVarName}` property:
+              console.log({@class.pageStreaming.resourceKeyVarName} + ': ' + JSON.stringify({@class.pageStreaming.pageVarName}[{@class.pageStreaming.resourceKeyVarName}], null, 2));
```
tcoffee-google: I believe JSON is always available with Node.js, but is that true for all Javascript environments we want to support?

src/main/resources/com/google/api/codegen/js/sample.snip
```diff
@@ -0,0 +1,181 @@
...
+          @end
+          {@pageStreamingFuncSignature(class)};
+        }
+      }, function(reason) {
+        console.log('error: ' + reason.result.error.message);
+      });
+    }
+
+  @end
+  function start() {
```
tcoffee-google: Top-level function name like this may need collision resolution.

src/main/resources/com/google/api/codegen/js/sample.snip
```diff
@@ -0,0 +1,181 @@
...
+    gapi.client.init({
+      @switch class.auth.type
+      @case "API_KEY"
+        'apiKey': '',  // TODO: Update placeholder with desired API key.'
```
tcoffee-google: Extra trailing `'`.

src/main/resources/com/google/api/codegen/js/sample.snip
```diff
@@ -0,0 +1,181 @@
...
+  @join resource : methodNameComponents on "."
+    {@resource}
+  @end
+@end
+
+@private scopeString(scopes)
+  @join scope : scopes on " "
+    {@scope}
+  @end
+@end
```
tcoffee-google: Missing newline at EOF.

src/main/resources/com/google/api/codegen/js/sample.snip
```diff
@@ -0,0 +1,181 @@
...
+
+  @end
+  function start() {
+    gapi.client.init({
+      @switch class.auth.type
+      @case "API_KEY"
+        'apiKey': '',  // TODO: Update placeholder with desired API key.'
+      @default
+        'clientId': '',  // TODO: Update placeholder with desired client ID.
+        'scope': '{@scopeString(class.auth.scopes)}',
```
tcoffee-google: Scopes are always singular?

src/main/resources/com/google/api/codegen/js/sample.snip
```diff
@@ -0,0 +1,181 @@
...
+      @else
+        makeApiCall();
+      @end
+    });
+  }
+  gapi.load('client', start);
+  </script>
+@end
+
+@private requestFuncSignature(class)
```
tcoffee-google: Looks to me like we should abstract out a snippet for the argument lists in this and `pageStreamingFuncSignature`, and then these two snippets will be unnecessary.

src/test/java/com/google/api/codegen/testdata/discoveries/js/js_cloudmonitoring.v2beta2.json.baseline
```diff
@@ -0,0 +1,328 @@
...
+    // The project ID to which this time series belongs. The value can be the numeric project ID or
+    // string-based project name.
+    project: '',  // TODO: Update placeholder value.
+
+    // Metric names are protocol-free URLs as listed in the Supported Metrics page. For example,
+    // compute.googleapis.com/instance/disk/read_ops_count.
+    metric: '',  // TODO: Update placeholder value.
+
+    // End of the time interval (inclusive), which is expressed as an RFC 3339 timestamp.
+    youngest: '',  // TODO: Update placeholder value.
```
tcoffee-google: (Hmm, this field is missing the proper stringFormat annotation in discovery.)

src/main/resources/com/google/api/codegen/js/sample.snip
saicheems: It's... not necessarily the case. It's possible that an API could require more than one scope to perform a function (there's a sentence in one of the docs about this, I can find it if you like), but most require only one. The trick becomes deciding which one to include... in the other languages we don't implement the OAuth2 flow and don't make a decision, so here I felt it was better to err towards including everything so the request works.

We don't actually have enough context to make an informed decision, API explorer also checks all scopes that a method declares.

build.gradle
```diff
@@ -456,12 +456,13 @@ task verifyLicense << {
   def srcFiles = []
   sourceSets
       .collectMany{it.allJava.getSrcDirs()}
-      .each{it.eachFileRecurse(FileType.FILES, {srcFiles << it})}
+      .each{it.eachFileRecurse(FileType.FILES, {srcFiles << new Tuple(it, it.text)})}
```
saicheems: @garrettjonesgoogle LGTY?

src/main/java/com/google/api/codegen/discovery/transformer/js/JSSampleMethodToViewTransformer.java
```diff
@@ -0,0 +1,159 @@
...
+
+    SampleView.Builder builder = SampleView.newBuilder();
+
+    String requestVarName = symbolTable.getNewSymbol(namer.getRequestVarName());
+
+    if (methodInfo.isPageStreaming()) {
+      builder.pageStreaming(createSamplePageStreamingView(context, symbolTable));
+    }
+
+    List<SampleFieldView> fields = new ArrayList<>();
```
saicheems: Not in this one and some others because either the auth portions create no variables, or they're members of another scope.

src/main/java/com/google/api/codegen/discovery/transformer/js/JSSampleMethodToViewTransformer.java
```diff
@@ -0,0 +1,159 @@
...
+    for (FieldInfo field : methodInfo.fields().values()) {
+      fields.add(
+          SampleFieldView.newBuilder()
+              .name(field.name())
```
saicheems: May be misunderstanding

The `params` variable name and `requestBody` variable names are kept in the symbol table, so they cannot collide. JS doesn't keep the request body as a member of the `params` struct, so we don't have to do that `resource` -> `resource_` rename here.

src/main/java/com/google/api/codegen/discovery/viewmodel/SampleView.java
```diff
@@ -146,6 +148,17 @@ public String resourceRoot() {
...
+  @Nullable
+  public abstract String paramsVarName();
+
+  /**
+   * Returns true if the page streaming call requires a `params` object to set the `pageToken`
+   * field, even if there are no required params.
```
saicheems: This should be JS specific, I don't think any other language should use it. I've edited the doc string to clarify it's JS only.

src/main/resources/com/google/api/codegen/js/sample.snip
saicheems: Done

src/main/resources/com/google/api/codegen/js/sample.snip
saicheems: I believe so

src/main/java/com/google/api/codegen/discovery/viewmodel/SampleView.java
tcoffee-google: Sorry, that's not what I meant ... by the usual case, I mean the case where there are required parameters. The doc string at https://github.com/saicheems/toolkit/blob/1bb05a49bf12e5f474cc284837c3b5069291a7a1/src/main/java/com/google/api/codegen/discovery/transformer/js/JSSampleMethodToViewTransformer.java#L99 is clear about this point ("If there are fields, or ...").

src/main/resources/com/google/api/codegen/js/sample.snip
tcoffee-google: Can we make a note somewhere of what that set of environments is?

saicheems: It's a part of the standard implementation, so the chart shows support for all versions of Chrome, Firefox from a very early stage, and IE8, which I think is basically all environments we could support.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON

I'll make an inline comment in the snippet

src/main/java/com/google/api/codegen/discovery/viewmodel/SampleView.java
saicheems: Gotcha, clarified some more

saicheems:
All baselines are up

src/test/java/com/google/api/codegen/testdata/discoveries/js/js_cloudbilling.v1.json.baseline
```diff
@@ -0,0 +1,246 @@
...
+1. If not already done, enable the Google Cloud Billing API
+   and check the quota for your project at
+   https://console.developers.google.com/apis/api/cloudbilling
+2. Get access keys for your application. See
+   https://developers.google.com/api-client-library/javascript/start/start-js#get-access-keys-for-your-application
+-->
+<script src="https://apis.google.com/js/api.js"></script>
+<script>
+function makeApiCall() {
+  var params = {
```
tcoffee-google: Should we make these empty initializers single-line?

src/main/resources/com/google/api/codegen/js/sample.snip
```diff
@@ -0,0 +1,172 @@
...
+
+    @end
+    @if class.isPageStreaming
+      {@class.pageStreaming.executeRequestFuncName}({@argList(class, @FALSE)});
+    @else
+      var {@class.requestVarName} = gapi.client.{@class.apiName}.{@callChain(class.methodNameComponents)}({@argList(class, @FALSE)});
+      @if class.hasResponse
+        {@class.requestVarName}.then(function({@class.responseVarName}) {
+          // TODO: Change code below to process the `{@class.responseVarName}` object:
+          console.log({@class.responseVarName}.result);
```
tcoffee-google: Why not use JSON.stringify here?

src/main/resources/com/google/api/codegen/js/sample.snip
```diff
@@ -0,0 +1,172 @@
...
+          // TODO: Add desired properties to the request body. All existing properties
+          // will be replaced.
+        @default
+          // TODO: Add desired properties to the request body.
+        @end
+      };
+
+    @end
+    @if class.hasMediaUpload
+      // TODO: To do media uploads, it's more appropriate to use CORS. See
```
tcoffee-google: "More appropriate" than ...?

src/main/resources/com/google/api/codegen/js/sample.snip
```diff
@@ -0,0 +1,172 @@
...
+      var {@class.paramsVarName} = {
+        @join field : class.fields if class.fields on BREAK.add(BREAK)
+          @if field.description
+            @join line : util.getDocLines(field.description, 100)
+              // {@line}
+            @end
+
+
+          @end
+          {@field.name}: {@field.defaultValue},  // TODO: Update placeholder value.
```
tcoffee-google: I believe trailing commas don't work prior to ES5 ... should we worry about this?

saicheems: I'm inclined to say no... admittedly for the selfish reason that it makes the snippet more complicated...

But I think it's safe to limit to ES5, all major browsers support it, minus IE8, but I think that's up to the developer at some point

src/main/resources/com/google/api/codegen/js/sample.snip
saicheems: I've changed the language to say `To do media uploads, use CORS.`, but I'm really not sure what's appropriate here. I had copied it directly from the link - could you take a look at it and let me know what a better wording might be?

src/main/resources/com/google/api/codegen/js/sample.snip
saicheems: Good point, done. I copied this mostly from the Node sample... so I'll make a CL to do it there as well.

src/test/java/com/google/api/codegen/testdata/discoveries/js/js_cloudbilling.v1.json.baseline
saicheems: Done

src/main/resources/com/google/api/codegen/js/sample.snip
saicheems: NVM, Node was right, the only thing that isn't stringified is when the page streaming resource isn't repeated.

saicheems:
Updated to add buttons that manage authorization if a sample doesn't use API key based auth.

vchudnov-g:
@saicheems  Please sync with me before merging this PR because we need to coordinate this with some other changes.

src/main/resources/com/google/api/codegen/js/sample.snip
tcoffee-google: I would have thought our samples should conform to the Google JavaScript Style Guide, which says "Your code must not rely on any of behaviors that change between EcmaScript 3 and EcmaScript 5."

src/main/resources/com/google/api/codegen/js/sample.snip
tcoffee-google: That sounds fine. I'd prefer "To upload media" and "To download media".

tcoffee-google:
There are some earlier comments still open; if they've all been reviewed, please let me know.

src/main/java/com/google/api/codegen/discovery/MainDiscoveryProviderFactory.java
saicheems: Done

src/main/java/com/google/api/codegen/discovery/transformer/js/JSSampleMethodToViewTransformer.java
saicheems: Done

src/main/java/com/google/api/codegen/discovery/viewmodel/SampleView.java
saicheems: Done

src/main/java/com/google/api/codegen/discovery/transformer/js/JSSampleMethodToViewTransformer.java
```diff
@@ -0,0 +1,159 @@
...
+              .build());
+    }
+
+    boolean hasRequestBody = methodInfo.requestBodyType() != null;
+    if (hasRequestBody) {
+      String requestBodyVarName =
+          symbolTable.getNewSymbol(
+              namer.getRequestBodyVarName(methodInfo.requestBodyType().message().typeName()));
+      builder.requestBodyVarName(requestBodyVarName);
+      //fieldVarNames.add(requestBodyVarName);
```
saicheems: Removed

src/main/resources/com/google/api/codegen/js/sample.snip
saicheems: Feedback from the library authors was not to use JSON.stringify, partially because of the IE8 thing... they suggested instead to print the result object directly. Doing it this way also gives some extra interactivity in the browser console.

src/main/resources/com/google/api/codegen/js/sample.snip
saicheems: Done

src/main/resources/com/google/api/codegen/js/sample.snip
saicheems: Fixed

src/main/resources/com/google/api/codegen/js/sample.snip
saicheems: Done

src/main/resources/com/google/api/codegen/js/sample.snip
saicheems: Done

saicheems:
I think all comments here have been addressed so far

src/main/resources/com/google/api/codegen/js/sample.snip
tcoffee-google: Do the same considerations apply to Node.js?

src/main/resources/com/google/api/codegen/js/sample.snip
tcoffee-google: What's the purpose of the second argument to `argList`? It appears to be always `@FALSE`, and I'm confused by the doc string.

src/main/resources/com/google/api/codegen/js/sample.snip
saicheems: I don't think so... because node is based on v8 and can't have the same issues, and also because there's no browser interactivity

saicheems:
I can't see your comment for some reason, though I got an email about it

Thanks for pointing out that I forgot to call `argList(class, @TRUE)` in the appropriate places, I fixed that in the latest commit. It's for cases where the call has no params but does have a request body.

src/main/resources/com/google/api/codegen/js/sample.snip
saicheems: I messed that up when refactoring, there are 2 cases where it should be `@TRUE`, I've fixed it now. It's for cases where there is no `params` object but there is a request body. The method call will then have `...({}, requestBody);`

src/main/resources/com/google/api/codegen/js/sample.snip
```diff
@@ -0,0 +1,206 @@
...
+      </script>
+      @if class.auth.type != "API_KEY"
+        <button id="signin-button" onclick="{@class.handleSignInClickFuncName}()">Sign in</button>
+        <button id="signout-button" onclick="{@class.handleSignOutClickFuncName}()">Sign out</button>
+      @end
+    </body>
+  </html>
+@end
+
+# If `isApiCall` is true, then if `needParams` is false and `hasRequestBody` is
```
tcoffee-google: I'd word this more clearly as:
```
If `needParams` is false and `hasRequestBody` is true, then the function will return "{}, {@class.requestBodyVarName}" if `isCall` is true, otherwise will return "{@class.requestBodyVarName}". The first case is needed for calls to the function dynamically generated by the client library, which takes the parameter list as first argument. The second case is used to simplify functions defined in the sample.
```

saicheems: Done

tcoffee-google: Sorry to be picky, but beginning a new sentence with the "Otherwise" retains the apparent inaccuracy I had meant to avoid. The whole comment pertains only to the case where `needParams` is false and `hasRequestBody` is true.

saicheems: No worries, fixed

vchudnov-g:
@saicheems I'm trying to test this PR and my changes together. Could you update the branch on your forked repo to resolve the merge conflicts, and then I'll try pulling from there again? Thanks!

saicheems:
Done

vchudnov-g:
OK, you can submit this because premature publication is blocked by the compilecheck code not knowing how to handle JS.

After you submit this, could you address https://github.com/googleapis/discovery-artifact-manager/issues/13? Don't submit *that* until I can test it with my changes.

PR #896, opened 2017-01-04
Split PHP unit tests into separate files for each class

michaelbausor:
Also move renderFileHeader into common.snip
Added: group annotations to test files

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTestTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTestTransformer.java
@@ -47,7 +47,6 @@ import com.google.api.codegen.viewmodel.testing.TestCaseView;
 import com.google.api.tools.framework.model.Interface;
 import com.google.api.tools.framework.model.Method;
 import com.google.api.tools.framework.model.Model;
-import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableList;
 import java.io.File;
 import java.util.ArrayList;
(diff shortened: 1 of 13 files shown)


michaelbausor:
Added group annotations to test classes, PTAL

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTestTransformer.java
```diff
@@ -61,32 +65,71 @@
...
   @Override
   public List<String> getTemplateFileNames() {
-    return Collections.singletonList(TEST_TEMPLATE_FILE);
+    return ImmutableList.<String>of(TEST_TEMPLATE_FILE, MOCK_SERVICE_TEMPLATE_FILE);
   }
 
   @Override
   public List<ViewModel> transform(Model model, ApiConfig apiConfig) {
     List<ViewModel> models = new ArrayList<ViewModel>();
     PhpSurfaceNamer namer = new PhpSurfaceNamer(apiConfig.getPackageName());
```
shinfan: Since you have a `testPackageNamer` you may rename this into something like `surfacePackageNamer`.

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTestTransformer.java
```diff
@@ -187,4 +233,29 @@ private void addUnitTestImports(ModelTypeTable typeTable) {
     typeTable.saveNicknameFor("\\Google\\GAX\\Testing\\MockStubTrait");
     typeTable.saveNicknameFor("\\PHPUnit_Framework_TestCase");
   }
+
+  /** Insert "Tests" into the package name after "Google\Cloud" */
+  private static String getTestPackageName(String packageName) {
```
shinfan: Should we move this logic to the namer?

src/main/resources/com/google/api/codegen/php/test.snip
```diff
@@ -2,39 +2,20 @@
...
-     * This file was automatically generated - do not edit!
+    /**
+     * @@group {@apiTest.testClass.apiName}
+     * @@group grpc
      */
+    class {@apiTest.testClass.name} extends PHPUnit_Framework_TestCase
+    {
+        {@setupMethods(apiTest.testClass)}
+        @join test : apiTest.testClass.testCases
+            {@testCase(test)}
```
shinfan: Don't you need `@group` for each test case as well?

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTestTransformer.java
michaelbausor: Yes, done. Also refactored other pieces of code using similar logic.

src/test/java/com/google/api/codegen/testdata/php_mock_service_library.baseline
```diff
@@ -0,0 +1,68 @@
...
+ * GENERATED CODE WARNING
+ * This file was automatically generated - do not edit!
+ */
+
+namespace Google\Tests\Example\Library\V1;
+
+use \Google\GAX\GrpcCredentialsHelper;
+use \Google\GAX\Testing\MockStubTrait;
+use \PHPUnit_Framework_TestCase;
+use \google\example\library\v1\LibraryServiceGrpcClient;
```
shinfan: Are these imports used?

michaelbausor: The MockStubTrait and LibraryServiceGrpcClient are, the others not. The unused imports are cleaned up in artman by formatting tools - is that still the preferred way to handle them? Or do we want to try to remove them from the generated output

shinfan: I think this should be fine as long as they are cleaned up by the formatting tool.

shinfan:
LGTM

PR #900, opened 2017-01-05
Cleanup of generated clients

michaelbausor:
- Insert newlines
- Add comments
- Remove empty page streaming descriptors

### DIFF
--- a/src/main/resources/com/google/api/codegen/php/main.snip
+++ b/src/main/resources/com/google/api/codegen/php/main.snip
@@ -80,6 +80,7 @@
          * The default address of the service.
          */
         const SERVICE_ADDRESS = '{@xapiClass.serviceAddress}';
+
     @end
 
     /**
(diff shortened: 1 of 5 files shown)


src/main/resources/com/google/api/codegen/php/main.snip
```diff
@@ -286,7 +297,7 @@
      *                              Google\Auth library.
      * }
      @if xapiClass.hasMissingDefaultOptions
-         * @throws ValidationException Throws a ValidationException if required arguments are missing
+         * @@throws ValidationException Throws a ValidationException if required arguments are missing
```
shinfan: Is there anyway to catch this in the baseline?

michaelbausor: The only way would be to create a new service that didn't have default service address/scopes. Maybe this is worth it? I don't know if other languages will also benefit.

shinfan: Ah this is fine. I don't think it would worth the effort.

shinfan:
LGTM

PR #902, opened 2017-01-05
Remove isset check when constructor will not throw

michaelbausor:


src/main/resources/com/google/api/codegen/php/method_sample.snip
```diff
@@ -1,14 +1,30 @@
...
+@snippet decorateSampleCode(xapiClass, apiMethod, coreSampleCode)
     try {
-        ${@apiMethod.apiVariableName} = new {@apiMethod.apiClassName}();
+        @if xapiClass.hasMissingDefaultOptions
+            $options = [
+                @if xapiClass.missingDefaultServiceAddress
+                    'serviceAddress' => 'my-service-address',
+                @end
+                @if xapiClass.missingDefaultServiceScopes
+                    'scopes' => ['my-service-scope'],
```
garrettjonesgoogle: Why not make my-service-scope and my-service-address defaults if you still have to set them here?

michaelbausor: I'm not sure I know what you mean - my-service-address and my-service-scope are placeholders in the method sample, which the user has to set to the appropriate values. Maybe they should be in [MY-SERVICE-ADDRESS] format?

garrettjonesgoogle: oh! I get it. Nevermind.

garrettjonesgoogle:
LGTM

PR #907, opened 2017-01-06
Fix PhpPackageMetadataNamer regex

michaelbausor:
- Fix regex to handle v1beta1 case
- Convert Name to simple lowercase string
- Add tests for PhpPackageMetadataNamer


garrettjonesgoogle:
LGTM

PR #910, opened 2017-01-09
Support future years in the license check

saicheems:
The `verifyLicense` command compares each file's header to a hardcoded text
file that specifies the year as 2016. Since it's no longer 2016, and will not
be 2016 for the foreseeable future, the command now does an in-memory regex
replace for "Copyright 20[0-9]{2}" with "Copyright 20xx" before comparison.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -456,12 +456,13 @@ task verifyLicense << {
   def srcFiles = []
   sourceSets
       .collectMany{it.allJava.getSrcDirs()}
-      .each{it.eachFileRecurse(FileType.FILES, {srcFiles << it})}
+      .each{it.eachFileRecurse(FileType.FILES, {srcFiles << new Tuple(it, it.text)})}
   srcFiles = srcFiles
-      .findAll{!it.path.contains("/generated/") && it.path.endsWith(".java")}
-      .findAll{!it.text.startsWith(licenseText)}
+      .collect{new Tuple(it.get(0), it.get(1).replaceAll("Copyright 20[0-9]{2}", "Copyright 20xx"))}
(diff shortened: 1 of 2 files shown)


shinfan:
Sorry I am not sure if I understand the change. We were using the regex to do the header matching so it would work for both 2016 and 2017, am I right?

build.gradle
```diff
@@ -456,12 +456,13 @@ task verifyLicense << {
   def srcFiles = []
   sourceSets
       .collectMany{it.allJava.getSrcDirs()}
-      .each{it.eachFileRecurse(FileType.FILES, {srcFiles << it})}
+      .each{it.eachFileRecurse(FileType.FILES, {srcFiles << new Tuple(it, it.text)})}
   srcFiles = srcFiles
-      .findAll{!it.path.contains("/generated/") && it.path.endsWith(".java")}
-      .findAll{!it.text.startsWith(licenseText)}
+      .collect{new Tuple(it.get(0), it.get(1).replaceAll("Copyright 20[0-9]{2}", "Copyright 20xx"))}
+      .findAll{!it.get(0).path.contains("/generated/") && it.get(0).path.endsWith(".java")}
```
garrettjonesgoogle: shouldn't we exclude files before doing the string replacement?

saicheems: Good point, done

saicheems:
@shinfan The check was hardcoded to only pass if the year was 2016. Now it accepts any year between 2000 and 2099 via regex.

garrettjonesgoogle:
LGTM

shinfan:
LGTM

PR #912, opened 2017-01-10
Revert some override functionality

saicheems:
To allow introducing new fields to objects, this commit removes the pattern
matching feature from all fields except the root level language nodes.

It also adds configuration in the Go sample transformer to only consider
methods whose verb is "GET" to be page streaming calls as specified in the Go
client library generator.

### DIFF
--- a/src/main/java/com/google/api/codegen/discovery/ViewModelProvider.java
+++ b/src/main/java/com/google/api/codegen/discovery/ViewModelProvider.java
@@ -26,13 +26,11 @@ import com.google.api.codegen.discovery.transformer.SampleMethodToViewTransforme
 import com.google.api.codegen.rendering.CommonSnippetSetRunner;
 import com.google.api.codegen.viewmodel.ViewModel;
 import com.google.api.tools.framework.snippet.Doc;
-import com.google.common.collect.Lists;
 import com.google.protobuf.Method;
-import java.util.Collections;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
-import java.util.regex.Pattern;
 
 /*
  * Calls the MethodToViewTransformer on a method with the provided ApiaryConfig.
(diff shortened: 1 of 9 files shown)


src/main/java/com/google/api/codegen/discovery/transformer/go/GoSampleMethodToViewTransformer.java
```diff
@@ -96,7 +96,10 @@ private SampleView createSampleView(SampleTransformerContext context) {
       fieldVarNames.add(requestBodyVarName);
     }
 
-    if (methodInfo.isPageStreaming()) {
+    // The Go client only considers methods whose verb is "GET" to be page
```
tcoffee-google: Is this correct behavior?

src/main/java/com/google/api/codegen/discovery/ViewModelProvider.java
```diff
@@ -125,29 +123,28 @@ private static SampleConfig override(
...
-            tree.remove(fieldName);
-          } else if (defaultValue.isObject() && overrideValue.isObject()) {
-            merge((ObjectNode) defaultValue, (ObjectNode) overrideValue);
-          } else if (overrideValue != null) {
-            tree.set(fieldName, overrideValue);
-          }
+    Iterator<String> fieldNames = overrideTree.fieldNames();
+    while (fieldNames.hasNext()) {
+      String fieldName = fieldNames.next();
+      JsonNode primaryValue = tree.get(fieldName);
```
tcoffee-google: Why call these 'primary' and 'backup' instead of 'default' and 'override'?

tcoffee-google:
Also, is the removal of some of these overrides contingent on submission of #880?

src/main/java/com/google/api/codegen/discovery/transformer/go/GoSampleMethodToViewTransformer.java
saicheems: At the moment, yes
https://github.com/google/google-api-go-client/blob/master/google-api-go-generator/gen.go#L1398

saicheems: Unless you're asking like metaphorically, in which case I don't think so, methods whose verb is "POST" can also be paging

tcoffee-google: This appears to be changed in https://code-review.googlesource.com/#/c/10110/3/google-api-go-generator/gen.go . For reference, here's a list of non-GET method IDs which we interpret to be paging for at least one API version:

resourceviews.regionViews.listresources                                                                                            
compute.instanceGroups.listInstances                                                                                               
genomics.referencesets.search                                                                                                      
genomics.variants.search                                                                                                           
compute.regionInstanceGroups.listInstances                                                                                         
logging.entries.list                                                                                                               
resourceviews.zoneViews.listresources                                                                                              
identitytoolkit.relyingparty.downloadAccount                                                                                       
genomics.annotations.search                                                                                                        
genomics.annotationsets.search                                                                                                     
cloudresourcemanager.organizations.search                                                                                          
genomics.variantsets.search                                                                                                        
genomics.reads.search                                                                                                              
genomics.annotationSets.search                                                                                                     
dfareporting.dimensionValues.query                                                                                                 
genomics.readgroupsets.search                                                                                                      
genomics.jobs.search                                                                                                               
cih.userInteractions.list                                                                                                          
genomics.callsets.search                                                                                                           
genomics.references.search

src/main/java/com/google/api/codegen/discovery/ViewModelProvider.java
saicheems: I lazily copied the old version of this method, I'll update to the latter.

src/main/java/com/google/api/codegen/discovery/transformer/go/GoSampleMethodToViewTransformer.java
saicheems: #915

saicheems:
wrt #880, I don't think this PR should have any effect

src/main/java/com/google/api/codegen/discovery/ViewModelProvider.java
saicheems: Done

tcoffee-google:
LGTM

PR #917, opened 2017-01-11
Update Nodejs package gen with latest toolkit change

shinfan:
- Add missing EOF newlines 
- Fix package.snip


shinfan:
Baseline updated.

PR #919, opened 2017-01-11
Support error testing in Nodejs

shinfan:
_Note: All new generated tests have been verified manually with Speech client._


src/main/resources/com/google/api/codegen/nodejs/test.snip
```diff
@@ -86,6 +90,27 @@
...
+      var client = {@moduleName}.{@test.serviceConstructorName}();
+      // Mock request
+      {@initCodeLines(test.initCode)}
+
+      // Mock Grpc layer
+      client._{@test.clientMethodName} = function(actualRequest, options, callback) {
+        assert.deepStrictEqual(actualRequest, request);
+        var error = new Error('Fake error');
+        error.code = FAKE_STATUS_CODE;
+        callback(error, null);
```
landrito: I think just using `callback(error);` is fine.

src/main/resources/com/google/api/codegen/nodejs/test.snip
```diff
@@ -86,6 +90,27 @@
...
+      };
+
+      client.{@test.clientMethodName}(request, function(err, response) {
+        assert.equal(err.code, FAKE_STATUS_CODE);
```
landrito: Add `assert.type(err, Error);`

src/main/resources/com/google/api/codegen/nodejs/test.snip
```diff
@@ -115,6 +140,27 @@
       });
     });
   });
+
```
landrito: Looks like this is duplicate code. Consider consolidating it.

src/main/resources/com/google/api/codegen/nodejs/test.snip
shinfan: Done

src/main/resources/com/google/api/codegen/nodejs/test.snip
shinfan: It seems like `Assert` does not have a type method (https://nodejs.org/api/assert.html). Used `assert(err instanceof Error)`

src/main/resources/com/google/api/codegen/nodejs/test.snip
shinfan: Done

shinfan:
@landrito PTAL

src/main/resources/com/google/api/codegen/nodejs/test.snip
landrito: Whoops! Looked at the wrong assertion library docs.

src/test/java/com/google/api/codegen/testdata/nodejs_test_library.baseline
```diff
@@ -51,6 +54,29 @@ describe('LibraryServiceClient', function() {
...
+    it('invokes createShelf with error', function(done) {
+      var client = libraryV1.libraryServiceClient();
+      // Mock request
+      var shelf = {};
+      var request = {
+          shelf : shelf
+      };
+
+      // Mock Grpc layer
+      client._createShelf = function(actualRequest, options, callback) {
```
landrito: I think the creation of the mock method can be consolidated to reduce code reuse as seen below:

Utility method used by non-streaming tests:
```js
function mockGrpcMethod(expectedRequest, response) {
  return function(actualRequest, options, callback) {
    assert.deepStrictEqual(actualRequest, expectedRequest);
    if (response instanceof Error) { // Unsure if idiomatic.
      callback(response);
    } else {
      callback(null, response);
    }
  };
}
```

When mocking the grpc layer for a regular test:
```js
// Non-error testing
client._getBook = mockGrpcMethod(expectedRequest, expectedResponse);
```

And for an error:
```js
// This will only need to be declared once.
var error = new Error('Fake error');
error.code = FAKE_STATUS_CODE;

// Error testing
client._getBook = mockGrpcMethod(expectedRequest, error);
```

Something similar can be made for streaming as well.

Note: Theres a line in my example that I am unsure if it is idiomatic. @jmuk can you take a quick look.

shinfan: Good point. I will have a try.

shinfan: I am not an export in NodeJs but I feel like

```
function mockGrpcMethod(expectedRequest, response, error) {
  return function(actualRequest, options, callback) {
    assert.deepStrictEqual(actualRequest, expectedRequest);
    if (error) {.
      callback(error);
    } else {
      callback(null, response);
    }
  };
}
```

might be better

src/test/java/com/google/api/codegen/testdata/nodejs_test_library.baseline
```diff
@@ -137,6 +208,29 @@ describe('LibraryServiceClient', function() {
...
+
+      // Mock Grpc layer
+      client._deleteShelf = function(actualRequest, options, callback) {
+        assert.deepStrictEqual(actualRequest, request);
+        var error = new Error('Fake error');
+        error.code = FAKE_STATUS_CODE;
+        callback(error);
+      };
+
+      client.deleteShelf(request, function(err, response) {
```
jmuk: `deleteShelf` has `Empty` return type, which should make the callback style as `function(err) { ... }`

src/test/java/com/google/api/codegen/testdata/nodejs_test_library.baseline
```diff
@@ -592,6 +1009,33 @@ describe('LibraryServiceClient', function() {
...
+        return mockStream;
+      };
+
+      client.streamShelves(request).on('data', function(response) {
+        assert.fail();
+      }).on('error', function(err) {
+        assert(err instanceof Error);
+        assert.equal(err.code, FAKE_STATUS_CODE);
+        done();
+        }
```
jmuk: This close brace seems redundant.

src/test/java/com/google/api/codegen/testdata/nodejs_test_library.baseline
landrito: I was trying to figure out which of the two were better as well. I went with mine to try and avoid this:

```js
// Looks a little ugly.
client._method = mockGrpcMethod(expectedRequest, null, error);
```

Maybe @jmuk can shed the light on whats best.

jmuk: I don't have strong feeling here, both would look good to me.  To me, skipping expected response as `null` isn't very bad.

landrito: @shinfan lets just go with 
```js 
function mockGrpcMethod(expectedRequest, response, error)
```

shinfan: sgtm

src/test/java/com/google/api/codegen/testdata/nodejs_test_library.baseline
```diff
@@ -749,17 +1117,34 @@ describe('LibraryServiceClient', function() {
...
+
+      client.addLabel(request, function(err, response) {
+        assert(err instanceof Error);
+        assert.equal(err.code, FAKE_STATUS_CODE);
+        done();
+      });
+    });
   });
 
   describe('getBigBookAsync', function() {
```
landrito: What is causing this to be `getBigBookAsync`. Shouldn't it be [`getBigBook`](https://github.com/googleapis/toolkit/blob/master/src/test/java/com/google/api/codegen/testdata/nodejs_doc_main_library.baseline#L1782)?

Furthermore, we need to do something different for longrunning methods since they will be wrapped with this [wrapper](https://github.com/googleapis/gax-nodejs/blob/master/lib/longrunning.js#L134).

src/test/java/com/google/api/codegen/testdata/nodejs_test_library.baseline
shinfan: Done

src/test/java/com/google/api/codegen/testdata/nodejs_test_library.baseline
shinfan: Fixed the async name.
Long running methods are not in the scope of this PR. Will fix in a separate one: https://github.com/googleapis/toolkit/issues/925

shinfan:
@landrito  PTAL

src/test/java/com/google/api/codegen/testdata/nodejs_test_library.baseline
```diff
@@ -749,21 +1117,38 @@ describe('LibraryServiceClient', function() {
...
+        done();
+      });
+    });
   });
 
-  describe('getBigBookAsync', function() {
-    it('invokes getBigBookAsync without error', function(done) {
+  describe('getBigBook', function() {
+    it('invokes getBigBook without error', function(done) {
       var client = libraryV1.libraryServiceClient();
```
landrito: This response of the method will be [wrapped](https://github.com/googleapis/gax-nodejs/blob/master/lib/longrunning.js#L94) by an operation so simply checking the response will not work. 

Maybe the most appropriate way would be a simple check like such:

```js
client.getBigBook(request, function(err, operation, response) {
    assert.ifError(err);
    assert.deepStrictEqual(operation.latestResponse, expectedResponse);
    assert.deepStrictEqual(response, expectedResponse);
    done();
  });
});
```

src/test/java/com/google/api/codegen/testdata/nodejs_test_library.baseline
```diff
@@ -749,21 +1117,38 @@ describe('LibraryServiceClient', function() {
...
       // Mock request
       var formattedName = client.bookPath("[SHELF_ID]", "[BOOK_ID]");
```
landrito: Not for this line but for the creation of the response. It looks like the response is not in the form of an [`google.longrunning.Operation`](https://github.com/googleapis/googleapis/blob/master/google/longrunning/operations.proto#L82). Any idea why?

shinfan: The mock response generation logic is shared across languages. I have looked at [that generation code](https://github.com/googleapis/toolkit/blob/master/src/main/java/com/google/api/codegen/transformer/TestCaseTransformer.java#L180), it looks like NodeJS may require some extra work for LRO. On the other hand, we may also need to update the sample gen code as well. I have an on-going LRO PR that I will send you for review as soon as it is ready.

landrito: Awesome SGTM.

src/test/java/com/google/api/codegen/testdata/nodejs_test_library.baseline
shinfan: I have skipped the LRO tests for now since they are incorrect anyway. I will add the tests back once the LRO issues are solved.

landrito:
LGTM pending CI failures.

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -596,7 +596,7 @@ public String getAsyncApiMethodName(Method method, VisibilityConfig visibility)
     return getApiMethodName(Name.upperCamel(method.getSimpleName()).join("async"), visibility);
   }
 
-  private String getApiMethodName(Name name, VisibilityConfig visibility) {
+  protected String getApiMethodName(Name name, VisibilityConfig visibility) {
```
landrito: Why did this happen?

shinfan: `getApiMethodName` is now used by `NodeJSSurfaceName`

landrito: Oh gotcha LGTM.

landrito:
LGTM

PR #922, opened 2017-01-12
implement package-private methods for Java

pongad:


src/test/java/com/google/api/codegen/testdata/java_main_library.baseline
```diff
@@ -2077,7 +2077,7 @@ public class LibraryServiceClient implements AutoCloseable {
    * @param label REQUIRED: The label to add.
    * @throws com.google.api.gax.grpc.ApiException if the remote call fails
    */
-  public final AddLabelResponse addLabel(String resource, String label) {
+   final AddLabelResponse addLabel(String resource, String label) {
```
pongad: The extra blank space here is OK. It will be stripped by the code formatter when we generate clients "for real"

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -588,12 +588,38 @@ public String getReroutedGrpcMethodName(MethodConfig methodConfig) {
...
+    }
+  }
+
+  /** The keyword controlling the visiblity, eg "public", "protected". */
+  public String getVisiblityKeyword(VisibilityConfig visibility) {
+    switch (visibility) {
+      case PUBLIC:
+        return "public";
+      case PACKAGE:
+        return "";
```
garrettjonesgoogle: alternatively, you could use `"/* package-private */"`

pongad: good idea, done

pongad:
@garrettjonesgoogle PTAL

garrettjonesgoogle:
LGTM

PR #923, opened 2017-01-12
Add streaming support to PHP

michaelbausor:
Enable streaming support for client, server and bidi streaming

Gax PR: https://github.com/googleapis/gax-php/pull/51


src/main/resources/com/google/api/codegen/php/main.snip
```diff
@@ -232,6 +233,25 @@
...
 @end
 
+@private grpcStreamingDescriptorFunction(xapiClass)
+    @if xapiClass.hasGrpcStreamingMethods
+        private static function getGrpcStreamingDescriptors()
+        {
+            return [
+                @join descriptor : xapiClass.grpcStreamingDescriptors
+                    '{@descriptor.methodName}' => [
+                        'grpcStreamingType' => '{@descriptor.grpcStreamingType.toString}',
```
garrettjonesgoogle: You don't need the `.toString`, the snippet language will do that automatically.

src/main/java/com/google/api/codegen/viewmodel/OptionalArrayMethodView.java
```diff
@@ -58,11 +58,18 @@
...
   public abstract LongRunningOperationDetailView longRunningView();
 
+  public boolean isLongRunningOperation() {
+    return longRunningView() != null;
+  }
+
+  public boolean isNotGrpcRequestStreamingMethod() {
+    return !(grpcStreamingType().equals(GrpcStreamingType.BidiStreaming)
+        || grpcStreamingType().equals(GrpcStreamingType.ClientStreaming));
+  }
```
garrettjonesgoogle: Since the logic is non-trivial here, I think I would prefer to have it in the transformer. Also, maybe the concept should be more general: `isSingularRequestMethod` or something like that.

src/main/java/com/google/api/codegen/viewmodel/DynamicLangXApiView.java
```diff
@@ -50,6 +50,8 @@
 
   public abstract List<LongRunningOperationDetailView> longRunningDescriptors();
 
+  public abstract List<GrpcStreamingDetailView> grpcStreamingDescriptors();
+
```
garrettjonesgoogle: Is PHP the first dynamic MVVM language with grpc streaming support?

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
```diff
@@ -178,6 +181,29 @@ public PhpGapicSurfaceTransformer(ApiConfig apiConfig, GapicCodePathMapper pathM
...
+      result.add(
+          GrpcStreamingDetailView.newBuilder()
+              .methodName(context.getNamer().getApiMethodName(method, VisibilityConfig.PUBLIC))
+              .grpcStreamingType(grpcStreamingConfig.getType())
+              .grpcResourcesField(
+                  grpcStreamingConfig.hasResourceField()
+                      ? context
+                          .getNamer()
+                          .getFieldGetFunctionName(grpcStreamingConfig.getResourcesField())
+                      : null)
```
garrettjonesgoogle: Maybe pull up this embedded logic

src/main/resources/com/google/api/codegen/php/method_sample.snip
```diff
@@ -83,9 +95,67 @@
...
+    $requests = [$request];
+    
+    // Write data to server and wait for a response
+    $streamingResponse = {@methodCallSampleCode(apiMethod)};
+    $result = $streamingResponse->writeAllAndWait($requests);
+    // doSomethingWith($result)
+    
+    // OR write data as it becomes available, then wait for a response
+    $streamingResponse = {@methodCallSampleCode(apiMethod)};
+    foreach (read_requests_from_stream() as $request) {
```
garrettjonesgoogle: yeah, the `$requests` array would be much clearer.

I think we should test what we have written, so if you are adding support for client streaming, add a baseline.

src/main/resources/com/google/api/codegen/php/method_sample.snip
```diff
@@ -83,9 +95,67 @@
     }
 @end
 
+@private clientStreamingMethodSampleCode(apiMethod)
+    $requests = [$request];
+    
+    // Write data to server and wait for a response
+    $streamingResponse = {@methodCallSampleCode(apiMethod)};
+    $result = $streamingResponse->writeAllAndWait($requests);
```
garrettjonesgoogle: The confusing part is "response" actually. This comment also applies to the types in GAX.

michaelbausor: What do you think about `$clientStream`, `$serverStream` and `$bidiStream`? I am not fully convinced that is right though, as they aren't really streams... maybe `$clientStreamConnection`? `$clientStreamingConnection`?

src/main/resources/com/google/api/codegen/php/method_sample.snip
michaelbausor: Updated to `$requests` and added client streaming method... although perhaps I should make the client streaming method part of a separate PR, as it changes multiple baselines in 6 languages, what do you think?

src/test/java/com/google/api/codegen/testdata/library_gapic.yaml
```diff
@@ -421,6 +421,14 @@ interfaces:
     retry_codes_name: non_idempotent
     retry_params_name: default
     timeout_millis: 30000
+  - name: MonologAboutBook
+    # required_fields makes difference on sample code.
```
garrettjonesgoogle: -> "makes a difference"

src/main/resources/com/google/api/codegen/php/method_sample.snip
garrettjonesgoogle: Maybe a name that corresponds to the type? So for streaming pull, `$pullStream` might be nice. Or if you just wanted to be generic, `$callStream`.

src/main/resources/com/google/api/codegen/php/method_sample.snip
garrettjonesgoogle: It seems inherent to the functionality you are enabling here, so I would say keep it in this PR.

src/test/java/com/google/api/codegen/testdata/java_test_library.baseline
```diff
@@ -1220,6 +1220,50 @@ public class LibraryServiceTest {
...
+      .setUserName(userName)
+      .setComment(comment)
+      .build();
+    mockLibraryService.addResponse(expectedResponse);
+    String name = "name3373707";
+    DiscussBookRequest request = DiscussBookRequest.newBuilder()
+      .setName(name)
+      .build();
+
+    $unhandled case: ClientStreaming$
```
garrettjonesgoogle: alert alert

src/test/java/com/google/api/codegen/testdata/nodejs_test_library.baseline
```diff
@@ -1001,6 +1001,8 @@ describe('LibraryServiceClient', function() {
     });
   });
 
+  $unhandled case: ClientStreaming$
```
garrettjonesgoogle: alert alert

src/test/java/com/google/api/codegen/testdata/php_main_library.baseline
```diff
@@ -1410,6 +1440,214 @@ class LibraryServiceClient
...
+     *     $clientStreamingResponse = $libraryServiceClient->monologAboutBook();
+     *     $result = $clientStreamingResponse->writeAllAndWait($requests);
+     *     // doSomethingWith($result)
+     *
+     *     // OR write data as it becomes available, then wait for a response
+     *     $clientStreamingResponse = $libraryServiceClient->monologAboutBook();
+     *     foreach ($requests as $request) {
+     *         $clientStreamingResponse->write($request);
+     *     }
+     *     $result = $clientStreamingResponse->wait();
```
garrettjonesgoogle: The name `wait` doesn't imply to me that it will return the final response

src/main/resources/com/google/api/codegen/php/method_sample.snip
michaelbausor: I like `$pullStream`, and perhaps also `$pushStream` and `$bidiStream`?

As for the GAX types, they could be named similarly.

garrettjonesgoogle: That's what I was thinking.

garrettjonesgoogle: Oh and as for your first statement - I suggested `pullStream` because the API method is `pull`. For the Speech API, the corresponding name would be `recognizeStream`.

michaelbausor: The variable names could be method-specific, but the GAX types are the same for all methods, so they need to be named for the streaming type

src/test/java/com/google/api/codegen/testdata/php_main_library.baseline
michaelbausor: Changed to readResponse()

michaelbausor:
Updated to skip test generation for client streaming in Java and NodeJS, which were causing the unhandled cases

src/test/java/com/google/api/codegen/testdata/library_gapic.yaml
michaelbausor: Fixed

michaelbausor:
PTAL

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
```diff
@@ -272,6 +273,9 @@ private ClientTestFileView createUnitTestFileView(SurfaceTransformerContext cont
     for (Method method : context.getSupportedMethods()) {
       MethodConfig methodConfig = context.getMethodConfig(method);
       if (methodConfig.isGrpcStreaming()) {
+        if (methodConfig.getGrpcStreamingType() == GrpcStreamingType.ClientStreaming) {
+          continue;
+        }
```
garrettjonesgoogle: Add a TODO & issue (for the Node one too) - I don't want to lose this.

garrettjonesgoogle:
LGTM after the TODO is added

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTestTransformer.java
michaelbausor: Added TODOs and created issue: https://github.com/googleapis/toolkit/issues/946

garrettjonesgoogle:
LGTM

PR #924, opened 2017-01-12
Port Python Sphinx doc files from Packman

### LINKED ISSUE #920: Port Python Sphinx configuration from Packman
### What

The [Sphinx configuration templates](https://github.com/googleapis/packman/tree/master/templates/gax/python/docs) previously generated by Packman need to be ported to `toolkit`, since the Python code generation pipeline no longer uses Packman.

### How

Translate the `.mustache` files to `.snip` files. Note that it will likely be necessary to make some changes to the [transformer](https://github.com/googleapis/toolkit/blob/baa2d3f832ddf8109577c9159a2ab1785de8ea03/src/main/java/com/google/api/codegen/transformer/py/PythonPackageMetadataTransformer.java)/ViewModel in order to pass the list of GAPIC Python modules to [the `index.rst` template](https://github.com/googleapis/packman/blob/master/templates/gax/python/docs/index.rst.mustache); this list of modules is used by Sphinx autosummary.

geigerj:
This is a translation of the mustache templates here:

  https://github.com/googleapis/packman/tree/master/templates/gax/python/docs

to SnippetSet templates.

Note that these templates require knowledge of the Python modules produced
during earlier stages of GAPIC generation. To do so, the PR passes the
`GapicProvider`s for these stages directly to the PackageMetadataTransformer,
and calls `generate()` on those providers to determine their output. After
Python is on MVVM, we can avoid actually running `generate()` twice by passing
the relevant transformers, and get the output module filenames from the
ViewModels they produce without actually rendering the templates twice.

Fixes #920


geigerj:
@lukesneeringer @shinfan PTAL

src/test/java/com/google/api/codegen/testdata/python_doc_docs_conf_library.baseline
```diff
@@ -0,0 +1,140 @@
...
+
+
+todo_include_todos = True
+
+
+
+html_theme = 'sphinx_rtd_theme'
+
+
+
```
geigerj: Ah, whoops. The `#` are comments in SnippetSet templates but not in mustache. I will fix this.

src/main/java/com/google/api/codegen/viewmodel/metadata/PackageMetadataView.java
```diff
@@ -75,8 +75,15 @@ public String resourceRoot() {
 
   public abstract boolean hasMultipleServices();
 
+  // Python-specific configuration
   @Nullable
-  public abstract List<String> namespacePackages();
+  public abstract List<String> pythonNamespacePackages();
```
geigerj: I assume this isn't the first time language-specific fields have popped up in MVVM viewmodels, but I'm not familiar enough with MVVM to point to another case. Is there are canonical solution to this?

(I might add a field `LanguagePackageMetadataView languageFields` to `PackageMetadataView` where `LanguagePackageMetadataView` is an empty interface implemented by `PythonPackageMetadataView`?)

garrettjonesgoogle: The canonical solution is to try and come up with a general name for the feature, so that if there is another language that implements a logically equivalent feature, that language could use the same field.

src/main/java/com/google/api/codegen/transformer/py/PythonPackageMetadataTransformer.java
```diff
@@ -69,23 +88,60 @@ public PythonPackageMetadataTransformer(PackageMetadataConfig packageConfig) {
...
         .build();
   }
 
+  /** Determines the Python files generated in the main phase of generation. */
+  private void computeModules(List<GapicProvider<? extends Object>> gapicProviders) {
+    // Only run generation once.
+    if (apiModules != null && typeModules != null) {
+      return;
+    }
+    Set<String> apiModuleSet = new TreeSet<>();
```
geigerj: Done.

I think I was somehow concerned that different providers could produce duplicate output files, but if that ever happens, that's a larger issue to address...

src/main/java/com/google/api/codegen/viewmodel/metadata/PackageMetadataView.java
geigerj: Removed references to Python in these field names.

src/test/java/com/google/api/codegen/testdata/python_doc_docs_conf_library.baseline
geigerj: Done

geigerj:
@shinfan Could you take another look? I think I've addressed your comments.

shinfan:
LGTM.

PR #927, opened 2017-01-13
support manual doc at interface level

pongad:


garrettjonesgoogle:
LGTM - add a couple more people to look at it.

pongad:
@garrettjonesgoogle I added @jba . Please let me know if you want to add anyone else.

geigerj:
Code for this LGTM.

What is the motivation for this PR? Is there an existing case where we want to use this, and what are the language-specific docs that are going here?

garrettjonesgoogle:
Yes, we are adding this feature because of a new use case. For Pub/Sub in Java, we want to hide the `publish` message on the GAPIC client, and then inject a doc at the class level pointing the user to the handwritten class that provides the functionality.

PR #928, opened 2017-01-13
Add unit test for PHP LRO

michaelbausor:


src/main/java/com/google/api/codegen/transformer/TestCaseTransformer.java
```diff
@@ -67,7 +67,11 @@ public TestCaseView createTestCaseView(
               methodContext.getTypeTable(),
               methodConfig.getPageStreaming().getResourcesFieldConfig());
     } else if (methodConfig.isLongRunningOperation()) {
-      clientMethodName = namer.getAsyncApiMethodName(method, methodConfig.getVisibility());
+      if (clientMethodType == ClientMethodType.OperationOptionalArrayMethod) {
+        clientMethodName = namer.getApiMethodName(method, methodConfig.getVisibility());
+      } else {
+        clientMethodName = namer.getAsyncApiMethodName(method, methodConfig.getVisibility());
```
shinfan: I am not sure what the purpose of this else case. It looks confusing to me that the method name is "async" while it is LRO. Do we want to distinguish those two types of methods?

src/test/java/com/google/api/codegen/testdata/php_test_library.baseline
```diff
@@ -794,4 +796,73 @@ class LibraryServiceTest extends PHPUnit_Framework_TestCase
...
+        $read = true;
+        $expectedResponse = new Book();
+        $expectedResponse->setName($name2);
+        $expectedResponse->setAuthor($author);
+        $expectedResponse->setTitle($title);
+        $expectedResponse->setRead($read);
+        $anyResponse = new Any();
+        $anyResponse->setValue($expectedResponse->serialize());
+        $completeOperation = new Operation();
+        $completeOperation->setName('operations/getBigBookTest')->setDone(true)->setResponse($anyResponse);
```
shinfan: Do we want to call `setDone(true)` after the actual API call? It is more similar to the actual sequence.

src/test/java/com/google/api/codegen/testdata/php_test_library.baseline
```diff
@@ -794,4 +796,73 @@ class LibraryServiceTest extends PHPUnit_Framework_TestCase
...
+        $grpcStub->addResponse($completeOperation);
+
+        // Mock request
+        $formattedName = LibraryServiceClient::formatBookName("[SHELF_ID]", "[BOOK_ID]");
+
+        $response = $client->getBigBook($formattedName);
+        $response->pollUntilComplete();
+        $actualRequests = $grpcStub->getReceivedCalls();
+        $this->assertSame(1, count($actualRequests));
+        list($actualFuncCall, $actualRequestObject) = $actualRequests[0];
```
shinfan: Not directly related to the scope of this PR, but do you think it is idiom to return a list and explode the string with `/` in PHP? It looks not very readable to me. Maybe consider using a class?

src/main/resources/com/google/api/codegen/php/test.snip
```diff
@@ -102,27 +127,39 @@
...
+        
+    @end
+    {@singleCallSuccessAsserts(test)}
     @if test.hasReturnValue
         $this->assertEquals($expectedResponse, $response);
     @end
 @end
 
-@private pagedOptionalArraySuccessAsserts(test)
+@private pagedSuccessAsserts(test)
```
garrettjonesgoogle: This name is confusing - should it be something like `pagedMethodSuccessAsserts`?

src/main/resources/com/google/api/codegen/php/test.snip
```diff
@@ -102,27 +127,39 @@
...
-@private optionalArraySuccessAsserts(test)
+@private successAsserts(test)
+    @if test.hasReturnValue
+        $response = $client->{@test.clientMethodName}({@sampleMethodCallArgList(test.initCode.fieldSettings)});
+        
+    @else
+        $client->{@test.clientMethodName}({@sampleMethodCallArgList(test.initCode.fieldSettings)});
+        
+    @end
+    {@singleCallSuccessAsserts(test)}
```
garrettjonesgoogle: It might be more readable to have fewer ifs but slightly more repetition:

```
@if test.hasReturnValue
    $response = ...
    {@singleCallSuccessAsserts(test)}
    $this->assert...
@else
    $client->...
    {@singleCallSuccessAsserts(test)}
@end
```

Or is it necessary to have the `$expectedResponse` assertion after `singleCallSuccessAsserts`? If not:


```
@if test.hasReturnValue
    $response = ...
    $this->assert...
@else
    $client->...
@end

{@singleCallSuccessAsserts(test)}
```

src/main/resources/com/google/api/codegen/php/test.snip
```diff
@@ -62,30 +62,55 @@
...
+        @case "PagedOptionalArrayMethod"
+            $grpcStub->addResponse($expectedResponse);
+        @case "OperationOptionalArrayMethod"
+            $anyResponse = new Any();
+            $anyResponse->setValue($expectedResponse->serialize());
+            $completeOperation = new Operation();
+            $completeOperation->setName('operations/{@test.name}')->setDone(true)->setResponse($anyResponse);
+            $grpcStub->addResponse($completeOperation);
+        @default
+            $unhandled case: {@test.clientMethodType.toString}$
```
garrettjonesgoogle: `.toString` not necessary here (other places too)

src/main/java/com/google/api/codegen/transformer/TestCaseTransformer.java
michaelbausor: I agree - the else case is to continue supporting NodeJS and Java, which use getAsyncApiMethodName (as they did before this change). I had to introduce the check because PHP uses the standard getApiMethodName for LRO methods.

Perhaps I could add a new method, namer.getLroApiMethodName?

src/test/java/com/google/api/codegen/testdata/php_test_library.baseline
michaelbausor: The string being exploded by '/' is controlled by the grpc stub, for example for Pubsub create topic it is '/google.pubsub.v1.Publisher/CreateTopic' (see here: https://github.com/googleapis/proto-client-php/blob/master/src/pubsub/v1/pubsub.php#L3635). Maybe I should construct and assert this complete string instead of just the last component?

I could create a class instead of each acutalRequest being an array [$actualFuncCall, $actualRequestObject]

src/test/java/com/google/api/codegen/testdata/php_test_library.baseline
michaelbausor: We don't want to change setDone after creating the Operation, but maybe we want to first return an incomplete operation, and then a complete one. This requires a fair bit of extra work to mock out the Operations API call as well as the original API call, but maybe worth it.

src/main/resources/com/google/api/codegen/php/test.snip
michaelbausor: Fixed

src/main/resources/com/google/api/codegen/php/test.snip
michaelbausor: Fixed

src/main/resources/com/google/api/codegen/php/test.snip
michaelbausor: Fixed

garrettjonesgoogle:
LGTM for the issues I called out - let @shinfan also approve.

src/main/java/com/google/api/codegen/transformer/TestCaseTransformer.java
shinfan: `namer.getLroApiMethodName` sounds good

src/test/java/com/google/api/codegen/testdata/php_test_library.baseline
shinfan: I think constructing a class would be good for better readability, so you can have something like: `$record->actualFuncCall` And for the string, I agree that it makes more sense to assert as a while instead of calling explode.

src/test/java/com/google/api/codegen/testdata/php_test_library.baseline
shinfan: Umm the test would be more complete if we can make the grpc return a incomplete operation first, right? But it is hard to reason whether the benefit worth the cost from my angle. If this requires a lot of engineering time, maybe it is not worth the effort.  Can you open an issue to track this and describe what needs to be done?

src/test/java/com/google/api/codegen/testdata/php_test_library.baseline
michaelbausor: Updated, see also https://github.com/googleapis/gax-php/pull/57

src/test/java/com/google/api/codegen/testdata/php_test_library.baseline
michaelbausor: Updated to test the incomplete operation case

michaelbausor:
Added the incomplete operation case and restructured test.snip, PTAL

PR #929, opened 2017-01-13
Fix Python GAPIC package metadata output path

geigerj:
Previously, the template file's directory was mistakenly included in
the output path.


lukesneeringer:
As soon as tests pass, I would like to merge this. It fixes a problem that is a prerequisite to the `protos`- change.

geigerj:
Tests pass -- LGTY?

lukesneeringer:
Looks good to me. I asked @landrito to give it a quick once-over so there is a second reviewer.

landrito:
LGTM

PR #930, opened 2017-01-13
Change grpc- prefix to proto-.

lukesneeringer:
We are changing our `grpc-` packages not to have a strict GRPC dependency. To avoid naming confusion, we are changing the prefix on these Python packages to `proto-`, since that is now the key things they contain.

It must be co-merged alongside:

  * googleapis/artman#132
  * googleapis/googleapis#239


geigerj:
Think you'll need to resolve some merge conflicts due to https://github.com/googleapis/toolkit/commit/2b2fad7955f49d754df3e3ed5f5725fab7bf211e

src/test/java/com/google/api/codegen/grpcmetadatagen/testdata/api_defaults.yaml
```diff
@@ -30,6 +30,6 @@ semver:
   #       publishing some 1.0.x versions to pypi already. They are now
   #       deleted but files still exist on the pypi servers and can prevent
   #       us from using the version number again.
-  python: '0.13.0'
```
geigerj: I'd actually advocate not generally updating this testdata file because it makes it easier to read the diff in the golden files caused by the actual change to the code generator.

lukesneeringer: Actually, it ended up working kind of well, because I caught a change where I inaccurately bumped `oauthclient`.

src/main/java/com/google/api/codegen/transformer/PackageMetadataTransformer.java
```diff
@@ -39,6 +39,7 @@
         .protoVersionBound(packageConfig.protoVersionBound(language))
         .commonProtosVersionBound(packageConfig.commonProtosVersionBound(language))
         .authVersionBound(packageConfig.authVersionBound(language))
+        .packageName(packageConfig.packageName(language))
```
geigerj: This was superseded by `grpcPackageName` and `gapicPackageName`. Please rename the `grpcPackageName` field to `protoPackageName` and reuse it here.

lukesneeringer: I'm concerned about that. It would be easy to update to `protoPackageName`, but if `grpcPackageName` is being used by _anything else_ (note that this is _not_ a language-dependent transformer), then it will be a breaking change.

lukesneeringer: Alternatively, would you prefer a `protoPackageName` method _in addition to_ the `gapic` and `grpc` equivalents? This gets around my concern.

lukesneeringer: This is changed to:

    .packageName(packageConfig.packageName(language))
    .protoPackageName("proto-" + packageConfig.packageName(language))

geigerj: Hmm, I don't see that change; did you push the commit? (Also, is the `packageName` field of `PackageMetadataView` used anywhere now that `protoPackageName` is available?)

PR #931, opened 2017-01-13
Add Sheets v4 API discovery baseline

saicheems:


saicheems:
This will make it easier to see changes in the, adding optional params, CL

PR #932, opened 2017-01-17
Fix some template mistakes

geigerj:
This change corrects some superficial errors in the templates that either had not been fixable before the templates had been ported from `packman` to `toolkit`, or were introduced in the process of porting them.

(See [this PR](https://github.com/googleapis/api-client-staging/pull/154/commits/ecfc1ce6e0980186ec235e5744e00537f1b778cd) for an example of how the generated output is currently hand edited.)

### DIFF
--- a/src/main/resources/com/google/api/codegen/py/README.rst.snip
+++ b/src/main/resources/com/google/api/codegen/py/README.rst.snip
@@ -2,7 +2,7 @@
   GAPIC library for the {@metadata.fullName}
   ================================================================================
   
-  {@metadata.packageName} uses google-gax_ (Google API extensions) to provide an
+  gapic-{@metadata.packageName} uses google-gax_ (Google API extensions) to provide an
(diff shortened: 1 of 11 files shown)


eoogbe:
LGTM

src/main/resources/com/google/api/codegen/py/README.rst.snip
```diff
@@ -2,7 +2,7 @@
   GAPIC library for the {@metadata.fullName}
   ================================================================================
   
-  {@metadata.packageName} uses google-gax_ (Google API extensions) to provide an
+  gapic-{@metadata.packageName} uses google-gax_ (Google API extensions) to provide an
```
garrettjonesgoogle: I would much prefer you to have a string in the view model named something like `gapicPackageName` - this here amounts to the type of concatenation I would like to avoid.

geigerj: Ok, sure. See #934

PR #935, opened 2017-01-17
Derive APIs Explorer link from service config name

geigerj:
Previously, it was assumed this was the same as the artman "short
name" for the API, which is not necessarily true.


src/main/java/com/google/api/codegen/transformer/PackageMetadataTransformer.java
```diff
@@ -27,6 +27,14 @@
       String template,
       String outputPath,
       TargetLanguage language) {
+    // Note that internally, this is overridable using a "discovery" section in the service config,
```
ethanbao: discovery is not part of the public service config, so let's not mention it here.

src/main/java/com/google/api/codegen/transformer/PackageMetadataTransformer.java
```diff
@@ -27,6 +27,14 @@
...
+    // which is not available externally. See:
+    //   https://github.com/googleapis/toolkit/issues/933
+    String discoveryApiName = model.getServiceConfig().getName();
```
ethanbao: This is actually service name. And I would suggest naming the first segment as apiName without mentioning discovery.

src/main/java/com/google/api/codegen/transformer/PackageMetadataTransformer.java
```diff
@@ -47,6 +55,7 @@
         .homepage(packageConfig.homepage())
         .licenseName(packageConfig.licenseName())
         .fullName(model.getServiceConfig().getTitle())
+        .discoveryApiName(discoveryApiName)
```
ethanbao: need to replace dash with underscore.

src/main/java/com/google/api/codegen/transformer/PackageMetadataTransformer.java
geigerj: The name `apiName` is confusing in the context of GAPIC given that we already have `shortName` and `fullName` to refer to the API's name. I want to clarify that this value should be used where the name of the discovery doc is needed.

src/main/java/com/google/api/codegen/transformer/PackageMetadataTransformer.java
geigerj: Done

src/main/java/com/google/api/codegen/transformer/PackageMetadataTransformer.java
geigerj: Done

src/main/java/com/google/api/codegen/transformer/PackageMetadataTransformer.java
ethanbao: Ack

ethanbao:
LGTM

PR #942, opened 2017-01-19
Support request body overrides in Java

saicheems:


src/main/java/com/google/api/codegen/discovery/transformer/java/JavaSampleTypeNameConverter.java
```diff
@@ -108,12 +97,25 @@ public TypeName getTypeNameForElementType(TypeInfo typeInfo) {
...
+      TypeName elementTypeName = getTypeNameForElementType(typeInfo, true);
+      return new TypeName(
+          listTypeName.getFullName(), listTypeName.getNickname(), "%s<%i>", elementTypeName);
+    }
+    return getTypeNameForElementType(typeInfo, false);
   }
 
-  private TypeName getNonMessageTypeName(TypeInfo typeInfo) {
+  @Override
+  public TypeName getTypeNameForElementType(TypeInfo typeInfo) {
```
tcoffee-google: With this refactoring, the meaning of this function has changed; how about `getTypeNameForNonRepeatedType`?

src/main/resources/com/google/api/codegen/java/sample.snip
```diff
@@ -64,6 +62,9 @@
       @end
 
       {@class.requestBodyTypeName} {@class.requestBodyVarName} = new {@class.requestBodyTypeName}();
+      @join field : class.requestBodyFields if class.requestBodyFields
```
tcoffee-google: Isn't this more properly written as an `@if` clause with `@join` inside, like the one added above? (I had understood `@join ... if` to be used for predicating on the iteration variable.

src/main/resources/com/google/api/codegen/java/sample.snip
```diff
@@ -73,6 +74,9 @@
     @end
     {@class.serviceTypeName} {@class.serviceVarName} = {@class.createServiceFuncName}();
     {@class.requestTypeName} {@class.requestVarName} = {@class.serviceVarName}.{@callChain(class.methodNameComponents)}({@argList(class.fieldVarNames)});
+    @join field : class.optionalFields if class.optionalFields
```
tcoffee-google: And similarly here.

src/test/java/com/google/api/codegen/testdata/discoveries/sheets.v4.json.overrides
```diff
@@ -1,12 +1,267 @@
 {
   "java": {
     "methods": {
+      "sheets.spreadsheets.batchUpdate": {
+        "requestBodyType": {
```
tcoffee-google: Seems like it would be much better if the overrides could just specify names of fields to include, and their properties would be determined from discovery doc. Otherwise API authors writing overrides need to be intimately familiar with our internals to enter all this metadata, some of it redundant even within itself.

src/main/java/com/google/api/codegen/discovery/config/ApiaryConfigToSampleConfigConverter.java
```diff
@@ -88,13 +88,21 @@ private MethodInfo createMethod(Method method) {
...
       // request body.
       if (fieldName.equals(DiscoveryImporter.REQUEST_FIELD_NAME)) {
         requestBodyType = createTypeInfo(field, method);
+
+        Type type = apiaryConfig.getType(field.getTypeUrl());
+        for (Field field2 : type.getFieldsList()) {
+          if (field2.getName().equals(PAGE_TOKEN_FIELD_NAME)) {
+            isPageStreamingResourceSetterInRequestBody = true;
+          }
+        }
```
saicheems: Sure, feel free to merge #880 whenever, I don't think there's anything else coming besides this

src/main/resources/com/google/api/codegen/java/sample.snip
saicheems: I actually added the `@join ... if` syntax to snippet language, it's semantically equivalent to
```
@if ...
  @join ...
  @end
@end
```

Unfortunately... in certain contexts the whitespace rules break down, so I fully expand the loops in those cases. I prefer where it works to use the short `@join if` syntax

saicheems: Specifically, the reason `@join if` was added was because a `@join` where no elements are available to iterate over prints an empty new line, whereas an `@join if` does not.

src/test/java/com/google/api/codegen/testdata/discoveries/sheets.v4.json.overrides
saicheems: This is true... I had considered both originally, and I am not opposed to doing so, but I figured I'd do the easy way first since it requires no extra work. Since no other APIs have asked yet, are you okay with deferring what you're suggesting for some time?

The method you're suggesting is an extension on what we're doing here that requires recursively exploring both non-required parameters and request body types, neither of which are implemented in `DiscoveryImporter`/`ApiaryConfigToSampleConfigConverter`. I can open a bug to track that feature, I wanted to do it eventually but haven't gotten to it.

src/main/java/com/google/api/codegen/discovery/transformer/java/JavaSampleMethodToViewTransformer.java
```diff
@@ -85,19 +85,29 @@ private SampleView createSampleView(SampleTransformerContext context) {
...
     for (FieldInfo field : methodInfo.fields().values()) {
-      SampleFieldView sampleFieldView = createSampleFieldView(field, typeTable, symbolTable);
+      SampleFieldView sampleFieldView = createSampleFieldView(field, context, symbolTable);
       fields.add(sampleFieldView);
-      fieldVarNames.add(sampleFieldView.name());
+      if (sampleFieldView.required()) {
+        fieldVarNames.add(sampleFieldView.name());
+      } else {
+        optionalFields.add(sampleFieldView);
+      }
```
saicheems: Probably the right move here is to fix the names,

`fields` -> `requiredFields`
`fieldVarNames` -> `apiCallArgs`
`optionalFields` -> `optionalFields`

does that seem reasonable?

tcoffee-google: I wonder if it would make sense to call these "parameters" like the API docs, instead of "fields" like in protobuf, which has always seemed confusing to me ... for instance,

* `fields` -> `requiredParams`
* `fieldVarNames` -> `methodCallParams`
* `optionalFields` -> `optionalParams`.

But if that's too inconsistent with other usage, I think your suggestion is fine.

src/main/resources/com/google/api/codegen/java/sample.snip
tcoffee-google: Okay, thanks for clarifying. I had thought it looked like a comprehension syntax.

src/test/java/com/google/api/codegen/testdata/discoveries/sheets.v4.json.overrides
tcoffee-google: The risk of deferring is if another API team wants to override and sees this as the reference way to do it, such that either (a) they do it this way and we have more trouble changing it later, or (b) they decide not to use it because it's too complicated. We could probably avoid that with some care in communications. @vchudnov-g your thoughts?

vchudnov-g: Sai and I talked last week and decided to do the quick approach for now until we get an idea of how prevalent the issue is. Let's stick with that.

You're right that we should be careful in communications. The way I see that part working is that we'll try to generate samples for a team, they'll fail, we'll see this is needed, we'll ask the team to provide this, and we'll say we're working to simplify this mechanism if it turns out to be common.

src/main/java/com/google/api/codegen/discovery/transformer/java/JavaSampleTypeNameConverter.java
saicheems: I think this name is still consistent (it was less appropriate before...) in that for any type you pass it in it returns the TypeName of its element.

A map type -> a map entry
An array type - > an array entry
A message -> a message
A primitive -> a boxed primitive

saicheems:
Rebased and ready

PR #943, opened 2017-01-20
Remove unused files

saicheems:


saicheems:
Many were removed by Michael in other PRs... I don't know if these are the last ones, but if there are any more I'll put up PRs when I see them.

PR #949, opened 2017-01-20
Change baseline Python docs to 0.15.0.

lukesneeringer:
Since we released the 0.15.0 releases of our tools this morning, the toolkit tests have started failing. I believe the version might be read from `googleapis`, which is now updated.

The ultimate fix is probably to make these independent, but for the moment, just updating the versions to make the tests pass again.

### DIFF
--- a/src/main/resources/com/google/api/codegen/py/docs/conf.py.snip
+++ b/src/main/resources/com/google/api/codegen/py/docs/conf.py.snip
@@ -317,5 +317,4 @@
   napoleon_use_ivar = False
   napoleon_use_param = True
   napoleon_use_rtype = True
-
 @end
--- a/src/test/java/com/google/api/codegen/testdata/python_doc_docs_conf_library.baseline
+++ b/src/test/java/com/google/api/codegen/testdata/python_doc_docs_conf_library.baseline
@@ -21,7 +21,7 @@ import shlex
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 sys.path.insert(0, os.path.abspath('..'))
 
-__version__ = '0.14.0'
+__version__ = '0.15.0'
 
 # -- General configuration ------------------------------------------------
 
@@ -317,4 +317,3 @@ napoleon_use_admonition_for_references = False
 napoleon_use_ivar = False
 napoleon_use_param = True
 napoleon_use_rtype = True
-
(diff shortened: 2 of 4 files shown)


src/main/resources/com/google/api/codegen/py/docs/conf.py.snip
```diff
@@ -317,5 +317,4 @@
   napoleon_use_ivar = False
   napoleon_use_param = True
   napoleon_use_rtype = True
-
```
geigerj: Also leave this in so that the rendered templates end in a newline.

geigerj:
Oh I see, that file is `0.15.0` at HEAD.

geigerj:
Would prefer @michaelbausor merge #948 instead of this one because #948 doesn't remove the trailing newline.

(Also, @lukesneeringer, would argue this is perhaps a reason not to update the testdata...)

src/main/resources/com/google/api/codegen/py/docs/conf.py.snip
lukesneeringer: Right now, the rendered templates end in _two_ newlines, which is why I removed this (so they would only end in one).

geigerj: Are you sure? If I recall correctly, you're seeing one newline from the rendered template, and one newline that's printed by the test itself. (Note that `*.baseline` is not a simple rendered template, but a list of newline-separated rendered templates.)

lukesneeringer: I am 100% positive. One sec, I can provide screenshots. :-)

michaelbausor:
I have merged https://github.com/googleapis/toolkit/pull/948, sorry for the confusion

src/main/resources/com/google/api/codegen/py/docs/conf.py.snip
lukesneeringer: `master`: screen shot 2017-01-20 at 12 54 45 pm

`0.15.0`: screen shot 2017-01-20 at 12 54 22 pm

lukesneeringer: This is true on all three `.baseline` files.

lukesneeringer: I tested this with artman, and @geigerj is correct and I am incorrect. The `.baseline` files do have two newlines, but one of them does not appear in the output files.

lukesneeringer:
Closed in favor of #948.

PR #951, opened 2017-01-20
Add PHP Exception tests

michaelbausor:
This PR looks like it changes a lot in the snip file, but it is only a minor restructuring that the git diff tool doesn't like for some reason. The existing tests are almost unchanged. Changes are:
- Added exception case tests
- Fixed bug in tests for methods without return values


src/main/resources/com/google/api/codegen/php/test.snip
```diff
@@ -67,128 +67,179 @@
...
+    @switch test.clientMethodType
+    @case "OptionalArrayMethod"
+        {@simpleTestCase(test)}
+        {@simpleExceptionTestCase(test)}
+    @case "PagedOptionalArrayMethod"
+        {@pageStreamingTestCase(test)}
+        {@simpleExceptionTestCase(test)}
+    @case "OperationOptionalArrayMethod"
+        {@lroTestCase(test)}
+        {@simpleExceptionTestCase(test)}
```
shinfan: Why is the LRO exception test the same as the others? I assume the test should be different because we should test the error handling of the operation object.

michaelbausor: Updated to throw exception from pollUntilComplete instead of initial call

src/test/java/com/google/api/codegen/testdata/php_test_library.baseline
```diff
@@ -151,6 +154,43 @@ class LibraryServiceClientTest extends PHPUnit_Framework_TestCase
...
+        $status->details = 'internal error';
+        $grpcStub->addResponse(null, $status);
+
+        // Mock request
+        $shelf = new Shelf();
+
+        try {
+            $client->createShelf($shelf);
+            // If the $client method call did not throw, fail the test
+            $this->assertTrue(false);
```
michaelbausor: Fixed

src/test/java/com/google/api/codegen/testdata/php_mock_service_library.baseline
```diff
@@ -23,15 +23,18 @@
 
 namespace Google\Tests\Example\Library\V1;
 
+use \Google\GAX\ApiException;
```
shinfan: Will these imports be cleaned up by the formatter?

src/test/java/com/google/api/codegen/testdata/php_test_library.baseline
```diff
@@ -152,6 +155,43 @@ class LibraryServiceClientTest extends PHPUnit_Framework_TestCase
...
+        try {
+            $client->createShelf($shelf);
+            // If the $client method call did not throw, fail the test
+            $this->fail("Expected an ApiException, but no exception was thrown.");
+        } catch (ApiException $ex) {
+            $this->assertEquals($status->code, $ex->getCode());
+            $this->assertEquals($status->details, $ex->getMessage());
+            $actualRequests = $grpcStub->getReceivedCalls();
+            $this->assertSame(1, count($actualRequests));
+            $actualFuncCall = $actualRequests[0]->getFuncCall();
```
shinfan: I feel this function call check is redundant, it has been checked in the previous test and the method call is the same for the exception test. Therefore we should just focus on the exception handling part.

src/test/java/com/google/api/codegen/testdata/php_mock_service_library.baseline
michaelbausor: Yes, imports will be cleaned up by the formatter

src/test/java/com/google/api/codegen/testdata/php_test_library.baseline
michaelbausor: Updated to remove these checks

michaelbausor:
@shinfan Finally came back to this PR, sorry for the ridiculous delay.

PR #953, opened 2017-01-23
Update discovery files and Ruby names yaml

saicheems:


saicheems:
Need to update the baselines too, will add those in a sec

PR #954, opened 2017-01-23
Support request body overrides in C#

saicheems:


src/test/java/com/google/api/codegen/testdata/discoveries/csharp/csharp_sheets.v4.json.baseline
```diff
@@ -197,10 +211,14 @@ namespace SheetsSample
...
 
+            // The ID of the spreadsheet to copy the sheet to.
+            string destinationSpreadsheetId = "";  // TODO: Update placeholder value.
+
             // TODO: Assign values to desired properties of `requestBody`:
             Data.CopySheetToAnotherSpreadsheetRequest requestBody = new Data.CopySheetToAnotherSpreadsheetRequest();
+            requestBody.DestinationSpreadsheetId = destinationSpreadsheetId;
 
-            SpreadsheetsResource.SheetsResource.CopyToRequest request = sheetsService.Spreadsheets.Sheets.CopyTo(requestBody, spreadsheetId, sheetId);
+            Spreadsheets.SheetsOperations.CopyTo request = sheetsService.Spreadsheets.Sheets.CopyTo(requestBody, spreadsheetId, sheetId);
```
tcoffee-google: Is this change due to an API update or a samplegen change?

saicheems: This is wrong, it's a Java rule that got copied over in the overrides file, it was right as it was.

Didn't actually want to start review yet, I put this pull up very prematurely, I'll tag you guys when I'm ready.

saicheems:
Ready now, not too much change from your last review, just shifting some bits out of `TypeNameGenerator` into the `CSharpSampleTypeNameConverter` and getting compilecheck working.

src/test/java/com/google/api/codegen/testdata/discoveries/sheets.v4.json.overrides
```diff
@@ -1,4 +1,15 @@
 {
+  "java": {
+    "methods": {
+      "sheets.spreadsheets.sheets.copyTo": {
+        "requestType": {
+          "message": {
+            "typeName": "Spreadsheets.SheetsOperations.CopyTo"
```
tcoffee-google: Why do we need this one override in Java?

src/main/java/com/google/api/codegen/discovery/transformer/csharp/CSharpSampleTypeNameConverter.java
```diff
@@ -71,19 +75,47 @@ public TypeName getServiceTypeName(String apiTypeName) {
...
   }
 
   @Override
   public TypeName getTypeName(TypeInfo typeInfo) {
-    if (typeInfo.isMessage()) {
-      return new TypeName(typeInfo.message().typeName());
+    if (typeInfo.isMap()) {
+      TypeName mapTypeName = typeNameConverter.getTypeName("System.Collections.Generic.Dictionary");
+      TypeName keyTypeName = getTypeNameForElementType(typeInfo.mapKey());
+      TypeName valueTypeName = getTypeNameForElementType(typeInfo.mapValue());
```
tcoffee-google: What happens if the value type of a map is a map? It seems like this would write it as a KeyValuePair instead.

src/main/java/com/google/api/codegen/discovery/transformer/csharp/CSharpSampleTypeNameConverter.java
```diff
@@ -71,19 +75,47 @@ public TypeName getServiceTypeName(String apiTypeName) {
...
   @Override
   public TypeName getRequestTypeName(String apiTypeName, TypeInfo typeInfo) {
-    return getTypeName(typeInfo);
+    List<String> copy = new ArrayList<>(methodNameComponents);
+    String requestTypeName = copy.remove(copy.size() - 1) + "Request";
+    String path = "";
+    for (String s : copy) {
+      path += s + "Resource.";
+    }
+    return new TypeName(path + requestTypeName);
```
tcoffee-google: Is this due to a change in the C# generator?

src/test/java/com/google/api/codegen/testdata/discoveries/sheets.v4.json.overrides
saicheems: The default name generated is `Sheets.Spreadsheets.Sheets.CopyTo`, which is wrong for whatever reason... I haven't had time to figure out the rule for this case. The actual type should be `Sheets.Spreadsheets.SheetsOperations.CopyTo`

src/main/java/com/google/api/codegen/discovery/transformer/csharp/CSharpSampleTypeNameConverter.java
saicheems: It's not really a change, I just moved the logic that decides these names from the `CSharpTypeNameGenerator`. The behavior is the exact same. I've been whittling down the language specific type name generators bit by bit and moving their logic to the converters.

src/main/java/com/google/api/codegen/discovery/transformer/csharp/CSharpSampleTypeNameConverter.java
saicheems: Yeah...... good point, I've changed it to be able to nest maps. But still, hope we never see that in production!

saicheems:
Any further thoughts?

src/main/java/com/google/api/codegen/discovery/transformer/java/JavaSampleTypeNameConverter.java
```diff
@@ -106,6 +106,13 @@ public TypeName getTypeName(TypeInfo typeInfo) {
     return getTypeNameForElementType(typeInfo, false);
   }
 
+  private TypeName getTypeName(TypeInfo typeInfo, boolean shouldBoxPrimitives) {
+    if (typeInfo.isMap() || typeInfo.isArray()) {
+      return getTypeName(typeInfo);
+    }
+    return getTypeNameForElementType(typeInfo, shouldBoxPrimitives);
+  }
+
```
tcoffee-google: Isn't it simpler to write the two functions above as:
```java
      return getTypeName(typeInfo, false)
   }

   private TypeName getTypeName(TypeInfo typeInfo, boolean shouldBoxPrimitives) {
      if (typeInfo.isMap()) {
        TypeName mapTypeName = typeNameConverter.getTypeName("java.util.Map");
        TypeName keyTypeName = getTypeName(typeInfo.mapKey(), true);
        TypeName valueTypeName = getTypeName(typeInfo.mapValue(), true);
        return new TypeName(
            mapTypeName.getFullName(),
            mapTypeName.getNickname(),
           "%s<%i, %i>",
           keyTypeName,
           valueTypeName);
     }
     if (typeInfo.isArray()) {
       TypeName listTypeName = typeNameConverter.getTypeName("java.util.List");
       TypeName elementTypeName = getTypeNameForElementType(typeInfo, true);
       return new TypeName(
           listTypeName.getFullName(), listTypeName.getNickname(), "%s<%i>", elementTypeName);
     }
      return getTypeNameForElementType(typeInfo, shouldBoxPrimitives);
    }
```

saicheems: Yes, good call. Done

src/main/java/com/google/api/codegen/discovery/transformer/csharp/CSharpSampleTypeNameConverter.java
```diff
@@ -71,33 +75,59 @@ public TypeName getServiceTypeName(String apiTypeName) {
...
+          mapTypeName.getFullName(),
+          mapTypeName.getNickname(),
+          "%s<%i, %i>",
+          keyTypeName,
+          valueTypeName);
     }
-    return getNonMessageTypeName(typeInfo);
+    if (typeInfo.isArray()) {
+      TypeName listTypeName = typeNameConverter.getTypeName("System.Collections.Generic.List");
+      TypeName elementTypeName = getTypeNameForElementType(typeInfo);
```
tcoffee-google: Er, don't we also need getTypeName here, if we have an array of arrays/maps?

saicheems: Just having `getTypeName` results in a stack overflow because whether or not a type is repeated is a property of the type, I figured it was more trouble than it was worth to refactor support for it at the moment because there is no field that can test it...

I filed #962 to track this, I don't think it's worth the trouble to solve right away
I can make an issue to track it, but I'm not sure it's worth spending time on right

tcoffee-google: Don't we just need a field like `typeInfo.arrayElement()` analogous to `typeInfo.mapKey()` and `typeInfo.mapValue()`? It feels bad that all these cases were supported back in the `DiscoveryContext` era but not anymore. While I don't recall yet seeing arrays of arrays, I think I have seen maps of maps.

I guess tracking the issue is fine for now.

saicheems: Yes, that's one of the solutions I was considering.
Give me until the end of this feature for the remaining languages and I'll do it right away

PR #956, opened 2017-01-24
C# gRPC duplex streaming, client & snippets

chrisdunelm:


chrisdunelm:
@jskeet: PTAL, baseline files updated, please can you check the generated C#.

src/main/resources/com/google/api/codegen/csharp/gapic_client.snip
```diff
@@ -108,41 +120,67 @@
...
+                    CallTiming.FromTimeout(TimeSpan.FromMilliseconds({@call.retryParamsView.totalTimeout.getMillis})));
+
+                /// <summary>
+                /// <see cref="BidirectionalStreamingSettings"/> for calls to
+                /// <c>{@api.name}.{@call.methodName}</c>.
+                /// </summary>
+                /// <remarks>
+                /// The default local send queue size is 100.
+                /// </remarks>
+                public BidirectionalStreamingSettings {@call.methodName}StreamingSettings { get; set; } =
```
jskeet: Presumably we want to parameterize this in the YAML. I think it's worth at least raising an issue so we can remember to do so.

PR #961, opened 2017-01-27
Support request body overrides in Go

saicheems:


src/main/resources/com/google/api/codegen/go/sample.snip
```diff
@@ -155,12 +164,42 @@
...
+
+
+  @end
+  {@field.name} := {@field.defaultValue}  // TODO: Update placeholder value.
+  @if field.example
+    // {@field.example}
+  @end
+@end
+
+@private callChain(class)
```
tcoffee-google: Can also include `{@class.ServiceVarName}` in this snippet. I'd prefer to call it something like `methodCall`.

src/main/resources/com/google/api/codegen/go/sample.snip
```diff
@@ -155,12 +164,42 @@
...
+  {@resourceCallChain(class)}({@paramList(class.methodCallFieldVarNames)}){@optionalFieldCallChain(class)}
+@end
+
+@private resourceCallChain(class)
```
tcoffee-google: This is not really a call; maybe just `methodResource`? Might make sense for the `{@class.ServiceVarName}` to go here.

src/main/resources/com/google/api/codegen/go/sample.snip
```diff
@@ -155,12 +164,42 @@
...
     {@resource}
   @end
 @end
 
+@private optionalFieldCallChain(class)
+  @join field : class.optionalFields on EMPTY
+    # Array types must be expanded if set on the request body.
+    # There's not really a good way to do this at the transformer level, so we do it here.
+    @if field.isArray
+      .{@field.setterFuncName}({@field.name}...)
```
tcoffee-google: Do I understand correctly that the client library implements array-field setters (and those alone) as variadics?

src/main/resources/com/google/api/codegen/go/sample.snip
```diff
@@ -155,12 +164,42 @@
...
+@end
+
+@private resourceCallChain(class)
+  @join resource : class.methodNameComponents on "."
     {@resource}
   @end
 @end
 
+@private optionalFieldCallChain(class)
+  @join field : class.optionalFields on EMPTY
```
tcoffee-google: Would it not be good style to line break and indent these (and avoid some very long lines)?

src/main/java/com/google/api/codegen/discovery/transformer/go/GoSampleTypeNameConverter.java
```diff
@@ -72,13 +72,9 @@ public TypeName getRequestTypeName(String apiTypeName, TypeInfo typeInfo) {
 
   @Override
   public TypeName getTypeName(TypeInfo typeInfo) {
-    if (typeInfo.isMessage()) {
-      String localName = GoSampleNamer.getServicePackageName(packagePrefix);
-      return new TypeName(packagePrefix + ";;;", localName + "." + typeInfo.message().typeName());
```
saicheems: The package prefix bit never really did anything, it can be used to make sure the type's package is imported, but we've already imported that package in the transformer

src/main/java/com/google/api/codegen/discovery/transformer/go/GoSampleTypeNameConverter.java
```diff
@@ -72,13 +72,9 @@ public TypeName getRequestTypeName(String apiTypeName, TypeInfo typeInfo) {
...
-    }
     if (typeInfo.isMap()) {
-      TypeName keyTypeName = getTypeNameForElementType(typeInfo.mapKey());
-      TypeName valueTypeName = getTypeNameForElementType(typeInfo.mapValue());
+      TypeName keyTypeName = getTypeName(typeInfo.mapKey());
+      TypeName valueTypeName = getTypeName(typeInfo.mapValue());
```
saicheems: Okay, I'll do it now

src/main/resources/com/google/api/codegen/go/sample.snip
saicheems: I believe so:
https://github.com/google/google-api-go-client/blob/master/google-api-go-generator/gen.go#L1652

src/main/resources/com/google/api/codegen/go/sample.snip
saicheems: Technically Go has no line length restrictions... it becomes a bit of a mess to get the snippet to do this properly

src/main/resources/com/google/api/codegen/go/sample.snip
saicheems: Done

src/main/resources/com/google/api/codegen/go/sample.snip
saicheems: Done

src/main/resources/com/google/api/codegen/go/sample.snip
tcoffee-google: Okay, maybe let's file an issue for this; it had been on my list since long ago to deal with line breaking more generally, for instance with long argument lists. The joining mechanisms in the snippet engine were designed with this in mind, so it's probably just a question of properly leveraging that machinery.

saicheems: #965

PR #966, opened 2017-01-27
Support request body overrides in Node.js

saicheems:


src/main/java/com/google/api/codegen/discovery/transformer/nodejs/NodeJSSampleMethodToViewTransformer.java
```diff
@@ -169,4 +182,19 @@ private SamplePageStreamingView createSamplePageStreamingView(
     builder.nextPageTokenName(methodInfo.responsePageTokenName());
     return builder.build();
   }
+
+  private SampleFieldView createSampleFieldView(
+      FieldInfo field, SampleTypeTable typeTable, boolean escapeReservedParamNames) {
+    String name = field.name();
+    if (escapeReservedParamNames && RESERVED_PARAM_NAMES.contains(name)) {
```
tcoffee-google: I couldn't remember why we handle these manually instead of through the normal symbol table mechanism. ?

saicheems: Because of the underlines thing, the symbol table still doesn't support underlines for disambiguation. These also hardcoded in the generator to map to specific names, so I felt it made more sense to copy the style.

Also node names don't go through the symbol table because they're all field names of the request object

PR #959, opened 2017-01-25
Ruby MVVM

eoogbe:
Refactors Ruby Gapic codegen to MVVM. Closes #349

PR #967, opened 2017-01-28
Version index file for ruby.

landrito:
Adds generation of a version file.


swcloud:
Let's wait until #959 is merged. There will be many conflicts.

landrito:
@swcloud This is ready. PTAL.

swcloud:
@landrito Can you resolve the conflicts first?

landrito:
@swcloud conflicts resolved!

src/main/java/com/google/api/codegen/viewmodel/InitCodeView.java
```diff
@@ -33,6 +33,9 @@
    */
   public abstract String apiFileName();
 
+  /** The file name of the index file. */
+  public abstract String indexFileImportName();
```
swcloud: I think 'versionIndexFileName' is better since if we use 'ImportName', it feels that this should be part of ImportSectionView

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -187,6 +187,7 @@ private InitCodeView buildInitCodeView(
         .topLevelLines(generateSurfaceInitCodeLines(context, argItems))
         .fieldSettings(getFieldSettings(context, argItems))
         .importSection(importSectionTransformer.generateImportSection(typeTable.getImports()))
+        .indexFileImportName(namer.getIndexFileImportName())
```
landrito: Using the version as the file name is a pattern that is only used in ruby. I suppose that the name 'versionIndex' will work for NodeJS since it exposes a version function (`v1`). Do you think it is confusing to call this a versionIndex when looking at the name from a NodeJS point of view.

swcloud: NodeJS index.js is essentially index for version as well. So I think it's consistent. It'll be ideal if the file is called v1.js, but it's too late to change. So I think we should use 'VersionIndex' in the code to make it clear.

landrito: Done!

src/main/java/com/google/api/codegen/viewmodel/InitCodeView.java
landrito: Done!

PR #971, opened 2017-01-30
Support request body overrides in PHP

saicheems:


src/test/java/com/google/api/codegen/testdata/discoveries/php/php_adexchangebuyer.v1.4.json.baseline
```diff
@@ -463,6 +463,7 @@ $client = getClient();
 $service = new Google_Service_AdExchangeBuyer($client);
 
 $optParams = [];
+
```
saicheems: I couldn't think of a great way to organize the `$optParams` variable with this change, I decided to put it after required params and before the request body. That means there's a newline separating this and the `do` in page streaming requests, but I thought that was alright. Lmk if you have other ideas

tcoffee-google: Seems okay to me.

PR #973, opened 2017-01-31
Implement getEnumValue

### LINKED ISSUE #944: Implement getEnumValue on each language's ModelTypeNameConverter
Related to #941, by specifying initial value for encoding, like `config.encoding=FLAC`, it ends up be an error `java.lang.UnsupportedOperationException: getEnumValue not supported by NodeJS.` It succeeds without the title, although the initial value is `ENCODING_UNSPECIFIED` which isn't great. This method should be implemented before we're going to add the config.

Affects C#, Go, NodeJS, PHP, and Ruby.

eoogbe:
Closes #944


src/main/java/com/google/api/codegen/transformer/csharp/CSharpModelTypeNameConverter.java
```diff
@@ -244,6 +239,10 @@ private TypeName getTypeNameForTypedResourceName(
 
   @Override
   public TypedValue getEnumValue(TypeRef type, String value) {
-    throw new UnsupportedOperationException("getEnumValue not supported by C#");
+    TypeName enumTypeName = getTypeName(type);
+    List<String> enumTypeNameParts = Splitter.on('+').splitToList(enumTypeName.getNickname());
```
jmuk: Sorry I'm newbie for C# -- What does this `'+'` mean?

src/test/java/com/google/api/codegen/transformer/testdata/myproto.proto
```diff
@@ -0,0 +1,21 @@
+syntax = "proto3";
```
jmuk: Instead of adding new protos like this -- can we reuse existing library.proto?
It has lots of complicated enum values for testing.

jmuk:
Also please look into the travis failures.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpModelTypeNameConverter.java
eoogbe: It's in [`getTypeName`](https://github.com/eoogbe/toolkit/blob/f83f2de6a51ee3aa38e3bfa4b364899e6e75c8e9/src/main/java/com/google/api/codegen/transformer/csharp/CSharpModelTypeNameConverter.java#L133).

eoogbe:
Removed tests that don't add to coverage and used existing testdata

src/main/java/com/google/api/codegen/transformer/csharp/CSharpModelTypeNameConverter.java
```diff
@@ -244,6 +239,16 @@ private TypeName getTypeNameForTypedResourceName(
 
   @Override
   public TypedValue getEnumValue(TypeRef type, String value) {
-    throw new UnsupportedOperationException("getEnumValue not supported by C#");
+    for (EnumValue enumValue : type.getEnumType().getValues()) {
+      if (enumValue.getSimpleName().equals(value)) {
```
jmuk: The exactly same pattern is seen among all languages. Also, `getEnumValue` is only referred from `ModelTypeTable`.

I think it's better to:
- extract `for...if...throw` code block into `ModelTypeTable`
- update `getEnumValue` interface as `public TypedValue getEnumValue(TypeRef type, EnumValue value)`

This will be beneficial for getting zero value too, it's gonna be `getEnumValue(type, type.getEnumType().getValues().get(0))`.

jmuk:
I believe it's better to keep the test. Don't over estimate the coverage reports -- they do cover the new features you've added. It looks like that the percentage number is down just because it does not cover for the `throw`, according to https://codecov.io/gh/googleapis/toolkit/pulls/973/diff

eoogbe:
@jmuk PTAL

PR #974, opened 2017-01-31
Support request body overrides in Python

saicheems:


src/test/java/com/google/api/codegen/testdata/discoveries/py/py_adexchangebuyer.v1.4.json.baseline
```diff
@@ -22,9 +22,9 @@ credentials = None
 service = discovery.build('adexchangebuyer', 'v1.4', credentials=credentials)
 
 # The account id
-id = 0  # TODO: Update placeholder value.
+id2 = 0  # TODO: Update placeholder value.
```
saicheems: What's happening here is that `id` is a built-in function in Python, and I forgot to hook up the variable names to the symbol table before, so now they're being corrected

saicheems: Fixed it by forcing underscores at the end

saicheems:
ping

PR #984, opened 2017-02-03
Add "proto_deps" to Metadata generation

geigerj:
This field already specifies the package's proto package
dependencies in the Artman config. Reading it in allow metadata
generation to specify these additional dependencies, e.g., PubSub's
dependency on the IAM proto package.

Note that because this field was previously only used by
Java/Packman, some additional configuration to specify the
language-specific package name is added via the "name_override"
field.

To be merged simultaneously with: https://github.com/googleapis/artman/pull/138, https://github.com/googleapis/googleapis/pull/248

### DIFF
--- a/src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
+++ b/src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
@@ -22,7 +22,9 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Set;
 import javax.annotation.Nullable;
 import org.yaml.snakeyaml.Yaml;
@@ -42,13 +44,13 @@ public abstract class PackageMetadataConfig {
 
   protected abstract Map<TargetLanguage, VersionBound> protoVersionBound();
 
-  protected abstract Map<TargetLanguage, VersionBound> commonProtosVersionBound();
-
   protected abstract Map<TargetLanguage, VersionBound> authVersionBound();
 
   protected abstract Map<TargetLanguage, VersionBound> generatedPackageVersionBound();
 
   protected abstract Map<TargetLanguage, String> packageName();
+
+  protected abstract Map<TargetLanguage, Map<String, VersionBound>> protoPackageDependencies();
 
   /** The version of GAX that this package depends on. Configured per language. */
   public VersionBound gaxVersionBound(TargetLanguage language) {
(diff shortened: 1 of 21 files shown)


geigerj:
cc: @bjwatson -- this PR fixes the last Python hand-edit.

src/test/java/com/google/api/codegen/testdata/library_pkg.yaml
```diff
@@ -6,6 +6,9 @@ author: Google, Inc.
 email: googleapis-packages@google.com
 homepage: https://github.com/googleapis/googleapis
 license: Apache-2.0
+proto_gen_pkg_deps:
```
shinfan: `proto_gen_pkg_deps` is a bit wordy. How about just `proto_gen_deps` or `proto_deps`

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
```diff
@@ -204,6 +201,42 @@ public static PackageMetadataConfig createFromString(String yamlContents) {
...
+      Map<String, Object> configMap) {
+    Map<TargetLanguage, Map<String, VersionBound>> packageDependencies = new HashMap<>();
+    List<String> packages = (List<String>) configMap.get("proto_gen_pkg_deps");
+
+    for (String packageName : packages) {
+      Map<String, Map<String, String>> config =
+          (Map<String, Map<String, String>>) configMap.get(packageName + "_version");
+      if (config == null) {
+        throw new IllegalArgumentException(
+            "'" + packageName + "' in proto_gen_pkg_deps was " + "not found in dependency list.");
```
shinfan: "'" + packageName + "' in proto_gen_pkg_deps was not found in dependency list."

src/main/java/com/google/api/codegen/transformer/PackageMetadataTransformer.java
```diff
@@ -27,6 +34,16 @@
...
+    return generateMetadataView(packageConfig, model, template, outputPath, language, null);
+  }
+
+  public PackageMetadataView.Builder generateMetadataView(
+      PackageMetadataConfig packageConfig,
+      Model model,
+      String template,
+      String outputPath,
+      TargetLanguage language,
+      Set<String> whitelistedDependencies) {
```
shinfan: It is not very clear that what this `whitelistedDependencies` does. Add some comments.

geigerj: Done

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
geigerj: Done

src/test/java/com/google/api/codegen/testdata/library_pkg.yaml
geigerj: Changed to `proto_deps` (and in googleapis/artman).

(I originally used `proto_gen_pkg_deps` because that was already the name of the field in the artman config.)

PR #987, opened 2017-02-03
Fix abnormal page token cases in PHP sample gen

tcoffee-google:
Addresses #692.

### DIFF
--- a/src/main/java/com/google/api/codegen/discovery/transformer/php/PhpSampleMethodToViewTransformer.java
+++ b/src/main/java/com/google/api/codegen/discovery/transformer/php/PhpSampleMethodToViewTransformer.java
@@ -169,7 +169,14 @@ public class PhpSampleMethodToViewTransformer implements SampleMethodToViewTrans
     builder.isResourceMap(fieldInfo.type().isMap());
     builder.pageVarName(
         symbolTable.getNewSymbol(namer.localVarName(Name.lowerCamel(fieldInfo.name()))));
-    builder.pageTokenName(methodInfo.requestPageTokenName());
+    boolean isResourceSetterInRequestBody = methodInfo.isPageStreamingResourceSetterInRequestBody();
+    builder.isResourceSetterInRequestBody(isResourceSetterInRequestBody);
+    // Use upper camel case for getter/setter function names in templates.
+    if (isResourceSetterInRequestBody) {
+      builder.pageTokenName(Name.lowerCamel(methodInfo.requestPageTokenName()).toUpperCamel());
+    } else {
+      builder.pageTokenName(methodInfo.requestPageTokenName());
+    }
     builder.nextPageTokenName(Name.lowerCamel(methodInfo.responsePageTokenName()).toUpperCamel());
     return builder.build();
   }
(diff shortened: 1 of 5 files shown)


saicheems:
You'll also want to edit this line:
https://github.com/googleapis/toolkit/blob/66d2df4a4ce2166b4a4115b11686e4767c0e7665/src/main/java/com/google/api/codegen/discovery/transformer/php/PhpSampleMethodToViewTransformer.java#L102

So `optParams` won't appear if a page streaming method doesn't assign the page token through `optParams`.

tcoffee-google:
> So `optParams` won't appear

Good catch.

src/main/java/com/google/api/codegen/discovery/transformer/php/PhpSampleMethodToViewTransformer.java
```diff
@@ -169,7 +170,14 @@ private SamplePageStreamingView createSamplePageStreamingView(
     builder.isResourceMap(fieldInfo.type().isMap());
     builder.pageVarName(
         symbolTable.getNewSymbol(namer.localVarName(Name.lowerCamel(fieldInfo.name()))));
-    builder.pageTokenName(methodInfo.requestPageTokenName());
+    boolean isResourceSetterInRequestBody = methodInfo.isPageStreamingResourceSetterInRequestBody();
+    builder.isResourceSetterInRequestBody(isResourceSetterInRequestBody);
+    // Use upper camel case for getter/setter function names in templates.
+    if (isResourceSetterInRequestBody) {
+      builder.pageTokenName(Name.lowerCamel(methodInfo.requestPageTokenName()).toUpperCamel());
```
saicheems: Use the namer for this, you can either add a `getPageTokenSetterFuncName` to the `PhpSampleNamer` or call `namer.publicMethodName(Name.lowerCamel("get", methodInfo.requestPageTokenName()))`

That way there's no name composition in the snippet

saicheems: Oh, I see, that's how I did it in Java, huh? :/

If you could make an issue then so I can clean it up later that'll be enough.

tcoffee-google: https://github.com/googleapis/toolkit/issues/991

src/main/resources/com/google/api/codegen/php/sample.snip
```diff
@@ -125,6 +125,26 @@
   ?>
 @end
 
+@private nextPageToken(class)
+  ${@class.responseVarName}->get{@class.pageStreaming.nextPageTokenName}()
+@end
+
+@private updatePageToken(class)
+  @if class.pageStreaming.isResourceSetterInRequestBody
+    ${@class.requestBodyVarName}->set{@class.pageStreaming.pageTokenName}({@nextPageToken(class)});
```
saicheems: Would also generally recommend not putting semicolons in the sub snippets and instead appending them to the actual call for consistency

```
...
updatePageToken(class);
```

saicheems:
^ With that last comment, everything else seems fine, lgtm

PR #992, opened 2017-02-07
Ruby: Migrate Packman Metadata Generation

landrito:
First part in stopping using packman.


swcloud:
@jmuk Could you please help take a look at this? Thanks!

src/test/java/com/google/api/codegen/testdata/ruby_gemspec_library.baseline
```diff
@@ -0,0 +1,26 @@
...
+  s.summary       = 'GRPC library for service {{{api.name}}}'
+
+  s.add_dependency 'grpc', '~> 1.0'
+  s.add_dependency 'googleauth', '~> 0.5.1'
+  s.add_dependency 'google-common-protos', '~> 1.3.1'
+  s.add_dependency 'google-some-other-package-v1', '~> 0.2.1'
+
+  s.add_development_dependency 'bundler', '~> 1.9'
+  s.add_development_dependency 'rake', '~> 10.4'
+end
```
jmuk: The last 'no newline' style is not consistent among this file and `ruby_doc_gemspec_library.baseline`. Is it expected?

src/main/java/com/google/api/codegen/transformer/ruby/RubyPackageMetadataNamer.java
```diff
@@ -0,0 +1,50 @@
...
+public class RubyPackageMetadataNamer extends PackageMetadataNamer {
+  private Name serviceName;
+
+  public RubyPackageMetadataNamer(String packageName) {
+    // Get the service name from the package name by removing the version suffix (if any).
+    List<String> names = Splitter.on("::").splitToList(packageName);
+    if (names.size() < 2) {
+      this.serviceName = Name.upperCamel(packageName);
+    } else {
+      this.serviceName = Name.upperCamel(names.get(0));
```
jmuk: Let me clarify what's in `packageName`. For example of Speech API, I am thinking like this is something like `Google::Cloud::Speech::V1beta1` -- is that right? In that case, this code defines `serviceName` as `Google`, I think that's not expected.

src/test/java/com/google/api/codegen/testdata/ruby_gemspec_library.baseline
landrito: Not expected. Fixed!

src/main/java/com/google/api/codegen/transformer/ruby/RubyPackageMetadataNamer.java
landrito: Ah good catch. Fixed.

landrito:
Added other generated files from packman.

src/main/java/com/google/api/codegen/transformer/ruby/RubyPackageMetadataNamer.java
```diff
@@ -0,0 +1,49 @@
...
+
+  @Override
+  public String getMetadataIdentifier() {
+    List<String> names = Splitter.on("::").splitToList(packageName);
+    if (names.size() > 1) {
+      names = names.subList(0, names.size() - 1);
+    }
+    ImmutableList.Builder<String> lowerNames = ImmutableList.builder();
+    for (String name : names) {
+      lowerNames.add(Name.upperCamel(name).toLowerCamel());
```
jmuk: According to the rubygems naming convention (see http://guides.rubygems.org/name-your-gem/), this would be `toLowerUnderscore()`.

src/main/java/com/google/api/codegen/transformer/ruby/RubyPackageMetadataTransformer.java
```diff
@@ -0,0 +1,79 @@
...
+    ImmutableList.Builder<ViewModel> views = ImmutableList.builder();
+    for (String template : TOP_LEVEL_FILES) {
+      views.add(generateMetadataView(model, template, namer));
+    }
+    return views.build();
+  }
+
+  private ViewModel generateMetadataView(
+      Model model, String template, RubyPackageMetadataNamer namer) {
+    String noLeadingRubyDir = template.startsWith("ruby/") ? template.substring(5) : template;
```
jmuk: I prefer avoiding the length value like `5`. This could be something like:
```java
private static final String RUBY_PREFIX = "ruby/";
...
...
String noLeadingRubyDir = template.startsWith(RUBY_PREFIX) ? template.substring(RUBY_PREFIX.length) : template;
```

src/main/java/com/google/api/codegen/transformer/ruby/RubyPackageMetadataTransformer.java
```diff
@@ -0,0 +1,79 @@
...
+    int extensionIndex = noLeadingRubyDir.lastIndexOf(".");
+    String outputPath = noLeadingRubyDir.substring(0, extensionIndex);
```
jmuk: For README -- the final result should be `README.md`, right?  You may want to rename the snip file as `README.md.snip`

src/main/java/com/google/api/codegen/transformer/ruby/RubyPackageMetadataNamer.java
landrito: Done!

src/main/java/com/google/api/codegen/transformer/ruby/RubyPackageMetadataTransformer.java
landrito: Done.

src/main/java/com/google/api/codegen/transformer/ruby/RubyPackageMetadataTransformer.java
landrito: Fixed!

PR #993, opened 2017-02-07
fix Go enum

pongad:


garrettjonesgoogle:
Is it hard to make a baseline that exercises this feature?

pongad:
@garrettjonesgoogle PTAL. BTW, the earlier zeros are valid Go. All enums are just constant numbers.

garrettjonesgoogle:
LGTM

PR #997, opened 2017-02-08
add deprecation config

pongad:


src/test/java/com/google/api/codegen/testdata/library_gapic.yaml
```diff
@@ -504,6 +504,7 @@ interfaces:
       - go
       - java
       visibility: PACKAGE
+      deprecated: TRUE
```
garrettjonesgoogle: I would prefer a more flexible enumeration. Maybe:

`lifecycle: DEPRECATED`

That allows us to do something like this too:

`lifecycle: EXPERIMENTAL`

This also implies that the SurfaceNamer for each language needs to produce that language's string for each concept.

pongad:
@garrettjonesgoogle PTAL

src/main/java/com/google/api/codegen/viewmodel/StaticLangApiMethodView.java
```diff
@@ -89,6 +89,8 @@ public boolean isStreaming() {
   @Nullable
   public abstract LongRunningOperationDetailView operationMethod();
 
+  public abstract String releaseAnnotation();
```
garrettjonesgoogle: I  think `releaseLevelAnnotation` would be clearer.

pongad: Done

garrettjonesgoogle:
LGTM

PR #998, opened 2017-02-08
update Go header

pongad:
A couple things of note:

The GRPC library doesn't expose a version yet, so that's not available
right now

The version string is in YYYYMMDD format. Rationale:
The semver spec states that "Once a versioned package has been released,
the contents of that version MUST NOT be modified". Since "go get" just
downloads the master branch, we are technically modifying the version
every time we submit code. If our users use the version package for
their own telemetry, they might be unpleasantly surprised.

### DIFF
--- a/src/main/resources/com/google/api/codegen/go/doc.snip
+++ b/src/main/resources/com/google/api/codegen/go/doc.snip
@@ -15,5 +15,20 @@
     @end
     package {@view.fileHeader.localPackageName} // import "{@view.importPath}"
 
-    const gapicNameVersion = "gapic/20170208"
+    import (
+        "golang.org/x/net/context"
+        "google.golang.org/grpc/metadata"
+    )
+
+    const (
+        gapicName = "gapic"
+        gapicVersion = "20170208"
+    )
+
+    func insertXGoog(ctx context.Context, val string) context.Context {
+        md, _ := metadata.FromContext(ctx)
+        md = md.Copy()
+        md["x-goog-api-client"] = []string{val}
+        return metadata.NewContext(ctx, md)
+    }
 @end
--- a/src/main/resources/com/google/api/codegen/go/doc.snip
+++ b/src/main/resources/com/google/api/codegen/go/doc.snip
@@ -15,5 +15,5 @@
     @end
     package {@view.fileHeader.localPackageName} // import "{@view.importPath}"
 
-    const gapicNameVersion = "gapic/0.1.0"
+    const gapicNameVersion = "gapic/20170208"
 @end
--- a/src/main/resources/com/google/api/codegen/go/doc.snip
+++ b/src/main/resources/com/google/api/codegen/go/doc.snip
@@ -20,11 +20,6 @@
         "google.golang.org/grpc/metadata"
     )
 
-    const (
-        gapicName = "gapic"
-        gapicVersion = "20170208"
-    )
-
     func insertXGoog(ctx context.Context, val string) context.Context {
         md, _ := metadata.FromContext(ctx)
         md = md.Copy()
--- a/src/main/resources/com/google/api/codegen/go/main.snip
+++ b/src/main/resources/com/google/api/codegen/go/main.snip
@@ -96,7 +96,7 @@
                 {@stub.name}: {@stub.createStubFunctionName}(conn),
             @end
         }
-        c.SetGoogleClientInfo(gapicName, gapicVersion)
+        c.SetGoogleClientInfo("gapic", version.Repo)
         return c, nil
     }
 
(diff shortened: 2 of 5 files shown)


src/main/resources/com/google/api/codegen/go/main.snip
```diff
@@ -116,7 +116,7 @@
     // use by Google-written clients.
     func (c *{@view.clientTypeName}) SetGoogleClientInfo(name, version string) {
         goVersion := strings.Replace(runtime.Version(), " ", "_", -1)
-        v := fmt.Sprintf("%s/%s %s gax/%s go/%s", name, version, gapicNameVersion, gax.Version, goVersion)
+        v := fmt.Sprintf("%s/%s %s gax/%s gl-go/%s", name, version, gapicNameVersion, gax.Version, goVersion)
```
bjwatson: This is the heart of the change, so I have several comments here:

- The first tuple contains `gax/{version}`, which is duplicative.
- How can a Veneer override `gapic/{version}` with `gccl/{version}`, so that we can tell when a Veneer is used vs. the direct GAPIC surface?
- The GAPIC version is not required when the `gccl` version is defined, since their relationship is deterministic (assuming we keep track of the version of GAPIC shipped with each Veneer version).
- Do you have access to the gRPC version? Even though it's reported in `user-agent`, the Chemist team does not want to look for it there. It should be reported in this custom header as `grpc/{version}`.
- Please see Luke's email today about possibly changing the ordering, and weigh in if you have opinions. Since you hard-code the string, I imagine it's all the same to you.

bjwatson: Oh, I just saw your comment about the gRPC version. I guess we table that for the future. It's a P1 metric anyway.

pongad: I'll fix the `gax/{version}` duplication. But to answer the remaining question, under the new logic
- handwritten clients can just call `client.SetGoogleClientInfo("gccl", version)`
- the "constructor" of the client uses the same method to set `"gapic"/gapicVersion` tuple

I'll wait for the update on the internal doc (5pm Pacific) then send another revision

bjwatson:
@pongad The Java implementation has a test case (and I requested one more). Is this easy to do for Go?

src/main/resources/com/google/api/codegen/go/doc.snip
```diff
@@ -15,5 +15,5 @@
     @end
     package {@view.fileHeader.localPackageName} // import "{@view.importPath}"
 
-    const gapicNameVersion = "gapic/0.1.0"
+    const gapicNameVersion = "gapic/20170208"
```
bjwatson: This definition should be moved somewhere that will be easy to remember to update whenever a new GAPIC release is cut.

src/main/resources/com/google/api/codegen/go/main.snip
bjwatson: The constructor is actually setting the following:

```
    c.SetGoogleClientInfo("gax", gax.Version)
```

IIUC, the Veneers don't call `NewClient()`? Or do they, but then immediately follow it with a `SetGoogleClientInfo()` call?

If the latter, do we know that there's only one `x-goog-api-client` in the header? It's actually valid in the HTTP header spec to duplicate header keys.

pongad:
@bjwatson The handwritten libraries do call `NewClient()`; they immediately follow with `SetGoogleClientInfo(...)`. The new `insertXGoog` method guarantees we only have one.

bjwatson:
@jba Can you take a look at the latest revisions? I'm not familiar with the Go aspects of the changes. I'll review the header-related stuff tomorrow.

jba:
@bjwatson LGTM.

src/main/resources/com/google/api/codegen/go/doc.snip
```diff
@@ -15,5 +15,20 @@
...
 
-    const gapicNameVersion = "gapic/0.1.0"
+    import (
+        "golang.org/x/net/context"
+        "google.golang.org/grpc/metadata"
+    )
+
+    const (
+        gapicName = "gapic"
+        gapicVersion = "20170208"
```
bjwatson: I'm afraid that we're still going to forget to update this. Could you add a context method to generate this date string for now, so that it always reflects the GAPIC build date? I'm less worried about the risk of multiple release cut on the same date.

src/main/resources/com/google/api/codegen/go/main.snip
```diff
@@ -116,8 +116,7 @@
     // use by Google-written clients.
     func (c *{@view.clientTypeName}) SetGoogleClientInfo(name, version string) {
         goVersion := strings.Replace(runtime.Version(), " ", "_", -1)
-        v := fmt.Sprintf("%s/%s %s gax/%s go/%s", name, version, gapicNameVersion, gax.Version, goVersion)
-        c.metadata = metadata.Pairs("x-goog-api-client", v)
+        c.xGoogHeader = fmt.Sprintf("gl-go/%s %s/%s gax/%s", goVersion, name, version, gax.Version)
```
bjwatson: Let's include `grpc/UNKNOWN` as the last element of this string. We're doing that in Java when we can't determine it, and I think will @jskeet do the same with .NET.

pongad:
@bjwatson PTAL

PR #999, opened 2017-02-08
Add CommentReformatter

eoogbe:
Many languages format doc comments in some way to match their standard documentation format. Factors out a `CommentReformatter` to handle language differences in rendering doc comments, while keeping common code in the `CommonRenderingUtil`.

1. **Add CommentReformatter**
2. Add BundlingDescriptorView
3. Add GrpcDocView
4. Add ImportSectionView
6. Add ParamDocTransformer
7. Ruby MVVM migration


jmuk:
I'm not sure how much beneficial is this change. To me, the main purpose of this PR is to share the same logic between the CommentFixers (JSDoc- and RDoc-) and RenderingUtils in MVVM. Since we are going to switch to MVVM very soon, those CommentFixers will be gone by then.

With this plan, I'd like to suggest to create `RubyRenderingUtil.java` and leave others as-is. The new file will copy the exact logic as the current `RDocCommentFixer.java` but that's fine.

garrettjonesgoogle:
I think the PR is good as is - LGTM - I'll leave it to others to decide the final course of action.

src/test/java/com/google/api/codegen/ProtoDocumentLinkTest.java
```diff
@@ -47,22 +47,23 @@ public void testMatchProtoLink() {
   }
 
   @Test
-  public void testRDocCommentFixer() {
```
landrito: Until the MVVM PRs are merged, we will still depend on the {lang}CommentFixers. I think its better to not lose coverage of the {lang}CommentFixers until we remove the {lang}CommentFixers in the MVVM PRs.

eoogbe: The `RDocCommentFixer` just calls the `RubyCommentFormatter` now. The logic has not changed; it's just been moved to MVVM.

landrito: Sure the logic has not been changed, but leaving it as is until MVVM deletes the RDocComment fixer protects us on the (admittedly very) off chance, between the time this PR is merged and the time MVVM is, from a bug being introduced in the RDocCommentFixer.

I agree, the same logic is continuing to be tested, so I'm fine with changing the tests now but would prefer changing them during the MVVM PR. Your call.

PR #1000, opened 2017-02-08
Add BundlingDescriptorView

eoogbe:
Adds bundling descriptors that Ruby needs for MVVM.

1. Add CommentReformatter
2. **Add BundlingDescriptorView**
3. Add GrpcDocView
4. Add ImportSectionView
6. Add ParamDocTransformer
7. Ruby MVVM migration


garrettjonesgoogle:
1000! ding ding ding ding

garrettjonesgoogle:
Hmm, this doesn't affect any baselines - any way to make this be used in the generation so that we can get coverage?

eoogbe:
I can add it to the `RubyGapicContext` to increase coverage

garrettjonesgoogle:
Why weren't any baselines affected by the snippet change?

eoogbe:
The change is part of the MVVM refactoring for Ruby. It shouldn't change codegen output.

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
```diff
@@ -127,6 +129,8 @@ public PhpGapicSurfaceTransformer(ApiConfig apiConfig, GapicCodePathMapper pathM
         pathTemplateTransformer.generatePathTemplateGetterFunctions(context));
     xapiClass.pageStreamingDescriptors(pageStreamingTransformer.generateDescriptors(context));
     xapiClass.hasPageStreamingMethods(context.getInterfaceConfig().hasPageStreamingMethods());
+    xapiClass.bundlingDescriptors(ImmutableList.<BundlingDescriptorView>of());
+    xapiClass.hasBundlingMethods(context.getInterfaceConfig().hasBundlingMethods());
```
garrettjonesgoogle: Don't add these lines if the PHP snippet isn't using these fields... If you remove them, then also `hasBundlingMethods` no longer needs to be added to `DynamicLangXApiView`.

src/main/resources/com/google/api/codegen/ruby/main.snip
```diff
@@ -132,14 +132,9 @@
...
       BUNDLE_DESCRIPTORS = {
-        @let ifaceConfig = context.getApiConfig.getInterfaceConfig(service)
-          @join method : context.messages.filterBundlingMethods(ifaceConfig, context.getSupportedMethods(service)) on {@", "}.add(BREAK)
-            @let bundling = ifaceConfig.getMethodConfig(method).getBundling(), \
-                 methodName = context.upperCamelToLowerUnderscore(method.getSimpleName)
-              "{@methodName}" => Google::Gax::BundleDescriptor.new(
-                {@bundleDescriptorBody(bundling, method)})
-            @end
-          @end
+        @join descriptor : context.getBundlingDescriptors(service) on {@", "}.add(BREAK)
```
garrettjonesgoogle: I'm assuming that the baseline doesn't change because your new form here exactly replicates the original output?

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
eoogbe: Ruby will use it when the MVVM migration is complete.

src/main/resources/com/google/api/codegen/ruby/main.snip
eoogbe: Yes, the output is the same. The new code is for migrating to MVVM.

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
garrettjonesgoogle: This is kind of why I didn't want the original PR split up...

src/main/java/com/google/api/codegen/viewmodel/DynamicLangXApiView.java
```diff
@@ -68,6 +72,8 @@
 
   public abstract boolean hasPageStreamingMethods();
```
landrito: Why not make this Nullable as well and removing it from the PHP surface transformer.

landrito: Nevermind, I think its fine to leave not nullable.

eoogbe: Primitive types cannot be Nullable

landrito: TIL.

PR #1001, opened 2017-02-08
Refactor gRPC doc message gen for Ruby to MVVM

eoogbe:
Will be used to generate gRPC docs for Ruby, Python, and Node.

1. Add CommentReformatter
2. Add BundlingDescriptorView
3. **<s>Add GrpcDocView</s> Migrate gRPC doc message gen**
4. Add ImportSectionView
6. Add ParamDocTransformer
7. Ruby MVVM migration


garrettjonesgoogle:
Either this PR needs coverage (it's clear that `GrpcElementDocTransformer` doesn't have any tests that exercise it) or it needs to be rolled into another one of the PRs that will actually use it.

landrito:
This is heavily tied to Ruby Doc Generation. It might be a good idea to do the Documentation MVVM migration in this PR. I only see it adding slight changes to the snip file, creating a ModelToViewTransformer, and providing it in the main gapic provider factory.

src/main/resources/com/google/api/codegen/ruby/message.snip
```diff
@@ -1,32 +1,42 @@
...
 
 @end
 
-@private documentMessages(containerElement)
-  @join element : context.filterDocumentingMessages(containerElement) on BREAK.add(BREAK)
-    @let document = context.methodDocComment(element)
-      @if document
-        {@toComments(document)}
+@private documentMessage(doc)
+  @if doc.lines
```
landrito: Seems like there is shared code between this if and the else. Is that because you are trying to avoid an extra line?

src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceDocTransformer.java
```diff
@@ -0,0 +1,78 @@
...
+  private ViewModel generateDoc(ProtoFile file, ApiConfig apiConfig) {
+    String subPath = pathMapper.getOutputPath(file, apiConfig);
+    String protoFilename = file.getSimpleName();
+    String baseFilename =
+        protoFilename.substring(0, protoFilename.length() - "proto".length()) + "rb";
+    ModelTypeTable typeTable =
+        new ModelTypeTable(
+            new RubyTypeTable(apiConfig.getPackageName()),
+            new RubyModelTypeNameConverter(apiConfig.getPackageName()));
+    SurfaceNamer namer = new RubySurfaceNamer(typeTable.getFullNameFor(file));
```
eoogbe: The docs for the files are not in the same package as the main API. You end up with docs in the tests for `Library::V1::Timestamp` instead of `Google::Protobuf::Timestamp`.

landrito: Thanks for the clarification.

src/main/java/com/google/api/codegen/transformer/GrpcElementDocTransformer.java
```diff
@@ -0,0 +1,97 @@
...
+    children.addAll(generateMessageDocs(typeTable, namer, containerElement));
+    children.addAll(generateEnumDocs(typeTable, namer, containerElement));
+    return children.build();
+  }
+
+  private List<GrpcElementDocView> generateMessageDocs(
+      ModelTypeTable typeTable, SurfaceNamer namer, ProtoContainerElement containerElement) {
+    ImmutableList.Builder<GrpcElementDocView> messageDocs = ImmutableList.builder();
+    for (MessageType message : containerElement.getMessages()) {
+      if (message.isMapEntry()) {
```
eoogbe: They are already covered by the dictionary type (e.g. `Hash` for Ruby).

landrito: Ah makes sense. Brain fart.

eoogbe:
@landrito @shinfan PTAL

garrettjonesgoogle:
Update the change description to match the current scope of this PR, then LGTM

PR #1002, opened 2017-02-08
Toolkit: HTTP header standardization for Python

lukesneeringer:
This is the toolkit portion of the HTTP header standardization for Python.

It mutually depends on two other pull requests and should not be merged until all three are tested and approved:

  * googleapis/googleapis#250
  * googleapis/gax-python#157


src/main/resources/com/google/api/codegen/py/main.snip
```diff
@@ -211,24 +212,52 @@
...
                 this data. Falls back to the default config if not specified
                 or the specified config is missing data points.
-              app_name (string): The codename of the calling service.
-              app_version (string): The version of the calling service.
+              app_name (string): The name of the application calling
+                the service. Recommended for analytics purposes.
+              app_version (string): The version of the application calling
+                the service. Recommended for analytics purposes.
+              lib_name (string): The API library software used for calling
+                the service. (Unless you are writing an API client itself,
```
geigerj: grammar: "an API client itself" --> "an API client"

src/main/resources/com/google/api/codegen/py/main.snip
```diff
@@ -100,8 +100,6 @@
         DEFAULT_SERVICE_PORT = {@context.getServiceConfig.getServicePort()}
         """The default port of the service."""
 
-        _CODE_GEN_NAME_VERSION = 'gapic/0.1.0'
-
         _GAX_VERSION = pkg_resources.get_distribution('google-gax').version
```
geigerj: It looks like `_GAX_VERSION` is now unused (and presumably can be derived in GAX itself); please remove.

src/main/resources/com/google/api/codegen/py/main.snip
```diff
@@ -211,24 +212,52 @@
...
+                the service. Recommended for analytics purposes.
+              app_version (string): The version of the application calling
+                the service. Recommended for analytics purposes.
+              lib_name (string): The API library software used for calling
+                the service. (Unless you are writing an API client itself,
+                leave this as default.)
+              lib_version (string): The API library software version used
+                for calling the service. (Unless you are writing an API client
+                itself, leave this as default.)
+              metrics_headers (dict): A dictionary of values for tracking
```
geigerj: @bjwatson I didn't see mention of this parameter (or allowing generic key-value pairs in the header) in the design doc. Is this how we want to do this in all languages? If not, can the design doc be amended to specify the list of expected configuration parameters available in the GAPIC client library object constructor?

lukesneeringer: I more or less consider `metrics_headers` to be a semi-private argument (like the preceding four arguments). I considered preceding it with an underscore but that seems not quite correct since it will be called as a "public" argument by Google-produced code elsewhere.

src/main/resources/com/google/api/codegen/py/main.snip
lukesneeringer: Concur. (And yes, I have a forthcoming PR to GAX where that is exactly what happens.)

src/main/resources/com/google/api/codegen/py/main.snip
```diff
@@ -174,24 +170,27 @@
...
                 credentials=None,
                 ssl_credentials=None,
                 scopes=None,
                 client_config=None,
-                app_name='gax',
-                app_version=_GAX_VERSION):
+                app_name=None,
+                app_version='0.0',
+                lib_name=None,
+                lib_version='0.0',
```
bjwatson: `lib_name` should be `'gax'`, and `lib_version` should be `_GAX_VERSION`

lukesneeringer: The `gax` portion of the header is added [here][1].

  [1]: https://github.com/lukesneeringer/gax-python/blob/headers/google/gax/api_callable.py#L341

lukesneeringer: Just spoke with @bjwatson offline and he confirmed that the current implementation is correct.

src/main/resources/com/google/api/codegen/py/main.snip
geigerj: (this is a nit, but I still think this is wrong...)

src/main/resources/com/google/api/codegen/py/main.snip
geigerj: SGTM assuming @bjwatson is OK with this.

(I'm not sure what "semi-private" means given that any developer will be able to see these arguments in the public docs, and this constructor will be the entry point to the client library for any API that doesn't have a handwritten layer.)

src/main/resources/com/google/api/codegen/py/main.snip
lukesneeringer: I like the emphasis (a valid use of the word), and would like to keep it. :-)

geigerj: Hm, to sound natural to my ears, "*self" always refers to the sentence subject. But definitely not something to block on.

src/main/resources/com/google/api/codegen/py/main.snip
geigerj: I understand from your comment on the GAX PR that this is intended for A/B testing, which makes sense to me.

lukesneeringer: One problem is that the GAPICs are using this to pass info to GAX, but there will later be A/B headers, which go at the end. Right now if we add those tomorrow in Python, they will be out of order.

The place to fix that is in GAX.

src/main/resources/com/google/api/codegen/py/main.snip
lukesneeringer: Note: The `'0.0'` (which should be `'UNKNOWN'`) is fixed in #1019.

PR #1005, opened 2017-02-09
Ruby: Request headers update.

landrito:
cc @bjwatson

### DIFF
--- a/src/main/resources/com/google/api/codegen/ruby/main.snip
+++ b/src/main/resources/com/google/api/codegen/ruby/main.snip
@@ -224,8 +224,8 @@
         timeout: DEFAULT_TIMEOUT,
         app_name: nil,
         app_version: nil,
-        lib_name: "gapic",
-        lib_version: "0.1.0"
+        lib_name: nil,
+        lib_version: "UNKNOWN"
       @# These require statements are intentionally placed here to initialize
       @# the gRPC module only when it's required.
       @# See https://github.com/googleapis/toolkit/issues/446
@@ -254,9 +254,11 @@
         warn "`app_name` and `app_version` are no longer being used in the request headers."
       end
 
-      google_api_client = "gl-ruby/#{RUBY_VERSION} " @\
-        "#{lib_name}/{lib_version} gax/#{Google::Gax::VERSION} " @\
-        "grpc/#{GRPC::VERSION}".freeze
+      google_api_client = "gl-ruby/#{RUBY_VERSION}"
+      google_api_client << " #{lib_name}/{lib_version}" if lib_name
+      google_api_client << " gapic/0.1.0 gax/#{Google::Gax::VERSION}"
+      google_api_client << " grpc/#{GRPC::VERSION}"
+      google_api_client.freeze
 
       headers = { :"x-goog-api-client" => google_api_client }
       {@constructDefaults(service)}
--- a/src/main/resources/com/google/api/codegen/ruby/main.snip
+++ b/src/main/resources/com/google/api/codegen/ruby/main.snip
@@ -254,9 +254,9 @@
         warn "`app_name` and `app_version` are no longer being used in the request headers."
       end
 
-      google_api_client = "#{lib_name}/{lib_version} " @\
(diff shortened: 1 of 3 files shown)


src/main/resources/com/google/api/codegen/ruby/main.snip
```diff
@@ -228,8 +222,10 @@
         scopes: ALL_SCOPES,
         client_config: {},
         timeout: DEFAULT_TIMEOUT,
-        app_name: "gax",
-        app_version: Google::Gax::VERSION
+        app_name: nil,
+        app_version: nil,
+        lib_name: "gapic",
+        lib_version: "0.1.0"
```
lukesneeringer: This seems like it is probably wrong (either that or _my_ implementation is wrong, which is totally possible!):

  * I was of the understanding that `lib_name` is supposed to be the library built _on top of_ the GAPIC, not the GAPIC itself. If another library is used, _both_ the `gccl` header _and_ the `gapic` header should be sent, correct?
  * `lib_version` probably should not be hard-coded.

src/main/resources/com/google/api/codegen/ruby/main.snip
```diff
@@ -248,13 +244,20 @@
...
-      google_api_client = "#{app_name}/#{app_version} " @\
-        "#{CODE_GEN_NAME_VERSION} gax/#{Google::Gax::VERSION} " @\
-        "ruby/#{RUBY_VERSION}".freeze
+      if app_name || app_version
+        warn "`app_name` and `app_version` are no longer being used in the request headers."
+      end
+
+      google_api_client = "#{lib_name}/{lib_version} " @\
+        "gax/#{Google::Gax::VERSION} grpc/#{GRPC::VERSION} " @\
+        "gl-ruby/#{RUBY_VERSION}".freeze
```
lukesneeringer: Move `gl-ruby` to the front of the line (based on changes to the spec from yesterday evening).

src/main/resources/com/google/api/codegen/ruby/main.snip
lukesneeringer: Update: After looking at the spec again, I think I am wrong about the first bullet (the Python implementation is specifically called out as expected to be slightly different from all the others).

landrito: Agreed that lib_version should not be hard-coded. The prior is hard coded, so lets just hard code for now.

Tracking as an issue: https://github.com/googleapis/toolkit/issues/1015.

src/main/resources/com/google/api/codegen/ruby/main.snip

src/main/resources/com/google/api/codegen/ruby/main.snip
lukesneeringer: Update: Based on this morning's changes, this is wrong again. :-)

src/main/resources/com/google/api/codegen/ruby/main.snip
```diff
@@ -248,13 +244,20 @@
...
 
-      google_api_client = "#{app_name}/#{app_version} " @\
-        "#{CODE_GEN_NAME_VERSION} gax/#{Google::Gax::VERSION} " @\
-        "ruby/#{RUBY_VERSION}".freeze
+      if app_name || app_version
+        warn "`app_name` and `app_version` are no longer being used in the request headers."
+      end
+
+      google_api_client = "gl-ruby/#{RUBY_VERSION} " @\
+        "#{lib_name}/{lib_version} gax/#{Google::Gax::VERSION} " @\
```
lukesneeringer: Add `gapic/#{CODE_GEN_NAME_VERSION}` and make `#{lib_name}/{lib_version}` only be added if `lib_name` is sent.

lukesneeringer: Also, how are you handling the `gl-ab*` headers?

lukesneeringer:
Looks good except for the lack of A/B.

landrito:
Thanks for the review @lukesneeringer.

PR #1012, opened 2017-02-09
Initial implementation of ruby smoke test

swcloud:
Use ApiMethodView in the SmokeTestClassView instead of TestCaseView, because this enables directly dropping in the sample-gen code snippet to the smoke test. Since sample-gen code already handles more complicated cases of making method call such as paging, streaming and LRO (which current TestCaseView-based approach does not handle), this approach makes smoke test generation a lot easier (essentially one-line call to the sample-gen), without going through redundant logic to make method calls via TestCaseView.


src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceTestTransformer.java
```diff
@@ -0,0 +1,165 @@
...
+
+    return testClass.build();
+  }
+
+  private OptionalArrayMethodView createSmokeTestCaseApiMethodView(
+      MethodTransformerContext context) {
+    OptionalArrayMethodView initApiMethodView =
+        new DynamicLangApiMethodTransformer().generateMethod(context);
+
+    OptionalArrayMethodView.Builder apiMethodView = OptionalArrayMethodView.newBuilder();
```
shinfan: Add a `toBuilder()` method to `OptionalArrayMethodView` then you can get rid of the following lines.

src/main/java/com/google/api/codegen/viewmodel/testing/TestCaseView.java
```diff
@@ -22,7 +22,6 @@
 
 @AutoValue
 public abstract class TestCaseView {
-
```
shinfan: Revert this blank line

src/main/java/com/google/api/codegen/viewmodel/testing/TestCaseView.java
```diff
@@ -65,7 +64,6 @@ public static Builder newBuilder() {
 
   @AutoValue.Builder
   public abstract static class Builder {
-
```
shinfan: ditto

src/main/java/com/google/api/codegen/transformer/TestCaseTransformer.java
```diff
@@ -222,4 +223,86 @@ private InitCodeContext createResponseInitCodeContext(
...
+    ClientMethodType methodType = ClientMethodType.FlattenedMethod;
+    if (context.getMethodConfig().isPageStreaming()) {
+      methodType = ClientMethodType.PagedFlattenedMethod;
+    }
+
+    return createTestCaseView(
+        context, new SymbolTable(), createSmokeTestInitContext(context), methodType);
+  }
+
+  public boolean requireProjectId(InitCodeView initCodeView, SurfaceNamer namer) {
```
shinfan: Add comments to indicate this method is used by smoke tests

src/main/java/com/google/api/codegen/transformer/TestCaseTransformer.java
```diff
@@ -222,4 +223,86 @@ private InitCodeContext createResponseInitCodeContext(
...
+      fieldConfigMap =
+          FieldConfig.toFieldConfigMap(
+              context.getFlatteningConfig().getFlattenedFieldConfigs().values());
+    } else {
+      outputType = InitCodeContext.InitCodeOutputType.SingleObject;
+      fieldConfigMap = null;
+    }
+
+    // Store project ID variable name into the symbol table since it is used by the execute method
+    // as a parameter. For more information please see smoke_test.snip.
```
shinfan: Remove the comment 'For more information please see smoke_test.snip.' since now this method is cross-language.

src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceTestTransformer.java
swcloud: That's exactly what I was looking for! Nice!

shinfan:
Test generator change LGTM

src/test/java/com/google/api/codegen/testdata/ruby_smoke_test_library.baseline
```diff
@@ -0,0 +1,41 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# AUTO GENERATED BY GAPIC
+
+require "minitest/autorun"
```
landrito: I think you are going to need to add development dependencies to package metadata generation snippets.

swcloud:
@landrito Can you take a look at generated the smoke test code?

landrito:
@swcloud Will do.

src/test/java/com/google/api/codegen/testdata/ruby_smoke_test_library.baseline
swcloud: But minitest is part of ruby distribution

landrito: Ah you are right my mistake.

src/test/java/com/google/api/codegen/testdata/ruby_smoke_test_library.baseline
```diff
@@ -0,0 +1,41 @@
+============== file: test/library/v1/library_service_smoke_test.rb ==============
```
landrito: Having this be generated into the same top level directory as the unit tests will cause problems automating the tests in the rake file. I think it would be better to have this generate something like `smoke_test/library/v1/library_service_smoke_test.rb`.

src/test/java/com/google/api/codegen/testdata/ruby_smoke_test_library.baseline
```diff
@@ -0,0 +1,41 @@
...
+# AUTO GENERATED BY GAPIC
+
+require "minitest/autorun"
+require "minitest/spec"
+require "library/v1/library_service_client"
+
+describe "LibraryServiceSmokeTest" do
+  it "runs one smoke test with update_book" do
+    if ARGV.empty?
+      fail "Usage: #{$0} <project_id>"
```
landrito: Is the `keyfile` not needed?

landrito:
Sorry about the approval -> requesting changes. I was working on updates to the Rakefile and found some points that might cause friction.

src/test/java/com/google/api/codegen/testdata/ruby_smoke_test_library.baseline
```diff
@@ -0,0 +1,41 @@
...
+    end
+    project_id = ARGV[0].freeze
```
landrito: I am thinking this would be better as an environment variable rather than an argument.
This will allow Rake to be able to run the test as well as being able to run the test from cli. This is because rake runs tests by require-ing the file and there is no way of passing an argument to a file using require. 

This part of code will end up looking like this if changed:
```rb
unless ENV["SMOKE_TEST_PROJECT"]
  fail "Usage: SMOKE_TEST_PROJECT=<project_id> ruby library_service_smoke_test.rb"
end
project_id = ENV["SMOKE_TEST_PROJECT"].freeze
```

For reference here is the bit in the Rakefile that I have in mind.
```rb
# Smoke tests
desc "Runs the smoke tests."
task :smoke_test, :project do |t, args|
  project = args[:project]
  project ||= ENV["SMOKE_TEST_PROJECT"]
  if project.nil?
    fail "You must provide a project. e.g. rake smoke_test[test123] or SMOKE_TEST_PROJECT=test123 rake smoke_test"
  end

  $LOAD_PATH.unshift "lib", "smoke_test"
  Dir.glob("smoke_test/**/*smoke_test.rb").each { |file| require_relative file }
end

src/test/java/com/google/api/codegen/testdata/ruby_smoke_test_library.baseline
swcloud: We could do this, but based on current structure of XXXGapicTestSurfaceTransformer, unit test and smoke test were generated together hence they are put into the same place. We can differentiate them based on their postfix such as library_service_smoke_test.rb and libray_service_unit_test.rb

src/test/java/com/google/api/codegen/testdata/ruby_smoke_test_library.baseline
swcloud: Sounds good. I'll make the change

src/test/java/com/google/api/codegen/testdata/ruby_smoke_test_library.baseline
landrito: That works for me.

PR #1016, opened 2017-02-10
replicate Go oneof bug

pongad:
The bug is in Go mock test
```go
var seriesString string = "foobar"
var seriesUuid = &librarypb.SeriesUuid{
    SeriesString: seriesString,
}
```
should actually be
```go
var seriesString string = "foobar"
var seriesUuid = &librarypb.SeriesUuid{
    Source: &librarypb.SeriesUuid_SeriesString{
        SeriesString: seriesString,
    },
}
```

This PR does not contain the fix, only the replication, to make changes in baseline easier to review.

### DIFF
--- a/src/test/java/com/google/api/codegen/testdata/library_gapic.yaml
+++ b/src/test/java/com/google/api/codegen/testdata/library_gapic.yaml
@@ -195,6 +195,7 @@ interfaces:
         - shelf
         - books
         - edition
+        - series_uuid
     required_fields:
       - shelf
       - books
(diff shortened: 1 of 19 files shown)


src/test/java/com/google/api/codegen/testdata/library_gapic.yaml
```diff
@@ -198,6 +198,7 @@ interfaces:
     required_fields:
       - shelf
       - books
+      - series_uuid
```
garrettjonesgoogle: It isn't consistent to have a field in required_fields but not have it in all flattenings.

pongad:
@garrettjonesgoogle PTAL

garrettjonesgoogle:
LGTM

PR #1019, opened 2017-02-10
Change the default lib_version from '0.0' to 'UNKNOWN'.

lukesneeringer:
@geigerj @bjwatson I just realized that there is one place where a potential unknown version is relevant for Python, which is the default `lib_version`. This corrects that.


bjwatson:
Good catch!

lukesneeringer:
Thanks. :-)

PR #1020, opened 2017-02-10
Introduce 'getProjectId' method to NodeJS GAPIC.

### LINKED ISSUE #1018: NodeJS: the generated class should provide the accessor to the project ID
From the request for the developers of google-cloud-node: since the auth library knows the current project ID, the generated class shouldn't block doing the information.

The class will have the method of `getProjectId` with the same interface as in the google-auto-auth library (see https://github.com/stephenplusplus/google-auto-auth#authgetprojectidcallback).

jmuk:
Fixes #1018

### DIFF
--- a/src/main/resources/com/google/api/codegen/nodejs/main.snip
+++ b/src/main/resources/com/google/api/codegen/nodejs/main.snip
@@ -54,6 +54,15 @@
       {@path_templates}
     @end
 
+    /**
+     * Get the project ID used by this class.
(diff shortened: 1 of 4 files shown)


landrito:
Have you confirmed that this works?

jmuk:
Yes, I generated a client locally and it successfully return the project ID.

landrito:
Is there any drawbacks to exposing the auth field?

PR #1021, opened 2017-02-10
Fix the x-goog-api-client header format for Node.JS.

jmuk:


src/test/java/com/google/api/codegen/testdata/nodejs_doc_main_library.baseline
```diff
@@ -106,14 +106,14 @@ function LibraryServiceClient(gaxGrpc, grpcClients, opts) {
   var port = opts.port || DEFAULT_SERVICE_PORT;
   var sslCreds = opts.sslCreds || null;
   var clientConfig = opts.clientConfig || {};
-  var appName = opts.appName || 'gax';
-  var appVersion = opts.appVersion || gax.version;
 
   var googleApiClient = [
-    appName + '/' + appVersion,
+    'gl-node/' + process.versions.node,
+    'gccl/' + require('../../package.json').version,
```
landrito: This should be optional parameters `lib_name` and `lib_version` since the wrapping library is what should define this header key and version. This is because we expect gapic to be wrapped by more than google cloud in the future.

lukesneeringer: To be clear:

  * [ ] Rename `gccl` here to `gapic`.
  * [ ] Add a `libName` and `libVersion` parameter which are inserted between `gl-node` and `gapic`.

jmuk: gapic is given as `CODE_GEN_NAME_VERSION` constant below. 
fixed the `gccl` part. Thanks.

src/main/resources/com/google/api/codegen/nodejs/main.snip
```diff
@@ -243,14 +243,15 @@
...
       var googleApiClient = [
         'gl-node/' + process.versions.node,
-        @if context.isGcloud
-          # TODO: might need to allow customizing the path to the package.json.
-          'gccl/' + require('../../package.json').version,
-        @end
-        CODE_GEN_NAME_VERSION,
+        CODE_GEN_NAME_VERSION
+      ];
+      if (opts.libName && opts.libVersion) {
```
lukesneeringer: Nitpick: This will do the wrong thing if someone `libName` in the `opts` object but not `libVersion`. In a perfect world, if `libName` is set, then you would use it and use `options.libVersion || ''` for the version.

I do not think that is important enough to block on, though.

bjwatson: We should fix this sometime soon, before we forget and it becomes an issue.

jmuk: Hmm, Sorry I missed replying.
I think the current code is okay, here's why.

I believe `libName` and `libVersion` comes only from google-cloud-node hand-written libraries. Under that assumption, I believe `libVersion` needs to exist, otherwise something is wrong on GCN part (or non-GCN code wrongly specifies the code). Either way, it's harmless to silently ignore the case that libVersion is missing.

bjwatson: Ok, I'll buy that.

lukesneeringer: Same.

PR #1024, opened 2017-02-13
fix Go top-level enum

pongad:
also adding @eoogbe since this touches cross-language enum code
and @swcloud since this touches Ruby baseline; I think this is expected behavior, but wanted to make sure.

Go names enums in two different ways.
If the enum is nested in a message, the format is <messagetype>_<enum value>,
respecting the C++ scoping rule used by protobuf,
where enum values are scoped at the same level as the
enum type, not as its children.
On the other hand, if the enum is at top-level,
there is no parent message,
and the format is <enum type>_<enum value>

### DIFF
--- a/src/test/java/com/google/api/codegen/transformer/ModelTypeNameConverterTestUtil.java
+++ b/src/test/java/com/google/api/codegen/transformer/ModelTypeNameConverterTestUtil.java
@@ -18,6 +18,7 @@ import com.google.api.codegen.CodegenTestUtil;
 import com.google.api.tools.framework.model.EnumType;
 import com.google.api.tools.framework.model.MessageType;
 import com.google.api.tools.framework.model.Model;
+import com.google.api.tools.framework.model.ProtoContainerElement;
 import com.google.api.tools.framework.model.ProtoFile;
 import com.google.api.tools.framework.model.TypeRef;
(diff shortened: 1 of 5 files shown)


garrettjonesgoogle:
LGTM

src/test/java/com/google/api/codegen/transformer/ModelTypeNameConverterTestUtil.java
```diff
@@ -18,36 +18,58 @@
...
-      for (MessageType message : file.getMessages()) {
-        if (!message.getSimpleName().equals("Book")) {
-          continue;
+    pathLoop:
+    for (int i = 0; i < path.length; i++) {
+      String pathElement = path[i];
+      for (MessageType message : container.getMessages()) {
+        if (message.getSimpleName().equals(pathElement)) {
+          container = message;
+          continue pathLoop;
```
eoogbe: Can you break it down into multiple methods? Named labels get kinda ugly.

garrettjonesgoogle: Crap I didn't notice this, I take back my LGTM

pongad:
@eoogbe @garrettjonesgoogle PTAL

garrettjonesgoogle:
LGTM

Issue #1022, opened 2017-02-11
googleapis url is broken in README.md from ruby packagemetadata gen


swcloud:
One example is monitoring:

[googleapis]: https://github.com/googleapis/googleapis/tree/master/google/google/cloud/monitoring/v3

landrito:
This issue was caused by an error in packman. After migrating metadata generation over to toolkit, this has been fixed.

PR #1027, opened 2017-02-13
Fixed ruby readme link.

landrito:
The `protoPath` for the library test data does not accurately depict the paths found in google apis which caused us to miss this. When generating clients the link would have the version repeated. 

ie: `https://github.com/googleapis/googleapis/tree/master/google/cloud/speech/v1beta1/v1beta1`

### DIFF
--- a/src/main/resources/com/google/api/codegen/ruby/README.md.snip
+++ b/src/main/resources/com/google/api/codegen/ruby/README.md.snip
@@ -6,7 +6,7 @@
   easy-to-use client library for the [{@metadata.fullName}][] ({@metadata.majorVersion}) defined in the [googleapis][] git repository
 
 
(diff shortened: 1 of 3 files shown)


swcloud:
Does this fix #1022?

landrito:
Correct.

swcloud:
Removing the version number is not the correct way to fix the URL.

With the PR,
https://github.com/googleapis/googleapis/tree/master/google/google/cloud/monitoring

is still incorrect.

The correct one should be:

https://github.com/googleapis/googleapis/tree/master/google/monitoring/v3

landrito:
Have you run the generator patched with this PR? When I run it I am getting this link: 
https://github.com/googleapis/googleapis/tree/master/google/monitoring/v3

`metadata.protoPath` for most apis produces the path starting with google ending with it's version.

PR #1029, opened 2017-02-14
Update Java header

shinfan:


src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTransformer.java
```diff
@@ -373,10 +375,20 @@ private StaticLangSettingsView generateSettingsClass(
...
     xsettingsClass.hasDefaultServiceScopes(interfaceConfig.hasDefaultServiceScopes());
     xsettingsClass.hasDefaultInstance(interfaceConfig.hasDefaultInstance());
+    xsettingsClass.generatorVersion(getGeneratorVersion());
 
     return xsettingsClass.build();
   }
 
+  private static String getGeneratorVersion() {
+    // TODO: Read version information from static file.
+    String version = JavaGapicSurfaceTransformer.class.getPackage().getImplementationVersion();
```
garrettjonesgoogle: Does this work when running through artman? It looks like it doesn't work in the baselines - maybe we need a versioning strategy like I introduced in gax?

shinfan: It only works if the caller uses the toolkit as a package. Yes, to fix this we need the versioning strategy from gax (That's why I added the TODO there). Do you prefer me to add the versioning in this PR or as a follow-up?

garrettjonesgoogle: It's not that much code - you should be able to pull the relevant stuff out of https://github.com/googleapis/gax-java/pull/182 for adding version.txt, populating the resource file in the build output, and reading the resource file at runtime. You wouldn't need to do the other version updating stuff for now.

garrettjonesgoogle: Explanation: I think a TODO could make sense if we wouldn't do it for a while, or if it's a cleanup task that didn't affect the output, but I don't think either apply here.

shinfan: Acknowledged. Done.

shinfan:
@garrettjonesgoogle PTAL

src/test/java/com/google/api/codegen/testdata/java_settings_library.baseline
```diff
@@ -147,6 +147,9 @@ public class LibraryServiceSettings extends ClientSettings {
       .build();
 
 
+  private static final String DEFAULT_GENERATOR_NAME = "gapic";
+  private static final String DEFAULT_GENERATOR_VERSION = "0.0.5";
```
garrettjonesgoogle: I didn't anticipate this before - this will have to be changed every time we bump up our version.

Let's wait until another day to fix it - we aren't frequently bumping up the version right now. I'm thinking maybe we need to support special directives in baselines for "match a dynamic value passed in" for when we do diffs.

shinfan: Sgtm. We might want to do something similar to the license matching, which allows the year number to be dynamic.

garrettjonesgoogle: Can you file an issue?

shinfan: Yup, issue is here: https://github.com/googleapis/toolkit/issues/1034

bjwatson: Go has a similar issue. FYI @pongad and @jba.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTransformer.java
```diff
@@ -85,6 +87,8 @@
   private static final String PAGE_STREAMING_RESPONSE_TEMPLATE_FILENAME =
       "java/page_streaming_response.snip";
 
+  private static final String DEFAULT_VERSION = "UNKNOWN";
```
bjwatson: Is there a PR for fixing this (i.e. changing it to `""`)?

bjwatson: @shinfan I guess I already knew about this PR. :)

Please note that I have comments on this and https://github.com/googleapis/gax-java/pull/203 about changing `"UNKNOWN"` to `""`, which is the decision we agreed upon this week with the service controller folks.

PR #1036, opened 2017-02-15
go: change x-goog header from `grpc/UNKNOWN` to `grpc/`

pongad:


garrettjonesgoogle:
We always have versions for the non-grpc values?

pongad:
@garrettjonesgoogle I think something must have gone seriously wrong if we don't have the versions. Every argument to printf, except from `version.Go()` is either constant string (`static final` in Java parlance) or parameters which constant strings are passed to.

The exception, `version.Go()` just takes a constant string and strips whitespace from it. While not technically a constant, it might as well be.

garrettjonesgoogle:
Alrighty, just wanted to check. LGTM.

PR #1042, opened 2017-02-17
Update PHP header

michaelbausor:
- Move version.txt file read to shared location and use for PHP
- Update data in PHP header

### DIFF
--- a/src/main/java/com/google/api/codegen/GeneratorVersionProvider.java
+++ b/src/main/java/com/google/api/codegen/GeneratorVersionProvider.java
@@ -1,4 +1,4 @@
-/* Copyright 2016 Google Inc
+/* Copyright 2017 Google Inc
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/main/java/com/google/api/codegen/GeneratorVersionProvider.java
+++ b/src/main/java/com/google/api/codegen/GeneratorVersionProvider.java
@@ -12,12 +12,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.google.api.codegen.transformer;
+package com.google.api.codegen;
 
 import java.io.IOException;
 import java.util.Properties;
 
-public class GeneratorVersion {
+public class GeneratorVersionProvider {
 
   private static final String DEFAULT_VERSION = "";
 
@@ -26,7 +26,8 @@ public class GeneratorVersion {
     Properties properties = new Properties();
     try {
       properties.load(
-          GeneratorVersion.class.getResourceAsStream("/com/google/api/codegen/codegen.properties"));
+          GeneratorVersionProvider.class
+              .getResourceAsStream("/com/google/api/codegen/codegen.properties"));
       version = properties.getProperty("version");
     } catch (IOException e) {
       e.printStackTrace(System.err);
--- a/src/main/java/com/google/api/codegen/transformer/GeneratorVersion.java
+++ b/src/main/java/com/google/api/codegen/transformer/GeneratorVersion.java
@@ -1,0 +1,36 @@
+/* Copyright 2016 Google Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen.transformer;
+
+import java.io.IOException;
+import java.util.Properties;
+
+public class GeneratorVersion {
+
+  private static final String DEFAULT_VERSION = "";
+
+  public static String getGeneratorVersion() {
+    String version = DEFAULT_VERSION;
+    Properties properties = new Properties();
+    try {
+      properties.load(
+          GeneratorVersion.class.getResourceAsStream("/com/google/api/codegen/codegen.properties"));
+      version = properties.getProperty("version");
+    } catch (IOException e) {
+      e.printStackTrace(System.err);
+    }
+    return version;
+  }
+}
--- a/src/main/java/com/google/api/codegen/viewmodel/DynamicLangXApiView.java
+++ b/src/main/java/com/google/api/codegen/viewmodel/DynamicLangXApiView.java
@@ -96,6 +96,8 @@ public abstract class DynamicLangXApiView implements ViewModel {
     return missingDefaultServiceAddress() || missingDefaultServiceScopes();
   }
 
+  public abstract String codeGenVersion();
+
   @Override
   public String resourceRoot() {
     return SnippetSetRunner.SNIPPET_RESOURCE_ROOT;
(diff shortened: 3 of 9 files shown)


bjwatson:
Thanks @michaelbausor! Will there be a corresponding https://github.com/googleapis/gax-php PR?

michaelbausor:
@bjwatson Yep just posted, https://github.com/googleapis/gax-php/pull/60

src/main/java/com/google/api/codegen/transformer/GeneratorVersion.java
```diff
@@ -0,0 +1,36 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen.transformer;
+
+import java.io.IOException;
+import java.util.Properties;
+
+public class GeneratorVersion {
```
garrettjonesgoogle: This isn't a transformer (it takes no input!), so I think it should go up one directory in the general area. Since this class doesn't hold data and just provides it, I think it should probably follow the naming convention of "doer" or "X doer" as opposed to "X".

michaelbausor: Done

src/main/java/com/google/api/codegen/GeneratorVersionProvider.java
```diff
@@ -0,0 +1,37 @@
+/* Copyright 2016 Google Inc
```
garrettjonesgoogle: copyright year -> 2017

src/main/java/com/google/api/codegen/viewmodel/DynamicLangXApiView.java
```diff
@@ -96,6 +96,8 @@ public boolean hasMissingDefaultOptions() {
     return missingDefaultServiceAddress() || missingDefaultServiceScopes();
   }
 
+  public abstract String codeGenVersion();
```
garrettjonesgoogle: maybe `toolkitVersion` instead? It doesn't necessarily have to be called that in the generated code, but in the view model classes, I think it makes it clearer that it is the version of toolkit itself.

src/main/java/com/google/api/codegen/GeneratorVersionProvider.java
michaelbausor: Done

src/main/java/com/google/api/codegen/viewmodel/DynamicLangXApiView.java
michaelbausor: Updated view model classes

garrettjonesgoogle:
LGTM

michaelbausor:
@bjwatson Do you want to take a look before merging?

michaelbausor:
Updated with appName->libName update, PTAL

src/main/resources/com/google/api/codegen/php/main.snip
```diff
@@ -332,9 +332,17 @@
...
-     *                              Default: the current version of GAX.
+     *     @@type string $libName The API library software used for calling the service.
+     *                            (Unless you are writing an API client itself, leave this as default.)
+     *     @@type string $libVersion The API library software version used for calling the service.
+     *                               (Unless you are writing an API client itself,
+     *                               leave this as default.)
+     *     @@type string $appName Currently unsupported - reserved for future use.
+     *                            The name of the application calling the service. Recommended
+     *                            for analytics purposes.
+     *     @@type string $appVersion Currently unsupported - reserved for future use.
```
garrettjonesgoogle: Why do we want to keep around $appName/$appVersion?

michaelbausor: This was mirroring Python's choice to keep them in the surface: https://github.com/googleapis/api-client-staging/blob/master/generated/python/gapic-google-cloud-pubsub-v1/google/cloud/gapic/pubsub/v1/publisher_client.py#L140

This may be unnecessary, as we will not break anyone if they keep using appName and appVersion (but the info will be dropped from the header) while python would have an unknown keyword arg if they removed.

bjwatson: Luke is a big fan of avoiding breaking changes wherever possible, so that's why this was done in Python (it's a cultural thing for that language community). Also, we will have a future use case for `$appName/$appVersion`, but we don't want to do anything with these values right now. I'm leaving that for a future design round.

src/test/java/com/google/api/codegen/testdata/php_main_library.baseline
```diff
@@ -435,12 +443,10 @@ class LibraryServiceClient
         }
 
         $headerDescriptor = new AgentHeaderDescriptor([
-            'clientName' => $options['appName'],
-            'clientVersion' => $options['appVersion'],
+            'libName' => $options['libName'],
+            'libVersion' => $options['libVersion'],
             'codeGenName' => self::CODEGEN_NAME,
```
bjwatson: `'codeGenName'` should just be hard-coded as `'gapic'` in GAX. We should not leave it open-ended.

src/test/java/com/google/api/codegen/testdata/php_main_library.baseline
```diff
@@ -401,9 +401,17 @@ class LibraryServiceClient
...
      *                              set the timeout in RetryOptions.
      *                              Default: 30000 (30 seconds)
-     *     @type string $appName The codename of the calling service. Default 'gax'.
-     *     @type string $appVersion The version of the calling service.
-     *                              Default: the current version of GAX.
+     *     @type string $libName The API library software used for calling the service.
+     *                            (Unless you are writing an API client itself, leave this as default.)
+     *     @type string $libVersion The API library software version used for calling the service.
+     *                               (Unless you are writing an API client itself,
+     *                               leave this as default.)
```
bjwatson: We could actually leave `$libName/$libVersion` undocumented, or discourage their use in some other way. Unless we recognize the `$libName` on the server side, we won't do anything with these values.

src/main/java/com/google/api/codegen/GeneratorVersionProvider.java
```diff
@@ -0,0 +1,37 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen;
+
+import java.io.IOException;
+import java.util.Properties;
+
+public class GeneratorVersionProvider {
```
bjwatson: @pongad Does this class provide a useful model for you to manage Go GAPIC versions, rather than remembering to update a value in the `.snip` file?

src/main/resources/com/google/api/codegen/php/main.snip
michaelbausor: I have removed these.

src/test/java/com/google/api/codegen/testdata/php_main_library.baseline
michaelbausor: Happy to leave them undocumented.

src/test/java/com/google/api/codegen/testdata/php_main_library.baseline
michaelbausor: Done in GAX, along with rename to gapicVersion

src/main/java/com/google/api/codegen/GeneratorVersionProvider.java
bjwatson: @swcloud @landrito This might also help with determining the Ruby GAPIC version.

PR #1045, opened 2017-02-20
Special characters in proto comments

chrisdunelm:
With tests, and C# code to perform required escaping.


chrisdunelm:
I don't think we need to wait.
This is a simple change to other language files, and I've seen previous similar changes that affect C# that I haven't been asked to review.
So I'll submit now.

PR #1049, opened 2017-02-21
Migrate Ruby to MVVM

### LINKED ISSUE #349: Refactor Ruby generation to use the MVVM pattern
FYI @jmuk @bjwatson @geigerj

eoogbe:
Refactors Ruby Gapic codegen to MVVM. Closes #349


eoogbe:
@landrito @swcloud @garrettjonesgoogle PTAL

swcloud:
On it right now.

src/main/java/com/google/api/codegen/transformer/ruby/RubyApiMethodParamTransformer.java
```diff
@@ -14,24 +14,127 @@
...
+
+        DynamicLangDefaultableParamView.Builder param =
+            DynamicLangDefaultableParamView.newBuilder();
+        param.name(context.getNamer().getVariableName(field));
+        param.defaultValue("nil");
+        methodParams.add(param.build());
+      }
+    }
+    methodParams.add(
+        DynamicLangDefaultableParamView.newBuilder().name("options").defaultValue("nil").build());
```
garrettjonesgoogle: This is under the control of google-java-format.

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -528,6 +528,11 @@ public String getPagedListResponseFactoryName(Method method) {
     return privateFieldName(Name.upperCamel(method.getSimpleName(), "PagedListResponseFactory"));
   }
 
+  /** The variable name of the gRPC request object. */
+  public String getRequestObjectName(Method method) {
+    return getNotImplementedString("SurfaceNamer.getRequestObjectName");
+  }
```
garrettjonesgoogle: Maybe instead name this `getRequestVariableName` for consistency with other methods implementing the same kind of thing. Also, it seems like you could provide a reasonable default implementation user using `localVarName`, similarly to `getApiWrapperVariableName`.

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -803,6 +808,19 @@ public String getOneofVariantTypeName(OneofConfig oneof) {
...
+   * metadata,
+   */
+  public String getLongRunningOperationTypeName(ModelTypeTable typeTable, TypeRef type) {
+    return getNotImplementedString("SurfaceNamer.getLongRunningOperationTypeName");
+  }
+
+  /** The type name for the gPRC request. */
+  public String getRequestTypeName(ModelTypeTable typeTable, TypeRef type) {
+    return getNotImplementedString("SurfaceNamer.getRequestTypeName");
+  }
```
garrettjonesgoogle: I think this could have a reasonable default implementation.

src/main/java/com/google/api/codegen/transformer/ruby/RubySurfaceNamer.java
```diff
@@ -47,12 +52,18 @@ public RubySurfaceNamer(String packageName) {
...
   /** The name of the class that implements snippets for a particular proto interface. */
+  @Override
   public String getApiSnippetsClassName(Interface interfaze) {
     return publicClassName(Name.upperCamel(interfaze.getSimpleName(), "ClientSnippets"));
   }
 
+  @Override
+  public String getPathTemplateName(
+      Interface service, SingleResourceNameConfig resourceNameConfig) {
+    return Name.from(resourceNameConfig.getEntityName(), "path", "template").toUpperUnderscore();
```
garrettjonesgoogle: What type of thing is this string used for - variable name? something else?

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
eoogbe: Some languages use `req`, others use `request`. Wasn't sure which one would make a better sensible default so I left it out.

src/main/java/com/google/api/codegen/transformer/ruby/RubySurfaceNamer.java
eoogbe: It's used for a constant variable name. The default implementation produces stuff like `ShelfPathTemplate` for Ruby when we need `SHELF_PATH_TEMPLATE`.

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
garrettjonesgoogle: Hmm ok.

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
eoogbe: I guess I could do the PHP implementation, but that assumes that the default uses the type nickname, rather than the full name. So far I have two uses of the method and they're too different to generalize from.

src/main/java/com/google/api/codegen/transformer/ruby/RubySurfaceNamer.java
garrettjonesgoogle: "constant variable" is a self-contradiction. If it's a constant, it seems like wrapping the name with `inittedConstantName` might make sense. The odd thing to me is that `inittedConstantName` seems to use upper camel instead of upper underscore:

https://github.com/googleapis/toolkit/blob/master/src/main/java/com/google/api/codegen/util/ruby/RubyNameFormatter.java#L79

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
garrettjonesgoogle: If it's hard to generalize, that leave it as is.

src/main/java/com/google/api/codegen/transformer/ruby/RubySurfaceNamer.java
eoogbe: constant name*

garrettjonesgoogle:
LGTM

Issue #1048, opened 2017-02-21
Honor "required_fields" in flattened methods


chrisdunelm:
E.g. https://github.com/googleapis/googleapis/blob/master/google/logging/v2/logging_gapic.yaml#L341
This should mean that all non-required flattened-method parameters can be passed null, which causes the relevant field(s) will be left unset on the request object.

- [ ] Java
- [x] C#

eoogbe:
@chrisdunelm Can you add a priority to this?

chrisdunelm:
This is done for C#, and I suspect also done for Java. @garrettjonesgoogle can you confirm.

garrettjonesgoogle:
I don't think we have done this for Java - @michaelbausor do you know?

michaelbausor:
Closing as no longer relevant - if this is still needed, please reopen.

PR #1051, opened 2017-02-22
C# honour required/optional flattened parameters/fields

chrisdunelm:
Updates #1048


src/main/java/com/google/api/codegen/viewmodel/StaticLangApiMethodView.java
```diff
@@ -57,6 +57,15 @@
 
   public abstract List<RequestObjectParamView> methodParams();
 
+  public boolean hasOptionalParams() {
+    for (RequestObjectParamView param : methodParams()) {
+      if (!param.isRequired()) {
+        return true;
+      }
+    }
```
garrettjonesgoogle: Don't put looping logic in the view.

src/main/resources/com/google/api/codegen/csharp/gapic_client.snip
```diff
@@ -833,34 +801,99 @@
...
         public virtual {@method.operationMethod.clientReturnTypeName} {@method.name}(
-            {@objectParams(method.methodParams)}) => {@method.name}(
-                new {@method.serviceRequestTypeName}
+        {@flattenedBody(method, FALSE)}
+    @end
+@end
+
+# method: StaticLangApiMethodView, isPaged: boolean
+@private flattenedBody(method, isPaged)
+    @if method.hasOptionalParams
```
garrettjonesgoogle: Break up the snippet into two, one snippet for each branch

src/test/java/com/google/api/codegen/testdata/csharp_gapic_client_library.baseline
```diff
@@ -1109,13 +1109,18 @@ namespace Google.Example.Library.V1
...
-                    Message = message,
-                },
-                callSettings);
+            CallSettings callSettings = null)
+        {
+            GetShelfRequest request = new GetShelfRequest
+            {
+                Name = name,
+            };
+            if (message != null)
```
jskeet: For reference types other than string, it's fine to just assign unconditionally; a null reference is just "this field is unset".

src/test/java/com/google/api/codegen/testdata/csharp_gapic_client_library.baseline
```diff
@@ -1748,17 +1774,22 @@ namespace Google.Example.Library.V1
         public virtual Task<PublishSeriesResponse> PublishSeriesAsync(
             Shelf shelf,
             IEnumerable<Book> books,
-            uint edition,
+            uint? edition,
```
jskeet: Providing a null value will always logically be the same as setting it to 0 for value types. Arguably clearer though.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_client_library.baseline
```diff
@@ -2590,14 +2652,20 @@ namespace Google.Example.Library.V1
...
+        {
+            ListStringsRequest request = new ListStringsRequest
+            {
+
+                PageToken = pageToken ?? "",
+                PageSize = pageSize ?? 0,
+            };
+            if (name != null)
+            {
+                request.AsResourceName = name;
```
jskeet: Again, we handle null for resource names, I think...

chrisdunelm:
PTAL.
Please re-review, the code has changed considerably since the initial version.
@jskeet as usual please mainly look at the generated C#
@garrettjonesgoogle please check everything else ;)

src/test/java/com/google/api/codegen/testdata/csharp_gapic_snippets_library.baseline
```diff
@@ -2204,6 +2204,195 @@ namespace Google.Example.Library.V1.Snippets
...
+            IEnumerable<int> requiredRepeatedInt32 = new List<int>();
+            IEnumerable<long> requiredRepeatedInt64 = new List<long>();
+            IEnumerable<float> requiredRepeatedFloat = new List<float>();
+            IEnumerable<double> requiredRepeatedDouble = new List<double>();
+            IEnumerable<bool> requiredRepeatedBool = new List<bool>();
+            IEnumerable<TestOptionalRequiredFlatteningParamsRequest.Types.InnerEnum> requiredRepeatedEnum = new List<TestOptionalRequiredFlatteningParamsRequest.Types.InnerEnum>();
+            IEnumerable<string> requiredRepeatedString = new List<string>();
+            IEnumerable<ByteString> requiredRepeatedBytes = new List<ByteString>();
+            IEnumerable<TestOptionalRequiredFlatteningParamsRequest.Types.InnerMessage> requiredRepeatedMessage = new List<TestOptionalRequiredFlatteningParamsRequest.Types.InnerMessage>();
+            IEnumerable<BookName> requiredRepeatedResourceName = new List<string>();
```
chrisdunelm: This is incorrect. #1085 filed.

src/main/java/com/google/api/codegen/transformer/ModelTypeTable.java
```diff
@@ -161,8 +161,8 @@ public String getAndSaveNicknameForContainer(
    * If the given type is not implicitly imported, the add it to the import set, then return the
    * zero value for that type.
    */
-  public String getZeroValueAndSaveNicknameFor(TypeRef type) {
-    return typeNameConverter.getZeroValue(type).getValueAndSaveTypeNicknameIn(typeTable);
+  public String getZeroValueAndSaveNicknameFor(TypeRef type, ZeroValuePurpose purpose) {
```
garrettjonesgoogle: I think I would prefer to introduce a new function instead of an enum for supporting the distinction you want. Something like `getParamZeroValue`.

For my understanding: why is this distinction necessary?

chrisdunelm: OK, I'll change it to use a new method instead.
The distinction is necessary because for lists, maps and bytes in C# we want to have snippets showing how to initialise fields of these types by creating a new List, Dictionary (map), or ByteString in the way a user would do so to populate such a type. However, there are more efficient ways of getting empty instances of these types that we use within the vkit layer when we need a "zero"/default value for an optional field that the user hasn't set.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
```diff
@@ -780,16 +780,16 @@ namespace Google.Example.Library.V1
     public partial class FindRelatedBooksRequest
     {
         /// <summary>
-        /// <see cref="IList{BookName}"/>-typed view over the <see cref="Names"/> resource name property.
+        /// <see cref="ResourceNameList{BookName}"/>-typed view over the <see cref="Names"/> resource name property.
         /// </summary>
-        public IList<BookName> BookNames =>
+        public ResourceNameList<BookName> BookNames =>
```
chrisdunelm: Because to initialise it using Object Initialisation syntax we need access to the Add(IEnumerable<>) method, which is not available on IList. Unfortunately.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_client_library.baseline
chrisdunelm: ack, and agreed this is clearer.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_client_library.baseline
```diff
@@ -2691,8 +2723,8 @@ namespace Google.Example.Library.V1
             CallSettings callSettings = null) => AddCommentsAsync(
                 new AddCommentsRequest
                 {
-                    Name = name,
-                    Comments = { comments },
+                    Name = GaxPreconditions.CheckNotNullOrEmpty(name, nameof(name)),
+                    Comments = { GaxPreconditions.CheckNotNull(comments, nameof(comments)) },
```
chrisdunelm: Yes, "comments" is required here.
If it's required, I think I'm happy that we forbid the user to pass null. I was also wondering about enforcing that it's non-empty, but this possibly seems like a step too far...

src/test/java/com/google/api/codegen/testdata/csharp_gapic_client_library.baseline
```diff
@@ -3758,6 +3784,779 @@ namespace Google.Example.Library.V1
                 LongRunningOperationsClient,
                 callSettings);
 
+        /// <summary>
+        /// Test optional flattening parameters of all types
```
chrisdunelm: It's only a test case, not sure its necessary?

src/test/java/com/google/api/codegen/testdata/csharp_gapic_client_library.baseline
```diff
@@ -3758,6 +3784,779 @@ namespace Google.Example.Library.V1
...
+            IDictionary<int, string> optionalMap,
+            CallSettings callSettings = null) => TestOptionalRequiredFlatteningParamsAsync(
+                new TestOptionalRequiredFlatteningParamsRequest
+                {
+                    RequiredSingularInt32 = requiredSingularInt32,
+                    RequiredSingularInt64 = requiredSingularInt64,
+                    RequiredSingularFloat = requiredSingularFloat,
+                    RequiredSingularDouble = requiredSingularDouble,
+                    RequiredSingularBool = requiredSingularBool,
+                    RequiredSingularEnum = requiredSingularEnum,
```
chrisdunelm: proto requires that the first defined enum value is 0, so I think keeping it like this is OK.

chrisdunelm:
PTAL
@garrettjonesgoogle split method as you requested.

src/test/java/com/google/api/codegen/testdata/csharp_gapic_resourcenames_library.baseline
jskeet: Nope, I'm still confused - this is a read-only, computed property... what do we need to initialize? (Can you refer to a file/line number where this is being used?)

jskeet: Ignore me, I get it now.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpSurfaceNamer.java
```diff
@@ -446,4 +450,73 @@ public String getGrpcStreamingApiReturnTypeName(Method method, ModelTypeTable ty
...
+                "%s.Empty<%i>",
+                elementTypeName);
+        return TypedValue.create(emptyTypeName, "%s()")
+            .getValueAndSaveTypeNicknameIn((CSharpTypeTable) typeNameConverter);
+      } else {
+        return null;
+      }
+    } else {
+      if (type.isPrimitive() || type.isEnum() || type.isRepeated()) {
+        return context.getTypeTable().getInternalZeroValueAndSaveNicknameFor(type);
```
garrettjonesgoogle: I can't tell why "internal" would be appropriate to distinguish this usage of zero values vs the existing usage. If it is a default value for optional fields, it's not really internal, right?

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -1151,4 +1151,20 @@ private static String removeSuffix(String original, String suffix) {
...
+
+  /** Is this type a primitive, according to target language. */
+  public boolean isPrimitive(TypeRef type) {
+    return type.isPrimitive();
+  }
+
+  /** The default value for an optional field, null if no default value required. */
+  public String getOptionalFieldDefaultValue(
+      FieldConfig fieldConfig, MethodTransformerContext context) {
+    return null;
```
garrettjonesgoogle: returning a non-implemented string seems better here.

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -366,7 +366,7 @@ private InitValueView getInitValue(MethodTransformerContext context, InitCodeNod
         default:
           throw new UnsupportedOperationException("unexpected entity name type");
       }
-    } else if (initValueConfig.hasFormattingConfig()) {
+    } else if (initValueConfig.hasFormattingConfig() && !item.getType().isRepeated()) {
```
garrettjonesgoogle: What was the bug that this is fixing?

src/main/java/com/google/api/codegen/transformer/csharp/CSharpSurfaceNamer.java
chrisdunelm: It's internal to vkit. That is, it's not exposed directly to users in snippets, which is where the existing getZeroValue method is used.
What was the rationale of your getParamZeroValue name? I can certainly change it to that if you'd rather.

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
chrisdunelm: It's related to #1085. It doesn't fix that issue, but it leads to still incorrect, but a bit more reasonable, generated code.
The problem at now is that the "zero-value" for a repeated resource-name is a list of string, rather than a list of the strongly-typed resource name.

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
chrisdunelm: OK, I'll change it shortly.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpSurfaceNamer.java
garrettjonesgoogle: Hmm, in that case I might prefer the distinction to be `getSnippetZeroValue` and `getImplZeroValue`.

garrettjonesgoogle: As for the rationale of `getParamZeroValue`: without looking deeper, I was assuming this would appear in the method signature. I have since discovered that it only appears in the implementation.

chrisdunelm: Done

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
chrisdunelm: Done

chrisdunelm:
@garrettjonesgoogle PTAL

garrettjonesgoogle:
LGTM

PR #1053, opened 2017-02-22
Migrate Python samplegen to MVVM

eoogbe:


lukesneeringer:
I actually do not feel all that comfortable reviewing this; I have not done anything nearly this complex in toolkit. @landrito or @geigerj, would you take a look?

src/main/java/com/google/api/codegen/transformer/py/PythonImportSectionTransformer.java
```diff
@@ -0,0 +1,173 @@
...
+
+      ImportFileView.Builder fileImport = ImportFileView.newBuilder();
+      fileImport.moduleName("google.gax");
+      fileImport.types(ImmutableList.of(callOptionsImport.build(), initialPageImport.build()));
+      pageStreamingImports.add(fileImport.build());
+    }
+    return pageStreamingImports.build();
+  }
+
+  private Comparator<ImportFileView> importFileViewComparator() {
```
geigerj: Is this needed to ensure deterministic output? If so, please leave a comment (and describe the comparison process)

src/main/java/com/google/api/codegen/transformer/py/PythonImportSectionTransformer.java
```diff
@@ -0,0 +1,173 @@
...
+    return new Comparator<ImportFileView>() {
+      @Override
+      public int compare(ImportFileView o1, ImportFileView o2) {
+        if (!o1.moduleName().equals(o2.moduleName())) {
+          return o1.moduleName().compareTo(o2.moduleName());
+        }
+
+        if (o1.types().size() != o2.types().size()) {
```
geigerj: This logic seems odd to me: does this mean that if an API adds a few types to a proto file, all of the imports in the GAPIC might get shuffled around? When is it not possible to compare based on the file name alone?

src/main/java/com/google/api/codegen/transformer/py/PythonImportSectionTransformer.java
```diff
@@ -0,0 +1,173 @@
...
+            return attributeName1.compareTo(attributeName2);
+          }
+        }
+
+        return 0;
+      }
+    };
+  }
+
+  private ImportFileView generateAppImport(String fullName, String nickname) {
```
geigerj: Please add a method-level comment, maybe with an example

src/main/java/com/google/api/codegen/transformer/py/PythonImportSectionTransformer.java
```diff
@@ -0,0 +1,173 @@
...
+import com.google.api.codegen.viewmodel.ImportTypeView;
+import com.google.api.tools.framework.model.TypeRef;
+import com.google.common.collect.ImmutableList;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Set;
+import java.util.TreeSet;
+
+public class PythonImportSectionTransformer implements ImportSectionTransformer {
```
geigerj: (Throughout this file)

Prefer to chain builder methods, i.e.,

```
Foo foo = Foo.newBuilder()
    .xxx("a")
    .yyy("b")
    .zzz("c")
    .build();
```

instead of

```
Foo.Builder foo = Foo.newBuilder();
foo.xxx("a");
foo.yyy("b");
foo.zzz("c");
Foo result = foo.build();
```

src/main/java/com/google/api/codegen/transformer/py/PythonModelTypeNameConverter.java
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  @Override
+  public TypeName getTypeNameForResourceNameElementType(
+      FieldConfig fieldConfig, String typedResourceShortName) {
+    throw new UnsupportedOperationException(
+        "getTypeNameForResourceNameElementType not supported by Python");
+  }
+
+  private String protoPackageToPythonPackage(String protoPackage) {
```
geigerj: IIUC this file duplicates some logic in `PythonImportHandler`. Is there a way to avoid the duplication? (Or if PythonImportHandler is going away with full MVVM, add a TODO?)

src/main/java/com/google/api/codegen/transformer/go/GoImportSectionTransformer.java
```diff
@@ -25,7 +27,13 @@
 
 public class GoImportSectionTransformer implements ImportSectionTransformer {
   @Override
-  public ImportSectionView generateImportSection(SurfaceTransformerContext context) {
+  public ImportSectionView generateFileHeaderImportSection(SurfaceTransformerContext context) {
+    return generateImportSection(context.getTypeTable().getImports());
```
eoogbe: `generateImportSection` is the private, generically-named helper function

src/main/java/com/google/api/codegen/transformer/py/PythonImportSectionTransformer.java
```diff
@@ -0,0 +1,173 @@
...
+        context.getNamer().getApiWrapperVariableName(context.getInterface()));
+  }
+
+  private List<ImportFileView> generateProtoImports(
+      MethodTransformerContext context, Iterable<InitCodeNode> specItemNodes) {
+    ModelTypeTable typeTable = context.getTypeTable();
+    Set<ImportFileView> protoImports = new TreeSet<>(importFileViewComparator());
+    for (InitCodeNode item : specItemNodes) {
+      TypeRef type = item.getType();
+      if (!type.isRepeated() && type.isMessage()) {
```
eoogbe: The sample will do
```python
fooElement = foo_pb2.Foo()
foos = [fooElement]
```

src/main/java/com/google/api/codegen/transformer/py/PythonImportSectionTransformer.java
```diff
@@ -0,0 +1,173 @@
...
+        String fullName = typeTable.getFullNameFor(type);
+        String nickname = typeTable.getNicknameFor(type);
+        protoImports.add(generateAppImport(fullName, nickname));
+      } else if (type.isEnum()) {
+        protoImports.add(createImport(context.getApiConfig().getPackageName(), "enums"));
+      }
```
eoogbe: It should've been handled in the original code as well. I'm not sure why the TODO was there. In any case, it looks like it's handled now.

src/main/java/com/google/api/codegen/transformer/py/PythonImportSectionTransformer.java
eoogbe: `types` is a list of `ImportTypeView`. In python, it holds the attribute name and local name.
Say, I have this:
```python
import os
from foo import bar as baz
from google.gax import CallOptions, INITIAL_PAGE
```
The types of `foo` contain `bar as baz`. The types of `google.gax` contain `CallOptions` and `INITIAL_PAGE`. The types of `os` are empty.

If we need to add another attribute to `foo`, say `qux`, it usually gets added like this:
```python
#...
from foo import bar as baz
from foo import qux
#...
```
So, no shuffling. In fact, the filename is typically the same as the attribute name, so most of the sorting is already done lexicographically on file path.

src/main/java/com/google/api/codegen/transformer/py/PythonImportSectionTransformer.java
geigerj: Sorry, I don't follow you. In your example, we still need to import `foo_pb2`. (Or are you saying that there's a implicit precondition that `!item.getType().isRepeated()`? In that case please add a `Preconditions.checkState()` to make this explicit.)

src/main/java/com/google/api/codegen/transformer/py/PythonImportSectionTransformer.java
geigerj: Hm, this still doesn't seem right. The imports should be sorted first by their type (std lib/third party/app), then lexicographically ([like here](https://github.com/googleapis/toolkit/blob/48dc12c24dab21bdb98661293e5dffb79866490d/src/main/java/com/google/api/codegen/py/PythonImportHandler.java#L256)).

src/main/java/com/google/api/codegen/transformer/py/PythonImportSectionTransformer.java
```diff
@@ -0,0 +1,179 @@
...
+      pageStreamingImports.add(fileImport);
+    }
+    return pageStreamingImports.build();
+  }
+
+  /**
+   * Orders the imports by: (1) number of attributes (least to most), (2) module name (A-Z), (3)
+   * attribute name (A-Z)
+   */
+  private Comparator<ImportFileView> importFileViewComparator() {
```
geigerj: Hm, this still doesn't seem right. The imports should be sorted first by their type (std lib/third party/app), then lexicographically ([like here](https://github.com/googleapis/toolkit/blob/48dc12c24dab21bdb98661293e5dffb79866490d/src/main/java/com/google/api/codegen/py/PythonImportHandler.java#L256)).

eoogbe: `types` doesn't refer to std/external/app. The import section already divides the imports by those categories, though the init code section only has app imports. `types` refers to `ImportTypeView`s. Those contain a fullName (attributeName in python) and a nickname (localName in python). This comparator follows the lexicographic sort from `PythonImportHandler`, but doesn't convert everything to a string first.

src/main/java/com/google/api/codegen/transformer/py/PythonImportSectionTransformer.java
eoogbe: `foo_pb2` will be imported because of the first line. It's not used in the second. `isRepeated` is excluded because it includes map types that don't end up showing up.

src/main/java/com/google/api/codegen/transformer/go/GoImportSectionTransformer.java
eoogbe: I think I can separate it into two interfaces, so only the Python one and the standard one needs to implement both.

src/main/java/com/google/api/codegen/transformer/py/PythonImportSectionTransformer.java
```diff
@@ -0,0 +1,173 @@
...
+    return new Comparator<ImportFileView>() {
+      @Override
+      public int compare(ImportFileView o1, ImportFileView o2) {
+        if (o1.types().size() != o2.types().size()) {
```
landrito: I'm failing to find where the python [style guide](https://google.github.io/styleguide/pyguide.html?showone=Imports_formatting#Imports_formatting) states to sort by number of import types. Am I missing something?

src/test/java/com/google/api/codegen/testdata/python_doc_main_library.baseline
```diff
@@ -1473,9 +1473,9 @@ class LibraryServiceClient(object):
           >>> required_repeated_string = []
           >>> required_repeated_bytes = []
           >>> required_repeated_message = []
-          >>> required_repeated_resource_name = api.book_path('[SHELF_ID]', '[BOOK_ID]')
-          >>> required_repeated_resource_name_oneof = api.book_path('[SHELF_ID]', '[BOOK_ID]')
-          >>> required_map = []
+          >>> required_repeated_resource_name = []
```
eoogbe: It's repeated; it should be a list. Not sure why it wasn't a list in the first place.

src/main/java/com/google/api/codegen/transformer/py/PythonImportSectionTransformer.java
eoogbe: It doesn't explicitly say, but that's how every example shows it. `import <module>` is always before `from <module> import <attribute>`. That's how we've been doing it, and checking it makes it easier to sort the attribute names lexicographically.

lukesneeringer: This is one of the relatively few places where I have seen different Python programmers do it different ways. Some people actually put `from <m> import <a>` _above_ `import <m>` to align with the requirement that `from __future__ import anything` be above any other import. Others place that by itself and do what @eoogbe has.

For what it is worth, I think what she has is fine. There is no hard and fast rule here and I have seen a couple different ways to go about it. Consistency is what matters.

For my part, I will probably try to make sure in the stuff that we feed into toolkit that any given module only gets one of the two styles of imports, thus making this sort not matter.

src/main/java/com/google/api/codegen/transformer/py/PythonModelTypeNameConverter.java
```diff
@@ -0,0 +1,245 @@
...
+          .put(Type.TYPE_INT32, "0")
+          .put(Type.TYPE_UINT32, "0")
+          .put(Type.TYPE_SINT32, "0")
+          .put(Type.TYPE_FIXED32, "0")
+          .put(Type.TYPE_SFIXED32, "0")
+          .put(Type.TYPE_STRING, "\'\'")
+          .put(Type.TYPE_BYTES, "b\'\'")
+          .build();
+
+  // TODO (geigerj): Read this from configuration?
```
bjwatson: Yes, let's create an issue for this. Otherwise this will break in six months or so when we add something else to `googleapis-common-protos` and forget to update this. @geigerj will this list be driven from `toolkit` or `artman` when you're done killing `packman`?

src/main/resources/com/google/api/codegen/py/method_sample.snip
```diff
@@ -1,74 +1,61 @@
...
-            {@prompt}
-            {@prompt} # Handle metadata.
-            {@prompt} metadata = response.metadata()
-          @end
-        @end
-      @end
+@snippet sampleCode(apiMethod)
+  Example:
+  @join line : util.getDocLines(sampleCode_(apiMethod))
+    {@""}  >>> {@line}
```
bjwatson: @lukesneeringer Do you have an opinion on prepending our sample lines with `">>> "`? I mainly did this because my predecessor Python TL thought we should, and because it makes it easier for us to run this through doctest. However, it makes it harder for users to copy and paste code.

bjwatson:
@garrettjonesgoogle or @shinfan should also take a look.

src/main/java/com/google/api/codegen/transformer/FeatureConfig.java
```diff
@@ -38,7 +38,7 @@ public boolean enableMixins() {
 
   /** Returns true if streaming APIs are supported. */
   public boolean enableGrpcStreaming() {
-    return false;
+    return true;
```
lukesneeringer: Since moving into mobile is a possibility, it is worth asking: Do Swift and Android Java have this support?

src/main/java/com/google/api/codegen/transformer/py/PythonImportSectionTransformer.java
```diff
@@ -0,0 +1,173 @@
...
+import com.google.api.codegen.viewmodel.ImportTypeView;
+import com.google.api.tools.framework.model.TypeRef;
+import com.google.common.collect.ImmutableList;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Set;
+import java.util.TreeSet;
+
+public class PythonImportSectionTransformer implements InitCodeImportSectionTransformer {
```
eoogbe: I have future plans to use the same class to implement the `ImportSectionTransformer` as some of the private helper methods will be shared. The implementation will look very different from the `StandardImportSectionTransformer` so it doesn't really make sense to extend it.

src/main/java/com/google/api/codegen/transformer/py/PythonApiMethodParamTransformer.java
```diff
@@ -0,0 +1,37 @@
...
+import com.google.api.codegen.viewmodel.ParamDocView;
+import com.google.common.collect.ImmutableList;
+import java.util.List;
+
+public class PythonApiMethodParamTransformer implements ApiMethodParamTransformer {
+  @Override
+  public List<DynamicLangDefaultableParamView> generateMethodParams(
+      MethodTransformerContext context) {
+    // TODO(eoogbe): implement this method when migrating to MVVM
+    return ImmutableList.<DynamicLangDefaultableParamView>of();
```
eoogbe: The methods get called in `DynamicLangApiMethodTransformer`, just the values don't get used. Need to return dummy values, not throw an exception.

src/test/java/com/google/api/codegen/testdata/discoveries/py/py_deploymentmanager.v2.json.baseline
```diff
@@ -668,8 +668,8 @@ request = service.types().list(project=project)
 while request is not None:
     response = request.execute()
 
-    for type2 in response['types']:
-        # TODO: Change code below to process each `type2` resource:
-        pprint(type2)
+    for type_ in response['types']:
+        # TODO: Change code below to process each `type_` resource:
+        pprint(type_)
```
eoogbe: When I fixed the `PythonNameFormatter` to wrap keywords and builtins, it overrided the check in discovery that does a similar thing.

src/main/java/com/google/api/codegen/transformer/FeatureConfig.java
eoogbe: Yeah, my concern is that when we add another language down the line, it won't have gRPC streaming support and we'll have to put this method back in.

src/main/resources/com/google/api/codegen/py/method_sample.snip
lukesneeringer: I am used to seeing samples prepended with `>>>`. I agree it makes it harder to copy and paste, but I see it done relatively consistently, and it makes doctesting easier.

I would not raise a stink if they were absent (this is one of those issues that I do not think really matters), but I think it is fine to have it there. The doctest support makes me even slightly more supportive. Doctesting is not favored in the Python community at this point, but autogeneration is probably a pretty strong case for it.

src/main/java/com/google/api/codegen/transformer/py/PythonImportSectionTransformer.java
bjwatson: Ok, a `TODO(eoogbe)` comment that explains that here in the code would be helpful. It makes sense if you're going the direction of making this an `ImportSectionTransformer`, but that's out of scope for this PR.

src/main/java/com/google/api/codegen/transformer/FeatureConfig.java
bjwatson: Ok, makes sense.

src/main/java/com/google/api/codegen/transformer/py/PythonApiMethodParamTransformer.java
bjwatson: Ok, I see.

src/test/java/com/google/api/codegen/testdata/discoveries/py/py_deploymentmanager.v2.json.baseline
vchudnov-g: @saicheems , you did the python MVVM refactoring for Discovery. Could you take a look to verify this doesn't break us?

saicheems: I think this should be fine. If anything it's better now

src/main/java/com/google/api/codegen/transformer/py/PythonModelTypeNameConverter.java
geigerj: This will be driven from `artman` via configuration in `googleapis`. There is actually [some configuration in `googleapis` already for `googleapis-common-protos`](https://github.com/googleapis/googleapis/blob/master/gapic/packaging/common_protos.yaml), but I think this is actually somewhat different.

This list is not actually "common protos" per se (e.g., IAM is not in `googleapis-common-protos`), but rather a list of proto package names we don't modify to add `cloud` in the path. The packages contained in `googleapis-common-protos` is a subset of this list.

I don't think this needs to be addressed in this PR. Creating https://github.com/googleapis/toolkit/issues/1113 to track.

src/main/java/com/google/api/codegen/transformer/py/PythonModelTypeNameConverter.java
geigerj: (resolved)

bjwatson:
@garrettjonesgoogle @shinfan Do you have time to review these changes to MVVM to support Python samplegen?

src/main/java/com/google/api/codegen/transformer/py/PythonImportSectionTransformer.java
```diff
@@ -0,0 +1,174 @@
...
+
+        return 0;
+      }
+    };
+  }
+
+  /**
+   * Generates an import from the fullName and the nickname of a type.
+   *
+   * <p>Handles the following cases: (1) Module only -- generateAppImport("foo.Bar", "foo.Bar") =>
```
shinfan: Format the doc using <ul></ul>

src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSFeatureConfig.java
```diff
@@ -16,9 +16,4 @@
 
 import com.google.api.codegen.transformer.FeatureConfig;
 
-public class NodeJSFeatureConfig extends FeatureConfig {
-  @Override
-  public boolean enableGrpcStreaming() {
-    return true;
-  }
-}
+public class NodeJSFeatureConfig extends FeatureConfig {}
```
shinfan: We can rename `FeatureConfig` to `DefaultFeatureConfig` so the languages use default values will not need to create an empty subclass.

src/main/java/com/google/api/codegen/transformer/InitCodeImportSectionTransformer.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen.transformer;
+
+import com.google.api.codegen.metacode.InitCodeNode;
+import com.google.api.codegen.viewmodel.ImportSectionView;
+
+/** Generates an ImportSectionView for the InitCodeTransformer. */
+public interface InitCodeImportSectionTransformer {
```
eoogbe: I split them up because Python is the only language that needs special handling of the init code import section, but many languages need special handling of the file header import section. Keeping them split up means I make fewer changes.

shinfan: What exactly changes are needed if you combine these two interfaces? From the architecture perspective it is expensive to have one extra top-level interface which is only used by Python (And it only has one method). We should avoid adding interfaces with similar functionality if not necessary because it is hard to know the difference without looking into the implementation details.

eoogbe: If I combine them again, I have to reimplement the same method in Ruby, Go, Php, and Node. It seems much less messy to have separate interfaces for separate contexts.

garrettjonesgoogle: It seems to me that since most languages just implement both interfaces with a single class, it's easier to manage the code if it has a single interface. It's true that most languages will do nothing, but it would be a small function.

geigerj: I requested this change originally because I didn't like that Python's particularities were forcing a code change in all of the other languages. But I'm fine with deferring to @shinfan and @garrettjonesgoogle if the consensus is to combine the interfaces.

garrettjonesgoogle: It's a "conceptually neutral" change, in the sense that there aren't Python-specific names or concepts in it. Theoretically some other future language could have the same scenario as Python.

eoogbe:
The ImportSectionTransformer issue has been resolve. @shinfan PTAL

lukesneeringer:
Huzzah.

PR #1055, opened 2017-02-22
Accept AsyncOperationFlattenedMethod as example.

lukesneeringer:
One of the new APIs only has a method of type `AsyncOperationFlattenedMethod`, and the static languages build because there are no valid example methods.

@garrettjonesgoogle I am mostly (~80%) sure this change is necessary, but I am _not_ sure that it is sufficient. As I ran into it building a Java library, please take a look and make sure, as I am not the domain expert there.


lukesneeringer:
Looking further: It is definitely not a sufficient change; there also has to be the example documentation produced in `java/main.snip`. I lack the knowledge to do that.

garrettjonesgoogle:
It looks like the following needs to be triangulated:

https://github.com/googleapis/toolkit/blob/master/src/main/resources/com/google/api/codegen/java/main.snip#L23

and

https://github.com/googleapis/toolkit/blob/master/src/main/resources/com/google/api/codegen/java/main.snip#L459

shinfan:
This issue has been fixed by: https://github.com/googleapis/toolkit/pull/1235

lukesneeringer:
@shinfan Great, thanks!

PR #1056, opened 2017-02-23
Ruby: Fixed missing dependency and broken link in Ruby Metadata generator

landrito:
Found some small errors when doing the ruby regeneration.


landrito:
Found a couple of missing lines that were missing in the packman template for the gemspec. PTAL.

Issue #1061, opened 2017-02-24
Enforce oneofs with client-side validation


bjwatson:
### What
For an example proto request message:

```proto3
message SeekRequest {
  string subscription = 1;
  oneof target {
    google.protobuf.Timestamp time = 2;
    string snapshot = 3;
  }
}
```

We should enforce in our GAPIC client code that no more than one field in `oneof target` is allowed to be passed into a method.

### Why
We currently allow the service to enforce this, but the quality of the error message returned by the service might vary across APIs. We can do a better, more consistent job of telling the user what they did wrong in our GAPIC client libs.

Some languages might have an even worse status quo. We think Python arbitrarily selects the last field in a `oneof` block that the user specified. This is undefined behavior that appears to work right now, but that we could "break" at some point post-GA. It would be better to tighten this up to be defined behavior.

### Languages
- [ ] NodeJS
- [ ] PHP
- [x] Python (#1069)
- [ ] Ruby

bjwatson:
Language leads: if this issue does not apply to you, then please remove the label and bullet point above for your language.

garrettjonesgoogle:
We don't put one of fields in parameter lists in Java currently.

jmuk:
ProtoBuf.js reports errors when both of `oneof` are specified. I don't think GAPIC layer should do something for Node.

jskeet:
Removed C#:

- The protobuf message handles oneof appropriately (it only stores a single value, and remembers which field it's associated with, then serializes appropriately)
- So long as we never create a flattening with more than one field from the oneof in, we'd never be generating anything that leads to a problem

bjwatson:
My main concern is that the user either gets a helpful error message when they define more than one field in a oneof block, or that the GAPIC surface for a language makes this impossible.

It sounds like C# is the latter, is that correct @jskeet? @garrettjonesgoogle Is the same true for Java?

@jmuk Is the ProtoBuf.js error message is helpful enough for the user? If so, then I agree we don't need to do anything for Node.

jskeet:
Assuming the GAPIC config has been done separately, it is indeed the latter for C#, and I'd assume that's the case for Java too.

bjwatson:
@jskeet Do you think you can resolve https://github.com/googleapis/toolkit/issues/1057 without introducing the possibility that multiple field in a `oneof` block can be defined by the user?

jskeet:
[...]

PR #1069, opened 2017-02-27
Python: Sending multiple "one of" arguments should raise an exception.

lukesneeringer:
This commit:

  * Adds a (language-independent) `getOneofs` argument to the `MethodConfig` class which returns each applicable `Oneof` object. Unlike an iteration over the methods themselves, this will provide each `Oneof` once and exactly once, and from there you can get the applicable argument names.
  * Adds snippet code in Python to raise `ValueError` if the user provides more than one of a "one of" argument set.

This PR addresses #1061 for Python, and is dependent on googleapis/gax-python#164.

### DIFF
--- a/src/main/resources/com/google/api/codegen/py/main.snip
+++ b/src/main/resources/com/google/api/codegen/py/main.snip
@@ -506,7 +506,7 @@
                         {@defaultMutableValues(optionalParams)}
                     @end
                     @if oneOfParams
-                      {@checkOneOfParams(oneOfParams)}
+                        {@checkOneOfParams(oneOfParams)}
                     @end
                     @# Create the request object.
                     @if requiredParams
--- a/src/main/resources/com/google/api/codegen/py/main.snip
+++ b/src/main/resources/com/google/api/codegen/py/main.snip
@@ -473,7 +473,8 @@
              documentation = methodComments(service, method), \
              methodConfig = context.getApiConfig.getInterfaceConfig(service).getMethodConfig(method), \
              requiredParams = methodConfig.getRequiredFields(), \
-             optionalParams = context.removePageTokenFromFields(methodConfig.getOptionalFields(), methodConfig)
+             optionalParams = context.removePageTokenFromFields(methodConfig.getOptionalFields(), methodConfig), \
+             oneOfParams = methodConfig.getOneofs()
         @if method.getRequestStreaming
             def {@methodName(method)}(
                     self,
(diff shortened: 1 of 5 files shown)


lukesneeringer:
I like this idea. I will make that change shortly.

lukesneeringer:
After investigating, it turns out that what @geigerj wants is not possible because the logical order does not match up. The second "one of" argument is actually [dropped when the `Request` object is created][1], which occurs prior to sending that object to GAX.

  [1]: https://github.com/googleapis/toolkit/issues/1061#issuecomment-282347297

lukesneeringer:
Based on a side conversation with @geigerj, we decided to push some of the logic into GAX, and call into it via. a function that is written in the autogen.

GAX pull request is coming.

src/test/java/com/google/api/codegen/testdata/python_main_library.baseline
```diff
@@ -1374,4 +1404,3 @@ class LibraryServiceClient(object):
             empty_pb2.Empty,
             library_pb2.GetBigBookMetadata,
             options)
-
```
geigerj: I'm assuming this lost newline is due to a change in the testing framework, and not because we are generating files that don't end in newline?

lukesneeringer: I will double-check to be sure.

lukesneeringer: Confirmed; `main.snip` still ends in a newline (same as every other snip file).

src/main/resources/com/google/api/codegen/py/main.snip
```diff
@@ -504,6 +505,10 @@
                     @if {@defaultMutableValues(optionalParams)}
                         {@defaultMutableValues(optionalParams)}
                     @end
+                    @if oneOfParams
+                      {@checkOneOfParams(oneOfParams)}
```
bjwatson: This is using two space indentation, whereas it looks like everything else is four space.

src/test/java/com/google/api/codegen/testdata/python_doc_main_library.baseline
```diff
@@ -695,6 +709,7 @@ class LibraryServiceClient(object):
           :exc:`google.gax.errors.GaxError` if the RPC is aborted.
           :exc:`ValueError` if the parameters are invalid.
         """
+        # Create the request object.
```
bjwatson: These comments fall in the category of unnecessary. They indicate what the code is doing rather than why. See recent TOTT.

src/test/java/com/google/api/codegen/testdata/python_main_library.baseline
geigerj: I don't think it matters what `main.snip` ends in; it matters what `main.snip`'s `generateModule()` snippet ends in. (It seems not to have the newline...)

Either way, not worried about this so long as the actual generated output does end in a newline.

src/main/resources/com/google/api/codegen/py/main.snip
lukesneeringer: Sure enough. Fixed.

PR #1076, opened 2017-03-01
go: properly set headers in LRO methods

pongad:


bjwatson:
@pongad LGTM, but @jba should also sign off. Also, sounds like all the GAPIC code needs to be regenerated... sorry about that.

pongad:
@bjwatson Don't worry. Only speech and spanner :smile:

bjwatson:
ok, cool

bjwatson:
@jba Ping

PR #1079, opened 2017-03-01
Pass generated code version to http headers

### LINKED ISSUE #1077: Define GAPIC version for Ruby client libraries
### What
Currently the Ruby GAPIC client libraries report no version for the `gapic/` token in the `x-goog-api-client` header. This should be fixed.

### Why
The `gapic/` token indicates two things: that a GAPIC layer is being used, and what version it is. We're only satisfying the first requirement.

It's okay if the `gapic/` version is the same as what we use for the `gccl/` version, since Ruby is a source-delivery language. However, `gccl/` should only be present when a manual layer is present and used directly.

The primary reason for consistency across all languages, so that we can avoid special-cases when analyzing usage metrics.

landrito:
resolves #1077

### DIFF
--- a/src/main/java/com/google/api/codegen/gapic/MainGapicProviderFactory.java
+++ b/src/main/java/com/google/api/codegen/gapic/MainGapicProviderFactory.java
@@ -416,7 +416,7 @@ public class MainGapicProviderFactory
 
         providers.add(mainProvider);
         providers.add(clientConfigProvider);
-        providers.add(metadataProvider);
+        //        providers.add(metadataProvider);
 
         if (id.equals(RUBY_DOC)) {
           GapicProvider<? extends Object> messageProvider =
--- a/src/main/java/com/google/api/codegen/gapic/MainGapicProviderFactory.java
+++ b/src/main/java/com/google/api/codegen/gapic/MainGapicProviderFactory.java
@@ -389,7 +389,7 @@ public class MainGapicProviderFactory
             CommonGapicProvider.<Interface>newBuilder()
                 .setModel(model)
                 .setView(new InterfaceView())
-                .setContext(new RubyGapicContext(model, apiConfig))
+                .setContext(new RubyGapicContext(model, apiConfig, packageConfig))
                 .setSnippetSetRunner(
                     new RubySnippetSetRunner<Interface>(SnippetSetRunner.SNIPPET_RESOURCE_ROOT))
                 .setSnippetFileNames(Arrays.asList("ruby/main.snip"))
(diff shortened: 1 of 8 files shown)


src/test/java/com/google/api/codegen/testdata/ruby_doc_main_library.baseline
```diff
@@ -252,7 +252,7 @@ module Library
 
         google_api_client = "gl-ruby/#{RUBY_VERSION}"
         google_api_client << " #{lib_name}/#{lib_version}" if lib_name
-        google_api_client << " gapic/ gax/#{Google::Gax::VERSION}"
+        google_api_client << " gapic/0.1.0 gax/#{Google::Gax::VERSION}"
```
bjwatson: @landrito Is this the version of the code generator or the generated code? We want the latter.

landrito: Generated code found here in the package yaml.

It will eventually be this: [googleapis/googleapis/gapic/packaging/api_defaults.yaml#L35](https://github.com/googleapis/googleapis/blob/master/gapic/packaging/api_defaults.yaml#L35)

bjwatson: Just to verify, the actual version number will be something like `0.6.8`?

landrito: Correct.

landrito: I'll change the test data to reflect this.

landrito: Done.

src/main/java/com/google/api/codegen/gapic/MainGapicProviderFactory.java
```diff
@@ -416,7 +416,7 @@
 
         providers.add(mainProvider);
         providers.add(clientConfigProvider);
-        providers.add(metadataProvider);
+        //        providers.add(metadataProvider);
```
bjwatson: Let's not comment out code, unless there's a TODO comment explaining when we will uncomment it. Otherwise, just delete it.

landrito: Woops thats an artifact from a different repo. Good catch.

landrito: Done!

landrito:
SGTM!

PR #1082, opened 2017-03-02
Update PHP gapic header to client library version

michaelbausor:
When available, we use the libVersion passed by the client library. When not, we check for VERSION file and ServiceBuilder::VERSION.


bjwatson:
LGTM, but @garrettjonesgoogle or @shinfan should also review.

garrettjonesgoogle:
LGTM

shinfan:
LGTM

PR #1083, opened 2017-03-02
go: always report gapic version

pongad:


src/test/java/com/google/api/codegen/testdata/go_main_library.baseline
```diff
@@ -172,7 +172,7 @@ func NewClient(ctx context.Context, opts ...option.ClientOption) (*Client, error
         client: librarypb.NewLibraryServiceClient(conn),
         labelerClient: taggerpb.NewLabelerClient(conn),
     }
-    c.SetGoogleClientInfo("gapic", version.Repo)
+    c.SetGoogleClientInfo("", "")
```
jba: Can't you just omit this line?

src/main/resources/com/google/api/codegen/go/main.snip
```diff
@@ -115,7 +115,11 @@
     // the `x-goog-api-client` header passed on each request. Intended for
     // use by Google-written clients.
     func (c *{@view.clientTypeName}) SetGoogleClientInfo(clientName, clientVersion string) {
-        c.xGoogHeader = fmt.Sprintf("gl-go/%s %s/%s gax/%s grpc/", version.Go(), clientName, clientVersion, gax.Version)
+        if clientName == "" {
```
jba: While you're here, how about changing this to SetGoogleClientVersion(version string), and putting in the gccl yourself? Or if we want more flexibility, at least rename clientName to clientToken.

bjwatson: Let's not hard-code `gccl`. We may use this same code path for other value add libraries later. I'm fine either way with `clientName` or `clientToken`.

src/test/java/com/google/api/codegen/testdata/go_main_library.baseline
bjwatson: +1

pongad: The header string would be null if we omit this right?

bjwatson: Can you just initialize both values to `""` by default?

src/main/resources/com/google/api/codegen/go/main.snip
pongad: The library could benefit from more flexibility here. I'll update a design doc and ping.

src/test/java/com/google/api/codegen/testdata/go_main_library.baseline
pongad: This is called from the "constructor" of the client. I think this is as close to default as we'll get.

bjwatson: Oh, got it. Let's leave this alone then.

src/main/resources/com/google/api/codegen/go/main.snip
bjwatson: Is this good enough to ship for the moment? My understanding is that this is a private interface.

pongad: I have the "correct" implementation almost ready. I think we should hold off on shipping a bit. It will greatly reduce churn.

pongad:
@jba @bjwatson PTAL

bjwatson:
LGTM, but @jba should sign off too.

PR #1086, opened 2017-03-02
Fix PHP headers and unit tests

michaelbausor:
- Fix getGapicVersion function
- Fix unit test generation for files without return type


garrettjonesgoogle:
LGTM

PR #1087, opened 2017-03-03
WIP: Bundling restructure

michaelbausor:
Companion PR to https://github.com/googleapis/gax-java/pull/214


src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -397,6 +397,19 @@ message BundlingSettingsProto {
...
   uint32 request_byte_limit = 5;
+
+  uint32 flow_control_element_limit = 6;
+
+  uint32 flow_control_byte_limit = 7;
+
+  FlowControlLimitExceededBehavior flow_control_limit_exceeded_behavior = 8;
+}
+
+enum FlowControlLimitExceededBehavior {
```
pongad: Can we name this `FlowControlLimitExceededBehaviorProto`? Most types in this file has "Proto" suffix to distinguish it from similar view classes.

src/test/java/com/google/api/codegen/testdata/java_settings_library.baseline
```diff
@@ -921,6 +990,14 @@ public class LibraryServiceSettings extends ClientSettings {
           .setRetryableCodes(RETRYABLE_CODE_DEFINITIONS.get("idempotent"))
           .setRetrySettingsBuilder(RETRY_PARAM_DEFINITIONS.get("default"));
 
+      builder.addCommentsSettings().getBundlingSettingsBuilder()
+          .setElementCountThreshold(0)
+          .setRequestByteThreshold(0)
+          .setDelayThreshold(Duration.millis(0))
```
pongad: IIUC, it is invalid to set these values to zero. They should either be unset or set to positive values.

src/main/java/com/google/api/codegen/transformer/BundlingTransformer.java
```diff
@@ -75,10 +82,32 @@ public BundlingConfigView generateBundlingConfig(MethodTransformerContext contex
...
+            .toString());
 
     return bundlingConfigView.build();
   }
 
+  private static FlowControlLimitBehavior getFlowControlLimitBehavior(
+      FlowControlLimitExceededBehavior behavior) {
+    switch (behavior) {
+      case BLOCK:
+        return FlowControlLimitBehavior.Block;
```
garrettjonesgoogle: It seems a little silly to use an enum just to convert another constant/enum into a String. Maybe just return a String?

src/main/resources/com/google/api/codegen/java/settings.snip
```diff
@@ -316,33 +316,35 @@
...
-              elements.addAll(request.{@desc.bundledFieldGetFunction}());
-            }
-
-            {@desc.requestTypeName} bundleRequest =
-                {@desc.requestTypeName}.newBuilder()
-                    {@fieldCopyCalls(desc.discriminatorFieldCopies)}
-                    .{@desc.bundledFieldSetFunction}(elements)
-                    .build();
-            return bundleRequest;
+          public BundlingDescriptor.RequestBuilder<{@desc.requestTypeName}> getRequestBuilder() {
```
garrettjonesgoogle: I guess this is more of a comment on the GAX change, but I'd prefer not to use inner classes when code outside of the package has to extend/implement it.

michaelbausor:
This PR is replaced by https://github.com/googleapis/toolkit/pull/1095

PR #1090, opened 2017-03-03
Renaming RpcFuture to ApiFuture

garrettjonesgoogle:


src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -730,7 +730,7 @@ public String getGrpcStreamingApiReturnTypeName(Method method, ModelTypeTable ty
 
   /**
    * The generic-aware response type name for the given type. For example, in Java, this will be the
-   * type used for RpcFuture&lt;...&gt;.
+   * type used for Future&lt;...&gt;.
```
vam-google: Is it Future on purpose, or should it be ApiFuture?

garrettjonesgoogle: Yeah I chose `Future` on purpose, because it is just an example for helping people understand the intent.

PR #1091, opened 2017-03-03
Process markdown cloud links for PHP

michaelbausor:
Prepend https://cloud.google.com to cloud site links where necessary.

Borrows from https://github.com/googleapis/toolkit/blob/master/src/main/java/com/google/api/codegen/util/ruby/RubyCommentReformatter.java


michaelbausor:
Ping - sorry I forgot about this one!

michaelbausor:
Followup ping

src/main/java/com/google/api/codegen/util/php/PhpCommentReformatter.java
```diff
@@ -25,6 +27,24 @@
...
+    if (!m.find()) {
+      return comment;
+    }
+    do {
+      String url = "https://cloud.google.com" + m.group(2);
+      // cloud markdown links may contain '$' which needs to be escaped using Matcher.quoteReplacement
+      m.appendReplacement(sb, Matcher.quoteReplacement(String.format("[%s](%s)", m.group(1), url)));
+    } while (m.find());
+    m.appendTail(sb);
+    return sb.toString();
```
garrettjonesgoogle: This is the same as Ruby's, could the implementation be shared somewhere?

src/main/java/com/google/api/codegen/util/CommentReformatting.java
```diff
@@ -0,0 +1,101 @@
...
+    return sb.toString();
+  }
+
+  public static Formatter of(String comment) {
+    return new Formatter(comment);
+  }
+
+  /**
+   * The Formatter object allows the static methods in CommentReformatting to be used via a fluent
+   * interface.
```
garrettjonesgoogle: I would prefer for this class to be opinionated and only provide the fluent interface, instead of both static methods + the fluent interface. Thus, I would suggest lifting Formatter implementation up and making the static methods private. Maybe rename the existing `CommentFormatter` interface to `LanguageCommentFormatter`, so that the name `CommentFormatter` is free.

michaelbausor:
Ping

src/main/java/com/google/api/codegen/util/CommentReformatter.java
```diff
@@ -14,7 +14,79 @@
...
+  public String toString() {
+    return comment;
+  }
+
+  public static Function<Matcher, String> reformatLinkFunction(
+      final String linkFormat, final String urlPrefix) {
+    return new Function<Matcher, String>() {
+      @Override
+      public String apply(Matcher matcher) {
+        String url = urlPrefix + matcher.group(2);
```
garrettjonesgoogle: This has a very hard to follow coupling - it depends on the `Matcher` having the exact right state, e.g. that it was run on the right pattern with the right groupings. I think usage of `Matcher` is best localized to the context where it is created.

michaelbausor: Fixing this involved a fairly large restructure. This functionality is now in ProtoLinkPattern, which couples the patterns/matchers together with the logic using the groupings.

src/main/java/com/google/api/codegen/util/js/JSCommentReformatter.java
```diff
@@ -14,50 +14,22 @@
...
-    do {
-      String url = "https://cloud.google.com" + m.group(2);
-      // cloud markdown links may contain '$' which needs to be escaped using Matcher.quoteReplacement
-      m.appendReplacement(sb, Matcher.quoteReplacement(String.format("[%s](%s)", m.group(1), url)));
-    } while (m.find());
-    m.appendTail(sb);
-    return sb.toString();
+    return CommentTransformer.of(comment)
+        .transform(ProtoLinkPattern.PROTO.createTransformation("{@link %s}"))
+        .transform(ProtoLinkPattern.CLOUD.createTransformation("[%s](%s)"))
```
garrettjonesgoogle: I think the following might be easier to understand:

```
  .transform(PROTO_ELEM_REF.toFormat("{@link $TYPE_REF}"))
  .transform(RELATIVE_URL.toDomain("https://cloud.google.com").toFormat("[$TITLE]($URL)"))
```

Points:
* I don't think that a pattern matcher for relative URLs should know inherently that the URLs are for Cloud - I think that should be informed by the context (Think of non-Cloud APIs)
* I think we should do something that at least looks like named variables; this makes the transformations a lot clearer. Also, it could help remove `Matcher` from the interface of everything.
* The names of the pattern matchers should be more explicit

src/main/java/com/google/api/codegen/util/ruby/RubyCommentReformatter.java
```diff
@@ -55,31 +79,16 @@ public String reformat(String comment) {
...
-    line = reformatCloudMarkdownLinks(line);
-    line = reformatAbsoluteMarkdownLinks(line);
-    line = reformatHeadline(line);
-    return line;
+    return CommentTransformer.of(line)
+        .replace(CommentPatterns.BACK_QUOTE_PATTERN, "+")
+        .transform(PROTO_TO_RUBY_DOC_TRANSFORMATION)
+        .transform(ProtoLinkPattern.CLOUD.createTransformation("{%s}[%s]"))
+        .transform(ProtoLinkPattern.ABSOLUTE.createTransformation("{%s}[%s]"))
+        .transform(HEADLINE_TRANSFORMATION)
```
garrettjonesgoogle: I think this would be more easier to follow:

```
  .scopedReplace(HEADLINE_PATTERN, "#", "=")
```

garrettjonesgoogle:
LBTM (looks beautiful to me)

PR #1092, opened 2017-03-03
Fix gapic/ token to use "gapic package version"

jmuk:


jmuk:
This imitates the pattern used in Ruby

bjwatson:
LGTM, but @landrito should also take a quick look.

PR #1096, opened 2017-03-06
bump protoc to 3.2.0

pongad:
For Go specifically, this lets us write correct import paths for proto files under the google.api package.

I'm tagging all language leads, since this might have cross-language ramifications, even though I do not foresee any specific problems.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -47,7 +47,7 @@ sourceCompatibility = 1.7
 group = "com.google.api"
 project.version = file("version.txt").text.trim()
 
-def protoVersion = "3.0.0-beta-3"
+def protoVersion = "3.2.0"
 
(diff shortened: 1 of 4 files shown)


pongad:
@lukesneeringer I don't think so. This should just affect the generated source code. I'm not sure how Python's "package manifest" is created. If it's still by pacman, this should almost certainly not affect it. (Unless pacman somehow detects the protoc version toolkit uses...?)

It's hard to imagine this breaking. I just want to make sure I'm not being too unimaginative 😆

jmuk:
Fine for nodejs, I believe this will not affect nodejs as well.

garrettjonesgoogle:
LGTM

PR #1098, opened 2017-03-06
Fix API explorer links

geigerj:
Previously, we used shortName, which is configured in the Artman
config. Instead, use discoveryApiName, which is configured by in the
service config.

Note: these are now consistent with README.rst.

cc: @swcloud (since this touches a Ruby template)

### DIFF
--- a/src/main/resources/com/google/api/codegen/py/docs/index.rst.snip
+++ b/src/main/resources/com/google/api/codegen/py/docs/index.rst.snip
@@ -13,7 +13,7 @@
   
   .. _`google-gax`: https://github.com/googleapis/gax-python
   .. _`googleapis`: https://github.com/googleapis/googleapis/tree/master/{@metadata.protoPath}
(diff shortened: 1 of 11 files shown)


lukesneeringer:
Can you explain to me what the real-life difference this is fixing is? The diff in the test failure does not make it obvious. (Also, fix the tests obviously ;-))

geigerj:
Whoops, updated the tests.

@lukesneeringer Both `shortName` and `discoveryApiName` are single-word representations of the API name, but with different origins. `shortName` is comes from the artman config ([example](https://github.com/googleapis/googleapis/blob/master/gapic/api/artman_logging.yaml#L2)). Our team has typically configured this to build the name of the generated GAPIC/gRPC package. It doesn't actually have anything to do with APIs Explorer. `discoveryApiName` comes from the API name in the service config ([example](https://github.com/googleapis/googleapis/blob/master/google/logging/logging.yaml#L5)), and is configured by the API team, not by our team. It's from that name that the actual apis explorer link is derived. (More context: https://github.com/googleapis/toolkit/issues/933)

geigerj:
So, the short answer to the real-life difference is that this should fix some broken links in the docs.

PR #1101, opened 2017-03-06
NodeJS: Migrate Nodejs param documentation to MVVM.

landrito:


src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSApiMethodParamTransformer.java
```diff
@@ -14,24 +14,165 @@
...
-    return ImmutableList.<DynamicLangDefaultableParamView>of();
+    ImmutableList.Builder<DynamicLangDefaultableParamView> methodParams = ImmutableList.builder();
+    methodParams.addAll(generateDefaultableParams(context));
+
+    TypeRef arrayType = TypeRef.fromPrimitiveName("string").makeRepeated();
+
+    DynamicLangDefaultableParamView.Builder optionalArgs =
+        DynamicLangDefaultableParamView.newBuilder();
+    optionalArgs.name(context.getNamer().localVarName(Name.from("optional", "args")));
+    optionalArgs.defaultValue(context.getTypeTable().getZeroValueAndSaveNicknameFor(arrayType));
```
jmuk: why zero value for `arrayType` for the default value of the optional args?

src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSApiMethodParamTransformer.java
```diff
@@ -14,24 +14,165 @@
...
+    return methodParams.build();
+  }
+
+  private List<DynamicLangDefaultableParamView> generateDefaultableParams(
+      MethodTransformerContext context) {
+    if (context.getMethod().getRequestStreaming()) {
+      return ImmutableList.of();
+    }
+    ImmutableList.Builder<DynamicLangDefaultableParamView> methodParams = ImmutableList.builder();
+    for (Field field : context.getMethodConfig().getRequiredFields()) {
```
jmuk: note: this could be rewritten as `Iterables.transform` -- not sure how worthwhile to do so. It's gonna be
```java
return Iterables.transform(context.getMethodConfig().getRequiredFields(), new Function() {
  @Override
  DynamicLangDefaultableParamView apply(Field field) {
    return DynamicDefaultableParamView.newBuilder().name(...).defaultValue("").build();
  }
});
```

(see http://google.github.io/guava/releases/snapshot/api/docs/com/google/common/collect/Iterables.html#transform-java.lang.Iterable-com.google.common.base.Function-)

src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSApiMethodParamTransformer.java
```diff
@@ -14,24 +14,165 @@
...
+            "parameter does not affect the return value. If page streaming is",
+            "performed per-page, this determines the maximum number of",
+            "resources in a page.");
+      } else {
+        docLines.addAll(namer.getDocLines(field));
+      }
+
+      if (field.getType().isMessage() && !field.getType().isMap()) {
+        // Add an empty line if there was doclines added previously
+        if (docLines.build().size() > 0) {
```
jmuk: This line looks like applying immutable pattern wrongly; I believe ImmutableList should be built only once. You should have a boolean flag for this (which should be configured around line 145 based on whether `namer.getDocLines(field)` is empty or not), or you can just use an `ArrayList`.

src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSApiMethodParamTransformer.java
```diff
@@ -14,24 +14,165 @@
...
+          docLines.add("");
+        }
+        docLines.add(
+            "This object should have the same structure as "
+                + namer.getLinkedElementName(field.getType().getMessageType()));
+      } else if (field.getType().isEnum()) {
```
jmuk: Should it be an empty line similar to the line 151?

src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSSurfaceNamer.java
```diff
@@ -153,4 +160,90 @@ public String getFieldGetFunctionName(TypeRef type, Name identifier) {
...
+      ImmutableSet.of(
+          "google/api",
+          "google/bytestream",
+          "google/logging/type",
+          "google/longrunning",
+          "google/protobuf",
+          "google/rpc",
+          "google/type");
+
+  private static final ImmutableMap<Type, String> PRIMITIVE_TYPE_NAMES =
```
jmuk: Very similar or same mapping exists in `NodeJSModelTypeNameConverter`. Can we merge it?

src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSApiMethodParamTransformer.java
landrito: This was an unused artifact from the php transformer I used a template. This is fixed.

src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSApiMethodParamTransformer.java
landrito: Fixed!

src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSApiMethodParamTransformer.java
landrito: Fixed!

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -1027,6 +1027,14 @@ public String getReleaseAnnotation(ReleaseLevel releaseLevel) {
     return getNotImplementedString("SurfaceNamer.getReleaseAnnotation");
   }
 
+  public String getLinkedElementName(ProtoElement element) {
```
landrito: These are used outside of the `NodeJSSurfaceNamer`. I'll look into moving these methods to the JSCommentReformatter.

src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSApiMethodParamTransformer.java
```diff
@@ -14,24 +14,168 @@
...
+    ImmutableList.Builder<ParamDocView> docs = ImmutableList.builder();
+    for (Field field : fields) {
+      if (isRequestTokenParam(methodConfig, field)) {
+        continue;
+      }
+
+      SimpleParamDocView.Builder paramDoc = SimpleParamDocView.newBuilder();
+      paramDoc.paramName("request." + namer.getVariableName(field));
+
+      String typeName = namer.getFieldTypeDoc(context.getTypeTable(), field);
```
eoogbe: can use `getParamTypeName` instead

src/test/java/com/google/api/codegen/testdata/nodejs_doc_main_library.baseline
```diff
@@ -388,9 +388,9 @@ LibraryServiceClient.prototype.createShelf = function(request, options, callback
  *
  * @param {Object} request
  *   The request object that will be sent.
- * @param {string} request.name
+ * @param {String} request.name
```
jmuk: Hmm, this is tiny and maybe harmless update, but we actually prefer lower-cased `string` rather than `String`. It used to be `String` but I changed to `string`.  Probably your code started referring some piece of code unchanged since then.

Please update the name reference to `string` with small s.

src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSApiMethodParamTransformer.java
landrito: Fixed! Good catch!

src/test/java/com/google/api/codegen/testdata/nodejs_doc_main_library.baseline
landrito: Ah it seems the primitive type map in the type table was a bit off. Fixed!

src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSApiMethodParamTransformer.java
landrito: Ack. I feel this is more readable until we can fully support Java 8.

src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSApiMethodParamTransformer.java
```diff
@@ -14,24 +14,171 @@
...
+import com.google.api.codegen.util.js.JSCommentReformatter;
 import com.google.api.codegen.viewmodel.DynamicLangDefaultableParamView;
 import com.google.api.codegen.viewmodel.ParamDocView;
+import com.google.api.codegen.viewmodel.SimpleParamDocView;
+import com.google.api.tools.framework.model.Field;
 import com.google.common.collect.ImmutableList;
 import java.util.List;
 
 public class NodeJSApiMethodParamTransformer implements ApiMethodParamTransformer {
+  private static final JSCommentReformatter jsCommentReformatter = new JSCommentReformatter();
```
eoogbe: Nit: s/jsCommentReformatter/commentReformatter

PR #1103, opened 2017-03-06
NodeJS: Migrate Nodejs file header to MVVM.

landrito:


src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSImportSectionTransformer.java
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  private ImportFileView createImport(String localName, String moduleName) {
+    return ImportFileView.newBuilder()
+        .types(
+            Collections.singletonList(
+                ImportTypeView.newBuilder()
+                    .nickname(localName)
+                    .fullName("")
+                    .type(ImportType.AliasedImport)
```
eoogbe: These aren't really `AliasedImport`s if there's only a full name. In other cases when there's only one name for the `ImportTypeView`, we have set the name to the `fullName` and set the nickname to the empty string. We should probably just make the nickname `Nullable` to avoid confusion.

landrito: Fixed. Removed Aliased imports (to use the default simpleImport). Changed the name to be the fullname and made the nickname nullable.

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -990,6 +990,10 @@ public String getGrpcClientImportName(Interface service) {
     return getNotImplementedString("SurfaceNamer.getGrpcClientImportName");
   }
 
+  public String getClientConfigImportModule(Interface service) {
```
eoogbe: Please reuse `getClientConfigPath`.

landrito: Ah, saw an implementation for `getClientConfigPath` that didn't match what was needed. Found that it wasn't used elsewhere in the project. 

This is fixed.

src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSImportSectionTransformer.java
```diff
@@ -0,0 +1,54 @@
...
+    imports.add(createImport("gax", "google-gax"));
+    if (new GrpcStubTransformer().generateGrpcStubs(context).size() > 1) {
+      imports.add(createImport("merge", "lodash.merge"));
+    }
+    return imports.build();
+  }
+
+  private ImportFileView createImport(String localName, String moduleName) {
+    return ImportFileView.newBuilder()
+        .types(Collections.singletonList(ImportTypeView.newBuilder().fullName(localName).build()))
```
jmuk: This is nitpicking, but I think you should use `nickname` for the local name. 

The import line like `var gax = require('google-gax');` case, the `'google-gax'` is more like a module name, and `gax` part would be the nickname of the import.

full name would be used for the patterns like `var expect = require('chai').expect` -- it's not used in gapic nodejs right now, in that case nickname `expect` will be extracted from the module `chai` and the full-name in the module. You don't have to care about this, but I'm thinking this pattern will justify the use of `nickname` here.

This means nickname is still not `Nullable`.

eoogbe: Fair enough. If there's another pattern where the full name would be used, then that makes sense.

Nickname can probably stay nullable because other languages use full name instead of nickname. Currently, they set it to the empty string, but really they don't exist in simple imports.

landrito: Switched to using nickname. Kept nickname nullable.

PR #1109, opened 2017-03-07
Add class_name setting, update Java generation

michaelbausor:
Mass changes due to signature change of getApiWrapperClassName to accept InterfaceConfig instead of Interface

### DIFF
--- a/src/test/java/com/google/api/codegen/testdata/library_gapic.yaml
+++ b/src/test/java/com/google/api/codegen/testdata/library_gapic.yaml
@@ -3,6 +3,8 @@ language_settings:
   java:
     package_name: com.google.gcloud.pubsub.spi
     release_level: BETA
+    class_names:
+      google.example.library.v1.LibraryService: LibraryClient
   python:
     package_name: google.cloud.gapic.example.library.v1
   go:
--- a/src/test/java/com/google/api/codegen/testdata/library_gapic.yaml
+++ b/src/test/java/com/google/api/codegen/testdata/library_gapic.yaml
@@ -3,8 +3,8 @@ language_settings:
   java:
     package_name: com.google.gcloud.pubsub.spi
     release_level: BETA
-    class_names:
-      google.example.library.v1.LibraryService: LibraryClient
+    interface_names:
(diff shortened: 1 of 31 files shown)


src/test/java/com/google/api/codegen/testdata/library_gapic.yaml
```diff
@@ -3,6 +3,8 @@ language_settings:
   java:
     package_name: com.google.gcloud.pubsub.spi
     release_level: BETA
+    class_names:
+      google.example.library.v1.LibraryService: LibraryClient
```
garrettjonesgoogle: The settings class isn't getting renamed. The two of them need to be in sync.

I suggest a different approach where the base name is given. Something like:

```
  interface_names:
    google.example.library.v1.LibraryService: Library
```

And this generates `LibraryClient` and `LibrarySettings` for classes, and `libraryClient` and `librarySettings` for variable names.

shinfan: What is the major use case of this field? Is it only for renaming interface names or any class names?

michaelbausor: It is for renaming from LibraryServiceClient->LibraryClient, and all other appropriate classes/variables.

I had hoped to allow for overriding the standard Client suffix as well, but you are right that this change must apply to the settings classes, and test class. Have updated accordingly.

michaelbausor:
PTAL

src/test/java/com/google/api/codegen/testdata/library_gapic.yaml
shinfan: Calling it interface_names indicate it only overrides class names for interfaces. I would prefer to keep the field name as "class_names"

garrettjonesgoogle: That's what I wanted! I don't think we should use this for other renames. This rename applies to things besides classes anyway (variable names).

garrettjonesgoogle:
LGTM

PR #1115, opened 2017-03-15
WIP go: unclash LRO types

pongad:
Previously LRO types are named after their result types.
This causes problems if two methods use LROs that
have the same result type but different metadata types.

This commit names LRO types after method name instead.

Do not merge yet; still waiting for discussion to resolve.

### DIFF
--- a/src/test/java/com/google/api/codegen/testdata/go_main_library.baseline
+++ b/src/test/java/com/google/api/codegen/testdata/go_main_library.baseline
@@ -830,7 +830,7 @@ func (it *StringIterator) takeBuf() interface{} {
 }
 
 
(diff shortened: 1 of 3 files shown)


src/test/java/com/google/api/codegen/testdata/go_main_library.baseline
```diff
@@ -830,18 +830,18 @@ func (it *StringIterator) takeBuf() interface{} {
 }
 
 
-// BookOperation manages a long-running operation yielding librarypb.Book.
-type BookOperation struct {
+// GetBigBookOperation manages a long-running operation yielding librarypb.Book.
```
pongad: maybe "manages a long-running operation from GetBigBook method"?

PR #1117, opened 2017-03-16
NodeJS: Migrate Grpc Doc Gen to MVVM

landrito:
#### Still need to be completed.
- [x] Remove NodeJSGapicContext when NodeJS MVVM Surface Gen happens.


src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSSurfaceNamer.java
```diff
@@ -336,4 +341,66 @@ private String getParamTypeNoCardinality(ModelTypeTable typeTable, TypeRef type)
...
+        // Add a break if there was field docs and an extra description.
+        lines.add("");
+      }
+      lines.add(extraFieldDescription);
+    }
+    return lines.build();
+  }
+
+  private String getExtraFieldDescription(Field field) {
+    boolean fieldIsMap = field.getType().isMessage() && !field.getType().isMap();
```
eoogbe: called `fieldIsMap`, but checks if `!isMap`. I think you want `fieldIsMessage`

landrito: Good catch. Done!

src/test/java/com/google/api/codegen/testdata/nodejs_doc_message_library.baseline
```diff
@@ -999,9 +1001,13 @@ var Book = {
    * @enum {number}
    */
   Rating: {
-    GOOD: 0,
+    /**
+     * GOOD enum description
+     */  GOOD: 0,
```
jmuk: Probably line breaking rule has been wrong (or at least weird) here, could you make a fix on the new template?  There should be a linebreak between the comment and the definition, like
```javascript
/**
 * GOOD enum description
 */
GOOD: 0,
```

landrito: Done!

landrito:
Just removed the leftover files after the migration. PTAL.

src/main/java/com/google/api/codegen/viewmodel/GrpcMessageDocView.java
```diff
@@ -21,12 +21,20 @@
...
+
   public abstract List<String> lines();
 
   public abstract List<ParamDocView> properties();
 
   public abstract List<GrpcElementDocView> elementDocs();
 
+  public abstract String fileUrl();
+
+  public boolean hasProperties() {
```
eoogbe: Unnecessary method. Can just check `@if message.properties` in the snippet.

landrito: Done!

PR #1118, opened 2017-03-17
go: small codegen fixes

pongad:
Previously we unit-test error cases with error Canceled.
However, since pubsub is configured to retry on this error,
we retry forever and never gives up.
This commit changes the error to PermissionDenied instead,
since it seems unlikely that there's a retry case for it.

This commit also addresses a naming inconsistency.
Unlike Java, Go spells "canceled", "canceling", etc
with only one L.

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/go/GoSurfaceNamer.java
+++ b/src/main/java/com/google/api/codegen/transformer/go/GoSurfaceNamer.java
@@ -234,7 +234,11 @@ public class GoSurfaceNamer extends SurfaceNamer {
 
   @Override
   public String getStatusCodeName(Status.Code code) {
-    return publicFieldName(Name.upperUnderscore(code.toString()));
+    String codeString = code.toString();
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/api/codegen/transformer/go/GoSurfaceNamer.java
```diff
@@ -234,7 +234,11 @@ Name clientNamePrefix(Interface service) {
 
   @Override
   public String getStatusCodeName(Status.Code code) {
-    return publicFieldName(Name.upperUnderscore(code.toString()));
+    String codeString = code.toString();
+    if (code.equals(Status.Code.CANCELLED)) {
+      codeString = "CANCELED";
```
pongad: There's currently no test for this, but it seems silly to add a CANCELLED to some method, since it will change baselines for all languages. @garrettjonesgoogle WDYT?

PR #1123, opened 2017-03-20
Ruby: Add updater-proc as a constructor param.

landrito:
This will allow users to specify an updater proc to transform metadata for requests. There has been requests for this in order to allow users to add authentication parameters to their requests.

### DIFF
--- a/src/main/resources/com/google/api/codegen/ruby/main.snip
+++ b/src/main/resources/com/google/api/codegen/ruby/main.snip
@@ -148,6 +148,12 @@
   @#   A Channel object through which to make calls.
   @# @@param chan_creds [Grpc::ChannelCredentials]
   @#   A ChannelCredentials for the setting up the RPC client.
+  @# @@param updater_proc [Proc]
+  @#   A function that transforms the metadata for requests, e.g., to give
+  @#   OAuth credentials.
+  @# @@param scopes [Array<String>]
+  @#   The OAuth scopes for this service. This parameter is ignored if
+  @#   an updater_proc is supplied.
   @# @@param client_config[Hash]
   @#   A Hash for call options for each method. See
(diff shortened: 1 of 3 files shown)


landrito:
@jmuk was there a specific reason the scopes parameter was not being documented?

jmuk:
I don't remember why `scopes` hasn't been documented -- probably I missed that with no reasons. LGTM.

src/main/resources/com/google/api/codegen/ruby/main.snip
```diff
@@ -148,6 +148,12 @@
   @#   A Channel object through which to make calls.
   @# @@param chan_creds [Grpc::ChannelCredentials]
   @#   A ChannelCredentials for the setting up the RPC client.
+  @# @@param updater_proc [Proc]
+  @#   A function that transforms the metadata for requests, e.g., to give
```
swcloud: I'm curious why it's called 'updater_proc', instead of 'authenticator_proc'?

src/main/resources/com/google/api/codegen/ruby/main.snip
```diff
@@ -148,6 +148,12 @@
...
+  @#   OAuth credentials.
+  @# @@param scopes [Array<String>]
```
swcloud: Surprised we missed the comment for this parameter.

src/main/resources/com/google/api/codegen/ruby/main.snip
landrito: I think the terminology of updater_proc is from `googleauth`: https://github.com/google/google-auth-library-ruby/blob/37ed189b2e5165243918fcde127ebe323e9a06b5/lib/googleauth/signet.rb#L58

PR #1127, opened 2017-03-22
Use .json extension for .overrides files

### LINKED ISSUE #602: Overrides files cleanup: directory, extension
- The overrides files are used in production, not just in tests, so they should be moved out of the testdata directory.
- The overrides files are valid json, so maybe instead of being of the form `api.version.json.overrides` they should be `api.version.overrides.json`. This will require slightly more processing on the calling side, but not much (in bash: using `%` in variable substitution; in Go, stripping the extension returned by path.Ext())

saicheems:
Resolves #602


saicheems:
Note that this will require some changes elsewhere, and I'll get that done once this is approved.

vchudnov-g:
This change looks alright, but will this test pass without  more changes to toolkit to look for the overrides file? You'll get those done as part of this PR?

vchudnov-g:
OK, the overrides are always explicitly specified when toolkit  code is invoked.

LGTM then

PR #1132, opened 2017-03-23
Change test status code to INVALID_ARGUMENT

michaelbausor:
Pubsub has been updated to retry on INTERNAL errors for some methods - this breaks our tests, as we are not expecting retries. (https://github.com/googleapis/googleapis/blob/master/google/pubsub/v1/pubsub_gapic.yaml#L401)

This changes the code returned by our exception tests to something that should never be retried, i.e. a client-side error, INVALID_ARGUMENT


garrettjonesgoogle:
LGTM

PR #1134, opened 2017-03-24
Fix gapic version tracking in x-goog-api header

shinfan:
The original approach uses `getImplementationVersion()` which sometimes returns null at runtime. Switch to use pom
version (Same as `ServiceOptions`)

Test: 
Baseline tests.
Manually verified e2e

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceNamer.java
+++ b/src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceNamer.java
@@ -34,8 +34,10 @@ import com.google.api.codegen.viewmodel.ServiceMethodType;
 import com.google.api.tools.framework.model.Field;
 import com.google.api.tools.framework.model.Interface;
 import com.google.api.tools.framework.model.Method;
+import com.google.api.tools.framework.model.Model;
 import com.google.api.tools.framework.model.TypeRef;
 import com.google.common.base.Joiner;
+import com.google.common.base.Splitter;
 import java.io.File;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -195,5 +197,11 @@ public class JavaSurfaceNamer extends SurfaceNamer {
   @Override
   public String getBatchingDescriptorConstName(Method method) {
     return inittedConstantName(Name.upperCamel(method.getSimpleName()).join("batching_desc"));
+  }
+
+  @Override
+  public String getArtifactId(Model model) {
+    List<String> nameTokens = Splitter.on(".").splitToList(model.getServiceConfig().getName());
+    return "google-cloud-" + nameTokens.get(0);
   }
 }
--- a/src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceNamer.java
+++ b/src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceNamer.java
@@ -34,10 +34,8 @@ import com.google.api.codegen.viewmodel.ServiceMethodType;
 import com.google.api.tools.framework.model.Field;
 import com.google.api.tools.framework.model.Interface;
 import com.google.api.tools.framework.model.Method;
-import com.google.api.tools.framework.model.Model;
 import com.google.api.tools.framework.model.TypeRef;
 import com.google.common.base.Joiner;
-import com.google.common.base.Splitter;
 import java.io.File;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -197,11 +195,5 @@ public class JavaSurfaceNamer extends SurfaceNamer {
   @Override
   public String getBatchingDescriptorConstName(Method method) {
     return inittedConstantName(Name.upperCamel(method.getSimpleName()).join("batching_desc"));
-  }
-
-  @Override
-  public String getArtifactId(Model model) {
-    List<String> nameTokens = Splitter.on(".").splitToList(model.getServiceConfig().getName());
-    return "google-cloud-" + nameTokens.get(0);
   }
 }
--- a/src/main/resources/com/google/api/codegen/java/settings.snip
+++ b/src/main/resources/com/google/api/codegen/java/settings.snip
@@ -83,6 +83,8 @@
   private static final String GROUP_ID = "com.google.cloud";
   private static final String META_FILE_ROOT = "/META-INF/maven/";
   private static final String META_VERSION_KEY = "version";
+
+  private static String gapicVersion;
   {@""}
 @end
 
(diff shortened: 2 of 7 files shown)


garrettjonesgoogle:
@michaelbausor could you do the first pass of this PR?

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceNamer.java
```diff
@@ -196,4 +198,10 @@ public String getReleaseAnnotation(ReleaseLevel releaseLevel) {
   public String getBatchingDescriptorConstName(Method method) {
     return inittedConstantName(Name.upperCamel(method.getSimpleName()).join("batching_desc"));
   }
+
+  @Override
+  public String getArtifactId(Model model) {
+    List<String> nameTokens = Splitter.on(".").splitToList(model.getServiceConfig().getName());
```
michaelbausor: Is this the standard way of getting the API name in toolkit? It would be nice if we had access to the `api_name` in the artman config: https://github.com/googleapis/googleapis/blob/master/gapic/api/artman_pubsub.yaml#L2

src/test/java/com/google/api/codegen/testdata/java_settings_library.baseline
```diff
@@ -423,8 +429,21 @@ public class LibrarySettings extends ClientSettings {
...
-    return packageVersion != null ? packageVersion : DEFAULT_GAPIC_VERSION;
+    try {
+      Properties properties = new Properties();
+      String mavenPropertiesPath = META_FILE_ROOT
+          + GROUP_ID + "/"
+          + ARTIFACT_ID + "/pom.properties";
+      InputStream inputStream =
+          ImageAnnotatorSettings.class.getResourceAsStream(mavenPropertiesPath);
+      if (inputStream != null) {
+        properties.load(inputStream);
```
michaelbausor: We should do this once, not every time we call `getGapicVersion()`

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceNamer.java
shinfan: I was hoping to use that exact field but unfortunately I don't think there is access to the artman config data from toolkit (these configs files are designed for artman usage)

src/test/java/com/google/api/codegen/testdata/java_settings_library.baseline
shinfan: Done

shinfan:
@michaelbausor PTAL

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceNamer.java
michaelbausor: Fair enough. We may want to file an issue to make this information available in toolkit, either via the command line or a gapic yaml setting.

src/main/resources/com/google/api/codegen/java/settings.snip
```diff
@@ -144,21 +146,22 @@
...
-          + GROUP_ID + "/"
-          + ARTIFACT_ID + "/pom.properties";
-      InputStream inputStream =
-          ImageAnnotatorSettings.class.getResourceAsStream(mavenPropertiesPath);
-      if (inputStream != null) {
-        properties.load(inputStream);
-        return properties.getProperty(META_VERSION_KEY, "");
+    if (gapicVersion == null) {
+      try {
+        Properties properties = new Properties();
```
michaelbausor: Maybe not necessary in this PR, but it feels like the loading logic here could be handled by generalizing GaxPropertiesProvider (https://github.com/googleapis/gax-java/blob/master/src/main/java/com/google/api/gax/core/GaxPropertiesProvider.java)

That way we have less logic in generated code.

michaelbausor:
LGTM, wait for @garrettjonesgoogle and @bjwatson to review.

src/main/resources/com/google/api/codegen/java/settings.snip
shinfan: GaxPropertiesProvider is currently used for retrieving data from gax.properties. We can consider renaming it to something like `MetadataProvider` to support more general usage.

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceNamer.java
garrettjonesgoogle: @shinfan could you file the issue?

src/main/resources/com/google/api/codegen/java/settings.snip
```diff
@@ -140,8 +146,22 @@
...
   private static String getGapicVersion() {
-    String packageVersion = {@xsettingsClass.name}.class.getPackage().getImplementationVersion();
-    return packageVersion != null ? packageVersion : DEFAULT_GAPIC_VERSION;
+    if (gapicVersion == null) {
+      try {
+        Properties properties = new Properties();
+        String mavenPropertiesPath =
+            META_FILE_ROOT + GROUP_ID + "/" + ARTIFACT_ID + "/pom.properties";
+        InputStream inputStream =
+            ImageAnnotatorSettings.class.getResourceAsStream(mavenPropertiesPath);
```
garrettjonesgoogle: You have a fixed class name in here (`ImageAnnotatorSettings`) - please run the tests in api-client-staging with your generated code before submitting.

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceNamer.java
shinfan: Done. https://github.com/googleapis/toolkit/issues/1137

src/main/resources/com/google/api/codegen/java/settings.snip
shinfan: Issue filed against gax: https://github.com/googleapis/gax-java/issues/252

src/main/resources/com/google/api/codegen/java/settings.snip
```diff
@@ -79,6 +79,12 @@
 
   private static final String DEFAULT_GAPIC_NAME = "gapic";
   private static final String DEFAULT_GAPIC_VERSION = "";
+  private static final String ARTIFACT_ID = "{@xsettingsClass.artifactId}";
+  private static final String GROUP_ID = "com.google.cloud";
```
garrettjonesgoogle: I'm not comfortable with com.google.cloud being fixed in the templates - remember, we will be generating code for APIs outside of cloud in not too long.

src/main/resources/com/google/api/codegen/java/settings.snip
```diff
@@ -79,6 +79,12 @@
...
+  private static final String META_FILE_ROOT = "/META-INF/maven/";
```
garrettjonesgoogle: At a more general level, I'm not comfortable with assuming that our generated code will be compiled with maven.

shinfan: Our generated code will most likely be used in google-cloud-java package which uses maven. I can make it default to use the `getImplementationVersion()` approach and fall back to use the maven version.

src/main/resources/com/google/api/codegen/java/settings.snip
shinfan: Right this is hard-coded and won't work for non-cloud APIs. But right now there is no way for toolkit to know the maven group ID. We probably need to add it as a field in the API gapic config yaml.

src/main/resources/com/google/api/codegen/java/settings.snip
```diff
@@ -147,18 +146,20 @@
...
+          Properties properties = new Properties();
+          InputStream inputStream =
+              {@xsettingsClass.name}.class.getResourceAsStream(PROPERTIES_FILE);
+          if (inputStream != null) {
+            properties.load(inputStream);
+            gapicVersion = properties.getProperty(META_VERSION_KEY, DEFAULT_GAPIC_VERSION);
+          }
+        } catch (Exception ignore) {
+          // Ignore
+        }
```
garrettjonesgoogle: Could we maybe make a gax helper function so that this same code doesn't have to be generated for every API?

shinfan: Done https://github.com/googleapis/gax-java/pull/253

garrettjonesgoogle:
LGTM

PR #1139, opened 2017-03-28
Update generated page response

michaelbausor:
Work in progress, wanted to get feedback on the approach. Each generated paged response object and page object has a lot more methods that before, but the implementations are very simple, with as much work as possible being delegated to the context object.


src/test/java/com/google/api/codegen/testdata/java_page_streaming_response_library.baseline
```diff
@@ -47,43 +47,282 @@ import javax.annotation.Generated;
 @ExperimentalApi
 public class PagedResponseWrappers {
 
-  public static class ListShelvesPagedResponse extends PagedListResponseImpl<ListShelvesRequest, ListShelvesResponse, Shelf> {
+  public static class ListShelvesPagedResponse implements PagedListResponse<Shelf> {
+
+    private final PagedListResponseContext<RequestT, ResponseT, ResourceT> context;
```
garrettjonesgoogle: I don't think you can parameterize a member variable with more parameters than are present in the type. I think you need the specific types here: `<ListShelvesRequest, ListShelvesResponse, Shelf>`.

src/test/java/com/google/api/codegen/testdata/java_page_streaming_response_library.baseline
```diff
@@ -47,43 +47,282 @@ import javax.annotation.Generated;
...
+      return new PageContext.PageIterator<ListShelvesPage>(
+        new PageContext.PageFetcher<>() {
+          @Override
+          public ListShelvesPage getNextPage(ListShelvesPage currentPage) {
+            return currentPage.getNextPage();
+          }
+      }
+    }
+
+    public Object getNextPageToken() {
```
garrettjonesgoogle: I think there was going to be some kind of standardization on the page token type, instead of allowing for both String and ByteString. Let's figure out what it is. 

Related: https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1764

src/test/java/com/google/api/codegen/testdata/java_page_streaming_response_library.baseline
michaelbausor: Yes, this is just an error on my part, it is meant to be the specific types

src/test/java/com/google/api/codegen/testdata/java_page_streaming_response_library.baseline
michaelbausor: The style guide says string: https://cloud.google.com/apis/design/design_patterns#list_pagination

And all APIs in googleapis use string. So I think we should go with that.

src/main/java/com/google/api/codegen/viewmodel/StaticLangPagedResponseView.java
```diff
@@ -22,6 +22,10 @@
 
   public abstract String name();
 
+  public abstract String pageName();
```
garrettjonesgoogle: This should be `pageTypeName` since it is the name of a type.

garrettjonesgoogle:
Thanks for catching more that I didn't notice - LGTM

PR #1141, opened 2017-03-28
Add exception testing to Python

eoogbe:


src/test/java/com/google/api/codegen/testdata/python_test_library.baseline
```diff
@@ -51,6 +57,23 @@ class TestLibraryServiceClient(unittest2.TestCase):
...
+
+        client = library_service_client.LibraryServiceClient()
+
+        # Mock request
+        shelf = library_pb2.Shelf()
+
+        # Mock exception response
+        grpc_stub.CreateShelf.side_effect = CustomException()
+
+        self.assertRaises(errors.GaxError, client.create_shelf, shelf)
```
shinfan: Do you need to test for the error code and message here? We need to ensure the original exception info is correctly wrapped.

lukesneeringer: That is not particularly easy to do with `unittest.assertRaises`. The correct way to do this stuff in Python is to have _subclasses_ of `GaxError` for each code (such as [what I almost did here][1]), which would then be what we would test for. But we do not right now, so what is here should be fine.

  [1]: https://github.com/googleapis/gax-python/pull/142#discussion_r94071999

geigerj: It's not as clean as using `assertRaises`, but it's possible to inspect the error with something like

```
try:
  ...
except GaxError as e:
  self.assert(<something about e>)
else:
  fail('shouldn't be here')
```

lukesneeringer: I would rather use `assertRaises`, and come back to this when we rethink errors. There is a pretty big push to try to defenestrate `GaxError` completely, so I do not see a huge purpose in trying to make the perfect the enemy of the good here.

shinfan: Hmmm can you use try/except block to catch the exception object? Here is what we did in Java:
https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-trace/src/test/java/com/google/cloud/trace/spi/v1/TraceServiceClientTest.java#L109

shinfan: Oops @geigerj already replied with the same content...

geigerj: SGTM, but want to make sure this isn't lost is the shuffle. @eoogbe, could you create an issue to improve the testing coverage here once the dust has settled around `GaxError`?

shinfan: @lukesneeringer Fair enough, we can revisit this part when we settle the GaxError design.

shinfan:
LGTM

src/test/java/com/google/api/codegen/testdata/python_test_library.baseline
eoogbe: #1143

PR #1142, opened 2017-03-28
Move construction time of Ruby page streaming data

saicheems:
Ensure that renames due to symbol table collisions affect the required
parameters first, and page streaming parameters last.


src/main/java/com/google/api/codegen/discovery/transformer/js/JSSampleMethodToViewTransformer.java
```diff
@@ -98,6 +94,10 @@ private SampleView createSampleView(SampleTransformerContext context) {
       }
     }
 
+    if (methodInfo.isPageStreaming()) {
```
tcoffee-google: Can we add a comment before this block in all 8 languages to note this issue? The order dependence is not immediately obvious.

vchudnov-g: +1

saicheems: Done

PR #1144, opened 2017-03-29
Add paging to Python unit test

eoogbe:


src/test/java/com/google/api/codegen/testdata/python_test_library.baseline
```diff
@@ -117,6 +124,42 @@ class TestLibraryServiceClient(unittest2.TestCase):
...
+
+        paged_list_response = client.list_shelves()
+        resources = list(paged_list_response)
+        self.assertEqual(1, len(resources))
+        self.assertEqual(expected_response.shelves[0], resources[0])
+
+        grpc_stub.ListShelves.assert_called_once()
+
+    @mock.patch('google.gax.config.API_ERRORS', (CustomException,))
+    @mock.patch('google.gax.config.create_stub')
```
lukesneeringer: We should use `spec=` when creating `MagicMock` objects if possible.

This does *not* block merge so long as an issue is made.

Issue #1033, opened 2017-02-14
Update proto comment regexes to match unordered lists


geigerj:
### What

The comment fixers (e.g., https://github.com/googleapis/toolkit/blob/master/src/main/java/com/google/api/codegen/py/PythonSphinxCommentFixer.java) don't handle Markdown unordered lists correctly. They seem to be mis-recognized as code blocks.

### Example

proto: https://github.com/googleapis/googleapis/blob/af42d585cfd43537fc21f3e950b1db2210e0cd9e/google/logging/v2/logging_config.proto#L245

GAPIC:
https://github.com/lukesneeringer/api-client-staging/blob/c6e2da47cc3ed03305ac84506eefbeea6d614394/generated/python/gapic-google-cloud-logging-v2/google/cloud/gapic/logging/v2/config_service_v2_client.py#L411

eoogbe:
This appears to be resolved

PR #1146, opened 2017-03-29
Match unordered list in Ruby proto comments

eoogbe:
Ruby-only fix for #1033. Python and NodeJS still left to do.

### DIFF
--- a/src/main/java/com/google/api/codegen/util/ruby/RubyCommentReformatter.java
+++ b/src/main/java/com/google/api/codegen/util/ruby/RubyCommentReformatter.java
@@ -21,29 +21,35 @@ import com.google.common.base.Strings;
 import java.util.regex.Matcher;
 
 public class RubyCommentReformatter implements CommentReformatter {
+  private static final String BULLET = "* ";
+
   @Override
   public String reformat(String comment) {
     StringBuffer sb = new StringBuffer();
-    boolean inList = false;
+    int listIndent = 0;
+    boolean followsListItem = false;
+    boolean followsBlankLine = false;
     for (String line : Splitter.on("\n").split(comment)) {
       Matcher listMatcher = CommentPatterns.UNORDERED_LIST_PATTERN.matcher(line);
       boolean matchesList = listMatcher.lookingAt();
-      if (line.isEmpty()) {
-        inList = false;
-      } else if (matchesList) {
-        inList = true;
+      Matcher indentMatcher = CommentPatterns.INDENT_PATTERN.matcher(line);
+      indentMatcher.lookingAt();
+      int indent = indentMatcher.group().length();
+      if (matchesList) {
+        line = listMatcher.replaceFirst(BULLET);
       }
-      if (inList) {
-        Matcher indentMatcher = CommentPatterns.INDENT_PATTERN.matcher(line);
-        if (matchesList) {
-          line = listMatcher.replaceFirst("* ");
-        } else if (indentMatcher.lookingAt()) {
-          int leadingWhitespaceCount = indentMatcher.group().length();
-          sb.append(Strings.repeat(" ", leadingWhitespaceCount - 2));
-          line = line.trim();
-        }
+      if (indent < listIndent && (matchesList || followsBlankLine)) {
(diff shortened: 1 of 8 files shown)


src/main/java/com/google/api/codegen/util/ruby/RubyCommentReformatter.java
```diff
@@ -17,17 +17,44 @@
 import com.google.api.codegen.CommentPatterns;
 import com.google.api.codegen.util.CommentReformatter;
 import com.google.common.base.Splitter;
+import com.google.common.base.Strings;
 import java.util.regex.Matcher;
 
 public class RubyCommentReformatter implements CommentReformatter {
   @Override
   public String reformat(String comment) {
```
geigerj: Possible to separate the logic out into a private method, like the other `reformatXXX`?

src/main/java/com/google/api/codegen/util/ruby/RubyCommentReformatter.java
```diff
@@ -17,17 +17,44 @@
...
+      if (line.isEmpty()) {
+        inList = false;
+      } else if (matchesList) {
+        inList = true;
+      }
+      if (inList) {
+        Matcher indentMatcher = CommentPatterns.INDENT_PATTERN.matcher(line);
+        if (matchesList) {
+          line = listMatcher.replaceFirst("* ");
+        } else if (indentMatcher.lookingAt()) {
```
geigerj: Not sure this is right -- [list don't need to be indented](https://daringfireball.net/projects/markdown/syntax#list)

src/main/java/com/google/api/codegen/util/ruby/RubyCommentReformatter.java
eoogbe: It handles the whole reformating, including all the other `reformatXXX` methods in `applyTransformations`.

src/main/java/com/google/api/codegen/util/ruby/RubyCommentReformatter.java
geigerj: Ah got it. Could you comment the code here? Also, it might be good to make the test case cover more of the possible list syntax, like:

```
// +  Here is a sentence about the first element of the list that continues
//     into a second line.
// +  The second element of the list.
//
// +  Another element of the list after a blank space
// where the indentation isn't consistent
//
//    still part of the
// last list bullet.
```

src/main/java/com/google/api/codegen/util/ruby/RubyCommentReformatter.java
geigerj: I can see that, but for readability, the list reformatting logic would ideally be in a discrete code unit, unless there's something that prevents that.

eoogbe: The list formatting logic depends on many variables being manipulated on the whole comment rather than just a line. The other transformations are discrete to a single line. The list formatting logic can't easily be separated into its own method.

src/main/java/com/google/api/codegen/util/ruby/RubyCommentReformatter.java
eoogbe: I don't think the list can continue after a line break unless the whole thing is indented. I will modify your example to exclude the last part from the list.

src/main/java/com/google/api/codegen/util/ruby/RubyCommentReformatter.java
geigerj: Suggest to at least factor out the logic of the `if (inlist)` clause into its own method.

src/main/java/com/google/api/codegen/util/ruby/RubyCommentReformatter.java
geigerj: I think it can -- see the "This is a list item with two paragraphs" example in the page linked above.

eoogbe:
@swcloud LGTY?

PR #1148, opened 2017-03-30
Migrate Python method docs to MVVM

eoogbe:


src/main/java/com/google/api/codegen/transformer/py/PythonSurfaceNamer.java
```diff
@@ -130,6 +137,55 @@ public String getGrpcClientTypeName(Interface service) {
...
+    return lines.build();
+  }
+
+  private boolean hasParams(MethodConfig methodConfig) {
+    if (!Iterables.isEmpty(methodConfig.getRequiredFieldConfigs())) {
+      return true;
+    }
+
+    int optionalParamCount = Iterables.size(methodConfig.getOptionalFieldConfigs());
+    return optionalParamCount > (methodConfig.getPageStreaming() == null ? 0 : 1);
```
geigerj: nit: Hard to discern intent from the code here. Leave a comment that this is discounting the `page_size` parameter? (At least I assume that's what this is doing?)

src/main/resources/com/google/api/codegen/py/main.snip
```diff
@@ -458,9 +448,7 @@
                 def {@methodName(method)}(
                         self,
                         options=None):
-                    @if documentation
-                        {@documentation}
-                    @end
+                    {@methodComments(methodView, methodView.doc)}
```
geigerj: I'm a little surprised this works without the `if` guard; typically leaving out the guard causes an empty return value to render as a newline. Do we have a test case where the method comments are empty to confirm?

eoogbe: Method comments can't be empty.

PR #1151, opened 2017-03-31
Pretty print responses in C# discovery samplegen

saicheems:
It turns out that on success, the C# samples are just printing the fully
qualified name of the response object.

For example, for `language.detections.list` in `translate/v2`:

    Console.WriteLine(response);

prints

    Google.Apis.Translate.v2.Data.DetectionsListResponse

This change converts `response` to a formatted JSON string before printing.

### DIFF
--- a/src/main/resources/com/google/api/codegen/csharp/sample.snip
+++ b/src/main/resources/com/google/api/codegen/csharp/sample.snip
@@ -124,18 +124,18 @@
                         foreach ({@class.pageStreaming.resourceElementTypeName} {@class.pageStreaming.resourceVarName} in response.{@class.pageStreaming.resourceFieldName})
                         {
                             // TODO: Change code below to process each `{@class.pageStreaming.resourceVarName}` element:
-                            Console.WriteLine("[" + {@class.pageStreaming.resourceVarName}.Key + "] = " + {@class.pageStreaming.resourceVarName}.Value);
+                            Console.WriteLine(JsonConvert.SerializeObject({@class.pageStreaming.resourceVarName}));
                         }
                     @else
                         foreach ({@class.pageStreaming.resourceElementTypeName} {@class.pageStreaming.resourceVarName} in response.{@class.pageStreaming.resourceFieldName})
                         {
                             // TODO: Change code below to process each `{@class.pageStreaming.resourceVarName}` resource:
-                            Console.WriteLine({@class.pageStreaming.resourceVarName});
+                            Console.WriteLine(JsonConvert.SerializeObject({@class.pageStreaming.resourceVarName}));
                         }
                     @end
                 @else
                     // TODO: Change code below to process each `response.{@class.pageStreaming.resourceFieldName}` resource:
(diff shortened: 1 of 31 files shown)


saicheems:
@chrisdunelm 

Curious what you think about this one, is this the right approach?

src/test/java/com/google/api/codegen/testdata/discoveries/csharp/csharp_appengine.v1beta5.json.baseline
```diff
@@ -17,6 +17,7 @@
 using Google.Apis.Appengine.v1beta5;
 using Google.Apis.Auth.OAuth2;
 using Google.Apis.Services;
+using Newtonsoft.Json;
```
saicheems: @chrisdunelm 

In the header above, in step 3 we tell users how to add the `Google.Apis` dependency (honestly, looking at it now, I think we could be a bit more clear...).
Do I need to mention `Newtonsoft.Json` as well? I noticed when I ran this sample that I didn't need to add it to `project.json`.

chrisdunelm: Newtonsoft.Json is included transitively, so it's not necessary to explicitly add it as a dep to the project.
I.e. Google.Apis.* depends on Google.Apis, which depends on Google.Apis.Core, which depends on Newtonsoft.Json

src/main/resources/com/google/api/codegen/csharp/sample.snip
```diff
@@ -124,18 +124,18 @@
                         foreach ({@class.pageStreaming.resourceElementTypeName} {@class.pageStreaming.resourceVarName} in response.{@class.pageStreaming.resourceFieldName})
                         {
                             // TODO: Change code below to process each `{@class.pageStreaming.resourceVarName}` element:
-                            Console.WriteLine("[" + {@class.pageStreaming.resourceVarName}.Key + "] = " + {@class.pageStreaming.resourceVarName}.Value);
+                            Console.WriteLine(JsonConvert.SerializeObject({@class.pageStreaming.resourceVarName}));
```
chrisdunelm: It may be better to keep the distinction between Key and Value, to make it clear these are always expected to be present on the result elements.

I would guess that the key would normally be a primitive type (e.g. string) that should print correctly, so maybe:
`Console.WriteLine("[" + {@class.pageStreaming.resourceVarName}.Key + "] = " + JsonConvert.SerializeObject({@class.pageStreaming.resourceVarName}.Value));`

saicheems: Done

tcoffee-google:
Recalling offline conversation, we had originally tried to do this with .NET framework, but it was unreliable:
https://github.com/googleapis/toolkit/commit/93a73df1e2257d72d7872bcd0f0eb958f471cb12

If Newtonsoft.Json is part of Google.Apis, this is definitely a better solution.

PR #1152, opened 2017-03-31
Use process.exit(1) instead of return on failure

saicheems:
All other samples (except JS) return a non-zero status code on failure.


saicheems:
Let me know what you think about this one...

The reasoning is this new testing strategy,
Every language fails with an exception on error, so the sample process returns with a non-zero status. That makes it really easy to verify whether or not a sample executed successfully against the mock server.

I wanted to align Node.js to that behavior, and it doesn't seem like too drastic or weird behavior to use `process.exit(1)`

saicheems:
Actually... after looking at this again, it looks pretty ugly. I'll have to find out another way to figure out if an error occurred.

PR #1154, opened 2017-03-31
Print Node.js errors to stderr

saicheems:


tcoffee-google:
I'm assuming that in a typical Node.js dev environment, these errors will still appear in the console.

saicheems:
In the CLI environment? Yes, they're sent to `stdout`

PR #1155, opened 2017-03-31
Print JS errors to stderr

saicheems:


saicheems:
Note that 404 errors aren't printed properly, because the `reason` doesn't have a `result` property in that scenario

Maybe we should print just `reason` instead?

tcoffee-google:
If we're going to output the whole object, should we use the native `JSON.stringify`?

saicheems:
I'll come back to the 404 thing in a follow-up

PR #1157, opened 2017-03-31
Treating Beta APIs as experimental in Java

garrettjonesgoogle:
Since we are going GA, this needs to be done...


michaelbausor:
Won't the beta APIs be in their own, non-GA packages?

garrettjonesgoogle:
So this is the release level setting for GAPIC clients, which can be at a different release level than the corresponding handwritten layer. For logging in particular, the GAPIC client will be beta and the handwritten layer will be GA.

michaelbausor:
Got it. LGTM

PR #1158, opened 2017-03-31
Add no_auth flag

saicheems:
Add `NONE` as an `AuthType`.


src/main/java/com/google/api/codegen/discovery/ViewModelProvider.java
```diff
@@ -26,6 +26,7 @@
 import com.google.api.codegen.rendering.CommonSnippetSetRunner;
```
saicheems: Note that all the changes in this file are purely cosmetic, I thought it was way too verbose

src/main/java/com/google/api/codegen/discovery/ViewModelProvider.java
```diff
@@ -72,15 +66,28 @@ private ViewModelProvider(
...
-    sampleConfig = override(sampleConfig, sampleConfigOverrides);
-    ViewModel surfaceDoc = methodToViewTransformer.transform(method, sampleConfig);
-    Doc doc = snippetSetRunner.generate(surfaceDoc);
+        new ApiaryConfigToSampleConfigConverter(methods(), apiaryConfig(), typeNameGenerator())
+            .convert();
+    ObjectMapper mapper = new ObjectMapper().registerModule(new GuavaModule());
+    JsonNode tree = mapper.valueToTree(sampleConfig);
+    tree = override(tree, sampleConfigOverrides());
+
+    if (sampleOptions().noAuth()) {
```
tcoffee-google: Seems like it might be more elegant---and readable---here to keep the surrounding plumbing in the `override` function and use it again with the `noAuth` flag formatted as an additional override.

tcoffee-google:
Also, do we need to add snippet support for the `NONE` option?

saicheems:
The motivation for this flag is to avoid calling auth code in the mock tests - but it could also be used to generate samples without auth boilerplate.

I will add snippet support language by language, I didn't want to make this PR too big.

src/main/java/com/google/api/codegen/discovery/ViewModelProvider.java
saicheems: Done, is this what you meant?

tcoffee-google:
Is this still in flight?

saicheems:
Doing it a little differently, will put the PR up shortly.

PR #1163, opened 2017-04-04
Support custom discovery URLs and no auth in samples

saicheems:
This commit adds support for:

1. The `NONE` auth type in all languages, which if specified via override will
   cause samples to be generated with no auth boilerplate.
2. Custom Discovery doc URLs in JS and Python via the `discoveryDocUrl`
   override field.

The handcrafted and trivial Discovery doc `foo.v1.json` and its corresponding
overrides file `foo.v1.json.overrides` are included so baselines in each
language are available to demonstrate the two changes above.


src/main/resources/com/google/api/codegen/js/sample.snip
```diff
@@ -115,30 +115,44 @@
       @end
       function {@class.initClientFuncName}() {
         gapi.client.init({
-          'apiKey': '',  // TODO: Update placeholder with desired API key.
-          @if class.auth.type != "API_KEY"
+          @if class.auth.type != "NONE"
+            'apiKey': '',  // TODO: Update placeholder with desired API key.
```
tcoffee-google: Is this right?

saicheems: In the JS generator review Bochun stated that the API key should be included for all auth types, so I believe so

src/main/java/com/google/api/codegen/discovery/config/js/JSTypeNameGenerator.java
```diff
@@ -35,4 +35,10 @@ public String stringDelimiter() {
   public String getStringFormatExample(String format) {
     return getStringFormatExample(format, "Date.toISOString()", "Date.toISOString()");
   }
+
+  @Override
+  public String getDiscoveryDocUrl(String apiName, String apiVersion) {
+    return String.format(
+        "https://content.googleapis.com/discovery/v1/apis/%s/%s/rest", apiName, apiVersion);
```
saicheems: The other client libraries are not dynamic, so don't include the discovery doc URL implicitly. Python constructs the URL on the fly from the API name and version, so we don't include it there for any Google API. For JavaScript though the recommended method of constructing the service is to actually pass in the discovery URL, so we have to include it.

It was already there in the snippet, but I moved it to the `SampleConfig`/`JSTypeNameGenerator` so we can target a custom discovery doc in the dynamic clients.

So no other language uses this by default, which is why only JS has it, and Python only considers it if it has been explicitly overridden. That is, by default for any discovery doc the string will be empty, and the python build statement will look like:

```discovery.build('foo', 'v1')```

and if it's specified, it'll look like:

```discovery.build('foo', 'v1', discoveryServiceUrl='{discoveryDocUrl}'```

None of the other languages use this property.

saicheems:
OK, I've cleaned up the snippets a bit, it's a bit tricky with these templates. I've only done it for the statements where there's no code duplication or extra lines.

I wish there was snippet support for more complex statements :/ at least a continue in a switch case or something...

Issue #1162, opened 2017-04-04
Ensure that gapic/ header version is incremented


bjwatson:
### What

The `gapic/` header version is currently `0.1.0` for NodeJS and PHP. We need to ensure that this increments with each new release.

### Why

If the `gapic/` version does not change, then its useless data. It would be better to report no version (i.e. `'gapic/'`) rather than a stale version.

lukesneeringer:
This is not broken in Node; however, many GAPICs had not been regenerated since it was fixed.

PR #1164, opened 2017-04-04
Add a GAPIC_VERSION variable in Node.js GAPICs.

lukesneeringer:
This is a prerequisite to having headers report them properly, since their `package.json` files are stripped away before delivery.


lukesneeringer:
Tagging #1162.

lukesneeringer:
It turns out that this is not necessary for #1162.
However, I would like to merge it anyway. Having an index level version is expected and will have other uses.

landrito:
What other uses do you have in mind?

lukesneeringer:
Mostly I just think it is good behavior to have the version listed on the outer-level object.

PR #1166, opened 2017-04-05
Update properties file path in Java XSettings

shinfan:
More context: https://github.com/GoogleCloudPlatform/google-cloud-java/pull/1863

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
+++ b/src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
@@ -43,7 +43,6 @@ import com.google.api.tools.framework.model.Method;
 import com.google.api.tools.framework.model.ProtoElement;
 import com.google.api.tools.framework.model.ProtoFile;
 import com.google.api.tools.framework.model.TypeRef;
-import com.google.common.base.Splitter;
 import com.google.common.collect.ImmutableList;
 import io.grpc.Status;
 import java.util.ArrayList;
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -113,6 +114,13 @@ public String getServicePhraseName(Interface service) {
     return Name.upperCamel(service.getSimpleName()).toPhrase();
   }
 
+  /** Artifact ID of the service */
+  public String getArtifactId(Interface service) {
+    List<String> tokens =
+        Splitter.on('.').splitToList(service.getModel().getServiceConfig().getName());
+    return tokens.size() > 0 ? "google-cloud-" + tokens.get(0) : null;
```
garrettjonesgoogle: You are hardcoding the knowledge of google-cloud. This doesn't work for non-google-cloud services. We need to get this from config somehow.

garrettjonesgoogle: Or it's probably better if it is provided at runtime from the library calling the GAPIC client.

shinfan: Done

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceNamer.java
```diff
@@ -198,4 +199,14 @@ public String getReleaseAnnotation(ReleaseLevel releaseLevel) {
   public String getBatchingDescriptorConstName(Method method) {
     return inittedConstantName(Name.upperCamel(method.getSimpleName()).join("batching_desc"));
   }
+
+  @Override
+  public String getPackagePath() {
+    List<String> packagePath = Splitter.on(".").splitToList(getPackageName());
+    if (packagePath.size() > 3) {
+      // Remove the "spi.{version}" suffix
+      return Joiner.on("/").join(packagePath.subList(0, packagePath.size() - 2));
```
garrettjonesgoogle: We should check to make sure our assumption is correct (that the package actually has the spi.version suffix), and if not, either throw an exception or do something different.

shinfan: Switched to use indexOf("spi"). Consider done.

shinfan:
PTAL

garrettjonesgoogle:
LGTM

PR #1168, opened 2017-04-05
Use placeholders for strings that are URL vars

saicheems:
Parameters which are URL variables have the attribute `location` set to
`path` in the Discovery file. This commit adds support for that field,
and adds a non-empty placeholder format for URL variable strings.

The reasoning is that samples executed with URL variables that are empty
strings will return a 404 error, because client libraries combine
contiguous slashes.

For example, the following sample for the method `/foo/{bar}/baz`:

    bar = ""
    service.get_baz(bar)

will send a request to `/foo/baz` instead.

This commit will generate the following sample instead:

    bar = "my-bar"
    service.get_baz(bar)

which correctly sends the request to `/bar/my-bar/baz`.

Note that URL variables are always strings, but parameters which have a
non-string `format` specified are not affected by this change.


src/main/java/com/google/api/codegen/discovery/DefaultString.java
```diff
@@ -94,16 +97,15 @@ public static String getNonTrivialPlaceholder(String pattern) {
...
-  public static String getPlaceholder(String fieldName, String pattern) {
-    if (pattern != null) {
-      // If the pattern has a specially-recognized default, use the default. No sample.
-      String def = forPattern(pattern, "{MY-%s}", true);
-      if (def != null) {
-        return def;
-      }
-    }
+  /** Returns true if field is a string and a path parameter. */
+  public static boolean shouldReplace(FieldInfo field) {
```
tcoffee-google: This name seems a bit confusing; how about just `isPathString`?

src/main/java/com/google/api/codegen/DiscoveryImporter.java
```diff
@@ -453,11 +469,19 @@ private String responseTypeUrl(JsonNode root) {
...
+      String kindName,
+      JsonNode formatNode,
+      JsonNode patternNode,
+      JsonNode locationNode) {
     String format = "";
     if (formatNode != null) {
       format = formatNode.asText();
     }
+    if (locationNode != null) {
+      config.getFieldLocation().put(type, field, locationNode.asText());
```
tcoffee-google: It might be more robust to establish an enum for the two options we know about, just in case there are options we don't know about.

saicheems:
Huh, I totally forgot we had changed those. What if I put them back? Instead of printing them into the field, we decided to add an `ex: "foo/my-bar"` comment below ([example](https://github.com/googleapis/toolkit/blob/4fc0207a03a155aadbff1cff6502ee228b212b37/src/test/java/com/google/api/codegen/testdata/discoveries/go/go_cloudbilling.v1.json.baseline#L43)).

Should we change that back?

I'll see if there's a nice way to drop the repetition. I'd like to come back to it later though, in the end I'd like to do a survey of the different transformers and drop all of the common functionality into a series of abstract class.

tcoffee-google:
Right, IIRC we had taken them out of the actual placeholders due to concern about confusing errors if they were not changed; but you observed that in the case of path parameters, users get confusing errors if the placeholders are empty, too, so it's probably just as well to put them back.

vchudnov-g:
Would capitalizing MY-FOO make it stand out more? I want to make it really clear to users and ourselves what is a placeholder.

tcoffee-google:
Like the braces, I always felt that the all-caps encouraged (perhaps even seemed to instruct) the user to use a similar format for actual values. I feel that the placeholder TODOs are sufficient to call attention.

src/main/java/com/google/api/codegen/discovery/DefaultString.java
saicheems: Removed the function

src/main/java/com/google/api/codegen/DiscoveryImporter.java
saicheems: Done

saicheems:
Used lowercase dash format for now, feel free to suggest alternatives. Personally, I think it's fine to leave as-is. In most cases where there is a path-template style string the documentation makes the intention clear.

src/main/java/com/google/api/codegen/discovery/config/TypeNameGenerator.java
```diff
@@ -146,18 +147,27 @@ public String getStringFormatExample(String format, String dateRef, String dateT
...
    * <p>If not the empty string, the returned value will be enclosed within the correct
    * language-specific quotes.
    */
-  public String getFieldPatternExample(String pattern) {
-    String def = DefaultString.getNonTrivialPlaceholder(pattern);
-    if (Strings.isNullOrEmpty(def)) {
+  public String getStringFieldPlaceholder(String fieldName, String pattern, boolean inPath) {
+    String format = "my-%s";
+    String val = DefaultString.getNonTrivialPlaceholder(pattern, format);
+    if (!Strings.isNullOrEmpty(val)) {
```
tcoffee-google: Are there string fields with recognized patterns that are not path parameters?

src/main/java/com/google/api/codegen/discovery/transformer/csharp/CSharpSampleMethodToViewTransformer.java
```diff
@@ -224,8 +224,8 @@ private SampleFieldView createSampleFieldView(
       defaultValue = String.format(typedValue.getValuePattern(), typeName);
     } else {
       defaultValue = typeTable.getZeroValueAndSaveNicknameFor(typeInfo);
```
tcoffee-google: This segment seems more appropriately an if-else. Also seems like the initial values for `defaultValue` and `typeName` above are unnecessary. Similarly in other `...SampleMethodToViewTransformer`s.

src/main/java/com/google/api/codegen/discovery/config/TypeNameGenerator.java
saicheems: It's theoretically possible according to the discovery doc format. This function should allow for it, it checks first to see if the pattern is valid. Otherwise it only returns a placeholder for path parameters.

tcoffee-google: Sure --- I suppose this is the behavior we would want (even query parameters get a nontrivial placeholder if they match a pattern), just curious if this was a consequential decision.

src/main/java/com/google/api/codegen/discovery/transformer/csharp/CSharpSampleMethodToViewTransformer.java
saicheems: Done

src/main/java/com/google/api/codegen/discovery/config/TypeNameGenerator.java
saicheems: Yeah I hadn't thought of that... I guess we may as well, so long as we have the ability.

PR #1170, opened 2017-04-05
Migrate Python path templates to MVVM

eoogbe:


src/main/java/com/google/api/codegen/py/PythonGapicContext.java
```diff
@@ -143,6 +149,21 @@ public ApiMethodView getApiMethodView(Interface service, Method method) {
     return apiMethodTransformer.generateMethod(methodContext);
   }
 
+  public List<PathTemplateView> getPathTemplates(Interface service) {
+    SurfaceTransformerContext context = getSurfaceTransformerContextFromService(service);
```
geigerj: Optional: worth saving as an instance variable Map<Interface, SurfaceTransformerContext> to avoid having to instantiate so many contexts?

eoogbe: This is intended to be very temp. Right now, I'm planning  the following PRs for full MVVM:
1. gRPC stubs
2. api methods
3. gRPC doc message
4. enums
5. finish

geigerj: Got it -- that sounds good to me.

eoogbe:
@lukesneeringer LGTY?

PR #1171, opened 2017-04-05
Use trailing commas

saicheems:
This commit adds a trailing comma to the last item of each object. Due
to our limitations in snippet language, and the fact that this code is
autogenerated, it's easier to use this style than try to ensure that the
last item in an object has no comma following it.

This commit also achieves consistency, as the samples are currently
inconsistent because all optional request body fields are generated with
trailing commas.

### DIFF
--- a/src/main/resources/com/google/api/codegen/nodejs/sample.snip
+++ b/src/main/resources/com/google/api/codegen/nodejs/sample.snip
@@ -83,7 +83,7 @@
         // TODO: Add desired media content for upload. See
         // https://github.com/google/google-api-nodejs-client#media-uploads
         mimeType: '',  // See https://www.w3.org/Protocols/rfc1341/4_Content-Type.html
-        body: {}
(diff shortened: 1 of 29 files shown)


tcoffee-google:
IIRC we had not done this before because trailing commas are not supported in ES3.

saicheems:
This is true for browser JS, and I think Bochun suggested at the end of the day that we go ahead and use trailing commas.

For Node.js though, the only runtime is built on v8. I don't think it's possible for a reasonably recent distribution (last 6-7 years) to raise a syntax error.

saicheems:
This change is specifically for JS, but it should work for any reasonably modern JS engine. JSON itself doesn't allow trailing commas, but all JS objects do since ES5.

This also brings us in compliance with the Google style guide for JS:
https://google.github.io/styleguide/jsguide.html#features-objects-use-trailing-comma

PR #1177, opened 2017-04-10
go: accept CallOption

pongad:
Earlier PRs have added plumbing for options,
but did not provide a way for users to pass the options.

This commit fixes the problem.

Note that the append uses the three-argument slicing syntax,
to avoid overwriting parts of the original slice.

### DIFF
--- a/src/test/java/com/google/api/codegen/testdata/go_main_library.baseline
+++ b/src/test/java/com/google/api/codegen/testdata/go_main_library.baseline
@@ -241,14 +241,15 @@ func (c *Client) BookIAM(book *librarypb.Book) *iam.Handle {
 
 // CreateShelf creates a shelf, and returns the new Shelf.
 // RPC method comment may include special characters: <>&"`'@.
-func (c *Client) CreateShelf(ctx context.Context, req *librarypb.CreateShelfRequest) (*librarypb.Shelf, error) {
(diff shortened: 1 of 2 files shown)


src/test/java/com/google/api/codegen/testdata/go_main_library.baseline
```diff
@@ -670,14 +694,15 @@ func (c *Client) GetBigBook(ctx context.Context, req *librarypb.GetBookRequest)
...
+func (c *Client) GetBigNothing(ctx context.Context, req *librarypb.GetBookRequest, opts ...gax.CallOption) (*GetBigNothingOperation, error) {
     ctx = insertXGoog(ctx, c.xGoogHeader)
+    opts = append(c.CallOptions.GetBigNothing[0:len(c.CallOptions.GetBigNothing):len(c.CallOptions.GetBigNothing)], opts...)
     var resp *longrunningpb.Operation
     err := gax.Invoke(ctx, func(ctx context.Context, settings gax.CallSettings) error {
         var err error
         resp, err = c.client.GetBigNothing(ctx, req, settings.GRPC...)
         return err
-    }, c.CallOptions.GetBigNothing...)
+    }, opts...)
```
pongad: We resolve the call option for getting the first operation. We don't have call options for calls to Poll/Wait though, since the longrunning package itself doesn't use gapic-generated code. Are you OK with this?

jba: Can we we add ...CallOption to the longrunning methods?

PR #1180, opened 2017-04-11
Renames for terminology sanity

garrettjonesgoogle:
We have a terminology mess in toolkit, partly due to factors outside our control, but this PR is an attempt to straighten out some of it. Another goal in this PR is to start making some room for other types of clients besides Gapic.

Here are the terminology problems we have:

* "service" is inherently ambiguous: In proto2/3, it is a group of API methods; however, in service config, a Service applies to a full API product, containing multiple instances of "service" definitions from proto.
* "api" is inherently ambiguous: In service config, an Api applies to a "service" definition from proto, but in product names, it applies to the whole product.

The toolkit codebase has mirrored different instances of these usages in different places, resulting in a pretty confusing situation:

* ApiConfig (Gapic generation) applies to full product; Api (used in Discovery) applies to a group of API methods (subset of a product)
* Instances of the type Interface (a group of API methods) are called "service" all over the place in Gapic config; ServiceConfig applies to a product, and Service (used in Discovery) also applies to a product.

In this PR, I'm doing the following:

* Renaming ApiConfig to GapicProductConfig, introducing an interface ProductConfig that is input-agnostic
* Renaming MethodConfig to GapicMethodConfig
* Renaming InterfaceConfig to GapicInterfaceConfig, introducing an interface InterfaceConfig that is input-agnostic
* Renaming ServiceConfig to ProductServiceConfig
* Renaming SurfaceTransformerContext to GapicInterfaceContext, introducing an interface InterfaceContext that is input-agnostic
* Renaming MethodTransformerContext to GapicMethodContext
* Renaming all variables holding Interface to "apiInterface" (former names: "service", "interfaze", "iface")

I am not going to take on extra renames in this PR; if there are additional suggestions, we can consider that for follow-up.


src/main/java/com/google/api/codegen/config/GapicInterfaceConfig.java
```diff
@@ -0,0 +1,405 @@
+/* Copyright 2016 Google Inc
```
michaelbausor: A hidden rename? Or should this be 2017

src/main/java/com/google/api/codegen/grpcmetadatagen/GrpcMetadataGeneratorTool.java
```diff
@@ -94,7 +94,7 @@ public static void main(String[] args) throws Exception {
 
   private static void generate(
       String descriptorSet,
-      String[] apiConfigs,
+      String[] configs,
```
michaelbausor: Same as above, not that descriptive

src/main/java/com/google/api/codegen/transformer/InterfaceContext.java
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen.transformer;
+
+import com.google.api.codegen.config.ProductConfig;
+import com.google.api.codegen.util.TypeTable;
+
+/**
+ * The context for transforming an API interface, in an input-agnostic way, into a view model to use
```
michaelbausor: Should this be described as a Product interface?

src/main/java/com/google/api/codegen/GapicContext.java
```diff
@@ -36,42 +36,49 @@
...
+   */
+  public ProductServiceConfig getServiceConfig() {
+    return productServiceConfig;
   }
 
-  /** Return the name of the class which is the GAPIC wrapper for this service interface. */
-  public String getApiWrapperName(Interface service) {
-    return service.getSimpleName() + API_WRAPPER_SUFFIX;
+  /** Return the name of the class which is the GAPIC wrapper for this API interface. */
+  public String getApiWrapperName(Interface apiInterface) {
```
michaelbausor: Why did you pick variable name `apiInterface`? The term `api` is confusing as you mention, and you have removed it from elsewhere. I agree that the variable name `interface` is also confusing. I think we should consider renaming the `Interface` type to something more descriptive, if that is possible at all.

src/main/java/com/google/api/codegen/CodeGeneratorTool.java
```diff
@@ -101,14 +101,14 @@ public static void main(String[] args) throws Exception {
 
   private static int generate(
       String descriptorSet,
-      String[] apiConfigs,
+      String[] configs,
```
garrettjonesgoogle: Right, but the tool option name itself is "configs". It's easier to just match them I think.

src/main/java/com/google/api/codegen/GapicContext.java
```diff
@@ -36,42 +36,49 @@
...
     return model;
   }
 
-  /** Returns the associated config. */
-  public ApiConfig getApiConfig() {
-    return apiConfig;
+  /*
+   * Returns the associated config.
+   *
+   * NOTE: The name here is out of date, but this whole class will be deprecated soon.
```
garrettjonesgoogle: Well it would be a lot of disruption for little gain - I didn't want to touch templates.

This is the big advantage of the MVVM architecture - changes to the transformers don't affect the templates.

src/main/java/com/google/api/codegen/GapicContext.java
garrettjonesgoogle: The compound term "API interface" has a clear referent, even if the components (API, interface) don't. 

We can't rename `Interface`, it comes from api-compiler.

src/main/java/com/google/api/codegen/config/GapicInterfaceConfig.java
garrettjonesgoogle: Hidden rename.

src/main/java/com/google/api/codegen/transformer/InterfaceContext.java
garrettjonesgoogle: No, "API interface" is correct here.

src/main/java/com/google/api/codegen/config/InterfaceConfig.java
```diff
@@ -14,381 +14,10 @@
...
-
-  public boolean hasLongRunningOperations() {
-    for (MethodConfig methodConfig : getMethodConfigs()) {
-      if (methodConfig.isLongRunningOperation()) {
-        return true;
-      }
-    }
-    return false;
-  }
+public interface InterfaceConfig {
```
geigerj: Where is this used?

garrettjonesgoogle: See changes in `SurfaceNamer`.

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -546,9 +551,7 @@ public String getRequestVariableName(Method method) {
   /////////////////////////////////////// Type names /////////////////////////////////////////////
 
   protected String getInterfaceName(InterfaceConfig interfaceConfig) {
-    return interfaceConfig.hasInterfaceNameOverride()
-        ? interfaceConfig.getInterfaceNameOverride()
-        : interfaceConfig.getInterface().getSimpleName();
+    return interfaceConfig.getName();
```
geigerj: Does this method still need to exist now that it's just an alias for another method?

src/main/java/com/google/api/codegen/config/InterfaceConfig.java
geigerj: Is `SurfaceNamer` not GAPIC-only? I'm wondering why this interface should exist separately of `GapicInterfaceConfig`.

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -102,22 +103,25 @@ public String getNotImplementedString(String feature) {
...
+    return Name.upperCamel(apiInterface.getSimpleName()).toPhrase();
   }
 
   /////////////////////////////////////// Constructors /////////////////////////////////////////////
 
-  /** The name of the constructor for the service client. The client is VKit generated, not GRPC. */
-  public String getApiWrapperClassConstructorName(Interface interfaze) {
-    return publicClassName(Name.upperCamel(interfaze.getSimpleName(), "Client"));
+  /**
+   * The name of the constructor for the apiInterface client. The client is VKit generated, not
```
geigerj: Unrelated to this change, but the second sentence of this comment should be removed.

src/main/java/com/google/api/codegen/config/InterfaceConfig.java
garrettjonesgoogle: `SurfaceNamer` definitely started as GAPIC-based, but over time, I want to convert it to support other inputs. Basically I don't want to repeat all of the naming work for each input. 

Background of this PR: I was starting to bootstrap Discovery-based GAPIC generation, but then encountered the terminology hell. I started fixing the terminology, but some of the generalization work came along for the ride. This interface needs to exist because I have a local `DiscoInterfaceConfig` that is part of the bootstrapping.

src/main/java/com/google/api/codegen/transformer/ModelTypeTable.java
```diff
@@ -173,4 +173,8 @@ public String getImplZeroValueAndSaveNicknameFor(TypeRef type) {
   public Map<String, TypeAlias> getImports() {
     return typeTable.getImports();
   }
+
+  public TypeTable getTypeTable() {
```
geigerj: Note: needing to expose this seems like a sort of strange effect of having `GapicInterfaceContext` implement the new `InterfaceContext`, but if that's what's needed to make it work (or I'm missing some other use for this method) that's fine

garrettjonesgoogle: Yeah, it's needed for it to work. Essentially, `ModelTypeTable` is a decorator around `TypeTable`.

michaelbausor:
LGTM

PR #1182, opened 2017-04-11
go: properly detect LRO methods

pongad:
Methods that return Operation but has no LRO config
should not be treated as LRO by the importer.
This PR makes it possible to generate longrunning gapic client.

### DIFF
--- a/src/test/java/com/google/api/codegen/transformer/go/testdata/myproto.proto
+++ b/src/test/java/com/google/api/codegen/transformer/go/testdata/myproto.proto
@@ -9,7 +9,11 @@ option go_package = "google.golang.org/genproto/googleapis/example/myproto/v1;my
 
 service Gopher {
   rpc SimpleMethod(SimpleRequest) returns (SimpleResponse);
+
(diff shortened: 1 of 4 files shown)


src/test/java/com/google/api/codegen/transformer/go/testdata/myproto.proto
```diff
@@ -9,7 +9,11 @@ option go_package = "google.golang.org/genproto/googleapis/example/myproto/v1;my
 
 service Gopher {
   rpc SimpleMethod(SimpleRequest) returns (SimpleResponse);
+
   rpc LroMethod(SimpleRequest) returns (google.longrunning.Operation);
+  // Returns Operation, but doens't have long_running config in gapic.yaml
```
garrettjonesgoogle: spelling (doens't)

garrettjonesgoogle:
LGTM after fixing the doc

src/test/java/com/google/api/codegen/transformer/go/testdata/myproto.proto
pongad: good catch, done

PR #1183, opened 2017-04-11
Fix multi-segment default strings

saicheems:
The Discovery service is now returning path wildcards as `[^/]+` instead
of `[^/]*`. This commit fixes that case, and adds support for two more
wildcard types: `.+` and `.*`.

### DIFF
--- a/src/main/java/com/google/api/codegen/discovery/DefaultString.java
+++ b/src/main/java/com/google/api/codegen/discovery/DefaultString.java
@@ -75,7 +75,8 @@ public class DefaultString {
     return "";
   }
 
-  private static final Set<String> WILDCARD_PATTERNS = ImmutableSet.of("[^/]+", ".+", ".*");
+  private static final Set<String> WILDCARD_PATTERNS =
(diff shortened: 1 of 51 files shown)


src/main/java/com/google/api/codegen/discovery/DefaultString.java
```diff
@@ -96,7 +75,7 @@ public static String getNonTrivialPlaceholder(String pattern, String format) {
     return "";
   }
 
-  private static final String WILDCARD_PATTERN = "[^/]*";
+  private static final Set<String> WILDCARD_PATTERNS = ImmutableSet.of("[^/]+", ".+", ".*");
```
vchudnov-g: Does it make sense to have `[^/]+` and `[^/]*` as well?

saicheems: I had excluded the latter because it doesn't appear in any discovery docs anymore, but just in case it makes a comeback I've put it in.

PR #1185, opened 2017-04-12
Ruby make enableTestGenerator separate from enableSurfaceGenerator

swcloud:


garrettjonesgoogle:
Why? It looks like they were already independent...

swcloud:
It's inside 

if (generatorConfig.enableSurfaceGenerator()) {
}

should be moved outside

PR #1186, opened 2017-04-13
Support multiple overrides files

saicheems:
Add support for a comma-delimited list of overrides files. These files
are parsed and applied in the order they are listed.

### DIFF
--- a/src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorApi.java
+++ b/src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorApi.java
@@ -66,7 +66,7 @@ public class DiscoveryFragmentGeneratorApi {
       ToolOptions.createOption(
           String.class,
           "overrides",
-          "A comma dielimited list of paths to sample config override files.",
+          "A comma delimited list of paths to sample config override files.",
           "");
 
   public static final Option<List<String>> GENERATOR_CONFIG_FILES =
--- a/src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorApi.java
+++ b/src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorApi.java
@@ -62,9 +62,12 @@ public class DiscoveryFragmentGeneratorApi {
           "The name of the output file or folder to put generated code.",
           "");
 
-  public static final Option<String> OVERRIDES_FILE =
+  public static final Option<String> OVERRIDE_FILES =
       ToolOptions.createOption(
-          String.class, "overrides", "The path to the sample config overrides file.", "");
+          String.class,
+          "overrides",
+          "A comma dielimited list of paths to sample config override files.",
+          "");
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorApi.java
```diff
@@ -62,9 +62,12 @@
           "The name of the output file or folder to put generated code.",
           "");
 
-  public static final Option<String> OVERRIDES_FILE =
+  public static final Option<String> OVERRIDE_FILES =
       ToolOptions.createOption(
-          String.class, "overrides", "The path to the sample config overrides file.", "");
+          String.class,
+          "overrides",
+          "A comma dielimited list of paths to sample config override files.",
```
vchudnov-g: "delimited"

src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorTool.java
```diff
@@ -43,7 +43,7 @@ public static void main(String[] args) throws Exception {
     options.addOption(
         Option.builder()
             .longOpt("overrides")
-            .desc("The path to the sample config overrides file")
+            .desc("A comma dielimited list of paths to sample config override files.")
```
vchudnov-g: "delimited"

src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorApi.java
```diff
@@ -107,15 +110,14 @@ protected void process() throws Exception {
...
+    List<JsonNode> overrides = new ArrayList<>();
+    for (String filename : filenames) {
       try {
         BufferedReader reader =
-            com.google.common.io.Files.newReader(
-                new File(overridesFilename), Charset.forName("UTF8"));
+            com.google.common.io.Files.newReader(new File(filename), Charset.forName("UTF8"));
         ObjectMapper mapper = new ObjectMapper();
-        overridesJson = mapper.readTree(reader);
+        overrides.add(mapper.readTree(reader));
```
vchudnov-g: If the overrides file is in a wrong format, will this also cause an exception? Should we catch it and print the name of the offending file?

saicheems: An exception will be raised when parsing into the `SampleConfig` class. Since it's an `AutoValue`, it'll complain if any types are wrong, or if the expected structure isn't matched.

src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorApi.java
saicheems: done

src/main/java/com/google/api/codegen/DiscoveryFragmentGeneratorTool.java
saicheems: done

PR #1192, opened 2017-04-18
config: Provide a way to override license

pongad:
Previously all languages use the same license for a given service.

This PR provides a way for a language to override the license.
This is necessary for Go longrunning gapic client,
since it will be distributed along with cloud.google.com/go, not gax.


garrettjonesgoogle:
LGTM

PR #1193, opened 2017-04-18
Ruby: Support for LRO tests

landrito:


src/main/resources/com/google/api/codegen/ruby/test.snip
```diff
@@ -122,6 +123,39 @@
...
+    @end
+    @# Create expected grpc response
+    {@initCode(test.mockResponse.initCode)}
+    result = Google::Protobuf::Any.new
+    result.pack(expected_response)
+    operation = Google::Longrunning::Operation.new(
+      name: 'operations/{@test.name}',
+      done: true,
+      response: result
+    )
```
swcloud: Can we add a test case where the result is an 'error'?

landrito: Hm. That sounds feasible let me work on it.

landrito: Done.

src/main/resources/com/google/api/codegen/ruby/test.snip
```diff
@@ -122,6 +123,68 @@
...
+
+    {@test.grpcStubCallString}.stub(:new, mock_stub) do
+      client = {@test.serviceConstructorName}.new
+
+      @# Call method
+      {@methodCallWithResponse(test)}
+
+      @# Verify the response
+      assert(response.error?)
+      assert_equal(error, response.error)
```
swcloud: Where does the 'error' here come from? I think it should be custom_error.

landrito: Ah realized the error must also be of type `Google::Rpc::Status`. Fixed.

PR #1197, opened 2017-04-19
Java: migrating off deprecated methods

garrettjonesgoogle:
Yes, `getIndexMapMap` looks funny, but that's because the name of the field itself ends in "map". It turns out that proto generates both `get${FieldName}` and `get${FieldName}Map`, and the first one is `@Deprecated`. 

This PR coordinates with upcoming PRs in gax-java & google-cloud-java.


michaelbausor:
LGTM

PR #1201, opened 2017-04-20
Support separate grpc and proto generation for Java

michaelbausor:
This PR is a companion to https://github.com/googleapis/artman/pull/161

It splits grpc/proto metadata generation to support two separate packages. I also relies on artman to execute protoc.


shinfan:
If we execute protoc in the artman is will depend on the local version of the protoc installed. I prefer to let Gradle handles the protoc version for us automatically.

geigerj:
@shinfan There will a single preferred version of protoc that artman runs soon, namely the one installed in the docker image. I expect once @lukesneeringer's automation work is complete, we won't have issues with differences between local environments -- Luke, do you know what the timeframe for this is?

michaelbausor:
@shinfan We have that dependency in other language's proto generation - Java is the exception. We are also already somewhat dependent on protoc, because the `ProtoDescGenTask` and `ProtoCodeGenTask` tasks are already being executed for Java.

Having gradle manage the protoc version is nice, but is also complicated by the proto compiler plugin. This means that building the gradle package is dependent on a local installation of the protoc compiler plugin. By pushing that requirement into artman, we keep all our dependencies in one place. (It will also help that once the plugin is published as a pip package, it can be added as an explicit dependency by artman).

shinfan:
Okay, I am okay with handling this within Artman as long as we have a central solution of protoc version (docker image I guess).

src/main/resources/com/google/api/codegen/metadatagen/java/grpc/grpc_package.snip
```diff
@@ -127,7 +72,7 @@
           }
 
           pom.project {
-            name "com.google.api.grpc:{@metadata.identifier}"
+            name "com.google.api.grpc:grpc-{@metadata.identifier}"
```
garrettjonesgoogle: Don't compose the package name in the snippet, pass in the full name through the model (other places too)

src/main/java/com/google/api/codegen/grpcmetadatagen/PackageDivision.java
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen.grpcmetadatagen;
+
+public enum PackageDivision {
```
garrettjonesgoogle: This name doesn't quite feel right.

src/main/resources/com/google/api/codegen/metadatagen/java/grpc/grpc_package.snip
michaelbausor: Done, also restructured to share a snippet file and added a test

src/main/java/com/google/api/codegen/grpcmetadatagen/PackageDivision.java
michaelbausor: As discussed offline, renamed to generation_layer

michaelbausor:
PTAL

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
```diff
@@ -223,6 +230,7 @@ public static PackageMetadataConfig createFromString(String yamlContents) {
         .packageName(buildMapWithDefault((Map<String, String>) configMap.get("package_name")))
         .shortName((String) configMap.get("short_name"))
         .packageType(PackageType.of((String) configMap.get("package_type")))
+        .generationLayer(GenerationLayer.of((String) configMap.get("generation_layer")))
```
shinfan: Do you think it makes sense to combine the generation_layer with package_type? (e.g. we can have something like raw_proto, grpc etc)

garrettjonesgoogle:
LGTM

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
michaelbausor: I think this could make sense, so long as we don't want to generate grpc stubs for core/common proto packages, but I think that we do want to do this so having them be independent settings makes sense. If we don't want to publish the stubs, then I think we can combine them.

shinfan:
LGTM

lukesneeringer:
> I expect once @lukesneeringer's automation work is complete, we won't have issues with differences between local environments -- Luke, do you know what the timeframe for this is?

A couple weeks, I expect.

PR #1206, opened 2017-04-21
Use kwargs for protos

eoogbe:


lukesneeringer:
This looks good. Since you have multiple PRs to be able to successfully run tests, I am not going to block this on being able to run the full suite. I will add additional issues if, after merging all of your PRs, I still see unit test failures.

PR #1214, opened 2017-04-24
Java updates to annotations and package generation

michaelbausor:
Change from ExperimentalApi to BetaApi
Update the java grpc protoc plugin version
Change the dependency package prefix from `grpc` to `proto`


garrettjonesgoogle:
LGTM

PR #1218, opened 2017-04-26
go: implement smoke test

pongad:
Some shared code is modified so that request-object methods
work properly.

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/TestCaseTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/TestCaseTransformer.java
@@ -227,15 +227,20 @@ public class TestCaseTransformer {
 
   public TestCaseView createSmokeTestCaseView(GapicMethodContext context) {
     GapicMethodConfig methodConfig = context.getMethodConfig();
-    ClientMethodType methodType = null;
-    if (methodConfig.isPageStreaming() && context.isFlattenedMethodContext()) {
-      methodType = ClientMethodType.PagedFlattenedMethod;
-    } else if (methodConfig.isPageStreaming() && !context.isFlattenedMethodContext()) {
-      methodType = ClientMethodType.PagedRequestObjectMethod;
-    } else if (!methodConfig.isPageStreaming() && context.isFlattenedMethodContext()) {
-      methodType = ClientMethodType.FlattenedMethod;
-    } else if (!methodConfig.isPageStreaming() && !context.isFlattenedMethodContext()) {
-      methodType = ClientMethodType.RequestObjectMethod;
+    ClientMethodType methodType;
+
+    if (methodConfig.isPageStreaming()) {
+      if (context.isFlattenedMethodContext()) {
+        methodType = ClientMethodType.PagedFlattenedMethod;
+      } else {
+        methodType = ClientMethodType.PagedRequestObjectMethod;
+      }
+    } else {
+      if (context.isFlattenedMethodContext()) {
+        methodType = ClientMethodType.FlattenedMethod;
+      } else {
+        methodType = ClientMethodType.RequestObjectMethod;
+      }
     }
 
     return createTestCaseView(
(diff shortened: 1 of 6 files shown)


src/main/resources/com/google/api/codegen/go/smoke.snip
```diff
@@ -0,0 +1,77 @@
...
+  )
+
+  var _ = iterator.Done
+
+  func {@view.name}(t *testing.T) {
+    if testing.Short() {
+      t.Skip("skipping smoke test in short mode")
+    }
+
+    projectId := testutil.ProjID()
```
shinfan: I have some concerns with the projectId here:
1. projectId is not always needed for every test
2. We should allow the user to pass in the ID from command line. I am not sure what it should look like in Go but in Java we use option args:
https://github.com/googleapis/toolkit/blob/master/src/test/java/com/google/api/codegen/testdata/java_smoke_test_library.baseline#L57

src/test/java/com/google/api/codegen/testdata/go_smoke_library.baseline
```diff
@@ -0,0 +1,64 @@
...
+      Name: formattedName,
+      Book: book,
+  }
+
+  resp, err := c.UpdateBook(ctx, request)
+  if err != nil {
+    t.Error(err)
+  } else {
+    t.Log(resp)
+  }
```
shinfan: Print "OK" if the test succeeded.

src/main/java/com/google/api/codegen/transformer/TestCaseTransformer.java
```diff
@@ -226,9 +226,16 @@ private InitCodeContext createResponseInitCodeContext(
   }
 
   public TestCaseView createSmokeTestCaseView(GapicMethodContext context) {
-    ClientMethodType methodType = ClientMethodType.FlattenedMethod;
-    if (context.getMethodConfig().isPageStreaming()) {
+    GapicMethodConfig methodConfig = context.getMethodConfig();
+    ClientMethodType methodType = null;
+    if (methodConfig.isPageStreaming() && context.isFlattenedMethodContext()) {
```
pongad: Done

src/main/resources/com/google/api/codegen/go/smoke.snip
pongad: For context: The `testutil` package is already used by handwritten clients. We're just reusing it.

> 2. We should allow the user to pass in the ID from command line

It is configurable with `GCLOUD_TESTS_GOLANG_PROJECT_ID` environment variable. So you can run
```shell
GCLOUD_TESTS_GOLANG_PROJECT_ID=my-project go test cloud.google.com/go/...
```
and test all handwritten and gapic clients with the same project.

> 1. projectId is not always needed for every test

It's OK. If you don't set the env var, `testutil.ProjID` returns empty string. If the test doesn't need it, it doesn't matter. It won't throw exception and fail the test.

src/test/java/com/google/api/codegen/testdata/go_smoke_library.baseline
pongad: The `go test` tool already does it when tests succeed. For example:
```
ok  	cloud.google.com/go/speech/apiv1	0.133s
ok  	cloud.google.com/go/speech/apiv1beta1	0.120s
ok  	cloud.google.com/go/trace/apiv1	2.675s
ok  	cloud.google.com/go/vision/apiv1	3.230s
```

pongad:
@shinfan PTAL

src/main/resources/com/google/api/codegen/go/smoke.snip
shinfan: Ok fair enough

pongad:
Sorry for the churn: as generated the smoke test will work on my desktop because it has application default cred. However, it won't work on CI, because creds are passed in another way. https://github.com/googleapis/toolkit/pull/1238 will help with this.

pongad:
@shinfan @jba PTAL

jba:
To avoid a lot of churn, you can remove the t.Log in the future.

pongad:
@jba sounds good, I'll merge this now for expedience. Reported your concern at https://github.com/googleapis/toolkit/issues/1249 so I don't loose track.

PR #1239, opened 2017-05-02
go: move auth scope to a shared location

pongad:
This makes the scope visible to handwritten clients and tests,
particularly testutil.TokenSource.
This will later be used by smoke testing.


garrettjonesgoogle:
Generator code LGTM

PR #1241, opened 2017-05-04
Update Ruby names

saicheems:
This commit also removes the `dataflow:v1b3` discovery file due to a
generation issue in the Ruby client library. Since the client can't
generate it, the names file is wrong, which breaks Discovery samplegen
in turn.


saicheems:
Note that I've asked sbazyl for help on the Ruby client issue, and I'll fix that once I get a reply.

saicheems:
Yeah, now that you mention it.......

It seems two methods have vaguely similar names and signatures, which causes the idiomatic renamer gem to produce a conflicting method name.

The exact error is this:
```
Duplicate method create_job_from_template generated, conflicting paths /dataflow:v1b3/dataflow.projects.locations.templates.create and /dataflow:v1b3/dataflow.projects.templates.create
```

tcoffee-google:
I did not recall that specific issue --- sounds like it should be filed as an issue for the client library generator.

saicheems:
Right, I think it's something I'll probably have to fix myself.
I'll make an issue as well to track.

PR #1243, opened 2017-05-04
Migrate import disambiguation to Python MVVM

eoogbe:
Ports the disambiguation functionality from the `PythonImport` into the `PythonTypeTable` for MVVM. Should also fix disambiguation for test gen.

### DIFF
--- a/src/main/java/com/google/api/codegen/util/py/PythonTypeTable.java
+++ b/src/main/java/com/google/api/codegen/util/py/PythonTypeTable.java
@@ -14,31 +14,43 @@
  */
 package com.google.api.codegen.util.py;
 
-import com.google.api.codegen.util.DynamicLangTypeTable;
 import com.google.api.codegen.util.NamePath;
 import com.google.api.codegen.util.TypeAlias;
 import com.google.api.codegen.util.TypeName;
 import com.google.api.codegen.util.TypeTable;
 import com.google.common.base.Joiner;
 import com.google.common.base.Splitter;
+import com.google.common.collect.BiMap;
 import com.google.common.collect.HashBiMap;
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Multimap;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
 import java.util.Map;
-import java.util.TreeMap;
 
 /** The TypeTable for Python. */
 public class PythonTypeTable implements TypeTable {
 
-  private final DynamicLangTypeTable dynamicTypeTable;
+  /** A bi-map from short names to file names. Should be kept 1:1 with moduleImports keys. */
+  private BiMap<String, String> usedShortNames = HashBiMap.create();
+
+  /**
+   * A multimap from module names to type aliases in the module. Keys should be kept 1:1 with
+   * usedShortNames.
+   */
+  private Multimap<String, TypeAlias> moduleImports = HashMultimap.create();
+
+  private final String implicitPackageName;
 
   public PythonTypeTable(String implicitPackageName) {
-    dynamicTypeTable = new DynamicLangTypeTable(implicitPackageName, ".");
+    this.implicitPackageName = implicitPackageName;
   }
 
   @Override
   public TypeTable cloneEmpty() {
-    return new PythonTypeTable(dynamicTypeTable.getImplicitPackageName());
+    return new PythonTypeTable(implicitPackageName);
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


geigerj:
Can you add/modify a `.baseline` test case that captures the failure that originally prompted this?

eoogbe:
I don't think it needs a baseline if the unit tests sufficiently cover it. I can't migrate the main imports to MVVM without migrating everything to MVVM.

src/main/java/com/google/api/codegen/util/py/PythonTypeTable.java
```diff
@@ -82,19 +94,84 @@ public String getAndSaveNicknameFor(String fullName) {
...
   @Override
   public String getAndSaveNicknameFor(TypeName typeName) {
-    return dynamicTypeTable.getAndSaveNicknameFor(typeName);
+    return typeName.getAndSaveNicknameIn(this);
   }
 
   @Override
   public String getAndSaveNicknameFor(TypeAlias alias) {
-    return dynamicTypeTable.getAndSaveNicknameFor(alias);
+    if (!alias.needsImport()) {
```
geigerj: Can you copy over the comment giving examples, as well? It's very hard to read this code without examples, IMO.

geigerj:
Re: baselines -- the baselines shouldn't care whether or not the main imports are migrated, but I think this is fine just being unit tested, as well.

src/main/java/com/google/api/codegen/util/py/PythonTypeTable.java
eoogbe: Done

src/main/java/com/google/api/codegen/util/py/PythonTypeTable.java
```diff
@@ -82,19 +94,101 @@ public String getAndSaveNicknameFor(String fullName) {
...
+    String shortName = alias.getNickname().substring(0, alias.getNickname().indexOf("."));
+    String className = alias.getNickname().substring(alias.getNickname().indexOf(".") + 1);
+    String moduleName =
+        alias.getFullName().substring(0, alias.getFullName().length() - className.length() - 1);
+    if (usedShortNames.containsKey(shortName)) {
+      // Short name already there.
+      String oldModuleName = usedShortNames.get(shortName);
+      if (moduleName.equals(oldModuleName)) {
+        // New alias for existing module import, no clash
+        moduleImports.put(moduleName, alias);
```
geigerj: IIUC this branch executes if the file in question has already been taken care of. Why do we need to update the moduleImports map, then?

eoogbe: We need to reference all the aliases for a given module name in case we need to replace them with a disambiguated name in `updateOldImports`.

src/main/java/com/google/api/codegen/util/py/PythonTypeTable.java
```diff
@@ -82,19 +94,101 @@ public String getAndSaveNicknameFor(String fullName) {
...
+    // Move the highest-level single package name not already present in the alias into the alias.
+    for (int i = nameParts.size() - 2; i >= 0; --i) {
+      localNamePackagePrefix = nameParts.get(i) + "_" + localNamePackagePrefix;
+      if (!localName.contains(localNamePackagePrefix)) {
+        return nameParts.get(i) + "_" + localName;
+      }
+    }
+
+    // If all packages are present in the alias, mangle.
+    return localName + "_";
```
eoogbe: Realized this case was unreachable, so removed it.

src/test/java/com/google/api/codegen/util/py/PythonTypeTableTest.java
```diff
@@ -0,0 +1,46 @@
...
+import com.google.common.truth.Truth;
+import org.junit.Test;
+
+public class PythonTypeTableTest {
+
+  @Test
+  public void testGetAndSaveNicknameFor_disambiguate_movePackage() {
+    PythonTypeTable typeTable = new PythonTypeTable("foo.bar");
+    typeTable.getAndSaveNicknameFor(TypeAlias.create("a.c.D", "c.D"));
+    Truth.assertThat(typeTable.getAndSaveNicknameFor(TypeAlias.create("b.c.E", "c.E")))
```
geigerj: Does saving this import change the the previous import to `a_c.D`?

eoogbe: It does, but that isn't registered until after the second import is saved. I could check the `getImports` to make sure the changes are there.

geigerj: Could you add that assertion as well?

lukesneeringer:
@eoogbe Can we merge this so I can re-generate monitoring?

lukesneeringer:
Wait, I just remembered that @eoogbe is OOO also. Merging. :-)

PR #1244, opened 2017-05-04
Support C# array default value and fix Go bug

saicheems:
This commit adds support for a default value for C# arrays and fixes a
bug in the NONE auth configuration for Go.

### DIFF
--- a/src/test/java/com/google/api/codegen/testdata/discoveries/foo.v1.json.overrides
+++ b/src/test/java/com/google/api/codegen/testdata/discoveries/foo.v1.json.overrides
@@ -4,5 +4,27 @@
   },
   "js|python": {
     "discoveryDocUrl": "localhost:8080/$discovery/foo?version=v1"
+  },
+  "csharp|go|java": {
+    "methods": {
+      "foo.bar.get": {
+        "fields": {
+          "foobar": {
+            "defaultValue": "\" \""
+          }
+        }
+      }
+    }
+  },
+  "js|nodejs|php|python|ruby": {
+    "methods": {
+      "foo.bar.get": {
+        "fields": {
+          "foobar": {
+            "defaultValue": "' '"
+          }
+        }
+      }
+    }
   }
 }
--- a/src/main/java/com/google/api/codegen/discovery/config/TypeInfo.java
+++ b/src/main/java/com/google/api/codegen/discovery/config/TypeInfo.java
@@ -59,6 +59,8 @@ public abstract class TypeInfo {
     return new AutoValue_TypeInfo.Builder();
   }
 
+  public abstract Builder toBuilder();
+
   @AutoValue.Builder
   public abstract static class Builder {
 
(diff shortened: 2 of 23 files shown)


src/test/java/com/google/api/codegen/testdata/discoveries/foo.v1.json.overrides
```diff
@@ -4,5 +4,27 @@
   },
   "js|python": {
     "discoveryDocUrl": "localhost:8080/$discovery/foo?version=v1"
+  },
+  "csharp|go|java": {
+    "methods": {
+      "foo.bar.get": {
+        "fields": {
+          "foobar": {
+            "defaultValue": "\" \""
```
tcoffee-google: These overrides may get unwieldy if even trivial literals must be specified with language-specific syntax. Could our override parsing do appropriate conversions from JSON types?

saicheems: I've included the override for foo as a demonstration, it's not actually necessary.

I need this field for my overrides generator for the mock tests, which automatically populates these fields with values. The reason is that C# requires a non-empty value in a repeated field.

No production sample should ever use this.

saicheems: I forgot to add the baselines... will add shortly, it should make a bit more sense.

saicheems: Specifically, for array fields, if a default value is specified, that element is added to the array. So if no override is specified, as usual, this is what shows up:

```
List<Foo> foos = new List<Foo>();
```

If an override is applied (in this case, `"defaultValue": "bar"`), this shows up:
```
List<Foo> foos = new List<Foo>();
foos.add("bar");
```

tcoffee-google: I meant that as it is, for instance, we need different overrides for different languages depending on the quote delimiter used for their string literals. Can we do something more intelligent with the override parsing?

saicheems: Technically, yes, but it could be annoying in the future if we do so. For example, we might want to insert `null`, or `x.toString();` or something like that without quotations.

saicheems: I suppose what we could do is, if we see double quotes, treat that as a string and then format that accordingly.

But since this really is intended only for mock overrides and should be generated only by automated tooling, I'd rather hold off on complicating the logic if possible.

tcoffee-google: I would think that JSON null would map to `null`. Not sure what you mean about `x.toString()`. Would we really have overrides that embody code and not just literal values?

saicheems: I hadn't ruled it out, but now that I think about it... that doesn't seem worth supporting.

The change you're suggesting is non-trivial, I'll follow up with it later if that's alright. I just want this for my tests. I'll open an issue to track.

saicheems: #1260

saicheems: Actually I'll just bail on this part of the PR for now, might as well do the whole thing later. There are other issues with default values anyway.

PR #1246, opened 2017-05-05
Update the Discovery parser

saicheems:


saicheems:
An update pulled from my refactoring work - it adds support for some new features, including array items and enums. Also adds auth detection support, and a new field for methods and schemas called `path`, which is the fully qualified path to a resource (ex: `schemas.foo.properties.bar.properties.baz`.

See the tests for details.

garrettjonesgoogle:
LGTM - everything seems reasonable.

PR #1253, opened 2017-05-09
Ruby: Remove aliasing from method samples and tests

landrito:
Often, when getting reviews internally and externally we get the feedback that aliasing in the method samples were weird and unidiomatic. This makes it such that samples and tests simply use fully qualified types. 

Note: Enums are not using fully qualified types. Rather they are just using their ruby symbol. Based on [protobuf#enum](https://developers.google.com/protocol-buffers/docs/reference/ruby-generated#enum) reference for ruby.


landrito:
cc @remi, @omaray

PR #1254, opened 2017-05-09
Fix JSON pretty print in Ruby sample

saicheems:


saicheems:
I think the print statement has been broken for quite a while...

tcoffee-google:
What is the failure?

saicheems:
Don't have the exact stack trace, but something along the lines of `pretty_generate expected 0 arguments but got 1`

tcoffee-google:
Can you try
```ruby
puts JSON.pretty_generate {@class.pageStreaming.resourceVarName}
```
instead of
```ruby
puts JSON.pretty_generate({@class.pageStreaming.resourceVarName})
```
?

saicheems:
I did, they're semantically the same, and result in the same failure

tcoffee-google:
Okay, that's what I thought. 

Seems to contradict documentation here: http://ruby-doc.org/stdlib-2.0.0/libdoc/json/rdoc/JSON.html#method-i-pretty_generate

And examples here:
http://stackoverflow.com/questions/2567670/ruby-json-pretty-generate-is-pretty-unpretty

saicheems:
I believe, BTW, that the failure is because the gem that handled the JSON conversion in the library (`representable`) was upgraded a few months ago. While `response.to_json` works fine, the fact that the superclass of all request/response objects in the Ruby client overrides the `to_json` method is messing with the `JSON.pretty_generate` call.

I suppose I could dig through the client library to get the same behavior as before, but it's more work than I want to do at the moment, and it may also not be a bug...

I also saw the posts you provided. When I get time I'll try to debug the implementation of `to_json` in the Ruby client. For now, this should work for the foreseeable future.

tcoffee-google:
Sure --- working is certainly better than not working, but it is uglier and IIUC produces less pretty output, so let's track in an issue.

saicheems:
I'll file a bug to `google-api-ruby-client`

FWIW, the output is the same as it was before (it's still pretty printed as you'd expect), it's just that the route we take to convert it into a pretty string is different than before.

I grabbed the stack trace:
```
/usr/local/lib/ruby/gems/2.4.0/gems/google-api-client-0.11.2/lib/google/apis/core/json_representation.rb:142:in `to_json': wrong number of arguments (given 1, expected 0) (ArgumentError)
	from /usr/local/lib/ruby/gems/2.4.0/gems/json-2.1.0/lib/json/common.rb:286:in `generate'
	from /usr/local/lib/ruby/gems/2.4.0/gems/json-2.1.0/lib/json/common.rb:286:in `pretty_generate'
	from dfareporting.accounts.get.frag.rb:22:in `<main>'
```

saicheems:
#576

PR #1265, opened 2017-05-12
Update Java grpc metadata for new staging structure

shinfan:
Make the build file depending on the local proto artifact instead of Maven.

cc/ @lukesneeringer


michaelbausor:
Is the intention that the updated build.gradle file in this PR is specifically for api-client-staging, and then we will also generate a second build.gradle file that can be used to release the package? Won't they both need to sit in the same location in api-client-staging?

shinfan:
@michaelbausor Yes there will be both build.gradle sitting in the same location (the release one will be named as build_release.gradle) while during the releasing process we will need some extra steps to swap the staging build file and the release build file.

shinfan:
@michaelbausor  As discussed I have updated the PR to support both build files. I have also updated the api-client-staging PR as well https://github.com/googleapis/api-client-staging/pull/297

src/main/java/com/google/api/codegen/grpcmetadatagen/java/JavaGrpcPackageMetadataTransformer.java
```diff
@@ -22,7 +22,8 @@
   private final Map<String, String> snippetsOutput =
       ImmutableMap.of(
           "LICENSE.snip", "LICENSE",
-          "metadatagen/java/grpc/package.snip", "build.gradle");
+          "metadatagen/java/grpc/package.snip", "build.gradle",
```
garrettjonesgoogle: Can we rename the snip file to "build.gradle.snip"? It's a little annoying that the file name bears no resemblance to the output.

src/main/resources/com/google/api/codegen/metadatagen/java/grpc/package_staging.snip
```diff
@@ -0,0 +1,59 @@
...
+      @end
+    @case "GRPC"
+      compile "io.grpc:grpc-stub:{@metadata.grpcVersionBound.lower}"
+      compile "io.grpc:grpc-protobuf:{@metadata.grpcVersionBound.lower}"
+      compile project(":{@metadata.protoPackageName}")
+    @default
+      $unhandledCase: {@metadata.generationLayer}$
+    @end
+  }
+@end
```
garrettjonesgoogle: add newline

src/main/resources/com/google/api/codegen/metadatagen/java/grpc/package_staging.snip
```diff
@@ -0,0 +1,59 @@
+@snippet generate(metadata)
```
garrettjonesgoogle: What makes it so that we can't use a single build.gradle for both staging & releasing?

src/main/java/com/google/api/codegen/grpcmetadatagen/java/JavaGrpcPackageMetadataTransformer.java
shinfan: Done

src/main/resources/com/google/api/codegen/metadatagen/java/grpc/package_staging.snip
shinfan: Done

src/main/resources/com/google/api/codegen/metadatagen/java/grpc/package_staging.snip
shinfan: Done

shinfan:
@garrettjonesgoogle Sorry that I accidentally force pushed the PR so the history might get lost. PTAL

garrettjonesgoogle:
LGTM

PR #1267, opened 2017-05-15
NodeJS: Update version index to support partial veneers

landrito:
Note: I need to add another service in library.proto to cover the case where there is multiple services in an api.

Here is an example of that generated for pubsub: https://gist.github.com/landrito/32c78c19f6a72865da4afa8fa529d4a9


lukesneeringer:
@landrito is OOO for the rest of the week, and I am wanting to get Node unit testing working in api-client-staging. Therefore, merging this on his behalf.

PR #1268, opened 2017-05-15
NodeJS: Updates to package.json

landrito:
This is to update the `package.json` with the manual updates made to the `package.json` found in GCN. I used monitoring as an example.


lukesneeringer:
Could you use the Vision semi-gapic as an example instead? It should be the canonical approach (and we should make monitoring look like that).

lukesneeringer:
@landrito is OOO for the rest of the week, and I am wanting to get Node unit testing working in api-client-staging. Therefore, merging this on his behalf.

PR #1274, opened 2017-05-17
Fix JSON print in Ruby

saicheems:
It turns out that `to_h` is not a method available on every object
returned by the samples at the moment. Also, there's some kind of
decoding bug showing up in certain samples where the JSON module is
unable to unparse the response.

It's troublesome to have so much boilerplate just to pretty print. This
commit removes the pretty calls in favor of the vanilla `to_json` method
implemented by the base model class in the Ruby client. In the future,
once the bugs w.r.t `pretty_generate` have been resolved, we can revert.


saicheems:
```
/home/travis/.rvm/rubies/ruby-2.3.1/lib/ruby/2.3.0/json/common.rb:286:in `encode': "\x8A" from ASCII-8BIT to UTF-8 (Encoding::UndefinedConversionError)
	from /home/travis/.rvm/rubies/ruby-2.3.1/lib/ruby/2.3.0/json/common.rb:286:in `generate'
	from /home/travis/.rvm/rubies/ruby-2.3.1/lib/ruby/2.3.0/json/common.rb:286:in `pretty_generate'
```

^ Here's one of the errors I'm getting from my runtime tests...

vchudnov-g:
LGTM

Just curious: What was `to_h` used for? To print the hash of the response object? Why would we not  always print the JSON representation?

saicheems:
What we had originally done was this:

```
JSON.pretty_generate(response)
```

Which converted the Ruby object to a pretty JSON string. At some point, I updated the `representable` gem in `google-api-ruby-client` in response to a popular issue, which broke the JSON conversion in that specific gem. It's not a critical issue (P2, maybe) so I haven't had time to look into that yet.

I switched it to use `to_h` and `JSON.pretty_unparse` to try to get the same result - `to_h` does return the hash of the object, and `pretty_unparse` returns a pretty JSON string from the hash. Unfortunately, it also doesn't work in some circumstances described in the commit message...

`to_json` does always work (so far, in my tests, anyway), but it's not pretty printed. I'll try to fix the JSON bug when I can and revert back to what the sample was.

PR #1281, opened 2017-05-18
Migrate PHP to new protobuf implementation

michaelbausor:
Significant changes:
- Protobuf fields are private, so we use getters and setters in paging descriptor instead of field access
- Repeated and map fields use different getters and setters
- Nested messages are named differently, with an underscore separator instead of updated namespace

This PR should not be merged until changes in GAX and google-cloud-php are also ready


src/main/java/com/google/api/codegen/transformer/php/PhpModelTypeNameConverter.java
```diff
@@ -119,10 +127,76 @@ public TypeName getTypeNameForElementType(TypeRef type) {
 
   @Override
   public TypeName getTypeName(ProtoElement elem) {
-    if (elem.getFullName().equals("google.protobuf.Empty")) {
-      return typeNameConverter.getTypeName("\\google\\protobuf\\EmptyC");
+    try {
+      return getTypeName(elem, 20);
```
garrettjonesgoogle: make a constant out of the magic `20`

src/main/java/com/google/api/codegen/transformer/php/PhpModelTypeNameConverter.java
```diff
@@ -119,10 +133,76 @@ public TypeName getTypeNameForElementType(TypeRef type) {
...
+   * <ul>
+   *   <li>\Example\Top
+   *   <li>\Example\Top_Nested
+   *   <li>\Example\Top_Nested_DeepNested
+   * </ul>
+   *
+   * <p>To correctly output these type names, we need to check whether the parent of a proto element
+   * is a message, and if so use '_' as a separator.
+   */
+  private TypeName getTypeName(ProtoElement elem, int maxDepth) {
```
shinfan: You can remove the maxDepth parameter and use MAX_NESTED_DEPTH directly in this method.

src/main/java/com/google/api/codegen/transformer/php/PhpImportSectionTransformer.java
```diff
@@ -37,7 +37,7 @@ public ImportSectionView generateImportSection(
     return generateImportSection(context.getTypeTable().getImports());
   }
 
-  private ImportSectionView generateImportSection(Map<String, TypeAlias> typeImports) {
+  public ImportSectionView generateImportSection(Map<String, TypeAlias> typeImports) {
```
shinfan: Can this be package-private?

src/main/java/com/google/api/codegen/transformer/php/PhpModelTypeNameConverter.java
```diff
@@ -119,10 +133,76 @@ public TypeName getTypeNameForElementType(TypeRef type) {
 
   @Override
   public TypeName getTypeName(ProtoElement elem) {
-    if (elem.getFullName().equals("google.protobuf.Empty")) {
-      return typeNameConverter.getTypeName("\\google\\protobuf\\EmptyC");
+    try {
+      return getTypeName(elem, MAX_NESTED_DEPTH);
+    } catch (IllegalStateException e) {
```
shinfan: I don't think it is necessary to catch and re-throw the exception here.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpCommonTransformer.java
```diff
@@ -102,6 +102,7 @@ public ParamWithSimpleDoc makeParam(
         .typeName(typeName)
         .setCallName("")
         .addCallName("")
+        .getCallName("")
         .isMap(false)
         .isArray(false)
         .isPrimitive(false)
```
shinfan: Not directly related to your PR, but I think here we should create a Type enum (MAP, ARRAY, PRIMITIVE) instead of having 3 flags.

src/main/java/com/google/api/codegen/transformer/php/PhpModelTypeNameConverter.java
michaelbausor: No we can't - this is called recursively with different values on line 176

src/main/java/com/google/api/codegen/transformer/php/PhpModelTypeNameConverter.java
michaelbausor: This is to help with debugging, so that the top level `ProtoElement` object appears in the exception. This information is not available at the exception throw sites (lines 169 and 173)

src/main/java/com/google/api/codegen/transformer/php/PhpImportSectionTransformer.java
michaelbausor: Yes, done.

src/main/java/com/google/api/codegen/transformer/php/PhpModelTypeNameConverter.java
shinfan: Oh see, this is a recursion.

src/main/java/com/google/api/codegen/transformer/php/PhpModelTypeNameConverter.java
shinfan: Ok I see.

shinfan:
LGTM after tests pass

PR #1284, opened 2017-05-22
Ruby: Update Rakefile to include doc tasks and empty acceptance test task

landrito:


src/main/java/com/google/api/codegen/viewmodel/metadata/PackageMetadataView.java
```diff
@@ -88,6 +88,9 @@ public String resourceRoot() {
   @Nullable
   public abstract String versionPath();
```
geigerj: I know we haven't been consistent about this previously, but it would be useful to have javadoc either on the builder or the accessor.

src/main/java/com/google/api/codegen/transformer/ruby/RubyPackageMetadataTransformer.java
```diff
@@ -222,6 +222,9 @@ private ViewModel generateMetadataView(
                 productConfig, ImportSectionView.newBuilder().build(), surfaceNamer))
         .hasSmokeTests(hasSmokeTests)
         .versionPath(surfaceNamer.getVersionIndexFileImportName())
+        .versionNamespace(
+            surfaceNamer.getNamespace(
+                new InterfaceView().getElementIterable(model).iterator().next()))
```
geigerj: This just grabs any one of the proto services defined in the API, right? What happens when there are multiple (like with PubSub)?

src/main/resources/com/google/api/codegen/ruby/Rakefile.snip
```diff
@@ -4,11 +4,15 @@
   {@header(metadata.fileHeader)}
 
   {@unitTestsTask(metadata)}
-  @if smokeTestsTask(metadata)
+  @if metadata.hasSmokeTests
 
     {@smokeTestsTask(metadata)}
   @end
 
+  {@acceptanceTestTask(metadata)}
```
geigerj: Why do we need this if it's empty?

landrito: This is needed because the metapackages Rakefile calls the `:acceptance` task on all the gems in the metapackage.

https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/Rakefile#L92

src/main/java/com/google/api/codegen/transformer/ruby/RubyPackageMetadataTransformer.java
landrito: This makes the assumption that all services given to the generator will have the same namespace. For pubsub, both the publisher and the subscriber are in the `Google::Pubsub::V1` namespace so there is no discrepancy between the two. Unless there is a case where two services are passed to the generator that don't have the same namespace, this should be fine.

geigerj: Suggest to formalize this assumption by using a helper method that checks that all services have the same namespace and issues a warning or exception if not.

src/main/java/com/google/api/codegen/viewmodel/metadata/PackageMetadataView.java
landrito: Done.

src/main/java/com/google/api/codegen/transformer/ruby/RubyPackageMetadataTransformer.java
landrito: Done.

src/main/java/com/google/api/codegen/transformer/ruby/RubyPackageMetadataTransformer.java
```diff
@@ -222,9 +226,20 @@ private ViewModel generateMetadataView(
                 productConfig, ImportSectionView.newBuilder().build(), surfaceNamer))
         .hasSmokeTests(hasSmokeTests)
         .versionPath(surfaceNamer.getVersionIndexFileImportName())
+        .versionNamespace(surfaceNamer.getNamespace(interfaces.iterator().next()))
         .build();
   }
 
+  private void validateVersionNamespaces(Iterable<Interface> interfaces, SurfaceNamer namer) {
```
geigerj: nit: would prefer this return the version namespace string and be called directly as an argument to the `versionNamespace` metadataView builder method. This makes it immediately obvious why the check is necessary and prevents the check from being separated accidentally from the builder in a future refactoring.

landrito: Done!

PR #1292, opened 2017-05-24
Implement GAPIC metadata-gen for Java

shinfan:
This PR enables toolkit to generate `build.gradle` for each Java GAPIC library that is used for staging purpose. 

After this change there would be "zero" manual effort involved of adding new Java libraries (except for manually editing artman config).

cc @lukesneeringer


src/main/java/com/google/api/codegen/transformer/PackageMetadataNamer.java
```diff
@@ -34,6 +34,11 @@ public String getProtoPackageName() {
     return getNotImplementedString("PackageMetadataNamer.getProtoPackageName");
   }
 
+  /** Returns the artifact identifier for a gRPC classes package */
+  public String getGrpcPackageName() {
+    return getNotImplementedString("PackageMetadataNamer.getProtoPackageName");
```
garrettjonesgoogle: Wrong error string

src/main/java/com/google/api/codegen/transformer/java/JavaGapicMetadataTransformer.java
```diff
@@ -0,0 +1,58 @@
+/* Copyright 2016 Google Inc
```
garrettjonesgoogle: use 2017

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
```diff
@@ -59,6 +61,9 @@
 
   protected abstract Map<TargetLanguage, Map<String, VersionBound>> protoPackageDependencies();
 
+  @Nullable
+  protected abstract Map<TargetLanguage, Map<String, VersionBound>> protoPackageTestDependencies();
```
shinfan: Note: I have added this protoPackageTestDependencies field to specify test dependencies for GAPIC library (specifically for Pubsub since the pubsub tests depends on the IAM gRPC package)

src/main/java/com/google/api/codegen/transformer/PackageMetadataNamer.java
shinfan: Done

src/main/java/com/google/api/codegen/transformer/java/JavaGapicMetadataTransformer.java
shinfan: Done

garrettjonesgoogle:
LGTM

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
pongad: Is it possible to make this empty map instead of null?

Guava also have Table, which is essentially 2D map, but looks like we should stick to map of map to look like protoPackageDependencies.

src/main/java/com/google/api/codegen/transformer/PackageMetadataTransformer.java
```diff
@@ -101,4 +106,22 @@
...
   }
+
+  private List<PackageDependencyView> getDependencies(
+      Map<String, VersionBound> dependencies, Set<String> whitelistedDependencies) {
+    List<PackageDependencyView> protoPackageDependencies = new ArrayList<>();
+    if (dependencies != null) {
+      for (Map.Entry<String, VersionBound> entry : dependencies.entrySet()) {
+        if (entry.getValue() != null
+            && (whitelistedDependencies == null
+                || whitelistedDependencies.contains(entry.getKey()))) {
```
pongad: Apart from null checks, I think this can be simpler written (partly-pseudocode)
```java
Map dep = new Map(dependencies); // make copy
dep.keySet().retainAll(whitelistedDependencies);
for (entry : deps.entrySet()) {
  list.add(PackageDependencyView.create(...);
}
```

src/main/java/com/google/api/codegen/transformer/java/JavaGapicMetadataTransformer.java
```diff
@@ -0,0 +1,58 @@
...
+
+  private final PackageMetadataConfig packageConfig;
+
+  public JavaGapicMetadataTransformer(PackageMetadataConfig packageConfig) {
+    this.packageConfig = packageConfig;
+  }
+
+  @Override
+  public List<ViewModel> transform(Model model, GapicProductConfig productConfig) {
+    List<ViewModel> viewModels = Lists.newArrayList();
```
pongad: return this.transfrom()?

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
shinfan: This is a protected method and the public accessor is `Map<String, VersionBound> protoPackageTestDependencies()`. Similar to `protoPackageDependencies` this method returns null if the language key is not found.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicMetadataTransformer.java
shinfan: Sorry that this looks a little bit dumb. I tried your suggestion as my first attempt but my IDE complains that `this.transform` returns `List<PackageMetadataView>` but here we need `List<ViewModel>`.

The alternative would be:
```java
(List<ViewModel>)(List<?>)this.transform(model, packageConfig);
```

src/main/java/com/google/api/codegen/transformer/PackageMetadataTransformer.java
shinfan: Done

shinfan:
@pongad PTAL

src/main/java/com/google/api/codegen/transformer/java/JavaGapicMetadataTransformer.java
pongad: I see, this makes sense.

PR #1297, opened 2017-05-25
Add a command line option for a Ruby names file

saicheems:
This option allows a custom Ruby names file to be passed to the sample
generator. It's intended purpose is to make generation of samples for
the mock tests easier.

### DIFF
--- a/src/main/java/com/google/api/codegen/discovery/DiscoveryProviderFactory.java
+++ b/src/main/java/com/google/api/codegen/discovery/DiscoveryProviderFactory.java
@@ -17,11 +17,15 @@ package com.google.api.codegen.discovery;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.google.api.Service;
 import com.google.api.codegen.ApiaryConfig;
+import java.io.File;
 import java.util.List;
 
 /** A factory for DiscoveryProviders which perform code generation. */
 public interface DiscoveryProviderFactory {
-  /** Create the provider from the given service, apiaryConfig, sampleConfigOverrides, and id. */
   DiscoveryProvider create(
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/api/codegen/discovery/DiscoveryProviderFactory.java
```diff
@@ -17,11 +17,15 @@
...
 
 /** A factory for DiscoveryProviders which perform code generation. */
 public interface DiscoveryProviderFactory {
-  /** Create the provider from the given service, apiaryConfig, sampleConfigOverrides, and id. */
   DiscoveryProvider create(
-      Service service, ApiaryConfig apiaryConfig, List<JsonNode> sampleConfigOverrides, String id);
+      Service service,
+      ApiaryConfig apiaryConfig,
+      List<JsonNode> sampleConfigOverrides,
+      File rubyNamesFile,
```
vchudnov-g: This seems like a non-language-specific file (as do some of the others in this PR), so why is there a parameter for Ruby and not for any other language?

saicheems:
This is admittedly more for convenience than anything else - since once toolkit is packaged into jar it will always reference its internal names file.

For my tests it was a bit annoying because I'd have to generate the Ruby APIs first, then copy those back into toolkit, then build toolkit.

src/main/java/com/google/api/codegen/discovery/DiscoveryProviderFactory.java
saicheems: Only Ruby uses a separate file to determine the names of methods/variables. That file is generated by the Ruby client library. Toolkit has a hard reference to a file in `resources`, which requires us to update the repos copy frequently. So passing the file through provides some convenience for me.

PR #1298, opened 2017-05-25
NodeJS: Metadata updates.

landrito:
Based on review of DLP gen: https://github.com/GoogleCloudPlatform/google-cloud-node/pull/2323#pullrequestreview-40387018

### DIFF
--- a/src/main/resources/com/google/api/codegen/nodejs/version_index.snip
+++ b/src/main/resources/com/google/api/codegen/nodejs/version_index.snip
@@ -46,8 +46,10 @@
     {@comments(@service.doc.lines)}
     *
     * @@param {object=} options - [Configuration object](#/docs).
-    * @@param {number=} options.port - The port on which to connect to the remote host.
-    * @@param {string=} options.servicePath - The domain name of the API remote host.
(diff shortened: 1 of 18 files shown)


src/main/resources/com/google/api/codegen/nodejs/version_index.snip
```diff
@@ -46,8 +46,10 @@
     {@comments(@service.doc.lines)}
     *
     * @@param {object=} options - [Configuration object](#/docs).
-    * @@param {number=} options.port - The port on which to connect to the remote host.
-    * @@param {string=} options.servicePath - The domain name of the API remote host.
+    * @@param {number=} options.port - The port on which to connect to
+    *   the remote host.
```
stephenplusplus: We do 4 space indent -- can't remember if it is because of JSdocs/dox (the parser) having problems, or just convention. If it's not too annoying, feel free to change.

landrito: Done!

stephenplusplus: Thank you 🙇

PR #1299, opened 2017-05-25
Readme: Update readme templates to add api summaries.

landrito:
cc: @omaray

### DIFF
--- a/src/main/resources/com/google/api/codegen/readme_rst.snip
+++ b/src/main/resources/com/google/api/codegen/readme_rst.snip
@@ -5,7 +5,7 @@
 @end
 
 @private title(metadata)
-  {@metadata.targetLanguage} Client{@plurality(metadata.hasMultipleServices)} for `{@metadata.fullName}`_ (`{@metadata.developmentStatusTitle}`_)
+  {@metadata.targetLanguage} Client{@plurality(metadata.hasMultipleServices)} for {@metadata.fullName} (`{@metadata.developmentStatusTitle}`_)
   ==================================================================================================
 
   `{@metadata.fullName}`_: {@metadata.apiSummary}
--- a/src/main/resources/com/google/api/codegen/readme_rst.snip
+++ b/src/main/resources/com/google/api/codegen/readme_rst.snip
@@ -8,7 +8,7 @@
   {@metadata.targetLanguage} Client{@plurality(metadata.hasMultipleServices)} for {@metadata.fullName} (`{@metadata.developmentStatusTitle}`_)
   ==================================================================================================
 
-  Idiomatic {@metadata.targetLanguage} client for `{@metadata.fullName}`_
+  Idiomatic {@metadata.targetLanguage} client for `{@metadata.fullName}`_: {@metadata.apiSummary}
 
   - `Client Library Documentation`_
   - `Product Documentation`_
--- a/src/main/resources/com/google/api/codegen/readme_rst.snip
+++ b/src/main/resources/com/google/api/codegen/readme_rst.snip
@@ -5,12 +5,10 @@
 @end
 
 @private title(metadata)
-  {@metadata.targetLanguage} Client{@plurality(metadata.hasMultipleServices)} for {@metadata.fullName} (`{@metadata.developmentStatusTitle}`_)
+  {@metadata.targetLanguage} Client{@plurality(metadata.hasMultipleServices)} for `{@metadata.fullName}`_ (`{@metadata.developmentStatusTitle}`_)
   ==================================================================================================
 
-  Idiomatic {@metadata.targetLanguage} client for `{@metadata.fullName}`_
-  -----------------------------------------------------------------------
-  {@metadata.apiSummary}
+  `{@metadata.fullName}`_: {@metadata.apiSummary}
 
   - `Client Library Documentation`_
   - `Product Documentation`_
--- a/src/main/resources/com/google/api/codegen/readme_rst.snip
+++ b/src/main/resources/com/google/api/codegen/readme_rst.snip
@@ -8,8 +8,8 @@
   {@metadata.targetLanguage} Client{@plurality(metadata.hasMultipleServices)} for {@metadata.fullName} (`{@metadata.developmentStatusTitle}`_)
   ==================================================================================================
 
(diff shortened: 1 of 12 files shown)


landrito:
# Example:
# Node.js Clients for Stackdriver Logging API ([Alpha](https://github.com/GoogleCloudPlatform/google-cloud-node#versioning))

### Idiomatic Node.js clients for [Stackdriver Logging API][Product Documentation]: 
The Stackdriver Logging API lets you write log entries and manage your logs, log sinks and logs-based metrics.
- [Client Library Documentation][]
- [Product Documentation][]

## Quick Start
In order to use this library, you first need to go through the following steps:

1. [Select or create a Cloud Platform project.](https://console.cloud.google.com/project)
2. [Enable the logging api.](https://console.cloud.google.com/apis/api/logging)
3. [Setup Authentication.](https://googlecloudplatform.github.io/google-cloud-node/#/docs/google-cloud/master/guides/authentication)

### Installation
```
$ npm install @google-cloud/logging
```

### Preview
#### LoggingServiceV2Client
```js
 var loggingV2 = require('@google-cloud/logging').v2({
   // optional auth parameters.
 });

 var client = loggingV2.loggingServiceV2Client();
 var formattedLogName = client.logPath(projectId, "test-" + Date.now().toString());
 var resource = {};
 var labels = {};
 var entries = [];
 client.writeLogEntries({logName: formattedLogName}).then(function(responses) {
     var response = responses[0];
     // doThingsWith(response)
 }).catch(function(err) {
     console.error(err);
 });
```

### Next Steps
- Read the [Client Library Documentation][] for Stackdriver Logging API to see other available methods on the client.
- Read the [Stackdriver Logging API Product documentation][Product Documentation] to learn more about the product and see How-to Guides.
- View this [repository's main README](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/master/README.md) to see the full list of Cloud APIs that we cover.

[Client Library Documentation]: https://googlecloudplatform.github.io/google-cloud-node/#/docs/logging
[Product Documentation]: https://cloud.google.com/logging

landrito:
Probably only need one reviewer since its a pretty simple change. Just wanted to tag all three leads so they are aware of the change.

src/test/java/com/google/api/codegen/testdata/python_README_library.baseline
```diff
@@ -2,7 +2,9 @@
 Python Client for Google Example Library API (`Alpha`_)
 ==================================================================================================
 
-Idiomatic Python client for `Google Example Library API`_
+Idiomatic Python client for `Google Example Library API`_:
```
geigerj: nit: no colon if this is a header?

src/test/java/com/google/api/codegen/testdata/python_README_library.baseline
```diff
@@ -2,7 +2,9 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
geigerj: These indentation levels don't make sense to me: should this be dashes?

landrito: I was thinking it would look better to go from `h1` to `h3` on a purely subjective stylistic level. If you think it should be dashes that's fine!

src/test/java/com/google/api/codegen/testdata/python_README_library.baseline
landrito: Done.

src/test/java/com/google/api/codegen/testdata/python_README_library.baseline
landrito: Done.

geigerj: Ah, so I think with reST the header level is determined implicitly. That means there's no way to avoid having the tilde be interpreted as h2 here (which then throws off the other heading levels).

src/main/resources/com/google/api/codegen/readme_rst.snip
```diff
@@ -9,6 +9,8 @@
   ==================================================================================================
 
   Idiomatic {@metadata.targetLanguage} client for `{@metadata.fullName}`_
+  -----------------------------------------------------------------------
```
geigerj: optional: line 37 of this file incorrectly refers to `monitoring` -- feel free to address in a follow-up PR if you prefer, since it's not directly linked to this one.

omaray:
This is great folks! A few notes

- Should be singular: "Node.js Client..." instead of "Node.js Clients..."
- Then the subtitle "Idiomatic Node.js Client for Stackdriver API" seems to add redundancy a bit with the title above and the description underneath. Perhaps we could have the title as-is, remove the subtitle, and then have the first line be:

"Idiomatic Node.js client for Stackdriver Logging API: <description>". Where <description> would be "let's write logs....".

But this can also depend on whether or not <description> also contains the words "Stackdriver logging API" as the start. Then it would be weird!

Another option.

(1) Remove the subtitle that says "Idiomatic Node.js client for Stackdriver API".
(2) Make the "Stackdriver Logging API" in the title as a link (so the docs on cloud.google.com)
(3) Have the very first sentence of the readme (that is not a title) be the description we have right now (basically what is already there).

src/main/resources/com/google/api/codegen/readme_rst.snip
landrito: Fixed in https://github.com/googleapis/toolkit/pull/1298

landrito:
So the logic for the plurality of clients is correct. I think it's simply because some of the services outlined the [service yaml](https://github.com/googleapis/googleapis/blob/master/google/logging/logging.yaml#L9) are not generated. I'll have to think of a solution.

Gotta get going and will be OOO for the next week. I'll do the second option you said and see where that takes us!

landrito:
Updated PTAL. Feel free to merge if it looks good!

omaray:
Looks good. Understood for plurality (we have 2 clients). Last comment: I would remove the link from the title since it's in the first part of the description. Feels like too many links to the docs otherwise. But otherwise lgtm.

PR #1301, opened 2017-05-26
Fix print for map resources in Ruby

saicheems:
Also fix a missing ')' bug.
The '<<' was changed to '+' because the key string is immutable.

### DIFF
--- a/src/test/java/com/google/api/codegen/testdata/discoveries/ruby/ruby_compute.v1.json.baseline
+++ b/src/test/java/com/google/api/codegen/testdata/discoveries/ruby/ruby_compute.v1.json.baseline
@@ -31,7 +31,7 @@ end
 
 items.each do |name, addresses_scoped_list|
   # TODO: Change code below to process each (name, addresses_scoped_list) pair:
-  puts String(name << " => " << addresses_scoped_list.to_json
+  puts String(name) + " => " + addresses_scoped_list.to_json
 end
 # BEFORE RUNNING:
 # ---------------
@@ -210,7 +210,7 @@ end
 
 items.each do |name, autoscalers_scoped_list|
   # TODO: Change code below to process each (name, autoscalers_scoped_list) pair:
-  puts String(name << " => " << autoscalers_scoped_list.to_json
+  puts String(name) + " => " + autoscalers_scoped_list.to_json
 end
 # BEFORE RUNNING:
 # ---------------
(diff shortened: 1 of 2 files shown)


saicheems:
Well it's not so strange, see the Java counterpart as an example:
https://github.com/googleapis/toolkit/blob/master/src/test/java/com/google/api/codegen/testdata/discoveries/java/java_compute.v1.json.baseline#L372

Each response in contains a map, in this case we call `fetch_all` to grab all the responses, which returns a list of maps. Then we iterate in a loop over each map to print its contents.

I tried putting `pretty_generate` back now that the bug is fixed... it doesn't crash anymore, but it doesn't actually pretty print either. It behaves the same as `to_json` does at the moment :/

I did try to poke around the client library to try to get the old behavior back, but I couldn't get it to work. It seems there can be issues with supporting the interface with the native JSON library's `pretty_generate` while also having custom `to_json` functions.

saicheems:
Nvm, I see what you're saying - Ruby generally pulls out the page streaming resource in the `fetch_all` call. The reason it doesn't do so here I think is a result of this line:

https://github.com/google/google-api-ruby-client/blob/c645e7d2f893dc5165ce6354b07f147387bce109/lib/google/apis/core/base_service.rb#L72

Since the result of a list function that returns an object is a map, the generator yields the full object instead of a number of map entries like how arrays are split apart (see line 66).

saicheems:
I filed an issue here:

https://github.com/google/google-api-ruby-client/issues/584

src/test/java/com/google/api/codegen/testdata/discoveries/ruby/ruby_compute.v1.json.baseline
```diff
@@ -29,9 +29,11 @@ items = service.fetch_all do |token|
...
 end
 
-items.each do |name, addresses_scoped_list|
-  # TODO: Change code below to process each (name, addresses_scoped_list) pair:
-  puts String(name << " => " << addresses_scoped_list.to_json
+items.each do |item|
+  item.each do |name, addresses_scoped_list|
+    # TODO: Change code below to process each (name, addresses_scoped_list) pair:
+    puts String(name) + " => " + addresses_scoped_list.to_json
+  end
```
tcoffee-google: Given the way the library currently handles maps, it seems to me that to be conceptually right (and avoid confusing the user), the variable names for this case will need to change---something like: `items` => `page`, `item` => `items`.

saicheems: I'm not sure if `page` is the right name either, the collection returned by `fetch_all` is the list of `items` pulled from the response, not the page. I think that `items` is still the correct name.

https://github.com/saicheems/toolkit/blob/master/src/test/java/com/google/api/codegen/testdata/discoveries/compute.v1.json#L199

Here's the discovery, the items field is what's collected by `fetch_all`.

saicheems:
Note: This depends on the next client release with https://github.com/google/google-api-ruby-client/pull/586

saicheems:
Updated, PTAL

PR #1302, opened 2017-05-26
Fix pprint for map resources in Python

saicheems:


saicheems:
Compute samples with page streaming resources that are maps currently fail with this stack trace:

```
        Traceback (most recent call last):
          File "compute.addresses.aggregatedList.py", line 26, in <module>
            pprint(name, addresses_scoped_list)
          File "/usr/lib/python2.7/pprint.py", line 59, in pprint
            printer.pprint(object)
          File "/usr/lib/python2.7/pprint.py", line 117, in pprint
            self._format(object, self._stream, 0, 0, {}, 0)
          File "/usr/lib/python2.7/pprint.py", line 143, in _format
            write = stream.write
        AttributeError: 'dict' object has no attribute 'write'
```

`pprint` can't be called with multiple arguments the same way that `print` can.

src/main/resources/com/google/api/codegen/py/sample.snip
```diff
@@ -107,7 +107,7 @@
                 @if class.pageStreaming.isResourceMap
                     for {@class.pageStreaming.resourceKeyVarName}, {@class.pageStreaming.resourceValueVarName} in {@class.responseVarName}['{@class.pageStreaming.resourceFieldName}'].items():
                         @# TODO: Change code below to process each ({@class.pageStreaming.resourceKeyVarName}, {@class.pageStreaming.resourceValueVarName}) item:
-                        pprint({@class.pageStreaming.resourceKeyVarName}, {@class.pageStreaming.resourceValueVarName})
+                        pprint({{@class.pageStreaming.resourceKeyVarName}: {@class.pageStreaming.resourceValueVarName}})
```
tcoffee-google: I'd prefer my [original formulation](https://github.com/saicheems/toolkit/blob/8d9d0bd9e3059f8b20ffef6051635c7d7a85c8a1/src/main/resources/com/google/api/codegen/py/discovery_fragment.snip) that uses a tuple instead --- it seems too misleading to print a series of single-item dictionaries to represent a single map.

vchudnov-g:
What's the motivation for this? AFAICT, the values are being referenced the same way in the templates.

saicheems:
@vchudnov-g it's because the current `pprint` call for map entries is not correct and will raise an exception. The issue is that `pprint` does not take multiple arguments the same way that `print` does. It expects a single object and some args. To address that I've wrapped the key value pair we print for map entries in a tuple.

src/main/resources/com/google/api/codegen/py/sample.snip
saicheems: Changed to a tuple, ptal

PR #1305, opened 2017-05-27
Resource disco generation

andreamlin:
This generates a java Serializable class for every Discovery doc schema in a given Discovery doc. There are still a couple kinks to work out, including adding in the java package name. After manually putting in a package name, I manually tested the class by sending an Instance List http request and deserializing the JSON response into an InstanceList generated class, which also contains nested generated classes.


src/main/resources/com/google/api/codegen/java/disco_schema.snip
```diff
@@ -0,0 +1,141 @@
...
+    {@getters(apiFile.schema)}
+    {@newBuilder(apiFile.schema)}
+    {@builder(apiFile.schema)}
+    {@toString(apiFile.schema)}
+    {@equalsMethod(apiFile.schema)}
+    {@hashCode(apiFile.schema)}
+  }
+@end
+
+@private toString(xschema)
```
garrettjonesgoogle: The "x" prefix thing was something I started a while back but decided to stop using, and I haven't gotten a chance to remove it everywhere yet. Go ahead and remove it from your code (so use `schema` here).

src/main/java/com/google/api/codegen/discogapic/transformer/java/JavaDiscoGapicSchemaToViewTransformer.java
```diff
@@ -0,0 +1,195 @@
...
+    ModelTypeTable typeTable = context.getModelTypeTable();
+    typeTable.saveNicknameFor("com.google.api.core.BetaApi");
+    typeTable.saveNicknameFor("java.io.Serializable");
+    typeTable.saveNicknameFor("java.util.List");
+    typeTable.saveNicknameFor("javax.annotation.Generated");
+  }
+
+  // Return the corresponding Java identifier for a given Discovery doc typeName and format.
+  // https://developers.google.com/discovery/v1/type-format.
+  private String typeToJavaType(String name, Schema schema) {
```
garrettjonesgoogle: We should create something analogous to `JavaModelTypeNameConverter` for discovery types - it would be named something like `JavaDiscoTypeNameConverter`. Also, don't use `ModelTypeTable` (referenced above) because this is for processing the type model from protobuf. You will need a `DiscoTypeTable` (or `SchemaTypeTable`?)

src/main/java/com/google/api/codegen/discogapic/transformer/SchemaToViewTransformer.java
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+package com.google.api.codegen.discogapic.transformer;
+
+import com.google.api.codegen.config.GapicProductConfig;
+import com.google.api.codegen.discovery.Document;
+import com.google.api.codegen.viewmodel.ViewModel;
+import java.util.List;
+
+public interface SchemaToViewTransformer {
```
garrettjonesgoogle: This interface is the same as `DocumentToViewTransformer`, and they serve the same intent, so I don't think we need distinct interfaces.

src/main/java/com/google/api/codegen/discogapic/DiscoGapicProvider.java
```diff
@@ -28,30 +29,55 @@
...
+        // generation failed; failures are captured in the model.
+        continue;
+      }
+      docs.put(surfaceDoc.outputPath(), doc);
+    }
+
+    return docs;
+  }
 
+  public Map<String, Doc> generateSchemas(String snippetFileName) {
```
garrettjonesgoogle: Don't duplicate the code between this new function (`generateSchemas`) and the existing one (`generate`).

src/main/resources/com/google/api/codegen/java/disco_schema.snip
```diff
@@ -0,0 +1,141 @@
...
+  @@Override
+    public String toString() {
+      return "{@xschema.typeName}{"
+      @join property : xschema.properties
+          + "{@property.name}=" + {@property.name} + ", "
```
garrettjonesgoogle: Won't this mean there is a trailing comma?

src/main/resources/com/google/api/codegen/java/disco_schema.snip
```diff
@@ -0,0 +1,141 @@
...
+   public boolean equals(Object o) {
+     if (o == this) {
+       return true;
+     }
+     if (o instanceof {@xschema.typeName}) {
+       {@xschema.typeName} that = ({@xschema.typeName}) o;
+       return
+       @join property : xschema.properties on "&&"
+
+       this.{@property.name}.equals(that.{@property.name}())
```
garrettjonesgoogle: Indent inside of a `@join` (multiple instances in this file)

src/main/resources/com/google/api/codegen/java/disco_schema.snip
```diff
@@ -0,0 +1,141 @@
...
+@private newBuilder(xschema)
+  public static Builder newBuilder() {
+    return new {@xschema.typeName}.Builder();
+  }
+
+@end
+
+@private getters(xschema)
+  @join property : xschema.properties
+    public {@property.typeName} {@property.name}() {
```
garrettjonesgoogle: I think we should have the name of this function be something like `getFoo()` (match the naming conventions in protobuf)

src/main/resources/com/google/api/codegen/java/disco_schema.snip
```diff
@@ -0,0 +1,141 @@
...
+      return {@property.name};
+    }
+    {@""}
+  @end
+@end
+
+@private constructor(xschema)
+  private {@xschema.typeName}(
+      @join property : xschema.properties on ", "
```
garrettjonesgoogle: Indent by 2 spaces instead of 4

src/main/resources/com/google/api/codegen/java/disco_schema.snip
```diff
@@ -0,0 +1,141 @@
...
+    Builder() {}
+
+    Builder({@xschema.typeName} source) {
+      @join property : xschema.properties
+        this.{@property.name} = source.{@property.name};
+      @end
+    }
+
+    @join property : xschema.properties
+      public Builder {@property.name}({@property.typeName} {@property.name}) {
```
garrettjonesgoogle: I think we should have the name of this function be something like `setFoo(...)` (match the naming conventions in protobuf)

src/main/resources/com/google/api/codegen/java/disco_schema.snip
```diff
@@ -0,0 +1,141 @@
...
+        this.{@property.name} = {@property.name};
+        return this;
+      }
+
+    @end
```
garrettjonesgoogle: The builder should also have get functions.

src/main/resources/com/google/api/codegen/java/disco_schema.snip
```diff
@@ -0,0 +1,141 @@
+@extends "java/common.snip"
+@extends "java/method_sample.snip"
+
+@snippet generate(apiFile)
```
garrettjonesgoogle: In terms of naming/concepts, I want to decouple the input/transformer from the output/view model+templates. What that means concretely is that the transformers are discovery-specific, but the view model and templates shouldn't be discovery specific, so they shouldn't have 'disco' in names or in comments. This file in particular could probably be named `message.snip`, because it's generating a message class. 

The motivating reason for this is to support other input formats in the future, but which use the same generated message templates.

src/main/java/com/google/api/codegen/viewmodel/StaticLangApiSchemaView.java
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * This ViewModel defines the structure of a Discovery doc's "schemas", "properties",
+ * "additionalProperties", and "items".
+ *
+ * This contains a subset of properties in the JSON Schema
+ * https://tools.ietf.org/html/draft-zyp-json-schema-03#section-5.7.
+ */
+@AutoValue
+public abstract class StaticLangApiSchemaView implements ViewModel {
```
garrettjonesgoogle: I would suggest `StaticLangMessageView` (name the file view one in a corresponding way)

src/main/java/com/google/api/codegen/viewmodel/StaticLangApiSchemaView.java
```diff
@@ -0,0 +1,90 @@
...
+
+  // The possibly-transformed ID of the schema from the Discovery Doc
+  public abstract String typeName();
+
+  // The escaped class name for this Schema.
+  public abstract String className();
```
garrettjonesgoogle: Is this used in the template? (I don't see it) Also applies to some others too (`defaultValue`). I'm guessing this is intentional.

src/main/java/com/google/api/codegen/viewmodel/StaticLangApiSchemaView.java
```diff
@@ -0,0 +1,90 @@
...
+  @Nullable
+  public abstract Boolean repeated();
+
+  // There can be arbitrarily nested fields inside of this field.
+  @Nullable
+  public abstract List<SimplePropertyView> properties();
+
+  @Nullable
+  // The typeName of a Schema that contains the full representation of this schema.
+  public abstract StaticLangApiSchemaView ref();
```
garrettjonesgoogle: Is this necessary?

src/main/resources/com/google/api/codegen/java/disco_schema.snip
andreamlin: Changing the join to joining on the string ", " so there is no longer a trailing comma.

src/main/resources/com/google/api/codegen/java/disco_schema.snip
andreamlin: Done.

src/main/resources/com/google/api/codegen/java/disco_schema.snip
andreamlin: Done.

src/main/resources/com/google/api/codegen/java/disco_schema.snip
andreamlin: Done.

src/main/resources/com/google/api/codegen/java/disco_schema.snip
andreamlin: Changed to getFoo() and setFoo() methods.

src/main/resources/com/google/api/codegen/java/disco_schema.snip
andreamlin: Done.

src/main/resources/com/google/api/codegen/java/disco_schema.snip
andreamlin: Done.

src/main/resources/com/google/api/codegen/java/disco_schema.snip
andreamlin: Done.

src/main/java/com/google/api/codegen/discogapic/transformer/java/JavaDiscoGapicSchemaToViewTransformer.java
andreamlin: I started work for a JavaModelTypeNameConverter on a different branch and definitely plan on finishing the implementation for that, but it seems to be a big enough change to be a separate PR.

src/main/java/com/google/api/codegen/viewmodel/StaticLangApiSchemaView.java
andreamlin: Removed.

src/main/java/com/google/api/codegen/viewmodel/StaticLangApiSchemaView.java
andreamlin: Removed className. The other unused ones haven't been implemented yet or are part of the ViewModel interface.

src/main/java/com/google/api/codegen/viewmodel/StaticLangApiSchemaView.java
andreamlin: Done and done.

src/main/java/com/google/api/codegen/discogapic/transformer/SchemaToViewTransformer.java
andreamlin: Removed this and corresponding boilerplate.

src/main/java/com/google/api/codegen/discogapic/DiscoGapicProvider.java
andreamlin: Cleansed.

andreamlin:
Responded to all of the comments in the first round of comments.

src/main/java/com/google/api/codegen/discogapic/DiscoGapicProvider.java
```diff
@@ -28,29 +28,31 @@
   private final Document document;
   private final GapicProductConfig productConfig;
   private final CommonSnippetSetRunner snippetSetRunner;
-  private final DocumentToViewTransformer transformer;
+  private final DocumentToViewTransformer documentTransformer;
```
garrettjonesgoogle: Looks like you could just revert this file back to its original state.

src/main/java/com/google/api/codegen/discogapic/MainDiscoGapicProviderFactory.java
```diff
@@ -49,16 +50,25 @@
                 .setPrefix("src/main/java")
                 .setShouldAppendPackage(true)
                 .build();
-        DiscoGapicProvider mainProvider =
+        DiscoGapicProvider mainClientProvider =
```
garrettjonesgoogle: Maybe just call the variable `clientProvider`

src/main/java/com/google/api/codegen/viewmodel/StaticLangApiMessageView.java
```diff
@@ -0,0 +1,95 @@
...
+
+import com.google.api.codegen.SnippetSetRunner;
+import com.google.api.codegen.discovery.Schema;
+import com.google.auto.value.AutoValue;
+import java.util.List;
+import javax.annotation.Nullable;
+
+/**
+ * This ViewModel defines the structure of a Discovery doc's "schemas", "properties",
+ * "additionalProperties", and "items".
```
garrettjonesgoogle: Change the description here so that it is generically described as a message view model, and then mention Discovery as an example of how it can be used.

src/main/resources/com/google/api/codegen/java/message.snip
```diff
@@ -0,0 +1,146 @@
...
+    {@builder(apiFile.schema)}
+    {@toString(apiFile.schema)}
+    {@equalsMethod(apiFile.schema)}
+    {@hashCode(apiFile.schema)}
+  }
+@end
+
+@private toString(schema)
+  @@Override
+    public String toString() {
```
garrettjonesgoogle: Don't indent more than the `@@Override` line

src/main/java/com/google/api/codegen/discogapic/transformer/java/JavaDiscoGapicSchemaToViewTransformer.java
```diff
@@ -0,0 +1,202 @@
...
+    for (Map.Entry<String, Schema> propertyEntry : schema.properties().entrySet()) {
+      String propertyName = schemaSymbolTable.getNewSymbol(propertyEntry.getKey());
+      Schema property = propertyEntry.getValue();
+      SimplePropertyView.Builder simpleProperty =
+          SimplePropertyView.newBuilder().name(propertyName).repeated(property.repeated());
+      simpleProperty.typeName(typeToJavaType(property));
+
+      // Property class name is Capitalized
+      simpleProperty.capitalizedName(
+          propertyName.substring(0, 1).toUpperCase() + propertyName.substring(1));
```
garrettjonesgoogle: Don't do string casing changes in transformers - that's the job of the surface namer. Since you don't have a namer yet, at a minimum, use code like this to compose new identifiers from existing ones:

https://github.com/googleapis/toolkit/blob/2853924324aeef3cabfe9c9c50a66009ead1f0b3/src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java#L282

src/main/resources/com/google/api/codegen/java/message.snip
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+@end
+
+@private hashCode(schema)
+  @@Override
+  public int hashCode() {
+    int h = 1;
+    @join property : schema.properties
+      h *= 1000003;
```
andreamlin: Is it necessary to use a different prime number? tbh this is just a copy of what AutoValue does. Java strings use 31, and Protobuffers sometimes use "1 << 16 - 1" (although each object in those examples is just an array of things of the same type).

src/main/resources/com/google/api/codegen/java/message.snip
```diff
@@ -0,0 +1,146 @@
...
+    Builder() {}
+
+    Builder({@schema.typeName} source) {
+      @join property : schema.properties
+        this.{@property.name} = source.{@property.name};
+      @end
+    }
+
+    @join property : schema.properties
+      public {@property.typeName} get{@property.capitalizedName}() {
```
andreamlin: Ok, moved the identifier to the view model. Will use a surface namer in a future PR.

src/main/java/com/google/api/codegen/discogapic/DiscoGapicProvider.java
andreamlin: Done.

src/main/java/com/google/api/codegen/discogapic/MainDiscoGapicProviderFactory.java
andreamlin: Done.

src/main/java/com/google/api/codegen/viewmodel/StaticLangApiMessageView.java
andreamlin: Done.

src/main/java/com/google/api/codegen/discogapic/transformer/java/JavaDiscoGapicSchemaToViewTransformer.java
andreamlin: Ok, converted the raw strings into Names and transformed those entities. Will make a surface namer in a future PR.

src/main/resources/com/google/api/codegen/java/message.snip
andreamlin: Done.

andreamlin:
Addressed another round of comments. PTAL

src/main/java/com/google/api/codegen/discogapic/transformer/java/JavaDiscoGapicSchemaToViewTransformer.java
```diff
@@ -0,0 +1,209 @@
...
+    List<SimplePropertyView> properties = new LinkedList<>();
+    for (Map.Entry<String, Schema> propertyEntry : schema.properties().entrySet()) {
+      String propertyString = schemaSymbolTable.getNewSymbol(propertyEntry.getKey());
+      Schema property = propertyEntry.getValue();
+      SimplePropertyView.Builder simpleProperty =
+          SimplePropertyView.newBuilder().name(propertyString);
+      simpleProperty.typeName(typeToJavaType(property));
+
+      // TODO(andrealin) use a surface namer for the getter/setter.
+      FieldCopyView.Builder getterAndSetter = FieldCopyView.newBuilder();
```
garrettjonesgoogle: I can see that `FieldCopyView` is a good structural match for what you want, but I don't think it's a good semantic match. I think I'd prefer `SimplePropertyView` to just hold `fieldGetFunction` and `fieldSetFunction` directly.

src/main/java/com/google/api/codegen/viewmodel/SimplePropertyView.java
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * This ViewModel defines the structure of a Discovery doc's "schemas", "properties",
+ * "additionalProperties", and "items".
+ *
+ * <p>This contains a subset of properties in the JSON Schema
+ * https://tools.ietf.org/html/draft-zyp-json-schema-03#section-5.7.
+ */
+@AutoValue
+public abstract class SimplePropertyView {
```
garrettjonesgoogle: I think `SimpleMessagePropertyView` would be the best name.

src/main/resources/com/google/api/codegen/java/message.snip
garrettjonesgoogle: Oh if AutoValue does this, then leave it as is.

src/main/java/com/google/api/codegen/discogapic/transformer/java/JavaDiscoGapicSchemaToViewTransformer.java
```diff
@@ -0,0 +1,209 @@
...
+      case ARRAY:
+        return String.format("List<%s>", typeToJavaType(schema.items()));
+      case INTEGER:
+        switch (schema.format()) {
+          case INT32:
+            return "Integer";
+          case UINT32:
+            return "Long";
+          default:
+            System.err.println("Discovery doc had an INTEGER typeName that was not Integer/Long.");
```
garrettjonesgoogle: Change your println's to exceptions for now (`IllegalStateException` is fine) - we want to know early when we aren't handling cases that we are supposed to.

garrettjonesgoogle:
LGTM

src/main/java/com/google/api/codegen/discogapic/transformer/java/JavaDiscoGapicSchemaToViewTransformer.java
andreamlin: Done.

src/main/java/com/google/api/codegen/discogapic/transformer/java/JavaDiscoGapicSchemaToViewTransformer.java
andreamlin: Done.

src/main/java/com/google/api/codegen/viewmodel/SimplePropertyView.java
andreamlin: Done.

src/main/java/com/google/api/codegen/discogapic/MainDiscoGapicProviderFactory.java
```diff
@@ -49,16 +50,25 @@
...
+        DiscoGapicProvider clientProvider =
             DiscoGapicProvider.newBuilder()
                 .setDocument(document)
                 .setProductConfig(productConfig)
                 .setSnippetSetRunner(new CommonSnippetSetRunner(new JavaRenderingUtil()))
                 .setDocumentToViewTransformer(
                     new JavaDiscoGapicSurfaceTransformer(javaPathMapper, packageConfig))
                 .build();
+        DiscoGapicProvider messageProvider =
+            DiscoGapicProvider.newBuilder()
```
vchudnov-g: To avoid repetitive code, I would consider looping over a list containing  `JavaDiscoGapicSurfaceTransformer` and `JavaDiscoGapicSchemaToViewTransformer`, since all the other lines are the same.

When you add more languages, each language-specific transformer list can itself be a value in a map keyed by the possible `id` values. Then you would have something like the following (in pseudocode)
```
transformers, prefix = PerLanguageTransformers[id]
if transformers == nil {
   throw ...
}
GapicCodePathMapper mapper = ....newBuilder().....setPrefix(prefix)....build()
for transform in transformers {
  DiscoGapicProvider provider = ....NewBuilder().....setDocumentToViewTransformer(new transform(mapper, packageConfig)
  ...
  providers.add(provider)
}
```

src/main/java/com/google/api/codegen/viewmodel/SimpleMessagePropertyView.java
```diff
@@ -0,0 +1,60 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen.viewmodel;
+
+import com.google.auto.value.AutoValue;
+import javax.annotation.Nullable;
+
+/**
+ * This ViewModel defines the structure of a Discovery doc's "schemas", "properties",
```
vchudnov-g: The class below does not seem to implement `ViewModel`.

src/main/java/com/google/api/codegen/viewmodel/StaticLangApiMessageView.java
```diff
@@ -0,0 +1,93 @@
...
+package com.google.api.codegen.viewmodel;
+
+import com.google.api.codegen.SnippetSetRunner;
+import com.google.api.codegen.discovery.Schema;
+import com.google.auto.value.AutoValue;
+import java.util.List;
+import javax.annotation.Nullable;
+
+/**
+ * This ViewModel defines the view model structure of a generic message.
```
vchudnov-g: How does this relate to `SimpleMessagePropertyView`?

src/main/java/com/google/api/codegen/discogapic/MainDiscoGapicProviderFactory.java
```diff
@@ -49,16 +50,25 @@
...
+                .setDocument(document)
+                .setProductConfig(productConfig)
+                .setSnippetSetRunner(new CommonSnippetSetRunner(new JavaRenderingUtil()))
+                .setDocumentToViewTransformer(
+                    new JavaDiscoGapicSchemaToViewTransformer(javaPathMapper, packageConfig))
```
andreamlin: JavaDiscoGapicSchemaToViewTransformer creates the ViewModel for Discovery doc schemas, e.g. a compute Disk; each schema file just has getters and setters, nothing fancy.
JavaDiscoGapicSurfaceTransformer creates the ViewModel for the API client file, which is what the user interacts with to make the REST calls; it could have methods like ListDisks().

src/main/java/com/google/api/codegen/viewmodel/SimpleMessagePropertyView.java
andreamlin: Done.

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -159,7 +159,15 @@ public static GapicProductConfig create(Document document, ConfigProto configPro
...
     // TODO actually load data
-    return createDummyInstance(interfaceConfigMap, "", "", null);
+
+    LanguageSettingsProto settings =
+        configProto.getLanguageSettings().get(configProto.getLanguage());
+    if (settings == null) {
+      settings = LanguageSettingsProto.getDefaultInstance();
+    }
+
+    // TODO (end copy from above method)
```
andreamlin: Done.

src/main/java/com/google/api/codegen/viewmodel/StaticLangApiMessageView.java
andreamlin: A StaticLangApiMessageView may contain a list of SimpleMessagePropertyViews; e.g. a message Disk may contain properties creationTimestamp, name, and diskEncryptionKey.

src/main/java/com/google/api/codegen/discogapic/MainDiscoGapicProviderFactory.java
andreamlin: Done.
I'll keep that in mind when other languages are added.

andreamlin:
Addressed the open comments. PTAL

src/main/java/com/google/api/codegen/viewmodel/StaticLangApiMessageFileView.java
```diff
@@ -0,0 +1,59 @@
...
+package com.google.api.codegen.viewmodel;
+
+import com.google.api.codegen.SnippetSetRunner;
+import com.google.auto.value.AutoValue;
+
+/**
+ * ViewModel representing the file containing a message.
+ *
+ * <p>For example, this can be used to represent a class file containing a Discovery Document's
+ * schemas".
```
vchudnov-g: nit: stray quote

src/main/java/com/google/api/codegen/viewmodel/SimpleMessagePropertyView.java
```diff
@@ -25,7 +26,7 @@
  * https://tools.ietf.org/html/draft-zyp-json-schema-03#section-5.7.
  */
 @AutoValue
-public abstract class SimpleMessagePropertyView {
+public abstract class SimpleMessagePropertyView implements ViewModel {
```
garrettjonesgoogle: Only top-level view model classes need to implement `ViewModel`.

andreamlin: That makes sense, given what's in the ViewModel interface. I merged the branch already but I'll take that out in the next PR.

PR #1306, opened 2017-05-28
PHP streaming retry and exception tests

michaelbausor:
This PR:
- Removes retry settings from PHP streaming methods. Retrying does not make sense in that context, and GAX prevents the use of retry settings for streaming methods in PHP
- Changes from INTERNAL errors to DATA_LOSS errors, because some methods are configured to retry on INTERNAL errors(!)
- Call getReceivedCalls in exception tests so that the stub will be exhausted

### DIFF
--- a/src/main/java/com/google/api/codegen/clientconfig/php/PhpClientConfigGapicContext.java
+++ b/src/main/java/com/google/api/codegen/clientconfig/php/PhpClientConfigGapicContext.java
@@ -1,0 +1,33 @@
+/* Copyright 2016 Google Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen.clientconfig.php;
+
+import com.google.api.codegen.clientconfig.ClientConfigGapicContext;
+import com.google.api.codegen.config.GapicProductConfig;
+import com.google.api.tools.framework.model.Method;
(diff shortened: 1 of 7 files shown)


src/main/resources/com/google/api/codegen/clientconfig/json.snip
```diff
@@ -88,12 +88,12 @@
                 retryCodesName = methodConfig.getRetryCodesConfigName, \
                 retryParamsName = methodConfig.getRetrySettingsConfigName
             "{@method.getSimpleName}": {
-              @if or(and(retryCodesName, retryParamsName), isBatching)
+              @if or(and(and(retryCodesName, retryParamsName), context.isRetryingSupported(method)), isBatching)
```
garrettjonesgoogle: Ack I didn't realize that clientconfig generation wasn't on MVVM. Since this is totally outside the scope of this PR, can you make sure an issue is filed if one isn't yet?

src/main/java/com/google/api/codegen/clientconfig/php/PhpClientConfigGapicContext.java
```diff
@@ -0,0 +1,33 @@
...
+public class PhpClientConfigGapicContext extends ClientConfigGapicContext {
+
+  /** Constructs the client config codegen context. */
+  public PhpClientConfigGapicContext(Model model, GapicProductConfig config) {
+    super(model, config);
+  }
+
+  @Override
+  public boolean isRetryingSupported(Method method) {
+    return isSupported(method) && !method.getRequestStreaming() && !method.getResponseStreaming();
```
garrettjonesgoogle: This seems applicable to all languages - maybe get all languages to sign off?

src/main/resources/com/google/api/codegen/clientconfig/json.snip
michaelbausor: https://github.com/googleapis/toolkit/issues/1307

src/main/java/com/google/api/codegen/clientconfig/php/PhpClientConfigGapicContext.java
michaelbausor: I don't think that it is. Other languages use retries here, to retry stream creation. But that doesn't make sense in PHP because we do not receive a status immediately.

garrettjonesgoogle:
LGTM

PR #1315, opened 2017-06-01
Use absolute path in build.gradle

shinfan:
Relative paths do not work well in the new api-client-staging structure since it depends on the directory of execution. Switching to use absolute path with `projectDir` solves this problem.

I will update the api-client-staging PR once the CI is fixed


garrettjonesgoogle:
LGTM

PR #1322, opened 2017-06-03
Nodeify Discovery objects

andreamlin:
Document, Method, and Schema are all created from DiscoveryNodes, which should be part of a JSON tree structure. This PR unites Document, Method, and Schema under one Node interface, which exposes a method to get the parent of a node. 

The ability to get the parent of a node is useful in the context it provides when operating on a single Node; ex. when transforming a Method, it would be easy to fetch the name "pubsub" from the parent Document node.


src/main/java/com/google/api/codegen/discovery/Document.java
```diff
@@ -133,6 +147,22 @@ public Schema dereferenceSchema(Schema schema) {
...
 
   /** @return the auth instructions URL. */
+  @JsonIgnore @Nullable private Node parent;
+
+  public Node parent() {
+    return parent;
+  }
+
+  // Package private for use by other Node objects.
+  void setParent(Node parent) {
```
saicheems: Why have a setter? If the issue was null nodes, you can define it as:

```
@JsonProperty("parent")
@Nullable
public abstract Node parent();
```

Same suggestion for below.

src/main/java/com/google/api/codegen/discovery/Document.java
```diff
@@ -66,23 +68,35 @@ public static Document from(DiscoveryNode root) {
     String servicePath = root.getString("servicePath");
     String title = root.getString("title");
     String version = root.getString("version");
+    String id = root.getString("id");
```
saicheems: Small nit, but I tried to order the assignment of fields roughly in the order that their functions are declared in the `Document`, so this would be assigned right after description.

src/main/java/com/google/api/codegen/discovery/Node.java
```diff
@@ -0,0 +1,27 @@
...
+import javax.annotation.Nullable;
+
+/** Represents a node in a tree of nodes. */
+public interface Node {
+  /** @return the immediate parent of this node. */
+  @Nullable
+  Node parent();
+
+  /** @return the ID of this node. */
+  String id();
```
saicheems: Would sort (id above parent) - sorry to be nitpicky

src/main/java/com/google/api/codegen/discovery/Schema.java
```diff
@@ -77,22 +89,35 @@ public static Schema from(DiscoveryNode root, String path) {
...
+        new AutoValue_Schema(
+            additionalProperties,
+            defaultValue,
+            description,
+            format,
+            id,
+            isEnum,
+            items,
+            location,
+            path,
```
saicheems: The addition of `parent` mostly eliminates the need for `path`. I would go ahead and remove it. If you're willing, removing all the boilerplate for path construction can happen in this PR.

saicheems:
Nice work! This is a very useful interface.

src/main/java/com/google/api/codegen/discovery/Document.java
```diff
@@ -133,6 +147,22 @@ public Schema dereferenceSchema(Schema schema) {
...
+  }
+
+  // Package private for use by other Node objects.
+  void setParent(Node parent) {
+    this.parent = parent;
+  }
+
+  /** @return whether or not to version the module. */
+  @JsonProperty("id")
+  public abstract String id();
```
saicheems: Would sort this and the other functions alphabetically to match the current style.

src/main/java/com/google/api/codegen/discovery/Method.java
```diff
@@ -38,6 +38,17 @@
    * @return a method.
    */
   public static Method from(DiscoveryNode root, String path) {
```
saicheems: Is this method needed anymore?

saicheems: May be able to just stick to constructors w/ the signature
```
public static Document from(DiscoveryNode root);
public static Method from(DiscoveryNode root, Parent parent);
public static Schema from(DiscoveryNode root, Parent parent);
```

src/main/java/com/google/api/codegen/discovery/Document.java
saicheems: Oh, nevermind, I see how that could be a pain. Please disregard.

src/main/java/com/google/api/codegen/discovery/Document.java
```diff
@@ -133,6 +147,22 @@ public Schema dereferenceSchema(Schema schema) {
   }
 
   /** @return the auth instructions URL. */
+  @JsonIgnore @Nullable private Node parent;
```
saicheems: Change the javadoc

src/main/java/com/google/api/codegen/discovery/Schema.java
andreamlin: Done.

src/main/java/com/google/api/codegen/discovery/Node.java
andreamlin: Done.

src/main/java/com/google/api/codegen/discovery/Document.java
andreamlin: Done.

src/main/java/com/google/api/codegen/discovery/Method.java
andreamlin: Done.

src/main/java/com/google/api/codegen/discovery/Document.java
andreamlin: Done.

src/main/java/com/google/api/codegen/discovery/Document.java
andreamlin: Done.

andreamlin:
Addressed comments. PTAL

saicheems:
lgtm

PR #1327, opened 2017-06-06
Baseline Test for Java Compute Discovery

andreamlin:
Add baseline test for DiscoGapic generated code for Compute API.

The baseline file is the concatenated output of the runDiscogapicGen gradle task.


andreamlin:
Tests pass. PTAL

garrettjonesgoogle:
Thanks for turning this around so quick!

The baseline file is unfortunately too large to manage. What we have done for testing of the proto3 codepath is that we have defined a fake service which exercises the functionality of codegen, but has no more than that. I would recommend coming up with a fake discovery doc service like that; maybe you could take compute and strip it down, or use another discovery doc, or type one up manually (any strategy works). 

Here is the fake proto3 service: https://github.com/googleapis/toolkit/blob/master/src/test/java/com/google/api/codegen/testdata/library.proto

You don't have to make yours as thorough quite yet since you're just bootstrapping.

andreamlin:
Replaced large compute baseline with smaller cloudbilling API baseline.

src/main/java/com/google/api/codegen/discogapic/transformer/java/JavaDiscoGapicSchemaToViewTransformer.java
```diff
@@ -151,6 +153,15 @@ private StaticLangApiMessageView generateSchemaClass(
           nameFormatter.publicMethodName(Name.from("set").join(propertyName)));
       properties.add(simpleProperty.build());
     }
+    // TODO(andrealin): implement sorting in a conventional way.
```
garrettjonesgoogle: How is this unconventional?

src/test/java/com/google/api/codegen/DiscoGapicTestBase.java
```diff
@@ -0,0 +1,139 @@
...
+    config =
+        CodegenTestUtil.readConfig(
+            new SimpleDiagCollector(), getTestDataLocator(), gapicConfigFileNames);
+    if (config == null) {
+      return;
+    }
+  }
+
+  @Override
+  protected Object run() {
```
garrettjonesgoogle: If you return `Map<String, Doc>` like `GapicTestBase` ( https://github.com/andreamlin/toolkit/blob/112c8e4faeaa009b924fa4ca4be80a5843f9b5c8/src/test/java/com/google/api/codegen/GapicTestBase.java#L170 ), the baseline will have a separator line for each file that is generated (see https://github.com/googleapis/toolkit/blob/master/src/test/java/com/google/api/codegen/testdata/java_main_library.baseline for example), which makes the output easier to read.

src/main/java/com/google/api/codegen/discogapic/transformer/java/JavaDiscoGapicSchemaToViewTransformer.java
andreamlin: :D 
Comment removed.

src/test/java/com/google/api/codegen/DiscoGapicTestBase.java
andreamlin: Done.

andreamlin:
Fishing for comments! PTAL

garrettjonesgoogle:
So unfortunately, the diff is showing all of the merged commits, which makes it hard to find your commits to review. I think I'll have to help you clean up the commit history so it can be reviewed properly.

vchudnov-g:
LMK when the commit history has been cleaned and I'll review these changes.

garrettjonesgoogle:
LGTM

andreamlin:
Commit history cleaned up.

src/test/java/com/google/api/codegen/DiscoGapicTestBase.java
```diff
@@ -0,0 +1,147 @@
...
+              null,
+              new LinkedList<String>());
+    } catch (IOException e) {
+      throw new IllegalArgumentException("Problem creating DiscoGapic generator.");
+    }
+
+    config =
+        CodegenTestUtil.readConfig(
+            new SimpleDiagCollector(), getTestDataLocator(), gapicConfigFileNames);
+    if (config == null) {
```
vchudnov-g: why do we need this `if` statement? You're returning anyway. Is there something  to fill in?

src/test/java/com/google/api/codegen/JavaDiscoGapicGeneratorTest.java
```diff
@@ -0,0 +1,70 @@
+/* Copyright 2016 Google Inc
```
vchudnov-g: fix date

src/test/java/com/google/api/codegen/DiscoGapicTestBase.java
```diff
@@ -0,0 +1,147 @@
+/* Copyright 2016 Google Inc
```
vchudnov-g: fix date

src/test/java/com/google/api/codegen/DiscoGapicTestBase.java
andreamlin: Done. 
Adding throwing exception if config is null.

src/test/java/com/google/api/codegen/DiscoGapicTestBase.java
andreamlin: Done.

src/test/java/com/google/api/codegen/JavaDiscoGapicGeneratorTest.java
andreamlin: Done.

PR #1330, opened 2017-06-07
Add auth instructions URL to JS samples

saicheems:


saicheems:
Although JS samples have full auth boilerplate, it can still be useful to show a link if the API owner provides one.

Feel free to suggest better wording.

src/main/resources/com/google/api/codegen/js/sample.snip
```diff
@@ -10,6 +10,10 @@
          https://console.developers.google.com/apis/api/{@class.apiName}
       2. Get access keys for your application. See
          https://developers.google.com/api-client-library/javascript/start/start-js#get-access-keys-for-your-application
+      @if class.auth.instructionsUrl
+        3. For additional information on making an authenticated call, see
```
tcoffee-google: What do you think about s/making an authenticated call/authentication/? To me, the phrasing now slightly suggests that an authenticated call is an optional alternative to a more conventional sort of call.

saicheems: One reason I wanted to be more specific wrt to saying "call" was that the instructions might include more information than just the auth stuff. For example, it's a bit of a pain to run a JS sample because it has to come from `localhost`, so a local server has to be run, and the OAuth setup has to point to that local server...

See the JS instructions for `sheets:v4`: https://developers.google.com/sheets/api/quickstart/js#step_2_set_up_the_sample

I'm not sure if replacing with "authentication" would convey that the site includes more information.

vchudnov-g: On the wording, I tend to lean more towards @tcoffee-google's suggestion. To your objection, @saicheems, we're not saying that auth is the only thing on the page. What if we word it as "For more information on using auth with JavaScript..."? But any of these three wordings are OK by me (but if we keep the original, s/an authenticated call/authenticated calls/)

But a more fundamental question: what useful information is the link providing that is not in the sign-in part of the sample below? AFAICT, the only difference is that the link has slightly fancier HTML. I'm not convinced that adding the link clarifies anything for the user.

saicheems: I think the most useful information was to point out that the sample should be run on a server - it won't run if just put into `index.html` and loaded on a browser. Also, the general instructions on how to set up the OAuth callback seemed useful.

After thinking about it a bit more, I'm satisfied with changing the wording to 'authentication'. Done.

vchudnov-g:
lgtm

PR #1332, opened 2017-06-07
Ruby: Use native hash to protobuf coercion

landrito:
Updated the main gapic to use it as well as the samples and tests to verify that it works. I hand tested the generated unit tests for spanner_admin_database and monitoring and all tests pass.

### DIFF
--- a/src/test/java/com/google/api/codegen/testdata/ruby_doc_main_library.baseline
+++ b/src/test/java/com/google/api/codegen/testdata/ruby_doc_main_library.baseline
@@ -622,7 +622,7 @@ module Library
       #   can also be provided.
       # @param books [Array<Google::Example::Library::V1::Book | Hash>]
       #   The books to publish in the series.
-      #   A hash of the same form as `Array`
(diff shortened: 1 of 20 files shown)


src/test/java/com/google/api/codegen/testdata/ruby_doc_main_library.baseline
```diff
@@ -431,15 +431,16 @@ module Library
       #   require "library/v1"
       #
       #   library_service_client = Library::V1::LibraryServiceClient.new
-      #   shelf = Google::Example::Library::V1::Shelf.new
+      #   shelf = {}
```
geigerj: I think it might be more helpful to users if we retain the full type name when it's empty. WDYT?

landrito: Hm. I think empty hash is fine so long as we update the param docs. That being said, I totally forgot to update the param docs. I'll have that done soon!

PR #1333, opened 2017-06-07
Split out auth variables and scope declaration

saicheems:
Address JS sample feedback from the `sheets:v4` team.

### DIFF
--- a/src/main/resources/com/google/api/codegen/js/sample.snip
+++ b/src/main/resources/com/google/api/codegen/js/sample.snip
@@ -114,16 +114,37 @@
 
       @end
       function {@class.initClientFuncName}() {
+        @if class.auth.type != "NONE"
+          var API_KEY = '';  // TODO: Update placeholder with desired API key.
+
+        @end
+        @switch class.auth.type
+        @case "NONE"
+        @case "API_KEY"
+        @default
+          var CLIENT_ID = '';  // TODO: Update placeholder with desired client ID.
+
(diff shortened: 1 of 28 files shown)


saicheems:
Feedback was that including all scopes allowed by a method was too broad. Changed the sample to match roughly what the other languages have done wrt scopes.

src/main/resources/com/google/api/codegen/js/sample.snip
```diff
@@ -114,16 +114,37 @@
 
       @end
       function {@class.initClientFuncName}() {
+        @if class.auth.type != "NONE"
+          var API_KEY = '';  // TODO: Update placeholder with desired API key.
```
vchudnov-g: I'm confused, or maybe I just forgot what we talked about way back: If auth is not NONE, then we necessarily have an API  key? What about 3LO, etc?

saicheems: It was a recommendation from Bochun, who said that API key should included in the JS samples for both 3LO and API key based auth.

tcoffee-google: Did he say why?

saicheems: Sent an email with the link

tcoffee-google: Okay --- I still don't understand, but it seems to have been thoroughly reviewed.

PR #1334, opened 2017-06-07
Add link refs to docs for Ruby

### LINKED ISSUE #1312: Ruby types in documentation should be hyperlinked
### What

Documented types are not currently hyperlinked to the type documentation; we should do this.

### Context

https://github.com/GoogleCloudPlatform/google-cloud-ruby/issues/1473

eoogbe:
Fixes #1312


src/test/java/com/google/api/codegen/testdata/ruby_doc_main_library.baseline
```diff
@@ -39,10 +39,10 @@ module Library
     # resources and Book resources in the library. It defines the following
     # resource model:
     #
-    # * The API has a collection of Shelf
+    # * The API has a collection of {Shelf}[rdoc-ref:google.example.library.v1.Shelf]
```
geigerj: Is this supposed to be `.`-separated? I would expect `::`-separated.

eoogbe: Done.

eoogbe:
Can't generate until googleapis/artman#215 is fixed, but will try it out when I can.

eoogbe:
Works on the logging api. @geigerj @landrito PTAL

geigerj:
What happens if the proto link is to a type that is outside the package, e.g., to a type from `googleapis-common-protos`?

eoogbe:
It will treat all the links the same.

geigerj:
@eoogbe No, I mean, is YARD be able to dereference the type name correctly if the type lives outside of the package? I would expect this is only possible with some configuration. I do not want to introduce broken links into our docs.

eoogbe:
YARD will remove links it can't resolve. The only problem is that it doesn't seem to understand URL links and uses the link address instead of the text as a replacement. This seems to be a bug with YARD. I think I'm following the [guide](http://www.rubydoc.info/gems/yard/file/docs/GettingStarted.md#Linking_Objects______).

landrito:
Don't think I can get around to reviewing this this week.

eoogbe:
@geigerj What do you mean by follow up?

geigerj:
@eoogbe Figure out what's causing the issue, and if there's a workaround on our end. If it's a bug in YARD, it would be great if you could file and track an issue with them.

eoogbe:
Found the bug in gcloud-jsondoc and filed an issue, GoogleCloudPlatform/google-cloud-ruby#1584.

PR #1338, opened 2017-06-09
add Java and Go benchmarks

pongad:


pongad:
@jskeet PTAL

jskeet:
I'm at a conference at the moment, which makes things tricky - I'll see what I can do between slots.

PR #1341, opened 2017-06-09
NodeJS: Fix samples to reflect partial veneer entrypoint.

landrito:
## Context:
The version_index for NodeJS was updated to use the partial veneer style in #1267. This PR is to make it such that the unit tests and samples use that entry point.


landrito:
Bah. Looks like when a proto object is referenced in the samples (AFIK only when an enum is used) the export of require("src/vN") is used to access it. This complicates things because when trying to use the partial veneer entry point. The user will not even have access to that export.

I think a solution would be to extend the export of "src/index.js" to include the protos of the service."

landrito:
Maybe it would be also good to make it accessible at the client level too.

```js
var library = require("@google/library");
var request = new library.GetBookRequest({});

var libraryClient = require("@google/library")({
  //auth stuff
});
request = new libraryClient.GetBookRequest({});

**Comment by landrito** [comment]
@swcloud thoughts?

**Comment by swcloud** [comment]
Please add the test case to cover multiple services.

Also add the description for the problems this PR tries to solve.

**Comment by landrito** [comment]
Added the test multiple services test! I'll add the description in the morning after some sleep. 💤  Ran the unit tests for monitoring with these changes and they all passed. This verifies that the version index and types export is working!

File: src/test/java/com/google/api/codegen/testdata/multiple_services.yaml [hunk 1]
```diff
@@ -0,0 +1,14 @@
...
+
+# Included protobuf APIs
+apis:
+- name: google.cloud.example.v1.IncrementerService
+- name: google.cloud.example.v1.DecrementerService
+
+# Documentation section
+documentation:
+  summary:
+    Tests no path templates
```
swcloud: Add end of file line.

src/test/java/com/google/api/codegen/testdata/multiple_services_gapic.yaml
```diff
@@ -0,0 +1,33 @@
...
+    retry_params_name: default
+    timeout_millis: 10000
+- name: google.cloud.example.v1.DecrementerService
+  required_constructor_params:
+  - service_address
+  - scopes
+  methods:
+  - name: Decrement
+    retry_params_name: default
+    timeout_millis: 10000
```
swcloud: Add end of file line.

src/main/resources/com/google/api/codegen/nodejs/version_index.snip
```diff
@@ -77,12 +99,19 @@
     module.exports.{@service.serviceName} = {@service.localName};
   @end
 
+  module.exports.types = {@version}Protos;
+
+  module.exports.{@version} = {};
```
swcloud: The line is unnecessary I think

src/main/resources/com/google/api/codegen/nodejs/version_index.snip
```diff
@@ -64,6 +67,25 @@
    }
```
swcloud: Where does 'Create the image annotator' come from? :)

src/main/resources/com/google/api/codegen/nodejs/version_index.snip
```diff
@@ -77,12 +99,19 @@
...
   @join service : services
     module.exports.{@version}.{@service.serviceName} = {@service.localName};
   @end
+  module.exports.{@version}.types = {@version}Protos;
 @end
 
 @private singleServiceExport(service, version)
   module.exports = {@service.localName};
+  module.exports.types = {@version}Protos;
+  module.exports.{@version} = {};
```
swcloud: This line is unnecessary I think

src/main/resources/com/google/api/codegen/nodejs/version_index.snip
landrito: Fixed in #1346! Merged the change!

src/test/java/com/google/api/codegen/testdata/multiple_services.proto
```diff
@@ -0,0 +1,36 @@
...
+  // Increment.
+  rpc Increment(IncrementRequest) returns (google.protobuf.Empty) {
+    option (google.api.http) = { post: "/v1/me:increment" body: "*" };
+  }
+
+}
+
+service DecrementerService {
+
+    // Ban.
```
landrito: Done.

src/main/resources/com/google/api/codegen/nodejs/version_index.snip
landrito: Ah actually it's only necessary for multiple services since `vX` is undefined but we need to assign to `vX.serviceShortname`.

Fixed!

src/main/resources/com/google/api/codegen/nodejs/version_index.snip
landrito: See above.

src/test/java/com/google/api/codegen/testdata/multiple_services.yaml
landrito: Done.

src/test/java/com/google/api/codegen/testdata/multiple_services_gapic.yaml
landrito: Done.

landrito:
@swcloud Thanks for the review! I again checked this PR's generated tests with monitoring and everything passed.

PR #1345, opened 2017-06-09
Java: updating for gax multiple transport refactoring

garrettjonesgoogle:


src/main/java/com/google/api/codegen/config/GapicInterfaceConfig.java
```diff
@@ -86,6 +86,11 @@ public String getName() {
     return hasInterfaceNameOverride() ? getInterfaceNameOverride() : getInterface().getSimpleName();
   }
 
+  @Override
+  public String getRawName() {
+    return getInterface().getSimpleName();
+  }
+
```
pongad: This is only used in one place. Can we just make the calling site do `getInterface().getSimpleName()`?

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTransformer.java
```diff
@@ -402,15 +419,142 @@ private StaticLangSettingsView generateSettingsClass(
...
+    StaticLangFileView.Builder<StaticLangStubInterfaceView> fileView =
+        StaticLangFileView.<StaticLangStubInterfaceView>newBuilder();
+
+    fileView.classView(generateStubInterface(context));
+    fileView.templateFileName(STUB_INTERFACE_TEMPLATE_FILENAME);
+
+    String outputPath =
+        pathMapper.getOutputPath(context.getInterface(), context.getProductConfig());
+    String className = context.getNamer().getApiStubInterfaceName(context.getInterfaceConfig());
+    fileView.outputPath(outputPath + "/stub/" + className + ".java");
```
pongad: I think we should prefer `File.separator` to hardcoding '/'

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTransformer.java
```diff
@@ -402,15 +419,142 @@ private StaticLangSettingsView generateSettingsClass(
...
+    StaticLangFileView.Builder<StaticLangGrpcStubView> fileView =
+        StaticLangFileView.<StaticLangGrpcStubView>newBuilder();
+
+    fileView.classView(generateGrpcStubClass(context));
+    fileView.templateFileName(GRPC_STUB_TEMPLATE_FILENAME);
+
+    String outputPath =
+        pathMapper.getOutputPath(context.getInterface(), context.getProductConfig());
+    String className = context.getNamer().getApiGrpcStubClassName(context.getInterfaceConfig());
+    fileView.outputPath(outputPath + "/stub/" + className + ".java");
```
pongad: ditto

src/main/java/com/google/api/codegen/viewmodel/StaticLangApiView.java
```diff
@@ -61,6 +61,9 @@
 
   public abstract boolean hasLongRunningOperations();
 
+  @Nullable // Used in Java
+  public abstract String stubInterfaceName();
```
pongad: Nullable feels a little icky. I generally think "NOT IMPLEMENTED" string is better. Looks like you'll have to change all LanguageTransformers though... Up to you.

src/main/java/com/google/api/codegen/viewmodel/StaticLangSettingsView.java
```diff
@@ -76,6 +76,12 @@
 
   public abstract boolean hasDefaultInstance();
 
+  @Nullable // Used in Java
```
pongad: Ditto nullable comment above

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTransformer.java
```diff
@@ -402,15 +419,142 @@ private StaticLangSettingsView generateSettingsClass(
...
+    stubClass.hasLongRunningOperations(interfaceConfig.hasLongRunningOperations());
+
+    return stubClass.build();
+  }
+
+  private String getAndSaveNicknameForRootType(GapicInterfaceContext context, String nickname) {
+    SurfaceNamer namer = context.getNamer();
+    ModelTypeTable typeTable = context.getModelTypeTable();
+
+    String fqTypeName = namer.getRootPackageName() + "." + nickname;
```
michaelbausor: fq?

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTransformer.java
```diff
@@ -122,23 +128,26 @@ public JavaGapicSurfaceTransformer(
               JavaFeatureConfig.newBuilder()
                   .enableStringFormatFunctions(enableStringFormatFunctions)
                   .build());
-      StaticLangApiFileView apiFile = generateApiFile(context);
+      StaticLangFileView<StaticLangApiView> apiFile = generateApiFile(context);
```
michaelbausor: Adding a generic seems like a break with out existing patterns. Is this necessary? Maybe you could call `generateApiClass(context)` here directly if it is necessary to have access to the `classView()` field?

src/test/java/com/google/api/codegen/testdata/java_grpc_stub_library.baseline
```diff
@@ -0,0 +1,548 @@
...
+    return new GrpcLibraryServiceStub(settings, ClientContext.create(settings));
+  }
+
+  public static final GrpcLibraryServiceStub create(ClientContext clientContext) throws IOException {
+    return new GrpcLibraryServiceStub(LibrarySettings.defaultBuilder().build(), clientContext);
+  }
+
+  /**
+   * Constructs an instance of GrpcLibraryServiceStub, using the given settings.
+   * This is protected so that it easy to make a subclass, but otherwise, the static
```
michaelbausor: s/it easy/it is easy

Is it an expected use case to subclass the Grpc stub?

src/test/java/com/google/api/codegen/testdata/java_stub_interface_library.baseline
```diff
@@ -0,0 +1,227 @@
...
+import com.google.tagger.v1.AddLabelResponse;
+import com.google.tagger.v1.AddTagRequest;
+import com.google.tagger.v1.AddTagResponse;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Generated;
+
+// AUTO-GENERATED DOCUMENTATION AND CLASS
+@Generated("by GAPIC v0.0.5")
+public abstract class LibraryServiceStub implements BackgroundResource {
```
michaelbausor: The reason this is an abstract class and not an interface is so that we can add new methods without breaking existing stubs, right? Or is there another reason?

We could use this to provide a better exception message for not implemented methods by having an abstract method that returns the name on the transport. Then the exception message could be something like "<Method> is not supported by transport <Transport>".

src/main/java/com/google/api/codegen/config/GapicInterfaceConfig.java
garrettjonesgoogle: It's used by two functions in `SurfaceNamer`.

I think it's a useful abstraction - there is also a `getName()` function. This is an override of the method in `InterfaceConfig` so that DiscoGapicConfig can have its own implementation.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTransformer.java
garrettjonesgoogle: Sure, I see that this is used in some other places already.

src/main/java/com/google/api/codegen/viewmodel/StaticLangApiView.java
garrettjonesgoogle: I think it would be neat to have a custom variant of `AutoValue` that could default to a "not set" string for values that aren't set. That might be a path to yak shaving though...

src/main/java/com/google/api/codegen/transformer/GapicInterfaceContext.java
```diff
@@ -107,6 +107,16 @@ public GapicInterfaceContext withNewTypeTable() {
         getFeatureConfig());
   }
 
+  public GapicInterfaceContext withNewTypeTable(String packageName) {
+    return create(
+        getInterface(),
+        getProductConfig(),
+        // getPackageMetadataConfig(),
```
garrettjonesgoogle: This was copied dead code - I will remove the line in the original too.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTransformer.java
garrettjonesgoogle: So the thing that drove me to do this was that there were two classes, `StaticLangApiFileView` and `StaticLangSettingsFileView`, which were identical except for the type of the view for the class they contained, and I needed to add two more types of files, 1) stub interface and 2) grpc stub implementation. I think it would harm code navigation a lot to just put a generic type like `Object` or `ViewModel` as the type of `classView()`, even if I didn't need to call methods on `classView`. That's the reason everything else is explicit types too; the Snippets language doesn't care, and everything could be an `Object` and the Snippets rendering would work fine.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTransformer.java
garrettjonesgoogle: fully qualified... I can spell it out here.

src/test/java/com/google/api/codegen/testdata/java_grpc_stub_library.baseline
garrettjonesgoogle: Hmm, this comment migrated from the client template... Maybe people should subclass the stub interface instead, but I'm perfectly fine with them subclassing the Grpc stub for mocking.

src/test/java/com/google/api/codegen/testdata/java_stub_interface_library.baseline
garrettjonesgoogle: Indeed, I would like to have http-json stubs provide messaging along those lines for streaming methods. It probably shouldn't be done at the stub interface level here - I want to distinguish a method that is missing altogether from one that is explicitly acknowledged as not supported by the transport.

garrettjonesgoogle:
Addressed feedback, PTAL

pongad:
LGTM but others should sign off.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTransformer.java
michaelbausor: Hmm, maybe that is just my fault for not realizing that. If you think it's clear I am happy to leave it, otherwise maybe `fullTypeName` is a compromise?

michaelbausor: Nervermind, saw updated

michaelbausor:
LGTM from me also

PR #1346, opened 2017-06-09
NodeJS: Small hand-edits fixes

landrito:


src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSGapicSurfaceTestTransformer.java
```diff
@@ -126,10 +127,17 @@ private MockCombinedView generateTestView(
               .build());
     }
 
+    String outputPath = "test";
```
swcloud: Let's move this to a separate method.

If you look at https://github.com/googleapis/googleapis/blob/master/gapic/api/artman_videointelligence.yaml#L56,

The test file name should be gapic-<version>.js

landrito: Done.

PR #1350, opened 2017-06-10
Inner classes for nested schemas

andreamlin:
Recursively create nested classes for nested schemas.

Schema view models can contain lists of other schema view models.

If a JSON schema is:
```
{
    id: "Foo",
    properties: {
        "Bar": {
            "type": "Array",
            "items": {
                   "type": "Object",
                   "properties":  {
                       "Baz": {
                            "type": "String",
                       }
                   }
              }
         }
    }
}
```

Then a class will be generated as such:

```
public class Foo implements Serializable {
    private List<Bar> bar;

    public class Bar {
        private String baz;
    }
}
```


andreamlin:
PTAL

src/main/java/com/google/api/codegen/discogapic/transformer/java/JavaDiscoGapicSchemaToViewTransformer.java
```diff
@@ -105,22 +108,32 @@ public int compare(Schema o1, Schema o2) {
 
   private SchemaTypeTable createTypeTable(String implicitPackageName) {
     return new SchemaTypeTable(
-        new JavaTypeTable(implicitPackageName),
-        new JavaSchemaTypeNameConverter(implicitPackageName));
+        new JavaTypeTable(implicitPackageName, true),
```
garrettjonesgoogle: booleans can be hard to read as parameters, since you can't tell at the call site what true/false means. I would suggest using an enum.

src/main/java/com/google/api/codegen/transformer/java/JavaSchemaTypeNameConverter.java
```diff
@@ -33,6 +35,12 @@
...
 
   private TypeNameConverter typeNameConverter;
+  private JavaNameFormatter nameFormatter;
+
+  public JavaSchemaTypeNameConverter(String implicitPackageName, JavaNameFormatter nameFormatter) {
+    this.typeNameConverter = new JavaTypeTable(implicitPackageName);
+    this.nameFormatter = nameFormatter;
+  }
 
   public JavaSchemaTypeNameConverter(String implicitPackageName) {
```
garrettjonesgoogle: Are there remaining uses of this constructor?

src/main/java/com/google/api/codegen/transformer/java/JavaSchemaTypeNameConverter.java
```diff
@@ -149,8 +163,12 @@ private TypeName getTypeNameForElementType(Schema schema, boolean shouldBoxPrimi
         shortName = schema.additionalProperties().reference();
       } else {
         // TODO(andrealin): Treat nested schemas as inner classes.
-        shortName = "Object";
-        packageName = "java.lang";
+        //        shortName = "Object";
+        //        packageName = "java.lang";
```
garrettjonesgoogle: Is this TODO done in this PR?

src/main/java/com/google/api/codegen/transformer/java/JavaSchemaTypeNameConverter.java
```diff
@@ -92,32 +100,38 @@ public TypeName getTypeNameInImplicitPackage(String shortName) {
   }
 
   @Override
-  public TypeName getTypeName(Schema schema) {
+  public TypeName getTypeName(String key, Schema schema) {
```
garrettjonesgoogle: I can't quite tell what `key` means for this function when just looking at the function signature.

src/main/java/com/google/api/codegen/transformer/java/JavaSchemaTypeNameConverter.java
```diff
@@ -92,32 +100,38 @@ public TypeName getTypeNameInImplicitPackage(String shortName) {
...
     if (schema.type() == Type.ARRAY) {
       TypeName listTypeName = typeNameConverter.getTypeName("java.util.List");
-      TypeName elementTypeName = getTypeNameForElementType(schema, true);
+      TypeName elementTypeName = getTypeNameForElementType(null, schema, null, true);
```
garrettjonesgoogle: I think a different design should be considered for passing `key` around if there are so many call sites that pass `null` for it. 

Like my other comment about booleans, call sites with `null` can be harder to understand, especially when multiple nulls are passed to the same call.

src/main/resources/com/google/api/codegen/java/message.snip
```diff
@@ -6,18 +6,37 @@
...
+    {@equalsMethod(schema)}
+
+    {@hashCode(schema)}
   }
 @end
 
+@private innerSchemas(schema)
+  @join property : schema.properties
+    @if {@property.properties}
+      {@schemaClassDef(property)}
```
andreamlin: By the spec, indefinitely. A schema can contain properties, each of which is a schema.
https://tools.ietf.org/html/draft-zyp-json-schema-03#section-5.2

garrettjonesgoogle: I'm concerned about the deeply nested classes - it's not typical class design. I don't actually have a better suggestion though because of the potential problems with name clashes if nested classes aren't used.

andreamlin: Right now, the compute discovery doc's most nested class is 3-deep (Operation.Warnings.Data)

src/main/java/com/google/api/codegen/transformer/java/JavaSchemaTypeNameConverter.java
andreamlin: Done indeed.

src/main/java/com/google/api/codegen/transformer/java/JavaSchemaTypeNameConverter.java
andreamlin: Moving key to a field inside of schema to cull the nulls.

src/main/java/com/google/api/codegen/transformer/java/JavaSchemaTypeNameConverter.java
andreamlin: Gone.

src/main/java/com/google/api/codegen/transformer/java/JavaSchemaTypeNameConverter.java
andreamlin: Gone.

src/main/java/com/google/api/codegen/discogapic/transformer/java/JavaDiscoGapicSchemaToViewTransformer.java
andreamlin: Done.

src/main/java/com/google/api/codegen/util/java/JavaTypeTable.java
```diff
@@ -171,19 +180,16 @@ public static String getBoxedTypeName(String primitiveTypeName) {
...
       return yes;
     }
-    // Use reflection to determine whether the name exists in java.lang.
-    try {
-      Class.forName("java.lang." + name);
-      yes = true;
-    } catch (Exception e) {
-      yes = false;
-    }
+    yes = JavaNameFormatter.isJavaLangImport(name);
```
andreamlin: Done.

src/test/java/com/google/api/codegen/testdata/discoveries/java/java_cloudbilling.v1.json.discogapic.baseline
```diff
@@ -505,6 +514,10 @@ public final class ProjectBillingInfo implements Serializable {
     return projectId;
   }
 
+
+
+
+
```
andreamlin: Done.
Added a new discovery doc billingtest.v1.json and the corresponding baseline files.

andreamlin:
Fixed up things in response to comments. PTAL

src/main/java/com/google/api/codegen/transformer/java/JavaSchemaTypeNameConverter.java
andreamlin: Done.
Simply passing along "key" (refactored to escapedSchemaName) everywhere instead of "null".

andreamlin:
Responded to all comments. PTAL

src/test/java/com/google/api/codegen/testdata/discoveries/js/js_billingtest.v1.json.baseline
```diff
@@ -0,0 +1,361 @@
+
+<html>
```
garrettjonesgoogle: Is there a way to avoid having baselines for the languages we aren't supporting for disco gapic yet? Maybe we need a new directory for discogapic to cut down on noise.

src/test/java/com/google/api/codegen/testdata/billingtest_gapic.yaml
```diff
@@ -0,0 +1,30 @@
+type: com.google.api.codegen.ConfigProto
+language_settings:
+  java:
+    package_name: com.google.cloud.billingtest.v1
```
garrettjonesgoogle: Maybe it might be best to not end the API name in "test", because we will have tests for it later, so we could end up with "BillingtestTest" or things like that. Maybe "Fakebilling" would be the best name.

src/main/java/com/google/api/codegen/transformer/SchemaTypeFormatter.java
```diff
@@ -26,12 +26,16 @@
   /** Get the full name for the given short name, using the default package. */
   String getImplicitPackageFullNameFor(String shortName);
 
-  /** Get the full name for the given type. */
-  String getFullNameFor(Schema type);
+  /** Returns the full name for the given type (without adding the full name to the import set). */
+  String getFullNameFor(String escapedSchemaName, Schema type, String parentName);
```
garrettjonesgoogle: I want to understand `escapedSchemaName` better - let's discuss in person tomorrow. Also I have an idea for a different way to handle imports in toolkit.

src/test/java/com/google/api/codegen/testdata/discoveries/js/js_billingtest.v1.json.baseline
andreamlin: For sure! 
Removing all the other languages' baselines for the billing.json Discovery doc and moving discogapic test resources to a different directory.

src/test/java/com/google/api/codegen/testdata/billingtest_gapic.yaml
andreamlin: Done.

src/main/java/com/google/api/codegen/transformer/SchemaTypeFormatter.java
andreamlin: Removed escapedSchemaName as parameter everywhere.

andreamlin:
Decluttered a ton. PTAL

src/test/java/com/google/api/codegen/testdata/fakebilling_gapic.yaml
```diff
@@ -0,0 +1,30 @@
...
+interfaces:
+- name: google.fakebilling.v1.BillingAccounts
+  methods:
+  - name: List
+  - name: Get
+- name: google.fakebilling.v1.Projects
+  methods:
+  - name: UpdateBillingInfo
+  - name: GetBillingInfo
+  - name: BeginTransaction
```
garrettjonesgoogle: add a newline at the end

src/main/resources/com/google/api/codegen/ruby/apiary_names.yaml
```diff
@@ -1,6 +1,15 @@
 ---
 "/foo:v1/bar.baz.get": get_bar_baz
 "/foo:v1/Baz": baz
+"/billingtest:v1/billingtest.billingAccounts.get": get_billing_account
```
garrettjonesgoogle: It looks like this file can be reverted

src/main/java/com/google/api/codegen/util/java/JavaTypeTable.java
```diff
@@ -14,6 +14,8 @@
  */
 package com.google.api.codegen.util.java;
 
+import static com.google.api.codegen.util.java.JavaTypeTable.JavaLangResolution.ESCAPE_JAVA_LANG_CLASH;
```
garrettjonesgoogle: It seems like `JavaLangResolution` should actually be here, since this is the class that uses it for logic.

src/main/java/com/google/api/codegen/transformer/SchemaTypeFormatter.java
```diff
@@ -26,12 +26,16 @@
   /** Get the full name for the given short name, using the default package. */
   String getImplicitPackageFullNameFor(String shortName);
 
-  /** Get the full name for the given type. */
-  String getFullNameFor(Schema type);
+  /** Returns the full name for the given type (without adding the full name to the import set). */
+  String getFullNameFor(Schema type, String parentName);
```
garrettjonesgoogle: Is it possible to get the value of `parentName` from the schema instead of piping it through?

src/test/java/com/google/api/codegen/testdata/fakebilling_gapic.yaml
andreamlin: Done.

src/main/resources/com/google/api/codegen/ruby/apiary_names.yaml
andreamlin: Done.

src/main/java/com/google/api/codegen/util/java/JavaTypeTable.java
andreamlin: JavaLangResolution is in JavaTypeTable. The static import is there so the enum doesn't have to be qualified when it's used, even though the enum is defined in this class.

src/main/java/com/google/api/codegen/transformer/SchemaTypeFormatter.java
andreamlin: Done.

andreamlin:
Another round of comments addressed. PTAL

src/main/java/com/google/api/codegen/util/java/JavaTypeTable.java
garrettjonesgoogle: Oh! Sounds good.

PR #1355, opened 2017-06-14
Remove bad restriction for default strings

saicheems:
There is no requirement that path template patterns have an even number
of segments.


saicheems:
This adds support for cases where path templates terminate in a literal instead of a wildcard.

See:
https://github.com/googleapis/toolkit/blob/master/src/test/java/com/google/api/codegen/testdata/discoveries/genomics.v1.json#L2824

or 

https://github.com/googleapis/discovery-artifact-manager/blob/master/discoveries/dataproc.v1.json#L597

src/main/java/com/google/api/codegen/discovery/DefaultString.java
```diff
@@ -94,14 +94,22 @@ static String forPattern(String pattern, String placeholderFormat) {
     }
 
     StringBuilder ret = new StringBuilder();
```
tcoffee-google: What does this variable name mean? Maybe change to something more obvious.

src/main/java/com/google/api/codegen/discovery/DefaultString.java
```diff
@@ -94,14 +94,22 @@ static String forPattern(String pattern, String placeholderFormat) {
...
-    for (int i = 0; i < elems.size(); i += 2) {
-      String literal = elems.get(i).getLiteral();
-      String placeholder = Inflector.singularize(literal);
-      placeholder = LanguageUtil.lowerCamelToLowerUnderscore(placeholder).replace('_', '-');
-      ret.append('/')
-          .append(literal)
-          .append("/")
-          .append(String.format(placeholderFormat, placeholder));
+    for (int i = 0; i < elems.size(); i++) {
```
tcoffee-google: Given the revised specification, it seems more appropriate to keep the looping by two, and simply add an additional check at the end for an extra literal.

src/main/java/com/google/api/codegen/discovery/DefaultString.java
saicheems: Changed to `defaultString`

src/main/java/com/google/api/codegen/discovery/DefaultString.java
saicheems: Done

PR #1358, opened 2017-06-14
Add README to Ruby service module doc

eoogbe:


src/main/resources/com/google/api/codegen/ruby/version_index.snip
```diff
@@ -1,12 +1,59 @@
...
+    {@modules(index.modules.iterator)}
+  @end
 
   @join requireView : {@index.requireViews}
       require "{@requireView.fileName}"
   @end
 
 @end
+
+@private modules(iterator)
```
geigerj: Why snippet recursion instead of `join`?

src/test/java/com/google/api/codegen/testdata/ruby_version_index_library.baseline
```diff
@@ -12,4 +13,48 @@
...
+#
+# library_service_client = Library::V1::LibraryServiceClient.new
+# formatted_name = Library::V1::LibraryServiceClient.book_path("testShelf-" + Time.new.to_i.to_s, project_id)
+# rating = :GOOD
+# book = { rating: rating }
+# response = library_service_client.update_book(formatted_name, book)
+# ```
+#
+# ### Next Steps
+# - Read the [Client Library Documentation][] for Google Example Library API to see other available methods on the client.
```
geigerj: This link doesn't make sense in this context. (It will be seen from within the client library docs.) Can we show this in the actual README but not here?

src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceTransformer.java
```diff
@@ -169,6 +171,27 @@ private ViewModel generateApiClass(GapicInterfaceContext context) {
 
   private ViewModel generateVersionIndexView(Model model, GapicProductConfig productConfig) {
     SurfaceNamer namer = new RubySurfaceNamer(productConfig.getPackageName());
+
+    ImmutableList.Builder<VersionIndexModuleView> moduleViews = ImmutableList.builder();
+    List<String> apiModules = namer.getApiModules();
+    int moduleCount = apiModules.size();
+    for (int i = 0; i < moduleCount; ++i) {
+      if (i == moduleCount - 2) {
```
geigerj: The number `2` is a little magical here. (Presumably, that's 1 module up for the version, plus 1 module up for the product name?)  Some documentation of the reasoning here would be helpful.

src/main/resources/com/google/api/codegen/ruby/version_index.snip
eoogbe: It makes the indentation easier. These are nested modules rather than sequential.

eoogbe:
It parses the markdown we use correctly, though I'm not clear on whether it features a complete markdown parser.

eoogbe:
I can't generate an example for monitoring, because of googleapis/artman#215 but I did get [a screenshot using the library service](https://screenshot.googleplex.com/Oby62JkmOoY.png).

eoogbe:
@landrito LGTY?

PR #1359, opened 2017-06-15
Update test resources for discogapic

andreamlin:
This is purely a test resource change (with one hack in the src/ to make tests pass). 

Add "interface" and "Interface" to the test Discovery doc (fakebilling) to test escaping Java keywords.


garrettjonesgoogle:
Nice to see the fields trimmed down! LGTM after you fix the tests.

PR #1366, opened 2017-06-16
Fix python smoke format string

eoogbe:


geigerj:
@eoogbe If there's context in an original issue or some other PR, it would be helpful to link it here.

eoogbe:
No context, just noticed it was wrong the first time

Issue #1320, opened 2017-06-02
Python: Auto-gen Improvements


lukesneeringer:
This is a "catch-all ticket" of things I have identified that need doing in Python auto-gen.
Examples in this ticket will follow Vision (which I have hand-edited).

More important items are marked with :warning:; this ticket should be demoted to P2 once those are completed.

### *_client.py (e.g. image_annotator_client.py)

  * [x] :warning: In docstrings, references to `string` should be spelled `str`.
  * [x] :warning: In docstrings, references to `iterator` should be spelled `Iterable`.
  * [x] Docstrings should use 4-space indents.
  * [x] Docstring class references should use the `~.` prefix if a period is present in the class name.
  * [x] In docstring examples, use `...` for the prompt in hanging indents or continued statements (match the interpreter).
  * [x] :warning: @landrito The `pkg_resources.get_distribution` call needs to use the "manual" package name (e.g. `google-cloud-vision`), not the GAPIC package name. We are moving away from publishing stand-alone GAPIC packages.
  * [x] :warning: @landrito Use the new Future class from @jonparrott.
      * This has its own issue: #1426. See also GoogleCloudPlatform/google-cloud-python#3764.

### *_client.json (e.g. image_annotator_client.json)

  * [x] Rename to `image_annotator_config.py`, and write the value to a variable in the top line. Alter the corresponding GAPIC Python file to import this rather than using `pkg_resources.resource_string` to load the JSON file.

### proto package

  * [x] :warning: @lukesneeringer Add `protoc-docs-plugin` to artman. (googleapis/artman#217)

### Partial Veneer

  * [x] :warning: @landrito Add a versioned directory in the `google.cloud` namespace (e.g. `google.cloud.vision_v1`)
  * [x] :warning: @landrito Add a `types.py` file to the `google.cloud.api_vN` package which iterates over all the `_pb2.py` files in the proto and makes the messages available in a consistent namespace (see Vision for a reference implementation).
  * [x] :warning: @landrito Add an `__init__.py` file to the `google.cloud.api_vN` package which imports (1) the client and (2) the enums module from the GAPIC, and (3) the `types` module (above bullet). These should be exported wholesale in `__all__`. See Vision for a reference implementation.
[...]

PR #1368, opened 2017-06-16
Python: Introduced versioned api directory exporting types.

landrito:
Updates #1320

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
+++ b/src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
@@ -159,7 +159,7 @@ public class SurfaceNamer extends NameFormatterDelegator {
     return qualifiedName(namePath.withoutHead());
   }
 
-  /** The namespace of the package. */
+  /** The unversioned namespace of the package. */
   public String getPackageNamespace(String version) {
     NamePath namePath = typeNameConverter.getNamePath(packageName);
     if (namePath.getHead().equals(version)) {
--- a/src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
+++ b/src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
@@ -159,6 +159,20 @@ public class SurfaceNamer extends NameFormatterDelegator {
     return qualifiedName(namePath.withoutHead());
   }
 
+  /** The namespace of the package. */
+  public String getPackageNamespace(String version) {
+    NamePath namePath = typeNameConverter.getNamePath(packageName);
+    if (namePath.getHead().equals(version)) {
+      return qualifiedName(namePath.withoutHead());
+    }
+    return qualifiedName(namePath);
+  }
+
+  /** The versioned namespace of an api. Example: google.cloud.vision_v1 */
+  public String getVersionedDirectoryNamespace(Interface apiInterface) {
+    return getNotImplementedString("SurfaceNamer.getVersionedDirectoryNamespace");
+  }
+
   /** The modules of the package. */
   public ImmutableList<String> getApiModules() {
     return ImmutableList.<String>of();
@@ -475,6 +489,15 @@ public class SurfaceNamer extends NameFormatterDelegator {
     return getNotImplementedString("SurfaceNamer.getByteLengthFunctionName");
   }
 
+  /** The name of the class that ties in the helper functions to a client partial veneers. */
+  public String getHelpersClassName(String apiShortName) {
+    return publicClassName(Name.anyCamel(apiShortName).join("helpers"));
+  }
+
+  /** The name of a stub type of an interface */
+  public String getStubType(Interface apiInterface) {
+    return publicClassName(Name.upperCamel(apiInterface.getSimpleName(), "Stub"));
+  }
   /////////////////////////////////////// Variable names //////////////////////////////////////////
 
   /**
--- a/src/main/java/com/google/api/codegen/transformer/py/PythonImportSectionTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/py/PythonImportSectionTransformer.java
@@ -37,11 +37,8 @@ import com.google.api.tools.framework.model.TypeRef;
 import com.google.common.collect.ImmutableList;
 import java.util.ArrayList;
 import java.util.Comparator;
-import java.util.HashSet;
-import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
-import java.util.Queue;
 import java.util.Set;
 import java.util.TreeSet;
 
@@ -90,9 +87,12 @@ public class PythonImportSectionTransformer implements ImportSectionTransformer 
       Model model,
       GapicProductConfig productConfig,
       PackageMetadataConfig packageConfig,
-      SurfaceNamer namer) {
+      SurfaceNamer namer,
+      boolean packageHasEnums) {
     return ImportSectionView.newBuilder()
-        .appImports(generateVersionedInitAppImports(model, productConfig, packageConfig, namer))
+        .appImports(
+            generateVersionedInitAppImports(
+                model, productConfig, packageConfig, namer, packageHasEnums))
         .standardImports(generateVersionedInitStandardImports())
         .build();
   }
@@ -189,7 +189,7 @@ public class PythonImportSectionTransformer implements ImportSectionTransformer 
     Set<ImportFileView> imports = new TreeSet<>(importFileViewComparator());
 
     // Save proto file import names to the type table for disambiguation.
-    Set<ProtoFile> protoFileDependencies = getProtofileDependencies(model);
+    List<ProtoFile> protoFileDependencies = model.getFiles();
     populateTypeTable(protoFileDependencies, typeTable);
 
     // Get disambiguated imports.
@@ -213,7 +213,7 @@ public class PythonImportSectionTransformer implements ImportSectionTransformer 
     return ImmutableList.of(createImport("__future__", "absoulute_import"), createImport("sys"));
   }
 
-  private void populateTypeTable(Set<ProtoFile> protoFileDependencies, ModelTypeTable typeTable) {
+  private void populateTypeTable(List<ProtoFile> protoFileDependencies, ModelTypeTable typeTable) {
     for (ProtoFile protoFile : protoFileDependencies) {
       // For python, adding a single message from the proto file to the type table will populate
       // the type table with the correct imports.
@@ -224,28 +224,6 @@ public class PythonImportSectionTransformer implements ImportSectionTransformer 
     }
   }
 
-  private Set<ProtoFile> getProtofileDependencies(Model model) {
-    // Set up BFS.
-    Iterable<Interface> apiInterfaces = new InterfaceView().getElementIterable(model);
-    Queue<ProtoFile> protoFileQueue = new LinkedList<>();
-    for (Interface apiInterface : apiInterfaces) {
-      protoFileQueue.add(apiInterface.getFile());
-    }
-
-    // BFS to get all proto dependencies.
-    Set<ProtoFile> seenFiles = new HashSet<>();
-    while (!protoFileQueue.isEmpty()) {
-      ProtoFile current = protoFileQueue.remove();
-      seenFiles.add(current);
-      for (ProtoFile protoFile : current.getDependencies()) {
-        if (!seenFiles.contains(protoFile)) {
-          protoFileQueue.add(protoFile);
-        }
-      }
-    }
-    return seenFiles;
-  }
-
   private List<ImportFileView> generateVersionedInitStandardImports() {
     return ImmutableList.of(createImport("__future__", "absoulte_import"));
   }
@@ -254,7 +232,8 @@ public class PythonImportSectionTransformer implements ImportSectionTransformer 
       Model model,
       GapicProductConfig productConfig,
       PackageMetadataConfig packageConfig,
-      SurfaceNamer namer) {
+      SurfaceNamer namer,
+      boolean packageHasEnums) {
 
     Iterable<Interface> apiInterfaces = new InterfaceView().getElementIterable(model);
     ModelTypeTable typeTable = emptyTypeTable(productConfig);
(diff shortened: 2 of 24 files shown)


src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -159,6 +159,20 @@ public String getNamespace(Interface apiInterface) {
     return qualifiedName(namePath.withoutHead());
   }
 
+  /** The namespace of the package. */
```
geigerj: nit: "the _unversioned_ namespace..."

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -475,6 +489,15 @@ public String getByteLengthFunctionName(TypeRef typeRef) {
     return getNotImplementedString("SurfaceNamer.getByteLengthFunctionName");
   }
 
+  /** The name of the class that ties in the helper functions to a client partial veneers. */
+  public String getHelpersClassName(String apiShortName) {
+    return publicClassName(Name.anyCamel(apiShortName).join("helpers"));
+  }
+
+  /** The name of a stub type of an interface */
+  public String getStubType(Interface apiInterface) {
```
geigerj: How is this different from the value we pass to `main.snip` to instantiate the gRPC stub?

src/main/java/com/google/api/codegen/transformer/py/PythonImportSectionTransformer.java
```diff
@@ -151,6 +183,101 @@ private ImportFileView generateApiImport(GapicMethodContext context) {
...
+      // For python, adding a single message from the proto file to the type table will populate
+      // the type table with the correct imports.
+      ImmutableList<MessageType> messages = protoFile.getMessages();
+      if (!messages.isEmpty()) {
+        typeTable.getAndSaveNicknameFor(TypeRef.of(messages.get(0)));
+      }
+    }
+  }
+
+  private Set<ProtoFile> getProtofileDependencies(Model model) {
```
geigerj: optional: it is probably significantly less efficient, but the simpler way to do this would be to subclass the api-compiler [Visitor](https://github.com/googleapis/api-compiler/blob/cc52101f81eae31ebdac647f9664f5b32d2228a1/src/main/java/com/google/api/tools/framework/model/Visitor.java). Here's [a more complicated example](https://github.com/googleapis/api-compiler/blob/4621e9bb63a3903b6b2c4639f6560c9284585a0f/src/main/java/com/google/api/tools/framework/processors/resolver/SymbolTableBuilder.java#L45) of doing this; in your case, the subclass would be trivial (it just needs to add ProtoFiles to a Set when it visits them).

src/main/java/com/google/api/codegen/transformer/py/PythonImportSectionTransformer.java
```diff
@@ -151,6 +183,101 @@ private ImportFileView generateApiImport(GapicMethodContext context) {
...
+      SurfaceNamer namer) {
+
+    Iterable<Interface> apiInterfaces = new InterfaceView().getElementIterable(model);
+    ModelTypeTable typeTable = emptyTypeTable(productConfig);
+    for (Interface apiInterface : apiInterfaces) {
+      namer.getAndSaveNicknameForGrpcClientTypeName(typeTable, apiInterface);
+    }
+
+    Set<ImportFileView> imports = new TreeSet<>(importFileViewComparator());
+    // TODO: (landrito) Make this only generate the enums module if there are enums to export.
```
geigerj: [Easy check](https://github.com/googleapis/toolkit/blob/f0762cd69fa4286fca3cfe7445b369c887ba5061/src/main/java/com/google/api/codegen/py/PythonImportHandler.java#L83). It seems like this was cause an `ImportError` if you actually try to run a GAPIC with no enums, so it seems like it would be best just to address this now.

src/main/java/com/google/api/codegen/transformer/py/PythonSurfaceNamer.java
```diff
@@ -146,8 +146,8 @@ public String getParseFunctionName(String var, SingleResourceNameConfig resource
 
   @Override
   public String getGrpcClientTypeName(Interface apiInterface) {
-    String fullName = getModelTypeFormatter().getFullNameFor(apiInterface) + "Stub";
-    return getTypeNameConverter().getTypeName(fullName).getNickname();
+    String fullName = getModelTypeFormatter().getFullNameFor(apiInterface) + "Client";
```
geigerj: Confused about terminology: `grpcClientTypeName` vs `stubType`?

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
landrito: It's no different. The full stub type was stored in the property that held grpcClientTypeName. Which didn't seem correct. This was to fix that so that I could use the namer.getGrpcClientTypeName for the grpc client type name.

src/main/java/com/google/api/codegen/transformer/py/PythonImportSectionTransformer.java
landrito: OH LOL. I found an even simple solution. `Model` has the method `getFiles`.

src/main/java/com/google/api/codegen/transformer/py/PythonSurfaceNamer.java
landrito: This change is actually to make it such that the terminology is intuitive. The grpcClientTypeName is the type of the grpcClient. The stubType is the type of the grpcClientStub.

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
landrito: Done.

src/main/java/com/google/api/codegen/transformer/py/PythonImportSectionTransformer.java
landrito: Fixed. Had to include that data elsewhere so moved the check into the transformer.

src/test/java/com/google/api/codegen/testdata/python_doc_types_library.baseline
```diff
@@ -0,0 +1,65 @@
...
+               descriptor_pb2
+               duration_pb2
+               empty_pb2
+               protobuf_field_mask_pb2
+               struct_pb2
+               timestamp_pb2
+               wrappers_pb2
+               status_pb2):
+    for name, message in get_messages(module).items():
+        message.__module__ = 'google.cloud.proto.example.library_v1'
```
lukesneeringer: This line is _probably_ wrong? I would expect it to be `google.cloud.library_v1`.

src/test/java/com/google/api/codegen/testdata/python_doc_types_no_path_templates.baseline
```diff
@@ -0,0 +1,39 @@
...
+from google.protobuf import empty_pb2
+
+
+names = []
+for module in (http_pb2
+               no_path_templates_pb2
+               descriptor_pb2
+               empty_pb2):
+    for name, message in get_messages(module).items():
+        message.__module__ = 'google.cloud.proto.example_v1'
```
lukesneeringer: I would expect this to be `message.__module__ = 'google.cloud.example_v1'`. But I can not tell if it is wrong since it is hard to determine that from the bogus examples.

src/test/java/com/google/api/codegen/testdata/python_doc_types_library.baseline
```diff
@@ -0,0 +1,65 @@
+============== file: google/cloud/proto/example/library_v1/types.py ==============
```
lukesneeringer: I would have expected this file location to be `google/cloud/library_v1/types.py`.

src/test/java/com/google/api/codegen/testdata/python_doc_types_no_path_templates.baseline
```diff
@@ -0,0 +1,39 @@
+============== file: google/cloud/proto/example_v1/types.py ==============
```
lukesneeringer: I would have expected this file to be `google/cloud/example_v1/types.py`.

src/test/java/com/google/api/codegen/testdata/python_doc_versioned_directory__init___library.baseline
```diff
@@ -0,0 +1,34 @@
...
+
+from __future__ import absoulte_import
+
+from google.cloud.gapic.example.library.helpers import LibraryHelpers
+from google.cloud.gapic.example.library.v1 import enums
+from google.cloud.proto.example.library.v1 import library_pb2
+from google.cloud.proto.example.library_v1 import types
+
+
+class LibraryServiceClient(LibraryHelpers, library_pb2.LibraryServiceClient):
```
lukesneeringer: `LibraryHelpers` is manually written and probably should not be referenced by autogen.

src/test/java/com/google/api/codegen/testdata/python_doc_versioned_directory__init___no_path_templates.baseline
```diff
@@ -0,0 +1,31 @@
...
+
+
+from __future__ import absoulte_import
+
+from example.helpers import LibraryHelpers
+from google.cloud.proto.example.v1 import no_path_templates_pb2
+from google.cloud.proto.example_v1 import types
+
+
+class NoTemplatesAPIServiceClient(LibraryHelpers, no_path_templates_pb2.NoTemplatesAPIServiceClient):
```
lukesneeringer: `LibraryHelpers` is manually written and likely should not be referenced by autogen.

src/test/java/com/google/api/codegen/testdata/python_types_library.baseline
```diff
@@ -0,0 +1,65 @@
+============== file: google/cloud/proto/example/library_v1/types.py ==============
```
lukesneeringer: Same issue with before re: the file path (and __module__ below)

src/test/java/com/google/api/codegen/testdata/python_doc_types_library.baseline
landrito: I see. Hmm I'll need to think more about how to construct that path.

landrito: @geigerj maybe you have an idea.

I implemented it using the gapic package name but that produced:
google/cloud/**gapic**/example/library_v1/

Now I'm thinking to construct it from the package config shortname name and version but I am trying to figure out how to get the base path.
as in
{base/path}/{shortname}_{version}

Thoughts?

geigerj: What is `base/path`? Is it ever something other than `google/cloud`?

landrito: It will if we ever move towards non cloud apis.

geigerj: It might make sense to use the [`organization_name`](https://github.com/googleapis/googleapis/blob/master/gapic/api/artman_datastore.yaml#L6) configuration, then. (Just replace the hyphen with the file separator.) You probably will need to modify the [package metadata generation task](https://github.com/googleapis/artman/blob/master/artman/tasks/package_metadata_tasks.py#L26) in artman to pass this through to toolkit.

landrito: I think using `organization_name` in that manner seems a bit magical.

I think a more ideal solution would be to have a property in the gapic yaml that holds the gapic package name and a property that holds the top-level package name. I think we should lean towards making these package names explicit rather than parsing the name out of an existing property.

ex:
```yaml
language_settings:
  python:
  package_name: google.cloud.ex_v1
  gapic_package_name: google.cloud.ex_v1.gapic
  proto_package_name: google.cloud.ex_v1.proto
```
Thoughts?

geigerj: I might be missing the full context of this re-organization. Are the `google.cloud.{gapic,proto}.ex.v1` packages going away and being replaced by these `ex_v1.{gapic,proto}` packages?

landrito: Ah yes but I think that it is orthogonal to this decision.

Using the current pattern the yaml would look like.

```
language_settings:
  python:
    package_name: google.cloud.example_v1
    gapic_package_name: google.cloud.gapic.example.v1
    proto_package_name: google.cloud.proto.example.v1

Comment (geigerj) [inline src/test/java/com/google/api/codegen/testdata/python_doc_types_library.baseline]: Ack. I'm not sure there is utility in having `gapic_package_name` and `proto_package_name` if they are always derivable from `package_name`. But just changing the semantics of `package_name` SGTM.

Comment (landrito) [inline src/test/java/com/google/api/codegen/testdata/python_doc_types_library.baseline]: Ack. 

@lukesneeringer how does that sound to you? Simply changing the semantics of the package name to be of the form `google.cloud.example_v1` and having the gapic and proto paths just added to that package name.

Comment (lukesneeringer) [inline src/test/java/com/google/api/codegen/testdata/python_doc_types_library.baseline]: > I might be missing the full context of this re-organization. Are the `google.cloud.{gapic,proto}.ex.v1` packages going away and being replaced by these `ex_v1.{gapic,proto}` packages?

They do not have to be necessarily, but it matters that the ultimate names we expose to users are `google.cloud.example_v1` and only things underneath that.

The current implementation in Vision, Speech, NL, etc. has modules that import from the existing namespaces, but changing the namespaces would actually be easier for documentation and such.

Comment (lukesneeringer) [inline src/test/java/com/google/api/codegen/testdata/python_doc_types_library.baseline]: > @lukesneeringer how does that sound to you? Simply changing the semantics of the package name to be of the form google.cloud.example_v1 and having the gapic and proto paths just added to that package name.
  
That is fine by me. It might even be ideal by me. :-)

Comment (lukesneeringer) [inline src/test/java/com/google/api/codegen/testdata/python_doc_types_library.baseline]: Addendum: I am now of the understanding that the new package names will be:

This means the final paths will be:

```yaml
package_name: google.cloud.example_v1
gapic_package_name: google.cloud.example_v1.gapic
proto_package_name: google.cloud.example_v1.proto
```

But only the first will be explicitly configured. Is that accurate?

lukesneeringer: To be clear, the back-incompat change is fine. I am willing to go update the world to get this change. :-)

landrito: > But only the first will be explicitly configured. Is that accurate?
Correct.

landrito: Took a look at what it's gonna take to do the package name switch. It's a better fit to do that work in a separate PR.

PR #1369, opened 2017-06-19
Add toolkit support for new longrunning operation implementation in Java

vam-google:


src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -505,6 +505,16 @@ message LongRunningConfigProto {
   // Whether or not the server implements cancel.
   bool implements_cancel = 4;
 
-  // The interval of polling, in milliseconds.
-  uint64 polling_interval_millis = 5;
+  // Initial delay after which first poll request will be made.
+  uint64 initial_poll_delay_millis = 5;
+
+  // Multiplier to gradually increase delay between subsequent polls untill it
```
garrettjonesgoogle: spelling (untill)

src/main/java/com/google/api/codegen/viewmodel/LongRunningConfigView.java
```diff
@@ -0,0 +1,54 @@
...
+
+  public abstract Duration operationTotalPollTimeout();
+
+  public static LongRunningConfigView.Builder newBuilder() {
+    return new AutoValue_LongRunningConfigView.Builder();
+  }
+
+  @AutoValue.Builder
+  public abstract static class Builder {
+    public abstract LongRunningConfigView.Builder operationResultTypeName(String val);
```
garrettjonesgoogle: You can remove the `LongRunningConfigView.` qualifier in this file.

build.gradle
```diff
@@ -115,6 +117,8 @@ dependencies {
     libraries.mockito,
     libraries.truth,
     libraries.toolsFxTesting
+
+  apt libraries.autovalue
```
garrettjonesgoogle: Why is it necessary to add this in?

vam-google: I made the changes in build.gradle to add support for AutoValue generated files in Intellij, but you are right, that particular line was not necessary. Removing it.

src/main/java/com/google/api/codegen/viewmodel/LongRunningConfigView.java
vam-google: Refactored, reused LongRunningDetailView instead. Removed this class Completely

src/main/resources/com/google/api/codegen/java/main.snip
```diff
@@ -183,8 +183,13 @@
 @end
 
 @private apiCallableMember(apiCallable)
+  @if apiCallable.hasMetadataTypeName
+    private final {@apiCallable.interfaceTypeName}<{@apiCallable.requestTypeName}, \
+      {@apiCallable.responseTypeName}, {@apiCallable.metadataTypeName}> {@apiCallable.name};
+  @else
   private final {@apiCallable.interfaceTypeName}<{@apiCallable.requestTypeName}, \
```
garrettjonesgoogle: Indent within this `@else` block

PR #1370, opened 2017-06-19
benchmark: add PHP client

pongad:


benchmark/clients/php/local.php
```diff
@@ -0,0 +1,82 @@
...
+{
+  $client = new PublisherClient([
+    'serviceAddress' => $host,
+    'port' => $port,
+    'sslCreds' => $creds,
+  ]);
+  return function($top) use($client) {
+    return $client->getTopic($top);
+  };
+}
```
pongad: @michaelbausor Do we have a way to get run async RPC in PHP? All functions I see return synchronously but maybe I don't understand it well enough.

michaelbausor: No, there is no way currently to run an async RPC

michaelbausor: I assume you mean from the GAPIC client? If you are using the grpc stub, you can make a pseudo-async call just by calling the appropriate stub method, which returns a call object with a `wait()` method. The call will run in the background, and then you call `wait()` when you need the response. I say pseudo-async because your only option is to call `wait()`, which is blocking - you cannot check for completion without blocking.

benchmark/clients/php/composer.lock
```diff
@@ -0,0 +1,815 @@
+{
```
michaelbausor: This file should not be committed - it is generated by `composer install`.

benchmark/clients/php/local.php
pongad: ack

benchmark/clients/php/local.php
```diff
@@ -0,0 +1,82 @@
...
+function grpc_get($host, $port, $creds)
+{
+  $client = new PublisherGrpcClient(sprintf("%s:%d", $host, $port), [
+    'credentials' => $creds,
+  ]);
+  $meta = array('x-goog-api-client' => ['gl-php/5.5.9-1ubuntu4.21 gapic/0.5.1 gax/0.9.0 grpc/0.1.0']);
+
+  return function($top) use($client, $meta) {
+    $req = new GetTopicRequest();
+    $req->setTopic("foo");
```
michaelbausor: Should this use `$top`?

benchmark/clients/php/local.php
```diff
@@ -0,0 +1,82 @@
...
+} else {
+  die("unknown client: $client");
+}
+
+$fn = $getfn($host, $port, $creds);
+
+$warmtime = intval(isset($options['warmup_time']) ? $options['warmup_time'] : '60');
+$runtime = intval(isset($options['run_time']) ? $options['run_time'] : '60');
+
+define("TOPIC", "projects/benchmark-project/topics/benchmark-topic");
```
michaelbausor: Use `const` rather than `define`

pongad:
@michaelbausor Done. PTAL.

PR #1371, opened 2017-06-19
Don't create directory structure for C# output

jskeet:
This will make it easier to move the files into the right directory structure in Artman.


jskeet:
(Looks like I need to move the expected output files as well for tests... will do that in a separate commit.)

jskeet:
Merging before @ethanbao looks at it just on the grounds that it should be absolutely fine...

ethanbao:
LGTM

PR #1372, opened 2017-06-19
Derive the original Discovery document URL

saicheems:


src/main/java/com/google/api/codegen/discovery/config/js/JSTypeNameGenerator.java
```diff
@@ -37,8 +39,19 @@ public String getStringFormatExample(String format) {
...
+    try {
+      uri = new URI(rootUrl);
+    } catch (URISyntaxException e) {
+      throw new IllegalArgumentException(String.format("malformed URL: %s", rootUrl));
+    }
+    String host = uri.getHost();
+    // The host is either of the form "www.googleapis.com" or "www.{apiName}.googleapis.com".
+    if (host.equals("www.googleapis.com")) {
+      return String.format("https://%s/discovery/v1/apis/%s/%s/rest", host, apiName, apiVersion);
+    } else {
```
tcoffee-google: I think it would be safer to make an explicit case here, and fail otherwise.

saicheems: Done

PR #1376, opened 2017-06-20
Add uploadArchive to the Java sample build file.

shinfan:
This PR also fixes indentation of the snip file. It would be easier to review the diff by checking the baseline file.


src/test/java/com/google/api/codegen/testdata/java_sample_library.baseline
```diff
@@ -50,3 +52,131 @@ jar {
...
+      workingDir javaDocRoot
+      commandLine 'git', 'commit', '-m', "Regenerating docs for ${packageName} ${project.version}"
+    }
+    println 'New docs have been generated under tmp_gh-pages and have been committed;'
+    println 'Next they just need to be pushed with "git push".'
+  }
+}
+
+// Regenerates and push the gh-pages branch under tmp_gh-pages
+task updateAndPushDocs {
```
garrettjonesgoogle: This is an idea for later - maybe we should make our own plugin so we don't have to repeat this tasks in every single build.gradle we own.

garrettjonesgoogle:
LGTM

src/test/java/com/google/api/codegen/testdata/java_sample_library.baseline
shinfan: That sounds reasonable to me.

PR #1378, opened 2017-06-21
Relax NodeJS version regex

### LINKED ISSUE #1335: Incorrect node.js path mapping for alpha/beta versions
https://github.com/googleapis/toolkit/blob/master/src/main/java/com/google/api/codegen/nodejs/NodeJSCodePathMapper.java#L30

There shouldn't be any number required after alpha/beta. The regex doesn't match version names like v1alpha.

geigerj:
Instead of the custom, overrestrictive version regex, reuse the
version regex provided by api-compiler.

Fixes #1335

### DIFF
--- a/src/main/java/com/google/api/codegen/nodejs/NodeJSCodePathMapper.java
+++ b/src/main/java/com/google/api/codegen/nodejs/NodeJSCodePathMapper.java
@@ -16,6 +16,7 @@ package com.google.api.codegen.nodejs;
 
 import com.google.api.codegen.config.ProductConfig;
 import com.google.api.codegen.gapic.GapicCodePathMapper;
+import com.google.api.tools.framework.aspects.versioning.model.ApiVersionUtil;
 import com.google.api.tools.framework.model.ProtoElement;
 import com.google.common.base.Splitter;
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/api/codegen/nodejs/NodeJSCodePathMapper.java
```diff
@@ -27,7 +28,7 @@ public String getOutputPath(ProtoElement element, ProductConfig config) {
     List<String> packages = Splitter.on(".").splitToList(element.getFullName());
     if (packages.size() > 2) {
       String parentName = packages.get(packages.size() - 2);
-      if (parentName.matches("v[0-9]+((alpha|beta)[0-9]+)?")) {
+      if (ApiVersionUtil.SEMANTIC_VERSION_REGEX_PATTERN.matcher(parentName).matches()) {
```
eoogbe: You can use `ApiVersionUtil.isValidApiVersion(apiVersion)` instead.

eoogbe:
Closing as #1545 does the same thing but using the relatively newer `VersionMatcher`.

PR #1384, opened 2017-06-23
Nodeify Discovery objects

saicheems:
Document, Method, and Schema are all created from DiscoveryNodes, which
should be part of a JSON tree structure. This PR unites Document,
Method, and Schema under one Node interface, which exposes a method to
get the parent of a node.

The ability to get the parent of a node is useful in the context it
provides when operating on a single Node; ex. when transforming a
Method, it would be easy to fetch the name "pubsub" from the parent
Document node.


saicheems:
This is the same as your earlier PR, but I want it in master.

PR #1385, opened 2017-06-23
Add a dereference method to Schema

saicheems:
It's more convenient to allow Schemas to dereference themselves than it
is to pass around Document all over the place.

This commit adds a `dereference` function to Schema that

1. If the Schema references nothing, the Schema is returned
1. Walks up the Node tree to a Document, if one exists
2. Returns the Schema that represents the reference


garrettjonesgoogle:
LGTM

PR #1387, opened 2017-06-26
Ruby: add use of keyfile param and Google::Gax::Credentials

landrito:
Updates: https://github.com/googleapis/toolkit/issues/1324


src/main/resources/com/google/api/codegen/ruby/main.snip
```diff
@@ -138,13 +138,20 @@
   @#   The domain name of the API remote host.
   @# @@param port [Integer]
   @#   The port on which to connect to the remote host.
+  # TODO(landrito): add documentation specifying that the channel updater_proc and keyfile
```
geigerj: For trackability, would prefer an issue to a comment for this sort of thing, unless you're planning to fix this immediately in a follow-up PR.

src/main/resources/com/google/api/codegen/ruby/main.snip
```diff
@@ -138,13 +138,20 @@
...
+  # must be used independently in relation to each other.
   @# @@param channel [Channel]
   @#   A Channel object through which to make calls.
-  @# @@param chan_creds [Grpc::ChannelCredentials]
```
geigerj: We should not remove this in case the user wants to set, e.g., non-default SSL creds. Could we have one parameter that accepts either `String` or `Grpc::ChannelCredentials`?

src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceTransformer.java
```diff
@@ -191,4 +207,58 @@ private ViewModel generateVersionIndexView(Model model, GapicProductConfig produ
                 productConfig, ImportSectionView.newBuilder().build(), namer))
         .build();
   }
+
+  private ViewModel generateCredentialsView(Model model, GapicProductConfig productConfig) {
+    SurfaceNamer namer = new RubySurfaceNamer(productConfig.getPackageName());
+    CredentialsClassView credentialsClass = generateCredentialsClass(model, productConfig);
+    Iterable<Interface> apiInterfaces = new InterfaceView().getElementIterable(model);
+    // The following assumes that all generated interfaces are generated to the same output path.
```
geigerj: What happens if this assumption is broken? (I.e., a single product contains two services with different namespaces.)

src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceTransformer.java
```diff
@@ -47,12 +52,19 @@
...
 import com.google.common.collect.ImmutableList;
+import java.io.File;
 import java.util.List;
 
 /** The ModelToViewTransformer to transform a Model into the standard GAPIC surface in Ruby. */
 public class RubyGapicSurfaceTransformer implements ModelToViewTransformer {
   private static final String VERSION_INDEX_TEMPLATE_FILE = "ruby/version_index.snip";
   private static final String XAPI_TEMPLATE_FILENAME = "ruby/main.snip";
+  private static final String CREDENTIALS_CLASS_TEMPLATE_FILE = "ruby/credentials.snip";
+  // This assumes the api is a google-cloud api.
```
geigerj: Good point. We should not accept the `GOOGLE_CLOUD`/`GCLOUD` env vars for non-cloud APIs. We know what's cloud and non-cloud from the artman config's `organization` field, so that should be possible.

landrito: I think we can leave as-is for the time being and solve it in the future. I think it's a bit of work to pipe artman config's organization field to toolkit.

src/main/resources/com/google/api/codegen/ruby/main.snip

src/main/resources/com/google/api/codegen/ruby/main.snip
landrito: This todo no longer applies.

src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceTransformer.java
landrito: This comment no longer applies. I made the output path based on the package name.

landrito:
Hi Jacob this is ready for another round of review.

I didn't remove the old parameters yet. I simply removed the documentation and spat out a deprecation warning if someone tries to use it.

src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceTransformer.java
geigerj: SGTM. Please file an issue to track.

src/main/resources/com/google/api/codegen/ruby/main.snip
```diff
@@ -175,13 +184,23 @@
       require "{@import.moduleName}"
     @end
 
+    # TODO(landrito): remove these parameters after an appropriate amount of time.
```
geigerj: Nit: since this is autogen code, I think a more generic note is appropriate like,

`Warning: these parameters will be removed on 2017/09/08`.

(I think we can pick a date 8 weeks out or so.)

landrito:
Tracked issue is here: https://github.com/googleapis/toolkit/issues/1403

src/main/resources/com/google/api/codegen/ruby/main.snip

PR #1390, opened 2017-06-27
PHP support tests for clients without scopes or service address

michaelbausor:


src/main/java/com/google/api/codegen/viewmodel/testing/ClientTestClassView.java
```diff
@@ -42,6 +42,18 @@
   @Nullable
   public abstract String packageServiceName();
 
+  public abstract boolean hasDefaultServiceAddress();
+
+  public abstract boolean hasDefaultServiceScopes();
+
+  public boolean missingDefaultServiceAddress() {
```
shinfan: We can just have this missingDefaultServiceAddress() and in the snip file we can do:
```
@if missingDefaultServiceAddress
....
@else
...
@end
```

michaelbausor: Done

PR #1392, opened 2017-06-28
Generate request classes

andreamlin:
This PR generates Request objects analogous to protobuf RPC request objects.


andreamlin:
PTAL.

src/test/java/com/google/api/codegen/testdata/discogapic/java/java_fakebilling.v1.json.baseline
```diff
@@ -337,47 +337,179 @@ public final class BillingAccount implements Serializable {
...
+  private BillingAccountsGetHttpRequest() {
+    this.queryParams = (new ImmutableListMultimap.Builder()).build();
+    this.pathParams = (new ImmutableListMultimap.Builder()).build();
   }
 
-  public String getValue() {
-    return value;
+  // A list of key-value pairs, where each key is a query parameter and the value is the value
+  // of the parameter. A key may appear multiple times, in the case of repeatable parameters.
+  public ImmutableListMultimap<String, String> getQueryParams() {
```
garrettjonesgoogle: We don't want to expose Guava types on the surface anywhere (either generated types or types from GAX)

andreamlin: Ok, removed all Immutable* classes from generated code.

andreamlin:
Generated Request message now implements a to-be-written-gax-http Message interface, which has the abstract method populateFieldsInMap().

PTAL.

src/test/java/com/google/api/codegen/testdata/discogapic/java/java_fakebilling.v1.json.baseline
```diff
@@ -382,54 +381,55 @@ public final class BillingAccountsGetHttpRequest {
...
-    ImmutableListMultimap.Builder queryParams = ImmutableListMultimap.builder();
+  private BillingAccountsGetHttpRequest() {}
 
-    if (name != null) {
-      pathParams.put("name", name.toString());
+  public void populateFieldsInMap(Set<String> fieldNames, Map<String, List<String>> map) {
+    if (fieldNames.contains("name")) {
+      List<String> values = new LinkedList<>();
+      values.add(name);
+      map.put("name", values);
```
garrettjonesgoogle: I think it might be nice to make this denser:

`map.put("name", Lists.newArrayList(name));`

Or maybe we want to make it immutable...

`map.put("name", Collections.singletonList(name));`

Although if we're making it immutable, we could just return the result instead of having to accept it as a parameter.

src/test/java/com/google/api/codegen/testdata/discogapic/java/java_fakebilling.v1.json.baseline
```diff
@@ -548,6 +553,16 @@ public final class BillingAccountsGetHttpRequest {
       return this;
     }
 
+    public List<String> getFooRepeated() {
+      return fooRepeated;
+    }
+
+    public Builder setFooRepeated(List<String> fooRepeated) {
+      this.fooRepeated = new LinkedList<>();
```
garrettjonesgoogle: We should use an ImmutableList instead of LinkedList so that the message is immutable after construction.

src/test/java/com/google/api/codegen/testdata/discogapic/java/java_fakebilling.v1.json.baseline
```diff
@@ -337,19 +337,18 @@ public final class BillingAccount implements Serializable {
...
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
+import java.util.Set;
 import javax.annotation.Generated;
 
 @Generated("by GAPIC")
 @BetaApi
-public final class BillingAccountsGetHttpRequest {
+public final class BillingAccountsGetHttpRequest implements Message {
```
garrettjonesgoogle: I think we should use a more unique class name for the message interface - I suggest `ApiMessage`

src/main/java/com/google/api/codegen/discogapic/transformer/DiscoGapicNamer.java
```diff
@@ -32,11 +32,46 @@ public String getFieldVarName(String fieldName) {
 
   /** Returns the resource getter method name for a resource field. */
   public String getResourceGetterName(String fieldName) {
-    return publicMethodName(Name.anyCamel("get", fieldName));
+    Name name;
+    if (fieldName.contains("_")) {
+      name = Name.anyCamel(fieldName.split("_"));
```
garrettjonesgoogle: Interesting, why is this necessary?

src/main/java/com/google/api/codegen/discogapic/transformer/DiscoGapicNamer.java
```diff
@@ -32,11 +32,46 @@ public String getFieldVarName(String fieldName) {
...
+   * "compute.addresses.aggregatedList" returns "aggregatedList".
+   */
+  public String[] getMethodNamePieces(String longMethodName) {
+    String[] pieces = longMethodName.split("\\.");
+    if (pieces.length < 3) {
+      throw new IllegalArgumentException(
+          "Fully qualified method name must be in the form [api].[resource].[method]");
+    }
+    return pieces;
+  }
```
garrettjonesgoogle: The implementation doesn't seem to match the documentation

src/main/java/com/google/api/codegen/discogapic/transformer/java/JavaDiscoGapicRequestToViewTransformer.java
```diff
@@ -0,0 +1,248 @@
...
+  }
+
+  static {
+    reservedKeywords.addAll(JavaNameFormatter.RESERVED_IDENTIFIER_SET);
+    reservedKeywords.add("Builder");
+  }
+
+  private static final String XAPI_TEMPLATE_FILENAME = "java/main.snip";
+  private static final String PACKAGE_INFO_TEMPLATE_FILENAME = "java/package-info.snip";
+  private static final String REQUEST_TEMPLATE_FILENAME = "java/http_request.snip";
```
garrettjonesgoogle: Remove the template filenames you aren't using

src/main/java/com/google/api/codegen/discogapic/transformer/java/JavaDiscoGapicSchemaToViewTransformer.java
```diff
@@ -168,7 +167,7 @@ private StaticLangApiMessageView generateSchemaClasses(
     // Getters and setters use unescaped name for better readability on public methods.
     schemaView.fieldGetFunction(context.getDiscoGapicNamer().getResourceGetterName(schemaId));
     schemaView.fieldSetFunction(context.getDiscoGapicNamer().getResourceSetterName(schemaId));
-    String schemaTypeName = schemaTypeTable.getAndSaveNicknameForElementType(schema);
+    String schemaTypeName = schemaTypeTable.getAndSaveNicknameFor(schema, true);
```
garrettjonesgoogle: The boolean parameter is unclear at the calling site, i.e. I can't tell what `true` means here. Consider using an enum (since you get names that way).

I see that the function here was private and is now becoming public; it was fine for the parameter to be a boolean when private, because of the smaller scope (it's easier to find the definition and find what it means).

src/main/java/com/google/api/codegen/viewmodel/StaticLangApiHttpRequestView.java
```diff
@@ -0,0 +1,106 @@
...
+  public List<SimpleParamView> allParams() {
+    if (allParams != null) {
+      return allParams;
+    }
+    ImmutableList.Builder<SimpleParamView> builder = new ImmutableList.Builder<>();
+    builder.addAll(pathParams());
+    builder.addAll(queryParams());
+    allParams = builder.build();
+    return allParams;
+  }
```
garrettjonesgoogle: I would prefer logic like this to live outside of the view model (i.e. the transformer)

src/main/resources/com/google/api/codegen/java/http_request.snip
```diff
@@ -0,0 +1,233 @@
...
+@snippet generate(apiFile)
+  {@renderFileHeader(apiFile.fileHeader)}
+
+  @@Generated("by GAPIC")
+  @@BetaApi
+  {@requestClassDef(apiFile.request)}
+@end
+
+@private requestClassDef(request)
+  public final class {@request.typeName} implements Message {
```
garrettjonesgoogle: I actually had in mind using a single template for all message types (i.e. including request message types). This means that non-request message types will have some extra stuff they technically don't need (`populateMapMethod`), but I think that's fine.

src/test/java/com/google/api/codegen/testdata/discogapic/java/java_fakebilling.v1.json.baseline
```diff
@@ -437,14 +676,28 @@ public final class Data implements Serializable {
...
     h ^= this.key.hashCode();
     h *= 1000003;
-    h ^= this.value.hashCode();
+    h ^= this.prettyPrint.hashCode();
+    h *= 1000003;
+    h ^= this.quotaUser.hashCode();
+    h *= 1000003;
+    h ^= this.userIp.hashCode();
+    h *= 1000003;
+    h ^= this.fooRepeated.hashCode();
```
garrettjonesgoogle: We'd get a lot less implementation noise if we use `Objects.hash()`

src/test/java/com/google/api/codegen/testdata/discogapic/java/java_fakebilling.v1.json.baseline
```diff
@@ -422,13 +655,19 @@ public final class Data implements Serializable {
...
-      ;
+          this.name.equals(that.getName())&&
+          this.access_token.equals(that.getAccessToken())&&
+          this.callback.equals(that.getCallback())&&
+          this.fields.equals(that.getFields())&&
+          this.key.equals(that.getKey())&&
+          this.prettyPrint.equals(that.getPrettyPrint())&&
+          this.quotaUser.equals(that.getQuotaUser())&&
+          this.userIp.equals(that.getUserIp())&&
+          this.fooRepeated.equals(that.getFooRepeated())
```
garrettjonesgoogle: I think we would want to allow for fields to be null, so this would throw null pointer exceptions. Let's use `Objects.equals()` here.

src/test/java/com/google/api/codegen/testdata/discogapic/java/java_fakebilling.v1.json.baseline
andreamlin: sure

src/test/java/com/google/api/codegen/testdata/discogapic/java/java_fakebilling.v1.json.baseline
andreamlin: oh neat! i like methods without side effects, so just returning the result now.

src/test/java/com/google/api/codegen/testdata/discogapic/java/java_fakebilling.v1.json.baseline
andreamlin: ImmutableList is in the com.google.common.collect package of Guava. I thought that users were allergic to Guava?

src/test/java/com/google/api/codegen/testdata/discogapic/java/java_fakebilling.v1.json.baseline
andreamlin: Done.

src/test/java/com/google/api/codegen/testdata/discogapic/java/java_fakebilling.v1.json.baseline
garrettjonesgoogle: Ahh! Here is an important clarification: Guava shouldn't be used on the surface (i.e. you shouldn't see any Guava types when you look at the javadoc for our library), but it can be used in the implementation.

src/main/java/com/google/api/codegen/discogapic/transformer/DiscoGapicNamer.java
andreamlin: Some parameter names defined in the Discovery doc have an underscore, ex "order_by." All the Name.*Camel methods will throw an exception if the input has underscores (because it's not in camelcase).

src/main/java/com/google/api/codegen/discogapic/transformer/DiscoGapicNamer.java
andreamlin: Fixed documentation.

src/main/java/com/google/api/codegen/discogapic/transformer/java/JavaDiscoGapicRequestToViewTransformer.java
andreamlin: done.

src/test/java/com/google/api/codegen/testdata/discogapic/java/java_fakebilling.v1.json.baseline
andreamlin: Oh that makes sense. Done.

src/test/java/com/google/api/codegen/testdata/discogapic/java/java_fakebilling.v1.json.baseline
andreamlin: Done.

src/main/java/com/google/api/codegen/discogapic/transformer/java/JavaDiscoGapicSchemaToViewTransformer.java
andreamlin: Removing the boolean param for now, since all current invocations only pass in "true".

src/main/java/com/google/api/codegen/viewmodel/StaticLangApiHttpRequestView.java
andreamlin: Moved to the request transformer.

src/main/resources/com/google/api/codegen/java/http_request.snip
andreamlin: Done. Request and non-request messages share the same template and ViewModel. However, non-request message types won't implement ApiMessage because the only planned method in ApiMessage is populateMapMethod, and that is undefined for non-request message types.

garrettjonesgoogle: Why should it be undefined? It's generic enough to not care, right?

andreamlin:
Addressed all comments. PTAL.

src/test/java/com/google/api/codegen/testdata/discogapic/java/java_fakebilling.v1.json.baseline
```diff
@@ -3591,24 +3546,35 @@ public final class Warnings implements Serializable {
 
 
     public Warnings build() {
-      return new Warnings(
+      String missing = "";
```
andreamlin: It's similar to AutoValue's build() method, which enforces non-nullness on required fields. If a required field is missing, then the name of that field is appended to the String missing.

As of this comment, changed so that
- that line only shows up if there are required fields
- throw an exception if any required fields are null, with variable [missing] as the error message.

src/main/java/com/google/api/codegen/discogapic/transformer/java/JavaDiscoGapicRequestToViewTransformer.java
```diff
@@ -169,73 +165,75 @@ private StaticLangApiHttpRequestView generateRequestClass(
...
       paramView.typeName("String");
+      paramView.innerTypeName("String");
       paramView.isRequired(false);
       paramView.canRepeat(false);
-      paramView.getterFunction(context.getDiscoGapicNamer().getResourceGetterName(param));
-      paramView.setterFunction(context.getDiscoGapicNamer().getResourceSetterName(param));
-      queryParamViews.add(paramView.build());
+      paramView.fieldGetFunction(context.getDiscoGapicNamer().getResourceGetterName(param));
+      paramView.fieldSetFunction(context.getDiscoGapicNamer().getResourceSetterName(param));
+      paramView.innerTypeName("String");
```
andreamlin: Removed line 186.

src/main/java/com/google/api/codegen/discogapic/transformer/DiscoGapicNamer.java
```diff
@@ -52,8 +52,8 @@ public String getResourceSetterName(String fieldName) {
   }
 
   /**
-   * Returns the last substring after the input is split by periods. Ex: Input
-   * "compute.addresses.aggregatedList" returns "aggregatedList".
+   * Returns the last substring after the input is split by periods. Ex: Input of
```
andreamlin: Fixed method comment.

src/test/java/com/google/api/codegen/testdata/discogapic/java/java_fakebilling.v1.json.baseline
```diff
@@ -3620,28 +3586,23 @@ public final class Warnings implements Serializable {
...
       return
-
-      this.code.equals(that.getCode())&&
-      this.data.equals(that.getData())&&
-      this.message.equals(that.getMessage())&&
-      this.projectBillingInfo.equals(that.getProjectBillingInfo())
-      ;
+          Objects.equals(this.code, that.getCode())&&
+          Objects.equals(this.data, that.getData())&&
+          Objects.equals(this.message, that.getMessage())&&
```
andreamlin: Done.

src/main/resources/com/google/api/codegen/java/message.snip
```diff
@@ -10,10 +10,19 @@
 @end
 
 @private schemaClassDef(schema)
-  public final class {@schema.innerTypeName} implements Serializable {
+  public final class {@schema.innerTypeName} implements Serializable
+      @if schema.isRequestMessage()
+        , ApiMessage {
```
andreamlin: Done.

src/main/resources/com/google/api/codegen/java/message.snip
```diff
@@ -134,19 +156,27 @@
...
-@end
+    public Builder clone() {
+      Builder newBuilder = new Builder();
+      @join param : schema.properties
+        newBuilder.{@param.fieldSetFunction}(this.{@param.name});
+      @end
+      return newBuilder;
+    }
+  }
+@end
```
andreamlin: Done.

src/main/java/com/google/api/codegen/transformer/SchemaTypeNameConverter.java
```diff
@@ -24,6 +24,9 @@
   /** Provides a TypeName for the given Schema. */
   TypeName getTypeName(Schema type);
 
+  /** Provides a TypeName for the given Schema. */
+  TypeName getTypeName(Schema type, boolean shouldBoxPrimitives);
```
andreamlin: finally got around to enum-ing this :)

andreamlin:
Fixed up code in response to comments.
Travis build for Java 8 passes.
PTAL.

src/main/java/com/google/api/codegen/transformer/SchemaTypeNameConverter.java
```diff
@@ -21,11 +21,20 @@
 
 /** SchemaTypeNameConverter maps Schema instances to TypeName instances. */
 public interface SchemaTypeNameConverter {
+
+  enum BoxPrimitives {
```
garrettjonesgoogle: This sounds like a command - maybe name it `BoxingBehavior`?

src/main/java/com/google/api/codegen/transformer/java/JavaSchemaTypeNameConverter.java
```diff
@@ -111,19 +111,22 @@ public TypedValue getEnumValue(Schema schema, String value) {
...
-          primitiveType = new TypeName(JavaTypeTable.getBoxedTypeName(primitiveTypeName));
-        } else {
-          primitiveType = new TypeName(primitiveTypeName);
+        switch (shouldBoxPrimitives) {
+          case BOX_PRIMITIVES:
+            primitiveType = new TypeName(JavaTypeTable.getBoxedTypeName(primitiveTypeName));
+            break;
+          case NO_BOX_PRIMTIVES:
+            primitiveType = new TypeName(primitiveTypeName);
+            break;
```
garrettjonesgoogle: Add a default case (throw an exception)

src/main/resources/com/google/api/codegen/java/message.snip
```diff
@@ -10,12 +10,7 @@
 @end
 
 @private schemaClassDef(schema)
-  public final class {@schema.innerTypeName} implements Serializable
-      @if schema.isRequestMessage()
-        , ApiMessage {
-      @else
-        {
-      @end
+  public final class {@schema.innerTypeName} implements Serializable, ApiMessage {
```
garrettjonesgoogle: I think we can simplify this by assuming that `ApiMessage` implements `Serializable`.

src/main/java/com/google/api/codegen/transformer/SchemaTypeNameConverter.java
andreamlin: Done.

src/main/resources/com/google/api/codegen/java/message.snip
andreamlin: Done.

src/main/java/com/google/api/codegen/transformer/java/JavaSchemaTypeNameConverter.java
andreamlin: Done.

andreamlin:
Fixed up code in response to comments.
Travis build for Java 8 passes.
PTAL.

garrettjonesgoogle:
LGTM

PR #1395, opened 2017-06-29
Add scaffold for the revamped C# samplegen

saicheems:


saicheems:
Note that this is to a separate branch, not master, and I don't intend for the build to pass.

I have a very large code change that I intend to split into many small (<250 line) CLs to make review go smoother and help explain what I'm trying to do in this design.

This PR sets up some temporary scaffolding to support a new C# sample transformer that uses the new discovery parser I designed several weeks ago. Since it's intended as scaffolding for the time being, please accept any roughness here as it will be 100% polished before this branch merges with master.

Specifically, this PR constructs a `Document` and redirects the `ViewModelProvider` to use the new `CSharpSampleTransformer` class to construct a `SampleView`. The new snippet file, `discovery_sample.snip`, prints nothing right now.

vchudnov-g:
ok as temporary scaffolding, but why do we need sample transformer for C#?

saicheems:
I'm doing some refactoring, in this case `SampleTransformer` is analogous to `SampleMethodToViewTransformer`.

I'm making a new one mostly because it'll be much easier to make sense of the diffs this way, as a lot of things are moving around.

PR #1398, opened 2017-06-30
NodeJS: fix bug where initcode was printed twice in samples.

landrito:


swcloud:
@landrito Actually I looked at it again. The duplicated init code is displayed for different cases under page streaming or long running handling, so I feel that it is okay to repeat the init code for the different cases. I guess this PR might not needed.

PR #1402, opened 2017-07-05
Rename python test output path

### LINKED ISSUE #1401: Reorganize GAPIC tests
Python smoke tests have recently been added to autogen. Yay!

They are being put in a place where they will never run, though, because the default test loader for `unittest` requires a `test_` prefix for every file.

Right now the directory structure is:

```
test/
  google/
    cloud/
      {api}/
        {version}/
          {api}_smoketest.py
          test_{client}.py
```

This should become:

```
tests/  # Note the -s suffix.
  unit/
    gapic/
      {version}/
        test_{client}.py
  system/
    gapic/
      {version}/
        test_{api}.py
```

The split for `unit/` and `system/` is also critical because our test setup in CI knows when to run and not run system tests.

eoogbe:
Fixes #1401


eoogbe:
@lukesneeringer PTAL

PR #1411, opened 2017-07-13
harden firstLine

pongad:
Previously this method simply throws if the API defines no doc.
Make this return empty string instead.
The doc will render weird but at least we don't crash.

### DIFF
--- a/src/main/java/com/google/api/codegen/viewmodel/ServiceDocView.java
+++ b/src/main/java/com/google/api/codegen/viewmodel/ServiceDocView.java
@@ -15,6 +15,7 @@
 package com.google.api.codegen.viewmodel;
 
 import com.google.auto.value.AutoValue;
+import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/api/codegen/viewmodel/ServiceDocView.java
```diff
@@ -40,11 +41,17 @@ public static Builder newBuilder() {
...
+    }
+    return "";
   }
 
   public List<String> remainingLines() {
-    return lines().subList(1, lines().size());
+    if (lines().size() > 0) {
+      return lines().subList(1, lines().size());
+    }
+    return Collections.<String>emptyList();
```
garrettjonesgoogle: Putting more logic like this in the view model will make it more difficult to just store view models as data later (if we decide to do so). Could we move the firstLine/remainingLine logic somewhere else? There are two other view models that have this pattern too (`MapParamDocView` and `SimpleParamDocView`).

PR #1417, opened 2017-07-19
Update gradle

bootstraponline:
Update protobuf-gradle-plugin to the latest stable version
Fix #379 convert double-quoted strings to single-quoted strings
Fix #1419 Update gradle to the latest stable version

PR #1421, opened 2017-07-19
Fix attempt to modify immutable list.

vchudnov-g:


src/main/java/com/google/api/codegen/configgen/LanguageGenerator.java
```diff
@@ -128,8 +129,9 @@ public String getFormattedPackageName(String packageName) {
       // google.logging.v2 => cloud.google.com/go/logging/apiv2
       // Otherwise, fall back to backup
       if (!isApiGoogleCloud(nameComponents)) {
-        nameComponents.add(0, "google.golang.org");
-        return Joiner.on("/").join(nameComponents);
+        List<String> newComponents = Lists.newArrayList(nameComponents);
```
garrettjonesgoogle: I think it would be easier to follow the code if we used a mutable list everywhere instead of immutable in some places and mutable in others. (I don't think mutability is a concern since the list is local to the function). Concretely, I'm suggesting that the original list be created with `Lists.newArrayList(Splitter...);`

garrettjonesgoogle:
LGTM

ethanbao:
Has CI test failure. Please address.

vchudnov-g:
I talked with @garrettjonesgoogle earlier today. The Travis failures are appearing elsewhere as well; seems like a fix is in #1417

PR #1422, opened 2017-07-19
Add ResourceName compatibility between OneOf (on method) and Any (on message) types

vam-google:
This makes possible using the following configuration (when a flattened method expects OneOf and the message expects Any):

```yaml
methods:
  - name: GetBookFromAbsolutelyAnywhere
    resource_name_treatment: STATIC_TYPES
    field_name_patterns:
      name: book_oneof

resource_name_generation:
- message_name: GetBookFromAbsolutelyAnywhereRequest
  field_entity_map:
    name: "*"

```
This is required when the message belongs to common (not specific to any client) module, like IAM.

For example, if we have a common mixin method (like `google.iam.v1.TestIamPermissions`), which accepts a common protobuf message (like `google.iam.v1.TestIamPermissionsRequest`), and the message is configured to accept ANY resource name, then this change allows to reuse that common method/message (via `reroute_to_grpc_interface` option) in multiple different clients, when each client has its own (different) OneOf resource name for the ANY field in the common message:

in **iam_gapic.yaml** (uses `"*"` (i.e. any), is shared by multiple clients):
```yaml
resource_name_generation:
- message_name: TestIamPermissionsRequest
  field_entity_map:
    resource: "*
```

In **client1_gapic.yaml** (uses `client1_iam_oneof`):
```yaml
methods:
  - name: TestIamPermissions
    reroute_to_grpc_interface: google.iam.v1.IAMPolicy
    field_name_patterns:
      resource: client1_iam_oneof
    resource_name_treatment: STATIC_TYPES
resource_name_generation:
- message_name: google.iam.v1.TestIamPermissionsRequest
  field_entity_map:
    resource: "*"
```

In **client2_gapic.yaml** (uses `client2_iam_oneof`) :
```yaml
methods:
  - name: TestIamPermissions
    reroute_to_grpc_interface: google.iam.v1.IAMPolicy
    field_name_patterns:
      resource: client2_iam_oneof
    resource_name_treatment: STATIC_TYPES
resource_name_generation:
- message_name: google.iam.v1.TestIamPermissionsRequest
  field_entity_map:
    resource: "*"
```


src/main/java/com/google/api/codegen/config/FieldConfig.java
```diff
@@ -132,8 +132,9 @@ static FieldConfig createFieldConfig(
         && !messageFieldResourceNameConfig.equals(flattenedFieldResourceNameConfig)) {
       // We support the case of the flattenedField using a specific resource name type when the
       // messageField uses a oneof containing that type
```
michaelbausor: Maybe update this comment with something like: "We support the case of the flattenedField using a specific resource name type when the messageField uses a oneof containing that type **or when the messageField accepts any resource name.**"

src/main/java/com/google/api/codegen/config/FieldConfig.java
```diff
@@ -210,6 +211,7 @@ public FieldConfig withResourceNameConfig(ResourceNameConfig resourceNameConfig)
   public boolean hasDifferentMessageResourceNameConfig() {
     return getResourceNameConfig() != null
         && getMessageResourceNameConfig() != null
+        && getMessageResourceNameConfig().getResourceNameType() != ResourceNameType.ANY
```
michaelbausor: I don't understand why this is necessary. Doesn't this mean that the `hasDifferentMessageResourceNameConfig()` function will return false whenever the message resource name is Any? That is incorrect if the method resource name is some more specific resource name or oneof.

Edit: looking into it further, I see that this function is being used to determine whether we should apply a parameter transformation, and that in the ANY case we don't want to do that transformation. Given that, can we either rename this method so that it describes itself accurately, or else move the check for ANY into the places this function is called instead.

src/main/java/com/google/api/codegen/config/FieldConfig.java
vam-google: Updated

src/main/java/com/google/api/codegen/config/FieldConfig.java
vam-google: Renamed to requiresParamTransformation()

PR #1425, opened 2017-07-21
Ruby: Credentials bugs fix

landrito:
This also moves the credentials class to the top level scope since the scopes are not per interface but rather a property of the model.

Monitoring unit and smoke tests now pass with these changes.

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceTransformer.java
@@ -244,8 +244,7 @@ public class RubyGapicSurfaceTransformer implements ModelToViewTransformer {
                         .types(ImmutableList.<ImportTypeView>of())
                         .build()))
             .build();
-    List<String> modules = namer.getApiModules();
-    modules = modules.subList(0, modules.size() - 1);
+    List<String> modules = namer.getTopLevelApiModules();
     return CredentialsClassFileView.newBuilder()
         .outputPath("lib" + File.separator + namer.getCredentialsClassImportName() + ".rb")
         .templateFileName(CREDENTIALS_CLASS_TEMPLATE_FILE)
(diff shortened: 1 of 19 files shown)


src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceTransformer.java
```diff
@@ -419,6 +409,11 @@ private String topLevelPackagePath(SurfaceNamer namer) {
     return Joiner.on(File.separator).join(paths);
   }
 
+  private String topLevelNamespace(SurfaceNamer namer) {
```
geigerj: There are several places in this file (e.g., `generateTopLevelIndexView`) where we're calling `x.subList(0, x.size() - 1)`, seemingly all to get the topLevelNamespace. This logic should be consolidated.

src/main/java/com/google/api/codegen/transformer/ruby/RubySurfaceNamer.java
```diff
@@ -332,6 +332,19 @@ public String getTopLevelIndexFileImportName() {
     return Joiner.on(File.separator).join(newNames.toArray());
   }
 
+  @Override
+  public String getCredentialsClassImportName() {
+    List<String> parts = getApiModules();
+    // Place credentials in top-level namespace.
+    parts = parts.subList(0, parts.size() - 1);
```
geigerj: Another instance of popping the last module that ideally should be consolidated into common logic.

src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceTransformer.java
landrito: Done.

src/main/java/com/google/api/codegen/transformer/ruby/RubySurfaceNamer.java
landrito: Done.

PR #1427, opened 2017-07-24
Fix broken Travis build

### LINKED ISSUE #1424: Drop oracle jdk7 from CI?
> oraclejdk7 is unfortunately no longer available in Trusty after our latest updates, since we can no longer support oraclejdk7 on our recent build images due to Oracle's withdrawal. See: http://www.webupd8.org/2017/06/why-oracle-java-7-and-6-installers-no.html
>
> https://github.com/travis-ci/travis-ci/issues/7019#issuecomment-316769716

It seems like we should drop oraclejdk7. Is there any issue with requiring JDK 8 and above?

eoogbe:
Fixes Travis CI error seen in travis-ci/travis-ci#7884 and travis-ci/travis-ci#7019.

Closes #1424

Note: the workaround is to remove Java 7 from Travis.


.travis.yml
```diff
@@ -6,8 +6,6 @@ env:
     - GRADLE_OPTS="-Xmx512m"
 jdk:
   - oraclejdk8
-  - oraclejdk7
-  - openjdk7
```
garrettjonesgoogle: I'd prefer not to remove Java 7 builds while we still have Java 7 as our minimum required version of java.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,4 +1,4 @@
-#Tue Jul 18 22:46:11 EDT 2017
+#Mon Jul 24 13:38:01 PDT 2017
```
garrettjonesgoogle: Revert this since it isn't a meaningful change

.travis.yml
eoogbe: That appears to be the workaround recommended. In that case, I should close this.

PR #1428, opened 2017-07-25
Partial veneer for PHP

michaelbausor:
cc @dwsupplee @jdpedrie

This PR splits the generated GAPIC client into two files:
- `<API>Gapic`: the implementation
- `<API>Client`: an empty class which extends `<API>Gapic` and can be used to hold any partial veneer methods

An update to the Vision API is shown in this PR: https://github.com/GoogleCloudPlatform/google-cloud-php/pull/600


src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
```diff
@@ -59,7 +60,8 @@
   private final FileHeaderTransformer fileHeaderTransformer =
       new FileHeaderTransformer(new PhpImportSectionTransformer());
 
-  private static final String XAPI_TEMPLATE_FILENAME = "php/main.snip";
+  private static final String API_TEMPLATE_FILENAME = "php/main.snip";
```
shinfan: I feel the snip name is misleading, maybe something like partial_veneer.snip?

src/test/java/com/google/api/codegen/testdata/php_main_library.baseline
```diff
@@ -84,2270 +84,9 @@ use Google\Tagger\V1\AddTagRequest;
...
- *     $libraryServiceClient->close();
- * }
- * ```
- *
- * Many parameters require resource names to be formatted in a particular way. To assist
- * with these names, this class includes a format method for each type of name, and additionally
- * a parse method to extract the individual identifiers contained within names that are
- * returned.
- * @experimental
+ * {@inheritdoc}
```
shinfan: What is `inheritdoc`?

src/main/java/com/google/api/codegen/transformer/php/PhpSurfaceNamer.java
```diff
@@ -124,6 +125,11 @@ public String getFullyQualifiedApiWrapperClassName(GapicInterfaceConfig interfac
   }
 
   @Override
+  public String getApiWrapperClassImplName(InterfaceConfig interfaceConfig) {
+    return publicClassName(Name.upperCamel(getInterfaceName(interfaceConfig), "Gapic"));
```
shinfan: I feel "Gapic" does not reflect what the class is, maybe something like "ClientImp" ?

src/test/java/com/google/api/codegen/testdata/php_main_library.baseline
michaelbausor: This is a phpdoc tag that says to use the documentation from the parent class, in this case the client implementation.

src/main/java/com/google/api/codegen/transformer/php/PhpSurfaceNamer.java
michaelbausor: Yes, I was uncertain about this. I wanted to avoid `ClientImpl`, but didn't have many better ideas. `GapicClient` maybe?

shinfan: `GapicClient` sounds good to me

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
michaelbausor: Updated both main and main_impl

src/main/java/com/google/api/codegen/transformer/php/PhpSurfaceNamer.java
michaelbausor: Done

PR #1430, opened 2017-07-26
Fix ResourceName != UntypedResourceName runtime test failure bug

vam-google:
When a message expects any resource name, and actual resource name is used, in `Assert.equals()` they have type mismatch, because `getResourceNameAsName()` always returns `UntypedResourceName`, while equals implementation for resource names checks for `instanceof`.

### DIFF
--- a/src/main/java/com/google/api/codegen/config/FieldConfig.java
+++ b/src/main/java/com/google/api/codegen/config/FieldConfig.java
@@ -212,8 +212,12 @@ public abstract class FieldConfig {
   public boolean requiresParamTransformation() {
     return getResourceNameConfig() != null
         && getMessageResourceNameConfig() != null
-        && getMessageResourceNameConfig().getResourceNameType() != ResourceNameType.ANY
         && !getResourceNameConfig().equals(getMessageResourceNameConfig());
+  }
+
+  public boolean requiresAnyParamTransformation() {
+    return getMessageResourceNameConfig() != null
+        && getMessageResourceNameConfig().getResourceNameType() == ResourceNameType.ANY;
   }
 
   public FieldConfig getMessageFieldConfig() {
--- a/src/main/java/com/google/api/codegen/config/FieldConfig.java
+++ b/src/main/java/com/google/api/codegen/config/FieldConfig.java
@@ -215,7 +215,7 @@ public abstract class FieldConfig {
         && !getResourceNameConfig().equals(getMessageResourceNameConfig());
   }
 
-  public boolean requiresAnyParamTransformation() {
+  public boolean requiresParamTransformationFromAny() {
     return getMessageResourceNameConfig() != null
         && getMessageResourceNameConfig().getResourceNameType() == ResourceNameType.ANY;
   }
--- a/src/main/java/com/google/api/codegen/transformer/StaticLangApiMethodTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/StaticLangApiMethodTransformer.java
@@ -808,7 +808,7 @@ public class StaticLangApiMethodTransformer {
     String transformParamFunctionName = null;
     if (context.getFeatureConfig().useResourceNameFormatOption(fieldConfig)
         && fieldConfig.requiresParamTransformation()) {
-      if (!fieldConfig.requiresAnyParamTransformation()) {
+      if (!fieldConfig.requiresParamTransformationFromAny()) {
         transformParamFunctionName = namer.getResourceOneofCreateMethod(typeTable, fieldConfig);
       }
     }
--- a/src/main/java/com/google/api/codegen/transformer/StaticLangApiMethodTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/StaticLangApiMethodTransformer.java
@@ -808,7 +808,9 @@ public class StaticLangApiMethodTransformer {
     String transformParamFunctionName = null;
     if (context.getFeatureConfig().useResourceNameFormatOption(fieldConfig)
         && fieldConfig.requiresParamTransformation()) {
-      transformParamFunctionName = namer.getResourceOneofCreateMethod(typeTable, fieldConfig);
+      if (!fieldConfig.requiresAnyParamTransformation()) {
+        transformParamFunctionName = namer.getResourceOneofCreateMethod(typeTable, fieldConfig);
+      }
     }
 
     RequestObjectParamView.Builder param = RequestObjectParamView.newBuilder();
--- a/src/main/resources/com/google/api/codegen/java/test.snip
+++ b/src/main/resources/com/google/api/codegen/java/test.snip
@@ -283,7 +283,7 @@
       @end
     @else
       @if assert.hasActualValueTransformFunction
-        "$ NOT IMPLEMENTED: hasActualValueTransformFunction == true while hasExpectedValueTransformFunction == false";
+        "$ NOT IMPLEMENTED: hasActualValueTransformFunction == true while hasExpectedValueTransformFunction == false $";
       @else
         Assert.assertEquals({@assert.expectedValueIdentifier}, \
           actualRequest.{@assert.actualValueGetter}());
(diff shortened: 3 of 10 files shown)


src/main/resources/com/google/api/codegen/java/test.snip
```diff
@@ -272,9 +272,15 @@
 
   @join assert : test.asserts
     @if assert.hasExpectedValueTransformFunction
-      Assert.assertEquals({@assert.expectedValueTransformFunction}(\
-        {@assert.expectedValueIdentifier}), \
-        actualRequest.{@assert.actualValueGetter}());
+      @if assert.hasActualValueTransformFunction
```
michaelbausor: We don't handle case of having an actualValueTransform but not an expectedValueTransform. That seems fine if it isn't used, but we should at least comment that this case is unhandled

src/main/java/com/google/api/codegen/transformer/StaticLangApiMethodTransformer.java
```diff
@@ -808,7 +808,9 @@ private RequestObjectParamView generateRequestObjectParam(
     String transformParamFunctionName = null;
     if (context.getFeatureConfig().useResourceNameFormatOption(fieldConfig)
         && fieldConfig.requiresParamTransformation()) {
-      transformParamFunctionName = namer.getResourceOneofCreateMethod(typeTable, fieldConfig);
+      if (!fieldConfig.requiresAnyParamTransformation()) {
```
michaelbausor: Perhaps a comment here indicating why we don't need to apply the transformation in the Any case, because the setter will accept the supplied resource name without transformation.

michaelbausor:
Overall LGTM, two notes about adding comments

src/main/java/com/google/api/codegen/config/FieldConfig.java
```diff
@@ -212,10 +212,14 @@ public FieldConfig withResourceNameConfig(ResourceNameConfig resourceNameConfig)
   public boolean requiresParamTransformation() {
     return getResourceNameConfig() != null
         && getMessageResourceNameConfig() != null
-        && getMessageResourceNameConfig().getResourceNameType() != ResourceNameType.ANY
         && !getResourceNameConfig().equals(getMessageResourceNameConfig());
   }
 
+  public boolean requiresAnyParamTransformation() {
```
garrettjonesgoogle: The function name here implies something different from the implementation - maybe `requiresParamTransformationFromAny` would be clearer. This is indeed difficult because the "any" term can easily be interpreted differently from the proto meaning. 

Thinking about this further: the naming pattern "requires param transformation" is somewhat opaque, given the intent. I think I would favor a more direct approach with an enum (combine this method with `requiresParamTransformation`):

```
public TestAssertionTransformation getTestAssertionTransformation() {
  ...
}

public enum TestAssertionTransformation {
  CreateOneOfObject,
  ConvertToString
}
```

@michaelbausor what do you think?

src/main/resources/com/google/api/codegen/java/test.snip
garrettjonesgoogle: Or at least have the generator output text that doesn't compile, warning about the unhandled case.

src/main/java/com/google/api/codegen/config/FieldConfig.java
michaelbausor: I think that does make sense. But `TestAssertionTransformation` isn't quite right, because we also do the transformation in a non-test context when creating the oneof object (but NOT when converting to string for the ANY case).

Perhaps because the new way of handling the ANY case has introduced a difference between the test case and the non-test case, these should be handled separately, meaning that the logic should be pulled out of `FieldConfig` completely, and implemented where it is used in the `InitCodeTransformer` (for test assertion creation) and in the `StaticLangApiMethodTransformer` for method body generation.

vam-google: This is an older problem (the generated test cases were broken even without the recent resourcename changes, see the `ListStrings` method test case). So, even though the need to convert both is currently present only in test cases, it can potentially be relevant to non-test cases. Specifically when we talk about resource names there are 3 resource name types:

Regular resource name: `BookName` (for example)
Oneof resource name: `BookNameOneof`
AnyResource name: essentially resolves to `UntypedResourceName`.

All this 3 may have been initialized from exactly same string representation. Their `toString()` methods will return exactly same value, from the server side point of view they are equal. But as resourcename class instances all 3 will be distinct (**all** of the 3 possible pairs will be `x.equals(y) == false`). 

The fact that: `ResourceName instance (x) -> Server side representation (y) -> Resource Name instance (z)` conversion chain may (and often does) end up in` x.equals(z) == false` feels disturbing, and maybe the we should either change `equals` implementation of for those classes or at least introduce some other method in ResourceName interface like `equalsToResource()` or something like that.

src/main/resources/com/google/api/codegen/java/test.snip
vam-google: Yes, I didn't add the case because it was not possible to get to that case (plus it would make the snip file look worse). When you say comment what exactly do you mean (comments in snip file, if snip files support comments at all, or a comment in generated java code)?

src/main/java/com/google/api/codegen/config/FieldConfig.java
vam-google: **[EDIT]** Things get even more confusing for the cases when we compare `BookNameOneof`, initialized by same string representation, but one using `.from()` and another using `.formUntyped()` - we would have 2 oneof instances of same class, with same string representation, but still `x.equals(y) == false`

src/main/resources/com/google/api/codegen/java/test.snip
michaelbausor: I was imagining a comment in the snippet file (comments are supported, see https://github.com/googleapis/toolkit/blob/master/src/main/resources/com/google/api/codegen/java/main.snip#L275) but outputting a non-compiling string seems better.

src/main/java/com/google/api/codegen/config/FieldConfig.java
michaelbausor: Do you think it would be better to implement `.equals()` as reference equality, and rely on another method to check whether the string formats are equal? (Or just force the use of `toString()` to check that?) This could be outside the scope of the PR.

vam-google:
PTAL. 
The format of the non-compiling string is chosen to be similar to the output of `SufraceNamer.getNotImplementedString()`.

src/main/resources/com/google/api/codegen/java/test.snip
```diff
@@ -272,12 +272,22 @@
...
+      @else
+        Assert.assertEquals({@assert.expectedValueTransformFunction}(\
+          {@assert.expectedValueIdentifier}), \
+          actualRequest.{@assert.actualValueGetter}());
+      @end
     @else
-      Assert.assertEquals({@assert.expectedValueIdentifier}, \
-        actualRequest.{@assert.actualValueGetter}());
+      @if assert.hasActualValueTransformFunction
+        "$ NOT IMPLEMENTED: hasActualValueTransformFunction == true while hasExpectedValueTransformFunction == false";
```
michaelbausor: Nit: add trailing `$`

src/main/java/com/google/api/codegen/config/FieldConfig.java
vam-google: Honestly, I don't know I keep wondering how much unexpected complexity is in those resource names. 

I feel like forcing to string for equality check is somewhat against the resource name philosophy (as they exist to make things more static). At the same time that sill may be a reasonable approach here, to emphasize that resource names are only 1) client side 2) wrapper classes with utility methods and 3) the actual entity a user is working with is still a string.

src/main/resources/com/google/api/codegen/java/test.snip
vam-google: Nice catch, adding =)

garrettjonesgoogle:
LGTM

PR #1431, opened 2017-07-26
Disable rubocop in version index README

geigerj:
Due to variability in the product name length, it's difficult to place
line breaks appropriately in this file.

### DIFF
--- a/src/main/resources/com/google/api/codegen/ruby/version_index.snip
+++ b/src/main/resources/com/google/api/codegen/ruby/version_index.snip
@@ -36,8 +36,11 @@
 @end
 
 @private readmeModule(indexType, module, iterator, content)
+  @# rubocop:disable LineLength
   @##
   {@toComments(util.getDocLines(generateReadme(indexType, module)))}
+  @#
+  @# rubocop:enable LineLength
   @#
   {@simpleModule(indexType, module, iterator, content)}
 @end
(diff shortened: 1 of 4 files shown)


src/main/resources/com/google/api/codegen/ruby/version_index.snip
```diff
@@ -36,9 +36,12 @@
 @end
 
 @private readmeModule(indexType, module, iterator, content)
+  @# rubocop:disable LineLength
   @##
```
geigerj: Is this pre-existing `##` meaningful/conventional?

landrito: I don't think it's meaningful or conventional, simply aesthetic. FWIW I haven't found any references to it in a style guide.

geigerj: Ah, OK. Just curious.

geigerj:
@landrito I would love to see that. In other languages we use a formatter; it acutally looks like [rubocop has this capability](https://github.com/bbatsov/rubocop/blob/master/manual/formatters.md); hopefully we can use that or something similar. I filed https://github.com/googleapis/artman/issues/246.

src/main/resources/com/google/api/codegen/ruby/version_index.snip
```diff
@@ -36,9 +36,12 @@
 @end
 
 @private readmeModule(indexType, module, iterator, content)
+  @# rubocop:disable LineLength
```
blowmage: I'm pretty sure this will not give the intended results. This will add "rubocop:disable LineLength" at the beginning of the documentation and "rubocop:enable LineLength" at the end of the documentation.

PR #1444, opened 2017-08-02
all: make interface renaming work for go

pongad:
Go uses the interface name as parts of many other names
in the client.
This commit makes the rename available to SurfaceNamer implementations.
The renaming itself is not "threaded through" yet.
Since this change is already large, I plan to do this in a follow-up commit.

An alternative to this is to give GoSurfaceNamer the rename map.
This would put renaming logic in two places, but it will isolate the change
to Go and leave other languages alone.
Reviewers, please let me know if you'd rather I do that instead.

I tested that Go and Java googleapis clients remain the same after this change.

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/DynamicLangApiMethodTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/DynamicLangApiMethodTransformer.java
@@ -128,7 +128,7 @@ public class DynamicLangApiMethodTransformer {
 
     apiMethod.packageName(namer.getPackageName());
     apiMethod.packageHasMultipleServices(packageHasMultipleServices);
-    apiMethod.packageServiceName(namer.getPackageServiceName(context.getInterface()));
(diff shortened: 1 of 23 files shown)


pongad:
@garrettjonesgoogle I'll add only you as reviewer for now for MVVM stuff. Please let me know if you think the alternative approach I mention in commit msg is better. When we agree on the approach, I'll add others to make sure I don't break their stuff.

garrettjonesgoogle:
I'm concerned about the conflicts between this and Andrea's branch. cc @andreamlin

pongad:
@garrettjonesgoogle @andreamlin Ping. Can this work go ahead now?

garrettjonesgoogle:
Not yet. @andreamlin is working in https://github.com/googleapis/toolkit/tree/discogapic if you want to follow.

pongad:
superseded by #1654

PR #1445, opened 2017-08-03
Python: Change samples to use dictionaries rather than instantiated protos.

landrito:
Generated unit and system tests are working for monitoring with these changes.

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/py/PythonGapicSurfaceTestTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/py/PythonGapicSurfaceTestTransformer.java
@@ -151,37 +151,45 @@ public class PythonGapicSurfaceTestTransformer implements ModelToViewTransformer
 
   private List<TestCaseView> createTestCaseViews(GapicInterfaceContext context) {
     ImmutableList.Builder<TestCaseView> testCaseViews = ImmutableList.builder();
-    SymbolTable testNameTable = new SymbolTable();
-    for (Method method : context.getSupportedMethods()) {
-      GapicMethodContext methodContext = context.asRequestMethodContext(method);
-      ClientMethodType clientMethodType = ClientMethodType.OptionalArrayMethod;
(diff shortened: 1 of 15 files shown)


src/main/java/com/google/api/codegen/transformer/py/PythonGapicSurfaceTestTransformer.java
```diff
@@ -151,37 +151,45 @@ public PythonGapicSurfaceTestTransformer(
...
-              .initObjectType(methodContext.getMethod().getInputType())
-              .suggestedName(Name.from("request"))
-              .initFieldConfigStrings(methodContext.getMethodConfig().getSampleCodeInitFields())
-              .initValueConfigMap(InitCodeTransformer.createCollectionMap(methodContext))
-              .initFields(FieldConfig.toFieldIterable(fieldConfigs))
-              .outputType(initCodeOutputType)
-              .fieldConfigMap(FieldConfig.toFieldConfigMap(fieldConfigs))
-              .valueGenerator(valueGenerator)
-              .build();
+    // There are situations where a response type name has a collision with a response type name
```
eoogbe: If it's just the response type names that collide, you can save those before creating the test case views, rather than duplicating everything twice. Something like:
```
for (Method method : context.getSupportedMethods()) {
  context.getModelTypeTable().getAndSaveNicknameFor(method.getOutputType());
}

SymbolTable testNameTable = new SymbolTable();
for (Method method : context.getSupportedMethods()) {
  GapicMethodContext methodContext = context.asRequestMethodContext(method);
  // create test case views...
```

landrito: Trying that didn't resolve the issue. The issue is not only isolated to response type names. If there is a collision between types found in init code of two test cases, the first test case will become invalid. 

I updated the comment to reflect the fact that collisions can occur with types seen in init code as well by making the comment more general.

PR #1448, opened 2017-08-03
build.gradle: stop using << in tasks

pongad:
The << operation is deprecated. Gradle docs suggest we use doLast instead.


garrettjonesgoogle:
LGTM

PR #1449, opened 2017-08-03
go: un-markdown the docs

pongad:
Since Go comments are rendered plain text, most elements are rendered
with no formatting.

Lists are indented. Godoc will render these in monospace font and preserve
whitespaces so that they look tabular.


build.gradle
```diff
@@ -66,6 +66,7 @@ ext {
     commonsLang: 'org.apache.commons:commons-lang3:3.4',
     commonsCli: 'commons-cli:commons-cli:1.3',
     snakeyaml: 'org.yaml:snakeyaml:1.16',
+    commonmark: 'com.atlassian.commonmark:commonmark:0.9.0',
```
pongad: Commonmark is licensed under BSD 2 clause. I'm pretty sure we're OK to use it, but do we need to replicate the BSD notice or get approval, etc?

src/test/java/com/google/api/codegen/testdata/go_main_library.baseline
```diff
@@ -148,14 +148,14 @@ type Client struct {
...
+//   resources, named bookShelves/*
 //
-// - Each Shelf has a collection of [Book][google.example.library.v1.Book]
-//   resources, named `bookShelves/*/books/*`
+//   Each Shelf has a collection of [Book][google.example.library.v1.Book]
+//   resources, named bookShelves/*/books/*
 //
-// Check out [cloud docs!](/library/example/link).
-// This is [not a cloud link](http://www.google.com).
+// Check out cloud docs!.
```
jba: I think the Go docs would be poorer for the lack of these URLs. What if we render them as

*text* (at *URL*)

src/test/java/com/google/api/codegen/testdata/go_main_library.baseline
```diff
@@ -324,8 +324,8 @@ func (c *Client) DeleteShelf(ctx context.Context, req *librarypb.DeleteShelfRequ
 }
 
 // MergeShelves merges two shelves by adding all books from the shelf named
-// `other_shelf_name` to shelf `name`, and deletes
-// `other_shelf_name`. Returns the updated shelf.
+// other_shelf_name to shelf name, and deletes
+// other_shelf_name. Returns the updated shelf.
```
garrettjonesgoogle: I feel like stripping out the backticks and not replacing them with anything produces potential ambiguities in the text.

jba: Yes, but that's a pretty theoretical concern. 

The Go doc generator does no interpretation of any markdown-like characters (with the exception of indentation), and the convention is to just write identifiers and other inline bits of code without any decoration. So on the whole this will make the Go doc much more readable to Go programmers.

src/test/java/com/google/api/codegen/testdata/go_main_library.baseline
pongad: Done

src/main/java/com/google/api/codegen/util/ErrorMarkdownVisitor.java
```diff
@@ -0,0 +1,175 @@
...
+/**
+ * A dummy implementation of {@code Visitor} that always throws Exception.
+ *
+ * <p>The exception is always {@code UnimplementedRenderException}. We use this instead of {@link
+ * UnsupportedOperationException} so we we don't confuse it with exceptions we get trying to add to
+ * an unmodifiable list, etc.
+ */
+public class ErrorMarkdownVisitor implements Visitor {
+
+  private static class UnimplementedRenderException extends RuntimeException {
```
pongad: Should we fallback to copying text verbatim if we see a kind of markdown code we don't support?
Pros:
- If a markdown uses a new feature, we don't crash. We'll generate weird-looking docs, but that doesn't break users
Cons:
- Generating a lot of code, we might not look hard enough to see that some docs are now in markdown format and we never realize we need to support more Node types.

jba: I would copy verbatim, but issue a log.Warning, if that makes sense.

build.gradle
pongad: We don't need to do anything here. We don't actually distribute the dependency; the users pull it at build time.

pongad:
@jba PTAL. I think you need to select "approve" from the file tab, otherwise GH won't let me merge.

PR #1463, opened 2017-08-09
Add endpoint override example

saicheems:
Resolves https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2224


src/test/java/com/google/api/codegen/testdata/java_main_no_path_templates.baseline
```diff
@@ -81,6 +81,21 @@ import javax.annotation.Generated;
  *     NoTemplatesApiServiceClient.create(noTemplatesApiServiceSettings);
  * </code>
  * </pre>
+ *
+ * <pre>
```
garrettjonesgoogle: Add a note before this that this sample is for customizing the endpoint. Also, for the prior one, add a note about that sample being for customizing credentials.

saicheems: Done

PR #1465, opened 2017-08-10
C# generation of partial LRO/rerouted gRPC classes

chrisdunelm:


src/main/resources/com/google/api/codegen/csharp/gapic_client.snip
```diff
@@ -967,3 +969,39 @@
...
+            public partial class {@api.grpcTypeNameInner}
+            {
+                /// <summary>
+                /// Creates a new instance of <see cref="{@rerouted.typeName}"/> using the same call invoker as this client.
+                /// </summary>
+                /// <returns>A new {@rerouted.typeName} for the same target as this client.</returns>
+                public virtual {@rerouted.typeName} {@rerouted.getMethodName}() => new {@rerouted.typeName}(CallInvoker);
+            }
+        }
+
```
jskeet: Remove the extra whitespace here and above?

chrisdunelm: These appear to be needed to get the output formatting correct (ie one blank line).
Don't ask me why :(
I had a bit of a frustrating half hour figuring this out.

chrisdunelm:
Yes, that's the plan. I've verified pubsub and speech are both correctly generated.

PR #1468, opened 2017-08-11
DO NOT SUBMIT: go: hide SetGoogleClientInfo in gapic-only clients

### LINKED ISSUE #1467: go: hide SetGoogleClientInfo if there's no handwritten layer

pongad:
If the client has no corresponding handwritten client,
there's no reason for this method to be exported.

Fixes #1467.

Don't submit yet. It will break cloud.google.com/go/errors.
We have to set the domain_layer_location in gapic config.
That, in turn, waits for GoogleCloudPlatform/google-cloud-go#674.


pongad:
This is not an issue for the microgenerator. The microgenerator will never by itself export `setGoogleClientInfo`. Wherever the function is needed, google-cloud-go will create a separate tool to export them.

PR #1469, opened 2017-08-11
go: fix stutters in doc comments

### LINKED ISSUE #1439: Stutter in generated comments
In some proto files, the comments in the service declaration already begin with the service or method name (in a way reminiscent of Go style).  The generated Go GAPIC seems to assume that  the comments start with verbs, so it always prepends the identifier name. This results in stutter for these cases. For example, the proto
```
 // Create is used to ....
rpc Create(...)...
```
results in Go code
```
// Create create is used to ...
func (c *Client) Create(...)...
```

I suggest that instead we implement something along the lines of the following pseudo-code:
```
if lower(proto.comment.word[0])==lower(method.name) {
   gapic.comment = method.name + proto.comment.word[1:]
}
```

Do other languages have a similar issue?

pongad:
Fixes #1439.


src/main/java/com/google/api/codegen/transformer/go/GoSurfaceNamer.java
```diff
@@ -100,9 +100,31 @@ public String getLongRunningOperationTypeName(ModelTypeTable typeTable, TypeRef
...
+  }
+
+  /**
+   * Docs in Go usually starts with the name of the method. Prepend the method name if the name
+   * isn't already the first word.
+   */
+  @VisibleForTesting
+  String putDocMethodName(String methodName, String doc, VisibilityConfig visibility) {
+    doc = doc.trim();
+    String firstWord = doc.trim();
```
jba: I'm confused by this. According to https://docs.oracle.com/javase/7/docs/api/index.html?java/lang/String.html, trim just removes whitespace, so calling it a second time won't do anything. 

Maybe just declare firstWord, and loop over doc below?

src/main/java/com/google/api/codegen/transformer/go/GoSurfaceNamer.java
```diff
@@ -100,9 +100,31 @@ public String getLongRunningOperationTypeName(ModelTypeTable typeTable, TypeRef
...
+    for (int i = 0; i < firstWord.length(); i++) {
+      if (Character.isWhitespace(firstWord.charAt(i))) {
+        firstWord = firstWord.substring(0, i);
+        break;
+      }
+    }
+    if (firstWord.equalsIgnoreCase(methodName)) {
+      return doc;
```
vchudnov-g: Shouldn't we upperCamel the first word here, since we're not sure the casing is correct?

src/main/java/com/google/api/codegen/transformer/go/GoSurfaceNamer.java
pongad: You're right. The second trim won't do anything. On the other hand, we can't just declare `firstWord`. The if-condition might never be satisfied (maybe `doc` is empty or it only has one word); in these cases, we set `firstWord` to `doc` itself.

src/main/java/com/google/api/codegen/transformer/go/GoSurfaceNamer.java
pongad: I don't think we should use `Name.upperCamel`. It constructs a `Name` object given an upper camel string, but we aren't sure what case the comment is. Let me know if the fix looks OK to you.

src/test/java/com/google/api/codegen/transformer/go/GoSurfaceNamerTest.java
```diff
@@ -49,4 +50,17 @@ public void testOperationName() {
...
+  @Test
+  public void testPutDocMethodName() {
+    GoSurfaceNamer namer = new GoSurfaceNamer("cloud.google.com/go/gopher/apiv1");
+    Truth.assertThat(namer.putDocMethodName("Create", "creates things", VisibilityConfig.PUBLIC))
+        .isEqualTo("Create creates things");
+    Truth.assertThat(namer.putDocMethodName("Create", "Creates things", VisibilityConfig.PUBLIC))
+        .isEqualTo("Create creates things");
+    Truth.assertThat(
+            namer.putDocMethodName(
+                "Create", "Create is used to create things", VisibilityConfig.PUBLIC))
```
vchudnov-g: How about also 
`"Create" vs "creAte is used to create things"` and 
`"CreateThing" vs "creatething is used to create things"`

src/main/java/com/google/api/codegen/transformer/go/GoSurfaceNamer.java
vchudnov-g: Argh, I can't reconstruct what I was thinking. I remember I was curious about what seemed like an inconsistency in the caller, but I can't find it now. I think `equalsIgnoreCase` makes more sense here, since it helps in case the comment has weird capitalization. Sorry for the flip-flop.

Also in case there is weird capitalization of the method in the comment, maybe instead of `return doc` we should set `doc = doc.substring(firstWord.length()).trim()` here and have the `return` below apply to both cases. I think that's a better consistency guarantee. See the additional test cases I suggested. WDYT?

pongad:
@vchudnov-g PTAL

src/main/java/com/google/api/codegen/transformer/go/GoSurfaceNamer.java
```diff
@@ -100,9 +100,31 @@ public String getLongRunningOperationTypeName(ModelTypeTable typeTable, TypeRef
...
-    return super.getDocLines(getApiMethodName(method, methodConfig.getVisibility()) + " " + text);
+    return super.getDocLines(
+        putDocMethodName(
+            method.getSimpleName(),
+            DocumentationUtil.getDescription(method),
+            methodConfig.getVisibility()));
+  }
+
+  /**
+   * Docs in Go usually starts with the name of the method. Prepend the method name if the name
```
vchudnov-g: s/starts/start/

PR #1474, opened 2017-08-14
Ruby: fix credentials for longrunning and tests for clients with no major version

landrito:


src/main/java/com/google/api/codegen/ruby/RubyUtil.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+package com.google.api.codegen.ruby;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.regex.Pattern;
+
+public class RubyUtil {
+  private static final String LONGRUNNING_PACKAGE_NAME = "Google::Longrunning";
+  private static final String RUBY_PACKAGE_SEPERATOR = "::";
```
geigerj: Aside: I'm pretty surprised MVVM doesn't provide a special construct for this. (No action required.)

src/main/java/com/google/api/codegen/viewmodel/testing/ClientInitParamView.java
```diff
@@ -0,0 +1,41 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen.viewmodel.testing;
+
+import com.google.auto.value.AutoValue;
+import javax.annotation.Nullable;
+
+@AutoValue
+public abstract class ClientInitParamView {
```
geigerj: Class-level comments, please.

src/main/java/com/google/api/codegen/ruby/RubyUtil.java
landrito: Ah you're right. Updated.

src/main/java/com/google/api/codegen/viewmodel/testing/ClientInitParamView.java
landrito: Done.

PR #1475, opened 2017-08-15
Ruby: fix tests and version matcher.

landrito:
Now matches this regex: https://github.com/googleapis/gax-ruby/blob/master/lib/google/gax/util.rb#L34

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
@@ -145,8 +145,8 @@ public class InitCodeTransformer {
       }
 
       String messageTypeName = null;
-      if (fieldType.isMessage() && !fieldType.isRepeated()) {
-        messageTypeName = methodContext.getTypeTable().getFullNameFor(fieldType);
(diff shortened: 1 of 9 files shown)


src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -145,8 +145,8 @@ public InitCodeContext createRequestInitCodeContext(
       }
 
       String messageTypeName = null;
-      if (fieldType.isMessage() && !fieldType.isRepeated()) {
-        messageTypeName = methodContext.getTypeTable().getFullNameFor(fieldType);
+      if (fieldType.isMessage()) {
```
geigerj: What does this change do?

landrito: This makes it such that repeated messages will have their message types tracked. Before, getting the full name of the fieldType it would return "Array" since the field is repeated. With this change the full name of the message type will be tracked. This allows for the assert view to convert the array of hashes to its corresponding message type for comparison.

PR #1476, opened 2017-08-16
all: set up for faster testing

pongad:
Previously, for each generated file, baseline test computes the view model
for all files in that language, renders one file for comparison,
then throws the rest away. This causes us to do O(n^2) amount of work.

This commit computes the view model for all files in a particular
language only once, then render all files for comparison.
A side effect is that baseline files are now concatenated.
I don't think this is a deal-breaker;
I don't currently see a good way to spread out the baselines
given the test infrastructure.

With this change Go test finishes in 2 seconds vs 3 seconds before.
This change is small but should be magnified when we port other
languages. It should be particularly impactful for languages with many
files like Python and Java.

To keep the change small, the old test is left untouched,
and I ported only Go.
After porting all languages, the old test can be deleted.

Note that the new GoCodeGeneratorTest has very few mentions of Go.
I think it's possible to declare all baseline tests across multiple
languages in one file in the future. This will reduce code duplication
and probably make things go faster as we can share/parallelize more work.

This commit also declares the output file of createProperties Gradle
task. This lets Gradle skip running tests completely if it determines
that code has not been changed.


pongad:
@lukesneeringer I'm not sure who should review this from SVL. Can you nominate?

To reviewers: It's probably easier to `vimdiff GapicTestBase{,2}.java`

garrettjonesgoogle:
@pongad could you create a separate "DIFF PR" along the lines of the one that I did for the big GAX refactoring? Putting less work on reviewers allows the to review more easily :-)

pongad:
See diff here: https://github.com/googleapis/toolkit/pull/1492/files#diff-9fda9439de72feb2c59afcd34539daba

PR #1479, opened 2017-08-16
Divide streaming into client/server/bidi

shinfan:
The motivation behind this change is that currently user need to decide
the streaming type by themselves based on the server implementation. We should
only provide one streaming type in the client for better user
experience.

Corresponding GAX PR: https://github.com/googleapis/gax-java/pull/350

### DIFF
--- a/src/main/resources/com/google/api/codegen/java/settings.snip
+++ b/src/main/resources/com/google/api/codegen/java/settings.snip
@@ -201,26 +201,28 @@
 
 @private methodMembers(xsettingsClass)
   @join settings : xsettingsClass.callSettings
-    @switch settings.type
-    @case "SimpleApiCallable"
-      private final SimpleCallSettings<{@settings.requestTypeName}, \
-          {@settings.responseTypeName}> {@settings.memberName};
-    @case "PagedApiCallable"
-      private final PagedCallSettings<{@settings.requestTypeName}, \
-          {@settings.responseTypeName}, \
-          {@settings.pagedListResponseTypeName}> {@settings.memberName};
-    @case "BatchingApiCallable"
-      private final BatchingCallSettings<{@settings.requestTypeName}, \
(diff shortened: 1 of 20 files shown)


src/main/resources/com/google/api/codegen/java/settings.snip
```diff
@@ -201,26 +201,28 @@
...
-          {@settings.responseTypeName}> {@settings.memberName};
-    @case "PagedApiCallable"
-      private final PagedCallSettings<{@settings.requestTypeName}, \
-          {@settings.responseTypeName}, \
-          {@settings.pagedListResponseTypeName}> {@settings.memberName};
-    @case "BatchingApiCallable"
-      private final BatchingCallSettings<{@settings.requestTypeName}, \
-          {@settings.responseTypeName}> {@settings.memberName};
-    @case "StreamingApiCallable"
+    @if settings.isStreaming
```
michaelbausor: Does the snippet language support fallthrough switch cases? If so, can we keep everything in one switch, and have the three streaming cases handled by the same block?

shinfan: Good point, will do.

shinfan: Btw it is weird that the snippet supports fallthough but it does not require the 'break' statement in each case to prevent fallthrough...

michaelbausor: Yeah.. so the snippet lang does support fallthrough? I suppose if so, it only supports when block is empty. But yeah, weird

shinfan: Okay I am actually not sure about this..let me try it out

shinfan: @michaelbausor No it does not support fallthrough...

PR #1480, opened 2017-08-17
Check request_object_method and client streaming incompatibility

### LINKED ISSUE #850: Request object method incompatible with streaming
It should be an error to specify request_object_method as true for a streaming method

eoogbe:
Closes #850


michaelbausor:
This change LGTM - we should confirm that we are not going to break generation for any of our existing APIs before merging.

eoogbe:
Fixed incompatible gapic configs in googleapis/googleapis#425

PR #1482, opened 2017-08-17
Replace fakebilling test resources with simple compute API

andreamlin:
Replacing all fakebilling test files with a very small working subset of the Compute API. The new Discogapic generated surface can be used to make real calls against Cloud Compute, which is helpful for manual testing.

~Blocked on https://github.com/googleapis/toolkit/pull/1481.~


andreamlin:
PTAL.

PR #1484, opened 2017-08-17
Fix test generation for latest GAX

michaelbausor:


garrettjonesgoogle:
Could you give a link to the corresponding GAX change (if there is one)?

michaelbausor:
There is no open change, but the relevant change to the GeneratedTest class was introduced in https://github.com/googleapis/gax-php/pull/92

PR #1485, opened 2017-08-17
Remove PHP parse methods

michaelbausor:
- Remove parse methods
- Rename format methods
- Make format methods instance (not static)
- Cache gapic version

cc @bshaffer - PTAL at the new format methods in the `php_test_library.baseline` file and let me know if this is an improvement, and suggest any possible changes


michaelbausor:
Gentle ping on this

src/test/java/com/google/api/codegen/testdata/php_client_impl_library.baseline
```diff
@@ -645,7 +570,7 @@ class LibraryServiceGapicClient
      * ```
      * try {
      *     $libraryServiceClient = new LibraryServiceClient();
-     *     $formattedName = LibraryServiceClient::formatShelfName("[SHELF_ID]");
+     *     $formattedName = $this->shelfPath("[SHELF_ID]");
```
garrettjonesgoogle: Why "path" instead of "name"? These represent resource names. While they have the form of a path, that doesn't mean that they can ever be used as an actual path as a subcomponent of a server URL.

michaelbausor: This was copying Ruby, based on feedback from DPE (see https://github.com/googleapis/toolkit/blob/master/src/main/java/com/google/api/codegen/transformer/ruby/RubySurfaceNamer.java#L82)

I like "name" as well, although I don't have strong feelings between "name" and "path" - I think "path" is descriptive in a way that "name" isn't.

@bshaffer @dwsupplee @jdpedrie WDYT?

michaelbausor: @bshaffer @dwsupplee @jdpedrie @vam-google Gentle ping, any thoughts?

bshaffer:
This is a huge improvement! Nice work. Looks like we are using the suffix `Path` in the methods to avoid conflicts? 

👍 !

michaelbausor:
@bshaffer I think using either the `Path` or `Name` suffix works - it is a question of which one is a better descriptor. Do you have a preference between the two?

michaelbausor:
Also, re parsing methods: what are your thoughts on providing a single `parsePath` (or `parseName`) method that would (i) accept a name in any of the formats supported by that client, and (ii) return an associative array of components? Bringing this up because of feedback from @dwsupplee and @jdpedrie that the parse methods are used by the manual layers.

bshaffer:
Sounds awesome, I'm all for consistent nomenclature and I believe _path_ is what the APIs use, but otherwise I have no preference. As for the `parsePath/parseName` method, that sounds great.

src/test/java/com/google/api/codegen/testdata/php_client_impl_library.baseline
dwsupplee: I agree with @garrettjonesgoogle, especially after taking a look at the documentation for services such as spanner and pubsub. The resource seems to always be referred to as "names", so it makes sense to me to match that here as well.

michaelbausor: Updated to `*Name`

michaelbausor:
Updated to `*Name` and added `parseName` method.

michaelbausor:
This PR currently makes the format/parse methods instance methods instead of static. However, looking at google-cloud-php code, this is likely to cause difficulties (e.g. here: https://github.com/GoogleCloudPlatform/google-cloud-php/blob/master/src/Spanner/SpannerClient.php#L186) because there is no instance directly available.

I am going to change the methods back to static methods.

BUT: what should we use in the method samples? It is valid to call static methods on an instance - should we encourage this in our samples? @bshaffer @dwsupplee @jdpedrie please weigh in

bshaffer:
@michaelbausor I agree there's no real way to make the refactoring to instance methods. However, if we have the instances available I'd prefer to call them there, to make the samples more readable.

dwsupplee:
> I am going to change the methods back to static methods.

giphy

There isn't really a reason we can't call `formatProjectName` in the underlying connection class, which has access to the instance. I think I would prefer that over statics, as it's a big +1 for testing.

michaelbausor:
+100 for Princess Bride GIFs.

Isn't there a problem for uses like these: https://github.com/GoogleCloudPlatform/google-cloud-php/blob/master/src/Spanner/Instance.php#L271
https://github.com/GoogleCloudPlatform/google-cloud-php/blob/master/src/Spanner/Database.php#L1520
https://github.com/GoogleCloudPlatform/google-cloud-php/blob/master/src/Spanner/Session/Session.php#L136

If we make them static, we can still encourage users to call them from the instance in our samples, as @bshaffer mentioned. As for testing - yes, that is a downside.

dwsupplee:
> +100 for Princess Bride GIFs.

😄 one of my favorite movies!

So, for those cases - what we could do is modify the connection class to expose access to the instance methods we need to touch. In fact, I've been debating the necessity of the `Connection` based abstractions lately. The original intent was to provide an avenue for providing REST/gRPC support to the veneer, but now as we're shifting that abstraction towards the GAPIC layer I don't know what value it provides 🤔.

Anyways, I don't want to get in between moving forward with this work (as rethinking the connection layer will be a more substantial task). You are right that @bshaffer's point stands.

michaelbausor:
@dwsupplee That makes sense as something to think about moving to in future - I agree that it is weird to have a `connection` abstraction, and then to directly access the Gapic classes for format/parse methods as we do in Spanner.

I think we should move forward with this change, keeping the methods as static for now, but direct users towards instance usage.

I have created a WIP PR that includes a refresh of the generated clients, and the required updates in Spanner to support these changes: https://github.com/GoogleCloudPlatform/google-cloud-php/pull/673

@vam-google @bshaffer This is ready for final review.

PR #1486, opened 2017-08-18
update a bunch of dependencies

pongad:


build.gradle
```diff
@@ -58,34 +58,30 @@ ext {
   // Shortcuts for libraries we are using
   libraries = [
     // General
-    guava: 'com.google.guava:guava:19.0',
+    guava: 'com.google.guava:guava:23.0-android',
```
garrettjonesgoogle: why `-android`?

pongad: 23.0 is for Java 8, Java 7 and android needs to use 23.0-android. Added comment to the file

chrisdunelm:
C# part lgtm.

PR #1490, opened 2017-08-18
Add ResourceName methods to *HttpRequest classes

andreamlin:
Request objects, e.g. CreateTopicRequest, now define methods like getTopicAsTopicName so that generated clients like TopicAdminClient can use ResourceNames.

All DiscoGapic request and message classes have no compile errors.


andreamlin:
PTAL.

src/main/java/com/google/api/codegen/config/DiscoveryField.java
```diff
@@ -72,6 +73,13 @@ public String getFullName() {
   }
 
   @Override
+  public TypeName getTypeName(ImportTypeTable typeTable) {
+    return typeTable
+        .getTypeTable()
+        .getTypeName(((SchemaTypeTable) typeTable).getFullNameFor(schema));
```
garrettjonesgoogle: This has an odd look - I see "type table" referenced three times in this expression..

src/main/java/com/google/api/codegen/discogapic/transformer/java/JavaDiscoGapicRequestToViewTransformer.java
```diff
@@ -107,21 +117,46 @@ public JavaDiscoGapicRequestToViewTransformer(
...
-            JavaFeatureConfig.newBuilder().enableStringFormatFunctions(false).build());
-
-    for (Method method : context.getDocument().methods()) {
-      SchemaTransformationContext requestContext =
-          SchemaTransformationContext.create(method.id(), context.getSchemaTypeTable(), context);
-      StaticLangApiMessageView requestView = generateRequestClass(requestContext, method);
-      surfaceRequests.add(generateRequestFile(requestContext, requestView));
+    DiscoApiModel model = new DiscoApiModel(document);
+
+    for (InterfaceModel apiInterface : model.getInterfaces(productConfig)) {
```
garrettjonesgoogle: I think this function could benefit from being split up more

andreamlin: stuffed some implementation details in a private method.

src/main/java/com/google/api/codegen/config/DiscoveryField.java
andreamlin: less tables now

andreamlin:
made some code a bit more readable
PTAL.

PR #1499, opened 2017-08-23
build: add target to create jar

pongad:
Currently artman invokes CodeGeneratorTool through gradle command.
A lot of time is wasted as Gradle repeatedly resolve the workspace.
The added command instead creates one JAR,
which can be repeatedly used.


build.gradle
```diff
@@ -128,6 +128,18 @@ task discoJar(type: com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar)
   exclude 'META-INF/*.RSA'
 }
 
+task gapicJar(type: com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar) {
```
pongad: Any preference between this and a JAR without main class attribute. Instead of `java -jar gapic.jar`, we call `java -cp toolkit.jar com.google.api.codegen.CodeGeneratorTool`.

Pros: We have a generic JAR to use with any main class
Cons: If we change the path of the main class, we need to also make corresponding change to artman.

garrettjonesgoogle: Maybe name the task more generally - `fatJar`. 

I have no opinion on with/without main class attribute (yet).

ethanbao: +1 on calling it fatJar

pongad: I think whether or not we have main class should influence the name of the task. If we create an executable jar that runs `CodeGeneratorTool` the task name should "document" that the JAR runs code generator. Otherwise, it's like saying a task for compiling a C program should be called "ELF binary".

If we don't include the main method and require the user to provide the name of the main class, then I think "fatJar" makes sense.

pongad:
@garrettjonesgoogle @ethanbao PTAL

PR #1503, opened 2017-08-23
Python: Add api and types docs and refactor index.

landrito:


landrito:
@lukesneeringer can you take a look at the baselines to make sure they are correct?

landrito:
@eoogbe can you take a look at the changes to the generator?

PR #1505, opened 2017-08-24
ruby discovery: cache name map

pongad:
During tests, we previously open and parse YAML file for name mapping
every time.
This commit caches the name map.
Ruby discovery testing time reduces from ~13 seconds on my machine
to ~2 seconds.


saicheems:
Thanks for doing this!

PR #1508, opened 2017-08-24
Python: update metadata.

landrito:
Updates #1320:

Does the following

- Remove `requirements.txt` entirely.
- Remove `oauth2client` from the dependency list in `setup.py`; add `google-auth` and `requests`.
- Remove dependencies on `proto-google-cloud-*`.
- Add `google.cloud.proto`, `google.cloud.proto.{api}` to `namespace_packages`.
- Add Python 3.6 to the list of tested Python versions.

Need to be done:

- [x] Update the auth version in googleapis.
 - https://github.com/googleapis/googleapis/pull/435


src/main/java/com/google/api/codegen/transformer/py/PythonPackageMetadataTransformer.java
```diff
@@ -198,6 +203,50 @@ private ViewModel generateMetadataView(
...
+            "google-gax", packageConfig.gaxVersionBound(TargetLanguage.PYTHON)));
+    dependencies.add(
+        PackageDependencyView.create(
+            "google-auth", packageConfig.authVersionBound(TargetLanguage.PYTHON)));
+    dependencies.add(
+        PackageDependencyView.create("requests", VersionBound.create("2.18.4", "3.0dev")));
+    return dependencies.build();
+  }
+
+  private boolean hasLongrunningMethods(Model model, GapicProductConfig productConfig) {
```
eoogbe: This appears to be unused.

src/main/java/com/google/api/codegen/transformer/py/PythonPackageMetadataTransformer.java
```diff
@@ -267,6 +316,8 @@ private OptionalArrayMethodView createExampleApiMethodView(GapicMethodContext co
         namespacePackages.add(subPackage);
       }
     }
+    namespacePackages.add("google.cloud.proto");
```
eoogbe: This adds `google.cloud.proto` but not `google.cloud.gapic`.

landrito: Google.cloud.gapic was switched over to the new path "google.cloud.api_vx.gapic" and that should be computed already.

landrito: Updated the top comment. Sorry forgot to remove those after copy pasting.

src/main/java/com/google/api/codegen/transformer/py/PythonPackageMetadataTransformer.java
landrito: Oops thats for a different PR. Fixed.

PR #1509, opened 2017-08-24
Python: use google.api.core.Operation for long running operations.

landrito:
Updates #1320 

Here is the Operation class for reference:
https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/core/google/api/core/operation.py

### DIFF
--- a/src/main/java/com/google/api/codegen/viewmodel/DynamicLangXApiView.java
+++ b/src/main/java/com/google/api/codegen/viewmodel/DynamicLangXApiView.java
@@ -131,9 +131,6 @@ public abstract class DynamicLangXApiView implements ViewModel {
   @Nullable
   public abstract String gapicPackageName();
 
-  @Nullable
-  public abstract List<String> operationTypes();
-
   @Override
   public String resourceRoot() {
     return SnippetSetRunner.SNIPPET_RESOURCE_ROOT;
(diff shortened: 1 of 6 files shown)


src/test/java/com/google/api/codegen/testdata/py/python_setup_library.baseline
```diff
@@ -11,6 +11,7 @@ import sys
 
 install_requires = [
     'google-auth>=1.0.2, <2.0dev',
+    'google-cloud-core>=0.27.0, <1.0dev',
```
lukesneeringer: `< 0.28dev`

landrito: Done.

src/main/java/com/google/api/codegen/viewmodel/DynamicLangXApiView.java
```diff
@@ -131,6 +131,9 @@ public boolean hasMissingDefaultOptions() {
   @Nullable
   public abstract String gapicPackageName();
 
+  @Nullable
+  public abstract List<String> operationTypes();
```
eoogbe: What's this for?

landrito: Woops, Theres two implementations of operation in google-cloud-core. I accidentally implemented the wrong one and this is a byproduct. Removing.

landrito: Fixed.

landrito:
This is now obsolete.

PR #1510, opened 2017-08-24
Python: use a python file for the default client config

landrito:
Updates: #1320


src/main/java/com/google/api/codegen/viewmodel/DynamicLangXApiView.java
```diff
@@ -18,6 +18,7 @@
 import com.google.auto.value.AutoValue;
 import java.util.List;
 import javax.annotation.Nullable;
+import javax.validation.constraints.Null;
```
eoogbe: This appears to be unused

landrito: oops forgot to format! Should be fixed now.

PR #1511, opened 2017-08-24
Update toolkit with new exception types

shinfan:
GAX PR: https://github.com/googleapis/gax-java/pull/357

### DIFF
--- a/src/main/resources/com/google/api/codegen/java/test.snip
+++ b/src/main/resources/com/google/api/codegen/java/test.snip
@@ -102,7 +102,7 @@
           {@sampleMethodCallArgList(test.initCode.fieldSettings)});
       Assert.fail("No exception raised");
     } catch (InvalidArgumentException e) {
-      Assert.assertEquals(Status.INVALID_ARGUMENT.getCode(), ((GrpcStatusCode) e.getStatusCode()).getCode());
+      // Expected exception
     }
   }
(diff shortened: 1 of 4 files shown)


src/main/resources/com/google/api/codegen/java/test.snip
```diff
@@ -101,8 +101,8 @@
       client.{@test.clientMethodName}(\
           {@sampleMethodCallArgList(test.initCode.fieldSettings)});
       Assert.fail("No exception raised");
-    } catch (GrpcApiException e) {
-      Assert.assertEquals(Status.INVALID_ARGUMENT.getCode(), e.getStatusCode().getCode());
+    } catch (InvalidArgumentException e) {
+      Assert.assertEquals(Status.INVALID_ARGUMENT.getCode(), ((GrpcStatusCode) e.getStatusCode()).getCode());
```
garrettjonesgoogle: I don't feel that we need to actually check codes any more, since we are catching the exception that corresponds to the code. I think we should now just leave a comment like:

```
// expected exception
```

(This doesn't apply to the other type of test, which catches ExecutionException)

shinfan: Done

shinfan:
@garrettjonesgoogle PTAL

PR #1513, opened 2017-08-25
Add gapic config adviser

eoogbe:


eoogbe:
Actually, this PR is uncomfortably large. I'm going to split it into 8 parts.

PR #1515, opened 2017-08-25
Add initial gapic config adviser

eoogbe:
Adds only the language-settings rule.


src/main/java/com/google/api/codegen/advising/LanguageSettingsRule.java
```diff
@@ -0,0 +1,70 @@
...
+        LanguageTransformer.LANGUAGE_FORMATTERS.entrySet()) {
+      String language = entry.getKey();
+      LanguageSettingsProto languageSettingsProto =
+          configProto.getLanguageSettingsOrDefault(language, null);
+      if (languageSettingsProto == null) {
+        String formattedPackageName = entry.getValue().getFormattedPackageName(packageName);
+        messages.add(
+            String.format(
+                "Missing language_settings for key %s.%n%n"
+                    + "language_settings:%n"
```
geigerj: Maybe add "Did you mean:"

```
Did you mean:
  language settings: ...
  ...
```

As is, it's not immediately clear to me what the relationship between the advice and the warning is.

src/main/java/com/google/api/codegen/advising/Adviser.java
```diff
@@ -0,0 +1,55 @@
...
+  public void advise(Model model, ConfigProto configProto) {
+    for (AdviserRule rule : rules) {
+      if (suppressors.contains(rule.getName())) {
+        continue;
+      }
+
+      for (String message : rule.collectAdvice(model, configProto)) {
+        model
+            .getDiagCollector()
+            .addDiag(Diag.warning(SimpleLocation.TOPLEVEL, "(%s) %s", rule.getName(), message));
```
geigerj: Can we introduce a new `class GapicConfigLocation implements Location` instead of using `TOPLEVEL`? It would be really cool if we could eventually provide filename + line number in the warnings (not necessarily in this PR)

eoogbe: We can do that if we return a config source from the `MultiYamlReader` and change everywhere that uses it.

eoogbe:
@ethanbao @garrettjonesgoogle PTAL

src/main/java/com/google/api/codegen/advising/AdviserRule.java
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+package com.google.api.codegen.advising;
+
+import com.google.api.codegen.ConfigProto;
+import com.google.api.tools.framework.model.Model;
+import java.util.List;
+
+public interface AdviserRule {
+  String getName();
```
ethanbao: Explain the syntax of the rule name with examples?

ethanbao: I meant can we control the granularity for proto element?

eoogbe: How would we specify that on the command line?

ethanbao: rule-name@element-name (e.g. FooRule@example.Bar,example.Bak)

eoogbe: Ok, that functionality isn't there yet, but I think I can add it in.

ethanbao: You can extend it in your followup PR.

PR #1517, opened 2017-08-25
Replace Message return with ConfigSource in MultiYamlReader#read

eoogbe:
Needed to use line numbers in the gapic config adviser.


eoogbe:
@ethanbao @garrettjonesgoogle @geigerj PTAL

eoogbe:
@ethanbao Yes

PR #1520, opened 2017-08-25
Warn on oneofs in required_fields and parameters

### LINKED ISSUE #620: Language-independent one-of fixes

eoogbe:
Fixes #620


eoogbe:
@geigerj PTAL

PR #1522, opened 2017-08-28
ruby, python, node, c#: make tests faster

pongad:
This commit continues from #1476 and #1514.
Please see the former for more info on the speed up itself.

This reduces test execution time on my machine from ~38s
to ~27s.


pongad:
I tested this change by slightly modifying a few snippets and making sure tests fail at appropriate points.

Could language owners skim the concatenated baseline files to make sure it looks sane? I apologize for the large diffs :(

geigerj:
I think that makes sense. @lukesneeringer any objection?

src/main/java/com/google/api/codegen/gapic/MainGapicProviderFactory.java
```diff
@@ -72,8 +72,8 @@
   public static final String NODEJS = "nodejs";
   public static final String NODEJS_DOC = "nodejs_doc";
   public static final String PHP = "php";
-  public static final String PYTHON = "python";
-  public static final String PYTHON_DOC = "python_doc";
+  public static final String PYTHON = "py";
+  public static final String PYTHON_DOC = "py_doc";
```
lukesneeringer: Does PYTHON_DOC even need to exist anymore?

geigerj: We should probably kill this, since it is replaced by your plugin.

pongad: Done

pongad:
@lukesneeringer It is unfortunate but intentional. The lines are moved into one baseline file, not deleted. I unfortunately don't see a way to avoid this :(.

pongad:
Ping @lukesneeringer are you OK with this change?

pongad:
@chrisdunelm @geigerj I just got LGTM from Luke and resolved conflict. Could you please take another pass?

chrisdunelm:
Still LGTM )

PR #1523, opened 2017-08-28
Ruby: Deprecate service_path and port client constructor parameters

eoogbe:


src/test/java/com/google/api/codegen/testdata/ruby/ruby_main_longrunning.baseline
```diff
@@ -184,8 +178,8 @@ module Google
           )
         end
         @operations_stub = Google::Gax::Grpc.create_stub(
-          service_path,
-          port,
+          SERVICE_ADDRESS,
```
geigerj: We should to change these from frozen constants to ordinary attributes so that users are still able to mutate them if needed.

src/test/java/com/google/api/codegen/testdata/ruby/ruby_main_longrunning.baseline
```diff
@@ -83,10 +83,6 @@ module Google
       ALL_SCOPES = [
       ].freeze
 
-      # @param service_path [String]
```
geigerj: Let's deprecate these. We can remove when we declare GA. (See as example: https://github.com/googleapis/toolkit/pull/1387/files#diff-58544da0694fa700a3f27fc817ec47d4R160)

src/test/java/com/google/api/codegen/testdata/ruby/ruby_main_longrunning.baseline
eoogbe: They are only used in the constructor. I believe the suggestion was to override in a subclass or mixin.

geigerj: Ah, my mistake. You're right. Ignore that comment.

src/test/java/com/google/api/codegen/testdata/ruby/ruby_main_library.baseline
```diff
@@ -222,12 +218,13 @@ module Library
         if app_name || app_version
           warn "`app_name` and `app_version` are no longer being used in the request headers."
         end
+        if service_path || port
+          warn "`service_path` and `port` parameters will be removed"
```
geigerj: nit: can you change the wording here and above to "... are deprecated and will be removed"?

PR #1528, opened 2017-08-30
Fix Python packaging issues

### LINKED ISSUE #1129: Fix Python packaging issues
1. `setup.py`
   - [ ] use `io.open` over `open`.
   - [ ] use `Apache 2.0` instead of `apache-2.0`.
2. `setup.cfg`
   - [ ] Add [setup.cfg](https://github.com/GoogleCloudPlatform/google-auth-library-python/blob/master/setup.cfg) to enable wheel distribution. (We should additionally ensure packages are published as wheels).
3. `MANIFEST.in`
   - [ ] use `recursive-include google *.json *.proto` over `graft google` and `recursive-include *.json`.
   - [ ] remove unneeded `prune .tox` and `graft google`.

eoogbe:
Closes #1129


src/main/resources/com/google/api/codegen/py/setup.py.snip
```diff
@@ -38,7 +39,7 @@
       ],
       description='GAPIC library for the {@metadata.fullName}',
       include_package_data=True,
-      long_description=open('README.rst').read(),
+      long_description=io.open('README.rst', encoding='utf-8').read(),
```
lukesneeringer: This does not close the file handler. This matters because calls to `pkg_resources.get_distribution` will read this file.

Above the `setup(...)` call, we need:

```python
with io.open('README.rst', 'r', encoding='utf-8'): as readme_file:
    long_description = readme_file.read()
```

And then just use the `long_description` variable here.

eoogbe: Done.

eoogbe:
@lukesneeringer PTAL

src/main/resources/com/google/api/codegen/py/setup.py.snip
```diff
@@ -18,6 +18,9 @@
       @end
   ]
 
+  with io.open('README.rst', 'r', encoding='utf-8'): as readme_file:
```
lukesneeringer: This is a `SyntaxError` (note the two colons)

eoogbe: Done

PR #1531, opened 2017-08-31
Python: update config-generation to use new package path

### LINKED ISSUE #1502: Configgen: update Python package name
### What

Change the configgen to produce Python package names of the form `...{api}_{version}.gapic`

### Why

All configs in googleapis/googleapis use this newer form, but the generator produces the older form `{api}.{version}.gapic`

landrito:
Resolves #1502


landrito:
Added _ as a character in the point release matcher. This will be used for speech 1.1 point release.

src/main/java/com/google/api/codegen/configgen/transformer/LanguageTransformer.java
```diff
@@ -161,4 +161,21 @@ public String rewrite(String input) {
       return input.replaceAll(pattern, replacement);
     }
   }
+
+  private static class PythonLanguageFormatter implements LanguageFormatter {
+    private static final RewriteRule PYTHON_REWRITE_RULE =
```
eoogbe: This duplicates the `commonRewriteRules`. How about passing it in as a constructor parameter?

src/main/java/com/google/api/codegen/configgen/transformer/LanguageTransformer.java
```diff
@@ -161,4 +161,21 @@ public String rewrite(String input) {
...
+        new RewriteRule("^google(?!\\.cloud)", "google.cloud");
+
+    @Override
+    public String getFormattedPackageName(String packageName) {
+      packageName = PYTHON_REWRITE_RULE.rewrite(packageName);
+      List<String> names = Arrays.asList(packageName.split("\\."));
```
eoogbe: Replate with `Splitter.on(DEFAULT_PACKAGE_SEPARATOR).split(packageName)`

src/main/java/com/google/api/codegen/configgen/transformer/LanguageTransformer.java
```diff
@@ -161,4 +161,21 @@ public String rewrite(String input) {
...
+      if (VersionMatcher.isVersion(names.get(names.size() - 1))) {
```
eoogbe: Just trying to understand what this does: what's an example where this check would return true?

src/main/java/com/google/api/codegen/configgen/transformer/LanguageTransformer.java
landrito: Done.

src/main/java/com/google/api/codegen/configgen/transformer/LanguageTransformer.java
landrito: Done.

src/main/java/com/google/api/codegen/configgen/transformer/LanguageTransformer.java
landrito: This handles situations where the package name does not end with a major version such as with longrunning. I added a longrunning test to make sure this case is covered.

PR #1532, opened 2017-08-31
Update version matcher to see underscore as a point release.

landrito:


landrito:
Moved these changes to #1531

PR #1534, opened 2017-08-31
Support php_namespace option

michaelbausor:
This PR increases the protobuf version to 3.4 so that the php_namespace option is available.

@geigerj @lukesneeringer PTAL at the changes in the Ruby and Node baselines - they appear to be just updated documentation for the Any class.


geigerj:
LGTM for Ruby change

src/main/java/com/google/api/codegen/CodeGeneratorTool.java
```diff
@@ -31,6 +32,8 @@
 //        --output=$BASE
 public class CodeGeneratorTool {
   public static void main(String[] args) throws Exception {
+    System.out.println("Toolkit args:" + Arrays.toString(args));
```
geigerj: Remove this?

michaelbausor: Yes, done

PR #1535, opened 2017-09-01
Merge master

andreamlin:
Merge in the latest changes from master into discogapic branch.

The diff of test resource files (baselines and yamls) in this PR's discogapic branch compared to master is empty (with the exception of discogapic's simplecompute resources).
There is no change in discogapic branch's simplecompute baseline.


andreamlin:
PTAL

PR #1537, opened 2017-09-01
Python: change the import path of protos

landrito:
This change corresponds to an upcoming change in artman that will move the protos into the `google.cloud.example_v1.proto` path.

Going to explore changing the package name in the gapic yaml from `google.cloud.example_v1.gapic` -> `google.cloud.example_v1` in a subsequent PR.


landrito:
Thanks for the review. Will merge after https://github.com/googleapis/artman/pull/261 is approved.

PR #1538, opened 2017-09-01
Add InterfaceRule to gapic config adviser

eoogbe:


eoogbe:
@ethanbao @garrettjonesgoogle @geigerj PTAL

PR #1540, opened 2017-09-06
go: remove "experimental" doc for GA libraries

pongad:


pongad:
@jba would you like a test for this? It's awfully trivial I'm not even sure how to test it.

pongad:
The failing test seems to be a problem with Travis. The failure is caused by the program never finish compiling.

jba:
No need for a test.

PR #1542, opened 2017-09-07
Travis: add workaround for travis issue that causes buffer overflow for openjdk7.

landrito:


.travis.yml
```diff
@@ -12,3 +12,7 @@ script:
   ./gradlew check --info
 after_success:
   - bash <(curl -s https://codecov.io/bash)
+# Workaround for Travis CI issue 5227 causing buffer overflow in OpenJDK7.
+before_install:
+    - sudo hostname "$(hostname | cut -c1-63)"
```
pongad: forgive my ignorance, but how do we sudo when the first lines says `sudo: false`? Does the restriction only kick in after installation or something?

pongad:
Thank you so much for looking into this!

michaelbausor:
Agreed, thanks a lot!

Based on [this comment in issue 5227](https://github.com/travis-ci/travis-ci/issues/5227#issuecomment-165131913), should we be using the hosts addon?

landrito:
I think one of my PRs that I was using to test this fix got merged which is causing the out of date fix to get merged. This is why there are deletions in this PR.

@michaelbausor I switched it to use the addons/hosts workaround. Which looks like it works

.travis.yml
landrito: I honestly don't know. I switched to using the hosts/addons fix so we don't have to worry about sudo usage.

PR #1544, opened 2017-09-07
Ruby: Move ToC from version index to doc message

eoogbe:


geigerj:
As far as you know, this should not cause any problems if the proto module name is different from the GAPIC module name, right, for example `Google::Cloud::Logging::V2` vs `Google::Logging::V2`? (On second thought, I expect that doesn't matter as long as the `jsondoc` tool is able to put this on the right page.)

geigerj:
Note: updates https://github.com/GoogleCloudPlatform/google-cloud-ruby/issues/1181

eoogbe:
IIUC the toc.json is manually configured to get the right filename, so this shouldn't cause problems.

src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceDocTransformer.java
```diff
@@ -51,12 +63,12 @@ public RubyGapicSurfaceDocTransformer(GapicCodePathMapper pathMapper) {
...
     ImmutableList.Builder<ViewModel> surfaceDocs = ImmutableList.builder();
     for (ProtoFile file : new ProtoFileView().getElementIterable(model)) {
-      surfaceDocs.add(generateDoc(file, productConfig));
+      surfaceDocs.add(generateDoc(model, file, productConfig));
     }
     return surfaceDocs.build();
   }
 
-  private ViewModel generateDoc(ProtoFile file, GapicProductConfig productConfig) {
+  private ViewModel generateDoc(Model model, ProtoFile file, GapicProductConfig productConfig) {
```
geigerj: nit: it's unnecessary to pass `model` if you already have a `ProtoFile`; you can just call `file.getModel()`.

src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceDocTransformer.java
```diff
@@ -26,20 +29,29 @@
...
+import com.google.api.codegen.viewmodel.metadata.TocContentView;
+import com.google.api.codegen.viewmodel.metadata.TocModuleView;
+import com.google.api.tools.framework.model.Interface;
 import com.google.api.tools.framework.model.Model;
 import com.google.api.tools.framework.model.ProtoFile;
 import com.google.common.collect.ImmutableList;
 import java.util.List;
 
 public class RubyGapicSurfaceDocTransformer implements ModelToViewTransformer {
+  private static final int VERSION_MODULE_RINDEX = 1;
```
geigerj: Ideally this constant should not be defined twice (here and in `RubyGapicSurfaceTransformer`). Is there a common location it can be shared from?

eoogbe: Done

src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceDocTransformer.java
eoogbe: Done

src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceTransformer.java
```diff
@@ -344,19 +342,17 @@ private ViewModel generateTopLevelIndexView(Model model, GapicProductConfig prod
...
+      if (moduleName.equals(namer.getModuleServiceName())) {
         moduleViews.add(
             metadataTransformer
                 .generateReadmeMetadataView(model, productConfig, packageNamer)
-                .moduleName(apiModules.get(i))
+                .moduleName(moduleName)
                 .build());
-      } else if (i != moduleCount - VERSION_MODULE_RINDEX || includeVersionModule) {
-        moduleViews.add(SimpleModuleView.newBuilder().moduleName(apiModules.get(i)).build());
+      } else if (includeVersionModule || !moduleName.equals(namer.getModuleVersionName())) {
```
geigerj: Nice! This is more readable and seems more robust than checking an index, as well. 👍

geigerj: (Actually, I just realized it does the same thing under the hood in the namer. Still, I like the clearer intent.)

PR #1545, opened 2017-09-07
NodeJS: Allow for path mapper to accept versions with no digit after alpha or beta

landrito:


eoogbe:
I believe this duplicates the efforts of #1378

PR #1549, opened 2017-09-08
Ruby documentation fixes

geigerj:
Based on feedback from
  https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/1702

- Use `@overload` to prevent YARD from being confused when `@param` docs
  do not match the *args, **kwargs method signature
- Do not use *args, **kwargs when the true method signature is known
- End version index files in newline, per convention

### DIFF
--- a/src/test/java/com/google/api/codegen/testdata/ruby/ruby_doc_version_index_library.baseline
+++ b/src/test/java/com/google/api/codegen/testdata/ruby/ruby_doc_version_index_library.baseline
@@ -86,10 +86,10 @@ module Library
   #
   # Service comment may include special characters: <>&"+'@.
   #
+  # @param version [Symbol, String]
+  #   The major version of the service to be used. By default :v1
+  #   is used.
   # @overload
-  #   @param version [Symbol, String]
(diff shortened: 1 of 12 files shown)


src/test/java/com/google/api/codegen/testdata/ruby/ruby_doc_version_index_library.baseline
```diff
@@ -86,34 +86,34 @@ module Library
...
-  #   an updater_proc is supplied.
-  # @param client_config[Hash]
-  #   A Hash for call options for each method. See
-  #   Google::Gax#construct_settings for the structure of
-  #   this data. Falls back to the default config if not specified
-  #   or the specified config is missing data points.
-  # @param timeout [Numeric]
-  #   The default timeout, in seconds, for calls made through this client.
+  # @overload
+  #   @param version [Symbol, String]
```
landrito: `version` is not an overloaded parameter.

geigerj: Done.

PR #1550, opened 2017-09-09
Ruby: Fix isSource check in gRPC doc generation

eoogbe:


src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceDocTransformer.java
```diff
@@ -83,10 +84,22 @@ private ViewModel generateDoc(ProtoFile file, GapicProductConfig productConfig)
         fileHeaderTransformer.generateFileHeader(
             productConfig, ImportSectionView.newBuilder().build(), namer));
     doc.elementDocs(elementDocTransformer.generateElementDocs(typeTable, namer, file));
-    doc.modules(generateModuleViews(file.getModel(), productConfig, namer, file.isSource()));
+    doc.modules(generateModuleViews(file.getModel(), productConfig, namer, isSourceFile(file)));
     return doc.build();
   }
 
+  private boolean isSourceFile(ProtoFile file) {
```
geigerj: nit: let's call this `isSourceApiInterfaceFile` since this is not 100% equivalent logically to the `apiCompiler` `isSource` method.

PR #1551, opened 2017-09-10
Deprecate defaultBuilder in GAPIC java surface

shinfan:
This PR also marks `stub` classes and `getStub()` method as `@BetaApi`. The beta release level for stub is now hardcoded since it will stay beta even after the surface is switched to GA.


michaelbausor:
The changes LGTM, but is there an issue or something that discusses why we are making the changes in this PR and https://github.com/googleapis/gax-java/pull/370?

shinfan:
@michaelbausor this is a TODO item in the GAX GA blocker list. The motivation behind this is that we should be consistent of using either newBuilder or defaultBuilder.

PR #1553, opened 2017-09-11
NodeJS: Support protos being copied to gapic directory.

landrito:


landrito:
Thanks for the review! 
Will merge alongside https://github.com/googleapis/artman/pull/264.

src/test/java/com/google/api/codegen/testdata/nodejs/nodejs_package_library.baseline
```diff
@@ -27,7 +27,6 @@
   "dependencies": {
     "extend": "^3.0",
     "google-gax": "^0.7.0",
-    "google-proto-files": "^0.8.2",
```
lukesneeringer: Are we able to remove this altogether? No needed actually-common ones?

landrito: Hm. I think I was under the (incorrect) assumption that all of the protos would be copied over into the file. I need to look into how we can have the nodejs protobuf be able to understand that the proto file will be in a separate place than the common protos files.

eoogbe: Only the source protos get copied in. Dependencies don't.

lukesneeringer: We probably _can_ copy in imported protos, and that might be the easiest way.

eoogbe: We'd have to dig into the source protos. I think it's easier to keep google-proto-files as a dependency for google-common-protos. If an api happens to not need google-common-protos, then it shouldn't depend on google-proto-files. Otherwise, keep it around.

lukesneeringer: > We'd have to dig into the source protos.

protoc can totally do this for you. A `CodeGeneratorRequest` comes in with all of this nicely wrapped with a bow on top. (That said, I am not sure what solution @landrito landed on, and I am not picky.)

landrito:
Closing since this actually doesn't work based on my incorrect assumption that all protos were being copied.

PR #1555, opened 2017-09-12
Fix link to artman usage section

danoscarmike:


lukesneeringer:
This seems to already have been changed.

PR #1561, opened 2017-09-13
NodeJS: Support for new protobuf loading

landrito:
Also some small name changes to make things a little more readable.

This depends on:
https://github.com/googleapis/gax-nodejs/pull/146
https://github.com/googleapis/googleapis/pull/462

Tested this on my local machine using the dependant PRs above as well as deleting the cloud file directory from the google-proto-files node_module (to ensure that the protos were coming from the client proto directory) and the smoke test works for vision!


src/main/resources/com/google/api/codegen/nodejs/version_index.snip
```diff
@@ -93,17 +96,13 @@
...
-      @else
-        # This does not follow the import in main.snip since version index is not generated
-        # for unversioned Apis. This will be changed when top level indexes are supported
-        # for unversioned Apis.
-        root: require('path').join(__dirname, '..', '..', 'proto'),
-      @end
-      file: '{@stub.protoFileName}'
-    }]).{@stub.namespace});
+    extend({@version}Protos, gaxGrpc.loadObject(
+      # This does not follow the import in main.snip since version index is not generated
```
lukesneeringer: I do not understand what this comment means. (But that might be okay?)

src/test/java/com/google/api/codegen/testdata/nodejs/nodejs_doc_main_library.baseline
```diff
@@ -93,7 +95,7 @@ var ALL_SCOPES = [
  *
  * @class
  */
-function LibraryServiceClient(gaxGrpc, grpcClients, opts) {
+function LibraryServiceClient(gaxGrpc, loadedProtos, opts) {
```
lukesneeringer: `${deity}` bless you.

src/main/resources/com/google/api/codegen/nodejs/version_index.snip
landrito: This was an existing comment. Essentially says that the main.snip has some logic to determine how far back in the directory the proto files are. This file will always be located 1 directory away from the proto directory.

landrito:
The dependency version update will be to `google-proto-files` once we release a new package for it [here](https://github.com/googleapis/googleapis/blob/master/gapic/packaging/dependencies.yaml#L82). An update to grpc version is not necessary since the current version that is imported by gax already supports protobufjs v6 ([source](https://github.com/grpc/grpc/blob/master/src/node/src/protobuf_js_6_common.js)).

PR #1563, opened 2017-09-14
Remove a one-character indentation error.

lukesneeringer:
Yes, Sphinx is _that_ pedantic.

P. S. Also removes an inaccurate `Returns:` clause, since this is an `__init__` function (constructors in Python always return `None`).


lukesneeringer:
@landrito Feel free to merge yourself once you have reviewed.

PR #1564, opened 2017-09-14
Python: Fix missing service proto import

landrito:
### What
Service proto files that didn't use messages declared in that file would not be imported in the main client.

### Why
1) the type table was being given the nickname of the of the type rather than the full type.
2) the imports were being constructed after the grpc client stub was added to the type table. 

Commit [0316032](https://github.com/landrito/toolkit/commit/0316032e5d3ef65fe8456bb895e7c784bf199240) shows the problem.
Commit [82bd127](https://github.com/googleapis/toolkit/pull/1564/commits/82bd127275185c37d68ecde9d5f39208e7cdb65f) shows the fix.

cc @ethanbao


ethanbao:
What about types list in the normalized service config? Would that be used for this purpose?

landrito:
Hm I think thats a separate issue. Can you provide more context of that?

ethanbao:
I think I misunderstood the problem.

PR #1566, opened 2017-09-15
Ruby: Fix ToC links

eoogbe:
Updates GoogleCloudPlatform/google-cloud-ruby#1181


src/main/java/com/google/api/codegen/transformer/ruby/RubyPackageMetadataTransformer.java
```diff
@@ -122,28 +123,29 @@ public RubyPackageMetadataTransformer(PackageMetadataConfig packageConfig) {
...
-      String apiModule, RubyPackageMetadataNamer namer, String version) {
-    return generateTocContent("Data types for " + apiModule, namer, version, "Data Types");
+      String apiModule, RubyPackageMetadataNamer namer) {
+    return generateTocContent("Data types for " + apiModule, namer, "Data Types");
   }
 
   private TocContentView generateTocContent(
-      String description, RubyPackageMetadataNamer namer, String version, String clientName) {
+      String description, RubyPackageMetadataNamer namer, String clientName) {
+    String protoPath = packageConfig.protoPath();
```
geigerj: This seems fine, but as you mentioned in https://github.com/GoogleCloudPlatform/google-cloud-ruby/issues/1181, the PackageMetadata `protoPath` is a particularly unstable value that can change if the `srcProtoPath` changes. I think it would be more robust to use the real proto package for the service file in question, that is, `protoFile.getProto().getPackage()`.

geigerj: @eoogbe WDYT?

eoogbe: The package is the value of the package in the proto file, not the path to the proto. Using `protoFile.getProto().getPackage()` or `protoFile.getFullName()` yields the wrong value. I'm still unclear on how to access the path to the proto from the `ProtoFile`.

geigerj: >  package is the value of the package in the proto file, not the path to the proto

By convention (e.g., all protos in googleapis/googleapis) these are the same. As long as we are consistent in our use of the package rather than the path (we should be! there's no semantic value to the path) it is fine even in the case that the proto does not follow Google convention.

eoogbe: For the library tests, the package name is `google.example.library.v1` and the path is `google/library`. It's inconsistent for most of the tests.

Since this is forming a url, we need the correct path. I believe that comes from the path to the protos, not the package name. If that isn't the case, then I'll change the test baselines to better reflect the conventions.

geigerj: Sure, you can change the tests if you prefer.

> Since this is forming a url, we need the correct path

We control the URLs, so we just need them to follow the right convention. ISTM it makes more sense to derive the URL path from the Ruby namespace (which hopefully comes from the proto package, not the directory structure, although I'm not sure!) rather than the path to upstream inputs.

geigerj: Basically, my general feeling is that the proto path should have no semantic meaning at all. The only reason I passed the proto path value to toolkit via the package metadata config was so that we could link users back to the right directory of the googleapis/googleapis repo to view the source protos. I'm not sure any of the templates are doing this any more, so we might be able to remove this parameter altogether.

eoogbe: Ok, if we control the urls, then I will implement it using the namespace.

PR #1574, opened 2017-09-18
Switch from ServiceBuilder to Version

michaelbausor:
Context: The version is ServiceBuilder [has been deprecated](https://github.com/GoogleCloudPlatform/google-cloud-php/blob/master/src/ServiceBuilder.php#L31), and replaced with the [Version class](https://github.com/GoogleCloudPlatform/google-cloud-php/blob/master/src/Version.php).


michaelbausor:
PTAL: Updated to fix spacing, remove unused timeoutMillis, and add comments

PR #1577, opened 2017-09-19
Remove "spi" from java package names in config

andreamlin:
Resolves https://github.com/googleapis/toolkit/issues/1501.


andreamlin:
PTAL.

PR #1588, opened 2017-09-22
Add overview.rb file to Ruby docs

geigerj:
This overview serves as the landing page for documentation for
GAPIC-only APIs. Previously, it was not rendered due to the same
feature shortcoming in the jsondoc tool that motivated

  https://github.com/googleapis/toolkit/pull/1544

Retain readme text in the version index to serve as a per-version
code level example, but remove installation instructions, since
installation is not API version-dependent.

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceDocTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceDocTransformer.java
@@ -76,8 +76,6 @@ public class RubyGapicSurfaceDocTransformer implements ModelToViewTransformer {
             new RubyTypeTable(productConfig.getPackageName()),
             new RubyModelTypeNameConverter(productConfig.getPackageName()));
     // Use file path for package name to get file-specific package instead of package for the API.
-    // TODO(jgeiger): switch to use productConfig.getPackageName() instead of
-    // typeTable.getFullNameFor(file)
     SurfaceNamer namer = new RubySurfaceNamer(typeTable.getFullNameFor(file));
     String subPath = pathMapper.getOutputPath(file, productConfig);
     String baseFilename = namer.getProtoFileName(file);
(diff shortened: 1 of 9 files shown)


src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceDocTransformer.java
```diff
@@ -74,6 +76,8 @@ private ViewModel generateDoc(ProtoFile file, GapicProductConfig productConfig)
             new RubyTypeTable(productConfig.getPackageName()),
             new RubyModelTypeNameConverter(productConfig.getPackageName()));
     // Use file path for package name to get file-specific package instead of package for the API.
+    // TODO(jgeiger): switch to use productConfig.getPackageName() instead of
```
eoogbe: Why do we need to switch?

src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceDocTransformer.java
```diff
@@ -85,7 +89,23 @@ private ViewModel generateDoc(ProtoFile file, GapicProductConfig productConfig)
...
     doc.elementDocs(elementDocTransformer.generateElementDocs(typeTable, namer, file));
     doc.modules(
-        generateModuleViews(file.getModel(), productConfig, namer, isSourceApiInterfaceFile(file)));
+        generateModuleViews(
+            file.getModel(), productConfig, namer, isSourceApiInterfaceFile(file), false));
+    return doc.build();
+  }
+
+  private ViewModel generateOverview(Model model, GapicProductConfig productConfig) {
+    // TODO(jgeiger): switch to use productConfig.getPackageName() instead of
```
eoogbe: This namer already uses `productConfig.getPackageName()`

src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceDocTransformer.java
geigerj: Whoops, bad comment. Removed.

src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceDocTransformer.java
geigerj: Removed.

PR #1589, opened 2017-09-22
Ruby: exclude top level index path from .yardopts

### LINKED ISSUE #1558: Ruby .yardopts excludes wrong directory
### What

.yardopts should exclude proto files but instead seems to exclude GAPIC files. This breaks documentation generation.

### Context

With Video Intelligence we generate:

```
--exclude lib/google/cloud/video_intelligence/v1beta2
```

We should generate:

```
--exclude lib/google/cloud/videointelligence
```

eoogbe:
Fixes #1558

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/ruby/RubyPackageMetadataTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/ruby/RubyPackageMetadataTransformer.java
@@ -274,6 +274,7 @@ public class RubyPackageMetadataTransformer implements ModelToViewTransformer {
             fileHeaderTransformer.generateFileHeader(
                 productConfig, ImportSectionView.newBuilder().build(), surfaceNamer))
         .hasSmokeTests(hasSmokeTests)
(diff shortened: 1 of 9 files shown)


geigerj:
I haven't checked, but on inspection, this logic looks off; it should deal with the proto file package, not the GAPIC version index file (which is based on the GAPIC package). For example, for Video Intelligence, we want to exclude `lib/google/cloud/videointelligence` (proto package), not `lib/google/cloud/video_intelligence/` (GAPIC package).

geigerj:
This actually might be quite subtle in some other cases, like Speech, where the proto and GAPIC packages coincide; for example, [here we want to exclude the *_pb.rb files, but not the GAPIC client](https://github.com/GoogleCloudPlatform/google-cloud-ruby/tree/master/google-cloud-speech/lib/google/cloud/speech/v1).

Can `.yardopts` handle a glob expression like `--exclude **/*_pb.rb`?

eoogbe:
Exclude takes a regexp. This latest change was tested with vision.

PR #1592, opened 2017-09-28
NodeJS: Hand-edit fixes.

landrito:
Removes usage of protobufjs since thats in gax now and that line of code exporting toolkit version.

### DIFF
--- a/src/main/resources/com/google/api/codegen/nodejs/main.snip
+++ b/src/main/resources/com/google/api/codegen/nodejs/main.snip
@@ -14,7 +14,6 @@
    {@comments(fileHeader.licenseLines)}
    {@editingInstructions(protoFilename)}
    */
-  /* jscs: disable maximumLineLength */
   'use strict';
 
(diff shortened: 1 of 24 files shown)


src/main/resources/com/google/api/codegen/nodejs/main.snip
```diff
@@ -14,7 +14,6 @@
    {@comments(fileHeader.licenseLines)}
    {@editingInstructions(protoFilename)}
    */
-  /* TODO: introduce line-wrapping so that it never exceeds the limit. */
   /* jscs: disable maximumLineLength */
```
lukesneeringer: This line can go away too. We no longer use have.

lukesneeringer: *jscs

landrito: Done.

PR #1593, opened 2017-09-29
C# csproj generation

chrisdunelm:


src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicClientTransformer.java
```diff
@@ -129,6 +143,58 @@ private ModelTypeTable createTypeTable(String implicitPackageName) {
...
+    String outputPath = pathMapper.getOutputPath(context.getInterface(), productConfig);
+    view.outputPath(outputPath + File.separator + productConfig.getPackageName() + ".csproj");
+    view.fileHeader(fileHeaderTransformer.generateFileHeader(context));
+    view.serviceTitle(model.getServiceConfig().getTitle());
+    view.serviceDescription(model.getServiceConfig().getDocumentation().getSummary().trim());
+    view.domainLayerLocation(productConfig.getDomainLayerLocation());
+    view.authScopes(new ArrayList<String>()); // Unused in C#
+    view.releaseLevel(productConfig.getReleaseLevel());
+    String versionSuffix;
+    switch (productConfig.getReleaseLevel()) {
```
jskeet: Do we have a plan for what API producers will do after the first release?

src/main/resources/com/google/api/codegen/csharp/gapic_csproj.snip
```diff
@@ -0,0 +1,62 @@
+# packageInfo: PackageInfoView
+@snippet generate(packageInfo)
+  <?xml version="1.0" encoding="utf-8"?>
+  <Project Sdk="Microsoft.NET.Sdk">
+    <PropertyGroup>
+
+      <!-- TODO: Version defaults to 1.0.0, edit as required -->
+      <Version>{@packageInfo.version}</Version>
+      
+      <!-- TODO: Nuget packaging option -->
```
jskeet: Nit: NuGet :)

And "options" instead of "option"?

src/main/resources/com/google/api/codegen/csharp/gapic_csproj.snip
```diff
@@ -0,0 +1,62 @@
...
+          <Copyright>
+            @join line : packageInfo.fileHeader.copyrightLines
+              {@line}
+            @end
+          </Copyright>
+        @else
+          <Copyright>{@packageInfo.fileHeader.copyrightLines.get(0)}</Copyright>
+        @end
+        <Authors>{@packageInfo.packageMetadata.author}</Authors>
+        <PackageIconUrl>https://cloud.google.com/images/gcp-icon-64x64.png</PackageIconUrl>
```
jskeet: Possibly make it clearer that these need editing? We don't want non-Cloud packages with the Cloud icon
(Likewise the repo bits)

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicClientTransformer.java
chrisdunelm: Do you mean with regard to upping the version number from 1.0.0? If so, then they'll need to edit the version manually.

src/main/resources/com/google/api/codegen/csharp/gapic_csproj.snip
chrisdunelm: Done

src/main/resources/com/google/api/codegen/csharp/gapic_csproj.snip
chrisdunelm: Done, I think. Please let me know if you had something else in mind.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicClientTransformer.java
```diff
@@ -129,6 +143,58 @@ private ModelTypeTable createTypeTable(String implicitPackageName) {
...
+      case ALPHA:
+        versionSuffix = "-alpha01";
```
ethanbao: Will there ever have a "-alpha02"? If so, I assume it would require manual change on top of the genereated csproj file?

chrisdunelm: Yes, changing to -alpha02 will need a manual change to the generated csproj file. There is a TODO in the generated file about fixing the version.

PR #1595, opened 2017-10-02
NodeJS: Docs. Update to add member tags and remove external semantics.

landrito:
### What

1. If there was name collisions between grpc types and manually written types, the the documentation for the manual type would get overwritten by the rpc type.
2. DPEs would like the common protos types to be treated the same as  the rest of the types since it makes the docsite flow better.

### Fixes.

1. Add `@member` tag to the type docs
2. Remove external semantics.

### DIFF
--- a/src/main/java/com/google/api/codegen/util/js/JSCommentReformatter.java
+++ b/src/main/java/com/google/api/codegen/util/js/JSCommentReformatter.java
@@ -37,6 +37,7 @@ public class JSCommentReformatter implements CommentReformatter {
 
   public String getLinkedElementName(ProtoElement element) {
     String simpleName = element.getSimpleName();
-    return String.format("[%s]{@link %s}", simpleName, simpleName);
+    String packageName = element.getFile().getFullName();
+    return String.format("[%s]{@link %s.%s}", simpleName, packageName, simpleName);
   }
 }
--- a/src/main/java/com/google/api/codegen/util/js/JSCommentReformatter.java
+++ b/src/main/java/com/google/api/codegen/util/js/JSCommentReformatter.java
@@ -18,8 +18,6 @@ import com.google.api.codegen.util.CommentReformatter;
 import com.google.api.codegen.util.CommentTransformer;
 import com.google.api.codegen.util.LinkPattern;
 import com.google.api.tools.framework.model.ProtoElement;
-import com.google.api.tools.framework.model.ProtoFile;
-import com.google.common.collect.ImmutableSet;
 
 public class JSCommentReformatter implements CommentReformatter {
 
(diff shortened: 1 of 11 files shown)


src/test/java/com/google/api/codegen/testdata/nodejs/nodejs_main_library.baseline
```diff
@@ -1070,7 +1070,7 @@ LibraryServiceClient.prototype.deleteBook = function(request, options, callback)
  * @param {Object=} request.updateMask
  *   A field mask to apply, rendered as an HTTP parameter.
  *
- *   This object should have the same structure as [google.protobuf.FieldMask]{@link external:"google.protobuf.FieldMask"}
+ *   This object should have the same structure as [FieldMask]{@link FieldMask}
```
landrito: This change should be fine @jmdobry, correct?

jmdobry: Yes, and necessary. However, it should be `[FieldMask]{@link google.protobuf.FieldMask}`

src/main/java/com/google/api/codegen/util/js/JSCommentReformatter.java
```diff
@@ -38,32 +36,7 @@ public String reformat(String comment) {
...
   public String getLinkedElementName(ProtoElement element) {
-    if (isExternalFile(element.getFile())) {
-      String fullName = element.getFullName();
-      return String.format("[%s]{@link external:\"%s\"}", fullName, fullName);
-    } else {
-      String simpleName = element.getSimpleName();
-      return String.format("[%s]{@link %s}", simpleName, simpleName);
-    }
+    String simpleName = element.getSimpleName();
+    return String.format("[%s]{@link %s}", simpleName, simpleName);
```
jmdobry: `return String.format("[%s]{@link %s.%s}", simpleName, packageName, simpleName);`

landrito: I think using `element.fullName` will get us what we want. This will change the all of the type links.

landrito: JK I think what you have is correct!

jmdobry: I think we need to use `packageName`. For nested messages, `fullName` will contain an extra path element that JSDoc won't be able to resolve. For example

```proto
package "farm/animals"

message Horse {
  message Thing {}
}

message Cow {
}
```

```js
/**
 * @typedef Horse
 * @memberof farm.animals
 */
var Horse = {
  /**
   * @typedef Thing
   * @memberof farm.animals
   */ 
  Thing: {}
};
/**
 * @typedef Cow
 * @memberof farm.animals
 */
var Cow = {};
```

For `Thing`, `element.fullName` would try to link to it using `farm.animals.Horse.Thing`, but JSDoc wants to link to it using `farm.animals.Thing`.

PR #1600, opened 2017-10-04
Compute configgen

andreamlin:
Fixing issues in discogapic generation so that the resulting Compute surface has no compile errors.
- Use Empty as a placeholder for Callables that have no response.
- Append "resource" suffix to request object parameter names, in case they clash with a path parameter name.
- Allow for a chain of method calls in the generated implementation of PagedListDescriptor.extractResources(), e.g. "return payload.getItems().getAddresses()".


andreamlin:
PTAL

src/main/java/com/google/api/codegen/config/DiscoveryField.java
```diff
@@ -63,7 +65,7 @@ public Schema getDiscoveryField() {
   public String getSimpleName() {
     String name =
         Strings.isNullOrEmpty(schema.reference()) ? schema.getIdentifier() : schema.reference();
-    return Name.anyCamel(name).toLowerCamel();
+    return Name.fromUnderScoreOrCamel(name).toLowerCamel();
```
garrettjonesgoogle: Why will this name be in either form?

src/main/java/com/google/api/codegen/config/DiscoveryField.java
```diff
@@ -123,6 +135,11 @@ public boolean isRepeated() {
   }
 
   @Override
+  public boolean mayBeInResourceName() {
```
garrettjonesgoogle: Could you explain what this means?

src/main/java/com/google/api/codegen/config/PageStreamingConfig.java
```diff
@@ -176,18 +178,40 @@ static PageStreamingConfig createPageStreaming(
...
+            resourcesFieldPath.add(resourcesField);
+            break;
+          }
+        }
+        if (resourcesField != null) {
+          break;
+        }
       }
     }
+
```
garrettjonesgoogle: This loop got big, could you pull it out to a helper method?

src/main/java/com/google/api/codegen/config/ProtoField.java
```diff
@@ -61,6 +63,16 @@ public String getFullName() {
   }
 
   @Override
+  public Name getNameAsParameterName() {
+    return Name.from(getSimpleName());
+  }
+
+  @Override
+  public String getNameAsParameter() {
+    return getNameAsParameterName().toLowerUnderscore();
```
garrettjonesgoogle: token formatting (camel/underscore) should be up to the target language

src/main/java/com/google/api/codegen/configgen/transformer/DiscoConfigTransformer.java
```diff
@@ -108,34 +113,61 @@ private LicenseView generateLicense() {
 
   /**
    * Examines all of the resource paths used by the methods, and returns a map from each unique
-   * resource paths to a short name used by the collection configuration.
+   * resource paths to a short name used by the collection configuration. Each resource path is
+   * merely a string describing the fields in the entity, and the resource path may not be the same
```
garrettjonesgoogle: s/may not/might not/

("may not" means it's not allowed - based on context, it doesn't seem like that's what you want)

src/main/java/com/google/api/codegen/discogapic/transformer/DiscoGapicNamer.java
```diff
@@ -95,12 +100,26 @@ public static Name methodAsName(Method method) {
...
   }
 
-  /** Return the name of the resource from a given method's path. */
-  public static Name getResourceIdentifier(Method method) {
+  /** Return the name of the qualified resource from a given method's path. */
+  public static Name getQualifiedResourceIdentifier(Method method, String qualifyingResource) {
+    String methodPath = method.flatPath();
+    Name qualifier = Name.anyCamel(qualifyingResource);
+    String baseResource = getResourceIdentifier(methodPath).toLowerCamel();
+    Name baseResourceName = Name.anyCamel(baseResource);
```
garrettjonesgoogle: It seems like you could just initialize `baseResourceName` to the output of `getResourceIdentifier(methodPath)`

src/main/java/com/google/api/codegen/discogapic/transformer/DiscoGapicNamer.java
```diff
@@ -95,12 +100,26 @@ public static Name methodAsName(Method method) {
...
+
+    if (!Inflector.singularize(qualifier.toLowerCamel()).equals(baseResource)) {
+      baseResourceName = qualifier.join(baseResourceName);
+    }
```
garrettjonesgoogle: It doesn't seem like the output of singularization is actually used

src/main/java/com/google/api/codegen/viewmodel/ListMethodDetailView.java
```diff
@@ -28,7 +29,7 @@
 
   public abstract String iterateMethodName();
 
-  public abstract String resourcesFieldGetFunction();
+  public abstract List<String> resourcesFieldGetFunction();
```
garrettjonesgoogle: The field name should be plural if is a `List`

src/main/resources/com/google/api/codegen/java/settings.snip
```diff
@@ -345,7 +345,9 @@
           }
           @@Override
           public Iterable<{@desc.resourceTypeName}> extractResources({@desc.responseTypeName} payload) {
-            return payload.{@desc.resourcesFieldGetFunction}();
+            @let resourcesGetFunction = functionChain(desc.resourcesFieldGetFunction)
+              return payload.{@resourcesGetFunction};
+            @end
```
garrettjonesgoogle: Why are you introducing usage of `@let` here?

src/test/java/com/google/api/codegen/testdata/simplecompute_config.baseline
```diff
@@ -41,8 +41,8 @@ interfaces:
   # The entity_name is the name to be used as a basis for generated methods and
   # classes.
   collections:
-  - name_pattern: \{project}/regions/{region}/addresses
-    entity_name: region
+  - name_pattern: \{project}/regions/{region}
+    entity_name: addressesRegion
```
garrettjonesgoogle: It seems like this should just be `region`

src/test/java/com/google/api/codegen/testdata/simplecompute_config.baseline
```diff
@@ -41,8 +41,8 @@ interfaces:
   # The entity_name is the name to be used as a basis for generated methods and
   # classes.
   collections:
-  - name_pattern: \{project}/regions/{region}/addresses
-    entity_name: region
+  - name_pattern: \{project}/regions/{region}
```
garrettjonesgoogle: The fundamental problem with not having the "projects/" literal in the name patterns in this client is that it is inconsistent with One Platform clients, which do include it in all resource names that start with project.

src/main/java/com/google/api/codegen/config/DiscoveryField.java
garrettjonesgoogle: Hmm do they always start with a lowercase letter?

src/main/java/com/google/api/codegen/config/DiscoveryField.java
andreamlin: If the result is true, then it is possible that the Field is in a generated ResourceName. If the result is false, then that Field will never be in a generated ResourceName.

This is necessary because a ResourceNameTreatment in the gapic config applies to a whole method, but some Discovery methods have path parameter Fields that can go in the ResourceName in addition to a request body object that doesn't go in the ResourceName. Then the request body Field should return false in mayBeInResourceName() so that it doesn't get thrown in the ResourceName.

andreamlin: Also added a comment in the implementation to help explain.

src/main/java/com/google/api/codegen/config/ProtoField.java
andreamlin: Should I make a "gapic config" SurfaceNamer? This is used in config-gen.

andreamlin: Or should I assume that the gapic configs are case-insensitive, and adjust the transformers accordingly?

src/main/java/com/google/api/codegen/config/PageStreamingConfig.java
andreamlin: Done.

src/main/java/com/google/api/codegen/configgen/transformer/DiscoConfigTransformer.java
andreamlin: yep, done.

src/main/java/com/google/api/codegen/discogapic/transformer/DiscoGapicNamer.java
andreamlin: Done.

src/main/java/com/google/api/codegen/discogapic/transformer/DiscoGapicNamer.java
andreamlin: The output is used in checking to see if the `baseResouceName` should be qualified, but the output doesn't actually get returned in any way by this function, correct. 

`qualifyingResource` will be a plural noun. In the case that the `qualifyingResource` represents the same entity as the `baseResourceName`, it's cleaner to just use `baseResourceName`. E.g. instead of "TargetVpnGatewaysTargetVpnGatewayName", the name can just be "TargetVpnGatewayName".

src/main/java/com/google/api/codegen/config/ProtoField.java
garrettjonesgoogle: Oh, well "gapic config" is only one language, so actually I think what you have is fine.

src/main/java/com/google/api/codegen/discogapic/transformer/DiscoGapicNamer.java
garrettjonesgoogle: Oh! That makes sense. Maybe clarify that somewhere, because otherwise the logic looks odd.

src/main/resources/com/google/api/codegen/java/settings.snip
andreamlin: Removed `@let` and just put the function call in that `return payload...` line.

src/test/java/com/google/api/codegen/testdata/simplecompute_config.baseline
andreamlin: I started qualifying all the entity_names because there were a couple examples like `insertAddress(AddressName address, Address address)`, which has that variable name clash.

andreamlin: Ok, this is changed back to `region`. 
Updated as so: 
- Methods that use the same canonical resource path will use the same resource entity; e.g. compute.router.insert and compute.address.insert both have paths `"{project}/project/{region}/region"` so they both use `RegionName`.
- In the case where multiple canonical paths have the same simple resource identifiers, then and only then is the `entity_name` qualified. e.g. method 
`compute.regionOperations.get`
has path
`"{project}/regions/{region}/operations/{operation}"`
and method
`"compute.globalOperations.get"`
has path
`"{project}/global/operations/{operation}"`
The qualified resource name classes `RegionOperationsOperationName` and `GlobalOperationsOperationName` are generated.

src/main/java/com/google/api/codegen/config/DiscoveryField.java
andreamlin: Empirically, yes. The interface's function originated as a wrapper around com.google.api.tools.framework.model.Field.getSimpleName(), which, afaik, returns a string that starts with a lowercase letter.

src/main/java/com/google/api/codegen/discogapic/transformer/DiscoGapicNamer.java
andreamlin: Added a comment!

src/main/java/com/google/api/codegen/config/DiscoveryField.java
garrettjonesgoogle: Maybe we could have a slightly more targeted method `Name.anyLower`.

src/test/java/com/google/api/codegen/testdata/simplecompute_config.baseline
andreamlin: Done. As a plus, the name patterns don't weirdly begin with the escaping backslash!

src/main/java/com/google/api/codegen/config/DiscoveryField.java

andreamlin:
Addressed all comments. Round 2! PTAL

src/main/java/com/google/api/codegen/configgen/transformer/DiscoConfigTransformer.java
```diff
@@ -29,25 +29,15 @@
...
-import com.google.common.collect.Lists;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.TreeMap;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
+import com.google.common.collect.*;
+import java.util.*;
```
garrettjonesgoogle: no star imports

src/test/java/com/google/api/codegen/testdata/discogapic/java/java_simplecompute.v1.json.baseline
```diff
@@ -28,11 +28,12 @@ import javax.annotation.Generated;
...
 @Generated("by GAPIC")
 @BetaApi
-public final class AddressesRegionName implements ResourceName {
+public final class AddressName implements ResourceName {
+  private final String address;
   private final String project;
   private final String region;
   private static final PathTemplate PATH_TEMPLATE =
-        PathTemplate.createWithoutUrlEncoding("{project}/regions/{region}/addresses");
+        PathTemplate.createWithoutUrlEncoding("project/{project}/regions/{region}/addresses/{address}");
```
garrettjonesgoogle: this should start with `projects` (plural)

src/main/java/com/google/api/codegen/config/DiscoveryMethodModel.java
```diff
@@ -187,12 +190,16 @@ public String getAndSaveRequestTypeName(ImportTypeTable typeTable, SurfaceNamer
...
-    return typeTable.getAndSaveNicknameFor(fullName.getFullName());
+    Name responseName = DiscoGapicNamer.getResponseName(method);
+    if (responseName != null) {
+      TypeName fullName =
+          typeTable
+              .getTypeTable()
+              .getTypeNameInImplicitPackage(surfaceNamer.publicClassName(responseName));
+      return typeTable.getAndSaveNicknameFor(fullName.getFullName());
+    } else {
+      return typeTable.getAndSaveNicknameFor("com.google.protobuf.Empty");
```
garrettjonesgoogle: Why is there a protobuf type name in `DiscoveryMethodModel`?

src/main/java/com/google/api/codegen/configgen/transformer/DiscoConfigTransformer.java
```diff
@@ -100,42 +118,69 @@ private LicenseView generateLicense() {
...
+
+  /**
    * Examines all of the resource paths used by the methods, and returns a map from each unique
-   * resource paths to a short name used by the collection configuration.
+   * canonical resource path to a resource identifier (a short name used by the collection
+   * configuration). Each resource path is merely a string describing the fields in the entity, and
+   * the resource path might not be the same as the RPC endpoint URI. The resource identifier is
+   * globally unique within each API. Many methods may use the same resource collection. The
+   * resource identifier will be qualified (with the name of a parent resource) iff there are two
+   * resource identifiers with different canonical resource paths.
```
garrettjonesgoogle: Shouldn't this be: "iff there are two *or more* resource identifiers..."?

src/main/java/com/google/api/codegen/discogapic/transformer/DiscoGapicNamer.java
```diff
@@ -119,30 +138,77 @@ public static Name getRequestName(Method method) {
...
     TypeNameConverter typeNameConverter = languageNamer.getTypeNameConverter();
     return typeNameConverter.getTypeNameInImplicitPackage(
         languageNamer.publicClassName(getRequestName(method)));
   }
 
-  /** Get the response type name from a method. */
+  public static Name getInterfaceName(String defaultInterfaceName) {
+    String[] pieces = defaultInterfaceName.split(REGEX_DELIMITER);
+    String resource = pieces[pieces.length - 1];
+    return Name.anyCamel(Inflector.singularize(resource), "admin");
```
garrettjonesgoogle: Where did this "admin" come from?

src/main/java/com/google/api/codegen/discogapic/transformer/DiscoGapicNamer.java
```diff
@@ -119,30 +138,77 @@ public static Name getRequestName(Method method) {
...
-      String resourceName = pieces[pieces.length - 2];
-      if (!method.isPluralMethod()) {
-        resourceName = Inflector.singularize(resourceName);
+    }
+    return null;
+  }
+
+  /**
+   * Get the canonical path for a method, in the form "(%s/\{%s\})+" e.g. for a method path
+   * "{project}/region/{region}/addresses", this returns "project/{project}/region/{region}".
```
garrettjonesgoogle: The literals should be plural in the example paths.

src/main/java/com/google/api/codegen/configgen/transformer/DiscoConfigTransformer.java
andreamlin: Done.

src/main/java/com/google/api/codegen/configgen/transformer/DiscoConfigTransformer.java
andreamlin: Yes :) Done.

src/test/java/com/google/api/codegen/testdata/discogapic/java/java_simplecompute.v1.json.baseline
andreamlin: oooh good catch; done.

src/main/java/com/google/api/codegen/discogapic/transformer/DiscoGapicNamer.java
andreamlin: Done.

src/main/java/com/google/api/codegen/discogapic/transformer/DiscoGapicNamer.java
andreamlin: Removed.

src/main/java/com/google/api/codegen/config/DiscoveryMethodModel.java
andreamlin: Replaced `Empty` with `java.lang.Void`. This represents an empty type for the *Callable generic classes.

andreamlin:
PTAL

src/main/java/com/google/api/codegen/Inflector.java
```diff
@@ -35,4 +35,14 @@ public static String singularize(String in) {
 
     return in;
   }
+
+  // TODO (garrettjones) find an existing function that does this
```
garrettjonesgoogle: Maybe remove my name off of this (and on the other method too).

src/main/java/com/google/api/codegen/discogapic/transformer/DiscoGapicNamer.java
```diff
@@ -182,13 +182,14 @@ public static Name getResponseName(Method method) {
 
   /**
    * Get the canonical path for a method, in the form "(%s/\{%s\})+" e.g. for a method path
-   * "{project}/region/{region}/addresses", this returns "project/{project}/region/{region}".
+   * "{project}/regions/{region}/addresses", this returns "projects/{project}/regions/{region}".
```
garrettjonesgoogle: you still need `projects/` at the beginning of the first example.

src/main/java/com/google/api/codegen/discogapic/transformer/java/JavaDiscoGapicResourceNameToViewTransformer.java
```diff
@@ -44,15 +43,7 @@
...
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.HashSet;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
garrettjonesgoogle: still some star imports here

src/main/java/com/google/api/codegen/transformer/StaticLangResourceObjectTransformer.java
```diff
@@ -37,6 +37,7 @@ public RequestObjectParamView generateRequestObjectParam(
     for (FieldModel f : requiredFields) {
       if (f.getSimpleName().equals(field.getSimpleName())) {
         isRequired = true;
+        break;
```
garrettjonesgoogle: ooh ooh this is a good place for a streaming lambda expression (completely, utterly optional)

src/main/java/com/google/api/codegen/discogapic/transformer/DiscoGapicNamer.java
andreamlin: So the first example is an example of a non-normalized path, the input to the function. Specifically, it's verbatim the path for compute.addresses.list as given in the Discovery doc. 

I wanted to not have the "projects/" prefix in the first example in order to contrast it with the canonical form, which does have the "projects/" prefix.

garrettjonesgoogle: derp I wasn't paying close enough attention...

src/main/java/com/google/api/codegen/discogapic/transformer/java/JavaDiscoGapicResourceNameToViewTransformer.java
andreamlin: All the star imports are gone now!

src/main/java/com/google/api/codegen/Inflector.java
andreamlin: :)

src/main/java/com/google/api/codegen/transformer/StaticLangResourceObjectTransformer.java
andreamlin: I will definitely get around to using Java 8 features!

For this particular for-loop, it's iterating over an Iterable<T>, which doesn't have the .stream() method.

andreamlin:
PTAL

PR #1603, opened 2017-10-07
Add support for header-based routing ("x-goog-request-params" header)

vam-google:
The corresponding [gax-java PR](https://github.com/googleapis/gax-java/pull/375)


src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -270,13 +272,19 @@ message MethodConfigProto {
   // IAM methods to IamPolicy for Pub/Sub.
   string reroute_to_grpc_interface = 12;
 
+  // Specifies the list of the method's input message fields (including nested
+  // ones, using the dot notation), which should be url-encoded in
+  // "x-goog-request-params" header
```
garrettjonesgoogle: I think we should also have the header key configurable.

vam-google: This header is hardcoded on google server side. It is analogous to url path and url parameters in HTTP/1+rest request. In other words it is not just a header, it is a support for url-path/params in https/2+grpc world. The name of the header is not going to change, if we will ever want to add more headers, they will be something different (i.e. they will be legit headers, instead of being url-path imitation for grpc).

src/main/resources/com/google/api/codegen/java/grpc_stub.snip
```diff
@@ -56,7 +56,35 @@
...
+      @@Override
+      public Map<String, String> extract({@callable.requestTypeName} request) {
+        ImmutableMap.Builder<String, String> params = ImmutableMap.builder();
+        {@requestParamExtractor(callable.headerRequestParams)}
+        return params.build();
+      }
+    }
+  @else
+    EmptyRequestParamsExtractor.<{@callable.requestTypeName}>of()
+  @end
```
garrettjonesgoogle: I would prefer for the entry points that don't need a params extractor to not pass in this parameter at all. Otherwise, a *lot* of files will be touched in your regeneration (as you can see in the baseline diff), which we could avoid touching.

src/test/java/com/google/api/codegen/testdata/java/java_library.baseline
```diff
@@ -5125,196 +5128,232 @@ public class GrpcLibraryServiceStub extends LibraryServiceStub {
...
               "google.example.library.v1.LibraryService/CreateBook",
               io.grpc.protobuf.ProtoUtils.marshaller(CreateBookRequest.getDefaultInstance()),
-              io.grpc.protobuf.ProtoUtils.marshaller(Book.getDefaultInstance())));
+              io.grpc.protobuf.ProtoUtils.marshaller(Book.getDefaultInstance())),
+          new RequestParamsExtractor<CreateBookRequest>() {
+            @Override
+            public Map<String, String> extract(CreateBookRequest request) {
+              ImmutableMap.Builder<String, String> params = ImmutableMap.builder();
+              params.put("name", String.valueOf(request.getName()));
+              params.put("book.read", String.valueOf(request.getBook().getRead()));
```
garrettjonesgoogle: I think it could be possible for a field like `request.getBook()` to be null.

vam-google: And in that case it will be "null" in the string. This is by design.

src/main/resources/com/google/api/codegen/java/grpc_stub.snip
vam-google: This is by design. Yes they will be touched and will explicitly say that they don't need parameter extraction. This extraction thing is very close to grpc world (i.e. it would be in grpc spec if changing grpc spec was not so difficult). This is why I try to keep this implementation as close to grpc as possible (i.e. changes are in `DirectCallable`) and explicit (a grpc call should have an explicit name, parameter and header parameters).

vam-google:
PTAL

src/main/resources/com/google/api/codegen/java/grpc_stub.snip
garrettjonesgoogle: I understand and agree that the implementation should be close to grpc, but when it comes to codegen, there are some practical matters that come to play. If you don't pass a parameter when there is no extraction, then actually *no* files in google-cloud-java need to be touched right now (since this feature isn't configured for any current APIs). Additionally, I am going to change the interface soon anyway, and I don't want to have "regeneration whiplash" (widescale codegen change, then another one right away).

src/test/java/com/google/api/codegen/testdata/java/java_library.baseline
garrettjonesgoogle: I agree about failing fast, but I would prefer IllegalArgumentException over NullPointerException.

vam-google: I believe it is ok to allow NPE here. Otherwise it will significantly complicate snippets (the getters chain can be arbitrarily long, each component should be checked for non-null).
Also NPE are often considered ok and are thrown directly: `java.util.List.add(E)` and `com.google.common.base.Preconditions.checkNotNull()` are two basic examples, when `NPE` can be thrown directly.

src/main/resources/com/google/api/codegen/java/grpc_stub.snip
vam-google: Ok, changed.

vam-google:
PTAL

PR #1604, opened 2017-10-09
Node.js autogen refactor

lukesneeringer:
This ended up being a pretty substantial refactor for the Node.js autogen.

It does the following:

  * Converts clients to ES6 classes with Node 4 syntax.
  * Dramatically changes the interior imports/exports. You now get the client class straight up, and instantiate it straight up (e.g. `new language.v1.LanguageServiceClient()`).
      * The various passes at "builder functions" (e.g. `language.v1()` and `LanguageServiceClientBuilder()`) are completely removed.
      * The explicit version number use is still optional. Same rules as Python.
  * Removes the dependency on `extend` and uses `Object.assign`.
  * _Mostly_ complies with prettier. (There are a couple oddities where line lengths were hard to control, and while prettier-compliant, there are a few really stupidly excessive line breaks also.)
  * Documentation examples got a substantial syntax and style rehash.

To-do:

  * [x] Port auto-generated unit tests.
  * [x] Port auto-generated system ("smoke") tests.
  * [x] Make README/docs use the explicit constructor.

We need to do smoke tests in the following APIs to prove that the autogen still works:

  * [x] Natural language (base case)
  * [x] Pub/Sub (path templates)
  * [x] Speech (streaming)
  * [x] Video Intelligence (LRO)
  * [x] (pagination -- what uses this?)
  * [x] ~(batching -- what uses this?)~ Pub/Sub covers batching.

### DIFF
--- a/src/main/resources/com/google/api/codegen/nodejs/method_sample.snip
+++ b/src/main/resources/com/google/api/codegen/nodejs/method_sample.snip
@@ -182,7 +182,7 @@
     @end
 
     var options = {autoPaginate: false};
-    var callbacl = responses => {
+    var callback = responses => {
       // The actual resources in a response.
       var resources = responses[0];
       // The next request if the response shows that there are more responses.
--- a/src/main/resources/com/google/api/codegen/nodejs/method_sample.snip
+++ b/src/main/resources/com/google/api/codegen/nodejs/method_sample.snip
@@ -2,17 +2,11 @@
 
 @snippet decorateSampleCode(apiMethod, coreSampleCode)
     @if apiMethod.hasApiVersion
-        var {@apiMethod.localPackageName} = require('{@apiMethod.packageName}');
+        const {@apiMethod.localPackageName} = require('{@apiMethod.packageName}');
 
-        @if apiMethod.packageHasMultipleServices
-          var client = {@apiMethod.localPackageName}.{@apiMethod.apiVersion}.{@apiMethod.packageServiceName}({
-            // optional auth parameters.
-          });
-        @else
-          var client = {@apiMethod.localPackageName}.{@apiMethod.apiVersion}({
-            // optional auth parameters.
-          });
-        @end
+        var client = new {@apiMethod.localPackageName}.{@apiMethod.apiVersion}.{@apiMethod.apiClassName}({
+          // optional auth parameters.
+        });
 
         {@coreSampleCode}
     @else
@@ -21,17 +15,11 @@
 @end
 
 @snippet decorateSampleCodeUnversioned(apiMethod, coreSampleCode)
-    var {@apiMethod.localPackageName} = require('{@apiMethod.packageName}');
+    const {@apiMethod.localPackageName} = require('{@apiMethod.packageName}');
 
-    @if apiMethod.packageHasMultipleServices
-      var client = {@apiMethod.localPackageName}.{@apiMethod.packageServiceName}({
-        // optional auth parameters.
-      });
-    @else
-      var client = {@apiMethod.localPackageName}({
-        // optional auth parameters.
-      });
-    @end
+    var client = {@apiMethod.localPackageName}.{@apiMethod.apiClassName}({
+      // optional auth parameters.
+    });
 
     {@coreSampleCode}
 @end
--- a/src/test/java/com/google/api/codegen/testdata/nodejs/nodejs_doc_library.baseline
+++ b/src/test/java/com/google/api/codegen/testdata/nodejs/nodejs_doc_library.baseline
@@ -2474,7 +2474,7 @@ class LibraryServiceClient {
       gaxGrpc.loadProto(
         path.join(__dirname, '..', '..', 'protos'),
         'library.proto'
-      ),,
+      ),
       gaxGrpc.loadProto(
         path.join(__dirname, '..', '..', 'protos'),
         'tagger.proto'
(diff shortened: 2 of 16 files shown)


src/test/java/com/google/api/codegen/testdata/nodejs/nodejs_multiple_services.baseline
```diff
@@ -216,177 +123,171 @@ module.exports.v1.types = v1Protos;
...
+    // Ensure that options include the service address and port.
+    opts = Object.assign({
+      clientConfig: {},
+      port: this.constructor.port,
+      servicePath: this.constructor.servicePath,
+    }, opts);
+
+    // Create a `gaxGrpc` object, with any grpc-specific options
+    // sent to the client.
+    Object.assign(opts, {scopes: this.constructor.scopes});
```
stephenplusplus: I think just assigning opts.scopes directly would be easier.

lukesneeringer: Agreed; done.

src/test/java/com/google/api/codegen/testdata/nodejs/nodejs_library.baseline
```diff
@@ -123,38 +123,39 @@ $ npm install --save @google-cloud/library
...
+  it('successfully makes a call to the service', done => {
+    const library = require('../src');
 
-    var client = library.v1({
+    var client = new library.v1.LibraryServiceClient({
       // optional auth parameters.
     });
 
     var formattedName = client.bookPath("testShelf-" + Date.now().toString(), projectId);
     var rating = library.v1.types.Book.Rating.GOOD;
```
lukesneeringer: @landrito This line is now wrong, since we nixed `types`. It should just be the string `'GOOD'`. Any idea how to change that? (It was the only thing I could not track down.)

landrito:
I think if you changed this line:
https://github.com/googleapis/toolkit/blob/9402fbbe3c393f121f528ece2003b7a418188218/src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSModelTypeNameConverter.java#L187

to:
```java
return TypedValue.create(getTypeName(type), String.format("'%s'", value.getSimpleName()));

You'll get what you're looking for.

File: src/test/java/com/google/api/codegen/testdata/nodejs/nodejs_library.baseline (hunk 2 above)
Comment (landrito) [reply src/test/java/com/google/api/codegen/testdata/nodejs/nodejs_library.baseline]: I think if you changed this line:
https://github.com/googleapis/toolkit/blob/9402fbbe3c393f121f528ece2003b7a418188218/src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSModelTypeNameConverter.java#L187

to:
```java
return TypedValue.create(getTypeName(type), String.format("'%s'", value.getSimpleName()));

You'll get what you're looking for.

Comment (jmdobry) [reply src/test/java/com/google/api/codegen/testdata/nodejs/nodejs_library.baseline]: What do you mean we nixed `types`? Will the user still have access to them?

**Comment by lukesneeringer** [comment]
I will make one on `nodejs-language`.

File: src/test/java/com/google/api/codegen/testdata/nodejs/nodejs_doc_library.baseline [hunk 3]
```diff
@@ -137,7 +137,7 @@ describe('LibraryServiceSmokeTest', () => {
     });
 
     var formattedName = client.bookPath("testShelf-" + Date.now().toString(), projectId);
-    var rating = library.v1.types.Book.Rating.GOOD;
+    var rating = 'GOOD';
```
jmdobry: Why are we nixing the types?

src/main/resources/com/google/api/codegen/nodejs/method_sample.snip
```diff
@@ -193,242 +182,246 @@
...
-                console.log(resources[i])
-            @else
-                // doThingsWith(resources[i]);
-            @end
-        }
-        if (nextRequest) {
-            // Fetch the next page.
-            return client.{@apiMethod.name}(nextRequest, options).then(callback);
-        }
+    var callbacl = responses => {
```
landrito: s/callbacl/callback

src/test/java/com/google/api/codegen/testdata/nodejs/nodejs_doc_library.baseline
lukesneeringer: @landtiro pointed out that there is no way to properly merge types from multiple sources together and avoid collisions.

We are importing types from:

  * the API
  * longrunning
  * IAM
  * common protos

And there could potentially be conflicts. The `types` module essentially compresses the various namespaces onto one spot. I _really_ like that (because the excessive namespacing is a confusing point for users) but our documentation is sourced from the `doc_` files, so the types module seems superfluous.

I am okay with having it, though, if it is valuable to users.

jmdobry: The `types` wasn't being used for docs, but rather was user-accessible. It's useful because they don't need to know the magic strings/numbers, they can just access the generated type, e.g. `library.v1.types.Book.Rating.GOOD` instead of `'GOOD'`.

I vote that we keep the types, but un-squash them. When the user accesses them they should follow the path as it would be in the protos. We really should give users access to the stuff loaded by Protobuf.js

lukesneeringer: Merging this as-is for now (this is an additive change so we do not need to block on it, and we have other repos waiting on this).

In principle, I am willing to do this, but I want to make sure we think through the strategy. I do not love `types.google.cloud.language.v1.Document` as something we ask a user to put in their code, for example, and it also introduces inconsistency in nomenclature (e.g. `video-intelligence` vs. `videointelligence`).

We can circle back and find the right thing,

PR #1605, opened 2017-10-09
C# partial ServerStreamingBase subclasses

evildour:


evildour:
@chrisdunelm Can you take a look at this one? Thanks.

src/main/resources/com/google/api/codegen/csharp/gapic_client.snip
```diff
@@ -601,7 +601,7 @@
         /// <summary>
         /// Server streaming methods for <c>{@method.name}</c>.
         /// </summary>
-        public abstract class {@method.name}Stream : ServerStreamingBase<{@method.serviceResponseTypeName}>
```
chrisdunelm: This looks fine.
But please can you also make the impl class partial (line 576), and the bi-directional abstract and impl classes partial too? May as well do the whole lot in one go.

The client-streaming classes aren't yet written, which is why they don't need doing ;)

Thanks!

evildour: Done

PR #1607, opened 2017-10-09
Fix Node constructor comments.

lukesneeringer:
These were in index.js before #1604, moved to the correct place (to the class itself, not the builder function).


src/main/resources/com/google/api/codegen/nodejs/main.snip
```diff
@@ -40,8 +40,30 @@
      {@comments(xapi.doc.lines)}
      *
    @end
-   *
    * @@class
+   *
+   * @@param {object=} options - [Configuration object](#/docs).
```
jmdobry: `[Configuration object](#/docs)` won't link to anything real, just remove the link.

PR #1608, opened 2017-10-10
Ruby: Use the credentials class in google-auth-library-ruby

landrito:
Depends on https://github.com/google/google-auth-library-ruby/pull/122


src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceTransformer.java
```diff
@@ -73,6 +73,8 @@
       ImmutableList.of("GOOGLE_CLOUD_KEYFILE", "GCLOUD_KEYFILE");
   private static final List<String> DEFAULT_JSON_ENV_VARS =
       ImmutableList.of("GOOGLE_CLOUD_KEYFILE_JSON", "GCLOUD_KEYFILE_JSON");
+  private static final List<String> DEFAULT_PATHS =
+      ImmutableList.of("~/.config/gcloud/application_default_credentials.json");
```
geigerj: Guessing this locates the key you get when you run `gcloud auth application-default login`? What was the previous behavior/the behavior for the manual Veneer creds?

landrito: It was just a default path defined in google-cloud-core. https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-core/lib/google/cloud/credentials.rb#L34

PR #1609, opened 2017-10-10
Docs updates

lukesneeringer:
This makes docs updates pointed out by @jmdobry in googleapis/nodejs-language#2:

  * [x] Changing `@member` to `@memberof`
  * [x] Namespacing the service client classes
  * [x] Providing `@memberof` on enums.
  * [x] Fixing enum links.
  * [x] Providing `@module` on index.js (currently wrong).

It also changes the license documentation to `//` comments.

### DIFF
--- a/src/test/java/com/google/api/codegen/testdata/nodejs/nodejs_library.baseline
+++ b/src/test/java/com/google/api/codegen/testdata/nodejs/nodejs_library.baseline
@@ -245,6 +245,7 @@ const protobuf = require('protobufjs');
  * Service comment may include special characters: <>&"`'@.
  *
  * @class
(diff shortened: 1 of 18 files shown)


src/test/java/com/google/api/codegen/testdata/nodejs/nodejs_library.baseline
```diff
@@ -169,6 +169,7 @@ describe('LibraryServiceSmokeTest', () => {
 // limitations under the License.
 
 /**
+ * @module library
```
lukesneeringer: Is this correct? Will be be `@google-cloud/${api}` on a real API?

PR #1610, opened 2017-10-10
Add PHP smoke test

saicheems:
This commit adds a smoke test snippet and transformer for PHP.

As part of that addition, the generated test directory for the PHP GAPIC
client is reorganized into `tests/unit`, which contains unit tests, and
`tests/system`, which contains the smoke test. This also mirrors the
layout on google-cloud-php.

`PhpGapicSurfaceTestTransformer.java` is rewritten to improve code
clarity.


src/main/java/com/google/api/codegen/util/ruby/RubyNameFormatter.java
```diff
@@ -57,7 +57,7 @@ public String publicFieldName(Name name) {
 
   @Override
   public String varReference(Name name) {
-    return "@" + name.toLowerUnderscore();
+    return name.toLowerUnderscore();
```
saicheems: This change is okay IMO because an `@` variable in Ruby is not a reference -- `@` variables are instance variables.

Also, it's never used by the Ruby transformers.

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -505,7 +505,7 @@ private InitValueView getInitValue(GapicMethodContext context, InitCodeNode item
         InitValue initValue = initValueConfig.getResourceNameBindingValues().get(entityName);
         switch (initValue.getType()) {
           case Variable:
-            entityValue = context.getNamer().localVarName(Name.from(initValue.getValue()));
+            entityValue = context.getNamer().varReference(Name.from(initValue.getValue()));
```
saicheems: This is changed because we need `projectId` to be printed as `$projectId` in PHP.

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTestTransformer.java
```diff
@@ -16,11 +16,13 @@
 
 import com.google.api.codegen.InterfaceView;
```
saicheems: The diff for this class is difficult to walk through because I reorganized it to be much more clear. There are improvements throughout, but the most important are in the new `createSmokeTest.*` methods.

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTestTransformer.java
```diff
@@ -256,6 +215,117 @@ private static ModelTypeTable createTypeTable(String packageName) {
...
+            .grpcMethods(new ArrayList<MockGrpcMethodView>())
+            .build());
+
+    mockServiceImplFile.outputPath(namer.getSourceFilePath(outputPath, name));
+    mockServiceImplFile.templateFileName(MOCK_SERVICE_IMPL_TEMPLATE_FILE);
+
+    ImportSectionView importSection =
+        importSectionTransformer.generateImportSection(context.getModelTypeTable().getImports());
+    PhpSurfaceNamer testPackageNamer =
+        new PhpSurfaceNamer(namer.getTestPackageName(SurfaceNamer.TestKind.UNIT));
```
garrettjonesgoogle: You could instead do `namer.cloneWithPackageName(...)`

src/main/java/com/google/api/codegen/util/ruby/RubyNameFormatter.java
garrettjonesgoogle: Wait do we need to distinguish between localVarReference and memberVarReference?

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTestTransformer.java
saicheems: Cool! Done

src/main/java/com/google/api/codegen/util/ruby/RubyNameFormatter.java
saicheems: Good point, I renamed the function to `localVarReference`. Looks like nobody would use a `memberVarReference` method yet, so I haven't added it.

PR #1611, opened 2017-10-11
Remove redundant if block for Python test provider

saicheems:
Generate Python test files only if the test generator is enabled in the
generator config.


eoogbe:
I believe the test provider was set to generate whenever Python generated because the test generator enabling was not working properly. Has that been fixed for Python?

saicheems:
I didn't notice a difference, running `artman2` with this change did generate a `tests` directory IIRC. I can check again.

saicheems:
Yeah if I generate with this change a `tests` directory is created. I guess it was fixed? Lmk.

PR #1614, opened 2017-10-12
Improve Node.js index.js comments

lukesneeringer:


src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSGapicSurfaceTransformer.java
```diff
@@ -330,7 +329,8 @@ private ViewModel generateApiClass(GapicInterfaceContext context, boolean hasMul
               .fileHeader(
                   fileHeaderTransformer.generateFileHeader(
                       productConfig, ImportSectionView.newBuilder().build(), namer))
-              .packageName(packageMetadataNamer.getMetadataIdentifier());
+              .packageName(packageMetadataNamer.getMetadataIdentifier())
```
landrito: Can you add a todo here to transform this into an acceptable name. Perceivably this name can have a hyphen in it like shown in the multiple-services baseline, which would cause an error.

src/main/resources/com/google/api/codegen/nodejs/version_index.snip
```diff
@@ -6,18 +6,70 @@
   {@lineComments(index.fileHeader.licenseLines)}
 
   /**
-   * @@module {@index.packageName}
+   * @@namespace google
```
landrito: Add a todo above this to fix this.

src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSGapicSurfaceTransformer.java
lukesneeringer: Filed #1615

src/main/resources/com/google/api/codegen/nodejs/version_index.snip
lukesneeringer: Filed #1616

PR #1617, opened 2017-10-12
Fix method calls in samples with optional args

### LINKED ISSUE #1255: Ruby/Node: Generated smoke tests are broken when config uses optional parameters
# What
Smoke tests that are given a config that has optional parameters set causes the optional parameters to be passed in a required parameters

### Example in Logging
```rb 
logging_service_v2_client = LoggingServiceV2Client.new
formatted_log_name = LoggingServiceV2Client.log_path(project_id, "test-" + Time.new.to_i.to_s)
resource = MonitoredResource.new
labels = {}
entries = []
response = logging_service_v2_client.write_log_entries(formatted_log_name, resource, labels, entries)
```

The method call should actually look like this:
```rb
logging_service_v2_client.write_log_entries(
  formatted_log_name, 
  resource: resource, 
  labels: labels, 
  entries: entries)
```

saicheems:
Fixes #1255


src/main/resources/com/google/api/codegen/php/method_sample.snip
```diff
@@ -173,24 +173,31 @@
...
+            {@sampleMethodCallArgList(apiMethod.initCode)})
     @else
         ${@apiMethod.apiVariableName}->{@apiMethod.name}()
     @end
 @end
 
-@private pageSizeMethodCallSampleCode(apiMethod)
-    @if apiMethod.hasRequiredParameters
-        ${@apiMethod.apiVariableName}->{@apiMethod.name}(\
-            {@sampleMethodCallArgList(apiMethod.initCode.fieldSettings)}, ['pageSize' => 5])
```
saicheems: I've changed the samples for PHP to not print `['pageSize' => 5]` as an optional parameter, mostly because it becomes difficult to actually insert an override in the gapic yaml for paging samples.

I think this is okay, as no other language is explicit about page sizes in the samples.

saicheems:
Without this change (but with the changes to `library.proto` and `library_gapic.yaml`):

```
PHP

$libraryServiceClient->updateBook($formattedName, $optionalFoo, $book, $updateMask, $physicalMask);
```

```
Python

old:
response = client.update_book(name, optional_foo, book, update_mask, physical_mask)

new:
response = client.update_book(name, book, optional_foo=optional_foo, update_mask=update_mask, physical_mask=physical_mask)
```

```
Ruby

old:
response = library_service_client.update_book(formatted_name, optional_foo, book, update_mask, physical_mask)

new:
response = library_service_client->update_book(formatted_name, book, optional_foo: optional_foo, update_mask: update_mask, physical_mask: physical_mask)
```

Before, all requests above would fail in all languages. In Python, `optional_foo` is listed before book, even though `optional_foo` is a named parameter, and `book` is not. Same for Ruby.

saicheems:
I'm asking you to take a look first @garrettjonesgoogle (mostly for `InitCodeTransformer.java`) but I'll tag the other language owners after.

saicheems:
@geigerj @lukesneeringer 

Tagged you to take a look at the changes to Ruby and Python baselines.

saicheems:
Jk made a mistake with the Ruby baselines somehow, looks like `.` changed to `->`... I'll fix that.

saicheems:
OK all good again.

garrettjonesgoogle:
I still intend on looking at this, I just have to read up on streams :-)

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -565,10 +589,19 @@ private InitValueView getInitValue(GapicMethodContext context, InitCodeNode item
...
 
-  private List<FieldSettingView> getFieldSettings(
+  /** Returns true if fieldConfigs contains fieldConfig. */
+  private boolean hasFieldConfig(Iterable<FieldConfig> fieldConfigs, FieldConfig fieldConfig) {
+    return StreamSupport.stream(fieldConfigs.spliterator(), false)
+        .anyMatch(f -> f.equals(fieldConfig));
+  }
+
+  /** Returns an insertion order map of FieldSettingView to whether or not the field is required. */
+  private LinkedHashMap<FieldSettingView, Boolean> getFieldSettings(
```
geigerj: nit: using a `Map<FieldSettingView, Boolean>` to store whether a field is required is unintuitive for me. Why not just make whether it is required part of the `View`?

saicheems: That's a great point, done.

saicheems:
@lukesneeringer all good for the Python baseline?

saicheems:
@vam-google all good for the PHP baselines?

See https://github.com/googleapis/toolkit/pull/1617#issuecomment-336287692 for what this PR is about. Also note that I dropped `['pageSize' => 5]` argument from paging samples -- the reasoning is that those samples break if any overrides are applied which specify optional parameters. I think it's fine to do, since no other language specifies that limit.

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -599,7 +614,8 @@ private InitValueView getInitValue(GapicMethodContext context, InitCodeNode item
                 .variantType(namer.getOneofVariantTypeName(item.getOneofConfig()))
                 .build());
       }
-
+      fieldSetting.required(
+          hasFieldConfig(context.getMethodConfig().getRequiredFieldConfigs(), fieldConfig));
```
garrettjonesgoogle: This looks like it has O(n^2) runtime characteristics - would that be concerning?

saicheems: Depends... n in this case is almost certainly < 10. So probably not a big deal. If it's concerning to you, the alternative is to initialize a `Map<MethodConfig, Set<RequiredFieldConfig>>` or something in the parent scope.

Lmk.

saicheems: Actually I'll just change it.

saicheems: Changed my mind, I tried using a `Set`, but it doesn't work. The hash codes of the `FieldConfig` derived from `InitCodeNode` is different from the ones derived from `GapicMethodContext`. The equals comparison does work, however.

garrettjonesgoogle: Can we do comparison on the field name instead of the entirety of the field config?

saicheems: Done

garrettjonesgoogle:
toolkit implementation LGTM - submit after all languages are approved.

vam-google:
LGTM for PHP

PR #1618, opened 2017-10-12
PHP: Prefer single quotes

saicheems:
String literals which are

- generated by the PHP transformers
- present in the PHP snippets

prefer single quotes where applicable. Double quoted strings are still
used in cases where variables are substituted or escape characters
are present.

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/php/PhpModelTypeNameConverter.java
+++ b/src/main/java/com/google/api/codegen/transformer/php/PhpModelTypeNameConverter.java
@@ -72,8 +72,8 @@ public class PhpModelTypeNameConverter implements ModelTypeNameConverter {
           .put(Type.TYPE_SINT32, "0")
           .put(Type.TYPE_FIXED32, "0")
           .put(Type.TYPE_SFIXED32, "0")
-          .put(Type.TYPE_STRING, "\"\"")
-          .put(Type.TYPE_BYTES, "\"\"")
+          .put(Type.TYPE_STRING, "\'\'")
+          .put(Type.TYPE_BYTES, "\'\'")
           .build();
 
   /** A map from protobuf message type names to PHP types for special case messages. */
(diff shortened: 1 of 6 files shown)


saicheems:
This should just be a style change, I was bothered by the mixing of double and single quotes everywhere. If you want me to drop this change lmk.

src/main/java/com/google/api/codegen/transformer/php/PhpModelTypeNameConverter.java
```diff
@@ -262,7 +262,7 @@ public String renderPrimitiveValue(TypeRef type, String value) {
         return value.toLowerCase();
       case TYPE_STRING:
       case TYPE_BYTES:
-        return "\"" + value + "\"";
```
vam-google: Please change to `'\''` here and in all other places (except `PRIMITIVE_ZERO_VALUE` initialization, since that one actually expects `String`). It is not a big deal, but `'\''` is slightly less heavy, since it is just a `char` not a `String`.

src/main/java/com/google/api/codegen/transformer/php/PhpSurfaceNamer.java
```diff
@@ -243,10 +243,15 @@ public String injectRandomStringGeneratorCode(String randomString) {
...
-          stringParts.add("\"" + token + "\"");
+          stringParts.add(quoted(token));
         }
       }
     }
     return Joiner.on(". ").join(stringParts);
   }
+
+  @Override
+  public String quoted(String text) {
```
vam-google: Please ensure that this will not start putting single quotes where double quotes are expected.

saicheems: I think it should be okay, the only place this function is used in the project is to enquote entity names (ex: `'[PROJECT_ID]'`):

```
InitCodeTransformer.java
...
String entityValue =
  context.getNamer().quoted("[" + Name.from(entityName).toUpperUnderscore() + "]");
```

PR #1619, opened 2017-10-12
Use the type table to enquote placeholder strings

saicheems:
Delete the function `SurfaceNamer.quoted(String)`.


saicheems:
There is a side effect in the Node.js baseline, but I think this commit is correct. Node.js uses single quotes for all other strings, presumably the use of double quotes for the placeholders was an oversight.

PR #1627, opened 2017-10-16
NodeJS: fix batching config

landrito:


landrito:
It was failing in node v6.6.0 for me.

PR #1629, opened 2017-10-17
C# routing header support

chrisdunelm:


chrisdunelm:
I've built pubsub with this change, and confirmed it builds.

chrisdunelm:
This requires an as-yet-unreleased version of gax; so will require syncing with gax and [dependencies.yaml](https://github.com/googleapis/googleapis/blob/master/gapic/packaging/dependencies.yaml#L56).

src/test/java/com/google/api/codegen/testdata/csharp/csharp_library.baseline
```diff
@@ -5079,7 +5079,8 @@ namespace Google.Example.Library.V1
             _callMergeShelves = clientHelper.BuildApiCall<MergeShelvesRequest, Shelf>(
                 GrpcClient.MergeShelvesAsync, GrpcClient.MergeShelves, effectiveSettings.MergeShelvesSettings);
             _callCreateBook = clientHelper.BuildApiCall<CreateBookRequest, Book>(
-                GrpcClient.CreateBookAsync, GrpcClient.CreateBook, effectiveSettings.CreateBookSettings);
+                GrpcClient.CreateBookAsync, GrpcClient.CreateBook, effectiveSettings.CreateBookSettings)
```
jskeet: Would it make sense to put x-goog-request-params into GAX as a constant?

jskeet: Or treat it the same way as field masks - add `CallSettings.FromRequestParamsHeader`?

chrisdunelm: Hmm, yes. That would be more consistent.
But I'm thinking probably not, because we certainly don't want anyone else using this header; and having it as a publicly available method might encourage people to think they should use it. Or at least cause confusion as to why it's there.
In fact, because no-one will ever use this header, I'm now thinking it may well be best to just leave it as it is now.

jskeet: Sure - that sounds fine to me. It's easy enough to change that later and old generated code (with the string) would still work. The reverse approach, not so much...

PR #1630, opened 2017-10-17
 Introduce required config_schema_version field

ethanbao:
conffig_schema_version is a required field to specify the version of GAPIC config schema. It is versioned using Semantic Versioning 2.0.0 (semver) and follows the semver specification. Currently, the valid values is "1.0.0".


src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -96,6 +96,11 @@ message ConfigProto {
 
   // A list of message resource name configurations.
   repeated ResourceNameMessageConfigProto resource_name_generation = 20;
+
+  // A required field to specify the version of GAPIC config schema.
+  // It is versioned using Semantic Versioning 2.0.0 (semver) and follows the
+  // semver specification. Currently, the valid values is "1.0.0".
```
geigerj: nit: 'the valid values is "1.0.0"' -> 'the only valid value is "1.0.0"'

src/test/java/com/google/api/codegen/config/GapicConfigProducerTest.java
```diff
@@ -0,0 +1,53 @@
...
+import com.google.api.tools.framework.model.Diag;
+import com.google.api.tools.framework.model.Model;
+import com.google.api.tools.framework.model.SimpleLocation;
+import com.google.api.tools.framework.model.testing.TestDataLocator;
+import com.google.common.truth.Truth;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+public class GapicConfigProducerTest {
```
geigerj: We already have a config generation test here: https://github.com/googleapis/toolkit/blob/master/src/test/java/com/google/api/codegen/ConfigGenerationTest.java. There are also already existing copies of `library.proto` checked into the `testsrc` directory and a baseline at https://github.com/googleapis/toolkit/blob/master/src/test/java/com/google/api/codegen/testdata/library_config.baseline.

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -228,6 +247,13 @@ public static GapicProductConfig create(
       throw new RuntimeException(e);
     }
 
+    String configSchemaVersion = configProto.getConfigSchemaVersion();
```
geigerj: This validation logic should not be duplicated. If the two `create` methods share logic, they should probably be refactored.

src/main/proto/com/google/api/codegen/config.proto
ethanbao: Done.

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
ethanbao: Agree. I don't have enough knowledge to refactor these two create's into one. In theory, the validation logic can be done in GAPIC config advisor. I left two TODOs to move the checking into GAPIC config, so that there won't be two identical validation logic in this class.

src/test/java/com/google/api/codegen/config/GapicConfigProducerTest.java
garrettjonesgoogle: +1

ethanbao: This is not about generation but parsing.

geigerj: Regardless, we shouldn't need to introduce another copy of library.proto and library.yaml.

ethanbao: The library.proto under testsrc contains "book_from_anywhere.proto", meaning it is not a pure copy of library.proto from googleapis repo. I don't necessarily need to use library.proto, any canonical proto will work. OTOH, I see other test set defined their own testdata subfolder and add proto/yamls there. What is the best practice here in toolkit repo. I am okay with any practice.

ethanbao: Switched to use myproto.proto, which is simpler and good enough for my case. PTAL.

ethanbao:
Ping @geigerj @garrettjonesgoogle

ethanbao:
Thanks for the review. Merging.

PR #1634, opened 2017-10-18
C# Fix casing of rerouting types

chrisdunelm:


chrisdunelm:
OK, I've discovered what's happened.
It's due to this PR: https://github.com/googleapis/toolkit/pull/1622
Which merged a branch into master 5 days ago.

It changes the case handling of `anyCamel()`, hence the change we're seeing. Unfortunately there is no test case for rerouting.
This does also affect a few other C# code-gen things, but they appear to be variable/method names that are internal to the generated code, so the change is fine. Only the one change here is required, as it's for a type that is external to the generated code.

chrisdunelm:
PTAL

PR #1635, opened 2017-10-18
Add xmldoc to remove build warnings

### LINKED ISSUE #1632: C# snippets build warnings
Generated snippets have two issues that cause warnings:

* Missing `.ConfigureAwait(false)`
* Missing XmlDoc

chrisdunelm:
Fixes #1632

### DIFF
--- a/src/main/resources/com/google/api/codegen/csharp/gapic_snippets.snip
+++ b/src/main/resources/com/google/api/codegen/csharp/gapic_snippets.snip
@@ -1,5 +1,6 @@
 @extends "csharp/common.snip"
 
+# snippets: SnippetsFileView
 @snippet generate(snippets)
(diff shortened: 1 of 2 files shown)


chrisdunelm:
Running artman creates all .cs files in the same dir, plus the csproj with warnings-as-errors enabled; so if someone just runs artman then without this there are many compile errors.
It's not essential, but is cleaner generating xmldoc here.

jskeet:
I thought Artman ended up separating the source files out as a final stage? It's intended to. If we're building the snippets into the library, I'd say that's a problem in itself. Maybe we need to think about the bigger picture again.

chrisdunelm:
My mistake.
I meant that running *toolkit* creates all .cs files in the same dir. Artman does indeed move them around.

PR #1637, opened 2017-10-18
Samplegen for discogapic

andreamlin:
Abstracts the TypeRef class into a TypeModel interface, which is consumed by the InitCode transformers and views to generate method samples for the surface.


andreamlin:
PTAL

src/main/java/com/google/api/codegen/config/DiscoGapicMethodConfig.java
```diff
@@ -162,8 +162,13 @@ static DiscoGapicMethodConfig createDiscoGapicMethodConfig(
     }
 
     Iterable<FieldConfig> requiredFieldConfigs =
-        DiscoGapicMethodConfig.createFieldNameConfigs(
-            DiscoGapicMethodConfig.getRequiredFields(
+        createFieldNameConfigs(
```
saicheems: You'll also want to do this below, for the `optionalFieldConfigs`. See `GapicMethodConfig.java:180`.

src/main/java/com/google/api/codegen/config/DiscoveryField.java
```diff
@@ -29,25 +29,40 @@
...
 import java.util.LinkedList;
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import javax.annotation.Nullable;
 
-/** Created by andrealin on 7/31/17. */
-public class DiscoveryField implements FieldModel {
+/** A field declaration wrapper around a Discovery Schema. */
+public class DiscoveryField implements FieldModel, TypeModel {
```
saicheems: Out of curiosity why not split this into `DiscoveryField implements FieldModel` and `DiscoveryTypeRef implements TypeModel`?

src/main/java/com/google/api/codegen/config/DiscoveryField.java
```diff
@@ -29,25 +29,40 @@
...
+      propertiesBuilder.add(DiscoveryField.create(child, discoGapicNamer));
+    }
+    this.properties = propertiesBuilder.build();
+  }
+
+  /* Create a FieldModel object from a non-null Schema object. */
+  public static DiscoveryField create(Schema schema, DiscoGapicNamer discoGapicNamer) {
+    Preconditions.checkNotNull(schema);
+    DiscoveryField discoField = new DiscoveryField(schema, discoGapicNamer);
+    return discoField;
```
saicheems: Can combine this and the above line.

src/main/java/com/google/api/codegen/config/DiscoveryField.java
```diff
@@ -250,4 +245,185 @@ public Oneof getOneof() {
...
+  }
+
+  @Override
+  public DiscoveryField getType() {
+    return this;
+  }
+
+  @Override
+  public boolean isEmptyType() {
+    return false;
```
saicheems: The empty type is an actual type, here's an example:
https://github.com/googleapis/toolkit/blob/master/src/test/java/com/google/api/codegen/testsrc/discoveries/pubsub.v1.json#L26

Dunno if you're already dealing with it somewhere else.

src/main/java/com/google/api/codegen/transformer/MethodContext.java
```diff
@@ -14,14 +14,7 @@
...
 
-import com.google.api.codegen.config.ApiSource;
-import com.google.api.codegen.config.FlatteningConfig;
-import com.google.api.codegen.config.GapicProductConfig;
-import com.google.api.codegen.config.InterfaceConfig;
-import com.google.api.codegen.config.InterfaceModel;
-import com.google.api.codegen.config.MethodConfig;
-import com.google.api.codegen.config.MethodModel;
-import com.google.api.codegen.config.SingleResourceNameConfig;
+import com.google.api.codegen.config.*;
```
saicheems: Should reconfigure your IDE to not do this, I'm assuming you're using IntelliJ,

Editor > Code Style > Java

you can raise the `'*'` import settings to 9999 or something huge.

src/main/java/com/google/api/codegen/transformer/ModelTypeNameConverter.java
```diff
@@ -14,10 +14,9 @@
  */
 package com.google.api.codegen.transformer;
 
-import com.google.api.codegen.config.FieldConfig;
-import com.google.api.codegen.config.FieldModel;
-import com.google.api.codegen.config.InterfaceModel;
-import com.google.api.codegen.config.ProtoInterfaceModel;
+import static com.google.protobuf.DescriptorProtos.FieldDescriptorProto.Type.TYPE_STRING;
```
saicheems: Probably shouldn't import the value itself, but just `Type`. Or something lower in the chain.

src/main/java/com/google/api/codegen/transformer/java/JavaSchemaTypeNameConverter.java
```diff
@@ -208,6 +217,11 @@ public String renderPrimitiveValue(Schema schema, String value) {
     throw new IllegalArgumentException("Schema is of unknown type.");
   }
 
+  @Override
+  public String renderValueAsString(String value) {
```
saicheems: Would be nice to replace calls to this with something like, `renderPrimitiveValue(Bla.STRING, value)`, but it looks like it's much harder with the current method signature of `renderPrimitiveValue`.

src/main/java/com/google/api/codegen/config/DiscoGapicMethodConfig.java
andreamlin: Done.

src/main/java/com/google/api/codegen/config/DiscoveryField.java

src/main/java/com/google/api/codegen/config/DiscoveryField.java
andreamlin: Conceptually, a FieldModel contains an instance of TypeModel. It seemed redundant to create another class that to separately implement TypeModel because the underlying Discovery Schema can take care of filling in the methods for both FieldModel and TypeModel.

src/main/java/com/google/api/codegen/transformer/MethodContext.java
andreamlin: oh thanks for the tip!
there are no more star imports anywhere in the codebase.

src/main/java/com/google/api/codegen/transformer/ModelTypeNameConverter.java
andreamlin: done.

src/main/java/com/google/api/codegen/transformer/java/JavaSchemaTypeNameConverter.java
andreamlin: I agree, it was just a black hole of type wrapping I didn't want to get into.

src/main/java/com/google/api/codegen/config/DiscoveryField.java
andreamlin: cool this has been updated to reflect Empty being a concrete type.

andreamlin:
Addressed open comments. 
PTAL

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
```diff
@@ -60,6 +59,13 @@ public String getKey() {
   }
 
   /*
+   * Get the variable name, which may be configured differently from the key.
+   */
+  public String getVarName() {
```
saicheems: Is this different from the `identifier` field?

src/main/java/com/google/api/codegen/transformer/java/JavaSchemaTypeNameConverter.java
saicheems: A possible approach is to create a static field `STRING` in `Schema` which contains a `Schema` of type `STRING`. Up to you if you want to do that.

saicheems:
I'm okay with it, let Garrett take a look too before merging.

PR #1640, opened 2017-10-19
Delete Discovery samplegen

saicheems:
The contents of this repo as of 2017-10-20 12:40 are captured in
github.com/googleapis/discovery-artifact-manager/toolkit/. This is a
one-time sync to cut development costs on the toolkit repo, which will
continue to evolve. We expect few or no changes to the Discovery
samplegen code, which will henceforth only live in the
discovery-artifact-manager repo, not in toolkit.


garrettjonesgoogle:
LGTM

saicheems:
I think we're all good,

my test running script expects `discoJar` to be in the same directory, which can be generated and copied over. The internal script can be configured by specifying a particular directory for toolkit.

vchudnov-g:
I'll approve this so as to unblock submission, but please address the following before merging:

1. Let's make sure the scripts run successfully by themselves off the new location
2. Please add details to this PR description like you did for the Dartman PR in order to give context as to why this code is being deleted, where it now lives, why, etc.

PR #1642, opened 2017-10-19
C# smoketest generator

chrisdunelm:


chrisdunelm:
I'll be changing artman to move the smoketest files into their own directory.

src/main/resources/com/google/api/codegen/csharp/gapic_smoketest_csproj.snip
```diff
@@ -0,0 +1,22 @@
+# packageMetadata: PackageMetadataView
+@snippet generate(packageMetadata)
+  <?xml version="1.0" encoding="utf-8"?>
+  <Project Sdk="Microsoft.NET.Sdk">
+  
+    <PropertyGroup>
+      <TargetFramework>netcoreapp2.0</TargetFramework>
```
jskeet: Any reason for targeting 2 rather than 1?

src/main/resources/com/google/api/codegen/csharp/gapic_smoketest.snip
```diff
@@ -0,0 +1,70 @@
...
+                        Console.WriteLine();
+                        return 1;
+                    }
+                    string projectId = args[0];
+
+                @end
+                // Create client
+                {@smoke.apiClassName} client = {@smoke.apiClassName}.Create();
+
+                @if smoke.hasMethod
```
jskeet: Do we only support a single test method per service?

chrisdunelm: Apparently yes, only a single test per service.

src/main/resources/com/google/api/codegen/csharp/gapic_smoketest_csproj.snip
chrisdunelm: No reason. Changed to 1.0

PR #1650, opened 2017-10-23
Fix Node.js JSDoc examples. Fixes #1649

### LINKED ISSUE #1649: Node.js: example captions need to be one line or they break
For example:

```html
<caption>Install the client library with
  <a href="https://www.npmjs.com/">npm</a>:</caption>
```
needs to be changed to
```html
<caption>Install the client library with <a href="https://www.npmjs.com/">npm</a>:</caption>
```

jmdobry:
Unfortunately the captions don't render correctly on cloud.google.com if they're not on one line.

### DIFF
--- a/src/test/java/com/google/api/codegen/testdata/nodejs/nodejs_multiple_services.baseline
+++ b/src/test/java/com/google/api/codegen/testdata/nodejs/nodejs_multiple_services.baseline
@@ -130,21 +130,17 @@ const gapic = Object.freeze({
  * @module {object} @google-cloud/multiple-services
  * @alias nodejs-multiple-services
  *
- * @example <caption>Install the client library with
- *   <a href="https://www.npmjs.com/">npm</a>:</caption>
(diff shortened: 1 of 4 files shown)


src/test/java/com/google/api/codegen/testdata/nodejs/nodejs_multiple_services.baseline
```diff
@@ -130,21 +130,17 @@ const gapic = Object.freeze({
  * @module {object} @google-cloud/multiple-services
  * @alias nodejs-multiple-services
  *
- * @example <caption>Install the client library with
- *   <a href="https://www.npmjs.com/">npm</a>:</caption>
+ * @example <caption>Install the client library with <a href="https://www.npmjs.com/">npm</a>:</caption>
  * npm install --save @google-cloud/multiple-services
  *
  * @example <caption>Import the client library:</caption>
  * const multiple-services = require('@google-cloud/multiple-services');
```
jmdobry: btw `const multiple-services = require('@google-cloud/multiple-services');` is not valid javascript. The variable needs to be changed to camelCase.

PR #1654, opened 2017-10-24
all: make SurfaceNamer use InterfaceConfig more

pongad:
Go needs InterfaceConfig in a bunch of places to get surface renaming
right.
This commit makes SurfaceNamer use InterfaceConfig instead of
InterfaceModel where possible.
The two places where this isn't possible are

- grpc stubs
- grpc clients

because they might refer to rerouted interfaces that we don't
have configs for.


pongad:
Adding @andreamlin for recent refactor and @garrettjonesgoogle for MVVM.

Since the changes are pretty mechanical, I don't think per-lang reviewers are necessary but feel free to add others.

PR #1659, opened 2017-10-26
Changes to compile against gax-httpjson

garrettjonesgoogle:


src/main/java/com/google/api/codegen/config/GapicInterfaceConfig.java
```diff
@@ -274,11 +274,7 @@ public GapicMethodConfig getMethodConfig(Method method) {
   }
 
   public GapicMethodConfig getMethodConfig(String methodSimpleName, String fullName) {
```
andreamlin: is `fullName` still required as a parameter?

src/main/java/com/google/api/codegen/transformer/StaticLangApiMethodTransformer.java
```diff
@@ -690,8 +690,6 @@ private void setCallableMethodFields(
       callableImplType = ApiCallableImplType.of(methodConfig.getGrpcStreamingType());
     } else if (methodConfig.isBatching()) {
       callableImplType = ApiCallableImplType.BatchingApiCallable;
-    } else if (methodConfig.isLongRunningOperation()) {
-      callableImplType = ApiCallableImplType.InitialOperationApiCallable;
```
andreamlin: Why is this `if {}` block being removed?

src/main/resources/com/google/api/codegen/java/settings.snip
```diff
@@ -464,7 +465,7 @@
    * Builder for {@xsettingsClass.name}.
    */
   public static class Builder extends ClientSettings.Builder {
-    private final ImmutableList<UnaryCallSettings.Builder> unaryMethodSettingsBuilders;
+    private final ImmutableList<UnaryCallSettings.Builder<?, ?>> unaryMethodSettingsBuilders;
```
andreamlin: Why aren't the UnaryCallSettings type pararmeters here aren't predictable?

src/main/java/com/google/api/codegen/config/GapicInterfaceConfig.java
garrettjonesgoogle: Looks like it's not, I'll remove it!

src/main/java/com/google/api/codegen/transformer/StaticLangApiMethodTransformer.java
garrettjonesgoogle: Because now we don't need to distinguish SimpleApiCallable and InitialOperationApiCallable; the callable for initiating operations will now just be a SimpleApiCallable.

src/main/resources/com/google/api/codegen/java/settings.snip
garrettjonesgoogle: because this is a list of settings for all of the API methods, which each have different request and response types.

garrettjonesgoogle:
Addressed feedback, PTAL

garrettjonesgoogle:
I made an update because I missed a few create() -> of() renames, PTAL

Issue #1652, opened 2017-10-23
Ruby: exclude all GAPIC files from rubocop until style issues are fixed


geigerj:
### What

GAPIC files have rubocop violations (tracked in #1120 and #1536). Until these are fixed, GAPIC files should be excluded from the rubocop check so as not to cause CI failures.

geigerj:
Closed by #1660

Issue #1536, opened 2017-09-01
Rubocop offenses for top-level API file


geigerj:
### Context

https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/1702

### What

Fix style-guide violations:

```
Offenses:



lib/google/cloud/video_intelligence.rb:54:81: C: Line is too long. [95/80]

      FILE_DIR = File.realdirpath(Pathname.new(__FILE__).join("..").join("video_intelligence"))

                                                                                ^^^^^^^^^^^^^^^

lib/google/cloud/video_intelligence.rb:57:9: C: Align the operands of an expression in an assignment spanning multiple lines.

        .select {|file| File.directory?(file)}

        ^^^^^^^

lib/google/cloud/video_intelligence.rb:57:17: C: Space between { and | missing.

        .select {|file| File.directory?(file)}

                ^^

lib/google/cloud/video_intelligence.rb:57:46: C: Space missing inside }.

        .select {|file| File.directory?(file)}

                                             ^

lib/google/cloud/video_intelligence.rb:58:9: C: Align the operands of an expression in an assignment spanning multiple lines.

        .select {|dir| Google::Gax::VERSION_MATCHER.match(File.basename(dir))}

        ^^^^^^^

lib/google/cloud/video_intelligence.rb:58:17: C: Space between { and | missing.

        .select {|dir| Google::Gax::VERSION_MATCHER.match(File.basename(dir))}

                ^^

lib/google/cloud/video_intelligence.rb:58:78: C: Space missing inside }.

        .select {|dir| Google::Gax::VERSION_MATCHER.match(File.basename(dir))}

                                                                             ^

lib/google/cloud/video_intelligence.rb:59:9: C: Align the operands of an expression in an assignment spanning multiple lines.

        .select {|dir| File.exist?(dir + ".rb")}

        ^^^^^^^

lib/google/cloud/video_intelligence.rb:59:17: C: Space between { and | missing.

        .select {|dir| File.exist?(dir + ".rb")}

                ^^

lib/google/cloud/video_intelligence.rb:59:48: C: Space missing inside }.

        .select {|dir| File.exist?(dir + ".rb")}

                                               ^

lib/google/cloud/video_intelligence.rb:60:9: C: Align the operands of an expression in an assignment spanning multiple lines.

        .map {|dir| File.basename(dir)}

        ^^^^

lib/google/cloud/video_intelligence.rb:60:14: C: Space between { and | missing.

[...]

Issue #1120, opened 2017-03-17
Ruby: Generate rubocop config and make sure generated clients pass rubocop.


landrito:
Currently, when running the rubocop task on the generated client. Many errors are found. We should resolve this by doing two things: 

- [ ] Generate a custom rubocop.yml to make rubocop more reasonable for generated clients. This should be based on the rubocop.ymls already present in google-cloud-ruby.
- [ ] Make the generated code pass for the generated clients using the custom rubocop.yml
- [ ] Add rubocop rake task

geigerj:
Bumping to P1 -- this is more important to have as we more forward with GAPIC-only releases.

landrito:
Hey @geigerj, I think you did some work similar to this recently. Is this still relevant? If so, what is left to be done?

geigerj:
Yup, still relevant. Unfortunately what I did recently was in the opposite direction; I excluded even more files from rubocop checking.

geigerj:
Reducing to P2, as this does not qualify as "release blocking".

hzyi-google:
Transferred to ruby microgenerator.

PR #1660, opened 2017-10-26
Various small Ruby GAPIC fixes

### LINKED ISSUE #1559: Ruby files should end in newline
### What

SnippetSet doesn't by default output a trailing newline. We should ensure one is added for all files. (For example, `Gemfile` currently lacks the trailing newline.) This isn't strictly a bug, but it does go against widespread convention.

geigerj:
- Ensure files end in newline 
- Style fix (double quotes) in Gemfile
- A workaround where a newer version of `public_suffix` breaks the google-cloud-ruby build
- Rename `ci:smoke_test` Rake task to `ci:acceptance` so that it runs correctly in google-cloud-ruby CI
- Exclude all generated files from rubocop due to numerous autogen style violations ("real" fixes tracked in #1536, #1120)
- YARD `@overloads` tag requires a method signature

Fixes #1559, #1652

### DIFF
--- a/src/main/resources/com/google/api/codegen/ruby/rubocop.yml.snip
+++ b/src/main/resources/com/google/api/codegen/ruby/rubocop.yml.snip
@@ -2,9 +2,8 @@
   AllCops:
     Exclude:
       - "{@metadata.identifier}.gemspec"
-      - "lib/{@metadata.protoPath}/**/*"
-      # This should be removed in the future after we are sanitizing client files.
-      - "lib/{@metadata.versionPath}/**/*"
+      # This should be removed in the future once GAPIC files conform more closely to rubocop checks.
+      - "lib/google/**/*"
       - "Rakefile"
       - "test/**/*"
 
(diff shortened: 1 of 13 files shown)


src/main/resources/com/google/api/codegen/ruby/rubocop.yml.snip
```diff
@@ -2,9 +2,8 @@
   AllCops:
     Exclude:
       - "{@metadata.identifier}.gemspec"
-      - "lib/{@metadata.protoPath}/**/*"
-      # This should be removed in the future after we are sanitizing client files.
-      - "lib/{@metadata.versionPath}/**/*"
+      # This should be removed in the future once GAPIC files conform more closely to rubocop checks.
```
landrito: nit: are we fine having this be specific to the google namespace rather than the the generic first namespace of a package?

geigerj: Agreed that this should be generic, but this is also just a workaround to get CI to pass in google-cloud-ruby until #1536 and #1120 are fixed, so leaving as-is.

PR #1661, opened 2017-10-26
go: add support for header-based routing

saicheems:
Configure "x-goog-request-params" before performing a request.

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/StaticLangApiMethodTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/StaticLangApiMethodTransformer.java
@@ -489,6 +489,7 @@ public class StaticLangApiMethodTransformer {
     } else {
       methodViewBuilder.hasReturnValue(method.hasReturnValue());
     }
+    methodViewBuilder.headerRequestParams(ApiCallableTransformer.getHeaderRequestParams(context));
   }
 
   protected void setServiceResponseTypeName(
--- a/src/main/resources/com/google/api/codegen/go/doc.snip
+++ b/src/main/resources/com/google/api/codegen/go/doc.snip
@@ -21,12 +21,9 @@
         "google.golang.org/grpc/metadata"
     )
 
-    func insertMetadata(ctx context.Context, md map[string][]string) context.Context {
+    func insertMetadata(ctx context.Context, md metadata.MD) context.Context {
         out, _ := metadata.FromOutgoingContext(ctx)
-        out = out.Copy()
-        for k, v := range md {
-            out[k] = v
-        }
+        out = metadata.Join(out, md)
         return metadata.NewOutgoingContext(ctx, out)
     }
 
--- a/src/main/resources/com/google/api/codegen/go/doc.snip
+++ b/src/main/resources/com/google/api/codegen/go/doc.snip
@@ -21,11 +21,13 @@
         "google.golang.org/grpc/metadata"
     )
 
-    func insertXGoog(ctx context.Context, val []string) context.Context {
-        md, _ := metadata.FromOutgoingContext(ctx)
-        md = md.Copy()
-        md["x-goog-api-client"] = val
-        return metadata.NewOutgoingContext(ctx, md)
+    func insertMetadata(ctx context.Context, md map[string][]string) context.Context {
+        out, _ := metadata.FromOutgoingContext(ctx)
+        out = out.Copy()
+        for k, v := range md {
+            out[k] = v
+        }
+        return metadata.NewOutgoingContext(ctx, out)
     }
 
     // DefaultAuthScopes reports the default set of authentication scopes to use with this package.
--- a/src/main/resources/com/google/api/codegen/go/doc.snip
+++ b/src/main/resources/com/google/api/codegen/go/doc.snip
@@ -21,9 +21,9 @@
         "google.golang.org/grpc/metadata"
     )
 
-    func insertMetadata(ctx context.Context, md metadata.MD) context.Context {
+    func insertMetadata(ctx context.Context, mds ...metadata.MD) context.Context {
         out, _ := metadata.FromOutgoingContext(ctx)
-        out = metadata.Join(out, md)
+        out = metadata.Join(append([]metadata.MD{out}, mds...)...)
         return metadata.NewOutgoingContext(ctx, out)
     }
 
--- a/src/main/resources/com/google/api/codegen/go/doc.snip
+++ b/src/main/resources/com/google/api/codegen/go/doc.snip
@@ -23,7 +23,12 @@
 
     func insertMetadata(ctx context.Context, mds ...metadata.MD) context.Context {
         out, _ := metadata.FromOutgoingContext(ctx)
-        out = metadata.Join(append([]metadata.MD{out}, mds...)...)
+        out = out.Copy()
+        for _, md := range mds {
+            for k, v := range md {
+                out[k] = append(out[k], v...)
+            }
+        }
         return metadata.NewOutgoingContext(ctx, out)
     }
 
--- a/src/main/resources/com/google/api/codegen/go/main.snip
+++ b/src/main/resources/com/google/api/codegen/go/main.snip
@@ -462,14 +462,15 @@
 @end
 
 @private mergeMetadata(method)
-    ctx = insertMetadata(ctx, c.Metadata)
     @if method.headerRequestParams
         md := metadata.Pairs(
             @join headerRequestParam : method.headerRequestParams
                 "x-goog-request-params", fmt.Sprintf("%s=%v", "{@headerRequestParam.fullyQualifiedName}", req.{@requestParamGetter(headerRequestParam.gettersChain)})
             @end
         )
-        ctx = insertMetadata(ctx, md)
+        ctx = insertMetadata(ctx, c.Metadata, md)
+    @else
+        ctx = insertMetadata(ctx, c.Metadata)
     @end
 @end
 
--- a/src/main/resources/com/google/api/codegen/go/main.snip
+++ b/src/main/resources/com/google/api/codegen/go/main.snip
@@ -461,13 +461,24 @@
     @end
 @end
 
+@private headerRequestParamString(headerRequestParams) horizontal on EMPTY
+    # This if block prevents a space from being printed after '"'.
+    @if TRUE
+        fmt.Sprintf("
+    @end
+    @join headerRequestParam : headerRequestParams on "&"
+        %s=%v
+    @end
+    ",
+    @join headerRequestParam : headerRequestParams on ", "
+        "{@headerRequestParam.fullyQualifiedName}", req.{@requestParamGetter(headerRequestParam.gettersChain)}
+    @end
+    )
+@end
+
 @private mergeMetadata(method)
     @if method.headerRequestParams
-        md := metadata.Pairs(
-            @join headerRequestParam : method.headerRequestParams
-                "x-goog-request-params", fmt.Sprintf("%s=%v", "{@headerRequestParam.fullyQualifiedName}", req.{@requestParamGetter(headerRequestParam.gettersChain)})
-            @end
-        )
+        md := metadata.Pairs("x-goog-request-params", {@headerRequestParamString(method.headerRequestParams)})
         ctx = insertMetadata(ctx, c.Metadata, md)
     @else
         ctx = insertMetadata(ctx, c.Metadata)
--- a/src/main/resources/com/google/api/codegen/go/main.snip
+++ b/src/main/resources/com/google/api/codegen/go/main.snip
@@ -69,7 +69,7 @@
         CallOptions *{@view.callOptionsTypeName}
 
         // The metadata to be sent with each request.
-        xGoogHeader []string
+        Metadata metadata.MD
     }
 
     // {@view.clientConstructorName} creates a new {@view.servicePhraseName} client.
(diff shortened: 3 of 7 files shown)


src/main/java/com/google/api/codegen/transformer/StaticLangApiMethodTransformer.java
```diff
@@ -489,6 +489,7 @@ private void setCommonFields(
     } else {
       methodViewBuilder.hasReturnValue(method.hasReturnValue());
     }
+    methodViewBuilder.headerRequestParams(ApiCallableTransformer.getHeaderRequestParams(context));
```
saicheems: Is this ok? I'm calling into the `ApiCallableTransformer` from here, don't know if that's considered ugly or not. Anyway, I need the `headerRequestParams` field to be part of the method to do this in Go.

src/test/java/com/google/api/codegen/testdata/go/go_library.baseline
```diff
@@ -100,10 +100,12 @@ import (
     "google.golang.org/grpc/metadata"
 )
 
-func insertXGoog(ctx context.Context, val []string) context.Context {
+func insertMetadata(ctx context.Context, m map[string][]string) context.Context {
```
saicheems: Note this is changed to accept an arbitrary metadata map.

src/test/java/com/google/api/codegen/testdata/go/go_library.baseline
```diff
@@ -100,11 +100,13 @@ import (
     "google.golang.org/grpc/metadata"
 )
 
-func insertXGoog(ctx context.Context, val []string) context.Context {
-    md, _ := metadata.FromOutgoingContext(ctx)
-    md = md.Copy()
-    md["x-goog-api-client"] = val
-    return metadata.NewOutgoingContext(ctx, md)
+func insertMetadata(ctx context.Context, md map[string][]string) context.Context {
```
saicheems: Note this is changed to accept an arbitrary metadata map.

src/test/java/com/google/api/codegen/testdata/go/go_library.baseline
```diff
@@ -459,7 +476,14 @@ func (c *Client) MergeShelves(ctx context.Context, req *librarypb.MergeShelvesRe
...
 func (c *Client) CreateBook(ctx context.Context, req *librarypb.CreateBookRequest, opts ...gax.CallOption) (*librarypb.Book, error) {
-    ctx = insertXGoog(ctx, c.xGoogHeader)
+    vals = []string{}
+    vals = append(vals, fmt.Sprintf("%s=%v", "name", req.GetName())
+    vals = append(vals, fmt.Sprintf("%s=%v", "book.read", req.GetBook().GetRead())
+    md := map[string][]string{
+        "x-goog-api-client": c.xGoogHeader,
+        "x-goog-request-params" = vals,
+    }
+    ctx = insertMetadata(ctx, md)
```
saicheems: This method demonstrates the new header.

src/test/java/com/google/api/codegen/testdata/go/go_library.baseline
```diff
@@ -459,7 +476,14 @@ func (c *Client) MergeShelves(ctx context.Context, req *librarypb.MergeShelvesRe
...
 func (c *Client) CreateBook(ctx context.Context, req *librarypb.CreateBookRequest, opts ...gax.CallOption) (*librarypb.Book, error) {
-    ctx = insertXGoog(ctx, c.xGoogHeader)
+    vals = []string{}
+    vals = append(vals, fmt.Sprintf("%s=%v", "name", req.GetName())
+    vals = append(vals, fmt.Sprintf("%s=%v", "book.read", req.GetBook().GetRead())
+    md := map[string][]string{
+        "x-goog-api-client": c.xGoogHeader,
+        "x-goog-request-params": vals,
+    }
+    ctx = insertMetadata(ctx, md)
```
saicheems: This method demonstrates the new header.

src/test/java/com/google/api/codegen/testdata/go/go_library.baseline
```diff
@@ -459,7 +476,14 @@ func (c *Client) MergeShelves(ctx context.Context, req *librarypb.MergeShelvesRe
 
 // CreateBook creates a book.
 func (c *Client) CreateBook(ctx context.Context, req *librarypb.CreateBookRequest, opts ...gax.CallOption) (*librarypb.Book, error) {
-    ctx = insertXGoog(ctx, c.xGoogHeader)
+    vals = []string{}
+    vals = append(vals, fmt.Sprintf("%s=%v", "name", req.GetName())
+    vals = append(vals, fmt.Sprintf("%s=%v", "book.read", req.GetBook().GetRead())
```
saicheems: Note that this sequence of getters cannot panic because nils are typed.

saicheems:
Hey all,

I mostly threw this together to demonstrate how this can work. As far as design goes... I'm sure it could be better. If you have any suggestions on how to organize my additions to the baseline, I'll be happy to incorporate.

src/test/java/com/google/api/codegen/testdata/go/go_library.baseline
```diff
@@ -459,7 +476,14 @@ func (c *Client) MergeShelves(ctx context.Context, req *librarypb.MergeShelvesRe
 
 // CreateBook creates a book.
 func (c *Client) CreateBook(ctx context.Context, req *librarypb.CreateBookRequest, opts ...gax.CallOption) (*librarypb.Book, error) {
-    ctx = insertXGoog(ctx, c.xGoogHeader)
+    vals = []string{}
+    vals = append(vals, fmt.Sprintf("%s=%v", "name", req.GetName()))
+    vals = append(vals, fmt.Sprintf("%s=%v", "book.read", req.GetBook().GetRead()))
```
saicheems: Note that this sequence of getters cannot panic because nils are typed.

jba:
It looks to me like the header request params are statically configured strings. Where's the part where they are combined with the client's current values? E.g. `fmt.Sprintf("projects/%s", client.projectID)`.

saicheems:
I'm not exactly sure what you mean, do our generated clients do substitution of client variables?
The way I thought this would work is as follows:

User creates a request:
```
req = Foo{name: "projects/my-project/foos/my-foo", ...}
```

User calls a function
```
foo, err := c.CreateFoo(ctx, req)
```

Outgoing request has these headers:
```
HEADERS
=======
"x-goog-api-client":...
"x-goog-request-params":name=projects/my-project/foos/my-foo&...
```

jba:
OK, take this with a grain of salt, because I may be confused.

The only example I have of this is Firestore. They want the resource to be the database resource name, `projects/P/databases/D`. That is true even if that string is not a request field; for instance if you're working on a document, the name field would be something like `projects/P/databases/D/documents/...`. So it's actually worse than what I said, you'd have to take a field from the request and maybe parse it, or at least take a prefix.

saicheems:
Ohhh I took a look at their proto, I see what you're saying. I feel that's something that we shouldn't be concerned about at our level though... it's going to be very difficult to decide how to or if to parse optional IDs in a collection.

For anyone else, here's an example request:
```
// The request for [Firestore.CreateDocument][google.firestore.v1beta1.Firestore.CreateDocument].
message CreateDocumentRequest {
  // The parent resource. For example:
  // `projects/{project_id}/databases/{database_id}/documents` or
  // `projects/{project_id}/databases/{database_id}/documents/chatrooms/{chatroom_id}`
  string parent = 1;

  // The collection ID, relative to `parent`, to list. For example: `chatrooms`.
  string collection_id = 2;

  // The client-assigned document ID to use for this document.
  //
  // Optional. If not specified, an ID will be assigned by the service.
  string document_id = 3;

  // The document to create. `name` must not be set.
  Document document = 4;

  // The fields to return. If not set, returns all fields.
  //
  // If the document has a field that is not present in this mask, that field
  // will not be returned in the response.
  DocumentMask mask = 5;
}
```

^ There's multiple ways to form a resource because there are optional IDs that can be specified by the caller.

Generally, I think this feature would be used to route to different regions, so the most important thing would be getting `us-west-1` or something like that in the headers...

saicheems:
Basically I'm saying that this would be configured by the API owner, and we can pass as many fields for a function into the request as they like. I'd say that it's Google backend's responsibility to route correctly, and they have more context on how to parse resource names at that point than we do.

src/main/resources/com/google/api/codegen/go/doc.snip
```diff
@@ -21,11 +21,13 @@
...
-        md, _ := metadata.FromOutgoingContext(ctx)
-        md = md.Copy()
-        md["x-goog-api-client"] = val
-        return metadata.NewOutgoingContext(ctx, md)
+    func insertMetadata(ctx context.Context, md map[string][]string) context.Context {
+        out, _ := metadata.FromOutgoingContext(ctx)
+        out = out.Copy()
+        for k, v := range md {
+            out[k] = v
+        }
```
pongad: `metadata.Join(out, md)`? No need to `Copy`, `Join` always allocates a new one.

src/test/java/com/google/api/codegen/testdata/go/go_library.baseline
```diff
@@ -459,7 +476,14 @@ func (c *Client) MergeShelves(ctx context.Context, req *librarypb.MergeShelvesRe
...
+    md := map[string][]string{
+        "x-goog-api-client": c.xGoogHeader,
+        "x-goog-request-params": vals,
+    }
+    ctx = insertMetadata(ctx, md)
```
pongad: I'm not too excited about creating one more map every call. Do you think we can make `c.xGoogHeader` a `metadata.MD` instead?

Then the calls become
```go
insertMetadata(ctx, c.xGoogHeader)
```
and
```go
reqParams := metadata.Pairs("x-goog-request-params", vals)
insertMetadata(ctx, c.xGoogHeader, reqParams)
```

src/test/java/com/google/api/codegen/testdata/go/go_library.baseline
pongad: I think we need the colon-equal operator to declare `vals` right?
Also consider slice literal:
```go
vals := []string{
  fmt.Sprintf("%s=%v", "name", req.GetName()),
  fmt.Sprintf("%s=%v", "book.read", req.GetBook().GetRead()),
}
```

src/test/java/com/google/api/codegen/testdata/go/go_library.baseline
saicheems: If we're going to change the type of `XGoogHeader`, can we also change it's name? `XGoogHeader` is not quite accurate anymore, if there are multiple `x-goog-*` headers... Maybe `Metadata` instead?

src/test/java/com/google/api/codegen/testdata/go/go_library.baseline
saicheems: Done

src/main/resources/com/google/api/codegen/go/doc.snip
saicheems: Good point, done

src/main/resources/com/google/api/codegen/go/main.snip
```diff
@@ -124,7 +124,7 @@
     func (c *{@view.clientTypeName}) {@setClientInfoFunc(view)}(keyval ...string) {
         kv := append([]string{"gl-go", version.Go()}, keyval...)
         kv = append(kv, "gapic", version.Repo, "gax", gax.Version, "grpc", grpc.Version)
-        c.xGoogHeader = []string{gax.XGoogHeader(kv...)}
+        c.Metadata = metadata.Pairs("x-goog-api-client", gax.XGoogHeader(kv...))
```
saicheems: Actually I think this is wrong... will probably have to cast a `map[string][]string` instead.

saicheems:
OK, PTAL

src/test/java/com/google/api/codegen/testdata/go/go_library.baseline
```diff
@@ -100,11 +100,10 @@ import (
...
 )
 
-func insertXGoog(ctx context.Context, val []string) context.Context {
-    md, _ := metadata.FromOutgoingContext(ctx)
-    md = md.Copy()
-    md["x-goog-api-client"] = val
-    return metadata.NewOutgoingContext(ctx, md)
+func insertMetadata(ctx context.Context, mds ...metadata.MD) context.Context {
+    out, _ := metadata.FromOutgoingContext(ctx)
+    out = metadata.Join(append([]metadata.MD{out}, mds...)...)
```
saicheems: Was a bit nicer in the library baseline if this accepted an arbitrary number of MDs, but it's a bit ugly here as a result. LMK if you're ok with it or want me to revert.

saicheems: I'm doing this mostly to avoid reconstructing the metadata maps over and over again, see the source of [`metadata.Join`](https://github.com/grpc/grpc-go/blob/master/metadata/metadata.go#L100).

pongad: You mean for `insertMetadata` to be variadic? I totally agree with this.

`Join` is a little inefficient (I wrote it a while back and still feel a little bad). It's commonly used by multiple goroutines though, so modifying an existing one is a little confusing.

jba:
@pongad It doesn't say that anywhere, but in conversations with Firestore team, they said that's what they want. They want routing by database, even for requests that refer only to documents or collections.

saicheems:
If that's the case then we can set the `parent` field to be included in the header, and in their backend they can parse out the `databaseId` right? It doesn't seem like our responsibility in the client to deal with that.

jba:
@saicheems That seems to be how it's going down.

pongad:
still lgtm

saicheems:
Feel free to approve, anyone, if you're cool with merging.

src/main/java/com/google/api/codegen/transformer/StaticLangApiMethodTransformer.java
eoogbe: That's kinda ugly. Wouldn't it be better to extract a `HeaderRequestParamsTransformer` that both transformers use?

saicheems: Sounds good, as discussed will leave to you to change in your follow up.

PR #1671, opened 2017-11-06
Rename Rakefile tasks

geigerj:
#1660 renamed some tasks to introduce `ci:acceptance`, but in the wrong
place. The acceptance task should be runnable in all cases, even when
there are no smoke tests.

### DIFF
--- a/src/main/resources/com/google/api/codegen/ruby/Rakefile.snip
+++ b/src/main/resources/com/google/api/codegen/ruby/Rakefile.snip
@@ -88,7 +88,11 @@
   @# Acceptance tests
   desc "Run the {@metadata.identifier} acceptance tests."
   task :acceptance do
-    puts "The {@metadata.identifier} gem does not have acceptance tests."
+    @if metadata.hasSmokeTests
+      Rake::Task["smoke_test"].invoke
+    @else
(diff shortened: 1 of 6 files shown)


landrito:
Should this also make the generated smoke tests run on `rake smoke_test`?

geigerj:
@landrito Isn't that already the case with the task defined here: https://github.com/googleapis/toolkit/pull/1671/files#diff-fdaf77e823c0b9f1416f7386beb45fc2R60

landrito:
Blah, brain fart. LGTM

landrito:
Oh wait, I retract my LGTM. 

I mean to say these two tasks should be merged.

https://github.com/googleapis/toolkit/blob/master/src/main/resources/com/google/api/codegen/ruby/Rakefile.snip#L62

https://github.com/googleapis/toolkit/blob/master/src/main/resources/com/google/api/codegen/ruby/Rakefile.snip#L90

geigerj:
@landrito I don't know about merged, but I made smoke tests are a subset of acceptance. (Especially with partial GAPICs, we might expect that there are manual, non-smoke test acceptance tests that you might want to be able to invoke separately.)

landrito:
Hm. Okay. I think I buy that a distinction is still needed.

PR #1672, opened 2017-11-06
Java: updating for builder parameterization, other

garrettjonesgoogle:


src/main/resources/com/google/api/codegen/java/settings.snip
```diff
@@ -149,15 +150,15 @@
...
-          .setEndpoint(getDefaultEndpoint());
+      return {@xsettingsClass.instantiatingChannelProvider}.newBuilder();
     }
 
     public static TransportChannelProvider defaultTransportChannelProvider() {
       return {@xsettingsClass.defaultTransportProviderBuilder}().build();
     }
   @end
 
+  @@BetaApi("The surface for customizing headers is not stable yet and my change in the future.")
```
michaelbausor: s/my/may

src/test/java/com/google/api/codegen/testdata/discogapic/java/java_simplecompute.v1.json.baseline
```diff
@@ -5324,6 +5321,7 @@ public class AddressSettings extends ClientSettings {
   }
 
 
+  @BetaApi("A restructuring of stub classes is planned, so this may break in the future")
```
michaelbausor: This is repeated in many places. Could/should it be a function?

garrettjonesgoogle: Hmm... I kind of like having this concrete instead of abstracted.

src/main/resources/com/google/api/codegen/java/settings.snip
garrettjonesgoogle: oy...

PR #1675, opened 2017-11-08
Java: surface cleanup

garrettjonesgoogle:


src/main/resources/com/google/api/codegen/java/settings.snip
```diff
@@ -176,23 +176,6 @@
...
-  @@Deprecated
-  public static Builder defaultBuilder() {
-    return Builder.createDefault();
-  }
-
-  /**
-   * Returns a builder for this class with recommended defaults for API methods, and the given
-   * ClientContext used for executor/transport/credentials.
-   */
-  @@Deprecated
```
pongad: are we keeping this for only one release?

garrettjonesgoogle: It's actually been deprecated for several releases. It was marked in https://github.com/GoogleCloudPlatform/google-cloud-java/pull/2429 and released in 0.23.1 as deprecated.

PR #1681, opened 2017-11-09
Java: updating set of services with no resource names

garrettjonesgoogle:
- longrunning is no longer a package we generate
- trace v2 has resource names


garrettjonesgoogle:
I just added it in https://github.com/googleapis/toolkit/pull/1669 so that the api-common dependency could be removed from some proto-generated packages in java.

michaelbausor:
SG thanks!

PR #1682, opened 2017-11-09
go: make metadata private

pongad:
This is technically a breaking change;
it was accidentally made public before.


src/main/resources/com/google/api/codegen/go/main.snip
```diff
@@ -69,7 +69,7 @@
         CallOptions *{@view.callOptionsTypeName}
 
         // The metadata to be sent with each request.
```
saicheems: Might be worth being more specific with the docstring if you're reducing the scope to `x-goog-*` headers only.

pongad: Done

PR #1683, opened 2017-11-10
Update discogapic 

andreamlin:
Discogapic pipeline needs to be updated to reflect gax-java changes.
Mostly just fixing imports.


andreamlin:
PTAL

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceTransformer.java
```diff
@@ -676,6 +675,9 @@ private void addSettingsImports(InterfaceContext context) {
     typeTable.saveNicknameFor("javax.annotation.Generated");
     typeTable.saveNicknameFor("org.threeten.bp.Duration");
 
+    // TODO(andrealin): remove this from httpjson
+    typeTable.saveNicknameFor("com.google.api.gax.grpc.GrpcExtraHeaderData");
```
garrettjonesgoogle: I don't quite understand this TODO - this class isn't even accessible from httpjson.

andreamlin: Ok, this import has been applied only to proto-input surfaces.

andreamlin:
PTAL

src/main/resources/com/google/api/codegen/java/settings.snip
```diff
@@ -162,8 +162,11 @@
   public static ApiClientHeaderProvider.Builder defaultApiClientHeaderProviderBuilder() {
     return ApiClientHeaderProvider.newBuilder()
         .setGeneratorHeader(DEFAULT_GAPIC_NAME, getGapicVersion())
-        .setApiClientHeaderLineKey("x-goog-api-client")
-        .addApiClientHeaderLineData(GrpcExtraHeaderData.getXGoogApiClientData());
+        @if xsettingsClass.transportProtocol == "GRPC"
+          .setApiClientHeaderLineKey("x-goog-api-client")
```
garrettjonesgoogle: We want the header line key for httpjson also.

andreamlin: Can this be hardcoded as HTTP/1.1, given how stable that config will be?

garrettjonesgoogle: I just meant that x-goog-api-client should be included for discogapic clients. I don't fully understand your question.

andreamlin: Oh gotcha. Moved the x-goog-api-client line out of the @if block.

andreamlin:
PTAL

PR #1697, opened 2017-11-16
Node.js: Accept point versions.

lukesneeringer:
This was making vision `v1p1beta1` toss files in the wrong place.


eoogbe:
There are literally, two open PRs to try to address this. #1545 uses `VersionMatcher` which is a better solution. I believe they haven't been merged because they didn't have accompanying tests. If we use the existing `VersionMatcher`, the existing tests that already test it should be sufficient. @landrito Please fix the merge conflict in #1545 and merge that PR. It has two approvals.

PR #1702, opened 2017-11-20
Remove deprecated parameters

geigerj:
Also, allow initializer to read the service address and port from
self.class in case these constants have been overridden.


geigerj:
Note: requires https://github.com/googleapis/gax-ruby/pull/109

PR #1710, opened 2017-11-22
Use underscore for smoke test env var

geigerj:
Previously the variable name was derived directly from the package
config, which could use an incorrect separator for multiword API names
(e.g., "video-intelligence" instead of "video_intelligence").

See similar fix for artman at googleapis/artman#344.


geigerj:
@landrito Good call. Added tests.

geigerj:
@landrito PTAL?

PR #1714, opened 2017-11-27
Determine x-goog-api-client GAPIC version dynmaically

### LINKED ISSUE #1701: Ruby: Load x-goog-api-client gapic version from gemspec
### What

The GAPIC version sent in the x-goog-api-client header should be loaded from the gemspec, rather than hardcoded into the generated service client file. One way to do this might be to do:

```
  ...
  google_api_client << " gapic/#{Gem.loaded_specs[{@metadata.identifier}]"
  ...
```

### Why

Hardcoding does not work if the gemspec is maintained manually, as happens in practice for google-cloud-ruby.

geigerj:
Previously the GAPIC version in the x-goog-api-client header was
hardcoded. This can lead to stale values being recorded if the library
is modified manually after generation, or if versioning is maintained
manually. Instead, compute the version to send dynamically based on the
installed gem metadata.

Fixes #1701


geigerj:
@landrito Could you take a look at this when you have a chance?

src/test/java/com/google/api/codegen/testdata/ruby/ruby_doc_library.baseline
```diff
@@ -2607,9 +2607,11 @@ module Library
           updater_proc = credentials.updater_proc
         end
 
+        package_version = Gem.loaded_specs['library'].version.version
```
landrito: This looks wrong? Can you double check this?

geigerj: What about it?

geigerj: (If it's the double `version`, that's actually how it works... the first `.version` returns a `Gem::Version`, and the second returns a string.)

geigerj:
Yep, sanity checked by hand-editing this change into the video-intelligence library and confirming that the right parameter is passed to GAX.

Issue #1585, opened 2017-09-21
NodeJS: through2 incorrectly being added as a dependency 


landrito:
if an API doesn't have a streaming method it is incorrect to add through2 as a dependency.

eoogbe:
@landrito Can you add a priority?

PR #1717, opened 2017-11-28
NodeJS: fix incorrect dependency on through2

landrito:
Through2 should not be used if the api does not have a streaming method.

Fixes: #1585


JustinBeckwith:
@alexander-fenster @lukesneeringer this has been sitting around for quite a while :) Can we merge it?

src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSPackageMetadataTransformer.java
```diff
@@ -215,34 +216,34 @@ private ViewModel generateMetadataView(
...
 
-  private boolean hasLongrunning(ApiModel model, ProductConfig productConfig) {
-    for (InterfaceModel apiInterface : model.getInterfaces()) {
-      if (productConfig.getInterfaceConfig(apiInterface).hasLongRunningOperations()) {
-        return true;
-      }
+  private List<PackageDependencyView> generateDevDependencies(
+      ApiModel model, ProductConfig productConfig) {
+    ImmutableList.Builder<PackageDependencyView> dependencies = ImmutableList.builder();
+    dependencies.add(PackageDependencyView.create("mocha", VersionBound.create("3.2.0", "")));
```
alexander-fenster: hey, I hate this: this feels very wrong to have version numbers hardwired in the code, even though it's Node-specific class.

alexander-fenster:
I don't really know the code and I don't want to sound picky but I'd rather leave the `package.json` content and especially versions of the dependencies to snip files. I might as well be very wrong here so feel free to approve and merge but I'd like someone to express the second opinion :)  @lukesneeringer, @eoogbe?

eoogbe:
I would prefer to move as much of the hard coded stuff as possible to the snippets. In fact, that is going to be a part of the view simplification refactoring for Ruby. That said, the current standard is to keep as much stuff as possible out of the snippets and inside the transformers. Your call if you want to keep things more consistent with the rest of the codebase or change to my preference.

garrettjonesgoogle:
Potentially dependencies seem like a good thing to keep out of the snippets, because they are a point of frequent updates and they probably need to be updated together.

PR #1719, opened 2017-11-28
NodeJS: change '*/' in comments

### LINKED ISSUE #615: Possible end of block comments not being detected.
### What

Some proto-file comments are causing `/* */` style block comments to end preemptively.
### Example Comment

From [LRO operations.proto](https://github.com/googleapis/googleapis/blob/master/google/longrunning/operations.proto#L50)

``` proto
  // NOTE: the `name` binding below allows API services to override the binding
  // to use different resource name schemes, such as `users/*/operations`.
```

The `users/*/operations` comment causes any block comment it is inside to end.
### Example Gapic Output

Node:

``` js
/**
 * Lists operations that match the specified filter in the request. If the
 * server doesn't support this method, it returns `UNIMPLEMENTED`.
 *
 * NOTE: the `name` binding below allows API services to override the binding
 * to use different resource name schemes, such as `users/*/operations`
 *
 * @param {Object} request
 *   The request object that will be sent.
 * @param {string} request.name
 *   The name of the operation collection.
 * @param {string} request.filter
 *   The standard list filter.
 *  . . .
 */
```

landrito:
This helps avoid incorrectly ended block comments.

Fixes #615


alexander-fenster:
I would prefer replacing */ with *[space]/ if it does not break anything else. Putting HTML escape sequence to Javascript code does not make much sense for me (and for users). I agree that both solutions are actually bad.

landrito:
Done.

PR #1728, opened 2017-11-30
Document mixins

### LINKED ISSUE #603: _doc does not generate the documents for mixin protos
`python_doc_message_library.baseline`, `ruby_doc_message_library.baseline`, and `nodejs_doc_message_library.baseline` don't have any references about tagger.proto. It seems the message generator does nothing on the mixin protos.

This would be problematic when generating documents for pubsub. Cc: @geigerj @bjwatson @swcloud

eoogbe:
Fixes #603

### DIFF
--- a/src/main/java/com/google/api/codegen/ProtoFileView.java
+++ b/src/main/java/com/google/api/codegen/ProtoFileView.java
@@ -14,55 +14,65 @@
  */
 package com.google.api.codegen;
 
+import com.google.api.codegen.config.GapicProductConfig;
+import com.google.api.codegen.config.InterfaceConfig;
+import com.google.api.codegen.config.MethodConfig;
+import com.google.api.codegen.config.MethodModel;
+import com.google.api.codegen.config.ProtoTypeRef;
 import com.google.api.tools.framework.model.Field;
-import com.google.api.tools.framework.model.Interface;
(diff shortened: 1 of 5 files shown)


eoogbe:
@geigerj PTAL

src/main/java/com/google/api/codegen/ProtoFileView.java
```diff
@@ -14,55 +14,65 @@
...
+import com.google.api.tools.framework.model.TypeRef;
 import java.util.Set;
+import java.util.TreeSet;
 
 /**
  * A file-based view of model, consisting of a strategy for getting the protocol buffer files from
  * which the model was constructed.
  */
-public class ProtoFileView implements InputElementView<ProtoFile> {
+public class ProtoFileView {
```
geigerj: Why is it necessary to break away from this interface? It's surprising that `ProtoFileView` would not implement `InputElementView<ProtoFile>`. Shouldn't mixins be part of the Model via the mixin config aspect?

eoogbe: `reroute_to_grpc_interface` would not be included if I depend solely on the `Model`.

The `InputElementView` interface is legacy from pre-MVVM. It remains because client config is still not migrated, but #1724 will change that.

PR #1735, opened 2017-12-05
add more stuff to package.json

alexander-fenster:
Since nodejs packages now go to separate repos, let's make package.json generated correctly so that we don't need to edit it manually after generating.

### DIFF
--- a/src/main/resources/com/google/api/codegen/nodejs/package.json.snip
+++ b/src/main/resources/com/google/api/codegen/nodejs/package.json.snip
@@ -1,10 +1,14 @@
 @snippet generate(metadata)
   {
-    "repository": "GoogleCloudPlatform/google-cloud-node",
-    "name": "{@metadata.identifier}",
-    "version": "{@metadata.packageVersionBound.lower}",
-    "author": "{@metadata.author}",
+    "name": "@@google-cloud/{@metadata.identifier}",
     "description": "{@metadata.fullName} client for Node.js",
+    "version": "{@metadata.packageVersionBound.lower}",
+    "license": "Apache-2.0",
+    "author": "{@metadata.author}",
+    "engines": {
+      "node": ">=4.0.0"
+    },
+    "repository": "googleapis/nodejs-{@metadata.shortName}",
     "main": "src/index.js",
     "files": [
       "protos",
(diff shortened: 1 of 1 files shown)


src/main/resources/com/google/api/codegen/nodejs/package.json.snip
```diff
@@ -1,10 +1,14 @@
 @snippet generate(metadata)
   {
-    "repository": "GoogleCloudPlatform/google-cloud-node",
-    "name": "{@metadata.identifier}",
+    "name": "@@google-cloud/{@metadata.identifier}",
```
landrito: This should be based on if teh domain layer location set in teh gapic client is set to google-cloud.

src/main/resources/com/google/api/codegen/nodejs/package.json.snip
```diff
@@ -25,23 +29,38 @@
       "{@metadata.shortName}",
       "{@metadata.fullName}"
     ],
+    "scripts": {
+      "cover": "nyc --reporter=lcov mocha --require intelli-espower-loader test/*.js && nyc report",
+      "docs": "repo-tools exec -- jsdoc -c .jsdoc.js",
+      "generate-scaffolding": "repo-tools generate all && repo-tools generate lib_samples_readme -l samples/ --config ../.cloud-repo-tools.json",
+      "lint": "repo-tools lint --cmd eslint -- src/ samples/ system-test/ test/",
+      "prettier": "repo-tools exec -- prettier --write src/*.js src/*/*.js samples/*.js samples/*/*.js test/*.js test/*/*.js system-test/*.js system-test/*/*.js",
+      "samples-test": "cd samples/ && npm link ../ && npm test && cd ../",
```
landrito: Is this always a valid script. I assume this will be added to the package.json when samples are introduced to the repo.

src/main/resources/com/google/api/codegen/nodejs/package.json.snip
```diff
@@ -25,23 +29,38 @@
...
+      "test": "repo-tools test run --cmd npm -- run cover"
+    },
     "dependencies": {
       @join packageDep : metadata.dependencies on ",".add(BREAK)
         "{@packageDep.name}": "^{@packageDep.versionBound.lower}"
       @end
     },
     "devDependencies": {
-      "mocha": "^3.2.0",
+      "@@google-cloud/nodejs-repo-tools": "^2.1.0",
```
landrito: https://github.com/googleapis/toolkit/pull/1717 would make this be more similar to how dependencies are added above.

src/main/resources/com/google/api/codegen/nodejs/package.json.snip
```diff
@@ -1,10 +1,14 @@
 @snippet generate(metadata)
   {
-    "repository": "GoogleCloudPlatform/google-cloud-node",
-    "name": "{@metadata.identifier}",
+    "name": "@@google-cloud/{@metadata.identifier}",
+    "description": "{@metadata.fullName} client for Node.js",
     "version": "{@metadata.packageVersionBound.lower}",
+    "license": "Apache-2.0",
```
landrito: This should use the {@metadata.licenseName} which is specified in gapic config.

alexander-fenster:
Note: Tests changes pending after the package.json format is agreed.

src/main/resources/com/google/api/codegen/nodejs/package.json.snip
alexander-fenster: It might not be always valid (obviously, if there is no samples, the scripts will not work). I'm still considering different options where this stuff can be put.

lukesneeringer: It would be valid to have repo-tools add the samples-test item if and only if it found a `samples/` directory. (Samples will often be added _post hoc_ so toolkit literally can not know.)

src/main/resources/com/google/api/codegen/nodejs/package.json.snip
```diff
@@ -25,23 +29,38 @@
...
       @join packageDep : metadata.dependencies on ",".add(BREAK)
         "{@packageDep.name}": "^{@packageDep.versionBound.lower}"
       @end
     },
     "devDependencies": {
-      "mocha": "^3.2.0",
+      "@@google-cloud/nodejs-repo-tools": "^2.1.0",
+      "async": "^2.5.0",
+      "codecov": "^3.0.0",
+      "eslint": "^4.9.0",
```
lukesneeringer: These are not really properties of the package itself. It would be valid to output this tool and use `semistandard` instead of `prettier`, for example. Can we have repo-tools add this sort of thing, so we put out a generic package that any third party could run through their preferred tooling?

src/main/resources/com/google/api/codegen/nodejs/package.json.snip
lukesneeringer: I do think it is valid to keep repo-tools and the `.cloud-repo-tools.json` file here, and delegate other _post hoc_ automation to repo-tools. Someone who wants to do something else can just not use it, and the cruft is minimal.

Thoughts?

PR #1740, opened 2017-12-11
Fix smoke test generation in PHP

michaelbausor:
- Also fix cross-language issue with PROJECT_ID reference in fields
other than the first field


michaelbausor:
@landrito PTAL for Ruby output

PR #1743, opened 2017-12-13
Make some small changes to setup.py

lukesneeringer:
Fixes some of the items raised by @jonparrott in https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4577.


src/main/resources/com/google/api/codegen/py/setup.py.snip
```diff
@@ -1,6 +1,7 @@
-@extends "common.snip"
+@extends "py/common.snip"
 
 @snippet generate(metadata)
+  #{@licenseSection(metadata.fileHeader)}
```
geigerj: This line should probably be removed.

PR #1746, opened 2017-12-14
Fix PHP empty service docs

michaelbausor:
Fixes https://github.com/googleapis/toolkit/issues/1433
Supercedes https://github.com/googleapis/toolkit/pull/1411

Adds generation of no_path_templates API to test.


michaelbausor:
Gentle ping on this

src/main/java/com/google/api/codegen/transformer/ServiceTransformer.java
```diff
@@ -34,7 +36,20 @@ public ServiceDocView generateServiceDoc(
       docLines.add("");
       docLines.addAll(namer.getDocLines(conf.getManualDoc()));
     }
-    serviceDoc.lines(docLines.build());
+    List<String> lines = docLines.build();
```
vchudnov-g: For encapsulation, would it make sense to keep `serviceDoc.lines(docLines.build())` here and move the logic below into `firstLine() and `remainingLines()`

vchudnov-g: Oh, I see the recommendation in #1411 was to not do this. I would prefer to do it, but we don't need to re-litigate.

src/main/java/com/google/api/codegen/transformer/ServiceTransformer.java
```diff
@@ -34,7 +36,20 @@ public ServiceDocView generateServiceDoc(
...
+      if (lines.size() > 0) {
+        serviceDoc.firstLine(lines.get(0));
+      } else {
+        serviceDoc.firstLine("");
+      }
+      if (lines.size() > 1) {
+        serviceDoc.remainingLines(lines.subList(1, lines.size()));
+      } else {
+        serviceDoc.remainingLines(new ArrayList<String>());
+      }
```
pongad:
```java
if (lines.isEmpty()) {
  serviceDoc.firstLine("");
  serviceDoc.remainingLines(Collections.emptyList());
} else {
  serviceDoc.firstLine(lines.get(0));

  // I think this works even if the size is 1. I also think we should copy the list, since sublist is an "image" of the original: changes in one is visible in another.
  serviceDoc.remainingLines(ImmutableList.copyOf(lines.subList(1, lines.size()));
}
```
?

michaelbausor: You are right - updated to something very similar.

I think that we don't need to copy - we are working with `ImmutableList` objects, so we don't need to worry about potential changes. Is that right, or am I not understanding the implications of sublist correctly?

src/main/java/com/google/api/codegen/transformer/ServiceTransformer.java
michaelbausor: I will leave as-is for now.

src/main/java/com/google/api/codegen/transformer/ServiceTransformer.java
pongad: Oh yeah you're right.

pongad:
LGTM

michaelbausor:
Gentle ping on this - I need an approval to merge.

PR #1747, opened 2017-12-14
Move client construction out of try block

michaelbausor:
Fixes https://github.com/googleapis/toolkit/issues/1173


michaelbausor:
@vam-google No, it doesn't. That part of the issue is fixed by requiring those dependencies in the composer.json file.

PR #1754, opened 2017-12-19
PHP: Add x-goog-request-params header support (header routing)

vam-google:


src/test/java/com/google/api/codegen/testdata/php/php_library.baseline
```diff
@@ -926,6 +927,14 @@ class LibraryServiceGapicClient
                 $optionalArgs['retrySettings']
             );
         }
+        $requestParams = new RequestParamsHeaderDescriptor([
+          'name' => $request->getName(),
+          'book.read' => $request->getBook()->getRead(),
+        ]);
+        $optionalArgs['userHeaders'] = isset($optionalArgs['userHeaders'])
+            ? array_merge($optionalArgs['userHeaders'], $requestParams->getHeader())
```
michaelbausor: To confirm the expected behaviour: if the user has manually specified the `x-goog-request-params` header, we will overwrite it with our own header.

vam-google: Modified the order, as we agreed.

michaelbausor:
LGTM

Issue #1704, opened 2017-11-21
Ruby: Smoke tests should use standard google cloud env variables for cloud APIs


geigerj:
### What

Currently the Ruby rake smoke test task attempts to use the environment variable `SMOKE_TEST_PROJECT` to determine the project to use when running smoke tests. For non-cloud APIs, that's a fine default, but if we know it's a cloud API, we should use `GCLOUD_TEST_PROJECT || {API}_TEST_PROJECT`.

### Why

Consistency with existing APIs in google-cloud-ruby.

geigerj:
`{API}_TEST_PROJECT` is now accepted. Reclassify as enhancement to support `GCLOUD_TEST_PROJECT`.

jbolinger:
No longer relevant given the customizations available in https://github.com/googleapis/gapic-generator-ruby.

PR #1773, opened 2017-12-27
Rakefile fixes

### LINKED ISSUE #1750: Ruby: Improve Rakefile smoke_test task
### What

The Rake task `ci:smoke_test` is unused and should be removed from the Rakefile template.

geigerj:
* Remove ci:smoke_test task
* Fix some env variable logic

Updates #1704 (only supports Cloud APIs)
Fixes #1750


geigerj:
Don't merge until updated to reflect https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/1920/files.

geigerj:
Updated and ready for review.

src/main/java/com/google/api/codegen/transformer/ruby/RubyImportSectionTransformer.java
```diff
@@ -62,6 +63,21 @@ public ImportSectionView generateTestImportSection(GapicInterfaceContext context
     return importSection.build();
   }
 
+  public ImportSectionView generateRakefileAcceptanceTaskImportSection(
+      List<GapicInterfaceContext> contexts) {
+    List<ImportFileView> none = ImmutableList.of();
+    ImportSectionView.Builder importSection = ImportSectionView.newBuilder();
+    importSection.standardImports(none);
+    importSection.externalImports(none);
+    List<ImportFileView> credentialsImports = new ArrayList<>();
```
landrito: This should be a set since multiple interfaces can map to the same credentials class.

geigerj: Good catch! Fixed, PTAL.

geigerj:
Sanity check revealed that the credentials imports generated for the `Rakefile` are incorrect and should not contain a version module. Holding off merge to fix.

geigerj:
@landrito Sanity check with container now works. PTAL before I merge?

landrito:
LGTM.

PR #1775, opened 2017-12-28
Update sample generation for optional args with sample init fields

michaelbausor:
This PR:
  - Adds a sample code init field for an optional argument
  - Stops automatically adding all required fields to the sampleCodeInitFields list
  - Stops filtering sample initialization trees by initFields(), which excludes optional args
  - Stops setting sampleCodeInitFields for the response object, which is a bug
  - Fixes snippets for Ruby and Node

### DIFF
--- a/src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
+++ b/src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
@@ -28,7 +28,6 @@ import com.google.common.collect.Lists;
 import com.google.protobuf.DescriptorProtos.FieldDescriptorProto.Type;
 import java.util.ArrayList;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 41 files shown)


michaelbausor:
Fixes https://github.com/googleapis/toolkit/issues/1706

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
```diff
@@ -238,16 +237,7 @@ private static InitValueConfig mergeInitValueConfig(
           newSubTrees.add(InitCodeNode.createWithValue(nameString, initValueConfig));
         }
       }
-      // Filter subTrees using fieldSet
```
geigerj: What does this filtering do, and why should it no longer happen?

src/main/java/com/google/api/codegen/transformer/TestCaseTransformer.java
```diff
@@ -270,7 +271,7 @@ private InitCodeContext createResponseInitCodeContext(
         .initObjectType(outputType)
         .symbolTable(symbolTable)
         .suggestedName(Name.from("expected_response"))
-        .initFieldConfigStrings(context.getMethodConfig().getSampleCodeInitFields())
+        .initFieldConfigStrings(ImmutableList.<String>of())
```
geigerj: What does this change do?

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
michaelbausor: This filtering ensured that **only** the set of fields listed in `context.initFields()` would be given an initialized value. Removing the filtering means that while all fields listed in `context.initFields()` will still get an intialized value, fields that are not listed can also be given initialized values.

The problem this fixes is that `context.initFields()` contains the required fields, but not the optional fields. So even if an optional field is given an sample_code_init_value, it would not be initialized because it will be filtered out.

src/main/java/com/google/api/codegen/transformer/TestCaseTransformer.java
michaelbausor: This stops using the sampleCodeInitFields when initializing the expected response object for generated tests. This is simply a bugfix - it makes no sense to use the sampleCodeInitFields, which are for the **request** object, when initializing the **response** object, which will have completely different fields.

PR #1777, opened 2017-12-28
PHP: Remove aliasing of nested types

michaelbausor:
Fixes https://github.com/googleapis/toolkit/issues/1438

This PR removes the aliasing currently used by nested messages and enums. While the aliasing makes our samples much prettier, it also makes them broken for users, who do not know to add the correct "use X_Y as Y" statements in order to make the samples work.


michaelbausor:
@vam-google gentle ping on this

src/main/java/com/google/api/codegen/transformer/php/PhpModelTypeNameConverter.java
```diff
@@ -174,11 +173,11 @@ private TypeName getTypeName(ProtoElement elem, int maxDepth) {
       }
 
       String parentFullName = getTypeName(parent, maxDepth - 1).getFullName();
-      String nickname = elem.getSimpleName();
-      String fullName = String.format("%s_%s", parentFullName, nickname);
+      String fullName = String.format("%s_%s", parentFullName, elem.getSimpleName());
+      String[] nameSplit = fullName.split("\\\\");
```
vam-google: Since you need only the last element something like the following may fit better (also allows to avoid double-escaping of the backslash):
```java
String nickName = fullName.substring(fullName.lastIndexOf("\\") + 1);
```

michaelbausor: Done

PR #1781, opened 2018-01-03
go: stop generating path functions

pongad:
The path functions are not terribly useful and adds much weight to the
surface.
We'll re-implement these functions in existing clients to not break
users.


pongad:
This creates a lot of dead code in toolkit. I'll remove them in a subsequent PR.

jba:
I don't care if tests have things like `fmt.Sprintf("shelves/%s", "[SHELF_ID]")`, but I think examples should just use constant strings. I can't tell if this change affects examples.

pongad:
@jba I agree. This change doesn't affect examples. Currently examples always have empty request structs. Currently the init code is somewhat Java-centric and doesn't map well to Go.

For example, it's easy to generate
```go
x := 5
req := &Request{
  X: x,
}
```
Unfortunately, it's much harder to generate the more idiomatic
```go
req := &Request{
  X: 5,
}
```

I think making init code use constant strings should be a part of the effort to make better examples. Are you OK with this?

jba:
@pongad Yes.

PR #1787, opened 2018-01-08
PHP HTTP 1.1/JSON Support

dwsupplee:
This PR introduces HTTP 1.1/JSON support for PHP. It utilizes the [HttpRule option](https://github.com/googleapis/googleapis/blob/master/google/firestore/v1beta1/firestore.proto#L59) to derive request information.

In progress (but otherwise ready for review!):

- [x] Update baseline snippets
- [x] Update test snippets

/cc @michaelbausor @bshaffer

What do you guys think about the approach I have taken with including the descriptor/rest config views in the `PhpGapicSurfaceTransformer`? Do you think they warrant their own transformers?


src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
```diff
@@ -228,6 +255,117 @@ private ViewModel buildClientViewModel(GapicInterfaceContext context) {
...
+    HttpRule httpRule = getHttpRule(method.getOptionFields());
+    String body = httpRule.getBody();
+
+    httpMethodMap.put("get", httpRule.getGet());
+    httpMethodMap.put("put", httpRule.getPut());
+    httpMethodMap.put("delete", httpRule.getDelete());
+    httpMethodMap.put("post", httpRule.getPost());
+    httpMethodMap.put("patch", httpRule.getPatch());
+
+    for (Map.Entry<String, String> entry : httpMethodMap.entrySet()) {
```
michaelbausor: This feels like an odd way to use a map - we don't do any lookups, just use it as a way to create a 2-tuple. I am not sure what the correct restructuring is - maybe this whole for loop should be a helper function?

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
```diff
@@ -228,6 +255,117 @@ private ViewModel buildClientViewModel(GapicInterfaceContext context) {
...
+  private RestPlaceholderConfigView generateRestPlaceholderConfigView(
+      GapicMethodContext context, String var) {
+    SurfaceNamer namer = context.getNamer();
+    MethodModel methodModel = context.getMethodConfig().getMethodModel();
+    RestPlaceholderConfigView.Builder placeholderView = RestPlaceholderConfigView.newBuilder();
+    String[] getters = var.split("\\.");
+    String fullName = context.getSurfaceInterfaceContext().getInterface().getFullName();
+
+    for (int i = 0; i < getters.length; i++) {
+      getters[i] =
```
michaelbausor: I think I would prefer adding to a new list instead of modifying the string array in place.

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
```diff
@@ -228,6 +255,117 @@ private ViewModel buildClientViewModel(GapicInterfaceContext context) {
...
+          namer.getFieldGetFunctionName(methodModel.getInputType(), Name.anyLower(getters[i]));
+    }
+
+    if (fullName.equals("google.longrunning.Operations")) {
```
michaelbausor: It would be great to have a comment here explaining what is going on with this special case, perhaps with a link to a relevant proto in googleapis

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
```diff
@@ -292,4 +434,27 @@ private void addApiImports(GapicInterfaceContext context) {
 
     return apiMethods;
   }
+
+  private String generateConfigOutputPath(GapicInterfaceContext context, String name) {
+    GapicInterfaceConfig interfaceConfig = context.getInterfaceConfig();
+    String outputPath =
+        pathMapper.getOutputPath(
+            context.getInterfaceModel().getFullName(), context.getProductConfig());
+    return outputPath.substring(0, outputPath.length() - 5)
```
michaelbausor: This looks odd to me - why do we need to manipulate the path like this?

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
```diff
@@ -292,4 +434,27 @@ private void addApiImports(GapicInterfaceContext context) {
...
+        + "/resources/"
+        + Name.upperCamel(interfaceConfig.getInterfaceModel().getSimpleName())
```
michaelbausor: We should probably push naming logic like this into the SurfaceNamer class, even if we need to create a new unimplemented method in the base class and implement it in the PhpSurfaceNamer.

src/main/java/com/google/api/codegen/transformer/php/PhpSurfaceNamer.java
```diff
@@ -121,7 +121,7 @@ public boolean shouldImportRequestObjectParamType(FieldModel field) {
 
   @Override
   public String getRetrySettingsTypeName() {
-    return "\\Google\\ApiCore\\RetrySettings";
+    return "RetrySettings";
```
michaelbausor: Looking at the code, it seems very odd that this method even exists in the SurfaceNamer, since it is only implemented and used in the PhpSurfaceNamer, and in fact the RetrySettings class name is hardcoded elsewhere. WDYT of just removing this method from SurfaceNamer and PhpSurfaceNamer?

src/main/java/com/google/api/codegen/viewmodel/DescriptorConfigView.java
```diff
@@ -0,0 +1,82 @@
...
+  public abstract String outputPath();
+
+  public abstract boolean hasPageStreamingMethods();
+
+  public abstract boolean hasBatchingMethods();
+
+  public abstract boolean hasLongRunningOperations();
+
+  public boolean hasGrpcStreamingMethods() {
+    return grpcStreamingDescriptors().size() > 0;
```
michaelbausor: Can we be consistent with the `hasX` methods? Either have all of them be set manually (like hasPageStreamingMethods, hasBatchingMethods, hasLongRunningOperations) or have all of them check the length of the corresponding list field (like hasGrpcStreamingMethods)

michaelbausor:
Re the transformers - I think including them in `PhpGapicSurfaceTransformer` is good.

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
dwsupplee: Originally, the code looked like this:

```java
return outputPath
        + "/../resources/"
        + Name.upperCamel(interfaceConfig.getInterfaceModel().getSimpleName())
            .join(name)
            .toLowerUnderscore()
        + ".php";
```

However, at generation time I was receiving the following error:

`java.io.FileNotFoundException: /Users/dave/dev/google-php-cloud-speech-v1beta1/src/Speech/V1beta1/Gapic/../resources/speech_rest_client_config.php (No such file or directory)`

This workaround solved the issue for me, however, I am looking into some other cleaner options :).

src/main/java/com/google/api/codegen/viewmodel/LongRunningOperationDetailView.java
```diff
@@ -35,6 +36,9 @@
 
   public abstract String methodName();
 
+  @Nullable
+  public abstract String upperCamelMethodName();
```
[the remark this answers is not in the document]
dwsupplee: `transportMethodName` sounds great. My apprehension in using `grpcMethodName` was due to just that.

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
```diff
@@ -292,4 +471,36 @@ private void addApiImports(GapicInterfaceContext context) {
...
+  }
+
+  private HttpRule getHttpRule(Map<FieldDescriptor, Object> optionFields) {
+    for (FieldDescriptor fieldDescriptor : optionFields.keySet()) {
+      if (fieldDescriptor.getFullName().equals("google.api.http")) {
+        return (HttpRule) optionFields.get(fieldDescriptor);
+      }
+    }
+
+    throw new IllegalStateException("A HttpRule option must be defined.");
```
michaelbausor: Another possibility we could consider - if there is no HttpRule option defined, instead of throwing a generation error, we just don't support the REST transport for that method, and we throw a runtime error if someone tries to use it, the same way we do for streaming. Is this:
1. Sensible, and does it ever come up?
2. Worth the effort?

src/main/java/com/google/api/codegen/viewmodel/DescriptorConfigView.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.List;
+import javax.annotation.Nullable;
+
+@AutoValue
+public abstract class DescriptorConfigView implements ViewModel {
+  public abstract String templateFileName();
+
+  public abstract List<PageStreamingDescriptorView> pageStreamingDescriptors();
+
+  @Nullable
```
michaelbausor: Why is this field nullable, but all others are not and use an empty list?

src/test/java/com/google/api/codegen/testdata/php/php_library.baseline
```diff
@@ -2728,7 +2319,323 @@ class LibraryServiceClient extends LibraryServiceGapicClient
...
+            ],
+        ]
+    ]
+];
+============== file: src/Example/Library/V1/resources/library_service_rest_client_config.php ==============
+<?php
+
+return [
+    'interfaces' => [
+        'google.example.library.v1.Labeler' => [
```
michaelbausor: A question about this approach - why do we need to place methods that use a rerouteToGrpcInterface setting in a different array? Can we not follow the approach of the settings json file and put them under the main interface (see https://github.com/googleapis/toolkit/pull/1787/files#diff-36f32ff1b2217a381ca1eaeeea45dfe6R2298) an then have a rerouteToInterface setting in the array itself? It seems like this would be consistent, and would also simplify the java implementation.

src/main/java/com/google/api/codegen/viewmodel/DescriptorConfigView.java
dwsupplee: I followed the standard set [here]( 
https://github.com/googleapis/toolkit/blob/master/src/main/java/com/google/api/codegen/viewmodel/DynamicLangXApiView.java#L56-L57). I agree it does not seem there is a good reason. Should we update this and the DynamicLangXApiView?

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
dwsupplee: It seems if an HttpRule isn't defined and the method is not streaming that it would indicate an issue in the protos that needs to be resolved. I like the idea of notifying us of that being a potential issue early, rather than down the line when a user may run into it.

src/test/java/com/google/api/codegen/testdata/php/php_library.baseline
dwsupplee: The interface name gets passed into the `Call` class [here](https://github.com/googleapis/gax-php/blob/json-transport/src/ApiCore/GapicClientTrait.php#L235). We use that combination of the interface name/method name to register the correct gRPC call or to perform a lookup for REST configuration data.

In order to achieve what you mention for the case of REST, I believe we would need to have transport specific logic living in the trait to pass in the main interface name instead. If possible, I would like to keep transport specific logic out of the trait.

Is there a purpose to the "interfaces" key in the .json config if we end up putting everything under the same blanket?

dwsupplee: Thought on this a little more, WDYT about modifying the `Call` class to look like the following?

```php
$call = new Call(
    $interface,
    $method,
    $decodeType,
    $message,
    $descriptor,
    $callType,
    $rerouteToInterface
);
```

Then in the transport implementations, if `$rerouteToInterface` is set use that over the `$interface` when constructing the request?

src/main/java/com/google/api/codegen/viewmodel/DescriptorConfigView.java
michaelbausor: Yes, lets do that. If you find that removing the nullability from the DynamicLangXApiView class, then leave as-is

src/test/java/com/google/api/codegen/testdata/php/php_library.baseline
michaelbausor: Hmm, but why is this transport dependent? Both gRPC and REST need to support it. How does grpc do it now?

bshaffer: I don't understand why having multiple interfaces is a bad idea. Isn't the call actually made to a separate service URL, so it makes sense? I feel like `rerouteToInterface` may be a gRPC-specific idea, or at least I'm confused by it.

dwsupplee: > Hmm, but why is this transport dependent? Both gRPC and REST need to support it. How does grpc do it now?

gRPC needs the `rerouteToInterface` when triggering [`\Grpc\BaseStub::_simpleRequest`](https://github.com/googleapis/gax-php/blob/json-transport/src/ApiCore/Transport/GrpcTransport.php#L146), but REST only really needs it in order to look up the correct URI template in the configuration file.

My original concern was if we move everything under one interface, we would need to do something like the following in the trait:

```php
    if ($transportType === 'REST') {
        $interfaceName = $this->serviceName;
    }

    $call = new Call(
        $this->buildMethod($interfaceName, $methodName),
        $decodeType,
        $request,
        $descriptor,
        $callType
    );
```

But this was superseded by the suggestion which followed.

>I don't understand why having multiple interfaces is a bad idea. Isn't the call actually made to a separate service URL, so it makes sense?

I side with this. I am having trouble understanding why currently everything lives under one interface (in the config.json, for example), even if they technically may be defined under another.

michaelbausor: I don't think having multiple interfaces is a bad idea, but I suppose I didn't understand why it was necessary for REST when it isn't (or doesn't seem to be) necessary for grpc.

It doesn't call a separate service URL. In fact in the REST case, I think the final result is completely unchanged. We only use the interface name as a key into the settings: https://github.com/googleapis/gax-php/blob/json-transport/src/ApiCore/RequestBuilder.php#L74

In the grpc case, we need it to make the _simpleRequest call. E.g. for IAM we are replicating this call: https://github.com/googleapis/gax-php/blob/json-transport/src/Cloud/Iam/V1/IAMPolicyGrpcClient.php#L68

I didn't mean to say the current approach is definitely wrong, only that it seemed different and slightly more complex than the grpc case and I didn't understand why.

michaelbausor: Sorry, my last comment did not see yours @dwsupplee

dwsupplee: No worries!

I think if we do keep multiple interfaces, it should be under the condition we eventually do the same for the descriptor and json config. If that seems like something which is off the table, does https://github.com/googleapis/toolkit/pull/1787#discussion_r162453900 seem reasonable?

michaelbausor: I think that comment looks reasonable, especially splitting the interface and method name. Rather than a separate parameter for rerouteToInterface, what about having that as part of the descriptor array?

dwsupplee: That sounds like a nice idea. For clarification on my end, what do we qualify a descriptor as? Is it a more concrete concept in toolkit (with batchDescriptors, pageDescriptors, etc), or can we use it more generically for "anything which describes this call?"

michaelbausor: It is more generic. It should be (I believe) features of a call that are fixed/permanent (and perhaps overridable by the user, but perhaps not). In the gapic yaml config, the rerouteToGrpcInterface setting (see here: https://github.com/googleapis/googleapis/blob/master/google/pubsub/v1/pubsub_gapic.yaml#L601) is the same sort of setting as page_streaming settings: https://github.com/googleapis/googleapis/blob/master/google/pubsub/v1/pubsub_gapic.yaml#L560

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
michaelbausor: That makes sense to me.

dwsupplee:
/cc @lukesneeringer, @geigerj, @alexander-fenster:

Hi guys! This PR introduced a few minor changes to the baselines for your languages. We just wanted to touch base and make sure everything looks good.

In order to support [decode types in GAPIC calls](https://github.com/GoogleCloudPlatform/google-cloud-php/blob/transport/src/Bigtable/V2/Gapic/BigtableGapicClient.php#L357) we needed to expose [the output type](https://github.com/googleapis/toolkit/pull/1787/files#diff-a819888b211ce238161de05ef1571023R106). This looks like it may have had  the side effect of some unused imports being introduced in python baselines, does the following seem acceptable?:

https://github.com/googleapis/toolkit/pull/1787/commits/e679f97e27da0c0e52317b1a9da513781993f00c

The googleapis submodule was updated, and it introduced a few minor changes to ruby/node doc baselines: 

https://github.com/googleapis/toolkit/pull/1787/commits/f9ff20fa41214b2c94c5f10c251722df4bf99a26

Thanks!

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTestTransformer.java
```diff
@@ -95,23 +90,7 @@
...
-        hasLongRunningOperations = true;
-      }
-    }
-    if (hasLongRunningOperations) {
-      InterfaceModel apiInterface = apiModel.getInterface("google.longrunning.Operations");
-      GapicInterfaceContext context =
-          createContext(apiInterface, productConfig, PhpSurfaceNamer.TestKind.UNIT);
-      views.add(createMockServiceImplView(context));
-    }
+
```
michaelbausor: Nit: remove newline

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
```diff
@@ -105,8 +133,11 @@ public PhpGapicSurfaceTransformer(
         context.withNewTypeTable(context.getNamer().getGapicImplNamespace());
 
     List<ViewModel> surfaceData = new ArrayList<>();
+
```
michaelbausor: Nit: remove newline

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
```diff
@@ -292,4 +464,67 @@ private void addApiImports(GapicInterfaceContext context) {
 
     return apiMethods;
   }
+
+  private String generateConfigOutputPath(GapicInterfaceContext context, String name) {
+    GapicInterfaceConfig interfaceConfig = context.getInterfaceConfig();
+    String outputPath =
+        pathMapper.getOutputPath(
+            context.getInterfaceModel().getFullName(), context.getProductConfig());
+    return outputPath.substring(0, outputPath.length() - 6)
```
michaelbausor: As discussed, lets restructure this.

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
```diff
@@ -292,4 +464,67 @@ private void addApiImports(GapicInterfaceContext context) {
...
+    if (interfaces.containsKey(interfaceName)) {
+      List<HttpRule> httpRules = interfaces.get(interfaceName);
+
+      for (int i = 0; i < httpRules.size(); i++) {
+        // If a rule is already set, override it. This exists to allow rules
+        // defined in the service configuration to take precedence as they are
+        // added last.
+        if (httpRules.get(i).getSelector().equals(httpRule.getSelector())) {
+          httpRules.set(i, httpRule);
+
```
michaelbausor: Nit: remove newline

src/test/java/com/google/api/codegen/testdata/php/php_library.baseline
```diff
@@ -157,11 +170,19 @@ use Google\Tagger\CustomNamespace\V1\LabelerGrpcClient;
...
+     * The name of the service.
+     */
+    const SERVICE_NAME = 'google.example.library.v1.LibraryService';
+
     /**
      * The default address of the service.
      */
     const SERVICE_ADDRESS = 'library-example.googleapis.com';
 
+
```
michaelbausor: Nit: remove newline (if this is difficult because of snippet lang, then don't worry)

src/test/java/com/google/api/codegen/testdata/php/php_library.baseline
```diff
@@ -219,6 +248,7 @@ class LibraryServiceGapicClient
         return self::$returnNameTemplate;
     }
 
+
```
michaelbausor: Nit: remove newline (if this is difficult because of snippet lang, then don't worry)

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
dwsupplee: On it!

How does `public String getConfigPath(InterfaceConfig interfaceConfig, String name)` in SurfaceNamer sound?

I tossed around the idea of refactoring `getClientConfigPath` into the more generic `getConfigPath`, but it seems rather disruptive, what about just having these two methods live next to each other?

michaelbausor: SGTM

src/main/java/com/google/api/codegen/transformer/DynamicLangApiMethodTransformer.java
```diff
@@ -103,9 +103,12 @@ public OptionalArrayMethodView generateMethod(
     apiMethod.requestVariableName(namer.getRequestVariableName(method));
     apiMethod.requestTypeName(
         namer.getRequestTypeName(context.getTypeTable(), context.getMethod().getInputType()));
+    apiMethod.responseTypeName(
+        namer.getRequestTypeName(context.getTypeTable(), context.getMethod().getOutputType()));
```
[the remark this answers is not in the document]
dwsupplee: What about just `getTypeName`?

src/test/java/com/google/api/codegen/testdata/php/php_library.baseline
dwsupplee: SG. I was able to get this guy, but it looks like there are lots of other extraneous newlines in this file as well. We should earmark coming back and cleaning it up a bit.

src/main/java/com/google/api/codegen/transformer/DynamicLangApiMethodTransformer.java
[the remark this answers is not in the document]
michaelbausor: Hmm, my instinct is that we want something more descriptive, but I don't know what. `getProtoElementTypeName`? `getTypeRefTypeName`? I don't actually think that I prefer either of those to just `getTypeName`...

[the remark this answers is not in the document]
dwsupplee: I was able to find an alternative route which allows us to leave the name as is :),

```java
apiMethod.responseTypeName(
        namer.getMessageTypeName(
            context.getTypeTable(), context.getMethod().getOutputType().getMessageType()));
```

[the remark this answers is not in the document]
dwsupplee: Looks like I jumped the gun. That doesn't add the classes to our imports, so it won't work.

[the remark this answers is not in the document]
dwsupplee: What about renaming `getRequestTypeName` to an overloaded `getMessageTypeName` which accepts a `TypeRef`?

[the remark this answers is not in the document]
michaelbausor: I think the reason not to do an overload is because the method supports types other than message types, such as primitives like string.

I think maybe `getAndSaveTypeName` (the `AndSave` because that is the part that the `getMessageTypeName` function you tried out isn't doing).

But I don't feel strongly - go with your best judgement :-)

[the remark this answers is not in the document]
dwsupplee: That sounds nice! Updated

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
dwsupplee: We discussed this item offline and decided the current strategy is acceptable.

PR #1790, opened 2018-01-09
go: properly make smoke test for List methods

pongad:


src/main/java/com/google/api/codegen/transformer/go/GoGapicSurfaceTestTransformer.java
```diff
@@ -213,8 +213,14 @@ private SmokeTestClassView createSmokeTestClassView(InterfaceContext context) {
...
   private StaticLangApiMethodView createSmokeTestCaseApiMethodView(MethodContext methodContext) {
-    StaticLangApiMethodView initialApiMethodView =
-        new StaticLangApiMethodTransformer().generateRequestObjectMethod(methodContext);
+    StaticLangApiMethodView initialApiMethodView;
+    if (methodContext.getMethodConfig().isPageStreaming()) {
+      initialApiMethodView =
+          new StaticLangApiMethodTransformer().generatePagedRequestObjectMethod(methodContext);
+    } else {
+      initialApiMethodView =
+          new StaticLangApiMethodTransformer().generateRequestObjectMethod(methodContext);
```
garrettjonesgoogle: Could you store the transformer as a member variable? This makes the code easier to read because it's more predictable where the transformers are defined and instantiated.

pongad: Ah right, done

PR #1792, opened 2018-01-09
Python: add license to namespace init files.

landrito:


src/test/java/com/google/api/codegen/testdata/py/py_library.baseline
```diff
@@ -736,13 +736,41 @@ Api Reference
     gapic/v1/api
     gapic/v1/types
 ============== file: google/__init__.py ==============
+# Copyright 2017 Google LLC
```
lukesneeringer: 2018, but not blocking the PR on that.

PR #1803, opened 2018-01-12
Java: Moving off deprecated proto-inserted methods

garrettjonesgoogle:
This depends on https://github.com/googleapis/protoc-java-resource-names-plugin/pull/23 .

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
@@ -621,7 +621,7 @@ public class InitCodeTransformer {
       String transformParamFunctionName = "";
       if (context.getFeatureConfig().useResourceNameConverters(fieldConfig)) {
         if (fieldConfig.getField().isRepeated()) {
-          // TODO support repeated one-ofs
+          // TODO (https://github.com/googleapis/toolkit/issues/1806) support repeated one-ofs
           transformParamFunctionName =
               namer.getResourceTypeFormatListMethodName(context.getTypeTable(), fieldConfig);
         } else {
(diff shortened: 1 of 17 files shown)


src/test/java/com/google/api/codegen/testdata/discogapic/java/java_simplecompute.v1.json.baseline
```diff
@@ -3100,7 +3100,7 @@ public final class DeleteAddressHttpRequest implements ApiMessage {
       return newBuilder;
     }
 
-    public final Builder setAddressWithAddressName(AddressName value) {
+    public final Builder setAddress(AddressName value) {
```
michaelbausor: I'm confused about this renaming. Where is it coming from?

garrettjonesgoogle: This is because of the FeatureConfig change for Java. As a follow-up change, we need to change the input and output types of these methods to String, but I didn't want to tackle that within this PR.

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -605,6 +617,19 @@ private InitValueView getInitValue(MethodContext context, InitCodeNode item) {
       fieldSetting.required(
           fieldConfig != null
               && requiredFieldSimpleNames.contains(fieldConfig.getField().getSimpleName()));
+      String formatMethodName = "";
+      String transformParamFunctionName = "";
+      if (context.getFeatureConfig().useResourceNameConverters(fieldConfig)) {
+        if (fieldConfig.getField().isRepeated()) {
+          // TODO support repeated one-ofs
```
michaelbausor: Should we create an issue to track this?

src/test/java/com/google/api/codegen/testdata/discogapic/java/java_simplecompute.v1.json.baseline
garrettjonesgoogle: https://github.com/googleapis/toolkit/issues/1804

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
garrettjonesgoogle: https://github.com/googleapis/toolkit/issues/1806

PR #1812, opened 2018-01-18
Ruby package metadata and linter updates

geigerj:
- Upgrade rubocop to 0.50.0 (with corresponding config updates)
- Do not check smoke test environment variables if there is no
  smoke test


geigerj:
Sanity checked on BigTable and NL.

Issue #1810, opened 2018-01-17
Uninitialized required fields should have a comment 


geigerj:
### Context
https://github.com/GoogleCloudPlatform/google-cloud-ruby/issues/1866#issuecomment-358061944

### What

If a field is required, it should be populated. Currently, an empty object may be created if the sample initialization values are not configured. In this case, we should generate an explanatory comment to populate the object.

For example, currently, one unconfigured Ruby NL method sample is

```
require "google/cloud/language/v1"

language_service_client = Google::Cloud::Language::V1.new
document = {}
response = language_service_client.analyze_syntax(document)
```

This might incorrectly imply that it is valid to send an empty document. Instead, display

```
require "google/cloud/language/v1"

language_service_client = Google::Cloud::Language::V1.new
// TODO: initialize document
document = {}
response = language_service_client.analyze_syntax(document)
```

geigerj:
Temporarily assigning to myself to look into for Ruby, but ultimately, this could be done cross-language.

cc: @vchudnov-g

vchudnov-g:
I agree we should do this cross-language. This has some implications for samplegen.

geigerj:
Done for Ruby in #1813. I'll look into other languages as time permits (or open this up to language leads if not).

jbolinger:
Python added in #1840 - not sure if I'll have time to do the others but it was a good introduction.

jbolinger:
I think this issue is no longer relevant given the current sample gen spec. Feel free to reopen if you feel otherwise.

PR #1813, opened 2018-01-19
Allow user initialization TODOs in samples and apply this to Ruby

geigerj:
In some autogenerated samples, additional user input is required to
obtail working code if the sample has not been fully configured. In
the case where a required field is not configured, add a comment to
the sample code.

Updates #1810


src/test/java/com/google/api/codegen/testdata/ruby/ruby_doc_library.baseline
```diff
@@ -2823,6 +2823,7 @@ module Library
       #   require "library/v1"
       #
       #   library_service_client = Library::V1.new
+      #   # TODO: Initialize this field.
```
vchudnov-g: Is it possible to have the TODO in line with the empty initialization?

Alternatively, either include the field name in the text of the TODO, or end the TODO with a colon?

geigerj: Whether it's inline is controlled by the templates on a language-by-language basis. For Ruby, I think I'd rather not because we don't have a code formatter, and I'd worry about the frequency of line length overruns. (I also don't feel strongly about that, since the Ruby generated code already fails the linter in several other places, so if you think it's better inline, I can change it to be inline.)

I did make both of the other changes (using the field name in the text, and ending with a colon).

src/test/java/com/google/api/codegen/testdata/ruby/ruby_doc_library.baseline
```diff
@@ -2823,7 +2823,7 @@ module Library
       #   require "library/v1"
       #
       #   library_service_client = Library::V1.new
-      #   # TODO: Initialize this field.
+      #   # TODO: Initialize +shelf+:
```
vchudnov-g: nit: Do you think it would make the comment more obvious to have a blank line preceding? This might conflict with some style rules, possibly. I guess it's not a biggie if people are using syntax highlighting: they'll see the comment.

vchudnov-g:
lgtm, but do amend the PR description that clarify this only modifies the output for Ruby (or go ahead and do the other languages, which would be great!)

PR #1814, opened 2018-01-19
Modify GAPIC config schema for GAPIC samplegen

vchudnov-g:
- Add config_sample_values and samples to config.proto:MethodConfigProto and
  library_gapic.yaml
- Add CodeSampleValues and SampleConfiguration to config.proto
- This change modifies the spec previous advertised on the design doc,
  primarily due to protobuf restrictions (default values; no maps with "repeated"
  values) but also to accommodate specifying Customary Samples more easily


src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -292,6 +300,84 @@ message MethodConfigProto {
...
+  // * `DOTTED.RESOURCE.NAME=VALUE`, to specify the entire value of
+  //   `NAME`.  If `VALUE` is not specified, it is treated as the
+  //    default zero value for `NAME`'s type.
+  // *  `DOTTED.RESOURCE.NAME#FIELD=VALUE`, if `NAME` is required to be
+  //    a specially formatted string using `FIELD` as one of possibly
+  //    several placeholders.
+  repeated string parameters = 4;
+}
+
+// `SampleConfigurations` defines combinations of "molds" (sample code
```
garrettjonesgoogle: Shouldn't `SampleConfigurations` be singular to match the message name?

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -292,6 +300,84 @@ message MethodConfigProto {
...
+
+    // Whether the samples specified by this configuration should also
+    // be used as configurable "Explorer Samples" suitable for
+    // inclusion in a UI element.
+    bool explorer_sample = 2;
+
+    // Whether the samples specified by this configuration should also
+    // be used as "Customary Samples" that are included within code
+    // code comments in the surface of each generated client library.
+    bool customary_sample = 3;
```
garrettjonesgoogle: "customary" doesn't really have any meaning out of context. Maybe "surface_sample" would be better, or alternatively "in_code_sample".

src/test/java/com/google/api/codegen/testsrc/library_gapic.yaml
```diff
@@ -227,6 +227,28 @@ interfaces:
     resource_name_treatment: STATIC_TYPES
     sample_code_init_fields:
     - series_uuid.series_string=foobar
+    code_sample_values:
+      - id: review
```
garrettjonesgoogle: The convention in our yamls is for the `-` lines to start at the same indentation as the key.

src/test/java/com/google/api/codegen/testsrc/library_gapic.yaml
```diff
@@ -227,6 +227,28 @@ interfaces:
...
+        title: "Second edition"
+        description: "Calling PublishSeries for a specified edition"
+        parameters:
+          - edition=2
+    samples:
+      - molds: ["edition"]                       # TODO(vchudnov-g) use a real mold once implemented
+        values:
+          - ids: ["second_edition","review"]
+            explorer_sample: true
+            position: 1
```
garrettjonesgoogle: It feels like this configuration could make it difficult to reason about the complete set of explorer samples that will be presented and which order they will be in. The reader has to check for `explorer_sample` and then check for the position fields, which might or might not be there.

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -292,6 +300,84 @@ message MethodConfigProto {
...
+  // A short, user-visible name for this set of values. This name may
+  // be used in a UI menu, for example.
+  string title = 2;
+
+  // A longer description for this set of values. This name may be
+  // used in a UI hoverbox, for example.
+  string description = 3;
+
+  // A set of parameter values to be used together in samples. Each
+  // member of `parameters` can be specified in one of two formats:
```
geigerj: Note: our current initialization fields have the following grammar (copied from an internal doc), which has special syntax for maps and arrays, and which doesn't support the special syntax for `FIELD` that you propose:

```
<field-path> ::= <field> | <field>"." <field-path>
<field> ::= <field-name> | <field-name> <array-index> | <field-name> <map-key>
<array-index> ::= "[" <unsigned integer> "]"
<map-key> ::= "{" <map-key-name> "}"
<map-key-name> ::= [a-zA-Z0-9_]*
```

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -292,6 +300,84 @@ message MethodConfigProto {
...
+  repeated string parameters = 4;
+}
+
+// `SampleConfigurations` defines combinations of "molds" (sample code
+// structures as output by the client library generator) and
+// `CodeSampleValues` that should be combined to generate samples.
+message SampleConfiguration {
+  // `Value` references one or more `CodeSampleValues` and provides a
+  // hint as to where it should be presented.
+  message Value{
```
geigerj: nit: add a space before `{`

src/test/java/com/google/api/codegen/testsrc/library_gapic.yaml
geigerj: YAML lists are ordered, so one alternative is to derive the position from the list position. We could also adopt the convention of listing explorer samples before non-explorer samples for readability. (It's not clear to me what the expected cardinality of samples and explorer samples is, which will have a large impact on how much of an issue this is in practice.)

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -292,6 +300,84 @@ message MethodConfigProto {
...
+    bool customary_sample = 3;
+
+    // A hint as to the user-visible position of this
+    // `SampleConfiguration` relative to other
+    // `SampleConfiguration`s. This is only used when generating
+    // "Explorer Samples", which only happens if `explorer_sample` is
+    // set.
+    uint32 position = 4;
+
+    // Whether or not to delete (rather than create) any previously
```
geigerj: Does this belong in `toolkit` or `artman`? So far, `toolkit` has had no knowledge of what's already "out there" -- that's entirely the realm of `artman`.

src/test/java/com/google/api/codegen/testsrc/library_gapic.yaml
vchudnov-g: I can see how it could be, but I'm not sure it will be terribly complicated. If the user simply specified `explorer_sample` and no `position`, they samples will show in order. If they have any reason to put the values in a different order in the file than the order they want in the UI, then they need to specify `position`.

1. One alternative is to remove `explorer_sample` and only those entries with non-zero `position` are considered Explorer Samples. The drawbacks are that (a) it overloads `position` (but granted, `position` only makes sense if `explorer_sample` is specified, so they're also kinda redundant), (b) makes it less obvious what it's used for, and (c) requires the user to update all the subsequent positions in case they want to reorder (as opposed to leaving `position` unspecified and using the implicit order  in the file). 
I suppose we could address these objections by calling it `ui_position`, letting a negative value denote "use it as an ES, but toolkit can figure out the relative placement", and positive values to denote the desired relative placement.

2. A different alternative it to instead remove `position` altogether, and always use the relative order of the items in the file with `explorer_sample` set as the order in which they appear in the UI. I don't like this as much because it forces the file order to be the same as the UI order, when there's no reason they should be.

What do you think? I slightly prefer the current version (call it 0) for clarity, but I'm fine doing alternative 1 instead. I'm not convinced about alternative 2. Any other ideas?

src/main/proto/com/google/api/codegen/config.proto
vchudnov-g: Fair enough. Going for "in_code_sample" for clarity and specificity.

src/test/java/com/google/api/codegen/testsrc/library_gapic.yaml
vchudnov-g: Fixed

src/main/proto/com/google/api/codegen/config.proto
vchudnov-g: Good catch!

src/main/proto/com/google/api/codegen/config.proto
vchudnov-g: Done

src/main/proto/com/google/api/codegen/config.proto
vchudnov-g: I wasn't proposing it. I was trying to capture the syntax for formatted field names as it already exists in the [code](https://github.com/googleapis/toolkit/blob/master/src/test/java/com/google/api/codegen/testsrc/library_gapic.yaml#L85), but I used the wrong field placeholder! Ooops.

Fixed.

src/main/proto/com/google/api/codegen/config.proto
vchudnov-g: Let me think about this for a bit....

vchudnov-g: My first inclination is to leave this in the GAPIC config for the following reasons:

1. the artman config does not deal with method-level details, correct?

2. the actual location of the method samples directory is still handled by the artman config, which can move directories around with its `directory_mappings`. It is only the content of that directory that is configured here (and yes, with the new wrinkle of specifying behavior that only artman can implement if it encounters a pre-existing artifact that needs to be deleted)

3. if the GAPIC config is responsible for causing specific samples to be created, it should also be responsible for capturing the information that specific samples need to be deleted. Splitting it up seems confusing and counter-intuitive.

One alternative is to always have the generated samples directory overwrite the repository's sample directory; in that case, if you want to "delete" a sample, you simply make sure it is not configured and it will never be generated and hence will not show up in the new version of the samples directory. The problem with this is that we want to allow manual samples to be added to the samples directory, and we don't want to clobber them every time we regenerate by overwriting the samples directory entirely. If we want to go with this alternative, we have to figure out a different place for manual samples, and I fear that splitting samples for a given API+language among different directories will make the generated libraries less usable.

Thoughts?

geigerj: > the artman config does not deal with method-level details, correct?

Correct. I don't think this is doable in artman with configuration as-is.

> (and yes, with the new wrinkle of specifying behavior that only artman can implement if it encounters a pre-existing artifact that needs to be deleted)

Could you clarify this point: are you suggesting that toolkit would communicate to artman that artman needed to perform the deletion?

> if the GAPIC config is responsible for causing specific samples to be created, it should also be responsible for capturing the information that specific samples need to be deleted. Splitting it up seems confusing and counter-intuitive.

To be honest, I'm not sure I'm convinced that "delete existing sample" is something that should be captured in declarative configuration at all. One possible approach is that toolkit could generate a samples directory that could then be merged intelligently with a manual samples directory by artman. Assuming autogenerated samples are "tagged" as such, `artman` should know to delete and replace all existing autogenerated samples while leaving manual samples alone.

Another possible approach that avoids the need for configuration is to develop a lightweight CLI (integrated into `artman`) that takes an artman config as input and can handle deleting samples by ID.

lukesneeringer: I agree with @geigerj here:

  * If we need a delete feature (we could always add this complexity later), this is not the place for it. We would be introducing a heretofore unknown assumption into toolkit (knowledge of surroundings).
  * It would be possible to `os.walk` the samples in a sample directory and delete samples that have been removed, if necessary, without touching manually written samples. This could also allow an auto-generated sample to be _modified_ and not replaced by removing the autogen comment line, which has value.

vchudnov-g: A proposal I like better than the current design, and then replying to the specific comments you made (on the design I now favor less :wink: )


## Proposal
**tl;dr:**  (1) don't worry about preserving manual samples for the first iteration (focus on autogen); (2) reserve a filename prefix for manual samples, and toolkit will never autogen samples with that prefix; (3) to allow not-clobbering manual samples, alter artman to not overwrite files with that prefix

**Details:**

The motivation for the delete feature is to allow changing the set of autogen samples without clobbering manual samples (as a full-on directory overwrite would do) and without having a second samples directory for manual samples (which is not great usability).

So then let's use a different method to disambiguate manual from autogenerated samples: a requirement that filenames with a given prefix are always preserved by artman in the publishing process. For example, we could say that all manual sample files should begin with `ex.` and none of the autogenerated files will. Artman will take care to preserve any such files that may exist when cloning the repo.

(This could be extended to allow manual files anywhere in our repos, if we so choose, by having one or more non-cloberable-while-publishing prefixes.)

What this would mean in the short term is that:
* we do not need to implement the manual-sample-preserving functionality right away (let's call it "M1"), since this is less important than getting autogen samples out the door
  * for M1, artman can simply overwrite the whole samples directory if it chooses and not worry about preserving manual samples
* we do need to reserve the filename prefix that we will use to identify and preserve manual samples, simply so toolkit never autogens samples with that prefix, not even in M1
* when we do implement the manual-sample-preserving functionality ("M2"), artman will take care to not delete files with the given prefix

_Variant_:  Instead of a filename prefix, we could use the delimiter tags within the samples, as Luke suggests. If we do so, I maintain the tag names themselves would have to follow a naming convention for either manual or autogen samples, because we want to still allow including the tags in manual samples too (since the tags allows the sample files to include content that can be easily excluded from devsite). So I think having the prefix at the filename level makes it substantially easier for humans to see which samples are manual, and just a hair easier for artman to distinguish them (globbing filenames rather than inspecting each file, essentially).

## Replies to comments

(many of these are moot if we go with the Proposal above)

> Could you clarify this point: are you suggesting that toolkit would communicate to artman that artman needed to perform the deletion?

That had been my thinking: just as we configure samples to appear and artman takes care of copying them to the right location, we could anti-configure samples to _not_ appear, and artman would take care of deleting them if it happens to encounter them. 

> To be honest, I'm not sure I'm convinced that "delete existing sample" is something that should be captured in declarative configuration at all. 

I mostly disagree (but the Proposal avoids this issue altogether). If we use the config to declare things that should be there, we could also use it to declare things that should not be there.

> One possible approach is that toolkit could generate a samples directory that could then be merged intelligently with a manual samples directory by artman.

I'm opposed to this. Having two separate directories in the final output optimizes for our development, but not for the API client experience. Users should not have to track whether samples were done manually or automatically: they should just have a single samples location where the samples just work. And if we merge, we still have the manual samples directory just dangling somewhere with duplicate content (as far as the API client is concerned)---and we also have another location to track and manipulate.

> Assuming autogenerated samples are "tagged" as such, artman should know to delete and replace all existing autogenerated samples while leaving manual samples alone.

See Proposal.

> Another possible approach that avoids the need for configuration is to develop a lightweight CLI (integrated into artman) that takes an artman config as input and can handle deleting samples by ID.

I'm opposed to this too because, again, this would introduce method-level configuration into artman. And also because, as I mention above, where we configure the allowed samples can certainly be where we configure the disallowed (ie delete-if-you-see-it) samples.

> If we need a delete feature (we could always add this complexity later), 

I'm OK with deferring this feature until we need it (as in the Proposal)

> this is not the place for it.

I'm not yet convinced it's not.

> We would be introducing a heretofore unknown assumption into toolkit (knowledge of surroundings).

How so? The functionality of the (pre-Proposal) design is "when you go to publish: after you clone the repo, if you see any of the samples that match this delete config, remove them; then add the newly generated samples; then commit.". The delete configuration is idempotent as far as toolkit is concerned and requires no prior "knowledge" as to whether the disallowed samples exist or ever existed.  It simply says, "if you encounter it, delete it."

But again, I prefer the Proposal above :point_up:. Let me know your thoughts.

vchudnov-g:
ping!

* On delete: I removed the delete functionality for MVP, but I think the Proposal I outlined in the comment above makes sense to implement in a later iteration. @geigerj @lukesneeringer Feedback?

* On Explorer Samples ordering: I slightly prefer the current PR code (option 0) to option 1 in the comment above, but could certainly implement option 1 if it addresses your concerns, @garrettjonesgoogle . But I don't want change things if it doesn't. Feedback?

geigerj:
@vchudnov-g This PR LGTM without the `delete` field. I think it makes sense to move the sample-deleting discussion to a different forum and proceed with the autogen functionality. I'm broadly in agreement with your proposal in that I think it makes sense to distinguish autogen samples from manual samples through metadata (whether tags or filename or something else).

I think the one point where I still disagree with you is that I don't think the deletion functionality should be configurable (either in `artman` or `toolkit`), since it's a one-time operation whose configuration immediately becomes stale once the delete has been executed. I don't see that as being in conflict with your proposal, though.

garrettjonesgoogle:
I'd still like to take another look before approving from my perspective

vchudnov-g:
@geigerj Correct, the Proposal would take deletion out of the configuration and it would be something that artifact maintainers can do however they wish. Regeneration would  always re-populate automatic samples and not touch any files that have the correct metadata identifying them as manual samples.

@garrettjonesgoogle I'll wait for your feedback.

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -292,6 +301,96 @@ message MethodConfigProto {
 
 }
 
+// `SampleValueSet` defines a set of parameter values used in
+// generating samples.
+message SampleValueSet {
+  // An identifier, unique to this method, for this set of values.
```
garrettjonesgoogle: maybe: "unique to a particular API method"

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -292,6 +301,96 @@ message MethodConfigProto {
...
+
+  // A set of parameter values to be used together in samples. Each
+  // member of `parameters` can be specified in one of two formats:
+  //
+  // * `DOTTED.RESOURCE.NAME=VALUE`, to specify the entire value of
+  //   `NAME`.  If `VALUE` is not specified, it is treated as the
+  //    default zero value for `NAME`'s type.
+  // *  `DOTTED.RESOURCE.NAME%FIELD=VALUE`, if `NAME` is required to be
+  //    a specially formatted string using `FIELD` as one of possibly
+  //    several placeholders.
```
garrettjonesgoogle: Your indentation is a little funky above - it transitions from 1 space in from the stars to 2 spaces in from the stars.

src/main/proto/com/google/api/codegen/config.proto
vchudnov-g: done

src/main/proto/com/google/api/codegen/config.proto
vchudnov-g: Fixed. Thanks for the catch!

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -316,62 +318,77 @@ message CodeSampleValues {
...
+    // One or more expressions matching the
+    // `MethodConfigProto.sample_value_sets.id`s defined for this
+    // method. Any expressions on this list that do not match a
+    // `SampleValueSet.id` cause a fatal error during code
+    // generation. If not specified or if empty, no samples of this
+    // type will be generated.
+    //
+    // An expression matches an ID
+    // (a) if the expression equals "*" (which matches all IDs), or
+    // (b) if the leading characters of the ID are identical to the
```
geigerj: When do you anticipate this "prefix matching" feature being used?

Did you consider making `*` a wildcard in the expression to make these semantics more explicit (i.e., expression `foo*` matches ID `foobar`, but expression `foo` does not match ID `foobar`)?

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -316,62 +318,77 @@ message CodeSampleValues {
...
-  repeated string molds = 1;
+  // `SampleTypeConfiguration` defines the samples that will show up
+  // as a particular sample type. If order matters for this sample
+  // type, the order of the samples is taken to be primarily the order
+  // in which the `calling_forms` appear, and secondarily, for each
+  // such form, the order in which the `value_sets` appear.
+  message SampleTypeConfiguration {
+    // One or more expressions matching the method calling forms
+    // (defined by the client generator) applicable to this
+    // method. Any expressions that do not match at least one calling
```
geigerj: > Any expressions that do not match at least one calling form cause a fatal error during code generation

Could you clarify this?

In the test configuration, I see `request_based` and `callable` being used as expressions here. I assume this refers to the distinction in Java between request-object methods and flattened method. (Is that right?) If so, `request_based` doesn't apply to most of the dynamic languages, which have no request-object methods. I believe the generators for those languages should simply ignore configuration for that calling form, but not raise an error.

vchudnov-g: That's a good point. i guess I was thinking forms "that do not match at least one calling form **in at least one language**" will cause an error, but otherwise they will be ignored by the languages that don't employ that calling form. My motivation was to catch configuration errors as early and prominently as possible; a situation like this is an obvious error.

However, now that you bring it up, I wonder whether this really confuses the semantics. Maybe it's better to say all unrecognized expressions are ignored, regardless of whether they exist for any language. I suppose we could issue an INFO message for each such unrecognized instance (though that would potentially be a lot of instances and I don't know whether people do/would look at those messages). Thoughts?

[Separately, I wonder whether eventually we'll want to expand `SampleTypeConfiguration` to include a `language` field. It's premature to do it now, though.]

src/main/proto/com/google/api/codegen/config.proto
vchudnov-g: I did, but I was trying to hew to the mental model in the document, where we use `#` separated fields to specify the method style and the calling form. However, what we ultimately care about in the config is the method form, so maybe it's easier to grok simply having wildcard matching here. The organization of the styles and forms can be completely contained in our naming convention for them, which we control.....

OK, changing it.

src/main/proto/com/google/api/codegen/config.proto
geigerj: > Maybe it's better to say all unrecognized expressions are ignored, regardless of whether they exist for any language.

I prefer this because it allows a (future, modular) world where a generator for a particular language doesn't need to know about the behavior other languages' generators. An INFO message for unrecognized expressions might still be helpful for debugging, as long as the tooling doesn't halt.

PR #1816, opened 2018-01-23
fix snippet error for node.js smoke test for long running operation

alexander-fenster:
Need to pass correct callback. There is no variable named `additionalCallback` in this scope.


ethanbao:
I will leave the review to @landrito as I have near zero knowledge about how NodeJS snip works.

alexander-fenster:
Looks like adding coverage for long_running: might be a bigger task than just changing this one snippet. Created #1825.

PR #1817, opened 2018-01-23
C# Fix smoketest regression

### LINKED ISSUE #1807: Smoke test regression in C#
The generated smoke tests now have two issues:

- If it's a paginated call, it expects the native response type instead of `PagedEnumerable<...>`. This used to work.
- If it's a long-running operation call, it expects a response type of `Operation` rather than `Operation<...>`, and just prints that response rather than waiting for the operation to complete. This may never have worked.

chrisdunelm:
Fixes #1807

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSmokeTestTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSmokeTestTransformer.java
@@ -17,6 +17,7 @@ package com.google.api.codegen.transformer.csharp;
 import com.google.api.codegen.InterfaceView;
 import com.google.api.codegen.config.FlatteningConfig;
 import com.google.api.codegen.config.GapicProductConfig;
+import com.google.api.codegen.config.MethodConfig;
 import com.google.api.codegen.config.MethodModel;
 import com.google.api.codegen.config.SmokeTestConfig;
 import com.google.api.codegen.gapic.GapicCodePathMapper;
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSmokeTestTransformer.java
```diff
@@ -145,8 +149,13 @@ private SmokeTestClassView generateSmokeTestCsproj(GapicInterfaceContext context
   }
 
   private StaticLangApiMethodView createSmokeTestCaseApiMethodView(MethodContext methodContext) {
-    StaticLangApiMethodView initialApiMethodView =
-        new CSharpApiMethodTransformer().generateFlattenedMethod(methodContext);
+    MethodConfig methodConfig = methodContext.getMethodConfig();
+    StaticLangApiMethodView initialApiMethodView;
+    if (methodConfig.isPageStreaming()) {
```
jskeet: Conditional operator?

chrisdunelm:
Yes, this just fixes methods that return lists. A further PR will add support for operations.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSmokeTestTransformer.java
chrisdunelm: Further `if`s will be required for support of things like operation-based results.
So I'd rather leave it as it is.

PR #1827, opened 2018-01-26
Stubsettings

andreamlin:
Generate a StubSettings class that provides implementation of the more visible Settings class.


src/main/java/com/google/api/codegen/viewmodel/StaticLangStubSettingsView.java
```diff
@@ -0,0 +1,164 @@
+/* Copyright 2016 Google LLC
```
garrettjonesgoogle: 2018

PR #1828, opened 2018-01-26
google-cloud-python #4295 - guard calls to setuptools.setup()

chemelnucfin:
[google-cloud-python/4295](https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4295)


chemelnucfin:
See [google-cloud-python/4800](https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4800)

PR #1832, opened 2018-02-01
Add "under development" disclaimer to new samplegen config fields

vchudnov-g:
Users can populate these fields, but they will have no effect yet as
we are still developing the back-end code to act on these settings.

### DIFF
--- a/src/main/proto/com/google/api/codegen/config.proto
+++ b/src/main/proto/com/google/api/codegen/config.proto
@@ -274,11 +274,19 @@ message MethodConfigProto {
 
   // Specifies sets of parameter values to be used together in a
   // sample. If this is not set, no samples are output.
+  //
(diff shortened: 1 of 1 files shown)


src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -274,11 +274,19 @@ message MethodConfigProto {
 
   // Specifies sets of parameter values to be used together in a
   // sample. If this is not set, no samples are output.
+  //
+  // UNDER DEVELOPMENT: Sample generation using this field is still
```
garrettjonesgoogle: Better wording: "Sample generation usage of this field..."

vchudnov-g: That seems a bit _too_ compact for clarity. I tweaked it slightly differently. LMKWYT.

PR #1833, opened 2018-02-01
Remove paged response wrappers

andreamlin:
Move the three static inner classes from PagedResponseWrappers into the client class files, and remove PagedResponseWrappers wrapping class.


garrettjonesgoogle:
This will still cause a name clash. I think we need to move the paged response classes to the individual client classes that they apply to.

andreamlin:
PTAL

See https://github.com/googleapis/gax-java/pull/471 for the resulting generated LRO API.

PR #1834, opened 2018-02-02
use google-common-protos for longrunning proto and grpc dependencies

neozwu:


src/main/resources/com/google/api/codegen/java/build_gapic.gradle.snip
```diff
@@ -18,11 +18,23 @@
...
   dependencies {
-    compile project(":{@metadata.protoPackageName}")
+    @if metadata.protoPackageName == "proto-google-longrunning-v1"
+      compile project(":proto-google-common-protos")
+    @else
+      compile project(":{@metadata.protoPackageName}")
+    @end
     compile libraries.gax, libraries.gaxGrpc
-    testCompile project(":{@metadata.grpcPackageName}")
+    @if metadata.grpcPackageName == "grpc-google-longrunning-v1"
```
garrettjonesgoogle: We don't want to generate grpc-google-longrunning-v1 at all... 

Can we figure out a way to configure this all more generically, so that we don't have these package names in toolkit logic?

neozwu: I agree. This is really a workaround. It seems currently 'java_gapic' will generate proto, grpc and gapic by default. Will look into this.

garrettjonesgoogle: Maybe we could support this by adding a config flag to turn off proto/grpc generation on a particular API, and then have it set to off for longrunning

garrettjonesgoogle:
This was solved in a different way already.

PR #1836, opened 2018-02-05
Java gen: Switching from Oneof to inheritance

garrettjonesgoogle:


src/main/java/com/google/api/codegen/transformer/StaticLangResourceObjectTransformer.java
```diff
@@ -75,12 +75,13 @@ public RequestObjectParamView generateRequestObjectParam(
     String formatMethodName = null;
     if (context.getFeatureConfig().useResourceNameFormatOption(fieldConfig)) {
       if (fieldConfig.requiresParamTransformation()
-          && !fieldConfig.requiresParamTransformationFromAny()) {
+          && !fieldConfig.requiresParamTransformationFromAny()
+          && !featureConfig.useInheritanceForOneofs()) {
```
michaelbausor: Is it possible to include this in the `requiresParamTransformation` function somehow? I can't see how on a quick look, but wanted to check whether you had considered this. It would mean we wouldn't need the new feature config.

garrettjonesgoogle: The answer needs to be different for C# vs Java, so it has to be controlled in feature config (which is a language-specific knob).

garrettjonesgoogle:
I have reworked the codegen for the new design - PTAL again

src/main/java/com/google/api/codegen/util/StringUtil.java
```diff
@@ -0,0 +1,26 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen.util;
+
+public class StringUtil {
+
+  /** TODO Do this operation in a more general way in the Name class */
```
michaelbausor: Why not place this in the Name class now?

garrettjonesgoogle: It would be 10x the work.

PR #1837, opened 2018-02-06
java: remove javadoc template

pongad:
These templates were used to generate javadoc from api-client-staging.
We don't need that anymore, since we can use `./gradlew pushDocs`
instead.


pongad:
@garrettjonesgoogle The way this is currently set up is not quite right. It won't link docs for grpc and proto packages properly. My thinking is this: keeping the template is just generating wrong code for the users, and we might as well remove it.

A non-cloud repo might want to create "aggregate" javadoc task like we did for api-client-staging. We could create a template for *that* if we want. It's only one-time work per repo though, so I'm honestly not sure how much work it will save.

pongad:
@garrettjonesgoogle I think you actually need to "approve" before GH will let me merge :/

PR #1840, opened 2018-02-08
Add TODO for field initialization for python

jbolinger:
Add TODO comments for python for #1810.

Managing the newlines in the templates is ugly, but this may be good enough for now.


src/main/resources/com/google/api/codegen/py/method_sample.snip
```diff
@@ -77,7 +77,14 @@
 @end
 
 # Generate a simple argument
-@private initLineSimple(line)
+@private initLineSimple(line, idx)
+  @if line.doc
+    @if util.toInt(idx) > util.toInt(0)
```
garrettjonesgoogle: We need to keep logic out of the templates - try to move this logic into transformers.

jbolinger: Sure, I tried. @geigerj suggested this could also be done as a post-processing step in artman.

src/main/resources/com/google/api/codegen/py/method_sample.snip
```diff
@@ -78,6 +78,14 @@
 
 # Generate a simple argument
 @private initLineSimple(line)
+  @if line.doc
+    @if line.isFirst
```
geigerj: `@if not(line.isFirst)`

src/main/java/com/google/api/codegen/viewmodel/SimpleInitCodeLineView.java
```diff
@@ -49,6 +52,8 @@ public static Builder newBuilder() {
 
     public abstract Builder doc(List<String> val);
 
+    public abstract Builder isFirst(boolean val);
```
geigerj: It would be better if this variable were more descriptive of its intention. Maybe, `leadingNewline`, which has the added benefit of being usable in other cases where a leading newline might be needed?

jbolinger:
@geigerj thanks for the not(x), I had guessed !x and not x... seems so obvious now

src/main/java/com/google/api/codegen/viewmodel/SimpleInitCodeLineView.java
```diff
@@ -31,8 +31,8 @@
 
   public abstract List<String> doc();
 
-  /** @return if this is the first item in a list of lines */
-  public abstract boolean isFirst();
+  /** @return if a newline has been added above this line */
```
geigerj: This is actually the opposite of what I had intended :P 

Optional: I'd prefer formulating this as "if a newline should be added above this line" because it makes the property about this object, and not about its context. I recognize that that might be a distinction without a difference given how it's used in the transformer.

geigerj:
@jbolinger "fun" fact: it's actually not a built-in, but a function defined here: https://github.com/googleapis/toolkit/blob/master/src/main/resources/com/google/api/codegen/common.snip

src/main/java/com/google/api/codegen/viewmodel/SimpleInitCodeLineView.java
jbolinger: Sounds the same to me. I liked isFirst because not all languages will want a newline here (ruby doesn't python does), but we can revisit the name when we do the others.

PR #1842, opened 2018-02-09
all: fix oneof clash in generated test

pongad:
We previously initialize all primitive fields in tests.
When we see multiple primitive fields in a oneof group,
we init all of them.

This commit dedupes by oneof group, so we only init only one.

Also, add "context" to Go list of "keywords". It's so commonly used
it might as well be reserved.


src/main/java/com/google/api/codegen/transformer/TestCaseTransformer.java
```diff
@@ -258,29 +262,37 @@ private MockGrpcResponseView createMockResponseView(
...
+  @VisibleForTesting
+  static <E extends FieldModel> List<FieldModel> responseInitFields(List<E> fields) {
+    HashSet<Oneof> oneofSet = new HashSet<>();
+    return fields
+        .stream()
+        .filter(f -> f.isPrimitive() && !f.isRepeated())
+        .filter(
+            f -> {
+              Oneof oneof = f.getOneof();
+              return oneof == null || oneofSet.add(oneof);
```
garrettjonesgoogle: Could you add a comment stating that this includes a field if it either isn't a oneof, or it's the first field on a oneof?

src/main/java/com/google/api/codegen/util/go/GoNameFormatter.java
```diff
@@ -106,72 +106,73 @@ public String classFileNameBase(Name name) {
...
-              "len",
-              "make",
-              "new",
-              "panic",
-              "print",
-              "println",
-              "real",
-              "recover")
+          // Keywords
+          .add("break")
```
garrettjonesgoogle: This seems like an unrelated change...

pongad: True. Moving to another change.

PR #1844, opened 2018-02-14
Quick Fix for Java static import issue 

vam-google:
It is done by completely disabling the feature. Without this fix the generator produces non-compilable code for all clients which use paging. Specifically it tries to **statically** import an internal class of **the same** class where the static import happens, which breaks compilation. 

For example, it would add the following line to the `DataTransferServiceClient.java` file (note the presence of `"DataTransferServiceClient"` in the import itself):

```java
import static com.google.cloud.bigquery.datatransfer.v1.DataTransferServiceClient.ListDataSourcesPagedResponse;
```

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/StandardImportSectionTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/StandardImportSectionTransformer.java
@@ -15,6 +15,7 @@
 package com.google.api.codegen.transformer;
 
 import com.google.api.codegen.metacode.InitCodeNode;
+import com.google.api.codegen.util.ImportType;
 import com.google.api.codegen.util.TypeAlias;
 import com.google.api.codegen.viewmodel.ImportFileView;
 import com.google.api.codegen.viewmodel.ImportSectionView;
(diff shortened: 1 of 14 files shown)


vam-google:
The static imports are actually used, so it has to be fixed properly now

src/main/java/com/google/api/codegen/transformer/StandardImportSectionTransformer.java
```diff
@@ -24,25 +25,48 @@
...
+        || !parentName.endsWith(className)) {
+      return false;
+    }
+
+    if (parentName.length() > className.length()) {
+      char packageSeparator = parentName.charAt(parentName.length() - className.length() - 1);
+      if (Character.isLetterOrDigit(packageSeparator)) {
+        return false;
+      }
+    }
```
garrettjonesgoogle: This is java-specific logic and thus should probably be in a java-specific class.

vam-google: Added comment as we discussed offline. Also included underscore as a valid classname character.

vam-google:
@garrettjonesgoogle PTAL

src/main/java/com/google/api/codegen/transformer/StandardImportSectionTransformer.java
```diff
@@ -53,17 +53,29 @@ public ImportSectionView generateImportSection(
...
+    // language-agnostic way. For example:
+    //     parentFullName = "package.path.ParentTheClass";
+    //     className = "TheClass" ;
+    // should return false, but:
+    //     parentFullName = "package.path.ParentTheClass";
+    //     className = "ParentTheClass";
+    // should return true.
+    //
+    // It is also assumed that everything in full class name which is not a letter/digit/underscore
+    // is a path seprator. This should be good enough for all languages that we support.
```
garrettjonesgoogle: spelling(seprator)

vam-google:
StaticImport is used only in java generation, this code checks the import type very early and if not StaticImport it does nothing (so should be pretty safe to say that Go is not affected) . Merging this now, since it is blocking pending release.

PR #1850, opened 2018-02-22
Add support for importing declared types in version index file

jbolinger:
This PR requires types defined in the service config in the version index file (Ruby only).

This was motivated by using custom exception types. Importing them in the index makes it more likely they will be included when an exception is raised.

### DIFF
--- a/src/main/java/com/google/api/codegen/InterfaceView.java
+++ b/src/main/java/com/google/api/codegen/InterfaceView.java
@@ -16,9 +16,7 @@ package com.google.api.codegen;
 
 import com.google.api.tools.framework.model.Interface;
 import com.google.api.tools.framework.model.Model;
-import com.google.api.tools.framework.model.TypeRef;
 import com.google.protobuf.Api;
-import com.google.protobuf.Type;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -36,20 +34,5 @@ public class InterfaceView implements InputElementView<Interface> {
       interfaces.add(model.getSymbolTable().lookupInterface(api.getName()));
     }
     return interfaces;
-  }
-
-  /**
-   * Helper to extract the types from the underlying model.
-   *
-   * @see com.google.api.Service#getTypesList()
-   * @param model model with service config
-   * @return types
-   */
-  public Iterable<TypeRef> getTypes(Model model) {
-    List<TypeRef> types = new ArrayList<>();
-    for (Type type : model.getServiceConfig().getTypesList()) {
-      types.add(model.getSymbolTable().lookupType(type.getName()));
-    }
-    return types;
   }
 }
--- a/src/main/java/com/google/api/codegen/InterfaceView.java
+++ b/src/main/java/com/google/api/codegen/InterfaceView.java
@@ -16,8 +16,11 @@ package com.google.api.codegen;
 
 import com.google.api.tools.framework.model.Interface;
 import com.google.api.tools.framework.model.Model;
+import com.google.api.tools.framework.model.TypeRef;
 import com.google.protobuf.Api;
+import com.google.protobuf.Type;
 import java.util.ArrayList;
+import java.util.List;
 
 /**
  * An interface-based view of model, consisting of a strategy for getting the interfaces of the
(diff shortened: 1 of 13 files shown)


src/main/java/com/google/api/codegen/config/DiscoApiModel.java
```diff
@@ -62,6 +62,12 @@ public boolean hasMultipleServices() {
     return interfaceModels;
   }
 
+  @Override
+  public Iterable<? extends TypeModel> getTypes() {
+    // TODO: is this supported?
```
geigerj: @landrito @andreamlin I believe this is supported in discovery (i.e., getting the types listed in the `types` section of the Service config); could one of you confirm?

src/main/java/com/google/api/codegen/InterfaceView.java
```diff
@@ -28,10 +31,25 @@
...
   }
+
+  /**
+   * Helper to extract the types from the underlying model.
+   *
+   * @see com.google.api.Service#getTypesList()
+   * @param model model with service config
+   * @return types
+   */
+  public Iterable<TypeRef> getTypes(Model model) {
```
geigerj: - Suggest renaming this to `getAdditionalTypes`. (This only returns the types listed in the `types` section of the service config, not all types in the Model's SymbolTable, right?)
- I think this code probably belongs in `ProtoApiModel`, rather than here. These `*View` classes aren't really designed to implement utility methods outside of `getElementIterable`.

src/main/java/com/google/api/codegen/viewmodel/metadata/VersionIndexView.java
```diff
@@ -51,6 +51,9 @@ public String resourceRoot() {
 
   public abstract List<VersionIndexRequireView> requireViews();
 
+  @Nullable
+  public abstract List<String> requireTypes();
```
geigerj: Should this be merged with the existing `requireViews` attribute?

src/main/java/com/google/api/codegen/config/DiscoApiModel.java
geigerj: (This doesn't need to be addressed in this PR, but I'd prefer a filed issue to a naked `TODO`.)

landrito: The types listed in the types section of the service config will be mapped to a schema in the schemas list inside the discovery document. I am unsure how that maps to disco-gen though.

src/main/java/com/google/api/codegen/InterfaceView.java
jbolinger: Sure, seems reasonable.

src/main/java/com/google/api/codegen/viewmodel/metadata/VersionIndexView.java
jbolinger: I have no idea. There wasn't much to work with from the comments so I thought it best to keep it standalone.

geigerj: I see. In that case, I would support merging them. (I would expect a `VersionIndexRequireView` to be a ViewModel for representing a Ruby `require` statement in the Version Index, so it would be surprising to need two separate lists of these.)

src/main/java/com/google/api/codegen/config/ProtoApiModel.java
```diff
@@ -114,14 +115,29 @@ public String getDocumentationSummary() {
...
   }
 
+  /**
+   * Helper to extract the types from the underlying model.
+   *
+   * @see com.google.api.Service#getTypesList()
+   * @param model model with service config
+   * @return types
+   */
+  protected Iterable<TypeRef> getTypes(Model model) {
```
geigerj: Why not `private`? (That is, will this be needed in subclasses of `ProtoApiModel`?)

src/main/java/com/google/api/codegen/viewmodel/metadata/VersionIndexView.java
jbolinger: Actually I take that back. That RequireView model thing has properties that are not really relevant to this type of require statement and would be confusing (clientName, etc.)

I think we'd want a new view model or need to split into a two types. Since there's only one property right now I'd propose we just leave it as a string and promote it to it's own type if, and when, we need others (and I can't think of any).

src/main/java/com/google/api/codegen/config/ProtoApiModel.java
jbolinger: Not really, I simply prefer protected per my own conventions. Do we have some standard conventions that would discourage this?

geigerj: The internal Java style guide would have this be private, but the external one seems to be more lenient. I'll follow up offline.

src/main/java/com/google/api/codegen/config/DiscoApiModel.java
```diff
@@ -62,6 +62,12 @@ public boolean hasMultipleServices() {
     return interfaceModels;
   }
 
+  @Override
+  public Iterable<? extends TypeModel> getAdditionalTypes() {
+    // TODO: is this supported?
+    return ImmutableList.of();
```
garrettjonesgoogle: I think it's better to throw `UnsupportedOperationException`

jbolinger: Yeah, I like that better too.

@landrito if you think there is a potential way to map this via discovery it's probably makes sense to file an issue as well. I'll defer that to you since I couldn't formulate a good description of the issue/work.

PR #1851, opened 2018-02-22
Java: Adding string-based overloads for resource name methods

garrettjonesgoogle:


garrettjonesgoogle:
@michaelbausor PTAL

src/main/java/com/google/api/codegen/transformer/FeatureConfig.java
```diff
@@ -41,6 +47,24 @@
...
+   * Returns true if useResourceNameFormatOptionInSampleOnly() is true but
+   * resourceNameProtoAccessorsEnabled() is false.
+   */
+  boolean useResourceNameConvertersInSampleOnly(FieldConfig fieldConfig);
+
+  /**
+   * Returns true if useResourceNameFormatOptionInSampleOnly() is true but
+   * resourceNameProtoAccessorsEnabled() is false.
+   */
+  boolean useResourceNameFormatOptionInSampleOnly(FieldConfig fieldConfig);
```
michaelbausor: What about a `useResourceNameFormatOptionInFlattenedSignature` method?

garrettjonesgoogle: That won't help the other branch in `InitCodeTransformer`, which needs to do a transformation when passing a resource name variable to a proto field.

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -669,7 +669,9 @@ private void setInitValueAndComments(
 
       String formatMethodName = "";
       String transformParamFunctionName = "";
-      if (context.getFeatureConfig().useResourceNameConverters(fieldConfig)) {
+      if (context.getFeatureConfig().useResourceNameConvertersInSampleOnly(fieldConfig)
```
michaelbausor: I think redo this conditional as:
```
if (context.getFeatureConfig().useResourceNameConvertersInSample(fieldConfig)
  && !(context.isFlattenedMethodContext() 
       && context.getFeatureConfig().useResourceNameConverters(fieldConfig)) {
```

This is as clear (or clearer?) and then you can remove both `SampleOnly` methods from feature config, because they are only used here.

garrettjonesgoogle: I feel that it is harder to reason about the boolean logic when expressed that way.

In the end, I'd prefer to do a simplification pass later by adding logic which analyzes source & target types, and automatically adds conversions when necessary. Our current approach of tracking every conversion site is too complex to reason about.

I have broken apart my current boolean expression and added comments so that it is easier to understand.

garrettjonesgoogle:
PTAL

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -669,9 +669,22 @@ private void setInitValueAndComments(
 
       String formatMethodName = "";
       String transformParamFunctionName = "";
-      if (context.getFeatureConfig().useResourceNameConvertersInSampleOnly(fieldConfig)
-          || (context.getFeatureConfig().useResourceNameConverters(fieldConfig)
-              && !context.isFlattenedMethodContext())) {
+
+      // If resource name converters should only used in the sample, we need to convert the resource
```
michaelbausor: `s/only used/only be used`

PR #1854, opened 2018-02-26
Java: remove apache commons dependency in smoke tests

neozwu:


src/test/java/com/google/api/codegen/testdata/java/java_library.baseline
```diff
@@ -4096,35 +4097,13 @@ import java.util.Arrays;
...
-              .argName("PROJECT-ID")
-              .required(true)
-              .build());
-      CommandLine cl = (new DefaultParser()).parse(options, args);
-      if (cl.hasOption("help")) {
-        HelpFormatter formater = new HelpFormatter();
-        formater.printHelp("SampleApp", options);
-      }
-      executeNoCatch(cl.getOptionValue("project_id"));
+      executeNoCatch(getProjectId());
```
garrettjonesgoogle: I think I still want to allow the user to pass in a project id on the command line for clients where we don't automatically look up the project id.

neozwu: With this setup, user can pass project id through sys prop and env var. Do you mean we still need a command line interface, treating smoke test as an app ?

garrettjonesgoogle: It's nice that we can configure through an environment variable, but I kind of wanted to retain the ability to pass in things like project id on the command line.

src/main/resources/com/google/api/codegen/java/smoke_test.snip
```diff
@@ -78,16 +77,16 @@
     {@apiMethod.responseTypeName} pagedResponse =
         client.{@apiMethod.name}(\
           {@sampleMethodCallArgList(apiMethod.initCode.fieldSettings)});
-    System.out.println(ReflectionToStringBuilder.toString(pagedResponse));
+    System.out.println(pagedResponse.toString());
```
vam-google: Please remove toStrings() as we discussed offline

src/main/resources/com/google/api/codegen/java/smoke_test.snip
```diff
@@ -24,32 +24,31 @@
...
-    }
-    executeNoCatch(cl.getOptionValue("project_id"));
+    executeNoCatch(getProjectId());
   @else
     executeNoCatch();
   @end
 @end
 
+@private getProjectIdForTest()
+  private static String getProjectId() {
```
vam-google: Please do it in same way as you did it here https://github.com/neozwu/google-cloud-java/blob/b9520a481c9c8fbac9b032eb77e0912eca21742a/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/v1/TopicAdminSmokeTest.java#L105

neozwu:
@vam-google PTAL

PR #1864, opened 2018-03-06
Add Cloud Compute methods to test API

andreamlin:
For more test coverage, tack on the following API methods in the testdata Compute API json:
- `compute.addresses.update`
- `compute.addresses.patch`
- `compute.addresses.aggregatedList`

Resolve ResourceName naming conflicts in configgen, e.g. when one resource name path template is `projects/{project}/region/{region}/addresses/{address}` and another is `projects/{project}/addresses/{address}`, create resource names `address` and `projectAddress` respectively.


src/main/java/com/google/api/codegen/discovery/Schema.java
```diff
@@ -56,6 +61,64 @@ public Schema dereference() {
...
+          localQueue.addAll(currentNode.properties().values());
+        }
+        if (currentNode.additionalProperties() != null) {
+          localQueue.add(currentNode.additionalProperties());
+        }
+      }
+
+      while (localQueue.peek() != null) {
+        Schema next = localQueue.poll();
+        nodeToPrevNode.put(next, currentNode);
```
michaelbausor: What if there are two or more paths to `next`, from different values of `currentNode`? This will overwrite any existing link from `next`.

If there can be no such link, does that mean this is a tree structure? If so, we don't need a full BFS.

andreamlin: It's ok if an existing link gets overriden because we only need to find one path, not all paths, from the target back to the source. As long as there is at least one value of `currentNode` for every key `next`, we can trace the back along an unbroken path from the target to start.

andreamlin: The schema is not guaranteed to be a tree.

andreamlin:
@michaelbausor this PR isn't as lean as it could be, so there is implementation of codegen outside of the configgen domain. The main point of this PR was to expose more of the test API, which did need some codegen changes in order to get around NPEs and other exceptions that cropped up when the generator was first run on the updated JSON API defintion. Thanks!

src/main/java/com/google/api/codegen/discovery/Schema.java
michaelbausor: What a structure with a bidirectional link (or any cycle) like this, searching for `D` from `A`:
```
A-->B<-->C-->D
```

Won't it go like this:
- currentNode=`A`, add `B` to `localQueue`, add `(B->A)` to `nodeToPrevNode`, add `B` to `queue`
- currentNode=`B`, add `C` to `localQueue`, add `(C->B)` to `nodeToPrevNode`, add `C` to `queue`
- currentNode=`C`, add `B` and `D` to `localQueue`
  - Process `B` in `localQueue`, add `(B->C)` to `nodeToPrevNode`, don't add `B` to `queue` as already visited
  - Process `D` in `localQueue`, add `(D->C` to `nodeToPrevNode`, terminate

Now there is no link from `B->A`, instead `B->C` causes an infinite loop. Is that right?

src/main/java/com/google/api/codegen/config/DiscoveryMethodModel.java
```diff
@@ -297,11 +297,9 @@ public boolean isIdempotent() {
   public Map<String, String> getResourcePatternNameMap(Map<String, String> nameMap) {
     Map<String, String> resources = new LinkedHashMap<>();
     for (Map.Entry<String, String> entry : nameMap.entrySet()) {
-      String resourceNameString =
-          DiscoGapicNamer.getResourceIdentifier(entry.getKey()).toLowerCamel();
-      if (DiscoGapicNamer.getResourceIdentifier(method.flatPath())
-          .toLowerCamel()
-          .equals(resourceNameString)) {
+      if (DiscoGapicNamer.getCanonicalPath(method).equals(entry.getKey())) {
```
garrettjonesgoogle: wait, I didn't know we were using static methods on DiscoGapicNamer - that prevents language-specific customization, right?

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
```diff
@@ -306,10 +307,12 @@ private static void validateType(
           throw new IllegalArgumentException(
               "typeRef " + typeRef + " not compatible with " + lineType);
         }
-        for (int i = 0; i < childKeys.size(); i++) {
-          if (!childKeys.contains(Integer.toString(i))) {
-            throw new IllegalArgumentException(
-                "typeRef " + typeRef + " must have ordered indices, got " + childKeys);
+        if (typeRef.getApiSource().equals(ApiSource.PROTO)) {
```
garrettjonesgoogle: Why is this block not done for Discovery? I'd really like to avoid source-specific logic in the init code logic.

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
```diff
@@ -353,7 +356,7 @@ private static void validateKeyValue(TypeModel parentType, String key) {
     if (parentType.isMap()) {
       TypeModel keyType = parentType.getMapKeyField().getType();
       validateValue(keyType, key);
-    } else if (parentType.isRepeated()) {
+    } else if (parentType.isRepeated() && parentType.getApiSource().equals(ApiSource.PROTO)) {
```
garrettjonesgoogle: Same question as above

src/main/java/com/google/api/codegen/config/DiscoveryMethodModel.java
andreamlin: The static methods are either
- language-agnostic functions, like in configgen or for parsing the `"list"` method name from the Discovery doc method name `compute.addresses.list`
- functions that return a Name object, so they can be formatted by a language-specific NameFormatter

I should be able to convert all the static methods to non-static methods if you'd think that'd make for better style or testing.

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
andreamlin: Whatever problem that hack had punted has gone away!
Removed.

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
andreamlin: Removed for now.
This may become a problem again when unit tests are generated from Discovery documents.

src/main/java/com/google/api/codegen/discovery/Schema.java
andreamlin: Thanks for the counterexample! Now only adding a node key to `nodeToPrevNode` iff that node has not been visited yet to guarantee that all insertions to the `nodeToPrevNode` map have unique keys.

Added a unit test in SchemaTest.java for this method on a very similar graph structure to above. It did run an infinite loop before I put in the code changes haha

andreamlin:
Baseline changes:
- addition of name resource path `"projects/{project}/addresses/{address}"` in test json requires
  - changes in DiscoGapicNamer to fully qualify resource names. This generates the ProjectAddress.java file, which is different from the existing Address.java file.
  - changes in ApiCallableTransformer to name an Address resource object parameter as `addressResource` rather than `address`, which would clash with the existing resource name path string `address`
  - changes in DiscoConfigTransformer to fully qualify resource names that have the same simple resource names, e.g. simple resource name `address` and the qualified resource name `projectAddress`
- addition of `AddressAggregatedList`, which contains new `AddressScopedList`, in test json requires
  - the `Schema.findChild()` method, because `AddressAggregatedList` must first traverse through `AddressScopedList` to retrieve the desired `List<Address>` response object.

The other changes are either for style or they result from bubbling the above diffs throughout the toolkit pipeline.

andreamlin:
Added unit test for Schema.findChild() and commented on baseline changes.

PTAL

src/main/java/com/google/api/codegen/config/DiscoveryMethodModel.java
garrettjonesgoogle: I think non-static would be best. In particular for functions returning Name, individual languages might want different name content, not simply different casing.

src/main/java/com/google/api/codegen/configgen/transformer/DiscoConfigTransformer.java
```diff
@@ -134,47 +141,30 @@ private LicenseView generateLicense() {
...
-
   /**
    * Examines all of the resource paths used by the methods, and returns a map from each unique
    * canonical resource path to a resource identifier (a short name used by the collection
    * configuration). Each resource path is merely a string describing the fields in the entity, and
    * the resource path might not be the same as the RPC endpoint URI. The resource identifier is
    * globally unique within each API. Many methods may use the same resource collection. The
-   * resource identifier will be qualified (with the name of a parent resource) iff there are two or
-   * more resource identifiers with different canonical resource paths.
+   * resource identifier will be full qualified iff there are two or more resource identifiers with
```
garrettjonesgoogle: s/full/fully/

src/main/java/com/google/api/codegen/config/DiscoveryMethodModel.java
andreamlin: Ok, removed all static methods from DiscoGapicNamer. In addition, removed the SurfaceNamer as a field inside DiscoGapicNamer and made the Surface a parameter in DiscoGapicNamer methods that require it. This allows DiscoGapicNamer to be used to generate language-agnostic config names without having a null inner SurfaceNamer.

src/main/java/com/google/api/codegen/configgen/transformer/DiscoConfigTransformer.java
andreamlin: Done.

andreamlin:
PTAL

src/main/java/com/google/api/codegen/config/DiscoGapicInterfaceConfig.java
```diff
@@ -139,7 +136,7 @@ static DiscoGapicInterfaceConfig createInterfaceConfig(
     if (methodConfigs != null) {
       for (MethodConfig methodConfig : methodConfigs) {
         Method method = ((DiscoveryMethodModel) methodConfig.getMethodModel()).getDiscoMethod();
-        String canonicalMethodPath = DiscoGapicNamer.getCanonicalPath(method);
+        String canonicalMethodPath = model.getDiscoGapicNamer().getCanonicalPath(method);
```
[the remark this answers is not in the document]
andreamlin: For sure. Moved the static-able util methods into a language-agnostic DiscoGapicParser class. Removed namers from the model classes.

andreamlin:
PTAL

src/main/java/com/google/api/codegen/discogapic/transformer/DiscoGapicParser.java
```diff
@@ -0,0 +1,155 @@
...
+import com.google.api.codegen.Inflector;
+import com.google.api.codegen.config.ResourceNameConfig;
+import com.google.api.codegen.discovery.Method;
+import com.google.api.codegen.discovery.Schema;
+import com.google.api.codegen.util.Name;
+import com.google.common.base.Strings;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+// Utility class for parsing Discovery document fields into canonical inputs for the discogapic transformer pipeline.
```
garrettjonesgoogle: Use javadoc style

PR #1865, opened 2018-03-07
Add block argument to all non-streaming generated methods

jbolinger:
This PR is related to a change in GAX to allow the underlying GRC operation to be exposed to the caller via an optional block: https://github.com/googleapis/gax-ruby/pull/118 

Streaming methods are not implemented. 

The current functionality is not changed:
```ruby
result = client.do_something(...)
```

This alternative calling style has been added (note that the variable same_result is shown for clarity, but it would not normally be used)
```ruby
same_result = client.do_something(...) do | result, op |
  # do stuff with result or metadata
  puts "#{result} with #{op.trailing_metadata}"
end
```

In addition, the following changes were also included:
  + An optional metadata parameter was added to generated clients. This metadata is sent on every call unless overridden with different call options.
  + A bug was fixed so that if metadata is specified via a call option the default metadata (x-goog-api-*, etc.) is not dropped.

### DIFF
--- a/src/test/java/com/google/api/codegen/testdata/ruby/ruby_library.baseline
+++ b/src/test/java/com/google/api/codegen/testdata/ruby/ruby_library.baseline
@@ -654,6 +654,8 @@ module Library
   #     or the specified config is missing data points.
   #   @param timeout [Numeric]
   #     The default timeout, in seconds, for calls made through this client.
+  #   @param metadata [Hash]
+  #     Default metadata to be sent with each request. This can be overridden on a per call basis.
   def self.new(*args, version: :v1, **kwargs)
     unless AVAILABLE_VERSIONS.include?(version.to_s.downcase)
       raise "The version: #{version} is not available. The available versions " \
@@ -802,11 +804,14 @@ module Library
     #   or the specified config is missing data points.
     # @param timeout [Numeric]
     #   The default timeout, in seconds, for calls made through this client.
+    # @param metadata [Hash]
+    #   Default metadata to be sent with each request. This can be overridden on a per call basis.
     def self.new \
         credentials: nil,
         scopes: nil,
         client_config: nil,
         timeout: nil,
+        metadata: nil,
         lib_name: nil,
         lib_version: nil
       kwargs = {
@@ -814,6 +819,7 @@ module Library
         scopes: scopes,
         client_config: client_config,
         timeout: timeout,
+        metadata: metadata,
         lib_name: lib_name,
         lib_version: lib_version
       }.select { |_, v| v != nil }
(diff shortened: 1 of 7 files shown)


src/test/java/com/google/api/codegen/testdata/ruby/ruby_longrunning.baseline
```diff
@@ -901,6 +908,7 @@ module Google
         google_api_client.freeze
 
         headers = { :"x-goog-api-client" => google_api_client }
+        headers.merge!(metadata) if metadata.is_a?(Hash)
```
geigerj: Silent failure may be confusing if `metadata` fails the type check; would it be better to remove the `if` and let `merge()` raise an error?

src/test/java/com/google/api/codegen/testdata/ruby/ruby_library.baseline
```diff
@@ -1259,12 +1269,13 @@ module Library
 
       def create_shelf \
           shelf,
-          options: nil
+          options: nil,
+          &block
```
geigerj: The method docs should probably mention this block?

src/test/java/com/google/api/codegen/testdata/ruby/ruby_longrunning.baseline
geigerj: I mean that with the current, if I call `MyClient.new(metadata: 'some-invalid-metadata')`, it will silently discard my (invalid) header. If you change this code to just

```
headers.merge!(metadata)
```

then the user will get `TypeError: no implicit conversion of String into Hash`, which is probably better than just ignoring the argument.

jbolinger: makes sense, how about:
```
headers.merge!(metadata) unless metadata.nil?
```

jbolinger:
@geigerj the unit test is certainly a good idea, but how do we deal with the releases? If I add a generated test then the gax dependency needs to be bumped to the latest.

src/test/java/com/google/api/codegen/testdata/ruby/ruby_longrunning.baseline
geigerj: LGTM

geigerj:
The autogen unit tests don't actually run on CI until you make a PR to google-cloud-ruby, at which point GAX should already have been updated, right?

jbolinger:
where is the gax version in the gemspec set?

geigerj:
The GAX version is set here: https://github.com/googleapis/googleapis/blob/master/gapic/packaging/dependencies.yaml

src/test/java/com/google/api/codegen/testdata/ruby/ruby_library.baseline
jbolinger: Good suggestion. The methods that already accepted a block appear to be missing this though.

jbolinger:
Let me know if the unit tests look like what you'd expect. In the meantime, I still need to find a way to run them all to verify they are valid.

src/test/java/com/google/api/codegen/testdata/ruby/ruby_library.baseline
geigerj: > The methods that already accepted a block appear to be missing this though.

Good catch! That should also probably be fixed (not necessarily in this PR).

geigerj:
Thanks, the new unit tests look good! I'm aiming to do another review pass sometime today.

PR #1867, opened 2018-03-12
Add true GAPIC_ONLY support.

vam-google:
Also fix proto dependencies generation, refactor and cleanup unused code.


src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
```diff
@@ -149,16 +143,8 @@ public String packageName(TargetLanguage language) {
...
-  /** Returns the type of the package */
   @Nullable
-  public abstract PackageType packageType();
-
-  /** Returns the type of the dependency */
-  @Nullable
-  public abstract DependencyType dependencyType();
-
-  @Nullable
-  public abstract GenerationLayer generationLayer();
```
garrettjonesgoogle: This looks like a breaking change, meaning we need to bump up the config schema version. With artman GA, I think we promised that we would backfill user config files if we made such breaking changes.

src/main/java/com/google/api/codegen/grpcmetadatagen/GrpcMetadataGeneratorTool.java
```diff
@@ -76,6 +76,16 @@ public static void main(String[] args) throws Exception {
             .argName("METADATA-CONFIG")
             .required(true)
             .build());
+    options.addOption(
+        Option.builder()
+            .longOpt("artifact_type")
+            .desc(
+                "Optional. Artifacts enabled for the generator. Currently supports "
+                    + "'grpc' and 'protobuf' and is ignored for all languges except Java")
```
garrettjonesgoogle: spelling (languges)

src/main/java/com/google/api/codegen/grpcmetadatagen/java/JavaProtoPackageMetadataTransformer.java
```diff
@@ -1,32 +0,0 @@
...
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-package com.google.api.codegen.grpcmetadatagen.java;
-
-import com.google.common.collect.ImmutableMap;
-import java.util.Map;
-
-/** Responsible for producing GRPC package meta-data related views for Java */
-public class JavaProtoPackageMetadataTransformer extends JavaPackageMetadataTransformer {
```
garrettjonesgoogle: clap clap clap (for deleting a subclass)

src/main/resources/com/google/api/codegen/metadatagen/java/grpc/build_base.gradle.snip
```diff
@@ -114,29 +116,6 @@
...
-      compile "io.grpc:grpc-protobuf:{@metadata.grpcVersionBound.lower}"
-      compile project(":{@metadata.protoPackageName}")
-    @default
-      $unhandledCase: {@metadata.generationLayer}$
-    @end
-  }
-@end
+@abstract dependencies(metadata)
+
+@abstract description(metadata)
```
garrettjonesgoogle: add newline (applies to other files too)

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
```diff
@@ -149,16 +143,8 @@ public String packageName(TargetLanguage language) {
...
+  public abstract ArtifactType artifactType();
```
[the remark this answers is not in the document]
vam-google: I did not choose this name. In essence this does not introduce a new config property (to tons other entities we already have), but instead propagates artman's artifact_type config in exactly same form to toolkit config. 

We already have `artman2_config -> artman_legacy_config -> toolkit_config` chain of conversions, which makes it super hard to understand (on each step same thing is called somehow differently, exercising in synonyms) . By keeping property name and its possible values same between artman and toolkit I hope to not add one more level of confusion here.

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
vam-google: This does not affect artman config, only toolkit config. Neither toolkit nor its config format are GA, so we should be ok here. Also nobody is using toolkit config directly (it is supposed to be generated by artman dynamically on generation step).

src/main/java/com/google/api/codegen/grpcmetadatagen/GrpcMetadataGeneratorTool.java
vam-google: Corrected

vam-google: Corrected

src/main/resources/com/google/api/codegen/metadatagen/java/grpc/build_base.gradle.snip
vam-google: added

vam-google:
@garrettjonesgoogle PTAL

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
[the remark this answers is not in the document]
garrettjonesgoogle: Ok, since it just propagates artman's input, it makes a lot of sense to keep it this way.

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
garrettjonesgoogle: Alright - as discussed offline, it's a generated intermediary config between artman & toolkit that isn't checked in anywhere. The consequence of this change basically means that after the push of artman & toolkit, you can't mix an older version of one & a newer version of the other.

src/main/java/com/google/api/codegen/grpcmetadatagen/java/JavaPackageMetadataTransformer.java
```diff
@@ -17,30 +17,42 @@
...
+  private final ArtifactType artifactType;
 
-  protected abstract Map<String, String> getSnippetsOutput();
+  public JavaPackageMetadataTransformer(
+      Map<String, String> snippetsOutput, ArtifactType artifactType) {
+    this.snippetsOutput = ImmutableMap.copyOf(snippetsOutput);
+    this.artifactType = artifactType;
+  }
+
+  protected Map<String, String> getSnippetsOutput() {
```
garrettjonesgoogle: Can we make this either private or public? We aren't expecting subclasses any more.

vam-google: There is still one subclass remaining: `JavaGapicMetadataTransformer`. But it makes sence, `JavaPackageMetadataTransformer` transforms for proto and grpc `build.gradle`, `JavaGapicMetadataTransformer` transforms for gapic.

garrettjonesgoogle: But does this need to remain `protected`?

vam-google: It does not, but when I make it private, IntelliJ highlights it and says it can be protected. I still can make it public, of course, do you feel strong about it?

vam-google: Btw, `JavaGapicMetadataTransformer` will go away in my next cleanup of toolkit (to remove sample app stuff). So, eventually this method can become private.

PR #1870, opened 2018-03-12
Use named parameters in snippets where ambiguous

### LINKED ISSUE #1841: C# possible snippet generation ambiguity on List methods
Occurs when there are two flattenings, where one has one extra string parameter.

chrisdunelm:
Fixes #1841

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSnippetsTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSnippetsTransformer.java
@@ -134,9 +134,13 @@ public class CSharpGapicSnippetsTransformer implements ModelToViewTransformer {
       } else if (methodConfig.isPageStreaming()) {
         if (methodConfig.isFlattening()) {
           ImmutableList<FlatteningConfig> flatteningGroups = methodConfig.getFlatteningConfigs();
-          // Find flattenings that have ambiguous parameters, and mark them to use named parameters.
-          // These are page-stream flattenings that have one or two string parameters (that are not
-          // resource-names) than another flattening.
+          // Find flattenings that have ambiguous parameters, and mark them to use named arguments.
+          // Ambiguity occurs in a page-stream flattening that has one or two extra string
+          // parameters (that are not resource-names) compared to any other flattening of this same
+          // method.
+          // Create a string for each flattening, encoding which parameters are strings and
+          // not-strings. Each character in the string refers to a parameter. Each string refers
+          // to a flattening.
           String[] stringParams =
               flatteningGroups
                   .stream()
@@ -157,6 +161,8 @@ public class CSharpGapicSnippetsTransformer implements ModelToViewTransformer {
                                   StringBuilder::append)
                               .toString())
                   .toArray(String[]::new);
+          // Array of which flattenings need to use named arguments.
+          // Each array entry refers to the correspondingly indexed flattening.
           Boolean[] requiresNamedParameters =
               Arrays.stream(stringParams)
                   .map(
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSnippetsTransformer.java
```diff
@@ -134,14 +134,49 @@ private SnippetsFileView generateSnippets(GapicInterfaceContext context) {
       } else if (methodConfig.isPageStreaming()) {
         if (methodConfig.isFlattening()) {
           ImmutableList<FlatteningConfig> flatteningGroups = methodConfig.getFlatteningConfigs();
+          // Find flattenings that have ambiguous parameters, and mark them to use named parameters.
```
jskeet: Nit: named arguments, not named parameters. (Including variable name, method names etc.)
I know it's a bit of a nit, but I'd rather we got it right from the start...

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSnippetsTransformer.java
```diff
@@ -134,14 +134,49 @@ private SnippetsFileView generateSnippets(GapicInterfaceContext context) {
...
+          // These are page-stream flattenings that have one or two string parameters (that are not
+          // resource-names) than another flattening.
+          String[] stringParams =
+              flatteningGroups
+                  .stream()
+                  .map(
+                      flat ->
```
jskeet: Is this "lambda parameter on its own" format a common one in Java? It looks very sparse to me...

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSnippetsTransformer.java
```diff
@@ -134,14 +134,49 @@ private SnippetsFileView generateSnippets(GapicInterfaceContext context) {
       } else if (methodConfig.isPageStreaming()) {
         if (methodConfig.isFlattening()) {
           ImmutableList<FlatteningConfig> flatteningGroups = methodConfig.getFlatteningConfigs();
+          // Find flattenings that have ambiguous parameters, and mark them to use named parameters.
+          // These are page-stream flattenings that have one or two string parameters (that are not
+          // resource-names) than another flattening.
```
jskeet: Not sure what "than another" means here.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSnippetsTransformer.java
```diff
@@ -134,14 +134,49 @@ private SnippetsFileView generateSnippets(GapicInterfaceContext context) {
...
+                                      field.getField().getType().isStringType()
+                                              && field.getResourceNameConfig() == null
+                                          ? 's'
+                                          : '.')
+                              .collect(
+                                  StringBuilder::new,
+                                  StringBuilder::appendCodePoint,
+                                  StringBuilder::append)
+                              .toString())
+                  .toArray(String[]::new);
```
jskeet: I have very little idea what the result is here. Is this code that exists somewhere else for another reasons?

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSnippetsTransformer.java
chrisdunelm: All changed.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSnippetsTransformer.java
chrisdunelm: Added more comments to try to explain. No, this code was written here specifically for this.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSnippetsTransformer.java
chrisdunelm: I've tried to explain better.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicSnippetsTransformer.java
chrisdunelm: (Thought I'd already answered this...)

This code is auto-formatted, and the formatting is enforced, so the sparseness cannot be changed,

PR #1874, opened 2018-03-15
Generate stub standalone sample files in Java and Python

vchudnov-g:
With these changes, **stubs** of Java and Python standalone samples are generated for
methods whose GAPIC configs are configured to do so via the method's
`sample_value_sets` and `samples` fields. Note that these changes are backwards-compatible: since currently released APIs do not have samples configured, no sample files are generated for those. The stubs currently in place use the existing in-code samples, which are not standalone.

I am also passing  the structured `initCode` for each sample to the standalone sample file templates. I was going to do this separately, but since it led to changes in the structure I had in earlier iterations of this PR, which was still under review, I am including it here.

I touch the path generator for other languages to account for a
modified interface. Let's not worry yet about the generated location of the
samples for those. At any rate, for all languages artman can be
configured to move the samples to the right final directory for the user.

High-level overview:

   - Main data structures:
     - Since API methods in general contain samples (in-code samples in the library surface itself, standalone samples in the standalone sample files), `{StaticLang,OptionalArray}MethodView` both now contain a getter `samples()` which returns a `List<MethodSampleView>`
     - `MethodSampleView` contains the view model for the method-level specific sample output, including the relevant `InitCodeView` (and `ApiMethodView.initCode` will be deprecated in favor of the new sample config)
     - `StaticLangSampleClassView` and `DynamicLangXSampleView` contain the top-level view model for the sample classes/files that show the usage of a particular method in the library surface.

   - High level flow
     - Configuration options get read into `SampleSpec`, now a part of `MethodConfig`
     - Transformation flow:
        `<Lang>Gapic{Surface,Samples}Transformer`: Entry point for generating the top level client or sample artifacts. Generates `StaticLang{Api,SampleClass}View`/`DynamicLangX{Api,Sample}View`
         -> `<Lang>MethodViewGenerator`: Logic for determining which method type (and eventually which calling forms) to select
         -> `{Static,Dynamic}LangApiMethodTransformer`: Generating the `ApiMethodView`s, including samples
         -> `SampleTransformer`: Generating the `MethodSampleView`s appropriate to the artifact being produced

   - Rendering entry points for samples are now `<LANG>/standalone_sample.snip`


vchudnov-g:
Please go ahead and review this PR. Subsequent work will go into a new PR I will push after this one gets committed.

src/main/java/com/google/api/codegen/transformer/py/PythonGapicSurfaceTransformer.java
```diff
@@ -175,6 +189,63 @@ private void addFieldsImports(ModelTypeTable typeTable, Iterable<FieldModel> fie
...
+  private Iterable<ViewModel> generateSampleClassesForApi(
+      GapicInterfaceContext context, DynamicLangXApiView apiClass) {
+    SurfaceNamer namer = context.getNamer();
+    ImmutableList.Builder<ViewModel> viewModelBuilder = ImmutableList.builder();
+
+    final Builder sampleClassBuilder = apiClass.toBuilder();
+    sampleClassBuilder.templateFileName(STANDALONE_SAMPLE_TEMPLATE_FILENAME);
+
+    for (ApiMethodView view : apiClass.apiMethods()) {
+      OptionalArrayMethodView methodView = (OptionalArrayMethodView) view;
```
garrettjonesgoogle: I'm uneasy of the general approach here, where the client surface is generated first, and then you crack it back open as a basis for the samples. For all of the other stuff we generate, it all starts from the source data - this is done for the client, in-code samples, and tests. I'd prefer to have your samples work the same way instead of working differently.

src/main/java/com/google/api/codegen/viewmodel/ApiMethodView.java
```diff
@@ -14,4 +14,6 @@
  */
 package com.google.api.codegen.viewmodel;
 
-public interface ApiMethodView {}
+public interface ApiMethodView {
+  String name();
```
garrettjonesgoogle: we use the `getX()` naming pattern in general - view models are the exception.

src/main/java/com/google/api/codegen/viewmodel/OptionalArrayMethodView.java
```diff
@@ -115,6 +116,16 @@ public boolean hasApiVersion() {
...
 
+  /** All the configured value sets for all samples for this method. */
+  public abstract SampleValueSetsModel sampleValueSetsModel();
+
+  /**
+   * When this class is being used as a view model for a single sample, the one value set to be used
+   * in the sample.
+   */
+  @Nullable
+  public abstract SampleValueSet sampleValueSet();
```
garrettjonesgoogle: Codegen proto-generated types should not be put in the view model classes.

src/main/resources/com/google/api/codegen/java/standalone_sample.snip
```diff
@@ -0,0 +1,49 @@
...
+     @let coreSampleCode = asyncOperationMethodSampleCode(apiMethod)
+       {@methodDoc(apiMethod.doc, decorateSampleCode(apiMethod, coreSampleCode))}
+     @end
+     */
+  @end
+@end
+
+# copied & modified from main.snip
+@private methodDoc(apiMethodDoc, methodSampleCode)
+  // AUTO-GENERATED DOCUMENTATION AND METHOD
```
garrettjonesgoogle: Maybe remove this

src/main/java/com/google/api/codegen/config/SampleSpec.java
```diff
@@ -0,0 +1,138 @@
...
+ * Class SampleSpec stores the sample specification for a given method, and provides methods to
+ * easily access them by calling form and sample type.
+ */
+public class SampleSpec {
+
+  /** A reference to the Sample Configuration from which the other fields are derived. */
+  private final SampleConfiguration sampleConfiguration;
+
+  /** All the SampleValueSets defined for this method, indexed by their IDs. */
+  private HashMap<String, SampleValueSet> valueSets;
```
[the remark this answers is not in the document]
vchudnov-g: Done

src/main/java/com/google/api/codegen/config/SampleSpec.java
```diff
@@ -0,0 +1,138 @@
...
+
+  /** The various types of supported samples. */
+  public enum SampleType {
+    INCODE,
```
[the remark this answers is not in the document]
vchudnov-g: Done

src/main/resources/com/google/api/codegen/java/standalone_sample.snip
vchudnov-g: It's just a stub, but OK.

src/main/java/com/google/api/codegen/viewmodel/OptionalArrayMethodView.java
vchudnov-g: OK, I created a `SampleValueSetsView` class and am using it in `SampleValueSetCollection` and elsewhere. I was going to do this sooner or later anyway because we'll need to pass the parsed representation of the initial values, as we do for the sample init fields.

    But I have to askwhy this blanket stricture if the proto-generated types are containers for the data we need that are already pre-defined and pre-populated? Is the concern about mutability? Would a deep copy work then?

    (As I said, this is moot right now, but I want to understand the thinking behind your comment.)

src/main/java/com/google/api/codegen/viewmodel/ApiMethodView.java
vchudnov-g: Is this just an FYI? This is a view model so this method name is fine, then, right?

src/main/java/com/google/api/codegen/transformer/py/PythonGapicSurfaceTransformer.java
vchudnov-g: OK, revamped this in both Python and Java. I'm not completely convinced it's simpler, but the extra structure did lead me to find an error earlier.

src/main/java/com/google/api/codegen/transformer/py/PythonGapicSurfaceTransformer.java
```diff
@@ -175,6 +189,63 @@ private void addFieldsImports(ModelTypeTable typeTable, Iterable<FieldModel> fie
...
+      SampleValueSetsModel valueSetsModel = (methodView).sampleValueSetsModel();
```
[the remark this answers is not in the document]
vchudnov-g: Done. Cruft!

src/main/java/com/google/api/codegen/transformer/py/PythonGapicSurfaceTransformer.java
```diff
@@ -129,18 +138,23 @@ public PythonGapicSurfaceTransformer(
...
+      serviceSurfaces.add(apiClass);
+      sampleSurfaces.addAll(generateSampleClassesForApi(context, (DynamicLangXApiView) apiClass));
     }
+
     GrpcDocView enumFile = generateEnumView(productConfig, modelTypeTable, namer, model.getFiles());
     if (!enumFile.elementDocs().isEmpty()) {
       serviceSurfaces.add(enumFile);
     }
-    return serviceSurfaces.build();
+    return serviceSurfaces.addAll(sampleSurfaces.build()).build();
```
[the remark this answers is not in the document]
vchudnov-g: Good point. Cruft? Done.

src/main/java/com/google/api/codegen/transformer/py/PythonGapicSurfaceTransformer.java
```diff
@@ -129,18 +138,23 @@ public PythonGapicSurfaceTransformer(
...
+
     for (InterfaceModel apiInterface : apiModel.getInterfaces()) {
       GapicInterfaceContext context =
           GapicInterfaceContext.create(
               apiInterface, productConfig, modelTypeTable, namer, featureConfig);
       addApiImports(context);
-      serviceSurfaces.add(generateApiClass(context));
+      final ViewModel apiClass = generateApiClass(context);
+      serviceSurfaces.add(apiClass);
+      sampleSurfaces.addAll(generateSampleClassesForApi(context, (DynamicLangXApiView) apiClass));
```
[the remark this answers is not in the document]
vchudnov-g: Done

src/main/java/com/google/api/codegen/transformer/py/PythonGapicSurfaceTransformer.java
```diff
@@ -175,6 +189,63 @@ private void addFieldsImports(ModelTypeTable typeTable, Iterable<FieldModel> fie
...
+   *     will be generated. The view models returned for each method are modified clones of the view
+   *     models for the methods in apiClass.
+   * @return A list of view models, each one corresponding to a distinct method sample
+   */
+  private Iterable<ViewModel> generateSampleClassesForApi(
+      GapicInterfaceContext context, DynamicLangXApiView apiClass) {
+    SurfaceNamer namer = context.getNamer();
+    ImmutableList.Builder<ViewModel> viewModelBuilder = ImmutableList.builder();
+
+    final Builder sampleClassBuilder = apiClass.toBuilder();
```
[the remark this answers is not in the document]
vchudnov-g: Done

src/main/java/com/google/api/codegen/transformer/py/PythonGapicSurfaceTransformer.java
```diff
@@ -175,6 +189,63 @@ private void addFieldsImports(ModelTypeTable typeTable, Iterable<FieldModel> fie
...
+  /**
+   * Generates a list of standalone sample view models for the the API, at most one for each calling
+   * form for each method, as specified in each method view model's sampleValueSetsModel (which
+   * ultimately derives from user-provided configuration). If no samples are configured, no sample
+   * files are produced.
+   *
+   * <p>TODO(vchudnov-g): Properly set the method form in the views thus generated.
+   *
+   * @param context the interface for whose methods the sample files will be generated
+   * @param apClass the previously generated API class view model for whose methods sample files
```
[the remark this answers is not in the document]
vchudnov-g: Done

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceTransformer.java
```diff
@@ -155,6 +166,81 @@ public JavaSurfaceTransformer(
...
+   *     models for the methods in apiFile.
+   * @return A list of view models, each one corresponding to a distinct method sample
+   */
+  private List<ViewModel> generateSampleFilesForApi(
+      InterfaceContext context, StaticLangFileView<StaticLangApiView> apiFile) {
+    List<ViewModel> sampleDocs = new ArrayList<>();
+
+    StaticLangApiView classView = apiFile.classView();
+
+    final StaticLangFileView.Builder<StaticLangApiView> sampleFileBuilder =
```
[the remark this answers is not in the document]
vchudnov-g: Done. But curious: why is that? Just because of the  clutter? I tend to like clarifying expectations for the compiler (and human readers) where the language allows. (I happen to like using `const int *const ptr` in C++)

src/main/java/com/google/api/codegen/transformer/StaticLangApiMethodTransformer.java
```diff
@@ -859,6 +873,18 @@ private boolean shouldAllowEmpty(MethodContext context, FieldModel field) {
...
+  private StaticLangApiMethodView buildMethodForCallingForm(
+      StaticLangApiMethodView.Builder methodViewBuilder,
+      MethodContext context,
+      ClientMethodType callingForm) {
+    return methodViewBuilder
+        .type(callingForm)
+        .sampleValueSetsModel(
+            new SampleValueSetsModel(context.getMethodConfig().getSampleSpec(), callingForm))
+        .build();
+  }
```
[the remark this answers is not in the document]
vchudnov-g: I think it clarifies the reading up above to have this helper method doing these common things

src/main/java/com/google/api/codegen/viewmodel/SampleValueSetsModel.java
```diff
@@ -0,0 +1,55 @@
...
+   * @param sampleSpec the sample specification from the user-provided config
+   * @param callingForm the calling form for which this SampleValueSetsModel will apply. This
+   *     selects which parts of sampleSpec get used.
+   */
+  public SampleValueSetsModel(SampleSpec sampleSpec, ClientMethodType callingForm) {
+    valueSetsBySampleType = new HashMap<>();
+    for (SampleType sampleType : SampleType.values()) {
+      valueSetsBySampleType.put(sampleType, sampleSpec.valueSetsMatching(callingForm, sampleType));
+    }
+  }
```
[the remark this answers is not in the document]
vchudnov-g: Could I put this logic in a Builder? I would prefer that for encapsulation purposes

src/main/java/com/google/api/codegen/viewmodel/SampleValueSetsModel.java
```diff
@@ -0,0 +1,55 @@
...
+import com.google.api.codegen.config.SampleSpec;
+import com.google.api.codegen.config.SampleSpec.SampleType;
+import java.util.HashMap;
+import java.util.Set;
+
+/**
+ * Implements a view model for the value sets specified in the configuration as relevant to a
+ * specific calling form.
+ */
+public class SampleValueSetsModel {
```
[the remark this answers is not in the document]
vchudnov-g: Done

src/main/java/com/google/api/codegen/gapic/GapicCodePathMapper.java
```diff
@@ -21,5 +21,16 @@
  * constructing a path to the GAPIC output.
  */
 public interface GapicCodePathMapper {
+
+  /** Returns the path to the directory containg the generated libraries */
```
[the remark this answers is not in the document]
vchudnov-g: Done

src/main/java/com/google/api/codegen/config/SampleSpec.java
```diff
@@ -0,0 +1,138 @@
...
+  private Iterable<String> getValueSetNamesMatchingExpression(String valueSetExpression) {
+    Predicate<String> expressionMatch =
+        new Predicate<String>() {
+          @Override
+          public boolean apply(@Nullable String id) {
+            return expressionMatchesId(valueSetExpression, id);
+          }
+        };
+
+    return Iterables.filter(valueSets.keySet(), expressionMatch);
```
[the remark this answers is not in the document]
vchudnov-g: Indeed. Inlined the one-liner.

src/main/java/com/google/api/codegen/config/SampleSpec.java
```diff
@@ -0,0 +1,138 @@
...
+   *
+   * <p>CAUTION: This is a stub at the moment, always returning true.
+   */
+  public static boolean expressionMatchesId(String expression, String id) {
+    // TODO(vchudnov-g): Implement more sophisticated matching as per design.
+    return id.matches(expression);
+  }
+
+  /** Returns the SampleValueSets that were specified for this methodForm and sampleType. */
+  public Set<SampleValueSet> valueSetsMatching(ClientMethodType methodForm, SampleType sampleType) {
```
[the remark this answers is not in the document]
vchudnov-g: Done

src/main/java/com/google/api/codegen/config/SampleSpec.java
```diff
@@ -0,0 +1,138 @@
...
+  public SampleSpec(MethodConfigProto methodConfigProto) {
+    this.sampleConfiguration = methodConfigProto.getSamples();
+    storeValueSets(methodConfigProto.getSampleValueSetsList(), methodConfigProto.getName());
+  }
+
+  /**
+   * Returns true if id is a match for the given expression. This is the function used to determine
+   * whether calling forms and value sets match expressions referencing them by ID.
+   *
+   * <p>CAUTION: This is a stub at the moment, always returning true.
```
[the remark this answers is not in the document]
vchudnov-g: Cruft. Good catch!

src/main/java/com/google/api/codegen/config/SampleSpec.java
```diff
@@ -0,0 +1,138 @@
...
+      case INCODE:
+        return sampleConfiguration.getInCodeList();
+      case EXPLORER:
+        return sampleConfiguration.getApiExplorerList();
+    }
+    return null;
+  }
+
+  /** Populates the map valueSets so as to be able to access each SampleValueSet by its ID. */
+  private void storeValueSets(List<SampleValueSet> sampleValueSets, String methodName) {
```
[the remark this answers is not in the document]
vchudnov-g: I disagree: This is an implementation detail that should not be exposed to clients. It's factored out merely as a helper method. I'd be OK to inline it back into the constructor.

src/main/java/com/google/api/codegen/transformer/StaticLangApiMethodTransformer.java
```diff
@@ -77,7 +78,8 @@ public StaticLangApiMethodView generatePagedFlattenedMethod(
     setListMethodFields(context, Synchronicity.Sync, methodViewBuilder);
     setFlattenedMethodFields(context, additionalParams, Synchronicity.Sync, methodViewBuilder);
 
-    return methodViewBuilder.type(ClientMethodType.PagedFlattenedMethod).build();
+    return buildMethodForCallingForm(
```
garrettjonesgoogle: The functions in this class aren't building methods for calling forms, they are building methods for method types. Another way to say it: a "calling form" is about the code to call a method; this class is about building the methods themselves.

src/main/java/com/google/api/codegen/transformer/StaticLangApiMethodTransformer.java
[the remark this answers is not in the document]
garrettjonesgoogle: There is not really anything substantial here, it's just plumbing parameters into a builder. I'd rather not have the abstraction...

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceTransformer.java
```diff
@@ -160,44 +180,80 @@ public JavaSurfaceTransformer(
   }
 
   private StaticLangFileView<StaticLangApiView> generateApiFile(
-      InterfaceContext context, GapicProductConfig productConfig) {
-    StaticLangFileView.Builder<StaticLangApiView> apiFile =
-        StaticLangFileView.<StaticLangApiView>newBuilder();
+      InterfaceContext context,
+      ApiAndSamples<StaticLangApiView, StaticLangApiMethodView> classView) {
```
garrettjonesgoogle: This feels weird to me - it the scope of the input is too large for the method. It should either be pure source input (the original structure, where it uses the API model from the context) or just the class view, not class view + sample view.

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceTransformer.java
```diff
@@ -122,9 +135,16 @@ public JavaSurfaceTransformer(
       InterfaceContext context =
           surfaceTransformer.createInterfaceContext(
               apiInterface, productConfig, namer, typeTable, enableStringFormatFunctions);
-      StaticLangFileView<StaticLangApiView> apiFile = generateApiFile(context, productConfig);
+
+      ApiAndSamples<StaticLangApiView, StaticLangApiMethodView> classViewWithSamples =
```
garrettjonesgoogle: If you have boolean words in your class name, you have the wrong abstraction.

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceTransformer.java
```diff
@@ -976,8 +1049,51 @@ private SettingsDocView generateSettingsDoc(
...
+    List<SampleInfo<StaticLangApiMethodView>> sampleInfos = new ArrayList<>();
+    for (SampleValueSetView values : matchingValueSets) {
+      sampleInfos.add(
+          sampleInfoBuilder
+              .valueSet(values)
+              .className(
+                  namer.getApiSampleClassName(
+                      methodView.name(),
+                      // TODO(vchudnov-g): Simplify so we don't need all these casing conversions
+                      Name.anyCamel(methodView.type().toString()).toLowerCamel(),
```
garrettjonesgoogle: Avoid casing conversions in transformer code

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceTransformer.java
```diff
@@ -976,8 +1049,51 @@ private SettingsDocView generateSettingsDoc(
     return settingsDoc.build();
   }
 
+  private ApiAndSamples<StaticLangApiMethodView, StaticLangApiMethodView> methodAndSamplesFor(
```
garrettjonesgoogle: This function doesn't create a method, it only creates samples.

Overall you have woven together method & sample generation, which mixes and couples concerns. I think it would be better for sample generation to run outside of the context of client+method generation.

src/main/java/com/google/api/codegen/transformer/py/PythonGapicSurfaceTransformer.java
```diff
@@ -176,75 +187,143 @@ private void addFieldsImports(ModelTypeTable typeTable, Iterable<FieldModel> fie
...
             context.getInterfaceModel().getFullName(), context.getProductConfig());
     String name = namer.getApiWrapperClassName(context.getInterfaceConfig());
-    List<ApiMethodView> methods = generateApiMethods(context);
+    List<ApiAndSamples<OptionalArrayMethodView, OptionalArrayMethodView>> methods =
+        generateApiMethodsAndSamples(context);
 
-    DynamicLangXApiView.Builder xapiClass = DynamicLangXApiView.newBuilder();
-    xapiClass.templateFileName(XAPI_TEMPLATE_FILENAME);
-    xapiClass.outputPath(namer.getSourceFilePath(subPath, name));
+    DynamicLangXApiView.Builder xapiClassBuilder = DynamicLangXApiView.newBuilder();
```
garrettjonesgoogle: I'd prefer not to add `Builder` on the variables all over the place because it adds so much text.

src/main/java/com/google/api/codegen/viewmodel/ApiMethodView.java
garrettjonesgoogle: Oh sorry, I didn't notice.

src/main/java/com/google/api/codegen/viewmodel/SampleInfo.java
```diff
@@ -0,0 +1,55 @@
...
+ * limitations under the License.
+ */
+package com.google.api.codegen.viewmodel;
+
+import com.google.auto.value.AutoValue;
+import javax.annotation.Nullable;
+
+/** Information for a single sample for a given method, to be included in other view models. */
+@AutoValue
+public abstract class SampleInfo<MethodViewT> {
```
garrettjonesgoogle: This looks like a view model, so it should have `View` at the end of the name

src/main/java/com/google/api/codegen/viewmodel/SampleInfo.java
```diff
@@ -0,0 +1,55 @@
...
+
+  /** The class name for the runnable sample */
+  @Nullable
+  public abstract String className();
+
+  /** The client library method for which this is a sample */
+  public abstract MethodViewT method();
+
+  /** The calling form exemplified in this sample */
+  public abstract String callingForm();
```
garrettjonesgoogle: This looks like an identifier for an object, not the object itself, so I suggest `callingFormName` or `callingFormId`

src/main/java/com/google/api/codegen/transformer/StaticLangApiMethodTransformer.java
vchudnov-g: Let's talk about calling forms. I set up a meeting.

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceTransformer.java
vchudnov-g: You're right, it only creates the samples because the method is passed in. The reason I chose the name is that it also creates and returns the method+samples bundle `ApiAndSamples`

"you have woven together method & sample generation, which mixes and couples concerns": Previously I had client generation and then the sample creation based off that. I changed it response to your comment on the earlier pass: "I'm uneasy of the general approach here, where the client surface is generated first, and then you crack it back open as a basis for the samples. For all of the other stuff we generate, it all starts from the source data - this is done for the client, in-code samples, and tests. I'd prefer to have your samples work the same way instead of working differently."

"I think it would be better for sample generation to run outside of the context of client+method generation.": To some extent yes, and that was what my first pass was trying to do. That said, the samples should be derived from the methods insofar as makes sense: I don't want to replicate the logic in two places. That's why my first approach inspected the set of methods created by the library.

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceTransformer.java
vchudnov-g: Yeah, not happy about that; that's what I put the TODO. I'll address it earlier, though, if you think it's a blocker.

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceTransformer.java
vchudnov-g: I'm happy to s/`And`/`With`/g, but that's not the point. This abstraction is there because samples are associated with methods. You don't have a sample floating off by itself: it always references a method. Contrariwise, a method will now come with one or more samples. That's the point of `ApiAndSamples`.

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceTransformer.java
vchudnov-g: This is in part in response to your concern about my first approach, which generating the client surface and then reading that to generate the samples. Let's discuss.

src/main/java/com/google/api/codegen/viewmodel/SampleInfo.java
vchudnov-g: moot

src/main/java/com/google/api/codegen/viewmodel/SampleInfo.java
vchudnov-g: moot

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceTransformer.java
vchudnov-g: ptal

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceTransformer.java
vchudnov-g: Done

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceTransformer.java
vchudnov-g: ptal

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceTransformer.java
vchudnov-g: ptal

src/main/java/com/google/api/codegen/transformer/StaticLangApiMethodTransformer.java
vchudnov-g: done

vchudnov-g:
PTAL. Refer to the commit comment/PR description for a high-level overview.

src/main/java/com/google/api/codegen/config/SampleSpec.java
```diff
@@ -0,0 +1,142 @@
...
+  /** The various types of supported samples. */
+  public enum SampleType {
+    IN_CODE,
+    STANDALONE,
+    EXPLORER,
+  }
+
+  public SampleSpec(MethodConfigProto methodConfigProto) {
+    specified = methodConfigProto.hasSamples();
+    this.sampleConfiguration = methodConfigProto.getSamples();
```
garrettjonesgoogle: Be consistent with using `this.` vs not using it

src/main/java/com/google/api/codegen/config/SampleSpec.java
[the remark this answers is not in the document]
garrettjonesgoogle: I wasn't suggesting that this be a public method, but instead a static private method.

src/main/java/com/google/api/codegen/transformer/java/JavaMethodViewGenerator.java
```diff
@@ -0,0 +1,147 @@
...
+public class JavaMethodViewGenerator {
+
+  final StaticLangApiMethodTransformer clientMethodTransformer;
+
+  public JavaMethodViewGenerator(SampleType sampleType) {
+    clientMethodTransformer = new StaticLangApiMethodTransformer(sampleType);
+  }
+
+  /**
+   * Generates the StaticLangApiMethodView for each of the methods in the InterfaceContext
```
garrettjonesgoogle: End sentence with a period.

src/main/java/com/google/api/codegen/transformer/java/JavaMethodViewGenerator.java
```diff
@@ -0,0 +1,147 @@
...
+   *
+   * @param context The context contianing the methods for which to generate StaticLangApiMethodView
```
garrettjonesgoogle: spelling(contianing)

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceTransformer.java
```diff
@@ -72,15 +70,19 @@
 public class JavaSurfaceTransformer {
   private final GapicCodePathMapper pathMapper;
   private final PackageMetadataConfig packageMetadataConfig;
+
+  // TODO: Simplify by making JavaSurfaceTransformer abstract and making
+  // JavaDiscoGapicSurfaceTransformer and JavaGapicSurfaceTransformer derive from it and override
+  // the methods that are used below.
```
garrettjonesgoogle: No, I don't want to make `JavaSurfaceTransformer` abstract. Such designs don't make the codebase simpler.

src/main/java/com/google/api/codegen/transformer/py/PythonGapicSurfaceTransformer.java
```diff
@@ -227,24 +234,30 @@ private ViewModel generateApiClass(GapicInterfaceContext context) {
...
+    xapiClass.outputPath(namer.getSourceFilePath(subPath, name));
+    xapiClass.name(name);
+    xapiClass.apiMethods(methods.stream().collect(Collectors.toList()));
+
     return xapiClass.build();
   }
 
-  private List<ApiMethodView> generateApiMethods(GapicInterfaceContext context) {
-    ImmutableList.Builder<ApiMethodView> apiMethods = ImmutableList.builder();
+  private List<ApiMethodView> generateApiMethodsAndSamples(GapicInterfaceContext context) {
```
garrettjonesgoogle: Shouldn't this just be `generateApiMethods`?

src/main/java/com/google/api/codegen/viewmodel/ApiMethodViewBuilder.java
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen.viewmodel;
+
+import java.util.List;
+
+// TODO(vchudnov-g): Consider making ApiMethodView an abstract class rather than an interface, and
+// including ApiMethodBuilder as an interface or abstract class within that.
```
garrettjonesgoogle: I don't think `ApiMethodView` needs to be an abstract class to contain this interface within it.

src/main/java/com/google/api/codegen/viewmodel/DynamicLangXSampleView.java
```diff
@@ -0,0 +1,56 @@
...
+  @Override
+  public String resourceRoot() {
+    return SnippetSetRunner.SNIPPET_RESOURCE_ROOT;
+  }
+
+  public abstract String outputPath();
+
+  public abstract String className();
+
+  public abstract OptionalArrayMethodView sampledMethod();
```
garrettjonesgoogle: should this be `sampleMethod`?

src/main/java/com/google/api/codegen/viewmodel/MethodSampleView.java
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen.viewmodel;
+
+import com.google.api.codegen.viewmodel.AutoValue_MethodSampleView.Builder;
```
garrettjonesgoogle: Remove this import

src/main/java/com/google/api/codegen/viewmodel/SampleValueSetView.java
```diff
@@ -0,0 +1,58 @@
...
+  /** The title as specified in the config */
+  public abstract String title();
+
+  /** The description as specified in the config */
+  public abstract String description();
+
+  /** The parameters as specified in the config */
+  public abstract List<String> parameters();
+
+  public static SampleValueSetView New(SampleValueSet config) {
```
garrettjonesgoogle: s/New/of/

src/main/java/com/google/api/codegen/viewmodel/StaticLangSampleClassView.java
```diff
@@ -0,0 +1,42 @@
...
+
+import com.google.auto.value.AutoValue;
+
+/** A ViewModel for standalone samples for static languages. */
+@AutoValue
+public abstract class StaticLangSampleClassView {
+
+  public abstract String name();
+
+  public abstract StaticLangApiMethodView sampledMethod();
```
garrettjonesgoogle: `sampleMethod`?

src/main/resources/com/google/api/codegen/java/method_sample.snip
```diff
@@ -53,22 +65,35 @@
...
+@snippet syncMethodSampleCode(apiMethod, initCode)
+  {@initCode(initCode)}
   @if apiMethod.hasReturnValue
     {@apiMethod.responseTypeName} response = {@sampleSyncMethodCall(apiMethod)};
   @else
     {@sampleSyncMethodCall(apiMethod)};
   @end
 @end
 
+# TODO(vchudnov-g): Deprecate this override
```
garrettjonesgoogle: Are all such overrides being deprecated? Also, deprecate or don't deprecate, don't "promise to deprecate."

src/main/resources/com/google/api/codegen/py/method_sample.snip
```diff
@@ -1,27 +1,38 @@
...
+          {@line}
+      @end
+@end
+
 @snippet sampleCode_(apiMethod)
-  {@importList(apiMethod.initCode.importSection.appImports)}
+  {@sampleCode_(apiMethod, apiMethod.initCode)}
+@end
+
+@snippet sampleCode_(apiMethod, initCode)
```
garrettjonesgoogle: Use a more informative naming scheme than adding `_`.

src/main/java/com/google/api/codegen/viewmodel/DynamicLangXSampleView.java
```diff
@@ -0,0 +1,56 @@
...
+ * limitations under the License.
+ */
+package com.google.api.codegen.viewmodel;
+
+import com.google.api.codegen.SnippetSetRunner;
+import com.google.auto.value.AutoValue;
+
+/** A ViewModel for standalone samples for dynamic languages. */
+@AutoValue
+public abstract class DynamicLangXSampleView implements ViewModel {
```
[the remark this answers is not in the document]
vchudnov-g: Oh, good. I was wondering why that X was everywhere. I take it you want me to do it here rather than doing it everywhere at once.

src/main/java/com/google/api/codegen/config/SampleSpec.java
vchudnov-g: Must have been cruft. I try to use `this` only to resolve ambiguity.

src/main/java/com/google/api/codegen/gapic/MainGapicProviderFactory.java
```diff
@@ -364,6 +374,13 @@
                 .setModelToViewTransformer(
                     new PythonGapicSurfaceTransformer(pythonPathMapper, packageConfig))
                 .build();
+        GapicProvider samplesProvider =
```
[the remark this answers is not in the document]
vchudnov-g: +1 for consistency

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
```diff
@@ -0,0 +1,181 @@
...
+  @FunctionalInterface
+  public interface generator {
+    public InitCodeView generate(InitCodeContext initCodeContext);
+  }
+
+  /**
+   * Constructs the SampleTransfomer
+   *
+   * @param sampleType The SampleType this transformer should concern itself with. Other SampleTypes
+   *     configured configured on the methods are ignored.
```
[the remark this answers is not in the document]
vchudnov-g: Done

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
```diff
@@ -0,0 +1,181 @@
...
+class SampleTransformer {
+
+  private final SampleType sampleType;
+
+  /**
+   * A functional interface provided by clients to generate an InitCodeView given an
+   * InitCodeContext.
+   */
+  @FunctionalInterface
+  public interface generator {
```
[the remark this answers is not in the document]
vchudnov-g: Done

src/main/java/com/google/api/codegen/transformer/java/JavaMethodViewGenerator.java
vchudnov-g: Done.  What's our convention for Java doc? It seems like in some cases we don't.

src/main/java/com/google/api/codegen/transformer/java/JavaMethodViewGenerator.java
vchudnov-g: Done

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceTransformer.java
vchudnov-g: I disagree. This seems like a case where inheritance would be quite appropriate, and certainly less confusing than both classes referencing each other and doing so by calling the other `surfaceTransformer` (which was the case before before I changed one to be `commonSurfaceTransformer`)

src/main/java/com/google/api/codegen/transformer/py/PythonGapicSurfaceTransformer.java
vchudnov-g: Actually, this is cruft that is no longer used. Removed.

src/main/java/com/google/api/codegen/viewmodel/ApiMethodViewBuilder.java
vchudnov-g: Oh, right you are! Did s/`ApiMethodViewBuilder`/`ApiMethodView.Builder`/g

src/main/java/com/google/api/codegen/viewmodel/DynamicLangXSampleView.java
vchudnov-g: No. With "sampled", I meant to convey that this is the client library method to which this sample pertains, not the interesting method of the sample. Added a comment to clarify

src/main/java/com/google/api/codegen/viewmodel/MethodSampleView.java
vchudnov-g: Done. Weird it didn't happen automatically.

src/main/java/com/google/api/codegen/viewmodel/SampleValueSetView.java
vchudnov-g: Done

src/main/java/com/google/api/codegen/viewmodel/StaticLangSampleClassView.java
vchudnov-g: Same rationale as in `DynamicLangSampleView`

src/main/resources/com/google/api/codegen/java/method_sample.snip
vchudnov-g: Yes. And fair enough.
Removed the now unused overloads.

src/main/resources/com/google/api/codegen/py/method_sample.snip
vchudnov-g: `samplecode_` is the name that was previously being used. I am just providing an override for this method.

I fully support a better name, but since this was pre-existing, maybe we should do that in a separate PR.

vchudnov-g:
PTAL, @garrettjonesgoogle

src/main/java/com/google/api/codegen/viewmodel/DynamicLangXSampleView.java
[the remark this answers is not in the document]
garrettjonesgoogle: Correct, I wanted to take care of this opportunistically in order to have minimal disturbance.

src/main/java/com/google/api/codegen/transformer/java/JavaMethodViewGenerator.java
garrettjonesgoogle: Convention: Add javadoc when it seems like an explanation is necessary.

When we decide we want more out-of-team adoption down the road, we'll have to start being more thorough.

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceTransformer.java
garrettjonesgoogle: This is a conversation about a design choice that doesn't impact the current PR. Since we have disagreement, I would prefer to have a separate conversation about this design choice instead of closing on it within the present PR (which would be required in order to codify your TODO).

src/main/java/com/google/api/codegen/viewmodel/DynamicLangXSampleView.java
garrettjonesgoogle: Hmm, the wording feels very weird - I've never heard the verb "sample" used in this way, e.g. "this code snippet samples API method X." It sounds like a statistical thing. I would suggest something else like `subjectMethod` or the like.

src/main/resources/com/google/api/codegen/py/method_sample.snip
garrettjonesgoogle: Oh! I didn't notice that. Carry on...

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceTransformer.java
vchudnov-g: I agree. That's why I put it as a TODO :-)

If you want, I can alter the wording s/`TODO: Simplify`/`TODO: Consider simplifying`/g

src/main/java/com/google/api/codegen/transformer/java/JavaMethodViewGenerator.java
vchudnov-g: _That_ part I got (even if in some cases I disagreed with the amount of information previously provided) and is what I've been doing: javadoc where it seems illuminating.

I meant my question more superficially: what is the convention about end-of-line punctuation in javadoc?

garrettjonesgoogle: The period drives javadoc logic - it defines the end of the short-form comment displayed in summary tables. http://www.oracle.com/technetwork/java/javase/tech/index-137868.html (search for "period")

src/main/java/com/google/api/codegen/viewmodel/DynamicLangXSampleView.java
vchudnov-g: I don't feel strongly about this, so OK, I'll change it to "libraryMethod" and keep the comment I added.

(While I often tend to be a stickler for correct usage, this doesn't bother me much. We neologize all the time in English. The meaning seems clear to me, though granted, I have author's bias)

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceTransformer.java
vchudnov-g: Oh, this will be an interesting discussion! I look forward to it.

In the meantime, I'll change the TODO as you suggest.

src/main/java/com/google/api/codegen/transformer/java/JavaMethodViewGenerator.java
vchudnov-g: Cool, thanks, I was looking for something like that. It seems we don't always follow that, but I'll use that as my style guide in this repo.

vchudnov-g:
PTAL, @garrettjonesgoogle

PR #1879, opened 2018-03-16
Remove paged response wrappers

andreamlin:


garrettjonesgoogle:
I think this is just something that was never put into use for DiscoGapic - @andreamlin can verify.

andreamlin:
Just removing the import line.

PR #1881, opened 2018-03-17
Generate unit tests on HTTP transport

andreamlin:
Add new http_unit_test.snip to render discogapic HTTP/1 unit tests.

Allow for multiple function calls to retrieve an paged resource from the RPC response message. e.g if `AddressClient.listAggregatedAddresses()` should return a `List<Address>`, but the RPC response message is an `AddressAggregatedList` which contains a `AddressScopedList getItems()` getter which itself contains the `List<Address> getAddressesList()` getter, then the generated paged list descriptor can then call `payload.getItems().getAddressesList()` to retrieve the paged resource list.


andreamlin:
This PR continues on https://github.com/googleapis/toolkit/pull/1864. I will clean up this commit history whenever that one gets merged in.

andreamlin:
PTAL

src/main/java/com/google/api/codegen/discogapic/ViewModelDiscoGapicProvider.java
```diff
@@ -1,4 +1,4 @@
-/* Copyright 2018 Google LLC
+/* Copyright 2016 Google LLC
```
garrettjonesgoogle: Why is the year changing here?

src/test/java/com/google/api/codegen/testdata/discogapic/java/java_simplecompute.v1.json.baseline
```diff
@@ -868,7 +868,11 @@ public final class Address implements ApiMessage {
       fieldMap.put("status", Collections.singletonList(String.valueOf(status)));
     }
     if (fieldNames.contains("users") && users != null) {
-      fieldMap.put("users", Collections.singletonList(String.valueOf(users)));
+      List<String> stringList = new LinkedList<>();
+      for (String item : users) {
+        stringList.add(item.toString());
+      }
+      fieldMap.put("users", stringList);
```
garrettjonesgoogle: I think it would be better to use an `ImmutableList` instead of `LinkedList`.

src/test/java/com/google/api/codegen/testdata/discogapic/java/java_simplecompute.v1.json.baseline
```diff
@@ -1117,12 +1121,20 @@ public final class Address implements ApiMessage {
...
       return users;
     }
 
     public Builder addAllUsers(List<String> users) {
-      this.users = users;
+      if (this.users == null) {
+        this.users = new LinkedList<>();
+      }
+      this.users.addAll(users);
+      return this;
```
garrettjonesgoogle: I think it would be better to use an `ArrayList` with a capacity initialized to the size of `users`

src/test/java/com/google/api/codegen/testdata/discogapic/java/java_simplecompute.v1.json.baseline
```diff
@@ -10077,30 +10149,23 @@ public class AddressClientTest {
...
 
     AddressName address = AddressName.of("[PROJECT]", "[REGION]", "[ADDRESS]");
 
     Operation actualResponse =
         client.deleteAddress(address);
     Assert.assertEquals(expectedResponse, actualResponse);
 
-    List<GeneratedMessageV3> actualRequests = mockAddresses.getRequests();
+    List<String> actualRequests = mockService.getRequestPaths();
     Assert.assertEquals(1, actualRequests.size());
```
garrettjonesgoogle: Why is there no comparison of the request content, or check for the header being sent?

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceTestTransformer.java
```diff
@@ -426,13 +429,25 @@ private void addUnitTestImports(InterfaceContext context) {
         typeTable.saveNicknameFor("java.util.concurrent.ExecutionException");
         break;
       case HTTP:
-        typeTable.saveNicknameFor("com.google.api.gax.rpc.StatusCode.Code");
-        typeTable.saveNicknameFor("com.google.api.gax.rpc.testing.FakeStatusCode");
-        typeTable.saveNicknameFor("com.google.api.gax.httpjson.MockHttpService");
+        typeTable.saveNicknameFor("com.google.api.gax.httpjson.ApiMessage");
+        typeTable.saveNicknameFor("com.google.api.gax.httpjson.ApiMethodDescriptor");
+        typeTable.saveNicknameFor("com.google.api.gax.httpjson.testing.MockHttpService");
```
garrettjonesgoogle: This depends on https://github.com/googleapis/gax-java/pull/493 which is still WIP, right?

src/main/resources/com/google/api/codegen/java/http_unit_test.snip
```diff
@@ -0,0 +1,135 @@
+@extends "java/common.snip"
+@extends "java/initcode.snip"
+
+@snippet generate(xapiTest)
```
garrettjonesgoogle: I would like to use the convention of an `x` prefix

src/main/resources/com/google/api/codegen/java/message.snip
```diff
@@ -169,13 +169,6 @@
     @join param : schema.properties
       this.{@param.name} = {@param.name};
     @end
-    ImmutableMap.Builder<String, String> mapBuilder = ImmutableMap.builder();
-    @if schema.hasResourceName
-      {@schema.pathAsResourceName.typeName} resourceName = {@schema.pathAsResourceName.typeName}.parse({@schema.pathAsResourceName.name});
-      @join nameField : schema.pathAsResourceName.fieldCopyMethods
-        mapBuilder.put("{@nameField.name}", resourceName.{@nameField.fieldGetFunction}());
-      @end
-    @end
```
garrettjonesgoogle: Why was this unnecessary?

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceTestTransformer.java
andreamlin: This depends on https://github.com/googleapis/gax-java/pull/492, which is a PR that can/should be reviewed in parallel with this toolkit PR.

andreamlin: Although yes good catch I'm reverting the MockHttpService package to `com.google.api.gax.httpjson`.

src/main/java/com/google/api/codegen/discogapic/ViewModelDiscoGapicProvider.java
andreamlin: whoops fixed now

src/main/resources/com/google/api/codegen/java/http_unit_test.snip
andreamlin: I assume this means you don't want the `x` prefix, so it's been removed.

src/test/java/com/google/api/codegen/testdata/discogapic/java/java_simplecompute.v1.json.baseline
andreamlin: I hadn't gotten that far in writing the unit tests. 
Put in TODOs in the http_unit_test.snip for both of those assertion checks.

src/main/resources/com/google/api/codegen/java/message.snip
andreamlin: Two things were removed there:
- the mapBuilder that was intended hold the map of field names to their values. This map wasn't ever used in later in the class because I hadn't finished implementing it. If you want, I put a TODO to store the field values in a map instead of in their own separate fields.
- the fields that made up a resource name. Recently, a resource name is just stored as a single String instead of its constituent String parts, and this ApiMessage instance has no idea how to parse a resource name.

src/test/java/com/google/api/codegen/testdata/discogapic/java/java_simplecompute.v1.json.baseline
andreamlin: Done. I'm going to assume the user knows not to call this method on a null param.

src/test/java/com/google/api/codegen/testdata/discogapic/java/java_simplecompute.v1.json.baseline
andreamlin: Done.

andreamlin:
Addressed comments. PTAL

src/main/resources/com/google/api/codegen/java/message.snip
garrettjonesgoogle: Sounds good. Storing in separate fields works good I think.

src/main/resources/com/google/api/codegen/java/http_unit_test.snip
```diff
@@ -0,0 +1,137 @@
+@extends "java/common.snip"
```
garrettjonesgoogle: It seems like either this file should be named http_test.snip or the grpc one should be named grpc_unit_test.snip.

PR #1882, opened 2018-03-19
Update php client constructor

michaelbausor:
- Merges service address and port settings
- Adds serviceScopes field
- Simplifies constructor options


src/main/resources/com/google/api/codegen/php/client_impl.snip
```diff
@@ -318,65 +319,36 @@
...
+         *     @@type string $serviceAddress Required. The address of the API remote host. Must be formatted as "<uri>:<port>".
      @end
+     *     @@type array $keyFile The contents of the service account credentials
+     *            .json file retrieved from the Google Developer's Console.
+     *            Ex: `json_decode(file_get_contents($path), true)`.
+     *     @@type string $keyFilePath The full path to your service account
+     *            credentials .json file retrieved from the Google Developers
+     *            Console.
+     *     @@type string $clientConfig
+     *            Array containing client method configuration, including retry settings. If this argument
```
vam-google: Does this documentation come from somewhere (like .proto file)? There are few inconsistencies in this documentation (like different maximum right margin for top and bottom fields (bottom is much wider), .json vs JSON when talking about json files etc. If it is external, then ok, if it is not, please make it moreconsistent.

PR #1886, opened 2018-03-21
Add support for explicit gem name

jbolinger:
Add a set of key value pairs to the language settings in the gapic config to enable language specific overrides. 

Use the new settings to implement a ruby "gem_name" override that can be used to explicitly control the name of the gem (by default it is a derivative of the package name).

Note the changes to `PackageMetadataTransformer.java` are primarily to avoid having to change all of the other transformers (i.e. non ruby). We could update those now instead or wait to change if/when they use an override setting as I've done here.


jbolinger:
Where could it be documented?

landrito:
It _could_ be documented in the configuration proto but it almost seems like if you're going to document it in that way, it should simply be a formally declared field.

jbolinger:
a declared field in what though? Like adding a `RubyLanguageSettingsProto` or something along those lines?

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -122,6 +122,9 @@ message LanguageSettingsProto {
 
   // The release level of the client in the language
   ReleaseLevel release_level = 6;
+
+  // Language specific flags to control fine-grained properties in package managers
+  map<string, string> packaging_settings = 7;
```
landrito: I mean to say that this field could changed to `string package_name = 7` which would provide more documentation of the configuration.

jbolinger: The issue with that is that there is already an `string package_name = 1` but that's not the same thing as the gem name.

landrito: Gah, unfortunately thats a bit of a misnomer. We could call it `metadata_identifier`.

jbolinger: I considered that too. The issue here is now we're projecting gem_name (or whatever we call it) on to all supported languages and I'm not sure that makes sense. i.e. how do I know that a metadata_identifier is the gem name for Ruby? What is it for C#, etc.?

To me, the extra indirection is at least as confusing as creating an arbitrary map of strings.

landrito: > how do I know that a metadata_identifier is the gem name for Ruby

I think the documentation can be both general enough to cover multiple languages but precise enough for one to imply that this field will override the gem_name that is inferred from the package_name field.

 Furthermore we can guard all of the languages that this doesn't make sense for by throwing an error by default if that field is set.

jbolinger: @lukesneeringer thoughts on this?

src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceTransformer.java
```diff
@@ -187,16 +192,24 @@ private ViewModel generateApiClass(GapicInterfaceContext context, String package
     xapiClass.apiMethods(methods);
 
     xapiClass.toolkitVersion(GeneratorVersionProvider.getGeneratorVersion());
+
+    final boolean isLongRunning =
+        RubyUtil.isLongrunning(context.getProductConfig().getPackageName());
+
+    // use explicit gem name if provided in the configuration
+    Map<String, String> packageSettings =
```
landrito: Note: We might be able to not have this ultra specific logic if we go with this solution.

jbolinger:
@lukesneeringer any thoughts on this? I can't think of a way to do this that I like, so maybe you can come up with one!

jbolinger:
Confirmed that we'll enforce the gem name matching the namespace. Closing as this enhancement is not required by anyone at this time.

PR #1890, opened 2018-03-23
Add more test object types and methods to discogapic test API

andreamlin:
See simplecompute.v1.json for test API changes.
- Add a method (`compute.dummyObjects.delete`) with a null response type
- Add a float field and a double field to a message type
- Add, in one message type, two fields with different names but same reference schemas

Changes in simplecompute_gapic.yaml and baseline files were automatically generated from running the test cases.

There are some hacks in the generator code put in just to avoid NPEs, but those will be removed after handling these corner cases is done. (Fixes in https://github.com/googleapis/toolkit/pull/1892).

### DIFF
--- a/src/test/java/com/google/api/codegen/testdata/discogapic/java/java_simplecompute.v1.json.baseline
+++ b/src/test/java/com/google/api/codegen/testdata/discogapic/java/java_simplecompute.v1.json.baseline
@@ -211,6 +211,195 @@ public final class AddressName implements ResourceName {
   }
 }
 
+============== file: src/main/java/com/google/cloud/simplecompute/v1/DummyObjectName.java ==============
+/*
+ * Copyright 2018 Google LLC
+ *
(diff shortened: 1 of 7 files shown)


andreamlin:
PTAL @garrettjonesgoogle

src/test/java/com/google/api/codegen/testdata/discogapic/java/java_simplecompute.v1.json.baseline
```diff
@@ -2365,88 +2590,352 @@ public final class Error implements ApiMessage {
...
+
+  public Double getPrecisionFloatie() {
+    return precisionFloatie;
+  }
+
+  public Address getAddress() {
+    return primaryAddress;
+  }
+
+  public Address getAddress() {
```
andreamlin: This duplicate naming issue will be fixed in https://github.com/googleapis/toolkit/pull/1892.

andreamlin:
PTAL

PR #1892, opened 2018-03-23
Handle API methods with empty response types

andreamlin:
Render empty response types as java.lang.Void in places like `UnaryCallable<RequestT, Void>`.
Create a new TypeModel, `EmptyTypeModel` as the in-memory placeholder for the null response types.


andreamlin:
This PR can be reviewed on its own, but it would be helpful to have https://github.com/googleapis/toolkit/pull/1890 checked in first. The other PR contains test resource changes, and this PR would have more helpful baseline changes once the test resources get updated.

andreamlin:
PTAL @garrettjonesgoogle

src/main/java/com/google/api/codegen/config/DiscoveryField.java
```diff
@@ -38,8 +38,11 @@
 /** A field declaration wrapper around a Discovery Schema. */
 public class DiscoveryField implements FieldModel, TypeModel {
   private final List<DiscoveryField> properties;
+  // Dereferenced schema for use rendering type names and determining properties, type, and format.
```
garrettjonesgoogle: s/for use/to use for/

src/main/java/com/google/api/codegen/transformer/SchemaTypeTable.java
```diff
@@ -289,9 +295,12 @@ public String getSnippetZeroValueAndSaveNicknameFor(FieldModel type) {
 
   @Override
   public String getSnippetZeroValueAndSaveNicknameFor(TypeModel type) {
-    return typeNameConverter
-        .getSnippetZeroValue(((DiscoveryField) type).getDiscoveryField())
-        .getValueAndSaveTypeNicknameIn(typeTable);
+    Schema schema = null;
+    if (!(type instanceof EmptyTypeModel)) {
```
garrettjonesgoogle: Doesn't `TypeModel` have `isEmptyType`?

src/main/java/com/google/api/codegen/config/DiscoveryField.java

src/main/java/com/google/api/codegen/transformer/SchemaTypeTable.java
andreamlin: Oh even better!

andreamlin:
Removed some `instanceof` checks. PTAL

src/main/resources/com/google/api/codegen/java/http_stub.snip
```diff
@@ -65,12 +65,14 @@
...
-          .setResponseParser(
-              ApiMessageHttpResponseParser.<{@methodDescriptor.responseTypeName}>newBuilder()
-                  @if methodDescriptor.hasResponse
-                    .setResponseInstance({@methodDescriptor.responseTypeName}.getDefaultInstance())
-                  @end
-                  .build())
+          @if methodDescriptor.hasResponse
+            .setResponseParser(
+                ApiMessageHttpResponseParser.<{@methodDescriptor.responseTypeName}>newBuilder()
+                    @if methodDescriptor.hasResponse
```
garrettjonesgoogle: This if-check is redundant with the outer one

PR #1898, opened 2018-03-26
enable running smoke tests in staging repo

neozwu:
This change is to enable running smoke tests in staging repo (by invoking "./gradlew smokeTest").

### DIFF
--- a/src/main/resources/com/google/api/codegen/java/build_gapic.gradle.snip
+++ b/src/main/resources/com/google/api/codegen/java/build_gapic.gradle.snip
@@ -21,7 +21,9 @@
     compile 'com.google.api:gax:{@metadata.gaxVersionBound.lower}'
     compile 'com.google.api:gax-grpc:{@metadata.gaxGrpcVersionBound.lower}'
     testCompile 'junit:junit:4.12'
+    testCompile 'commons-lang:commons-lang:2.6'
     testCompile 'com.google.api:gax-grpc:{@metadata.gaxGrpcVersionBound.lower}:testlib'
+    testCompile 'io.grpc:grpc-netty-shaded:{@metadata.grpcVersionBound.lower}'
     @switch metadata.artifactType.toString
     @case "GAPIC_ONLY"
       @join dependency : metadata.protoPackageDependencies
@@ -35,6 +37,13 @@
     @join dependency : metadata.protoPackageTestDependencies
       testCompile project(':grpc-{@dependency.name}')
     @end
+  }
+
+  task smokeTest(type: Test) {
+    filter {
+      includeTestsMatching "*SmokeTest"
+      setFailOnNoMatchingTests false
+    }
   }
 
   sourceSets {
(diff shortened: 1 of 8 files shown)


src/main/resources/com/google/api/codegen/java/build_gapic.gradle.snip
```diff
@@ -21,7 +21,9 @@
     compile 'com.google.api:gax:{@metadata.gaxVersionBound.lower}'
     compile 'com.google.api:gax-grpc:{@metadata.gaxGrpcVersionBound.lower}'
     testCompile 'junit:junit:4.12'
+    testCompile 'commons-lang:commons-lang:2.6'
```
garrettjonesgoogle: Shouldn't this be removed?

src/main/resources/com/google/api/codegen/java/build_gapic.gradle.snip
```diff
@@ -21,7 +21,9 @@
...
     testCompile 'com.google.api:gax-grpc:{@metadata.gaxGrpcVersionBound.lower}:testlib'
+    testCompile 'io.grpc:grpc-netty-shaded:{@metadata.grpcVersionBound.lower}'
```
garrettjonesgoogle: Why is this necessary?

src/main/resources/com/google/api/codegen/java/build_gapic.gradle.snip
neozwu: Done. Yes, I guess it's just really hard to actually remove commons-lang.

src/main/resources/com/google/api/codegen/java/build_gapic.gradle.snip
neozwu: I think we need this or the right combination of  (grpc-netty + netty-tcnative-boringssl-static) to send real requests. I figured using grpc-netty-shaded is a lot easier to manage. But you probably have more insights here.

neozwu:
@garrettjonesgoogle PTAL

src/main/resources/com/google/api/codegen/java/smoke_test.snip
```diff
@@ -11,6 +11,11 @@
     private static final String PROJECT_ENV_NAME = "GOOGLE_CLOUD_PROJECT";
     private static final String LEGACY_PROJECT_ENV_NAME = "GCLOUD_PROJECT";
 
+    @@Test
+    public void run() {
+      main(null);
+    }
```
garrettjonesgoogle: Will smoke tests run when we do `./gradlew test`? It seems like that would make api-client-staging fail, right? (since a project wouldn't be defined)

neozwu: It will. And it will fail unless you have an env var to denote project_id.  I now configured the build script so that smoke tests are excluded when running `./gradlew test`.

PR #1900, opened 2018-03-28
Handle map return type

andreamlin:
In Discovery doc schemas, the `additionalProperties` field is a [dynamically-keyed map type](https://developers.google.com/discovery/v1/reference/apis). For now, assume the keys are Strings. There are pageable methods, e.g. `compute.addresses.aggregatedList`, that must page over this `Map<String, ? extends ApiMessage>` type.

In discogapic configgen, make `page_streaming.response.resources_field` the name of the field in the response containing the list of resources belonging to the page, as the config.proto has dictated. See simplecompute_gapic.yaml for resulting output changes.

For paged methods, scrap the concept of a list of field getter methods to get from a response object to the paged resource object. It's too complex and hard to determine the series of fields to traverse. Assume the paged resource object is a direct field of the response object.

PageStreamingConfig constructor is now same regardless of model type (e.g. proto/discovery)


andreamlin:
PTAL

andreamlin:
PTAL

PR #1903, opened 2018-03-29
C# Prevent type name collisions

### LINKED ISSUE #1868: C#: use fully-qualified names for system Task types in snippets
Currently if an API describes a Task type, the generated snippets don't compile:

```csharp
/// <summary>Snippet for ListQueuesAsync</summary>
public async Task ListQueuesAsync()
```

... that `Task` isn't the right Task type.

chrisdunelm:
By using namespace aliasing

Fixes #1868


src/test/java/com/google/api/codegen/testdata/csharp/csharp_library.baseline
```diff
@@ -5218,28 +5235,28 @@ namespace Google.Example.Library.V1.Tests
...
-using System;
-using System.Collections;
-using System.Collections.Generic;
-using System.Collections.ObjectModel;
+using grpccore = Grpc.Core;
+using s = System;
+using sc = System.Collections;
+using scg = System.Collections.Generic;
+using sco = System.Collections.ObjectModel;
 using System.Linq;
```
jskeet: It feels odd for this to be unqualified. Any reasons?

src/test/java/com/google/api/codegen/testdata/csharp/csharp_library.baseline
```diff
@@ -5218,28 +5235,28 @@ namespace Google.Example.Library.V1.Tests
...
+using gax = Google.Api.Gax;
+using gaxgrpc = Google.Api.Gax.Grpc;
 using Google.Cloud.Tagger.V1;
-using Google.LongRunning;
-using Google.Protobuf;
-using Google.Protobuf.WellKnownTypes;
+using lro = Google.LongRunning;
+using proto = Google.Protobuf;
+using protowkt = Google.Protobuf.WellKnownTypes;
 using Google.Tagger.V1;
```
jskeet: Why do we have two different namespaces for Tagger? Is that deliberate?

src/test/java/com/google/api/codegen/testdata/csharp/csharp_library.baseline
```diff
@@ -5218,28 +5235,28 @@ namespace Google.Example.Library.V1.Tests
...
-using Grpc.Core;
-using System;
-using System.Collections;
-using System.Collections.Generic;
-using System.Collections.ObjectModel;
+using grpccore = Grpc.Core;
+using s = System;
```
jskeet: I'm wondering whether "sys", "syscoll" and "syscollgen" might be better names for these aliases. See what you think.

jskeet:
Issues found generating all clients:

- WithCallSettingsOverlay doesn't qualify CallSettings
- The Bigtable code generator will need to be altered (easy enough to do; just adding regular using directives is enough)
- In Datastore, an optional ByteString value generates `Transaction = transaction ?? ByteString.Empty, // Optional`; ByteString needs to be qualified. (Ditto in Spanner.)
- Logging: `IResourceName` isn't qualified in the LogSink.DestinationAsResourceName property (in either declaration or comment)

(Currently running unit tests, but that's all that's needed to get it all compiling. Pretty impressively small list of further changes required!)

src/main/java/com/google/api/codegen/util/csharp/CSharpImports.readme
```diff
@@ -0,0 +1,8 @@
+CSharpImports.java is an auto-generated file,
```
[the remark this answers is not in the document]
chrisdunelm: Done :)

src/test/java/com/google/api/codegen/testdata/csharp/csharp_library.baseline
```diff
@@ -1082,16 +1098,16 @@ namespace Google.Example.Library.V1.Snippets
         /// <summary>Snippet for UpdateBookAsync</summary>
         public async Task UpdateBookAsync2()
         {
-            // Snippet: UpdateBookAsync(BookName,string,Book,FieldMask,FieldMask,CallSettings)
-            // Additional: UpdateBookAsync(BookName,string,Book,FieldMask,FieldMask,CancellationToken)
+            // Snippet: UpdateBookAsync(BookName,string,Book,FieldMask,apis::FieldMask,CallSettings)
```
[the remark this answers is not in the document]
chrisdunelm: As discussed, the snippet generator will be changed to handle qualifiers.

src/test/java/com/google/api/codegen/testdata/csharp/csharp_library.baseline
chrisdunelm: Unqualified because of extension method use. I'd rather leave this as it is for now. If it causes problems, then it can be changed in a future PR.

src/test/java/com/google/api/codegen/testdata/csharp/csharp_library.baseline
chrisdunelm: This is not due to this PR, it had both imports before. I suspect this may be due to a bad configuration in the test library; but this can be followed up in a future PR.

src/test/java/com/google/api/codegen/testdata/csharp/csharp_library.baseline
chrisdunelm: Personally I'm happy with the shorter aliases, so would rather leave them as they are. They can always be changed in the future if they cause confusion.

chrisdunelm:
All issues now fixed.

PR #1905, opened 2018-03-29
Generate resources field in page streaming config.proto

andreamlin:
In discogapic configgen, make `page_streaming.response.resources_field` the name of the field in the response containing the list of resources belonging to the page, as the config.proto has dictated. See simplecompute_gapic.yaml for resulting output changes.

For paged methods, scrap the concept of a list of field getter methods to get from a response object to the paged resource object. It's too complex and hard to determine the series of fields to traverse. Assume the paged resource object is a direct field of the response object.

`PageStreamingConfig` constructor is now same regardless of model type (e.g. proto/discovery)

### DIFF
--- a/src/main/java/com/google/api/codegen/configgen/transformer/DiscoveryMethodTransformer.java
+++ b/src/main/java/com/google/api/codegen/configgen/transformer/DiscoveryMethodTransformer.java
@@ -29,6 +29,9 @@ import javax.annotation.Nullable;
 public class DiscoveryMethodTransformer implements InputSpecificMethodTransformer {
   private final PagingParameters pagingParameters = new HttpPagingParameters();
 
+  // For Discovery doc configgen, assume that paged resource field name is "items".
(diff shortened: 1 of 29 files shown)


src/main/java/com/google/api/codegen/configgen/transformer/DiscoveryMethodTransformer.java
```diff
@@ -29,6 +29,9 @@
 public class DiscoveryMethodTransformer implements InputSpecificMethodTransformer {
   private final PagingParameters pagingParameters = new HttpPagingParameters();
 
+  // For Discovery doc configgen, assume that paged resource field name is "items".
```
andreamlin: add in comments: only resource name seen for compute

andreamlin: done.

andreamlin:
Closing this PR and reopening https://github.com/googleapis/toolkit/pull/1900.

PR #1906, opened 2018-03-29
Remove exception from close method

michaelbausor:
Remove exception from close method. WIP while verifying that this generates correct code for current APIs.


src/main/resources/com/google/api/codegen/java/grpc_test.snip
```diff
@@ -39,7 +39,7 @@
     }
 
     @@After
-    public void tearDown() throws Exception {
+    public void tearDown() {
```
garrettjonesgoogle: These `throws Exception` clauses don't harm anyone, I think we should leave them. (JUnit is the only layer concerned really)

michaelbausor: OK, I will revert.

michaelbausor:
Note that while I have removed `throws Exception` from the static apply methods in https://github.com/googleapis/gax-java/pull/521, I have NOT removed them from the generated applyToAllUnaryMethods method in the settings class, because this would technically be a breaking change for anyone extending that class. Should I include a grep-able note saying that this should be removed in the next major version? It is less clear what the right approach here is for toolkit.

PR #1907, opened 2018-03-29
Discogapic: Handle map return type

andreamlin:
In Discovery doc schemas, the `additionalProperties` field is a [dynamically-keyed map type](https://developers.google.com/discovery/v1/reference/apis). For now, assume the keys are Strings. There are pageable methods, e.g. `compute.addresses.aggregatedList`, that must page over this `Map<String, ? extends ApiMessage>` type.

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/SchemaTypeTable.java
+++ b/src/main/java/com/google/api/codegen/transformer/SchemaTypeTable.java
@@ -178,9 +178,8 @@ public class SchemaTypeTable implements ImportTypeTable, SchemaTypeFormatter {
 
   @Override
   public String getFullNameFor(TypeModel type) {
-    // TODO(andrealin): Remove this hack when null response types are implemented.
-    if (type == null) {
-      return "nullFullName";
+    if (type.isEmptyType()) {
+      return "java.lang.Void";
     }
     if (type instanceof DiscoveryRequestType) {
       Method method = ((DiscoveryRequestType) type).parentMethod().getDiscoMethod();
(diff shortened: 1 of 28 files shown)


src/main/java/com/google/api/codegen/transformer/SchemaTypeTable.java
```diff
@@ -178,9 +178,8 @@ public String getFullNameFor(InterfaceModel type) {
 
   @Override
   public String getFullNameFor(TypeModel type) {
-    // TODO(andrealin): Remove this hack when null response types are implemented.
-    if (type == null) {
-      return "nullFullName";
+    if (type.isEmptyType()) {
+      return "java.lang.Void";
```
andreamlin: make a SchemaTypeNameConverter.getTypeNameForEmptyType() and call that here.

src/main/java/com/google/api/codegen/transformer/java/JavaSchemaTypeNameConverter.java
```diff
@@ -14,9 +14,11 @@
  */
 package com.google.api.codegen.transformer.java;
 
+import com.google.api.codegen.config.DiscoveryField;
```
andreamlin: move type switching logic to language subclasses

andreamlin: done.

src/main/java/com/google/api/codegen/transformer/SchemaTypeTable.java
andreamlin: nvm

andreamlin:
Closing this PR and reopening https://github.com/googleapis/toolkit/pull/1900.

PR #1908, opened 2018-03-30
Add static resource provider

vam-google:
Java: add gradle build scripts and gradle wrapper to allow building and testing generated files out of the box with no prerequisites except installed JDK.


garrettjonesgoogle:
@andreamlin could you do a first pass?

src/main/resources/com/google/api/codegen/java/sample.gradle.snip
```diff
@@ -1,185 +0,0 @@
-@snippet generate(metadata)
```
andreamlin: Why is this file being deleted?

vam-google: It is not used anymore. I forgot to delete this file during one of my previous PRs.

vam-google:
Addressed PR feedback, @garrettjonesgoogle PTAL. Note I had to keep the `Collection<String> getInputFileNames()` method in the interface, otherwise it caused too much ugliness in tests.

src/main/java/com/google/api/codegen/gapic/StaticResourcesProvider.java
```diff
@@ -0,0 +1,54 @@
+/* Copyright 2016 Google LLC
```
garrettjonesgoogle: 2018

src/main/java/com/google/api/codegen/grpcmetadatagen/py/PythonPackageCopier.java
```diff
@@ -72,12 +73,14 @@ public FileVisitResult postVisitDirectory(Path dir, IOException exc) {
...
-        docBuilder.put(outFile, Doc.text("\n"));
+        docBuilder.put(outFile, GeneratedResult.create(Doc.text("\n"), false));
 
         // All others get become namespace packages
       } else {
         docBuilder.put(
-            outFile, Doc.text("__import__('pkg_resources').declare_namespace(__name__)\n"));
+            outFile,
+            GeneratedResult.create(
+                Doc.text("__import__('pkg_resources').declare_namespace(__name__)\n"), false));
```
garrettjonesgoogle: In this context it's hard to figure out what the boolean parameter means. Could we either create an enum or use some other way to make it clear here?

src/main/java/com/google/api/codegen/util/StaticResourcesHandler.java
```diff
@@ -0,0 +1,45 @@
+/* Copyright 2016 Google LLC
```
garrettjonesgoogle: 2018

src/main/java/com/google/api/codegen/grpcmetadatagen/py/PythonPackageCopier.java
vam-google: I can't think of any reasonable enum name for this (and would like to avoid some enum, which results into. At long last the parameter has a name, so it should be enough to describe what it does. If param name `executable` is not descriptive, I'm happy to change it to something else.

I added documentation for the `GeneratedResult<T>` and its methods to makes things less confusing.

vam-google:
@garrettjonesgoogle PTAL

PR #1909, opened 2018-04-03
Make Discogapic a gradle executable

andreamlin:
Expose the Discogapic client library and config generator methods as runnable gradle tasks.


andreamlin:
PTAL

PR #1911, opened 2018-04-03
Discogapic: Fully qualify generated ResourceNames

andreamlin:
Context: in the newest compute.v1.json Discovery doc, there is a resource name generated that would be called class "ResourceName", which is a problem for Java client libraries because this conflicts with the api-common dependency ResourceName. ~Simplest solution seems to be to just fully qualify and remove the api-common import.~ Fully qualify all generated ResourceNames to solve this. In Google Cloud, all resource names will begin with "Project", because "project" is required in the path.

### DIFF
--- a/src/main/java/com/google/api/codegen/config/AnyResourceNameConfig.java
+++ b/src/main/java/com/google/api/codegen/config/AnyResourceNameConfig.java
@@ -15,6 +15,7 @@
 package com.google.api.codegen.config;
 
 import com.google.api.tools.framework.model.ProtoFile;
+import javax.annotation.Nullable;
 
 /**
  * AnyResourceNameConfig is a singleton configuration indicating acceptance of any resource name
@@ -47,6 +48,12 @@ public class AnyResourceNameConfig implements ResourceNameConfig {
   @Override
   public String getEntityName() {
     return ENTITY_NAME;
+  }
+
+  @Override
+  @Nullable
+  public String getCommonResourceName() {
+    return null;
   }
 
   @Override
--- a/src/main/java/com/google/api/codegen/config/FixedResourceNameConfig.java
+++ b/src/main/java/com/google/api/codegen/config/FixedResourceNameConfig.java
@@ -35,6 +35,12 @@ public abstract class FixedResourceNameConfig implements ResourceNameConfig {
   public abstract ProtoFile getAssignedProtoFile();
 
   @Override
+  @Nullable
+  public String getCommonResourceName() {
+    return null;
+  }
+
+  @Override
   public ResourceNameType getResourceNameType() {
     return ResourceNameType.FIXED;
   }
--- a/src/main/java/com/google/api/codegen/config/ResourceNameConfig.java
+++ b/src/main/java/com/google/api/codegen/config/ResourceNameConfig.java
@@ -15,6 +15,7 @@
 package com.google.api.codegen.config;
 
 import com.google.api.tools.framework.model.ProtoFile;
+import javax.annotation.Nullable;
 
 public interface ResourceNameConfig {
 
@@ -23,6 +24,9 @@ public interface ResourceNameConfig {
 
   /** Returns the name used as a basis for generating methods. */
   String getEntityName();
+
+  @Nullable
+  String getCommonResourceName();
 
   /** Returns the resource name type. */
   ResourceNameType getResourceNameType();
--- a/src/main/java/com/google/api/codegen/config/ResourceNameOneofConfig.java
+++ b/src/main/java/com/google/api/codegen/config/ResourceNameOneofConfig.java
@@ -38,6 +38,12 @@ public abstract class ResourceNameOneofConfig implements ResourceNameConfig {
   @Override
   public abstract ProtoFile getAssignedProtoFile();
 
+  @Override
+  @Nullable
+  public String getCommonResourceName() {
+    return null;
+  }
+
   public Iterable<SingleResourceNameConfig> getSingleResourceNameConfigs() {
     return Iterables.filter(getResourceNameConfigs(), SingleResourceNameConfig.class);
   }
--- a/src/main/java/com/google/api/codegen/config/SingleResourceNameConfig.java
+++ b/src/main/java/com/google/api/codegen/config/SingleResourceNameConfig.java
@@ -24,6 +24,7 @@ import com.google.api.tools.framework.model.DiagCollector;
 import com.google.api.tools.framework.model.ProtoFile;
 import com.google.api.tools.framework.model.SimpleLocation;
 import com.google.auto.value.AutoValue;
+import com.google.common.base.Strings;
 import javax.annotation.Nullable;
 
 /** SingleResourceNameConfig represents the collection configuration for a method. */
@@ -51,16 +52,22 @@ public abstract class SingleResourceNameConfig implements ResourceNameConfig {
     String entityId = collectionConfigProto.getEntityName();
     String entityName = entityId;
     String language = configProto.getLanguage();
+    String commonResourceName = null;
     if (language != null) {
       for (CollectionLanguageOverridesProto override :
           collectionConfigProto.getLanguageOverridesList()) {
         if (language.equals(override.getLanguage())) {
-          entityName = override.getEntityName();
+          if (!Strings.isNullOrEmpty(override.getEntityName())) {
+            entityName = override.getEntityName();
+          }
+          if (!Strings.isNullOrEmpty(override.getCommonResourceName())) {
+            commonResourceName = override.getCommonResourceName();
+          }
         }
       }
     }
     return new AutoValue_SingleResourceNameConfig(
-        namePattern, nameTemplate, entityId, entityName, file);
+        namePattern, nameTemplate, entityId, entityName, commonResourceName, file);
   }
 
   /** Returns the name pattern for the resource name config. */
@@ -76,6 +83,10 @@ public abstract class SingleResourceNameConfig implements ResourceNameConfig {
   /** Returns the name used as a basis for generating methods. */
   @Override
   public abstract String getEntityName();
+
+  @Override
+  @Nullable
+  public abstract String getCommonResourceName();
 
   @Override
   @Nullable
--- a/src/main/java/com/google/api/codegen/transformer/PathTemplateTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/PathTemplateTransformer.java
@@ -123,7 +123,8 @@ public class PathTemplateTransformer {
             .enumName(namer.getResourceEnumName(config))
             .docName(config.getEntityName())
             .index(index)
-            .pattern(config.getNamePattern());
+            .pattern(config.getNamePattern())
+            .commonResourceName(config.getCommonResourceName());
     List<ResourceNameParamView> params = new ArrayList<>();
     int varIndex = 0;
     for (String var : config.getNameTemplate().vars()) {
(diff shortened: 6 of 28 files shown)


src/test/java/com/google/api/codegen/testdata/discogapic/simplecompute.v1.json
```diff
@@ -948,6 +948,28 @@
   },
   "dummyObjects": {
    "methods": {
+    "getResource": {
+     "id": "compute.dummyObjects.getResource",
+     "path": "{project}/dummyObjects/{resource}",
```
andreamlin: This file change introduces a test case for a resource path that would be rendered as a class called `ResourceName`.

src/test/java/com/google/api/codegen/testdata/simplecompute_config.baseline
```diff
@@ -300,6 +300,8 @@ interfaces:
   collections:
   - name_pattern: projects/{project}/dummyObjects/{dummyObject}
     entity_name: dummyObject
+  - name_pattern: projects/{project}/dummyObjects/{resource}
```
andreamlin: Changes in this file are just a direct, autogenerated consequence of the changes in compute.v1.json.

src/test/java/com/google/api/codegen/testdata/simplecompute_gapic.yaml
```diff
@@ -242,6 +242,8 @@ interfaces:
   collections:
   - name_pattern: projects/{project}/dummyObjects/{dummyObject}
     entity_name: dummyObject
+  - name_pattern: projects/{project}/dummyObjects/{resource}
```
andreamlin: Changes in this file are also just a direct, autogenerated consequence of the changes in compute.v1.json.

michaelbausor:
Changes LGTM. Do we want to consider an alternative approach, which would be to avoid generating `resource` as the default entity name? The advantage of this is that it is less likely to be confusing for users, but it does seem like it would be harder to implement. Is there a sensible way to generate an alternate name, perhaps using the existing conflict resolution mechanism that is used when two names would conflict with each other? (e.g. `project/{project}/books/{book}` and `project/{project}/libraries/{library}/books/{book}`)

src/test/java/com/google/api/codegen/testdata/discogapic/java/java_simplecompute.v1.json.baseline
```diff
@@ -943,6 +938,194 @@ public final class RegionName implements ResourceName {
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import javax.annotation.Generated;
+
+@Generated("by GAPIC")
+@BetaApi
+public final class ResourceName implements com.google.api.resourcenames.ResourceName {
```
garrettjonesgoogle: Alternative: call this `ProjectResourceName`

andreamlin: Done. I liked this as a more consistently surfaced solution, where *all* resource names get fully qualified.

andreamlin:
PTAL
I think this newer solution is easier for users to understand and doesn't require a verbose `com.google.api.resourcenames.ResourceName` in the snippet.

Thanks for both your suggestions!

src/main/java/com/google/api/codegen/discogapic/transformer/DiscoGapicParser.java
```diff
@@ -100,7 +101,24 @@ public static String getCanonicalPath(Method method) {
       }
       namePattern = m.appendTail(sb).toString();
     }
-    return namePattern;
+
+    // Assume, based on the Google Compute API, that there is no more than one wildcard segment in a row.
+    // Now the path segments alternate between exactly one literal segment and exactly one wildcard segment.
```
michaelbausor: Does this always have to be the case? What about something like `projects/{project}/qualifier/regions/{region}` - is this possible? And if so, do we also need to handle potential collisions, such as `projects/{project}/alternate/regions/{region}`?

andreamlin: Definitely valid points, thanks!

- Removed the assumption that "there is no more than one wildcard segment in a row".
- Yes, now handling potential conflicts like the ones you gave as examples. Those would now be two different resource names, with the java class names `ProjectQualifierRegionName` and `ProjectAlternateRegionName` respectively.

andreamlin:
PTAL
Changed the implementation of qualifying ResourceName names to resolve more path template collisions. The resulting implementation has less total complexity, IMHO.
See DiscoGapicParserTest for more legible examples of method calls and results.

src/main/java/com/google/api/codegen/discogapic/transformer/DiscoGapicParser.java
```diff
@@ -142,20 +128,34 @@ public static Name methodAsName(Method method) {
...
-    for (String piece : pieces) {
-      if (!piece.contains("{")) {
-        if (name == null) {
-          name = Name.from(Inflector.singularize(piece));
-        } else {
-          name = name.join(stringToName(Inflector.singularize(piece)));
+    for (int i = 0; i < pieces.length; i++) {
+      if (i < pieces.length - 1 && pieces[i + 1].contains("{")) {
+        // Current index at a literal segment directly preceding a wildcard segment.
+        // Skip literal segment iff it is the same as or the pluralized version of the wildcard string.
```
michaelbausor: I may be reading this wrong, but if we skip the segments that precede `{bracketed}` segments, isn't that the same as singularizing the preceding segment then skipping the bracketed segment, which is what the original code is doing? The only difference is when the preceding segment is not a pluralized version of the bracketed segment, when we include both. Is that what we want?

If that is what we want, then let's always include the unbracketed segment, and skip the bracketed one when it matches the previous. This lets us avoid indexing and lookahead. We could even just always check for duplication, so something like:
```
Name name = Name.from();
String previous = null;
for (String piece : pieces) {
  String next = Inflector.singularize(piece);
  if (!next.equals(previous)) {
    // Only append to the name if this segment is not identical to the previous segment
    name = name.join(stringToName(next));
    previous = next;
  }
}
```
Does that work?

src/main/java/com/google/api/codegen/discogapic/transformer/DiscoGapicParser.java
```diff
@@ -142,20 +128,34 @@ public static Name methodAsName(Method method) {
...
+            || Inflector.singularize(pieces[i].toLowerCase()).equals(wildCard.toLowerCase())) {
+          continue;
         }
       }
+
+      String namePiece = pieces[i];
+      if (namePiece.contains("}")) {
+        namePiece = namePiece.substring(1, namePiece.length() - 1);
+      }
+      if (name == null) {
```
michaelbausor: Let's initialize `name` with `Name name = Name.from()`, then we can remove this if statement.

src/test/java/com/google/api/codegen/discogapic/transformer/DiscoGapicParserTest.java
```diff
@@ -0,0 +1,61 @@
...
+    Truth.assertThat(
+            DiscoGapicParser.getQualifiedResourceIdentifier(
+                    "{project}/global/backendBuckets/{backendBucket}")
+                .toUpperCamel())
+        .isEqualTo("ProjectGlobalBackendBucket");
+    Truth.assertThat(
+            DiscoGapicParser.getQualifiedResourceIdentifier(
+                    "{project}/zones/{zone}/disks/{resource}")
+                .toUpperCamel())
+        .isEqualTo("ProjectZoneDiskResource");
```
michaelbausor: Re above - would we prefer just `ProjectZoneDisk`?

src/main/java/com/google/api/codegen/discogapic/transformer/DiscoGapicParser.java
andreamlin: Yes, thanks so much!

src/main/java/com/google/api/codegen/discogapic/transformer/DiscoGapicParser.java
andreamlin: Sure.

src/test/java/com/google/api/codegen/discogapic/transformer/DiscoGapicParserTest.java
andreamlin: I just put in a test case statement right above that one to show that `ProjectZoneDiskResource` is preferable as the return value.

src/test/java/com/google/api/codegen/discogapic/transformer/DiscoGapicParserTest.java
```diff
@@ -48,6 +48,10 @@ public void testGetQualifiedResourceIdentifier() {
                     "{project}/global/backendBuckets/{backendBucket}")
                 .toUpperCamel())
         .isEqualTo("ProjectGlobalBackendBucket");
+    Truth.assertThat(
+            DiscoGapicParser.getQualifiedResourceIdentifier("{project}/zones/{zone}/disks/{disk}")
+                .toUpperCamel())
+        .isEqualTo("ProjectZoneDisk");
```
andreamlin: This extra test checks that 
`{project}/zones/{zone}/disks/{disk}`
 resolves to a different resourceName than
`{project}/zones/{zone}/disks/{resource}`
The is important because the respective gapic ResourceName classes will have a field/getter/setter name for the `disk` and `resource` field, respectively, so they need to be in separate classes.

andreamlin:
PTAL
Thanks so much for the simplification suggestions!

src/test/java/com/google/api/codegen/discogapic/transformer/DiscoGapicParserTest.java
michaelbausor: This is interesting - should we be saying "actually, these are the same format, so they should use the same ResourceName class, and having two ResourceName classes with the same format is invalid"? The argument would be otherwise you can have an odd situation where you want one type but you have the other. The argument against might be, if the same format is used in two different places for different purposes, then we don't want them to be interchangeable. In practice I don't know if this every happens, because resource names often map to REST URLs, so the same formats can't be used for different purposes in that way. I'm not sure what the right answer is here. @garrettjonesgoogle, thoughts?

garrettjonesgoogle: I think it comes down to interchangeability. @andreamlin , do you know if there are any places where both are accepted? If not, it would make sense for them to be separate. Can we check the original proto2 to see what's happening there?

It should be noted that such a situation isn't possible with One Platform APIs because our original data source for config gen has stars (e.g. https://github.com/googleapis/googleapis/blob/master/google/pubsub/v1/pubsub.proto#L49 ), not bracketed variable names, so we guess at what the variable names should be based on the plural words. In this case, we'd only have 'projects/*/zones/*/disks/*' to go off of.

andreamlin: The discovery doc will explicitly ask for the bracketed word as a parameter, e.g. method [compute.disks.setLabels](https://raw.githubusercontent.com/googleapis/discovery-artifact-manager/master/discoveries/compute.v1.json), path  `{project}/zones/{zone}/disks/{resource}/setLabels`  lists parameters `project`, `zone`, and `resource`. Also, in proto2, these bracketed words are the exact names of fields in the request object. 

I understand that this introduces a redundancy compared to One Platform APIs, where the wildcard word doesn't matter, but I would rather have this duplication of semantically-equivalent resource names than the possible inaccuracy of assuming two different path templates are the same resource name class.

andreamlin: PTAL again?

PR #1912, opened 2018-04-03
Inflector.singularize() cases and tests

andreamlin:
Add more cases to the `singularize()` method and add an InflectorTest.java to check it.

Context: a compute.v1.json resource is named "licenses", which currently gets truncated to "licens" when singularizing it.

### DIFF
--- a/src/test/java/com/google/api/codegen/InflectorTest.java
+++ b/src/test/java/com/google/api/codegen/InflectorTest.java
@@ -1,0 +1,35 @@
+/* Copyright 2018 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


andreamlin:
PTAL @pongad

src/test/java/com/google/api/codegen/InflectorTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+  @Test
+  public void testSingularize() {
+    Truth.assertThat(Inflector.singularize("scarves")).isEqualTo("scarf");
+    Truth.assertThat(Inflector.singularize("halves")).isEqualTo("half");
+    Truth.assertThat(Inflector.singularize("cars")).isEqualTo("car");
+    Truth.assertThat(Inflector.singularize("fuzzes")).isEqualTo("fuzz");
+    Truth.assertThat(Inflector.singularize("sneezes")).isEqualTo("sneeze");
+    Truth.assertThat(Inflector.singularize("blesses")).isEqualTo("bless");
+    Truth.assertThat(Inflector.singularize("scarves")).isEqualTo("scarf");
```
pongad: Looks like this case is written twice (another on L24)?
For this kind of tests, I like sorting alphabetically but I won't push on that :P

andreamlin: Ah sorting alphabetically would have caught the dup! Done.

andreamlin:
made changes; PTAL

PR #1913, opened 2018-04-04
Set page_streaming.resources_object to non-null value

andreamlin:
Currently, if no resources_object is found, configgen renders a null value for `page_streaming.resources_object`. This results in an unparsable gapic config yaml that throws hard-to-diagnose error messages when it is used to generate a GAPIC. 

If an obvious "TODO" is given as an alternative for an unknown page_streaming.resources_object, it is much easier to diagnose a failed GAPIC gen on the concrete string "TODO".

Context: the newest compute.v1.json has one paging resource object named "resources", which didn't follow the pattern of it being "items".

### DIFF
--- a/src/main/java/com/google/api/codegen/config/DiscoveryField.java
+++ b/src/main/java/com/google/api/codegen/config/DiscoveryField.java
@@ -283,6 +283,9 @@ public class DiscoveryField implements FieldModel, TypeModel {
 
     Schema parentTypeSchema = getDiscoveryField();
     List<Schema> pathToKeySchema = parentTypeSchema.findChild(key);
+    if (pathToKeySchema.size() == 0) {
+      return null; // key not found.
+    }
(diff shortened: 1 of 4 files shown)


andreamlin:
PTAL

garrettjonesgoogle:
Should we also add in a helpful error message when TODO is encountered during codegen? (I'm not sure how it will fail with TODO)

andreamlin:
re: error message
The DiagCollector collects error messages, and will log "Resources field missing for page streaming..." when the resourcesField is not found.

andreamlin:
PTAL
some minor changes:
- just return null when resources field is not found in method output fields instead of throwing exception; this follows the convention of treating a null resource field as an error that is reported to the DiagCollector
- make `"_TODO_"` the non-null filler string, which is a little more visible as an invalid argument than `"TODO"`

PR #1914, opened 2018-04-05
Escape variable names in generated test cases

andreamlin:
Generated test cases may use a language-formatted version of the string `"resources"`. Message types may have a field of the same name. Add `"resources"` to the SymbolTable for the generated test case so the rendered field name doesn't clash with the hard-coded string in the test case.

An extra test API method was added to simplecompute.v1.json to test this, which resulted in changes in all those test resources.


src/test/java/com/google/api/codegen/testdata/discogapic/java/java_simplecompute.v1.json.baseline
```diff
@@ -12831,6 +13977,54 @@ public class DummyObjectClientTest {
     client.close();
   }
 
+  @Test
+  @SuppressWarnings("all")
+  public void listResourcesDummyObjectsTest() {
+    String kind = "kind3292052";
+    String nextPageToken = "";
+    Address resourcesElement = Address.newBuilder().build();
+    List<Address> resources2 = Arrays.asList(resourcesElement);
```
andreamlin: Without this change, the `resources2` variable name would have been rendered as `resources`, which clashes with the below `List<Address> resources`.

src/test/java/com/google/api/codegen/testdata/discogapic/java/java_simplecompute.v1.json.baseline
```diff
@@ -12831,6 +13977,54 @@ public class DummyObjectClientTest {
...
+      .setNextPageToken(nextPageToken)
+      .addAllResources(resources2)
+      .build();
+    mockService.addResponse(expectedResponse);
+
+    ProjectName project = ProjectName.of("[PROJECT]");
+
+    ListResourcesDummyObjectsPagedResponse pagedListResponse = client.listResourcesDummyObjects(project);
+
+    List<Address> resources = Lists.newArrayList(pagedListResponse.iterateAll());
```
andreamlin: There would have been a clash with this hardcoded `resources` field.

andreamlin:
PTAL

PR #1916, opened 2018-04-10
Adding instructions, renaming output artifact, better failure messages

garrettjonesgoogle:
Renaming the entire github repo will be a separate step.


garrettjonesgoogle:
cc @timburks Check out the new documentation :-)

garrettjonesgoogle:
@vchudnov-g PTAL. Note: there are a couple steps that aren't completely documented for all languages; I'd like to defer that for now.

README.md
```diff
@@ -1,32 +1,254 @@
...
+Run the following command to generate the proto descriptor file:
+
+```
+protoc -I=${PROTOC_INCLUDE_DIR} -I=${GOOGLEAPIS_DIR} -I=${YOUR_PROTO_DIR} \
+  --include_imports --include_source_info -o ${YOUR_DESCRIPTOR_FILE} ${YOUR_PROTO_FILE}
+```
+
+### Generate proto message classes
+
+(Skip this section for Node.js - it loads proto files into memory at runtime.)
```
vam-google: Please highlight this Node.js discrepancy somehow (very easy to miss, I actually missed during first read).

README.md
```diff
@@ -1,32 +1,254 @@
...
+
+1. The output directory to contain the proto files. Hereafter, this will be referenced as `${GENERATED_PROTO_DIR}`.
+2. The language you are generating for. The possible values for protoc
+`["java", "go", "php", "ruby", "python", "csharp"]`.
+Hereafter, this will be referenced as `${PROTO_OUTPUT_LANG}`.
+
+Run the following command to generate the proto message classes:
+
+```
+rm -rf ${GENERATED_PROTO_DIR}
```
vam-google: That is strange. Can't these just be 2 different dirst, instead of one dirname being reused twise for different content?

README.md
```diff
@@ -1,32 +1,254 @@
...
+   referenced as `${PACKAGE_NAME}`.
+5. The proto path, which is the proto's package with dots converted to slashes. For example, Cloud Language v1's
+   proto path is `google/cloud/language/v1`. Hereafter, this will be referenced as `${PROTO_PATH}`.
+
+First, copy the contents of api_defaults.yaml and dependencies.yaml into a single file, hereafter
+referenced as `${PKG_META_CONFIG}`.
+
+```
+cp ${GOOGLEAPIS_DIR}/gapic/packaging/api_defaults.yaml ${PKG_META_CONFIG}
+cat < ${GOOGLEAPIS_DIR}/gapic/packaging/dependencies.yaml >> ${PKG_META_CONFIG}
```
vam-google: Wow... it looks like we are really not ready to distribute gapic-gen as a separate tool

README.md
```diff
@@ -1,32 +1,254 @@
...
+
+First, you need to install protoc if you don't have it yet: see
+[protocol-compiler-installation](https://github.com/google/protobuf#protocol-compiler-installation).
+The pre-built binaries are the easiest, but there aren't really instructions for installation -
+you have to just download the correct binaries for your platform at put them on your path.
+If you are using Linux x86_64:
+
+1. Download the latest zip from https://github.com/google/protobuf/releases ending in "-linux-x86_64.zip"
+2. Unzip the file, e.g. `unzip protoc-3.5.1-linux-x86_64.zip -d protoc-3.5.1`
+3. Copy protoc to a location on $PATH, e.g. `cp protoc-3.5.1/bin/protoc /usr/local/bin`
```
vchudnov-g: Backtick `$PATH` for consistency with `${TOOLKIT_DIR}` below

README.md
```diff
@@ -1,32 +1,254 @@
...
+
+Clone the present repository and build API Client Generator:
+
+```
+git clone https://github.com/googleapis/toolkit.git
+cd toolkit
+./gradlew fatJar
+```
+
+The googleapis/toolkit directory will hereafter be referenced as `${TOOLKIT_DIR}`.
```
vchudnov-g: Backtick dirs?

README.md
```diff
@@ -1,32 +1,254 @@
...
+
+1. The language-specific generator config for your desired output language. Given a language where
+   `${language}` is in `["csharp", "go", "java", "nodejs", "php", "py", "ruby"]`, the file is located at
+   `src/main/resources/com/google/api/codegen/${language}/${language}_gapic.yaml` (exception: the Python file is named
+   `python_gapic.yaml`). Hereafter, this will be referenced as `${LANGUAGE_CONFIG}`.
+2. The output directory for your generated client classes. Hereafter, this will be
+   referenced as `${GENERATED_CLIENT_DIR}`.
+
+```
+java -cp build/libs/gapic-generator-*-fatjar.jar com.google.api.codegen.CodeGeneratorTool \
```
vam-google: I think we should considering adding `.sh/.bat` runner scripts around toolkit for this purpose. It seems like a industry standard of doing it for java apps (for example both gradlew and IntelliJ IDEA do it this way). Btw, this may allow us to avoid adding fatJar task in gradle (correct classpath can be handled in the sh script itself).

README.md
```diff
@@ -1,32 +1,254 @@
...
+java -cp build/libs/gapic-generator-*-fatjar.jar com.google.api.codegen.configgen.ConfigGeneratorTool \
+  --descriptor_set=${YOUR_DESCRIPTOR_FILE} --service_yaml=${YOUR_SERVICE_YAML} \
+  -o=${YOUR_CLIENT_CONFIG}
+```
+
+You can safely ignore the warning about control-presence.
+
+### Manually tweak the generated client config
+
+The generated client config contains FIXME comments with instructions on how to choose values
```
vchudnov-g: backtick `FIXME`? Or quote?

README.md
```diff
@@ -1,32 +1,254 @@
...
+in the client config. The client config should work as-is though; tweaks are only necessary to improve
```
vchudnov-g: s/as-is though/as is, though/

README.md
```diff
@@ -1,32 +1,254 @@
...
+
+(There is an open issue to fix this: https://github.com/googleapis/toolkit/issues/1918 )
+
+### Perform fix-ups to get a working library
+
+The fix-ups here are language-specific.
+
+#### Java
+
+Copy the proto-generated classes into the client directory tree:
```
vam-google: Please, please, please lets not do it (more details on https://github.com/googleapis/toolkit/pull/1916/files#diff-5f526d3c7a4f88ae57f1cba301706d19R38).

README.md
```diff
@@ -1,32 +1,254 @@
...
+
+### Generate proto message classes
+
+(Skip this section for Node.js - it loads proto files into memory at runtime.)
+
+You need to locate/decide on the following before you can generate the proto message classes:
+
+1. The output directory to contain the proto files. Hereafter, this will be referenced as `${GENERATED_PROTO_DIR}`.
+2. The language you are generating for. The possible values for protoc
+`["java", "go", "php", "ruby", "python", "csharp"]`.
```
vchudnov-g: I don't think we need array notation here.

"""
The possible values for protoc are `java`, `go`, `php`, `ruby`, `python`, and `csharp`.
"""

README.md
```diff
@@ -1,32 +1,254 @@
...
+package_name:
+  default: ${PACKAGE_NAME}
+proto_path: ${PROTO_PATH}
+```
+
+### Run code generation
+
+You need to locate/decide on the following before you call code generation:
+
+1. The language-specific generator config for your desired output language. Given a language where
```
vchudnov-g: """
Given a `${language}` (one of `java`, `go`, `php`, `ruby`, `python`, `nodejs`, or `csharp`), the file ...
"""
(Note also order consistent with previous listing)
(not also the or

README.md
```diff
@@ -1,32 +1,254 @@
...
+
+Special note for java: several files will be dumped into the parent directory of `${GENERATED_CLIENT_DIR}`:
+
+- gradle/
+- gradlew
+- gradlew.bat
+
+(There is an open issue to fix this: https://github.com/googleapis/toolkit/issues/1918 )
+
+### Perform fix-ups to get a working library
```
vchudnov-g: Here and below, I suggest:

s/fix-ups/fixes/
or
s/fix-ups/touch-ups/

README.md
```diff
@@ -1,32 +1,254 @@
...
+
+```
+cp -r ${GENERATED_PROTO_DIR}/* ${GENERATED_CLIENT_DIR}/src/main/java
+```
+
+In `${GENERATED_CLIENT_DIR}/build.gradle`, there are a couple dependencies that need to be removed if you
+are bundling your proto-generated classes and grpc stubs in the same package as the client. They have a
+comment starting with "`// Remove this line if you are bundling`". Warning: If you regenerate again, you will
+need to remove the lines from the build.gradle file again. (There is an open issue to fix this:
+https://github.com/googleapis/toolkit/issues/1917 ).
```
vchudnov-g: Is this auto-hyperlinked?

build.gradle
```diff
@@ -118,8 +118,13 @@ dependencies {
 }
 
 task fatJar(type: com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar) {
-  archiveName = "gapic.jar"
-  from sourceSets.main.output
+  classifier = 'fatjar'
+  from sourceSets.main.output.filter {
+    // This jar file needs to be excluded or else the shadow plugin will actually explode the classes into
+    // the fat jar. We want to retain the file as a unit because we want to put it into the output directory.
+    // StaticResourcesHandler peforms the rename back to `gradle-wrapper.jar`.
```
vchudnov-g: "peforms"??

vchudnov-g: "peforms"??

src/main/java/com/google/api/codegen/util/StaticResourcesHandler.java
```diff
@@ -36,7 +37,16 @@ public StaticResourcesHandler(Map<String, String> resourceFilesMap) {
     ImmutableMap.Builder<String, byte[]> resources = ImmutableMap.builder();
     for (Map.Entry<String, String> entry : resourceFilesMap.entrySet()) {
       String resourcePath = SnippetSetRunner.SNIPPET_RESOURCE_ROOT + '/' + entry.getKey();
-      byte[] resource = ByteStreams.toByteArray(cl.getResourceAsStream(resourcePath));
+      InputStream resourceStream = cl.getResourceAsStream(resourcePath);
+      // Hack necessary to allow jar files to be included in the fat jar
+      if (resourceStream == null && resourcePath.endsWith(".jar")) {
```
vchudnov-g: Where do you do the renaming back to `.jar` that you mentioned above?

(I'm probably missing something since I don't have the full context of what's going in here.)

build.gradle
```diff
@@ -118,8 +118,13 @@ dependencies {
 }
 
 task fatJar(type: com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar) {
-  archiveName = "gapic.jar"
-  from sourceSets.main.output
+  classifier = 'fatjar'
+  from sourceSets.main.output.filter {
+    // This jar file needs to be excluded or else the shadow plugin will actually explode the classes into
```
vchudnov-g: I don't have full context here, but this reads a bit confusing to me. We are excluding the file but want to include it (as a jar) in the output directory. Is it excluded entirely, or just marked for not being exploded into the fat jar?

README.md
```diff
@@ -1,32 +1,254 @@
-# Introduction
+# GAPIC Generator
 
-Google API tools consist of two tools:
+GAPIC (Generated API Client) Generator is a client library code generator which auto-generates
```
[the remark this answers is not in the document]
garrettjonesgoogle: Will do.

README.md
```diff
@@ -1,32 +1,254 @@
...
+clients for both grpc-based APIs and Discovery-document-based APIs. Currently, the following
```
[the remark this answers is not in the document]
garrettjonesgoogle: This sounds odd to me...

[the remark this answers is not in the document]
garrettjonesgoogle: Sure, I will add a note.

README.md
```diff
@@ -1,32 +1,254 @@
...
+   [artman]( https://github.com/googleapis/artman/#usage). This takes care of several of the steps
+   below and is not documented here.
 
-1. Invoking directly through a gradle task (see the build.gradle:runCodeGen task).
-2. Invoking a code generation pipeline through
-   [artman]( https://github.com/googleapis/artman/#usage).
+### Prerequisites for code generation
 
-## The Google API Code Synchronizer
+1. You need a proto file describing your API. Proto files of existing Google APIs are available at
```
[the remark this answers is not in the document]
garrettjonesgoogle: Do you mean the common proto stuff built from the directories listed in https://github.com/googleapis/googleapis/blob/master/gapic/packaging/common_protos.yaml ? If so, yes, the output is coupled to that - well actually, coupled to the language-specific package built from that, which should automatically be part of the dependency list of the generated output for each language. Of interest, @lukesneeringer and I have reached an agreement to move these protos to their own repository.

README.md
```diff
@@ -1,32 +1,254 @@
...
+[googleapis](https://github.com/googleapis/googleapis).
+  - Example: https://github.com/googleapis/googleapis/blob/master/google/cloud/language/v1/language_service.proto
+2. You need a yaml file describing certain service values. This is unfortunately not
```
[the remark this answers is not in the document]
garrettjonesgoogle: Sure.

README.md
```diff
@@ -1,32 +1,254 @@
...
+well-documented.
+  - Example: https://github.com/googleapis/googleapis/blob/master/google/cloud/language/language_v1.yaml
```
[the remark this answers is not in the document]
garrettjonesgoogle: Fair, I'll move it.

README.md
```diff
@@ -1,32 +1,254 @@
...
+3. Java 8 JDK
```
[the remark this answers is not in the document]
garrettjonesgoogle: I guess it wouldn't harm anyone to mention it :-)

README.md
```diff
@@ -1,32 +1,254 @@
...
+### Process overview for generating a client library
 
-1. Invoking directly through a gradle task (see the build.gradle:runSynchronizer
-   task).
-2. Invoking a code generation pipeline which contains synchronization step
-   through [artman](https://github.com/googleapis/artman/#usage).
+1. Generate a descriptor file from the proto (once per API)
+2. Run client config generation (once per API)
+3. Manually tweak the generated client config
+4. Run code generation (once per API + language)
```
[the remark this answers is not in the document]
garrettjonesgoogle: Sounds good.

README.md
```diff
@@ -1,32 +1,254 @@
...
+
+Each of these steps are described in more detail below.
+
+### Set up prerequisites for building from source and running directly
+
+First, you need to install protoc if you don't have it yet: see
```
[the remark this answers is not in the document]
garrettjonesgoogle: Sounds like a good practice, I'll do that.

README.md
```diff
@@ -1,32 +1,254 @@
...
+[protocol-compiler-installation](https://github.com/google/protobuf#protocol-compiler-installation).
+The pre-built binaries are the easiest, but there aren't really instructions for installation -
```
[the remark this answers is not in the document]
garrettjonesgoogle: I will reword it so it sounds less reproach-y.

README.md
```diff
@@ -1,32 +1,254 @@
...
+you have to just download the correct binaries for your platform at put them on your path.
```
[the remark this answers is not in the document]
garrettjonesgoogle: Will do.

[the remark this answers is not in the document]
garrettjonesgoogle: oops!

README.md
```diff
@@ -1,32 +1,254 @@
...
+If you are using Linux x86_64:
+
+1. Download the latest zip from https://github.com/google/protobuf/releases ending in "-linux-x86_64.zip"
```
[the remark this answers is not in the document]
garrettjonesgoogle: Note: we don't require a specific version of protobuf. Fairly old versions can still be used, actually. I don't want a direct link because I don't want to imply that's the only platform that will work, and I also want people to prefer newer versions when protobuf releases them. This readme isn't specific to linux - it works on Mac too (in fact, so far I have only verified on Mac). 

That's a fair point about supported platforms - I'll add a note.

README.md
```diff
@@ -1,32 +1,254 @@
...
+2. Unzip the file, e.g. `unzip protoc-3.5.1-linux-x86_64.zip -d protoc-3.5.1`
```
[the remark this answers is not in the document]
garrettjonesgoogle: I'm using 'e.g.' because the command would be different if the user downloaded a different version or a different platform.

README.md
garrettjonesgoogle: Will do.

README.md
garrettjonesgoogle: Will do.

README.md
```diff
@@ -1,32 +1,254 @@
...
+git clone https://github.com/googleapis/googleapis
+```
+
+The googleapis/googleapis directory will hereafter be referenced as `${GOOGLEAPIS_DIR}`.
+
+### Generate a descriptor file from the proto
+
+You need to locate/decide on the following before you can generate the descriptor file:
+
+1. The include directory with the proto files bundled with protoc (from the protoc setup step).
```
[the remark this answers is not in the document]
garrettjonesgoogle: But if I still keep the variable-like strings, I still need to define them...

I like the current ordering because it can just be followed in a sequence. This isn't quite a man page, it is a tutorial.

README.md
```diff
@@ -1,32 +1,254 @@
...
+   Hereafter, this will be referenced as `${PROTOC_INCLUDE_DIR}`.
+2. Any directories containing protos that your API depends on. For Google APIs defined in
+   [googleapis](https://github.com/googleapis/googleapis), this will be `${GOOGLEAPIS_DIR}`.
+3. The directory containing your proto. Hereafter, this will be referenced as `${YOUR_PROTO_DIR}`.
+4. Your proto file. Hereafter, this will be referenced as `${YOUR_PROTO_FILE}`.
```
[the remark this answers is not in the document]
garrettjonesgoogle: I will add "(or files)".

README.md
garrettjonesgoogle: Ok I will italicize it, and also add a reminder in step `#2`.

README.md
garrettjonesgoogle: Ok.

README.md
garrettjonesgoogle: The current instructions are designed to dump it all together. When we restructure the output to create three modules, we can change this.

README.md
garrettjonesgoogle: I will backtick it.

README.md
garrettjonesgoogle: Will do.

README.md
garrettjonesgoogle: Nope! That is actually the key point of this whole exercise, to bring the mess out in the clear, so that we know what to clean up.

README.md
garrettjonesgoogle: Will do, except that `python` needs to be `py`.

README.md
garrettjonesgoogle: It's considerably easier to distribute a fat jar - it's actually somewhat difficult to grab all of the jars out of the gradle cache. 

I'm fine with adding `.sh`/`.bat` files later, but I don't want to tackle it at the moment.

README.md
garrettjonesgoogle: I will use "fixes".

README.md
garrettjonesgoogle: As I responded in that comment - I agree, but we need to restructure the output of toolkit before we can do that. I'm just trying to document what you have to do given the current state of toolkit.

README.md
garrettjonesgoogle: Yes.

build.gradle
garrettjonesgoogle: We are renaming it so that the shadow plugin doesn't extract all of the classes from it such that they live side-by-side with the toolkit classes.

build.gradle
garrettjonesgoogle: oops, I will fix that

src/main/java/com/google/api/codegen/util/StaticResourcesHandler.java
garrettjonesgoogle: `entry.getValue()` contains the file path, which will be `.../gradle-wrapper.jar`. The byte content of the .zjar file will end up in a .jar file.

garrettjonesgoogle:
PR feedback addressed, PTAL

README.md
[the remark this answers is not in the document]
vam-google: PHP requires at least 3.5.1 I think (because of various bugs in PHP support in protobuf)

README.md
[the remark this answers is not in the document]
vam-google: Maybe, but the string "Hereafter, this will be referenced" has 15 exact matches in the readme... I would still prefer the reversed order. Even in the way people read this kind of "tutorials" they just want some command to copy paste and run, see if it fails and if it fails fix it and run again. Nobody really wants to understand details, just run the thing (so we show the command first and provide explanation only for those who want explanation, if the placeholder is descriptive enough one would not even need to read the explanation).

README.md
vam-google: But we remove the directory, how can that mean dumping all together? They are together "geographically" (same folder) but not historically (one dies before the other is born). Am I missing something here?

README.md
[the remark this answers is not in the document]
garrettjonesgoogle: Oh ok. Well anyway, the instructions are correct for that (since they say "latest zip").

README.md
garrettjonesgoogle: I will remove the `rm -rf` line because it's confusing. The sequence of instructions did work because this is the first line, then the proto generation, and finally the grpc generation (which puts the files in the same place as the proto generation).

README.md
```diff
@@ -1,32 +1,279 @@
...
+2. Run client config generation (once per API)
+3. Manually tweak the generated client config
+4. Run code generation (once per API x language combination of interest)
+
+Each of these steps are described in more detail below. Note: the instructions assume you are running on a
+Unix-y system; if you are using Windows, you will need to tweak the steps for yourself.
+
+### Set up prerequisites for building from source and running directly
+
+Firstly, make sure you have the
```
vchudnov-g: s/Firstly/First/

README.md
```diff
@@ -1,32 +1,279 @@
...
+[Java 8 JDK](http://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html) installed.
+
+Next, you need to install `protoc` if you don't have it yet: see
+[protocol-compiler-installation](https://github.com/google/protobuf#protocol-compiler-installation).
+The pre-built binaries are the easiest. There aren't instructions for installation;
+you just have to download the correct binaries for your platform and put them on your path.
+Given a platform ${PROTOC_PLATFORM} and a protoc version ${PROTOC_VERSION}:
+
+1. Download the latest zip from https://github.com/google/protobuf/releases ending in "-${PROTOC_PLATFORM}.zip"
```
vchudnov-g: `the latest zip file ending in "..." from https://....`

README.md
```diff
@@ -1,32 +1,279 @@
...
+Run the following command to generate the proto descriptor file:
+
+```
+protoc -I=${PROTOC_INCLUDE_DIR} -I=${GOOGLEAPIS_DIR} -I=${YOUR_PROTO_DIR} \
+  --include_imports --include_source_info -o ${YOUR_DESCRIPTOR_FILE} ${YOUR_PROTO_FILE}
+```
+
+### Generate proto message classes
+
+*(Skip this section for Node.js - it loads proto files into memory at runtime.)*
```
vchudnov-g: s/ - /— /

README.md
garrettjonesgoogle: Will do.

README.md
garrettjonesgoogle: Will do.

README.md
garrettjonesgoogle: But... will anyone care or notice besides you?

vchudnov-g:
The point of my suggestion is not to provide automation in here (that's why I explicitly said it need not be polished nor robust). The point is to provide a more reliable place for the documentation via executable code. That way it's easier to notice when the documentation is or isn't stale. With the doc in a separate doc like you have here, it's very easy to forget to update it and then not notice until you go through the steps manually. Essentially, I'm suggesting condensing, not expanding, what you have written out here into a couple of files and adding #! lines. The fact that there are so many steps would continue to convey the current invocation complexity.

I'm OK with submitting this as is, but I strongly urge you to consider migrating this doc to executable scripts as suggested :)

PR #1920, opened 2018-04-11
avoid regex; it's quadratic time

pongad:


pongad:
@michaelbausor I don't completely follow. The regex doesn't exist anymore, so there's nothing to test right?

michaelbausor:
The regex has been replaced with new logic, so it would be good for that to be tested somehow. I meant to say that I think that logic will be tested sufficiently just by the existing tests of parsePartialDottedPathToInitCodeNode, but I wanted to confirm that that was your thinking as well.

pongad:
I see what you're saying. Yes I think so.

The regex test confirms that you parse correctly right-to-left. This is already tested by `testNestedMixedField`.

PR #1922, opened 2018-04-11
Ruby: allow for namespaces version module

landrito:


src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceTransformer.java
```diff
@@ -369,15 +373,21 @@ private ViewModel generateTopLevelIndexView(ApiModel model, GapicProductConfig p
...
             metadataTransformer
                 .generateReadmeMetadataView(model, productConfig, packageNamer)
                 .moduleName(moduleName)
                 .build());
-      } else if (includeVersionModule || !moduleName.equals(namer.getModuleVersionName())) {
+      } else if (includeVersionModule || !isVersion) {
         moduleViews.add(SimpleModuleView.newBuilder().moduleName(moduleName).build());
       }
+
+      if (!includeVersionModule && isVersion) {
```
jbolinger: I probably should have done this, but would you mind adding a comment around here or at the method level so we don't forget why we break here? Something to the effect of if `includedVersionModule` is `false` then the list of returned modules will not include the version or anything after it.

src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceTransformer.java
```diff
@@ -399,6 +409,36 @@ private String topLevelPackagePath(SurfaceNamer namer) {
...
+    boolean versionFound = false;
+    for (String part : parts) {
+      if (versionFound) {
+        paths.add(namer.packageFilePathPiece(Name.upperCamel(part)));
+      }
+      if (VersionMatcher.isVersion(part)) {
+        versionFound = true;
+      }
+    }
+    return Joiner.on(File.separator).join(paths);
```
jbolinger: Hmm, this is mostly unrelated.. but does it make sense to use `File.separator`? There is no guarantee that the generator is going to be run on the same platform as the actual code and (I don't work on Windows much) but I suspect the paths are still specified with `/`.

src/main/java/com/google/api/codegen/viewmodel/metadata/VersionIndexView.java
```diff
@@ -63,8 +64,17 @@ public String resourceRoot() {
 
   public abstract VersionIndexType type();
 
+  /* The path leading up to the version directory. */
```
jbolinger: 👍 Thank you! We don't have nearly enough of these!

src/main/resources/com/google/api/codegen/ruby/version_index.snip
```diff
@@ -164,15 +164,27 @@
...
     @else
-      {@requireView.topLevelNamespace}.const_get(version_module).new(*args, **kwargs)
+      {@requireView.topLevelNamespace}.const_get(version_module){@postVersionNamespace(requireView)}.new(*args, **kwargs)
     @end
   end
 @end
+
+@private postVersionDirPath(index)
+  @if index.hasPostVersionDirPath
+    /{@index.postVersionDirPath}
```
jbolinger: ah, this is what I was referring to with the `Files.seperator` thing.

src/test/java/com/google/api/codegen/testdata/nodejs/nodejs_multiple_services.baseline
```diff
@@ -79,96 +79,7 @@ $ npm install --save @google-cloud/multiple-services
...
- * @property {constructor} DecrementerServiceClient
- *   Reference to {@link v1.DecrementerServiceClient}
- */
-module.exports.v1 = gapic.v1;
-
-// Alias `module.exports` as `module.exports.default`, for future-proofing.
-module.exports.default = Object.assign({}, module.exports);
-
-============== file: src/v1/decrementer_service_client.js ==============
+============== file: src/decrementer_service_client.js ==============
```
jbolinger: Is this correct (i.e. moving to different directory without the version)?

src/test/java/com/google/api/codegen/testdata/nodejs/nodejs_multiple_services.baseline
```diff
@@ -383,10 +294,10 @@ class DecrementerServiceClient {
 
 module.exports = DecrementerServiceClient;
 
-============== file: src/v1/decrementer_service_client_config.json ==============
+============== file: src/decrementer_service_client_config.json ==============
```
jbolinger: Is this correct (same as previous comment)?

src/test/java/com/google/api/codegen/testdata/nodejs/nodejs_multiple_services.baseline
landrito: Ah I overlooked this change. Let me try and figure this out!

jbolinger:
Hey @landrito any news on this one? Should we ping Alex to help? Looks like the only thing missing is the potential node updates?

PR #1923, opened 2018-04-11
C# support for common resource names

chrisdunelm:
Java generation is not correct.
@garrettjonesgoogle I'll look at the Java generation tomorrow, but if you get a change to have a look at this PR that'd be helpful. Or can point someone else at it.
@jskeet The C# generated code should be correct. Please feel free to review if you have a few minutes.

### DIFF
--- a/src/test/java/com/google/api/codegen/testdata/csharp/csharp_library.baseline
+++ b/src/test/java/com/google/api/codegen/testdata/csharp/csharp_library.baseline
@@ -119,6 +119,7 @@ namespace Google.Example.Library.V1.Snippets
 {
     using Google.Api.Gax;
     using Google.Api.Gax.Grpc;
+    using Google.Api.Gax.ResourceNames;
     using apis = Google.Example.Library.V1;
     using Google.LongRunning;
     using Google.Protobuf;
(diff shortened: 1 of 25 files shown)


src/test/java/com/google/api/codegen/testdata/csharp/csharp_library.baseline
```diff
@@ -5237,6 +5271,7 @@ namespace Google.Example.Library.V1.Tests
 
 using gax = Google.Api.Gax;
 using gaxgrpc = Google.Api.Gax.Grpc;
+using gaxres = Google.Api.Gax.ResourceNames;
```
jskeet: Is this hard-coded somewhere? I'm just trying to work out what happens for the Bigtable scenario, where we might have Google.Api.Gax.ResourceNames.ProjectName and also Google.Cloud.Bigtable.CommonResources.V1.TableName. (I'm not sure what we're going to do about versioning for common names, admittedly. That could be fun.)

chrisdunelm: Yes, this specific namespace is hardcoded to the `gaxres` alias.
In the example you give, `Google.Cloud.Bigtable.CommonResources.V1` will be imported using a standard (non-aliasing) `using` directive.
We don't yet have any use-cases where aliases of arbitrary namespaces are required. We should probably discuss the best way to handle these, but I don't think that needs to be part of this PR.

jskeet: Okay, I'm fine with that, so long as there's a plan :)

chrisdunelm:
Java common resource name generation disabled, so this PR no longer affects Java except due to the new test proto fields.
@jskeet This is now ready for review. Thanks :)

PR #1926, opened 2018-04-12
C# Fully qualify resource-name types in resource partials

chrisdunelm:
To avoid name collisions between class name and property name

### DIFF
--- a/src/test/java/com/google/api/codegen/testdata/csharp/csharp_library.baseline
+++ b/src/test/java/com/google/api/codegen/testdata/csharp/csharp_library.baseline
@@ -12763,11 +12763,11 @@ namespace Google.Example.Library.V1
     public partial class Book
     {
         /// <summary>
-        /// <see cref="BookName"/>-typed view over the <see cref="Name"/> resource name property.
+        /// <see cref="Google.Example.Library.V1.BookName"/>-typed view over the <see cref="Name"/> resource name property.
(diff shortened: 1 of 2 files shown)


chrisdunelm:
This reverts the qualification behaviour within the resource-name message partials to as it was before the main aliasing change.

src/test/java/com/google/api/codegen/testdata/csharp/csharp_library.baseline
```diff
@@ -12763,11 +12763,11 @@ namespace Google.Example.Library.V1
     public partial class Book
     {
         /// <summary>
-        /// <see cref="BookName"/>-typed view over the <see cref="Name"/> resource name property.
+        /// <see cref="Google.Example.Library.V1.BookName"/>-typed view over the <see cref="Name"/> resource name property.
         /// </summary>
-        public BookName BookName
+        public Google.Example.Library.V1.BookName BookName
```
jskeet: This part doesn't need to be fully-qualified. It can only be a type name, so that's how it's looked up.
It doesn't do any harm though.

chrisdunelm: Yes, some of these fully-qualified names could be simplified; however that would entail a much larger change, so I'll keep it like this for now.

PR #1931, opened 2018-04-13
java: implement common resource name

pongad:


src/main/java/com/google/api/codegen/transformer/java/JavaModelTypeNameConverter.java
```diff
@@ -223,8 +223,18 @@ public TypeName getTypeName(ProtoElement elem) {
 
   private TypeName getTypeNameForTypedResourceName(
       ResourceNameConfig resourceNameConfig, TypeRef type, String typedResourceShortName) {
```
pongad: The short name isn't short anymore since the common resource name is fully-qualified. How does C# deal with this? (I looked through the C# commit but got confused :( )

chrisdunelm: I constantly struggle with this code, but I *think* C# deals with this by calling [`typeNameConverter.getAndSaveNicknameFor(...)`](https://github.com/googleapis/gapic-generator/blob/master/src/main/java/com/google/api/codegen/transformer/csharp/CSharpModelTypeNameConverter.java#L272), which handles both shortnames and fully-qualified names.

PR #1933, opened 2018-04-13
Make the generation instructions runnable.

vchudnov-g:
This allows us to easily run the documentation we are presenting to
users, thus verifying that it is still accurate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ More platforms are likely supported when building protoc from C++ source.
 
 There are two main ways to generate clients:
 
-1. Build from source and run directly. This is detailed below.
+1. Build from source and run directly. See below.
 2. Invoke a code generation pipeline through
    [artman]( https://github.com/googleapis/artman/#usage). This takes care of several of the steps
    below and is not documented here.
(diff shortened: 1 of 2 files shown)


vchudnov-g:
This is what I was referring to earlier in terms of runnable documentation. WDYT?

vchudnov-g:
Fair enough, but the infrastructure for the script is set up, so I don't see that this is any harder than maintaining the README file. And this brings the benefits of making the documentation executable.

When we make changes, all that's needed is to add variables and steps following the pattern already established. And more likely than adding is that is that we'll be removing steps (which is the whole goal of simplifying).

garrettjonesgoogle:
Others should weigh in here too. @lukesneeringer @timburks @vam-google

vchudnov-g:
All good points, @vam-google . Your script looks pretty sophisticated, and I agree artman is how we usually publish.

My points, however, are that:

1. this script/documentation is _not_ sophisticated. That's on purpose.
2. we're putting the commands in the README currently, so why not instead put them in a place where they can  also be executed? I claim going forward it's the same amount of work to do one or the other
3. the script is mostly as readable as the README. Admittedly. in the current form you have to get past the set-up section at the beginning of the script, but if that's really a problem we can restructure things so the command list comes first and the user can see that immediately. 
3. at any given point in time, checking that the instructions are correct is strictly easier with the script,  because we only have set the environment variables and then run the script, rather than setting the variables and the following along in a doc as we copy and paste each step. We're much more likely to do the former than the latter, I claim.

vchudnov-g:
PS. You can download the script from this PR and then run it with no arguments  to see the pure documentation output it produces,

alexander-fenster:
@vchudnov-g Do you think this year-old PR is still actual? I'm trying to close old PRs in gapic-generator, are you OK with closing this one or do you plan to do anything here?

PR #1935, opened 2018-04-13
dynamic: move LRO decision into generator code

pongad:
This commit moves the decision of whether to generate LRO code
into the generator and out of the snippet.
This makes the dynamic langs more similar to static langs,
and helps with sample code generation.

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/DynamicLangApiMethodTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/DynamicLangApiMethodTransformer.java
@@ -71,6 +71,9 @@ public class DynamicLangApiMethodTransformer {
       apiMethod.pageStreamingView(
           pageStreamingTransformer.generateDescriptor(
               context.getSurfaceInterfaceContext(), method));
+    } else if (context.getMethodConfig().isLongRunningOperation()) {
+      apiMethod.longRunningView(lroTransformer.generateDetailView(context));
+      apiMethod.type(ClientMethodType.OperationOptionalArrayMethod);
     } else {
       apiMethod.type(ClientMethodType.OptionalArrayMethod);
     }
@@ -138,10 +141,6 @@ public class DynamicLangApiMethodTransformer {
     apiMethod.packageHasMultipleServices(packageHasMultipleServices);
     apiMethod.packageServiceName(namer.getPackageServiceName(context.getInterfaceConfig()));
     apiMethod.apiVersion(namer.getApiWrapperModuleVersion());
-    apiMethod.longRunningView(
-        context.getMethodConfig().isLongRunningOperation()
-            ? lroTransformer.generateDetailView(context)
-            : null);
 
     apiMethod.oneofParams(context.getMethodConfig().getOneofNames(namer));
     apiMethod.headerRequestParams(
--- a/src/main/java/com/google/api/codegen/transformer/DynamicLangApiMethodTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/DynamicLangApiMethodTransformer.java
@@ -73,7 +73,7 @@ public class DynamicLangApiMethodTransformer {
               context.getSurfaceInterfaceContext(), method));
     } else if (context.getMethodConfig().isLongRunningOperation()) {
       apiMethod.longRunningView(lroTransformer.generateDetailView(context));
-      apiMethod.type(ClientMethodType.OperationOptionalArrayMethod);
+      apiMethod.type(ClientMethodType.LongRunningOptionalArrayMethod);
     } else {
       apiMethod.type(ClientMethodType.OptionalArrayMethod);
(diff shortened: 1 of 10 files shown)


src/main/java/com/google/api/codegen/transformer/DynamicLangApiMethodTransformer.java
```diff
@@ -71,6 +71,9 @@ public OptionalArrayMethodView generateMethod(
       apiMethod.pageStreamingView(
           pageStreamingTransformer.generateDescriptor(
               context.getSurfaceInterfaceContext(), method));
+    } else if (context.getMethodConfig().isLongRunningOperation()) {
+      apiMethod.longRunningView(lroTransformer.generateDetailView(context));
+      apiMethod.type(ClientMethodType.OperationOptionalArrayMethod);
```
vchudnov-g: One thing I was chatting to Michael B about, and which is independent of this but may diminish confusion:

At the moment, some of  the client method type enum values seem to be used in the same way (AFAICT) in different languages even though they we use distinct values. It might be desirable to use the same values across languages when they have the same semantics.

This would be analogous to, but independent of, using the same calling form enum value (once we create the enum type) across different languages, where applicable.

src/main/resources/com/google/api/codegen/nodejs/method_sample.snip
```diff
@@ -42,6 +35,8 @@
     @switch apiMethod.type.toString
     @case "OptionalArrayMethod"
         {@optionalArrayMethodSampleCode(apiMethod, additionalCallback)}
+    @case "OperationOptionalArrayMethod"
```
vchudnov-g: nit "Operation" sounds rather generic. Maybe we can s/Operation/LongRunning/

src/main/java/com/google/api/codegen/viewmodel/ClientMethodType.java
```diff
@@ -32,7 +32,7 @@
   // PHP
   OptionalArrayMethod,
   PagedOptionalArrayMethod,
-  OperationOptionalArrayMethod,
+  LongRunningOptionalArrayMethod,
```
vchudnov-g: And this is only used in PHP and Python, not in the other dynamic languages?

I would be in favor of changing the comment so that this line ends with `// Python, PHP` to indicate that.

src/main/resources/com/google/api/codegen/nodejs/method_sample.snip
```diff
@@ -25,14 +25,7 @@
...
 @snippet sampleCode(apiMethod)
-    @switch apiMethod.type.toString
-    @case "OptionalArrayMethod"
-        {@optionalArrayMethodSampleCode(apiMethod, empty())}
-    @case "PagedOptionalArrayMethod"
-        {@pagedOptionalArrayMethodSampleCode(apiMethod, empty())}
-    @default
-        $unhandledCase: {@apiMethod.type.toString}$
-    @end
+    {@sampleCode(apiMethod, empty())}
```
vchudnov-g: Thanks for removing the duplicated code!

src/main/resources/com/google/api/codegen/nodejs/method_sample.snip
```diff
@@ -216,24 +211,24 @@
...
-        var response = responses[0];
-        @if additionalCallback
-          console.log(response);
-        @else
-          // doThingsWith(response)
-        @end
-      })
-      {@promiseCallbacks(additionalCallback)}
-  @end
+@private operationSampleCode(apiMethod, additionalCallback)
```
vchudnov-g: What do you think about a more descriptive name, either `longRunningOperationSampleCode` or `lroSampleCode`?

src/main/resources/com/google/api/codegen/py/method_sample.snip
```diff
@@ -165,22 +167,24 @@
...
-
-      @# Handle metadata.
-      metadata = response.metadata()
-    @end
   @else
     {@methodCallSampleCode(apiMethod, initCode)}
   @end
 @end
 
+@private operationSampleCode(apiMethod, initCode)
```
vchudnov-g: More descriptive name, as above?

src/main/java/com/google/api/codegen/viewmodel/ClientMethodType.java
pongad: Done

src/main/resources/com/google/api/codegen/nodejs/method_sample.snip
pongad: Done

src/main/resources/com/google/api/codegen/py/method_sample.snip
pongad: Done

PR #1936, opened 2018-04-13
Allow nullable fields in PackageMetadataConfig

andreamlin:
Discogapic doesn't need package configs for protoc, grpc, etc, so allow those fields to be null.


andreamlin:
PTAL

PR #1937, opened 2018-04-14
Add discogapic generation of build.gradle

andreamlin:
Add `gax-httpjson` to PackageMetadataConfig.
Update build.gradle snippet with `gax-httpjson` and `gax:testlib`.

### DIFF
--- a/src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
+++ b/src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
@@ -42,24 +42,18 @@ public abstract class PackageMetadataConfig {
 
   protected abstract Map<TargetLanguage, VersionBound> gaxVersionBound();
 
-  @Nullable
   protected abstract Map<TargetLanguage, VersionBound> gaxGrpcVersionBound();
 
-  @Nullable
   protected abstract Map<TargetLanguage, VersionBound> gaxHttpVersionBound();
 
-  @Nullable
   protected abstract Map<TargetLanguage, VersionBound> grpcVersionBound();
 
-  @Nullable
   protected abstract Map<TargetLanguage, VersionBound> protoVersionBound();
 
   protected abstract Map<TargetLanguage, VersionBound> apiCommonVersionBound();
 
-  @Nullable
   protected abstract Map<TargetLanguage, VersionBound> authVersionBound();
 
-  @Nullable
   protected abstract Map<TargetLanguage, VersionBound> generatedNonGAPackageVersionBound();
 
   @Nullable
@@ -67,7 +61,6 @@ public abstract class PackageMetadataConfig {
 
   protected abstract Map<TargetLanguage, String> packageName();
 
-  @Nullable
   protected abstract Map<TargetLanguage, Map<String, VersionBound>> protoPackageDependencies();
 
   @Nullable
--- a/src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
+++ b/src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
@@ -45,20 +45,26 @@ public abstract class PackageMetadataConfig {
 
   protected abstract Map<TargetLanguage, VersionBound> gaxGrpcVersionBound();
 
+  @Nullable
   protected abstract Map<TargetLanguage, VersionBound> grpcVersionBound();
 
+  @Nullable
   protected abstract Map<TargetLanguage, VersionBound> protoVersionBound();
 
   protected abstract Map<TargetLanguage, VersionBound> apiCommonVersionBound();
 
+  @Nullable
   protected abstract Map<TargetLanguage, VersionBound> authVersionBound();
 
+  @Nullable
   protected abstract Map<TargetLanguage, VersionBound> generatedNonGAPackageVersionBound();
 
+  @Nullable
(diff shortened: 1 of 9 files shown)


andreamlin:
PTAL

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
```diff
@@ -307,7 +320,9 @@ public static PackageMetadataConfig createFromString(String yamlContents) {
       Map<String, Object> configMap, String key) {
     Map<TargetLanguage, Map<String, VersionBound>> packageDependencies = new HashMap<>();
     List<String> packages = (List<String>) configMap.get(key);
-    Preconditions.checkArgument(packages != null, key + " is missing from package metadata config");
+    if (packages == null) {
+      return packageDependencies;
```
vam-google: Why is null an acceptable value now? Looks weird. If in discogapic flow it can be null, it is better just to provide an empty collection where appropriate (in the place which writes into `configMap`) instead. This is minor, but is still weird.

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
```diff
@@ -381,6 +396,10 @@ public ReleaseLevel apply(@Nullable String releaseLevelName) {
    */
   private static <V> Map<TargetLanguage, V> buildMapWithDefault(Map<String, V> inputMap) {
     Map<TargetLanguage, V> outputMap = new HashMap<>();
+    if (inputMap == null) {
```
vam-google: Please see the previous comment, same applies here.

src/main/resources/com/google/api/codegen/java/build_gapic.gradle.snip
```diff
@@ -19,11 +19,22 @@
 
   dependencies {
     compile 'com.google.api:gax:{@metadata.gaxVersionBound.lower}'
-    compile 'com.google.api:gax-grpc:{@metadata.gaxGrpcVersionBound.lower}'
+    testCompile 'com.google.api:gax:{@metadata.gaxVersionBound.lower}:testlib'
```
vam-google: This is surprising. It seems that discogapic feature tries to go in parallel with gapic generation (basically having its own hierarcy of classes with unfortunately significant amount of code duplication between gapic and discogapic). For example discogapic has its own `MainDiscoGapicProviderFactory`,  but here it is trying to reuse `guilb_gapic.gradle.snip`. This is somewhat subjective, but I think it should be another way around: we try to reuse classes if we can, if we can reuse classes then we try to reuse related templates (.snip files). If we failed to reuse classes, then probably we should not even try to reuse snip files (since discogapic and gapic are already in "parallel worlds").

@garrettjonesgoogle what do you think about it?

andreamlin: The snippet files are explicitly tied to ViewModel classes.  I don't think it helps to extend that coupling of snippet files with generator classes all the way up with ProviderFactory classes.

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
andreamlin: Discogapic is missing a lot of config values (e.g. `gax_grpc_version`, `proto_deps`), and this would require many lines of `if(value == null)` checks and null handling.

The place that creates the `configMap` is in a 3rd party `org.yaml` package.

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
andreamlin: ditto :)

andreamlin:
Responded to comments. PTAL

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
vam-google: In case if it is  gapic execution path (not discogapic) this would make `null` a valid value (used to be invalid). I understand that it is minor, but if it is possible to keep it invalid in case of gapic (not discogapic) then please make it so (but only if it is simple to do and does not complicate or "uglify" the code significantly).

src/main/resources/com/google/api/codegen/java/build_gapic.gradle.snip
garrettjonesgoogle: +1  to @andreamlin - snippet files should be coupled to view model classes and not transformer classes. Where possible, it would be nice to have fewer if checks in snippets, which sometimes might mean making view models more abstract sometimes. For example, here we could put the name of the transport-specific gax dependency in the model, knowing that we just need one of them. (Not a strict requirement for this PR, it's just an illustration.)

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
andreamlin: I'm not sure of a way to implement gapic-only checks and discogapic-only checks in a clean way. One solution would be to put the input source (e.g. protobuf, discovery-doc) in input yaml, and then introduce `if` statements for every case that applies to different execution paths differently. 

This PR also introduces the config attribute `gax_http_version` which will be null for gapic execution paths, so null is a valid value for any path.

PR #1942, opened 2018-04-18
Add autogenerated marker at the top of the sample files.

vchudnov-g:
As per the design doc, this will be used to allow distinguishing autogenerated samples from manual samples.


src/test/java/com/google/api/codegen/testdata/py/py_library.baseline
```diff
@@ -3521,66 +3521,69 @@ def lint_setup_py(session):
     session.run(
         'python', 'setup.py', 'check', '--restructuredtext', '--strict')
 ============== file: samples/google/cloud/example/library_v1/gapic/get_book/get_book_sample_optional_array_method_sample_code_init_fields.py ==============
+#### [ This is an auto-generated sample file produced by the gapic-generator. Filename: "GetBookSampleOptionalArrayMethodSampleCodeInitFields" ]
```
garrettjonesgoogle: The filename here doesn't actually match the filename

vchudnov-g: Good catch! Thanks!

PR #1943, opened 2018-04-18
Don't emit standalone samples from sample_code_init_fields by default

vchudnov-g:
If standalone samples are configured but none matches the spec,
ensure that we don't accidentally create standalone samples for the sample_code_init_fields


src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
```diff
@@ -58,6 +58,13 @@ public SampleTransformer(SampleType sampleType) {
     this.sampleType = sampleType;
   }
 
+  /**
+   * A placeholder id for the ValueSet we synthesize in generateSamples() to accommodate legacy
+   * configurations that still rely on sample_code_init_fields. The format of this string is
+   * purposefully set to something that will cause errors if it makes it to an artifact.
+   */
+  private static String LEGACY_SAMPLE_CODE_INIT_VALUES = "[ sample_code_init_fields ]";
```
pongad: `private static final`?

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
```diff
@@ -89,13 +96,14 @@ public void generateSamples(
...
       methodViewBuilder.initCode(methodSampleViews.get(0).initCode());
-    }
 
-    // Make sure not to add the samples if they were not configured, since methodSampleViews may be
-    // non-empty solely because of the backward-compatibility accommodation.
-    if (!context.getMethodConfig().getSampleSpec().isConfigured()) {
-      methodSampleViews = new ArrayList<>();
+      // Don't emit samples that were specifically generated for backward-compatibility. We only
+      // wanted them for initCode above.
+      if (methodSampleViews.get(0).valueSet().id() == LEGACY_SAMPLE_CODE_INIT_VALUES) {
```
pongad: `.equals(LEGACY_SAMPLE_CODE_INIT_VALUES)`?

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
```diff
@@ -89,13 +96,14 @@ public void generateSamples(
...
+        methodSampleViews = new ArrayList<>();
```
pongad: Prefer `Collections.emptyList` so we can't add things to it by mistake.

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
vchudnov-g: Done.

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
vchudnov-g: Done.

PR #1944, opened 2018-04-19
Update generated php package

michaelbausor:
- Update generated code location
- Update namespaces
- Remove check for Google\Cloud prefix
- Add improved parameterized testing


src/main/java/com/google/api/codegen/util/php/PhpPackageUtil.java
```diff
@@ -62,4 +50,34 @@ public static String getFullyQualifiedName(String packageName, String objectName
...
+    }
+    if (foundVersion) {
+      return buildPackageName(Lists.reverse(packageComponents));
+    } else {
+      // If we did not find a version, then "FromVersionOnwards" is null
+      return null;
+    }
+  }
+
+  public static String getPackageNameBeforeVersion(String packageName) {
```
garrettjonesgoogle: It feels like the more fundamental thing you want is the "base package name", which everything else is a subpackage of. I think it would feel more intuitive to name things in terms of that.

src/main/java/com/google/api/codegen/transformer/php/PhpPackageMetadataNamer.java
```diff
@@ -21,12 +21,40 @@
...
+      composerPackage = composerVendor;
+    } else {
+      composerVendor = Name.anyLower(packagePieces[0]);
+      composerPackage = Name.from();
+      for (int i = 1; i < packagePieces.length; i++) {
+        composerPackage = composerPackage.join(Name.anyLower(packagePieces[i]));
+      }
+    }
+    metadataIdenfifier =
+        composerVendor.toSeparatedString("-") + "/" + composerPackage.toSeparatedString("-");
```
vam-google: Just in case: is it ok to use slash here explicitly (`"/"`)? I'm asking because there are abstractions around naming in gapic-generator, which abstract the actual separators used depending on a language (slash looks like a separator).

src/main/java/com/google/api/codegen/util/php/PhpPackageUtil.java
michaelbausor: Updated to use "base package name" language, PTAL

src/main/java/com/google/api/codegen/util/php/PhpPackageUtil.java
```diff
@@ -62,4 +50,40 @@ public static String getFullyQualifiedName(String packageName, String objectName
...
+   * version is present in the input packageName, returns null.
+   */
+  public static String removeBasePackageName(String packageName) {
+    ArrayList<String> packageComponents = new ArrayList<>();
+    List<String> pieces = Arrays.asList(PhpPackageUtil.splitPackageName(packageName));
+    boolean foundVersion = false;
+    for (String packageElement : Lists.reverse(pieces)) {
+      packageComponents.add(packageElement);
+      if (isPackageVersion(packageElement)) {
+        foundVersion = true;
```
vam-google: This looks redundant. You can just do `return buildPackageName(Lists.reverse(packageComponents));` here and `return null` at the end of the method (no `if` statement or `foundVersion` variable needed)

src/main/java/com/google/api/codegen/transformer/php/PhpPackageMetadataNamer.java
```diff
@@ -21,12 +21,40 @@
...
+    for (int i = 0; i < packagePieces.length; i++) {
+      packagePieces[i] = packagePieces[i].toLowerCase();
+    }
+
+    Name composerVendor;
+    Name composerPackage;
+    // The metadataIdentifier for composer is formatted as "vendor/package". If a
+    // domainLayerLocation is provided, set that as the vendor. Otherwise, take the first piece of
+    // packagePieces to use as the vendor.
+    if (domainLayerLocation != null && !domainLayerLocation.equals("")) {
```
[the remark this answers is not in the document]
michaelbausor: Fixed

src/main/java/com/google/api/codegen/transformer/php/PhpPackageMetadataNamer.java
michaelbausor: It is a separator in the package name. Maybe there should be a function in surface namer that contains this detail? I'm not sure. This is a php-specific class, so maybe it is acceptable to have php-specific formatting logic here.

src/main/java/com/google/api/codegen/util/php/PhpPackageUtil.java
michaelbausor: Fixed

src/main/java/com/google/api/codegen/transformer/php/PhpPackageMetadataNamer.java
```diff
@@ -21,12 +21,40 @@
...
-    this.domainLayerLocation = domainLayerLocation;
+    String packageNameWithoutVersion = PhpPackageUtil.getPackageNameBeforeVersion(packageName);
+    String[] packagePieces = PhpPackageUtil.splitPackageName(packageNameWithoutVersion);
+
+    this.serviceName = Name.upperCamel(packagePieces);
+
+    // To build metadata identifier, update pieces to be lowercase, so each piece is treated as a
+    // single piece in the Name object
+    for (int i = 0; i < packagePieces.length; i++) {
+      packagePieces[i] = packagePieces[i].toLowerCase();
```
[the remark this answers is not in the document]
michaelbausor: It will have the same effect here, but will slightly change the value in `serviceName`, which interprets the camel casing in the original. Rather than iterate here, we could perform two splits, one on `basePackageName` for `serviceName`, and another of `basePackageName.toLowerCase()` for `metadataIndentifier` - but I'm not sure if that is preferable.

garrettjonesgoogle:
You can proceed without an LGTM from me

PR #1945, opened 2018-04-19
Create and use CallingForm enum for standalone samples

vchudnov-g:


src/main/java/com/google/api/codegen/transformer/StaticLangApiMethodTransformer.java
```diff
@@ -324,7 +334,7 @@ public StaticLangApiMethodView generateRequestObjectAsyncMethod(
         Synchronicity.Async,
         additionalParams,
         methodViewBuilder,
-        Arrays.asList(type));
+        Arrays.asList(CallingForm.RequestAsync));
```
pongad: How do we know this is always `RequestAsync`?

src/main/java/com/google/api/codegen/transformer/StaticLangApiMethodTransformer.java
```diff
@@ -454,7 +464,11 @@ public StaticLangApiMethodView generateAsyncOperationFlattenedMethod(
     methodViewBuilder.exampleName(namer.getAsyncApiMethodExampleName(method));
     methodViewBuilder.callableName(namer.getCallableName(method));
     setFlattenedMethodFields(
-        context, additionalParams, Synchronicity.Async, methodViewBuilder, Arrays.asList(type));
+        context,
+        additionalParams,
+        Synchronicity.Async,
+        methodViewBuilder,
+        Arrays.asList(CallingForm.LongRunningFlattenedAsync));
```
pongad: Similar to above: how do we know this is always flattened and async?

src/main/java/com/google/api/codegen/transformer/StaticLangApiMethodTransformer.java
```diff
@@ -87,7 +88,7 @@ public StaticLangApiMethodView generatePagedFlattenedMethod(
         additionalParams,
         Synchronicity.Sync,
         methodViewBuilder,
-        Arrays.asList(ClientMethodType.PagedFlattenedMethod));
+        Arrays.asList(CallingForm.FlattenedPaging));
```
pongad: FYI, in general `Collections.singletonList` is more efficient, but no need to do it in this PR.

src/main/java/com/google/api/codegen/transformer/StaticLangApiMethodTransformer.java
vchudnov-g: We're in method `generateAsyncOperationFlattenedMethod`, so I'm taking the `Async` and `Flattened` at face value, and interpreting `Operation` as `LongRunning`. Are any of these inferred conventions incorrect? If so, we can trace the logic more thoroughly.

src/main/java/com/google/api/codegen/transformer/StaticLangApiMethodTransformer.java
vchudnov-g: We're in method `generateRequestObjectAsyncMethod`, so I'm taking the `Request` and `Async` at face value. Is this incorrect? If so, we can trace the logic more thoroughly.

src/main/java/com/google/api/codegen/transformer/StaticLangApiMethodTransformer.java
vchudnov-g: Ack.

src/main/java/com/google/api/codegen/transformer/StaticLangApiMethodTransformer.java
pongad: I looked through the code a little more. C# seems to also use the `type` for cancellation method. I'm good with this PR, but we might need to revisit.

PR #1946, opened 2018-04-19
Py rebase

pongad:


vchudnov-g:
lgtm. I want to change the structure of the sample somewhat, but I'll do that in a chaser somewhat

PR #1949, opened 2018-04-26
Java: prepare gapic-generator for proto/grpc artifact move to gcj

vam-google:
1) Generate non-standalone `pom.xml` for grpc and proto artifacts (assumes presence of a parent pom). The poms expect a parent to be present. This can be temporary (the pom generation can be moved somewhere else in the future). The artifacts are optional and are not required during self-service (oping the pom.xml out of self-service is TBD).
2) Remove `generated_package_version` and `generated_ga_package_version` support from java generation. By default (including self-service) all artifacts (including gapic) are generated with `0.0.0-SNAPSHOT` version.
3) "Resurrect" the usage of `dep-name_version` config properties (for example `google-common-protos_version` and `google-some-other-package-v1_version`). This corresponds to the plan of moving common dependencies to their own repo and releasing new versions of them only when necessary. Also this allows making self-service much more user friendly: now to build a client user needs to run generation only for that client and does not have to generate common dependencies (they will be pulled by gradle during build). Also this makes googleapis `api_defaults.yaml` obsolete for java.
4) Add `allJars` task in generated gapic `build.gradle`. This should simplify self-service, allowing to group all dependencies (and the generated jars themselves) in one folder, after which users of the self-service can choose which to copy to their build (i.e. it is an "exploded" fat jar, since the self-service users were already doing that: manipulating the dependencies content of the fat jar).
5) Upgrade `gradle-wrapper.propeities` used for generated artifacts to version `4.7`.

### DIFF
--- a/src/main/java/com/google/api/codegen/grpcmetadatagen/GrpcMetadataProviderFactory.java
+++ b/src/main/java/com/google/api/codegen/grpcmetadatagen/GrpcMetadataProviderFactory.java
@@ -50,14 +50,16 @@ public class GrpcMetadataProviderFactory {
             new JavaPackageMetadataTransformer(
                 ImmutableMap.of(
                     "LICENSE.snip", "LICENSE",
-                    "metadatagen/java/grpc/build_grpc.gradle.snip", "build.gradle"),
+                    "metadatagen/java/grpc/build_grpc.gradle.snip", "build.gradle",
+                    "metadatagen/java/grpc/pom_grpc.xml.snip", "pom.xml"),
                 artifactType));
       case PROTOBUF:
         return new JavaGrpcMetadataProvider(
             new JavaPackageMetadataTransformer(
                 ImmutableMap.of(
                     "LICENSE.snip", "LICENSE",
-                    "metadatagen/java/grpc/build_protobuf.gradle.snip", "build.gradle"),
+                    "metadatagen/java/grpc/build_protobuf.gradle.snip", "build.gradle",
+                    "metadatagen/java/grpc/pom_protobuf.xml.snip", "pom.xml"),
                 artifactType));
     }
 
--- a/src/main/resources/com/google/api/codegen/java/build_gapic.gradle.snip
+++ b/src/main/resources/com/google/api/codegen/java/build_gapic.gradle.snip
@@ -8,7 +8,8 @@
   apply plugin: 'java'
 
   description = 'GAPIC library for {@metadata.identifier}'
-  group = 'com.google.api'
+  group = 'com.google.cloud'
+  version = '0.0.0-SNAPSHOT'
   sourceCompatibility = 1.7
   targetCompatibility = 1.7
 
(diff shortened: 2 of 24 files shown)


src/main/java/com/google/api/codegen/grpcmetadatagen/GrpcMetadataProviderFactory.java
```diff
@@ -50,14 +50,16 @@
             new JavaPackageMetadataTransformer(
                 ImmutableMap.of(
                     "LICENSE.snip", "LICENSE",
-                    "metadatagen/java/grpc/build_grpc.gradle.snip", "build.gradle"),
+                    "metadatagen/java/grpc/build_grpc.gradle.snip", "build.gradle",
+                    "metadatagen/java/grpc/pom_grpc.xml.snip", "pom.xml"),
```
garrettjonesgoogle: Why generate both gradle and maven files?

src/main/resources/com/google/api/codegen/java/build_gapic.gradle.snip
```diff
@@ -8,7 +8,8 @@
   apply plugin: 'java'
 
   description = 'GAPIC library for {@metadata.identifier}'
-  group = 'com.google.api'
+  group = 'com.google.cloud'
```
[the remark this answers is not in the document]
vam-google: I believe it was wrong for them already. This group field is another problem which needs to be solved, but it is out of scope of this change (we should not have mention of `google.cloud` in the generator, but I believe we should also not have mention of google.api in the generator as well). For most "private" apis the `com.google.cloud` is what people would want. This is also what we would want. If somebody wants to test an updated api (for example DPE writing samples for some major update of an existing api), he can just run the self-serivce and get his client in proper group. The already are doing this actually. 

So I did this change not to fix something, but to mitigate issues (making the "wrong" thing "correct" for most common cases).

src/main/java/com/google/api/codegen/grpcmetadatagen/GrpcMetadataProviderFactory.java
vam-google: I mentioned it in the PR description: "(oping the pom.xml out of self-service is TBD)". We should not do it, but it seems it does not hurt (self-serivce workflow does not even imply people investigating the proto/grpc generated dirs, just run build from the root. Also before the recent cleanup we had a lot of random garbage in proto/gprc generated folders, so our customers are used to see random stuff there.

So this is a small issue, which shoudl be addressed, but after the bigger change (move to gcj).

garrettjonesgoogle: Ok, I'll allow it given impending work on codegen/packagegen splitting.

src/main/resources/com/google/api/codegen/java/build_gapic.gradle.snip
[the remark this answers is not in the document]
garrettjonesgoogle: I'll allow it with the understanding that we will make it configurable later.

PR #1950, opened 2018-04-26
Adding Bazel support; dep cleanup; other cleanup

garrettjonesgoogle:
For those not familiar with Bazel:

- BUILD.bazel declares the targets that can be built
- WORKSPACE declares the external "projects" (basically where to find dependencies)

Bazel made me add all directly referenced dependencies, which made it clear we were implicitly using code from several libraries but not declaring them in the build.gradle, so I added them there too. 

I had to do a couple other random things to get gapic-generator to build & run through Bazel.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,0 +1,57 @@
+java_plugin(
+    name = "auto_value_plugin",
+    processor_class = "com.google.auto.value.processor.AutoValueProcessor",
+    deps = ["@com_google_auto_value_auto_value//jar"],
+)
+
+java_library(
+    name = "auto_value",
+    exported_plugins = [":auto_value_plugin"],
+    exports = ["@com_google_auto_value_auto_value//jar"],
+)
+
+proto_library(
+    name = "config_proto",
+    srcs = ["src/main/proto/com/google/api/codegen/config.proto"],
+)
+
+java_proto_library(
+    name = "config_java_proto",
+    deps = [":config_proto"],
+)
+
+java_binary(
+    name = "gapic_generator",
+    srcs = glob(["src/main/java/**/*.java"]),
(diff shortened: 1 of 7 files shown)


BUILD.bazel
```diff
@@ -0,0 +1,57 @@
+java_plugin(
+    name = "auto_value_plugin",
+    processor_class = "com.google.auto.value.processor.AutoValueProcessor",
+    deps = ["@com_google_auto_value_auto_value//jar"],
+)
+
+java_library(
+    name = "auto_value",
+    exported_plugins = [":auto_value_plugin"],
+    exports = ["@com_google_auto_value_auto_value//jar"],
```
vchudnov-g: Does it make sense to move this `exports` to the `"auto_value_plugin"` target?

(Just learning about `exports` and `java_plugin`--I've not run into them before)

WORKSPACE
```diff
@@ -0,0 +1,142 @@
+workspace(name = "com_google_api_gapic_generator")
```
vchudnov-g: Would it make sense to add a file comment with the date these were generated?

BUILD.bazel
garrettjonesgoogle: This was the recommended way to do it... I'm not sure the reason for your suggestion, do you see a problem with it how it is?

WORKSPACE
garrettjonesgoogle: This wasn't generated, it was handwritten. (so I don't think it makes sense...)

BUILD.bazel
garrettjonesgoogle: I'm going to merge because it works as-is; we can discuss offline if you think something should change later.

PR #1952, opened 2018-04-26
Fix Node x-goog-api-client string

kjin:
`process.version.node` is undefined, so client libraries seem to be sending a header that starts with `gl-node/undefined`. Unless I'm mistaken, what we want is either `process.version` or `process.versions.node` (they are equivalent). ([docs](https://nodejs.org/dist/latest/docs/api/process.html#process_process_version))


ofrobots:
/cc @lukesneeringer. 

The fix looks good, but can a regression test be added for this. It seems like this regressed 7 months ago as a result of a refactor. Let's not have that happen again?

PR #1956, opened 2018-04-30
Case-insensitive Discovery message type names

andreamlin:
Windows NTFS is case-insensitive, so this PR escapes DiscoveryField identifiers that are the same, disregarding character casing.

Introduce global table of DiscoveryFields to prevent duplicate creation of DiscoveryFields from the same Schema.
Constructor for DiscoveryFields enforces unique, case-insensitive identifiers for message-type Schemas.
Refactor SchemaTypeName* utils to use DiscoveryFields instead of Schemas. This allows the TypeName utils to use the DiscoveryField's unique identifiers.

### DIFF
--- a/src/main/java/com/google/api/codegen/config/DiscoveryField.java
+++ b/src/main/java/com/google/api/codegen/config/DiscoveryField.java
@@ -30,8 +30,8 @@ import com.google.api.tools.framework.model.TypeRef.Cardinality;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
+import java.util.AbstractMap;
 import java.util.Comparator;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
@@ -55,7 +55,23 @@ public class DiscoveryField implements FieldModel, TypeModel {
   private static Comparator<Schema> messageSchemaComparator =
       (Schema s1, Schema s2) -> s1.getIdentifier().compareTo(s2.getIdentifier());
 
-  private static Map<Schema, DiscoveryField> globalObjects = new HashMap<>();
+  // Comparator for DiscoveryFields.
+  private static Comparator<Map.Entry<Schema, DiscoApiModel>> discoveryFieldComparator =
+      new Comparator<Map.Entry<Schema, DiscoApiModel>>() {
+        @Override
+        public int compare(
+            Map.Entry<Schema, DiscoApiModel> o1, Map.Entry<Schema, DiscoApiModel> o2) {
+          if (o1.getKey().hashCode() == o2.getKey().hashCode()) {
+            // if the schemas are the same, use the DiscoApiModel package name to differentiate.
+            return o1.getValue()
+                .getDefaultPackageName()
+                .compareTo(o2.getValue().getDefaultPackageName());
+          }
+          return o1.hashCode() - o2.hashCode();
+        }
+      };
+  private static Map<Map.Entry<Schema, DiscoApiModel>, DiscoveryField> globalObjects =
+      new TreeMap<>(discoveryFieldComparator);
 
   // Stores the escaped name for each message-type schema.
   private static Map<Schema, String> messageNames = new TreeMap<>(messageSchemaComparator);
@@ -70,9 +86,6 @@ public class DiscoveryField implements FieldModel, TypeModel {
   private DiscoveryField(Schema refSchema, DiscoApiModel apiModel) {
     Preconditions.checkNotNull(refSchema);
     Preconditions.checkNotNull(apiModel);
-    //    Preconditions.checkArgument(!Strings.isNullOrEmpty(apiModel.getDefaultPackageName()));
-    //    Preconditions.checkArgument(
-    //        apiModel.getDefaultPackageName().equals("com.google.cloud.simplecompute.v1"));
     this.originalSchema = refSchema;
     this.schema = refSchema.dereference();
     this.apiModel = apiModel;
@@ -104,16 +117,18 @@ public class DiscoveryField implements FieldModel, TypeModel {
   public static synchronized DiscoveryField create(Schema schema, DiscoApiModel rootApiModel) {
     Preconditions.checkNotNull(schema);
     Preconditions.checkNotNull(rootApiModel);
-    if (globalObjects.containsKey(schema)) {
+    Map.Entry<Schema, DiscoApiModel> fieldPrimaryKey =
+        new AbstractMap.SimpleEntry<>(schema, rootApiModel);
+    if (globalObjects.containsKey(fieldPrimaryKey)) {
       // DiscoveryField has already been created for this schema.
-      return globalObjects.get(schema);
+      return globalObjects.get(fieldPrimaryKey);
     }
     if (!Strings.isNullOrEmpty(schema.reference())) {
       // First create a DiscoveryField for the underlying referenced Schema.
       create(schema.dereference(), rootApiModel);
     }
     DiscoveryField field = new DiscoveryField(schema, rootApiModel);
-    globalObjects.put(schema, field);
+    globalObjects.put(fieldPrimaryKey, field);
     return field;
   }
 
@@ -213,7 +228,6 @@ public class DiscoveryField implements FieldModel, TypeModel {
       parentName = "";
     }
 
-    Preconditions.checkArgument(!Strings.isNullOrEmpty(apiModel.getDefaultPackageName()));
     return ResourceNameMessageConfig.getFullyQualifiedMessageName(
         apiModel.getDefaultPackageName(), parentName);
   }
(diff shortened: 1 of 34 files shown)


andreamlin:
PTAL

src/main/java/com/google/api/codegen/config/DiscoveryField.java
```diff
@@ -23,36 +23,89 @@
...
   private final List<DiscoveryField> properties;
+  private final DiscoApiModel apiModel;
   // Dereferenced schema to use for rendering type names and determining properties, type, and format.
   private final Schema schema;
-
   // Not dereferenced schema; used in rendering this FieldModel's parameter name.
   private final Schema originalSchema;
-  private final DiscoApiModel apiModel;
+
+  // Unformatted type name for this Field. For message-type Fields, this will be globally unique.
```
garrettjonesgoogle: Shouldn't this be unique per-package (namespace)?

src/main/java/com/google/api/codegen/discovery/Method.java
```diff
@@ -138,7 +138,7 @@ public Node parent() {
   /** @return the HTTP method. */
   public abstract String httpMethod();
 
-  /** @return the ID. */
+  /** @return the ID. This is unique. */
```
garrettjonesgoogle: Unique in what context?

src/main/java/com/google/api/codegen/transformer/SchemaTypeTable.java
```diff
@@ -69,12 +69,12 @@ public String renderPrimitiveValue(Schema type, String value) {
   }
 
   @Override
-  public String getNicknameFor(Schema type) {
+  public String getNicknameFor(DiscoveryField type) {
```
garrettjonesgoogle: What is the motivation for switching from `Schema` to `DiscoveryField` everywhere?

src/main/java/com/google/api/codegen/config/DiscoveryField.java
andreamlin: Great point! Yes indeed.

src/main/java/com/google/api/codegen/discovery/Method.java
andreamlin: Updated to say within the context of the parent Document.

src/main/java/com/google/api/codegen/transformer/SchemaTypeTable.java
andreamlin: There has to be a central place that contains Schemas and their escaped, unique names. I wanted to have DiscoveryField handle getting unique names. Every transformer must go through DiscoveryField to get that base, unique name. This ensures that these Schema names are consistent throughout the pipeline. 

An alternative would to put all the name-escaping logic in Schema, but Schema seemed like too much of a pure model kind of class that was just data and no logic.

andreamlin:
Made adjustments. PTAL

src/main/java/com/google/api/codegen/discovery/Schema.java
```diff
@@ -367,4 +368,30 @@ public boolean isPathParam() {
...
     return String.format("Schema \"%s\", type %s", getIdentifier(), type());
   }
+
+  /**
+   * @return hashCode that should be unique for each underlying Node in the Document. This function
+   *     includes the location of the node in its calculation, so two different nodes with the same
+   *     content but different parents will still have different hashCodes.
+   */
+  @Override
+  public int hashCode() {
```
garrettjonesgoogle: You need to either override both `hashCode` and `equals` or neither.

src/main/java/com/google/api/codegen/config/DiscoveryField.java
```diff
@@ -23,36 +23,93 @@
...
+          if (o1.getValue().getDefaultPackageName().equals(o2.getValue().getDefaultPackageName())) {
+            // if the namespaces are the same, use the schema's hashcode to differentiate.
+            return o1.getKey().hashCode() - o2.getKey().hashCode();
+          }
+          return o1.getValue()
+              .getDefaultPackageName()
+              .compareTo(o2.getValue().getDefaultPackageName());
+        }
+      };
+  private static Map<Map.Entry<Schema, DiscoApiModel>, DiscoveryField> globalObjects =
```
garrettjonesgoogle: This is... quite unconventional (use a map entry as a key). I would suggest making your own data class for the key.

garrettjonesgoogle:
+ @pongad in case I can't finish this review in a timely fashion - could you take a look too?

src/main/java/com/google/api/codegen/discovery/Schema.java
andreamlin: Everybody's invited to the override party! Done.

andreamlin: If I overwrite the hashCode function, then AutoValue doesn't generate a hashCode function. I wanted a custom hashCode function so that `parent` gets included as part of the hashCode, but AutoValue wouldn't have picked up `parent` because it isn't an abstract field getter.

src/main/java/com/google/api/codegen/config/DiscoveryField.java
```diff
@@ -23,36 +23,76 @@
...
+    this.originalSchema = refSchema;
+    this.schema = refSchema.dereference();
     this.apiModel = apiModel;
 
+    String simpleName = DiscoGapicParser.stringToName(refSchema.getIdentifier()).toLowerCamel();
+    String namespace = apiModel.getDefaultPackageName();
+    if (isTopLevelSchema(schema)) {
+      // Within this namespace, get a unique name for this message-type schema.
+      SchemaNamer messageNamer;
+      if (namespaceToSchemaNamer.containsKey(namespace)) {
```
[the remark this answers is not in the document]
andreamlin: Neato! Done.

src/main/java/com/google/api/codegen/config/DiscoveryField.java
```diff
@@ -61,25 +101,44 @@ private DiscoveryField(Schema schema, DiscoApiModel apiModel) {
   }
 
   /** Create a FieldModel object from a non-null Schema object. */
-  public static DiscoveryField create(Schema schema, DiscoApiModel rootApiModel) {
-    Preconditions.checkNotNull(schema);
-    Preconditions.checkNotNull(rootApiModel);
-    return new DiscoveryField(schema, rootApiModel);
+  public static synchronized DiscoveryField create(Schema schema, DiscoApiModel rootApiModel) {
```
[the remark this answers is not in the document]
andreamlin: `synchronized` because the method performs operations on a global `HashMap`. I can't predict how other people are going to use this class, so they could theoretically use multiple threads? I can take this out if it's overkill?

src/main/java/com/google/api/codegen/config/DiscoveryField.java
```diff
@@ -411,6 +473,65 @@ public String toString() {
...
+
+  // Util class for getting unique names within namespaces for message type schemas.
+  private class SchemaNamer {
+    private SymbolTable idSymbolTable = new SymbolTable(caseInsensitiveComparator);
+
+    // Stores the escaped name for each message-type schema.
+    private Map<Schema, String> messageNames = new TreeMap<>(messageSchemaComparator);
+
+    String getSchemaName(Schema schema, String basename) {
+      if (messageNames.containsKey(schema)) {
```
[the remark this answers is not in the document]
andreamlin: def def

src/main/java/com/google/api/codegen/config/DiscoveryField.java
```diff
@@ -411,6 +473,65 @@ public String toString() {
...
+    @Override
+    public boolean equals(Object other) {
+      if (!(other instanceof ModelData)) {
+        return false;
+      }
+      ModelData otherModelData = (ModelData) other;
+      return Objects.equals(
+              getApiModel().getDefaultPackageName(),
+              (otherModelData.getApiModel().getDefaultPackageName()))
+          && getSchema().hashCode() == otherModelData.getSchema().hashCode();
```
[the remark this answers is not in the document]
andreamlin: Sure.

src/main/java/com/google/api/codegen/util/SymbolTable.java
```diff
@@ -26,7 +28,22 @@
...
+
+  /**
+   * Create a SymbolTable with a custom comparison function. This can be used, for example, to make
+   * a case-insensitive symbol table by using a comparison function that orders two strings the same
+   * if they are the same in lowercase.
+   *
+   * @param comparator function to determine equality on Strings
+   */
+  public SymbolTable(Comparator<String> comparator) {
+    symbolTable = new ConcurrentSkipListSet<>(comparator);
```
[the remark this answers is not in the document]
andreamlin: Sure.

andreamlin:
Made fixes. PTAL

PR #1961, opened 2018-05-02
go: import "time" for LRO

pongad:
Package time is used by LRO methods.

Package time is usually already imported for retries,
so the fact that LRO didn't properly import time didn't cause us to
fail.

@jadekler


jadekler:
@pongad Not sure I grok, sorry :( What is LRO? Long running object? 

> LRO didn't properly import time didn't cause us to fail.

This change would cause the generator to fail?

pongad:
@jadekler LRO is "long running operation". APIs with long running operations need to import the time package, since we need to use it to configure how often we want to poll. Currently we use the time package without importing, leading to the failure. Does this make any sense?

jadekler:
Makes sense! Thanks for the quick fix! lgtm

PR #1962, opened 2018-05-04
Generate stub standalone sample files in NodeJS

vchudnov-g:
This follows the pattern established in https://github.com/googleapis/gapic-generator/pull/1874

### DIFF
--- a/src/main/resources/com/google/api/codegen/nodejs/standalone_sample.snip
+++ b/src/main/resources/com/google/api/codegen/nodejs/standalone_sample.snip
@@ -1,0 +1,36 @@
+@extends "nodejs/common.snip"
+@extends "nodejs/method_sample.snip"
+
+
+@snippet generate(sampleFile)
+   //// [ This is an auto-generated sample file produced by the gapic-generator. Sample name: "{@sampleFile.className}" ]
+   //// STUB standalone sample "{@sampleFile.className}" /////
+
+   // FIXME: Insert here set-up comments that we never want to display in cloudsite. These are seen by users perusing the samples directly in the repository.
+   
+   // [START full_sample]
+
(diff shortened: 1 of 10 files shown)


src/main/resources/com/google/api/codegen/nodejs/standalone_sample.snip
```diff
@@ -0,0 +1,36 @@
...
+
+@snippet generate(sampleFile)
+   //// [ This is an auto-generated sample file produced by the gapic-generator. Sample name: "{@sampleFile.className}" ]
+   //// STUB standalone sample "{@sampleFile.className}" /////
+
+   // FIXME: Insert here set-up comments that we never want to display in cloudsite. These are seen by users perusing the samples directly in the repository.
+   
+   // [START full_sample]
+
+   // FIXME: Insert here boilerplate code not directly related to the method call itself.
```
alexander-fenster: I wonder if it's possible to add two lines with client initialization here, e.g. like [these lines](https://github.com/googleapis/nodejs-dlp/blob/aeab8215ebdd1960b71e1594c727822ac7795a9b/samples/quickstart.js#L20) - if it's not hard it will make sample look much more usable.

vchudnov-g: Which lines are you referring to? The `require` line you linked to, or the `[START ...]` line?

These templates are stubs for now. I have the `FIXME`s as reminders to us/DPEs as to what content to insert where as we fill in the stubs.

vchudnov-g: OK, I see. Let me defer this to the next pass where we/DPE fill in the template files with real content. Let's just have the stubs for the moment.

PR #1965, opened 2018-05-04
all: make ApiModel return Lists

pongad:
Iterables are hard to stream.
Motivation: to make #1962 more convenient.


vchudnov-g:
I like the use of Java streams because it makes the code more compact and IMO slightly easier to understand, but what do you mean by "Motivation: to make #1962 more convenient"?

pongad:
@vchudnov-g There were a few places in that PR that, I think, would benefit from streams. Since that's already merged, there's no need to change it, but I thought it'd be good to be able to use streams in the future.

PR #1966, opened 2018-05-04
Fixes and updates for PHP toolkit constructor args

michaelbausor:
Corresponding gax PR: https://github.com/googleapis/gax-php/pull/174

### DIFF
--- a/src/test/java/com/google/api/codegen/testdata/php/php_library.baseline
+++ b/src/test/java/com/google/api/codegen/testdata/php/php_library.baseline
@@ -215,8 +215,6 @@ class LibraryServiceGapicClient
         ];
     }
 
-    private $operationsClient;
-
     private static function getShelfNameTemplate()
     {
         if (self::$shelfNameTemplate == null) {
@@ -2847,12 +2845,12 @@ use Google\LongRunning\GetOperationRequest;
 use Google\LongRunning\Operation;
 use Google\Protobuf\Any;
 use Google\Protobuf\GPBEmpty;
+use Google\Rpc\Code;
 use Google\Tagger\CustomNamespace\V1\AddLabelRequest;
 use Google\Tagger\CustomNamespace\V1\AddLabelResponse;
 use Google\Tagger\CustomNamespace\V1\AddTagRequest;
 use Google\Tagger\CustomNamespace\V1\AddTagResponse;
 use Google\Tagger\CustomNamespace\V1\LabelerGrpcClient;
-use Grpc;
 use PHPUnit\Framework\TestCase;
 use stdClass;
 
--- a/src/test/java/com/google/api/codegen/testdata/php/php_library.baseline
+++ b/src/test/java/com/google/api/codegen/testdata/php/php_library.baseline
@@ -203,13 +203,19 @@ class LibraryServiceGapicClient
         return [
             'serviceName' => self::SERVICE_NAME,
             'serviceAddress' => self::SERVICE_ADDRESS . ':' . self::DEFAULT_SERVICE_PORT,
-            'scopes' => self::$serviceScopes,
-            'clientConfigPath' => __DIR__ . '/../resources/library_service_client_config.json',
-            'restClientConfigPath' => __DIR__ . '/../resources/library_service_rest_client_config.php',
+            'clientConfig' => __DIR__ . '/../resources/library_service_client_config.json',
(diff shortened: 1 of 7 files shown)


src/test/java/com/google/api/codegen/testdata/php/php_library.baseline
```diff
@@ -203,11 +203,17 @@ class LibraryServiceGapicClient
...
-            'versionFile' => __DIR__ . '/../../VERSION'
+            'credentialsConfig' => [
+                'scopes' => self::$serviceScopes,
+            ],
+            'transportConfig' => [
+                'rest' => [
+                    'restClientConfigPath' => __DIR__ . '/../resources/library_service_rest_client_config.php',
+                ]
+            ],
+            'versionFile' => __DIR__ . '/../../../VERSION',
```
dwsupplee: Good catch here.

src/test/java/com/google/api/codegen/testdata/php/php_library.baseline
```diff
@@ -461,8 +467,8 @@ class LibraryServiceGapicClient
      *           The transport used for executing network requests. May be either the string `rest`
      *           or `grpc`. Defaults to `grpc` if gRPC support is detected on the system.
      *           *Advanced usage*: Additionally, it is possible to pass in an already instantiated
-     *           TransportInterface object. Note that when this objects is provided, any settings in
-     *           $transportConfig, and any $serviceAddress setting, will be ignored.
+     *           {@see TransportInterface} object. Note that when this object is provided, any
```
dwsupplee: There is an odd restriction that we don't have a workaround for in the doc generation yet. I believe the `@see` tag needs to include a FQNS in order to build out a link, if I remember correctly it doesn't take into account use statements.

src/test/java/com/google/api/codegen/testdata/php/php_library.baseline
```diff
@@ -437,18 +443,18 @@ class LibraryServiceGapicClient
      *     @type string $serviceAddress
      *           The address of the API remote host. May optionally include the port, formatted
      *           as "<uri>:<port>". Default 'library-example.googleapis.com:443'.
-     *     @type string|array $auth
+     *     @type string|array $credentials
```
dwsupplee: Could we port the documentation update from [here](https://github.com/googleapis/gax-php/blob/master/src/ApiCore/GapicClientTrait.php#L157) over as well?

src/test/java/com/google/api/codegen/testdata/php/php_longrunning.baseline
```diff
@@ -772,7 +782,10 @@ class OperationsClientTest extends GeneratedTest
      */
     private function createClient(array $options = [])
     {
-        return new OperationsClient($options);
+        return new OperationsClient($options + [
+            'serviceAddress' => '',
+            'scopes' => [],
```
dwsupplee: It looks like this, and the samples need to be updated. The scopes are passed in via `credentialsConfig` now.

src/test/java/com/google/api/codegen/testdata/php/php_library.baseline
dwsupplee: `FetchAuthTokenInterface` and `CredentialsWrapper` need imports. While checking to see if those existed, I noticed we now have a few other imports which are no longer used and can probably go.

```php
use Google\ApiCore\Call;
use Google\Auth\CredentialsLoader;
use Grpc\Channel;
use Grpc\ChannelCredentials;
```

src/test/java/com/google/api/codegen/testdata/php/php_longrunning.baseline
```diff
@@ -772,7 +783,12 @@ class OperationsClientTest extends GeneratedTest
      */
     private function createClient(array $options = [])
     {
-        return new OperationsClient($options);
+        return new OperationsClient($options + [
+            'serviceAddress' => '',
+            'transportConfig' => [
```
dwsupplee: Should this be `credentialsConfig`?

michaelbausor: Yep, fixed

src/test/java/com/google/api/codegen/testdata/php/php_library.baseline
michaelbausor: Updated imports

PR #1969, opened 2018-05-07
Python interceptor 2

landrito:
Depends on #1968. 

# What
This PR adds a class method to the generated clients which is used to create a gRPC channel that is scoped to the API. Currently, the way to use interceptors is a clunky flow since wrapped methods are generated at initialization making it such that all interceptors must be bound to the channel prior to initialization. This, in turn, makes it such that the user must know both the gRPC target and the auth scopes. This added method takes away the need for the user to know the target and scopes.

# Usage
## Simple Usage
```py
import grpc

from google.cloud.example import library_v1

channel = library_v1.LibraryServiceClient.create_channel()
channel = grpc.intercept_channel(channel, ExampleInterceptor())

client = library_v1.LibraryServiceClient(channel=channel)
```

## Wrapping with custom exceptions
```py 
import grpc

from google.cloud.example import library_v1

class CustomExceptionInterceptor(grpc.UnaryUnaryClientInterceptor):
    def intercept_unary_unary(self, continuation, client_call_details, request):
        response = continuation(client_call_details, request_or_iterator)
        if response.exception():
            raise CustomException(response.exception())
        return response


channel = library_v1.LibraryServiceClient.create_channel()
channel = grpc.intercept_channel(channel, CustomExceptionInterceptor())

client = library_v1.LibraryServiceClient(channel=channel)
```

## Adding custom metadata
```py
import grpc

from google.cloud.example import library_v1

class CustomExceptionInterceptor(grpc.UnaryUnaryClientInterceptor):
    def intercept_unary_unary(self, continuation, client_call_details, request):
        metadata = []
        if client_call_details.metadata is not None:
            metadata = list(client_call_details.metadata)
        metadata.append((
            '<HEADER>',
            '<VALUE'>,
        ))
        client_call_details = _ClientCallDetails(
            client_call_details.method, client_call_details.timeout, metadata,
            client_call_details.credentials)
        return continuation(client_call_details, request_or_iterator)


channel = library_v1.LibraryServiceClient.create_channel()
channel = grpc.intercept_channel(channel, CustomExceptionInterceptor())

client = library_v1.LibraryServiceClient(channel=channel)
```

### DIFF
--- a/src/main/resources/com/google/api/codegen/py/main.snip
+++ b/src/main/resources/com/google/api/codegen/py/main.snip
@@ -4,7 +4,7 @@
 @snippet generate(xapiClass)
     {@licenseSection(xapiClass.fileHeader)}
 
-    {@moduleDocstring(xapiClass)}
+    """Accesses the {@xapiClass.servicePhraseName}."""
     {@renderImportSection(xapiClass.fileHeader.importSection)}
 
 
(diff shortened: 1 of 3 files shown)


lukesneeringer:
@landrito This is being shelved in favor of #1974, yes? Can we close it?

lukesneeringer:
I assert this was shelved in favor of #1974. Reopen if not.

PR #1970, opened 2018-05-07
[DO NOT SUBMIT] Simplification proposal

garrettjonesgoogle:


alexander-fenster:
@hzyi-google Do you think this PR can be closed now? It's more than a year old.

PR #1976, opened 2018-05-08
use the latest google-gax version for Node.js

alexander-fenster:
The latest version of `google-gax` brings in new gRPC dependency that we want to use. 0.14.0 is way too old, and minor versions in the 0.x range are pinned (unfortunately). Let's use 0.16.1 explicitly.


garrettjonesgoogle:
This should be closed now - anyway, this change didn't impact the real generated code, only tests.

PR #1982, opened 2018-05-11
make Java standalone sample actually standalone

pongad:


pongad:
Too many conflicts. Sending separate PR

PR #1984, opened 2018-05-11
Moving api_defaults.yaml & dependencies.yaml into gapic-generator

garrettjonesgoogle:
- Moving dependencies.yaml & api_defaults.yaml from googleapis and into gapic-generator; creating DependenciesConfig and ApiDefaultsConfig classes to read them
- Creating PackagingConfig to read the values that are still passed from artman
  - Renaming some of the values so that they match what is in artman config
- PackageMetadataConfig is being retained as the "funnel" for the old way & the new way, so I'm intentionally not cleaning up anything there.
- The `package_yaml` parameter is retained for the old way, and a new `package_yaml2` parameter triggers the new way. If both are provided, `gapic-generator` barfs.

There is a corresponding `artman` change that goes with this (https://github.com/googleapis/artman/pull/418); since `gapic-generator` will still work with the old way, the changes don't need to be published simultaneously. I tested locally by creating a template which can render the values of PackageMetadataConfig per-language, then ran codegen for the old way & the new way, then did diffs for every language. I ran diffs for all APIs for Java, and most of the important ones for the other languages.


src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
```diff
@@ -349,69 +349,107 @@ public static PackageMetadataConfig createFromString(String yamlContents) {
...
+            .gaxHttpVersionBound(dependenciesConfig.gaxHttpVersionBound())
+            .grpcVersionBound(dependenciesConfig.grpcVersionBound())
+            .protoVersionBound(dependenciesConfig.protoVersionBound())
+            .authVersionBound(dependenciesConfig.authVersionBound())
+            .apiCommonVersionBound(dependenciesConfig.apiCommonVersionBound());
+
+    // api_defaults config
+    builder
+        .generatedNonGAPackageVersionBound(apiDefaultsConfig.generatedNonGAPackageVersionBound())
+        .author(apiDefaultsConfig.author())
```
michaelbausor: A question about the general approach here. We have had to introduce a yaml file and an autovalue file with non-trivial logic that can parse it, and contains a hardcoded reference to that particular yaml file, just to get these values. Can we just make them constants? We could have an ApiDefaultsConfig static class to hold them. The same applies to DependenciesConfig.

lukesneeringer:
Approved, but I wonder if we could not have `*_2.yaml`. I am not sure enough to block on it though.

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
garrettjonesgoogle: I agree with the sentiment, but I was aiming for a relatively straightforward move for this step. The status quo is already to parse these values; I'm now having it parse the values from a different location, so this change equates to moving the non-trivial logic from one place to another. 

I actually want to go even further that what you're talking about - for several of the defaults, I don't think we should even have them at all; I think we should require them to be passed in (e.g. "author" doesn't make sense to default IMO). For others, we probably don't need to hold them in a constants class - we can probably move them even closer to where they're used.

michaelbausor: Fair enough, and I don't want to block on this. If this is actually simpler that putting those values into constants somewhere, then we can make that change later and more thoroughly, as you suggest.

garrettjonesgoogle:
@lukesneeringer since this is an intermediate file, I'm not too concerned with the name. It already has a problem that the current language is not present in the file name, so generating a single API across multiple languages currently causes the file to be stomped. The gapic-generator side doesn't care about any patterns in the file name - it's the artman side that completely determines it.

PR #1990, opened 2018-05-14
travis: detect $unhandledCallingForm and fail

pongad:


testutils/check-baselines.sh
```diff
@@ -0,0 +1,59 @@
...
+callingFormsFile="src/main/java/com/google/api/codegen/viewmodel/CallingForm.java"
+
+# Calling forms are in an enum, so we look for capital identifiers that begins the line.
+callingForms=$(< $callingFormsFile grep -o '^[[:space:]]*[A-Z][a-zA-Z]*' | tr -d -c '[a-zA-Z\n]')
+
+unusedForms=$(
+	# Grep src/
+	#   recursively (-r)
+	#   in baseline files (--include='*.baseline')
+	#   for any calling forms: -f <(cat <<< $callingForms)
```
vchudnov-g: The use of the here string is neat, but `-f <(echo "$callingForms")` accomplishes the same thing, right?

vchudnov-g: The process substitution is fine and clever, but I'm just curious: any particular reason for  choosing `-f` rather than `-e`?
`grep ... -F $(echo $callingForms | sed 's/[^ ]* */-e &/g') src/`

The command line is similarly (not identically) long in both cases....

testutils/check-baselines.sh
```diff
@@ -0,0 +1,59 @@
...
+	#   match whole word (-w)
+	#   print only the calling form, not surrounding line (-o)
+	#   don't print file names (-h)
+	# Prints all occurences of used calling forms
+	grep --include='*.baseline' -h -w -o -r -F -f <(cat <<< $callingForms) src/ |
+
+	# Then turn it into a set.
+	sort -u |
+
+	# Both callingForms and used forms are sets, sorting them together means
```
vchudnov-g: Rephrase a bit:
"Both $callingForms  and the used forms determined so far in this pipeline are sets, so sorting ...."

testutils/check-baselines.sh
```diff
@@ -0,0 +1,59 @@
...
+	#   used forms appear twice together (one from callingForms, one from used forms)
```
vchudnov-g: "from $callingForms"
"the used forms"

testutils/check-baselines.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+
+set -e
+
+# TODO: We should figure out how to do this as part of GapicGeneratorTest at some point.
```
vchudnov-g: s/We should figure out how to do this/Do this/
s/at some point./so this check can be run locally./

testutils/check-baselines.sh
```diff
@@ -0,0 +1,59 @@
...
+fi
+
+## Check that every calling form is used somewhere.
+callingFormsFile="src/main/java/com/google/api/codegen/viewmodel/CallingForm.java"
+
+# Calling forms are in an enum, so we look for capital identifiers that begins the line.
+callingForms=$(< $callingFormsFile grep -o '^[[:space:]]*[A-Z][a-zA-Z]*' | tr -d -c '[a-zA-Z\n]')
+
+unusedForms=$(
+	# Grep src/
```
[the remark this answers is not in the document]
pongad: Thank you!

testutils/check-baselines.sh
```diff
@@ -0,0 +1,59 @@
...
+some sample code won't render properly.
+
+EOF
+	exitStatus=1
+fi
+
+## Check that every calling form is used somewhere.
+callingFormsFile="src/main/java/com/google/api/codegen/viewmodel/CallingForm.java"
+
+# Calling forms are in an enum, so we look for capital identifiers that begins the line.
```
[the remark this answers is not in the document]
pongad: Done

testutils/check-baselines.sh
pongad: Done

testutils/check-baselines.sh
pongad: Done

testutils/check-baselines.sh
pongad: Done. I edited slightly to "...so this check can be run with Gradle check command"

testutils/check-baselines.sh
pongad: If we have a lot of forms, splitting with `-e` might blow up the command line enough that the syscall won't execute. With proc substitution, the command line length is the same no matter how many forms we have.

pongad: This is slightly subtle but it does not. Documented this into the script.

pongad: We changed to `echo` to fix the issue around newlines in Travis (I should learn more about herestrings) so my objection to the change is mostly moot.

For information though: I think it's different. IIUC, expanding the string in bash is fine, but if the string gets too long, the kernel might not be able to start the process with so many args. By turning into a herestring, the long string is fed into stdin. The stdin gets big but that's OK; the process can still start.

PR #1991, opened 2018-05-15
Switching tests to use package_yaml2

garrettjonesgoogle:
Switching tests to use package_yaml2.

Also:
- Adding a `frozen_dependencies.yaml` file so that tests don't have to be updated when the main `dependencies.yaml` is updated
- Fixing grpc vs httpjson dependencies for Java - instead of triggering based on version presence in DependencyConfig (since all versions will be present), instead triggering based on artifact type
- Creating new files (_pkg2.yaml) instead of updating existing ones, in order to create less noise; I'll delete the old ones in a cleanup phase.
- Some Python/Node/Ruby baselines have diffs due to original dependencies on a multitude of versions in _pkg.yaml files, whereas now everything comes from a single source
- Note: I discovered as part of this update that I completely overlooked grpc metadata generation; I'll to a full testing pass between artman & gapic-generator as a later work item.

### DIFF
--- a/src/main/java/com/google/api/codegen/config/PackagingConfig.java
+++ b/src/main/java/com/google/api/codegen/config/PackagingConfig.java
@@ -19,10 +19,13 @@ import com.google.api.codegen.grpcmetadatagen.ArtifactType;
 import com.google.auto.value.AutoValue;
 import com.google.common.base.Joiner;
 import com.google.common.base.Strings;
+import com.google.common.io.Resources;
 import java.io.IOException;
+import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Paths;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import javax.annotation.Nullable;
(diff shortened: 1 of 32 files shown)


garrettjonesgoogle:
cc @andreamlin

src/test/java/com/google/api/codegen/testsrc/library_pkg2.yaml
```diff
@@ -0,0 +1,11 @@
...
+api_version: v1
+artifact_type: GAPIC
+organization_name: google-cloud
+proto_path: google/library
+proto_deps:
+- google-common-protos
+- google-some-other-package-v1
+proto_test_deps:
+- google-some-test-package-v1
+release_level: GA
```
michaelbausor: Would changing the release level here fix the baseline changes in dynamic langs? Or would it just cause other baseline changes elsewhere?

src/main/java/com/google/api/codegen/config/PackagingConfig.java
```diff
@@ -104,7 +107,7 @@ private static PackagingConfig createFromString(String yamlContents) {
             .apiName((String) configMap.get("api_name"))
             .apiVersion((String) configMap.get("api_version"))
             .organizationName((String) configMap.get("organization_name"))
-            .protoPackageDependencies((List<String>) configMap.get("proto_deps"))
+            .protoPackageDependencies(nullToEmpty((List<String>) configMap.get("proto_deps")))
```
michaelbausor: How about using Objects.firstNotNull here?

src/test/java/com/google/api/codegen/testsrc/library_pkg2.yaml
garrettjonesgoogle: Yeah it would make other baselines change (e.g. Java) - we previously had different inputs for different languages, and now this makes them all the same.

src/main/java/com/google/api/codegen/config/PackagingConfig.java
garrettjonesgoogle: ok I can do that

PR #2003, opened 2018-05-16
Fix NodeJS snippet function invocations missing new init parameter

vchudnov-g:
I believe there are no remaining calls to the deleted function signatures.


vchudnov-g:
FYI @lukesneeringer

PR #2009, opened 2018-05-18
Provide sample_code_init_fields to samples via sample_value_sets; apply to NodeJS

vchudnov-g:
For backward compatibility, `sample_init_fields` remains the source of
truth for `IN_CODE` samples for the moment. `sample_init_fields` is
now copied to the in-code sample view model so that as languages adopt
samplegen, those language's templates can stop depending on
`method.initCode`. Once all languages have done this mgration, we can
remove `method.initCode`

- Do the copying from `sample_init_fields` in `SampleTransformer`
- Make NodeJSPackageMetadataTransformer and
  NodeJSGapicSurfaceTestTransformer use NodeJSMethodViewGenerator
- Make the NodeJS in-code samples use the sample-configured init fields

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
@@ -62,13 +62,12 @@ public class SampleTransformer {
    * configurations that still rely on sample_code_init_fields. The format of this string is
    * purposefully set to something that will cause errors if it makes it to an artifact.
    */
-  private static final String LEGACY_SAMPLE_CODE_INIT_VALUES = "[ sample_code_init_fields ]";
+  private static final String INIT_CODE_SHIM = "[ shim for initcode ]";
 
   /**
-   * Populates methodViewBuilder.samples with the appropriate MethodSampleViews. This method is
-   * provided to provide backward compatibility for configuring in-code samples via
-   * MethodView.initCode. Once that configuration is removed, this function becomes a one-liner that
-   * should then be inlined at the callsites.
+   * Populates {@code methodViewBuilder.samples} with the appropriate {@code MethodSampleView}s.
+   * This method provides backward compatibility for configuring in-code samples via {@code
+   * MethodView.initCode}.
    *
    * @param methodViewBuilder
    * @param context
@@ -77,7 +76,7 @@ public class SampleTransformer {
    * @param generator The function (typically a lambda) to generate the InitCode for a sample given
    *     an InitCodeContext
    * @param callingForms The list of calling forms applicable to this method, for which we will
-   *     generate samples if so configured via context.getMethodConfig()
+   *     generate samples if so configured via {@code context.getMethodConfig()}
    */
   public void generateSamples(
       ApiMethodView.Builder methodViewBuilder,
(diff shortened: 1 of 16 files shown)


vchudnov-g:
One issue I have not figured out: how to create blank lines blank lines only between iterations of an `@join` in the templates. For spacing purposes, I had to work around my ignorance by creating a blank line at the end of each iteration, but that unfortunately leads to extra blank lines in the baselines (and doesn't look as pretty). If I can figure out how to do this properly, then the baseline files will not change in this PR.

I haven't spent much time researching how this might be doable. I vaguely recall it's being done somewhere else in our templates. If you happen to know where, that would save me some time. Otherwise, I will hunt for it next week.

alexander-fenster:
Changes LGTM if those nasty empty lines are removed. Unfortunately I don't really know where it should be changed.

vchudnov-g:
@alexander-fenster Should be fixed now (easier than I thought). PTAL.

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
```diff
@@ -77,18 +76,60 @@ public SampleTransformer(SampleType sampleType) {
    * @param generator The function (typically a lambda) to generate the InitCode for a sample given
    *     an InitCodeContext
    * @param callingForms The list of calling forms applicable to this method, for which we will
-   *     generate samples if so configured via context.getMethodConfig()
+   *     generate samples if so configured via {@code context.getMethodConfig()}
+   */
+  public void generateSamples(
+      ApiMethodView.Builder methodViewBuilder,
+      MethodContext context,
+      Iterable<FieldConfig> fieldConfigs,
```
pongad: Can we use `Collection` instead of `Iterable`? `Iterable` makes it really hard to use streaming API.
If it's too hard to do in this PR, I'm OK with it.

src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSMethodViewGenerator.java
```diff
@@ -43,49 +44,59 @@ public NodeJSMethodViewGenerator(DynamicLangApiMethodTransformer transformer) {
...
-                methodContext, packageHasMultipleServices, callingForms);
-      }
+    return context
+        .getSupportedMethods()
+        .stream()
+        .map(
+            methodModel ->
+                generateOneApiMethod(
+                    context.asDynamicMethodContext(methodModel), null, packageHasMultipleServices))
+        .collect((Collectors.toList()));
```
pongad: Unnecessary parens?

src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSMethodViewGenerator.java
```diff
@@ -43,49 +44,59 @@ public NodeJSMethodViewGenerator(DynamicLangApiMethodTransformer transformer) {
       GapicInterfaceContext context, boolean packageHasMultipleServices) {
     ImmutableList.Builder<OptionalArrayMethodView> apiMethodsAndSamples = ImmutableList.builder();
```
pongad: I think this variable isn't used anymore.

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
vchudnov-g: I suggest we do this separately, in a different PR.

src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSMethodViewGenerator.java
vchudnov-g: Guess I got carried away?

src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSMethodViewGenerator.java
vchudnov-g: Good catch!

PR #2011, opened 2018-05-21
Update PHP headers key

michaelbausor:


michaelbausor:
@garrettjonesgoogle Gentle ping

michaelbausor:
@dwsupplee @jdpedrie Can one of you PTAL?

This change is required to match the option key that Gax is expecting: https://github.com/googleapis/gax-php/blob/master/src/ApiCore/Transport/GrpcTransport.php#L172

PR #2019, opened 2018-05-23
go: don't overwrite the List requests

pongad:
In list calls, the iterator might make multiple RPCs with different page token to fetch items. Previously, we use the request object the user supplied and change its `PageToken` field in place. This behavior is probably unexpected and certainly unintentional.

This PR fixes the problem by making a copy of the request object for the iterator's own use and never modify the one the user passes in.

This is technically a breaking change: people might be counting on the request being modified, but this seems extremely unlikely.


jadekler:
I lack context - could you describe what this change does to generated code?

pongad:
@jadekler You can usually see the difference in the [baseline file](https://github.com/googleapis/gapic-generator/pull/2019/files#diff-b0f59f7d7d8379e79e9a826ca299d0da). But this is a good point, I'll update the issue description, and commit it when I squash.

PR #2023, opened 2018-05-24
Removing old package_config/package_yaml options and codepaths

garrettjonesgoogle:


src/main/java/com/google/api/codegen/config/ApiDefaultsConfig.java
```diff
@@ -40,15 +40,9 @@
...
-  protected abstract Map<TargetLanguage, ReleaseLevel> releaseLevel();
-
   /** The development status of the client library. Configured per language. */
-  public ReleaseLevel releaseLevel(TargetLanguage language) {
-    ReleaseLevel level = releaseLevel().get(language);
-    if (level == null) {
-      level = ReleaseLevel.UNSET_RELEASE_LEVEL;
-    }
-    return level;
+  public ReleaseLevel releaseLevel() {
```
michaelbausor: I don't understand this change, what is happening here?

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
```diff
@@ -400,24 +284,13 @@ public static PackageMetadataConfig createFromPackaging(
...
-      Map<TargetLanguage, ReleaseLevel> result = new HashMap<>();
-      for (Map.Entry<TargetLanguage, ReleaseLevel> entry : lhs.entrySet()) {
-        // This just sets every language's release level to the release level of the current
-        // context. This is admittedly a little heavy-hadned, but it doesn't matter since
-        // we only run one language in an invocation of gapic-generator.
-        result.put(entry.getKey(), thisReleaseLevel);
-      }
-      return result;
-    }
+  private static ReleaseLevel mergeReleaseLevels(
```
michaelbausor: Can we remove this function entirely and use firstNotNull?

src/main/java/com/google/api/codegen/config/ApiDefaultsConfig.java
garrettjonesgoogle: In the prior set of changes (moving api_defaults/dependencies into gapic-generator), I changed the package metadata format to contain only one release level instead of a mapping of release levels per language. The reasoning is that in the end, only the release level of the target language was actually used (this was verified through testing all languages through artman). In those prior changes, I syntactically preserved the language -> release level mapping in gapic-generator, and just populated all languages with the same value. Now that the transition is complete and I don't need to preserve the old structure, I am removing the parameterization by language.

garrettjonesgoogle:
I don't think there's terribly much value in fixing the "_2" naming, particularly with the parameter names, because I will be adding a third format later anyway. I think it can actually help people to remember what version they are looking at.

garrettjonesgoogle:
(by "parameter names" I meant to say "command line arguments")

src/main/java/com/google/api/codegen/config/PackageMetadataConfig.java
garrettjonesgoogle: Will do.

garrettjonesgoogle:
PTAL

Issue #1999, opened 2018-05-16
PHP generated unit tests take a long time to execute


michaelbausor:
The PHP generated unit tests take multiple minutes for the full suite of cloud apis - this should be an order of magnitude less. Watching test progress, it seems some tests take much longer than others - it isn't clear why this should be without further investigation.

PR #2026, opened 2018-05-25
Set PHP initial poll delay in LRO tests

michaelbausor:
This fixes https://github.com/googleapis/gapic-generator/issues/1999


michaelbausor:
@dwsupplee Can you PTAL?

PR #2027, opened 2018-05-25
Removing dependency on GAX

garrettjonesgoogle:


garrettjonesgoogle:
As discussed offline, I removed `RetrySettings` altogether and just used the proto-generated class directly. Now we have a net removal of lines! PTAL

PR #2028, opened 2018-05-28
Cleanup - renames, moves, deletes

garrettjonesgoogle:
Performing a bunch of renames, moves, and deletes. The core thing I am doing is merging the GapicProvider & PackageProvider interfaces, so that individual providers can be moved around more easily - later I will be moving certain providers from codegen to packagegen. At the same time, I am cleaning up the root package

Renames:
- CodeGeneratorTool -> GapicGeneratorTool, CodeGeneratorApi -> GapicGeneratorApi

Moves:
  - Moving Gapic-specific stuff from the root to gapic/
  - Moving DiscoGapic-specific stuff from the root to discogapic/
- Moving shared stuff from the root to a new suppackage common/
  - CodegenContext
  - GeneratedResult
  - TargetLanguage
- Moving util stuff from the root to util/
  - Inflector
  - LanguageUtil
  - CommentPatterns
Deletes:
- Deleting SynchronizerTool

Other changes:

- Merging GapicProvider and PackageProvider into CodeGenerator, using the GapicProvider design (all inputs are provided at construction time), putting it under common/
  - JavaGrpcPackageProvider and PythonGrpcPackageProvider now have to take Model and PackageMetadataConfig in their constructors
- Deleting InputElementView, which was an abstraction around iterating over interfaces that was not needed. 
  - Renaming its only child, InterfaceView -> ProtoModels
  - Renaming the only method getElements -> getInterfaces, and making it static.
  - Removing the parameterization of CommonGapicProvider by Element, which is always Interface.
- Renaming ProtoFileView to ProtoFiles, moving it under gapic/, making its methods static
- Moving ServiceMessages under gapic, removing unused methods


src/main/java/com/google/api/codegen/common/OutputProvider.java
```diff
@@ -12,19 +12,19 @@
...
- * A GapicProvider performs file generation (code, static resources, etc.) using a proto-based Model
- * for a particular language.
+ * An OutputProvider performs file generation (code, config, etc) or file copying (static resources)
+ * based on input provided at construction time.
  *
- * @param <T> Class which represent generated file body (may be {@code byte[]})
+ * @param <T> Class which represents the generated file body.
  */
-public interface GapicProvider<T> {
+public interface OutputProvider<T> {
```
michaelbausor: `OutputProvider` is a very generic name. And it doesn't seem to really be a "provider" - it is a generator. `OutputFileGenerator` may be better, or `OutputGenerator` (but that is even more generic).

src/main/java/com/google/api/codegen/gapic/ServiceMessages.java
```diff
@@ -0,0 +1,34 @@
...
+import com.google.api.tools.framework.model.TypeRef;
+import com.google.protobuf.Empty;
+
+/** Utility class with methods to work with service methods. */
+public class ServiceMessages {
+
+  private static final String LRO_TYPE = "google.longrunning.Operation";
+
+  /** Returns true if the message is the empty message. */
+  public boolean isEmptyType(TypeRef type) {
```
michaelbausor: Why is this pattern (having an instance method delegate to a static method) necessary?

src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSGapicSurfaceDocTransformer.java
```diff
@@ -51,7 +51,7 @@
   public List<ViewModel> transform(ApiModel apiModel, GapicProductConfig productConfig) {
     Model model = ((ProtoApiModel) apiModel).getProtoModel();
     ImmutableList.Builder<ViewModel> surfaceDocs = ImmutableList.builder();
-    for (ProtoFile file : new ProtoFileView().getProtoFiles(productConfig)) {
+    for (ProtoFile file : new ProtoFiles().getProtoFiles(productConfig)) {
```
michaelbausor: Should we use ProtoFiles statically?

src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceDocTransformer.java
```diff
@@ -68,7 +68,7 @@ public RubyGapicSurfaceDocTransformer(
   @Override
   public List<ViewModel> transform(ApiModel model, GapicProductConfig productConfig) {
     ImmutableList.Builder<ViewModel> surfaceDocs = ImmutableList.builder();
-    for (ProtoFile file : new ProtoFileView().getProtoFiles(productConfig)) {
+    for (ProtoFile file : new ProtoFiles().getProtoFiles(productConfig)) {
```
michaelbausor: Use ProtoFiles  statically?

src/main/java/com/google/api/codegen/common/OutputProvider.java
garrettjonesgoogle: It seems like the right name for this is `Generator`; I think the reason we didn't do this before is that there is already an inner interface of `SnippetSetRunner` named `Generator`. It looks like that is only used by client config generation though; all other generation goes through `CommonSnippetSetRunner`. Maybe we should rename that inner interface so we can use the name. I think `Generator` is appropriate for the use case here because the output is `GeneratedResult`, and the interface really is quite general. What do you think? I would also be ok with `OutputGenerator` as you suggest, although the `Output` part is potentially so generic that it is redundant. We could potentially go for `CodeGenerator` also, with the understanding that config, readmes, etc all count as code. I'm a little wary of `FileGenerator` because no files are actually generated at this stage.

src/main/java/com/google/api/codegen/gapic/ServiceMessages.java
garrettjonesgoogle: For the snippet language - it can't access static methods.

src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSGapicSurfaceDocTransformer.java
garrettjonesgoogle: yep, I overlooked this when I did the conversion.

src/main/java/com/google/api/codegen/common/OutputProvider.java
michaelbausor: `CodeGenerator` or `Generator` SGTM.

garrettjonesgoogle:
Renamed `OutputProvider` to `CodeGenerator`, and merged the `SnippetSetRunner.Generator` interface into `SnippetSetRunner` itself. PTAL

PR #2030, opened 2018-05-29
Ruby: move credentials file to version namespace

landrito:
# What
This moves the generated credentials class to the namespace of the package name rather than being a top level file.


landrito:
I don't think that will be a problem since there will only be one subpackage which is declared by the package name field in the gapic_yaml.

PR #2032, opened 2018-05-29
Add gRPC interceptor support

jbolinger:
Add support for enabling custom gRPC interceptors via a subclass hook (`GRPC_INTERCEPTORS` following the pattern from `SERVICE_ADDRESS`, etc.)

Associated GAX change: https://github.com/googleapis/gax-ruby/pull/127/files


jbolinger:
@lukesneeringer thanks - hoping to do this now just waiting for CI (seems to have passed but wasn't reported)

PR #2036, opened 2018-05-30
Merging main() methods, referencing concrete provider factories

garrettjonesgoogle:
- Merging `*Tool` classes into a single `GeneratorMain` class, consolidating command-line options
  - Artman is unaffected because it still goes through the gradle targets. Eventually, I want to have Artman just invoke the jar directly, using `GeneratorMain` with the desired artifact.
- Creating new `ArtifactType` enum, each of which maps to a previous `main()` method
  - Several start with `LEGACY_` because I will be replacing them with more specific artifact types later
  - Later, we'll probably have multiple artifacts map to a single tool driver class (e.g. for packaging stuff)
- Renaming `packagegen.ArtifactType` to `packagegen.PackagingArtifactType`
- Renaming `(*)Api` classes to `{$1}App` - I think the "Api" terminology is really confusing
- Deleting the dynamic factory lookup & instantiation code, just hardcoding the factory type to use in each `*App` class (this essentially means the factory class specified in files like https://github.com/googleapis/googleapis/blob/master/gapic/lang/java_gapic.yaml is now ignored, but I think this has really ossified anyway)
  - Also deleting provider factory interfaces
- Removing `Main` from the provider factory names


eoogbe:
Remove the `Main` prefix from the names of the concrete provider factories.

garrettjonesgoogle:
@eoogbe done.

garrettjonesgoogle:
I realized that after I removed `Main` from the provider factory names, it made a bit of a mess. Look at the first commit to see what actually changed for the provider factory classes.

As for reviewing more closely: I would say `GeneratorMain.main()`, because it is new code. The rest of `GeneratorMain` is copied + modified code (a lot of strings were converted to `${OPTION_CONSTANT}.getLongOpt()` instead).

PR #2037, opened 2018-05-30
java: give standalone samples their own classes

pongad:
This makes it easier to copy-paste.


src/main/java/com/google/api/codegen/util/java/JavaRenderingUtil.java
```diff
@@ -22,12 +22,20 @@
...
    * Splits given text into lines and returns an list of strings, each one representing a line.
-   * Performs escaping of certain html characters.
+   * Performs escaping of certain HTML characters.
    */
   public static List<String> getDocLines(String text) {
     return CommonRenderingUtil.getDocLines(new JavaCommentReformatter().reformat(text));
   }
 
+  /**
+   * Splits given text into lines and returns an list of strings, each one representing a line,
```
vchudnov-g: s/an/a/

PR #2038, opened 2018-05-31
Merge DocumentToViewTransformer into ModelToViewTransformer

eoogbe:
Merges redundant transformer interfaces and discogapic providers.

`ModelToViewTransformer` handles all transformations for `ApiModel`s. `DiscoGapicProvider` handles all code generation for discogapic.

This cleanup reduces references to view models as part of replacing view models with views in the GAPIC view simplification.


src/main/java/com/google/api/codegen/transformer/java/JavaGapicPackageTransformer.java
```diff
@@ -26,8 +26,8 @@
 import java.util.List;
 
 /** Responsible for producing package meta-data related views for Java GAPIC clients */
-public class JavaGapicPackageTransformer extends JavaPackageTransformer
-    implements ModelToViewTransformer {
+public class JavaGapicPackageTransformer<T extends ApiModel> extends JavaPackageTransformer
+    implements ModelToViewTransformer<T> {
```
garrettjonesgoogle: I think `ApiModelT` would be clearer than `T`

src/main/java/com/google/api/codegen/discogapic/DiscoGapicProviderFactory.java
```diff
@@ -69,17 +70,18 @@
...
-                .setDocumentToViewTransformers(transformers)
+                .setModelToViewTransformers(transformers)
                 .build();
 
         providers.add(provider);
 
         CodeGenerator metadataProvider =
-            ViewModelDiscoGapicProvider.newBuilder()
-                .setModel(model)
+            DiscoGapicProvider.newBuilder()
```
garrettjonesgoogle: I didn't notice the duplication between `DiscoGapicProvider` and `ViewModelDiscoGapicProvider` before.

I'm thinking we should also rename `ViewModelGapicProvider` to `GapicProvider` to be consistent. (The `GapicProvider` name was freed up when I renamed that interface to `CodeGenerator`)

Taking the naming even further, we could instead use `DiscoGapicGenerator` and `GapicGenerator`. (And then, `DiscoGapicGeneratorFactory` and `GapicGeneratorFactory`...)

I'll leave that last suggestion up to you (it would be cleaner to do it in a following PR anyway I think)

eoogbe: I will do this in a follow-up PR

PR #2039, opened 2018-05-31
Fix PHP doc reference to build methods

michaelbausor:
cc @dwsupplee


michaelbausor:
@garrettjonesgoogle PTAL for merging approval

PR #2040, opened 2018-05-31
Adding new --language command-line arg

garrettjonesgoogle:
This is backwards compatible with artman. 

- Plumbing `TargetLanguage` throughout the code
- Removing the "generator id" concept altogether.
  - Node: using the NODEJS_DOC codepath; updating the Node baselines to reflect generation of docs
  - Ruby: using the RUBY_DOC codepath; updating the Ruby baselines to reflect generation of docs
- Deleting `GapicGeneratorConfig`, passing around its components instead
- Deleting CLIENT_CONFIG support - Artman doesn't use it, and all relevant languages perform the same generation anyway


garrettjonesgoogle:
cc @andreamlin (for discogapic changes), cc @jbolinger (for Ruby changes - FYI only, it shouldn't change real API generation output), cc @alexander-fenster (for NodeJS changes - FYI only, it shouldn't change real API generation output)

michaelbausor:
Changes LGTM, but I didn't get why the baselines need to be renamed, and why there were changes in ruby and node?

alexander-fenster:
LGTM from my side, let's see :)

garrettjonesgoogle:
I decided to re-enable sample test generation within the baseline tests (it will stay off for real generation, which is the status quo)

PR #2042, opened 2018-05-31
Python: add transport class

landrito:
~DO NOT MERGE. THIS IS A WIP.~
Now theoretically ready for merge.


lukesneeringer:
I think this is ready now. Since @landrito and I both worked on it, and because her team will be affected, asking @theacodes to review.

src/test/java/com/google/api/codegen/testdata/py/python_no_path_templates.baseline
```diff
@@ -928,16 +893,23 @@ class NoTemplatesAPIServiceClient(object):
...
-        if 'increment' not in self._inner_api_calls:
-            self._inner_api_calls['increment'] = google.api_core.gapic_v1.method.wrap_method(
-                self._no_templates_api_service_stub.Increment,
+        # Wrap the transport method to add retry and timeout logic.
+        #
+        # Note: We use the hash of the original transport method
+        # in order to automatically invalidate the lookup if the
+        # transport method changes in-place (this can happen because
+        # the stubs must be regenerated if interceptors are added.)
+        cache_key = hash(self.transport.increment)
```
theacodes: This is *super* gross. This is breaking fundamental assumptions about the language and working around them in unfortunate ways. Transport should be immutable.

src/test/java/com/google/api/codegen/testdata/py/python_no_path_templates.baseline
```diff
@@ -970,9 +942,110 @@ config = {
 # limitations under the License.
 
 
+import google.api_core.grpc_helpers
+
+from example.proto import no_path_templates_pb2_grpc
+
+
+class NoTemplatesApiServiceGrpc(object):
```
theacodes: GrpcTransport?

src/test/java/com/google/api/codegen/testdata/py/python_no_path_templates.baseline
```diff
@@ -970,9 +942,110 @@ config = {
...
+            Callable: A callable which accepts the appropriate
+                deserialized request object and returns a
+                deserialized response object.
+        """
+        return self._stubs['no_templates_api_service_stub'].Increment
+
+
+    @property
+    def _stubs(self):
+        if '_stubs' not in self.__dict__:
```
theacodes: why are we doing late binding on stubs?

src/test/java/com/google/api/codegen/testdata/py/python_no_path_templates.baseline
lukesneeringer: The problem this is working around is that the stubs are created with the transport's channel, which can then change. I agree with you that this is not ideal, however.

One alternate idea would be to put the wrapper on interceptors on the primary client (I believe it is there already) and then simply make that write _a new transport object_ to the client. That would solve all of the issues on the transport and make it be frozen.

It does that at the expense of making the client's transport mutable, which feels pretty bad (but perhaps less bad).

theacodes:
Why do we consider the channel mutable? I totally expect both transport and
channel to be immutable.

src/test/java/com/google/api/codegen/testdata/py/python_no_path_templates.baseline
landrito: Would we be able to get around this simply by forcing interceptor binding to happen prior to the first API call a client makes?

theacodes:
That seems like a band-aid.

src/main/java/com/google/api/codegen/transformer/py/PythonImportSectionTransformer.java
```diff
@@ -14,20 +14,9 @@
  */
 package com.google.api.codegen.transformer.py;
 
-import com.google.api.codegen.config.ApiModel;
-import com.google.api.codegen.config.GapicProductConfig;
-import com.google.api.codegen.config.InterfaceModel;
-import com.google.api.codegen.config.ProductConfig;
-import com.google.api.codegen.config.TypeModel;
+import com.google.api.codegen.config.*;
```
eoogbe: I believe star imports are discouraged. What prompted the change?

landrito: Intellij handles my importing so /shrug. I'll look into it.

src/test/java/com/google/api/codegen/testdata/py/python_no_path_templates.baseline
lukesneeringer: @landrito Could you make this happen?

lukesneeringer:
@theacodes Could you do another surface review? I think we have this where you are hoping. :-)

src/test/java/com/google/api/codegen/testdata/py/python_no_path_templates.baseline
theacodes: Bump on this, I still think it should be `GrpcTransport`.

PR #2046, opened 2018-06-01
Assign and use the correct calling forms for Python samples

vchudnov-g:
* The logic for determining calling forms has been moved to the
  generator

* The standalone samples key off the calling form to call the correct
  sample-generating function (the routing is implemented here; the
  functions called at the moment as stubs are actually the in-code
  sample functions)

* The in-code samples are now also keyed off the calling forms


pongad:
LGTM but Luke should sign off on the minor change to Python doc.

PR #2047, opened 2018-06-01
Update default gax and api-common deps

andreamlin:
gax-java and api-common dependencies have already been updated in google-cloud-java.


andreamlin:
PTAL

PR #2050, opened 2018-06-04
test: check calling forms used in specified langs

### LINKED ISSUE #2045: samplegen: check-baselines should check that each calling form is used in the language listed
Right now `check-baselines` only checks that each calling form appears in at least one language baseline. 

It should also check that each calling form also appears in each of the languages listed in its comment in `CallingForm.java`. We can do this either by checking directly against the baseline, or by checking against the `library_gapic.yaml` (and then relying on `check-calling-forms-in-baseline` to check the baseline against the yaml.

pongad:
Previously `check-baselines.sh` only checks that all declared calling forms are used in some `*.baseline` files. This is insufficient, as we also want to make sure that the forms are used specifically by the languages in the comments.

Since the logic is now too complex to be profitably written in shell, this PR introduces a new test-script `check-lang-uses-form` to perform this check and deletes the corresponding portions of the script.

Fixes #2045.


vchudnov-g:
So now there are two go testing utils we run? Could you write in the PR description an overview of what this new one does and how it differs from the previous one? This looks like an additive change, so saying it "fixes" #2045 is  confusing.

testutils/check-lang-uses-form/main.go
```diff
@@ -0,0 +1,185 @@
...
+
+// readChecks reads calling forms and languages where calling forms are expected in from r.
+// The format is described by the program's usage text.
+func readChecks(r io.Reader) (map[string][]string, error) {
+	checks := make(map[string][]string)
+	sc := bufio.NewScanner(r)
+	for sc.Scan() {
+		ln := sc.Text()
+		p := strings.Index(ln, "//")
+		const prefix = "// used by:"
```
vchudnov-g: Ooh, the space here after the `//` is significant, which would not have been obvious to me as a user. Let's note that in the usage.

testutils/check-lang-uses-form/main.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+}
+
+// readChecks reads calling forms and languages where calling forms are expected in from r.
+// The format is described by the program's usage text.
+func readChecks(r io.Reader) (map[string][]string, error) {
+	checks := make(map[string][]string)
+	sc := bufio.NewScanner(r)
+	for sc.Scan() {
+		ln := sc.Text()
```
vchudnov-g: I suggest s/`ln`/`line`/ for readability

testutils/check-lang-uses-form/main.go
```diff
@@ -0,0 +1,185 @@
...
+		}
+		form := ln[:p]
+		form = strings.TrimSpace(form)
+		form = strings.TrimRight(form, ",")
+
+		for _, lang := range strings.Fields(ln[p+len(prefix):]) {
+			checks[lang] = append(checks[lang], form)
+		}
+	}
+	if err := sc.Err(); err != nil {
```
vchudnov-g: If you're ok returning non nil `checks` in case of error, the if can go away:
`return checks, sc.Err()`

vchudnov-g: My personal preference would be s/`sc`/`scanner`/ since by the time I got here I forgot was `sc` was.

testutils/check-lang-uses-form/main.go
```diff
@@ -0,0 +1,185 @@
...
+		for lang, forms := range checks {
+			for _, form := range forms {
+				fmt.Printf("{%q, %q}\n", lang, form)
+			}
+		}
+		os.Exit(1)
+	}
+}
+
+// readChecks reads calling forms and languages where calling forms are expected in from r.
```
vchudnov-g: "...reads in from r..."

testutils/check-lang-uses-form/main.go
```diff
@@ -0,0 +1,185 @@
...
+// The format is described by the program's usage text.
```
vchudnov-g: "The format for r is ..."

vchudnov-g: Clarify this returns a map of language to list of calling forms to check for that language.,

testutils/check-lang-uses-form/main.go
```diff
@@ -0,0 +1,185 @@
...
+		if len(words) == 0 {
+			return nil
+		}
+
+		baseline, err := ioutil.ReadFile(path)
+		if err != nil {
+			return err
+		}
+
+		words = notFoundWords(string(baseline), words)
```
vchudnov-g: I claim the following renaming would make the code easier to follow:
s/`words`/`formsToCheck`/
s/`notFoundWords`/`formsRemaining`/

testutils/check-lang-uses-form/main.go
```diff
@@ -0,0 +1,185 @@
...
+
+// hasWord reports wheter s contains word w. See notFoundWords for details.
+func hasWord(s, w string) bool {
+	p := 0
+	for {
+		dp := strings.Index(s[p:], w)
+		if dp < 0 {
+			return false
+		}
+		if wordChar(utf8.DecodeLastRuneInString(s[:p+dp])) {
```
vchudnov-g: How about:

```
if !wordChar( ) && !wordChar( ) {
  return true
}
p += dp + len(w)
```

testutils/check-lang-uses-form/main.go
```diff
@@ -0,0 +1,185 @@
...
+		if !hasWord(s, w) {
+			notFound = append(notFound, w)
+		}
+	}
+	return notFound
+}
+
+// hasWord reports wheter s contains word w. See notFoundWords for details.
+func hasWord(s, w string) bool {
+	p := 0
```
vchudnov-g: Why not implement this via a regex, which will be easier to follow and verify? For the word boundaries, I think you could even use one of the predefined character classes in https://github.com/google/re2/wiki/Syntax#ascii, since those are supported by https://golang.org/pkg/regexp/ Ex:
```
pattern := fmt.Sprintf("[^[[:word:]]]+%s[^[[:word:]]]+",w)
```

testutils/check-lang-uses-form/main_test.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	wants := map[string][]string{
+		"java": {"Form1", "FormNoComma"},
+		"py":   {"Form1"},
+	}
+
+	for lang, want := range wants {
```
vchudnov-g: consider inlining `wants`

testutils/check-lang-uses-form/main.go
```diff
@@ -0,0 +1,185 @@
...
+		flag.PrintDefaults()
+		fmt.Fprint(out, usage)
+	}
+	flag.Parse()
+
+	if *formFname == "" {
+		flag.Usage()
+		os.Exit(1)
+	}
+
```
vchudnov-g: I would recommend making main deal only with parsing arguments and printing usage, and have it pass the arguments to a function that does the logic based on its formal parameters. At the moment, it's surprising that you look at `Nargs` further down in this function, for example.

testutils/README.md
```diff
@@ -27,3 +24,7 @@ To run the test, run
   go run testutils/check-calling-form-in-baseline/main.go -yaml src/test/java/com/google/api/codegen/testsrc/library_gapic.yaml src
   
 from the root of the repository.
+
+check-lang-uses-form
+--------------------
+Checks that calling forms defined in `CallingForm.java` are used by the specified languages.
```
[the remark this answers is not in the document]
pongad: Done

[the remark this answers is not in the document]
pongad: Done

testutils/check-baselines.sh
```diff
@@ -28,47 +26,5 @@ Above baseline files contain string "$unhandledStr";
...
-)
-if [ "$unusedForms" ]; then
-	cat 1>&2 <<EOF
-Unused CallingForm:
-$unusedForms
-
-EOF
-	# There are unused calling forms right now, so commenting this off so we don't fail CI.
-	# exitStatus=1
+	exit 1
```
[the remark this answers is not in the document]
pongad: This conditionally `exit 1` depending on whether or not the if-stmt fires. You might be confused by the indentation of the `EOF` line. Unfortunately herestring terminators must begin the line and whitespaces are not allowed.

testutils/check-lang-uses-form/main.go
```diff
@@ -0,0 +1,185 @@
...
+	"io/ioutil"
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+	"unicode/utf8"
+)
+
+const usage = `Reads calling forms from -forms and makes sure each is used by
+the specified languages' *.baseline file in dir.
```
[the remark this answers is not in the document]
pongad: Done

testutils/check-lang-uses-form/main.go
```diff
@@ -0,0 +1,185 @@
...
+	"io"
+	"io/ioutil"
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+	"unicode/utf8"
+)
+
+const usage = `Reads calling forms from -forms and makes sure each is used by
```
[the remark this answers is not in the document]
pongad: Done

testutils/check-lang-uses-form/main.go
```diff
@@ -0,0 +1,185 @@
...
+  CallingForm[,] // used by: lang1 lang2 ...
+
+If the string "// used by" is itself preceeded by another "//" in the same line,
+it is ignored so that it is possible to talk about the comment in the file itself.
+
+Baseline files of the specified languages (in this case "lang1" and "lang2") will be checked
+`
+
+func main() {
+	formFname := flag.String("forms", "", "file containing calling forms")
```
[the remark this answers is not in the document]
pongad: It seems strange to put the default here. I think it makes more sense to let the test command provide it. I can't really articulate why...

testutils/check-lang-uses-form/main.go
```diff
@@ -0,0 +1,185 @@
...
+	flag.Usage = func() {
+		binName := os.Args[0]
+		out := os.Stderr // Should be flag.CommandLine.Output(), but Go version on Travis is too old.
+		fmt.Fprintf(out, "Usage of %s:\n", binName)
```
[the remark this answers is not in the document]
pongad: I was modeling it a bit after `man` pages. On my machine, `man grep` gives

```
NAME
       grep, egrep, fgrep, rgrep - print lines matching a pattern

SYNOPSIS
       grep [OPTIONS] PATTERN [FILE...]
```

So I thought it made sense to include bin name in the "preamble".

testutils/check-lang-uses-form/main.go
```diff
@@ -0,0 +1,185 @@
...
+		fmt.Fprint(out, usage)
+	}
+	flag.Parse()
+
+	if *formFname == "" {
+		flag.Usage()
+		os.Exit(1)
+	}
+
+	formf, err := os.Open(*formFname)
```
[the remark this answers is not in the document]
pongad: Done

testutils/check-lang-uses-form/main.go
pongad: Done

testutils/check-lang-uses-form/main.go
pongad: I thought about it. I remember Jonathan being against doing this so I've been using if-stmt ever since. I don't think it's a big deal either way though.

testutils/check-lang-uses-form/main.go
pongad: Done

testutils/check-lang-uses-form/main.go
pongad: Now that I think about it. We should just use `bufio.Reader`. I have nothing against the `Scanner`, but it's too easy to forget the error check.

testutils/check-lang-uses-form/main.go
pongad: Done

testutils/check-lang-uses-form/main.go
pongad: Done

pongad: Done

testutils/check-lang-uses-form/main.go
pongad: SGTM

testutils/check-lang-uses-form/main.go
pongad: Split `wordChar` but done.

testutils/check-lang-uses-form/main_test.go
pongad: Do you feel strongly about this? Sometimes I do this and find the loop harder to read.

testutils/check-lang-uses-form/main.go
pongad: Done

testutils/check-lang-uses-form/main.go
pongad: I thought about this and decided against it. Philosophically, regexps are useful for matching patterns known at runtime. When the pattern, or at least the format, is simple and known ahead of time, I'd argue that using the regexp is a bit of an overkill.

Technically (though it doesn't really matter here), regexp is usually slow even when implemented properly:
```
BenchmarkRegex-12      	   10000	    148171 ns/op	  81.06 MB/s
BenchmarkHasWord-12    	10000000	       147 ns/op	81468.69 MB/s

```

testutils/check-baselines.sh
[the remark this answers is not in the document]
vchudnov-g: Oh, I see. We're still inside the `if` statement here. Yep, I parsed incorrectly. My bad. Carry on.

(I would have liked the diff to show the `fi`lines here matching, rather than below!)

testutils/check-lang-uses-form/main.go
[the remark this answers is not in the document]
vchudnov-g: But for the foreseeable future, the same value will be provided (except maybe for testing this test), so I think making it the default makes sense.

testutils/check-lang-uses-form/main.go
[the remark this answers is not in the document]
vchudnov-g: I was thinking more of the output when you invoke a command incorrectly:

eg:
```
$ awk
Usage: awk [POSIX or GNU style options] -f progfile [--] file ...
Usage: awk [POSIX or GNU style options] [--] 'program' file ...
...
```

testutils/check-lang-uses-form/main_test.go
vchudnov-g: Don't feel strongly, but IIUC it's the idiomatic way to do it. More compact.

testutils/check-lang-uses-form/main.go
vchudnov-g: This is not performance-critical, and the regex is still sub millisecond, which is fine for what we're doing. In that case, the important criterion becomes clarity. People know regexs and they're easy to read (much more compact) and think about, as opposed to double-checking carefully this custom method.

Something to think about. Not going to block on this.

PR #2051, opened 2018-06-06
go: don't generate empty LRO Metadata()

pongad:
Most methods we generate have signatures like this:

  func (*Client) Method(context.Context, *pb.Request) (*pb.Response, error)

If the response is an empty type, it's not terribly useful, so we drop
it, making the signature:

  func (*Client) Method(context.Context, *pb.Request) error

Since we never write the name of the Empty type,
we have a special logic to never import the "emptypb" package
so we don't get the imported-but-not-used error.

The problem is that if a long-running operation has an Empty metadata,
we still create the

  func (*Operation) Metadata() (*emptypb.Empty, error)

method. This does not compile, because now we're using emptypb without
importing.

The solution is to simply not generating the Metadata() method at all,
since it doesn't contain any useful information anyway.

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/LongRunningTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/LongRunningTransformer.java
@@ -39,6 +39,7 @@ public class LongRunningTransformer {
         .clientReturnTypeName(clientReturnTypeName)
         .operationPayloadTypeName(operationPayloadTypeName)
         .isEmptyOperation(lroConfig.getReturnType().isEmptyType())
+        .isEmptyMetadata(lroConfig.getMetadataType().isEmptyType())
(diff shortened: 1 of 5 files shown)


pongad:
@garrettjonesgoogle @lukesneeringer Do you think it's worth adding a whole new baseline for this? Do other langs have special treatment for empty?

jadekler:
This is technically breaking the API surface of many clients, right? Is there any possible use of metadata at all? Does the error return anything useful - can it be used as a hacky ping or something?

jba:
@jadekler You're right, it is a breaking change, but it's unlikely anyone is using the method because it's useless.

pongad:
In some ways, it's not breaking because the current code won't compile :laughing:

PR #2052, opened 2018-06-07
ToolOptions.get(any_option) always returns a non-null string

andreamlin:
Modify the check on `toolOptions.get(option)` to see if the output is an empty string, because the return value is an empty string (not `null`) for options that have no value.


andreamlin:
PTAL

PR #2053, opened 2018-06-08
New createToolPaths task, add symlink to fatjar

garrettjonesgoogle:
These features allow artman to invoke gradle just twice (once to build gapic-generator, a second time to cache the tool paths), and then avoid gradle completely after that.

### DIFF
--- a/buildSrc/src/main/groovy/DependencyResolver.groovy
+++ b/buildSrc/src/main/groovy/DependencyResolver.groovy
@@ -1,3 +1,17 @@
+/* Copyright 2016 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/groovy/DependencyResolver.groovy
```diff
@@ -1,3 +1,17 @@
+/* Copyright 2016 Google LLC
```
michaelbausor: Fix year and formatting of license (or is that year correct?)

garrettjonesgoogle: This is actually correct - the file was created in 2016, I am backdating the license.

This format matches what we use in the rest of gapic-generator.

PR #2055, opened 2018-06-11
Add PHP type hints for constructor and optional args

michaelbausor:
Add type hints for array arguments in constructor and methods. cc @dwsupplee @jdpedrie

### DIFF
--- a/src/main/resources/com/google/api/codegen/php/client_impl.snip
+++ b/src/main/resources/com/google/api/codegen/php/client_impl.snip
@@ -372,7 +372,7 @@
      * @@throws ValidationException
      * @@experimental
      */
-    public function __construct($options = [])
+    public function __construct(array $options = [])
     {
         $clientOptions = $this->buildClientOptions($options);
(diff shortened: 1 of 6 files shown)


src/main/resources/com/google/api/codegen/php/client_impl.snip
```diff
@@ -503,13 +503,21 @@
...
-            ${@param.name} = {@param.defaultValue}
+            {@paramName(param)} = {@param.defaultValue}
         @else
-            ${@param.name}
+            {@paramName(param)}
         @end
     @end
 @end
 
+@private paramName(param)
```
garrettjonesgoogle: If this can have a type hint in it, it's not technically just a "name" any more, right?

michaelbausor: Updated to `paramDeclaration`, WDYT?

garrettjonesgoogle: That works!

Issue #2016, opened 2018-05-22
Python: routing_header object called as method (actually module)


crwilcox:
While generating IoT I found that there are calls to routing_header that should be modified as follows:

in `google/cloud/iot_v1/gapic/device_manager_client.py`:
```
-        routing_header = google.api_core.gapic_v1.routing_header(	
+        routing_header = google.api_core.gapic_v1.routing_header.to_grpc_metadata(
```

cc @theacodes

PR #2056, opened 2018-06-11
Python: routing_header object called as method (actually module)

andreamlin:
Resolves https://github.com/googleapis/gapic-generator/issues/2016.


andreamlin:
PTAL

PR #2057, opened 2018-06-11
turn on calling form checks for everyone

pongad:
The logic of the check is slightly tweaked to account
for minor differences between langs.


testutils/check-calling-form-in-baseline/main.go
```diff
@@ -175,45 +176,78 @@ type checkConfig struct {
...
-			files = append(files, baseline)
-			break
-		}
-		files = append(files, baseline[:p])
+func deleteFoundForms(r io.Reader, lang string, forms map[checkConfig]bool) error {
+	var callingForm, valueSet string
 
-		baseline = baseline[p:]
-		baseline = baseline[strings.IndexByte(baseline, '\n')+1:]
+	// del deletes the form specified in callingForm, valueSet, and lang from forms and reset.
```
vchudnov-g: "and resets "... what?

testutils/check-calling-form-in-baseline/main.go
```diff
@@ -175,45 +176,78 @@ type checkConfig struct {
...
-		valSet := valueSetRe.FindStringSubmatch(file)
-		if len(valSet) == 0 {
+		// The expected lines look like this:
+		//   // calling form: "form"
+		//   ## valueSet "set" ("title")
+		// The comment character differs between languages.
+		// Some langs also have colons and others don't. Instead of mandating
+		// exact format, just be a little resilient.
+
+		// Trim out commets and whitespaces
```
vchudnov-g: "comments"

testutils/check-calling-form-in-baseline/main.go
```diff
@@ -175,45 +176,78 @@ type checkConfig struct {
...
+		if err == io.EOF {
+			del()
+			return nil
+		}
+		if err != nil {
+			return err
+		}
+
+		if strings.HasPrefix(line, "============== file:") {
+			del()
```
vchudnov-g: It seems to me that it's clearer to call `del()` at the bottom of the `for` loop, rather than setting up all the info needed for the delete but deferring it until the next iteration. This would also get you to have the `del()` in one place rather than two (and incidentally, would allow you to inline it if you choose)

src/test/java/com/google/api/codegen/testsrc/library_gapic.yaml
```diff
@@ -239,12 +239,17 @@ interfaces:
     - series_uuid.series_string=foobar
     sample_value_sets:
       # callingFormCheck: pi_version java: Flattened Request Callable
+      # callingFormCheck: pi_version nodejs: Request
+      # callingFormCheck: pi_version py: Request
     - id: pi_version
       title: "Pi version"
       description: "Testing calling forms: {Java:[Flattened, Request, Callable], NodeJS:[Request], Python:[Request]}"
```
vchudnov-g: Consider also removing the calling form data from the description. We don't need it, keeping it sync willbe in pain, and if it falls out of sync it will be annoying.

testutils/check-calling-form-in-baseline/main.go
pongad: Done

testutils/check-calling-form-in-baseline/main.go
pongad: Hahaha done

src/test/java/com/google/api/codegen/testsrc/library_gapic.yaml
pongad: Let's do this in a separate PR? Enabling checks and removing this are arguably orthogonal.

pongad:
@vchudnov-g PTAL

testutils/check-calling-form-in-baseline/main.go
vchudnov-g: Oh, I see what you mean. That makes sense. It still seems like a bit of unfortunate cognitive load to handle the `del` in multiple places and at the start of the next iteration. I was hoping to differentiate reading the line from processing the file, but that seems to make the code more verbose.....

Maybe add a comment like 
```
// Handle the end of the file currently being processed
```

testutils/check-calling-form-in-baseline/main.go
```diff
@@ -175,45 +176,78 @@ type checkConfig struct {
...
-		}
-		files = append(files, baseline[:p])
+func deleteFoundForms(r io.Reader, lang string, forms map[checkConfig]bool) error {
+	var callingForm, valueSet string
 
-		baseline = baseline[p:]
-		baseline = baseline[strings.IndexByte(baseline, '\n')+1:]
+	// del deletes the form specified in callingForm, valueSet, and lang from forms and reset.
+	// We call this at the end of every logical file.
+	del := func() {
```
vchudnov-g: I suggest s/`del`/`removeForm`/ for clarity

testutils/check-calling-form-in-baseline/main.go
```diff
@@ -175,45 +176,78 @@ type checkConfig struct {
...
-	for {
-		const fileHeader = "============== file:"
-		p := strings.Index(baseline, fileHeader)
-		if p < 0 {
-			files = append(files, baseline)
-			break
-		}
-		files = append(files, baseline[:p])
+func deleteFoundForms(r io.Reader, lang string, forms map[checkConfig]bool) error {
+	var callingForm, valueSet string
```
vchudnov-g: Seems like just having a `var formInstance checkConfig` instead would be more succinct?

src/test/java/com/google/api/codegen/testsrc/library_gapic.yaml
vchudnov-g: OK, doing it in a chaser will make it easier to focus on the current changes in this PR.

testutils/check-calling-form-in-baseline/main.go
pongad: Done

PR #2060, opened 2018-06-11
go: initial setup for nested initializer

pongad:
Go idiomatically uses nested initializer like
```go
&ParentType{
  ChildField: &ChildType{
    X: 42,
  },
  Y: "43",
}
```

The current shared logic that "flattens" initializations does not
support this use case.

If the new logic fails, it prints the error message. The entire output is in a comment,
so this won't interfere with regeneration. The comments are placed in tests, so GoDoc won't display it.

Currently there's two known bugs:
- we don't handle map indices
- we don't handle resource names

Both will be fixed in separate PRs, since this one is already getting big.

The plan: we will slowly add features to nested initializer until it reaches feature-parity with the current init code. Then we'll replace the current flattened init code with nested ones. Gofmt, run as part of code generation pipeline, will handle indentation for us.


src/main/resources/com/google/api/codegen/go/mock.snip
```diff
@@ -185,7 +185,7 @@
         }
         return nil
     }
-@end
+@end
```
garrettjonesgoogle: Revert accidental whitespace?

pongad:
@garrettjonesgoogle Is there a reason you think it won't? I mostly wanted to make sure you're OK with the transformer doing all the initializations.

Issue #2015, opened 2018-05-22
Python: attempt to read nonexistent attr


crwilcox:
In the generated code for IoT, there are two locations (ln 380, 755) in    `google/cloud/iot_v1/gapic/device_manager_client.py` where attempts are made to read non-existent attributes.

```
-        routing_header = google.api_core.gapic_v1.routing_header.to_grpc_metadata(
-            [('device_registry.name', device_registry.name)], )
-        metadata.append(routing_header)
+        if hasattr(device_registry, 'name'):
+            routing_header = google.api_core.gapic_v1.routing_header.to_grpc_metadata(
+                [('device_registry.name', device_registry.name)], )
+            metadata.append(routing_header)
```

https://github.com/GoogleCloudPlatform/google-cloud-python/pull/5355/commits/2a97ee519ee0fd354a870d8e082e409696e6cbd5

cc: @theacodes

theacodes:
Since `device_registry` can also be a dict, this code is really brittle.

Issue #2014, opened 2018-05-22
Python: Attempting to append to None value object


crwilcox:
In the generated code for IoT, there are multiple methods that have `metadata=None` as a default and later we try to append. This will fail. The following code needs to be added to set a better default. See attached commit for more context.
```
+        if metadata is None:
+            metadata = []
+        metadata = list(metadata)
```

https://github.com/GoogleCloudPlatform/google-cloud-python/pull/5355/commits/2a97ee519ee0fd354a870d8e082e409696e6cbd5

cc: @theacodes

andreamlin:
Resolved by https://github.com/googleapis/gapic-generator/pull/2061.

PR #2061, opened 2018-06-12
Catch AttributeError; set None-valued var to empty list before appending

andreamlin:
- Catch AttributeError in case there is an attempt to read non-existent attribute in the params `to_grpc_metadata()`
- if `metadata is None`, set `metadata` to empty list before appending to it
 
See https://github.com/GoogleCloudPlatform/google-cloud-python/pull/5472 as a example of using this PR (on top of the latest gapic-generator:master) to update the IOT python client.

Resolves https://github.com/googleapis/gapic-generator/issues/2015 and https://github.com/googleapis/gapic-generator/issues/2014.

### DIFF
--- a/src/main/resources/com/google/api/codegen/py/main.snip
+++ b/src/main/resources/com/google/api/codegen/py/main.snip
@@ -235,9 +235,6 @@
                         @end
                     @end
 
-                    if metadata is None:
-                        metadata = []
-                    metadata = list(metadata)
                     @if apiMethod.headerRequestParams
                         try:
                             routing_header = google.api_core.gapic_v1.routing_header.to_grpc_metadata(
@@ -251,6 +248,9 @@
                 @end
                 @switch apiMethod.type
                 @case "OptionalArrayMethod"
+                    if metadata is None:
+                        metadata = []
+                    metadata = list(metadata)
                     @if apiMethod.hasReturnValue
                         return self._inner_api_calls['{@apiMethod.name}']({@apiMethod.requestVariableName}{@optionalParams(apiMethod)})
                     @else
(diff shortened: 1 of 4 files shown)


andreamlin:
PTAL

src/main/resources/com/google/api/codegen/py/main.snip
```diff
@@ -235,12 +235,17 @@
                         @end
                     @end
 
+                    if metadata is None:
```
eoogbe: This is only needed when appending the routing headers. Move it to the routing header conditional.

src/main/resources/com/google/api/codegen/py/main.snip
```diff
@@ -235,12 +235,17 @@
...
+                        metadata = []
+                    metadata = list(metadata)
                     @if apiMethod.headerRequestParams
-
-                        routing_header = google.api_core.gapic_v1.routing_header.to_grpc_metadata(
-                            [{@routingHeader(apiMethod.headerRequestParams)}],
-                        )
-                        metadata.append(routing_header)
+                        try:
+                            routing_header = google.api_core.gapic_v1.routing_header.to_grpc_metadata(
```
eoogbe: It's not entirely clear that the error is supposed to catch just the case where the nested params are missing. Move more of the logic to an `else` clause. Something like:
```python
try:
    routing_header = [('name', name), ('book.read', book.read)]
except AttributeError:
    pass
else:
    routing_metadata = google.api_core.gapic_v1.routing_header.to_grpc_metadata(routing_header)
    metadata.append(routing_metadata)
```

andreamlin: definitely

src/main/resources/com/google/api/codegen/py/main.snip
andreamlin: done.

andreamlin:
Thanks for the suggestions! PTAL

Issue #2017, opened 2018-05-22
Python: Update nox.py to use 3.7 for unit tests and system tests


crwilcox:
nox.py still uses 3.4 and doesn't specify 3.7. The generator should be updated to emit `@nox.parametrize('py', ['2.7', '3.5', '3.6', '3.7'])` for unit tests. system tests should be similarly updated to be `@nox.parametrize('py', ['2.7', '3.7'])`

PR #2062, opened 2018-06-12
Update nox python versions

andreamlin:
Resolves https://github.com/googleapis/gapic-generator/issues/2017.

See https://github.com/GoogleCloudPlatform/google-cloud-python/pull/5295/files for the desired output.


andreamlin:
PTAL

PR #2075, opened 2018-06-14
java: update bidi sample to use BidiStream

pongad:
also fix orphan closing-paren and overly-static type name


pongad:
@garrettjonesgoogle PTAL

pongad:
compile-checked!

Issue #2078, opened 2018-06-14
Make STATIC_TYPES the default for resource_name_treatment in configgen


andreamlin:
Make resource_name_treatment: STATIC_TYPES default in config gen.

PR #2083, opened 2018-06-15
Add resource_name_treatment to protobuf-based configgen

andreamlin:
If there are field name patterns, then also set a `resource_name_treatment` config value that defaults to `STATIC_TYPES`.

Resolves https://github.com/googleapis/gapic-generator/issues/2078


andreamlin:
PTAL

PR #2087, opened 2018-06-18
setup output printing

pongad:
Examples need to be able to print specific fields in responses.
This PR sets up the structure to do this.

Currently formatted prints should work.
We can commit this now so that we can parallelize work.

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
@@ -14,15 +14,15 @@
  */
 package com.google.api.codegen.transformer;
 
-import com.google.api.codegen.Accessor;
 import com.google.api.codegen.OutputSpec;
 import com.google.api.codegen.config.FieldModel;
 import com.google.api.codegen.config.MethodModel;
 import com.google.api.codegen.config.TypeModel;
+import com.google.api.codegen.viewmodel.OutputView;
 import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableList;
 import java.util.Collections;
 import java.util.List;
-import java.util.stream.Collectors;
 
 class OutputTransformer {
   private static final String RESPONSE_PLACEHOLDER = "$resp";
@@ -37,56 +37,54 @@ class OutputTransformer {
             .build());
   }
 
-  static OutputSpec toLanguage(OutputSpec config, MethodContext context) {
-    OutputSpec.Builder lang = OutputSpec.newBuilder();
+  static OutputView toView(OutputSpec config, MethodContext context) {
+    OutputView.Builder view = OutputView.newBuilder();
     switch (config.getStmtCase()) {
       case LOOP:
         throw new UnsupportedOperationException("loop not implemented yet");
       case PRINT:
-        lang.setPrint(toLanguage(config.getPrint(), context));
+        view.print(toView(config.getPrint(), context)).kind(OutputView.Kind.PRINT);
         break;
     }
-    return lang.build();
+    return view.build();
   }
 
-  private static OutputSpec.PrintStmt toLanguage(
-      OutputSpec.PrintStmt config, MethodContext context) {
-    return OutputSpec.PrintStmt.newBuilder()
-        .setSpec(context.getNamer().getPrintSpec(config.getSpec()))
-        .addAllArgsElems(
+  private static OutputView.PrintView toView(OutputSpec.PrintStmt config, MethodContext context) {
+    return OutputView.PrintView.newBuilder()
+        .printSpec(context.getNamer().getPrintSpec(config.getSpec()))
+        .printArgs(
             config
                 .getArgsList()
                 .stream()
                 .map(a -> accessor(a, context))
-                .collect(Collectors.toList()))
+                .collect(ImmutableList.toImmutableList()))
         .build();
   }
 
-  private static Accessor accessor(String config, MethodContext context) {
+  private static OutputView.PrintArgView accessor(String config, MethodContext context) {
     String[] configElems = config.split("\\.");
-    if (configElems.length == 0) {
-      return Accessor.getDefaultInstance();
-    }
+    Preconditions.checkArgument(configElems.length != 0, "field string cannot be empty");
 
-    Accessor.Builder accessor = Accessor.newBuilder();
+    OutputView.PrintArgView.Builder view = OutputView.PrintArgView.newBuilder();
     TypeModel type;
     if (configElems[0].equals(RESPONSE_PLACEHOLDER)) {
-      accessor.setVariable(context.getNamer().getSampleResponseVarName());
+      view.variable(context.getNamer().getSampleResponseVarName());
       type = context.getMethodModel().getOutputType();
     } else {
       throw new UnsupportedOperationException("local variable not implemented yet");
     }
 
+    ImmutableList.Builder<String> accessors = ImmutableList.builder();
     for (int i = 1; i < configElems.length; i++) {
       String fieldName = configElems[i];
       FieldModel field =
           Preconditions.checkNotNull(
-              type.getField(configElems[i]), "type %s does not have field %s", type, fieldName);
+              type.getField(fieldName), "type %s does not have field %s", type, fieldName);
       Preconditions.checkArgument(
           !field.isRepeated() && !field.isMap(), "%s.%s is not scalar", type, fieldName);
       type = field.getType();
-      accessor.addFields(context.getNamer().getFieldGetFunctionName(field));
+      accessors.add(context.getNamer().getFieldGetFunctionName(field));
     }
-    return accessor.build();
+    return view.accessors(accessors.build()).build();
   }
 }
--- a/src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
@@ -1,0 +1,92 @@
+/* Copyright 2018 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen.transformer;
+
+import com.google.api.codegen.Accessor;
+import com.google.api.codegen.OutputSpec;
+import com.google.api.codegen.config.FieldModel;
+import com.google.api.codegen.config.MethodModel;
+import com.google.api.codegen.config.TypeModel;
+import com.google.common.base.Preconditions;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+
+class OutputTransformer {
+  private static final String RESPONSE_PLACEHOLDER = "$resp";
+
+  static List<OutputSpec> defaultOutputSpecs(MethodModel method) {
+    if (method.isOutputTypeEmpty()) {
+      return Collections.emptyList();
+    }
+    return Collections.singletonList(
+        OutputSpec.newBuilder()
+            .setPrint(OutputSpec.PrintStmt.newBuilder().setSpec("%s").addArgs(RESPONSE_PLACEHOLDER))
+            .build());
+  }
+
+  static OutputSpec toLanguage(OutputSpec config, MethodContext context) {
+    OutputSpec.Builder lang = OutputSpec.newBuilder();
+    switch (config.getStmtCase()) {
+      case LOOP:
+        throw new UnsupportedOperationException("loop not implemented yet");
+      case PRINT:
+        lang.setPrint(toLanguage(config.getPrint(), context));
+        break;
+    }
+    return lang.build();
+  }
+
+  private static OutputSpec.PrintStmt toLanguage(
+      OutputSpec.PrintStmt config, MethodContext context) {
+    return OutputSpec.PrintStmt.newBuilder()
+        .setSpec(context.getNamer().getPrintSpec(config.getSpec()))
+        .addAllArgsElems(
+            config
+                .getArgsList()
+                .stream()
+                .map(a -> accessor(a, context))
+                .collect(Collectors.toList()))
+        .build();
+  }
+
+  private static Accessor accessor(String config, MethodContext context) {
+    String[] configElems = config.split("\\.");
+    if (configElems.length == 0) {
+      return Accessor.getDefaultInstance();
+    }
+
+    Accessor.Builder accessor = Accessor.newBuilder();
+    TypeModel type;
+    if (configElems[0].equals(RESPONSE_PLACEHOLDER)) {
+      accessor.setVariable(context.getNamer().getSampleResponseVarName());
+      type = context.getMethodModel().getOutputType();
+    } else {
+      throw new UnsupportedOperationException("local variable not implemented yet");
+    }
+
+    for (int i = 1; i < configElems.length; i++) {
+      String fieldName = configElems[i];
+      FieldModel field =
+          Preconditions.checkNotNull(
+              type.getField(configElems[i]), "type %s does not have field %s", type, fieldName);
+      Preconditions.checkArgument(
+          !field.isRepeated() && !field.isMap(), "%s.%s is not scalar", type, fieldName);
+      type = field.getType();
+      accessor.addFields(context.getNamer().getFieldGetFunctionName(field));
+    }
+    return accessor.build();
+  }
+}
--- a/src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
@@ -32,29 +32,42 @@ class OutputTransformer {
       return Collections.emptyList();
     }
     return Collections.singletonList(
-        OutputSpec.newBuilder()
-            .setPrint(OutputSpec.PrintStmt.newBuilder().setSpec("%s").addArgs(RESPONSE_PLACEHOLDER))
-            .build());
+        OutputSpec.newBuilder().addPrint("%s").addPrint(RESPONSE_PLACEHOLDER).build());
   }
 
   static OutputView toView(OutputSpec config, MethodContext context) {
+    Runnable once =
+        new Runnable() {
+          boolean ran;
+
+          @Override
+          public void run() {
+            if (ran) {
+              throw new IllegalArgumentException("only one field of OutputSpec may be set");
+            }
+            ran = true;
+          }
+        };
+
     OutputView.Builder view = OutputView.newBuilder();
-    switch (config.getStmtCase()) {
-      case LOOP:
-        throw new UnsupportedOperationException("loop not implemented yet");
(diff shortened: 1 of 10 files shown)


src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -656,3 +658,28 @@ message LongRunningConfigProto {
...
 }
+
+message OutputSpec {
+  message LoopStmt {
+    string collection = 1;
+    string var = 2;
+    repeated OutputSpec body = 3;
+  }
+  message PrintStmt {
+    string spec = 1;
```
vchudnov-g: What if we made the yaml more compact, like this:
```
- print:
  - "Person %s lives at %s\n"
  - response.name
  - response.address
```
which I guess means that `OutputSpec.print` should then be just a `repeated string`. I guess this would mean then an enum instead of a oneof to distinguish the element types.

(Originally I was going to suggest a `repeated string` inside the `PrintStmt`, but IIUC that would still be more verbose in the YAML than what I suggest above)

src/main/java/com/google/api/codegen/viewmodel/OutputView.java
```diff
@@ -0,0 +1,85 @@
...
+
+    public abstract Builder print(PrintView val);
+
+    public abstract OutputView build();
+  }
+
+  @AutoValue
+  public abstract static class PrintView {
+
+    public abstract String printSpec();
```
vchudnov-g: What if instead we only had `args`, where `args[0]` is the format string if `len(args)>1`? This makes it simpler in the .snip to concatenate all the arguments without having the empty `{}` format string in your current baseline.

In this case, if `len(args)==0`, the transformer could set `args[0]` to `"\n"`

(This ties in nicely, I think, with my other comment below.)

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -345,6 +345,8 @@ message SampleValueSet {
   //   a specially formatted string using `FIELD` as one of possibly
   //   several placeholders.
   repeated string parameters = 4;
+
+  repeated OutputSpec outputs = 5;
```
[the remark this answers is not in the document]
pongad: I think we should follow the convention that `repeated` fields have plural names.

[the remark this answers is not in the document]
vchudnov-g: Then I wonder whether we should use a different name, like `response_cmds` (but shorter than that). I'm thinking that it makes the yaml a bit misleading to have an `outputs` section when some of the commands are not `print`s:

```
- outputs:
  - print
    - "foo %s"
    - "$resp.name"
  - loop
    var: foo
    iterable: my_collection
    body:
    - print
      - "%s\n"
      - foo
```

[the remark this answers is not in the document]
lukesneeringer: FWIW: The API style guide would ask for `repeated OutputSpec output_specs`.

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -656,3 +658,28 @@ message LongRunningConfigProto {
   // Total polling timeout.
   uint64 total_poll_timeout_millis = 8;
 }
+
+message OutputSpec {
+  message LoopStmt {
+    string collection = 1;
+    string var = 2;
```
lukesneeringer: `var` is a reserved word in JavaScript; it is probably unwise to use it as a field name.

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -656,3 +658,28 @@ message LongRunningConfigProto {
   // Total polling timeout.
   uint64 total_poll_timeout_millis = 8;
 }
+
+message OutputSpec {
+  message LoopStmt {
```
lukesneeringer: Can we spell out `Statement`? (Here and for `PrintStmt` below.)

src/main/proto/com/google/api/codegen/config.proto
lukesneeringer: Also, we use `string variable` later in this file. :-)

vchudnov-g: Good point.

Using the same term both here and in `Accessor` makes sense, since they both reference a local variable. (And using it eventually in the `AliasStatement` as well for the same reason.) I'm fine with the term `variable`.

src/main/proto/com/google/api/codegen/config.proto

src/main/java/com/google/api/codegen/viewmodel/OutputView.java
pongad: Synced offline. We shouldn't do this since langs like Python does `"formatspec".format(...)`

src/main/proto/com/google/api/codegen/config.proto

src/main/proto/com/google/api/codegen/config.proto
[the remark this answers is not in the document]

pongad:
@vchudnov-g @lukesneeringer PTAL

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -656,3 +658,20 @@ message LongRunningConfigProto {
...
+  }
+
+  // Exactly one of the following should be set.
+
+  LoopStatement loop = 1;
+
+  // example ["%s=%s", thing.variable, thing.value]
+  // field name is singular because the array
+  // represents one print statement.
+  repeated string print = 2;
```
lukesneeringer: `print` is a reserved word in numerous languages. It is also a verb; proto fields should generally be nouns.

I can not actually determine what is intended -- does it represent a print _statement_ or does it represent the _string to be printed_?

vchudnov-g: @lukesneeringer See my previous comment. I'm the one who suggested the repeated string `print` because it would make the input yaml more readable, comprehensible, and compact, like so:

```
- print:
  - "Person %s lives at %s\n"
  - response.name
  - response.address
```

After all, the output_spec field is a description of the code to output, so I'd like it be as easy for humans as possible. This seems very readable, as it's the name of the statement (indeed, a verb and a common reserved word: that is on purpose) and then the list args that get used in printf-style functions in many languages.

Let me also say that generally I favor good encapsulation and reusable messages so my first inclination (eg in the design doc) was indeed to to have a `PrintStatement`. But since a big part of the purpose of this proto is to generate the yaml schema used by API teams to define their samples, I deem that the usability gains are worth it in this case. I'm open to convincing counterarguments.

src/main/proto/com/google/api/codegen/config.proto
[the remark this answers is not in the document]
vchudnov-g: (tangent for my own curiosity)

@lukesneeringer What does the style guide rule you cite imply for re-using messages? Imagine that we want to have an identical code specification for tests. We would have to have whatever field holds the spec be called `output_specs`? By analogy to types and variables in a programming language, I would like to be able to do `repeated OutputSpec sample_specs` and `repeated OutputSpec test_specs`. But it sounds like instead we need to maybe rename the type and fields and have the field names be identical within separate messages?

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -0,0 +1,103 @@
...
+
+  private static OutputView.PrintView toView(List<String> config, MethodContext context) {
+    Preconditions.checkArgument(!config.isEmpty(), "print spec cannot be empty");
+    return OutputView.PrintView.newBuilder()
+        .printSpec(context.getNamer().getPrintSpec(config.get(0)))
+        .printArgs(
+            config
+                .subList(1, config.size())
+                .stream()
+                .map(a -> accessor(a, context))
```
vchudnov-g: It would be helpful to our users if we could also check that the number of placeholders in the input string matches the number of accessors provided.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -0,0 +1,103 @@
...
+      once.run();
+      view.print(toView(config.getPrintList(), context)).kind(OutputView.Kind.PRINT);
+    }
+    return view.build();
+  }
+
+  private static OutputView.PrintView toView(List<String> config, MethodContext context) {
+    Preconditions.checkArgument(!config.isEmpty(), "print spec cannot be empty");
+    return OutputView.PrintView.newBuilder()
+        .printSpec(context.getNamer().getPrintSpec(config.get(0)))
```
vchudnov-g: Note for the future: as we expand to additional languages, we should check that the `%s` format specifier works for the corresponding value. If not, we may need language-specific substitutions. For example, maybe in Go `%v` is more appropriate than `%s`. I'm not sure about other languages, whether `%s` always prints ints, floats, bools correctly.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    ImmutableList.Builder<String> accessors = ImmutableList.builder();
+    for (int i = 1; i < configElems.length; i++) {
+      String fieldName = configElems[i];
+      FieldModel field =
+          Preconditions.checkNotNull(
+              type.getField(fieldName), "type %s does not have field %s", type, fieldName);
+      Preconditions.checkArgument(
+          !field.isRepeated() && !field.isMap(), "%s.%s is not scalar", type, fieldName);
```
vchudnov-g: We should support indexing here, so that if `fieldName` is of the form `"foo[bar]"`, then we `getField("foo")` and make sure that `foo` is indexable. We can do this in a subsequent PR, but we'll likely want to do it ASAP since many of our reponses (eg in the library service) have repeated fields.

[This also means that `bar` itself could be a dotted/subscripted accessor, so the simple split on `.` you have above  will need to get more sophisticated. I'm fine if we defer this, restricting indices to only be literals for now.]

vchudnov-g: The first substitution (`type`) is giving me `{[Protobuf FieldModel: {repeated string google.example.library.v1.FindRelatedBooksResponse.names}, Protobuf FieldModel: {string google.example.library.v1.FindRelatedBooksResponse.next_page_token}]}` instead of simply `FindRelatedBooksResponse`. Anyway to fix this? (This will affect the the other precondition check, since this one will go away as per the other comment on this line.)

vchudnov-g: We should allow the terminal segment of `configElems` to not be a scalar. This is to allow printing the whole iterable in the language-idiomatic way. This will also be needed for loop when we implement them.

src/main/java/com/google/api/codegen/transformer/py/PythonSurfaceNamer.java
```diff
@@ -453,4 +453,27 @@ public String getReleaseAnnotation(ReleaseLevel releaseLevel) {
...
+        return sb.append(spec, cursor, spec.length()).toString();
+      }
+      sb.append(spec, cursor, p);
+
+      if (spec.startsWith("%%", p)) {
+        sb.append('%');
+      } else if (spec.startsWith("%s", p)) {
+        sb.append("{}");
+      } else {
+        throw new IllegalArgumentException(String.format("bad format verb: %s", spec));
```
vchudnov-g: Something to think about for later: do we want to also support `%q` in the Go sense, so that it maps (for Python) to `sb.append("\"{}\"")`?

src/main/java/com/google/api/codegen/viewmodel/OutputView.java
```diff
@@ -0,0 +1,85 @@
...
+@AutoValue
+public abstract class OutputView {
+
+  public enum Kind {
+    PRINT
+  }
+
+  public abstract Kind kind();
+
+  public abstract PrintView print();
```
vchudnov-g: Are you thinking of doing essentially a oneof pattern here as we add more `Kind`s? How about making `OutputView` abstract and making `PrintView`, `LoopView`, etc be subclasses? That way each `OutputView` is guaranteed to be exactly one of those, and the`MethodSampleView` has the top-level collection of such `OutputView`s.

src/main/java/com/google/api/codegen/transformer/py/PythonSurfaceNamer.java
```diff
@@ -453,4 +453,27 @@ public String getReleaseAnnotation(ReleaseLevel releaseLevel) {
         throw new IllegalStateException("Invalid development status");
     }
   }
+
+  @Override
+  public String getPrintSpec(String spec) {
+    // Escaers don't work here. They only map from characters to strings.
```
vchudnov-g: "Escapers"? I assume they're previously existing classes. Maybe name the package so it's clearer what you're referring to.

src/main/java/com/google/api/codegen/transformer/py/PythonSurfaceNamer.java
```diff
@@ -453,4 +453,27 @@ public String getReleaseAnnotation(ReleaseLevel releaseLevel) {
         throw new IllegalStateException("Invalid development status");
     }
   }
+
+  @Override
+  public String getPrintSpec(String spec) {
```
vchudnov-g: I'm finding that `\n` in a quoted string in the YAML input does not show up in the output. I have to double-escape in the YAML `\\n` to get it to show up as `\n` in the output. It would be great if we don't have to double-escape in the YAML. Do you know how we can achieve that?

vchudnov-g: Oh, that was with double-quoted strings. Single-quoted strings appear to pass escape sequences without double escaping. Sweet.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+    if (config.getPrintCount() > 0) {
+      once.run();
+      view.print(toView(config.getPrintList(), context)).kind(OutputView.Kind.PRINT);
+    }
+    return view.build();
+  }
+
+  private static OutputView.PrintView toView(List<String> config, MethodContext context) {
+    Preconditions.checkArgument(!config.isEmpty(), "print spec cannot be empty");
```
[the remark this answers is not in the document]

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
pongad: I thought of ways to do this and decided to write a general algorithm for it. Let's do this in a separate PR.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
pongad: `%s` will not print floats and bools correctly in Go. `%s` is mean to be a placeholder; in Go, we'll translate `%s` into `%v` like you suggested.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
pongad: Definitely let's do this in a separate PR

src/main/java/com/google/api/codegen/transformer/py/PythonSurfaceNamer.java
pongad: Done

src/main/java/com/google/api/codegen/viewmodel/OutputView.java
vchudnov-g: No, we can  still have a `kind()` method in`OutputView` that gets overriden by subclasses. Wouldn't that make the `.snip` side of things just as easy as it is now?

src/main/java/com/google/api/codegen/transformer/py/PythonSurfaceNamer.java
vchudnov-g: :+1:   Agree on the P2 (hence the "later")

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
vchudnov-g: Agreed. I'm mostly concerned about languages where different types take different placeholders, but we'll deal with that when we get there, as we discussed. It shouldn't be too hard.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
vchudnov-g: I think this could be just a couple of lines in this file, no? Just check the two counts.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -56,27 +59,38 @@ public void run() {
...
                 .subList(1, config.size())
                 .stream()
-                .map(a -> accessor(a, context))
+                .map(a -> accessor(a, context, valueSet))
                 .collect(ImmutableList.toImmutableList()))
         .build();
   }
 
-  private static OutputView.PrintArgView accessor(String config, MethodContext context) {
+  private static OutputView.PrintArgView accessor(
```
vchudnov-g: minor nit: I personally would find the name "fieldAccessor" more illuminating.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -35,16 +36,18 @@
...
     Runnable once =
         new Runnable() {
           boolean ran;
 
           @Override
           public void run() {
-            if (ran) {
-              throw new IllegalArgumentException("only one field of OutputSpec may be set");
-            }
+            Preconditions.checkArgument(
```
vchudnov-g: We should test this. I'm fine with adding a `TODO(pongad)` for now if we're moving quickly for parallel development and things may change, but I don't want this to be untested for too long.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -56,27 +59,38 @@ public void run() {
...
   }
 
-  private static OutputView.PrintArgView accessor(String config, MethodContext context) {
+  private static OutputView.PrintArgView accessor(
+      String config, MethodContext context, SampleValueSet valueSet) {
     String[] configElems = config.split("\\.");
-    Preconditions.checkArgument(configElems.length != 0, "field string cannot be empty");
+    Preconditions.checkArgument(
+        configElems.length != 0,
+        "%s:%s: field string cannot be empty",
```
vchudnov-g: We should test this. I'm fine with adding a `TODO(pongad)` for now

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -92,9 +106,19 @@ public void run() {
       String fieldName = configElems[i];
       FieldModel field =
           Preconditions.checkNotNull(
-              type.getField(fieldName), "type %s does not have field %s", type, fieldName);
+              type.getField(fieldName),
```
vchudnov-g: We should test this. I'm fine with adding a `TODO(pongad)` for now

src/main/java/com/google/api/codegen/transformer/py/PythonSurfaceNamer.java
vchudnov-g: We should test this. I'm fine with adding a `TODO(pongad)` for now

src/main/java/com/google/api/codegen/viewmodel/OutputView.java
pongad: Done

src/main/java/com/google/api/codegen/transformer/py/PythonSurfaceNamer.java
pongad: I'll just add it in the todo table in the doc.

PR #2101, opened 2018-06-22
samples: allow printing arrays

pongad:


src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -88,44 +88,78 @@ public void run() {
...
+    while (end < config.length()) {
+      if (config.charAt(end) == '.') {
+        Preconditions.checkArgument(
+            !type.isRepeated() && !type.isMap(),
+            "%s:%s: %s is not scalar",
+            context.getMethodModel().getSimpleName(),
+            valueSet.getId(),
+            config.substring(0, end));
+
+        cursor = end + 1;
```
vchudnov-g: It would be good if you could factor out this stanza, which is used identically above (but with one more matching character, `$`). It would make the code at this level more compact and easier to follow.

PR #2107, opened 2018-06-22
Samplegen/py: Tweak the PrintView output

vchudnov-g:
- print a response if no OutputView is specified or if an empty string provided in the PrintView
  (there seems to be no way to provide an empty format string via the yaml)
- print just the format string if no arguments provided

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
@@ -33,7 +33,10 @@ class OutputTransformer {
       return Collections.emptyList();
     }
     return Collections.singletonList(
-        OutputSpec.newBuilder().addPrint("%s").addPrint(RESPONSE_PLACEHOLDER).build());
+        OutputSpec.newBuilder()
+            .addPrint("") // empty string means just the first placeholder, unformatted
+            .addPrint(RESPONSE_PLACEHOLDER)
+            .build());
   }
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -33,7 +33,10 @@
       return Collections.emptyList();
     }
     return Collections.singletonList(
-        OutputSpec.newBuilder().addPrint("%s").addPrint(RESPONSE_PLACEHOLDER).build());
+        OutputSpec.newBuilder()
```
pongad: I'm confused what this is doing. Is it so that we don't create any output here, and use the template logic to write `print(response)`. In that case, I think we should delete this method altogether.

vchudnov-g: Not quite. The result of this, with the template I have, is that we get a `print(response)` rather than `print('{}'.format(response))`. The former is more idiomatic if we're not formatting.

vchudnov-g: (BTW, this only works if no `OutputSpec` is specified at all, which is when we call this. It seems we can't pass an empty string explicitly as the first `print` list element; the yaml-to-proto process seems to barf.)

vchudnov-g: As discussed offline, changed so that the view model still corresponds to `%s`. The logic for making the output more idiomatic will be contained in the `.snip` file.

PR #2108, opened 2018-06-22
Use pandoc for Python doc comment reformatting.

### LINKED ISSUE #1587: Move Markdown -> RST conversion to pandoc
Right now, toolkit converts documentation from protos to the `*_client.py` files using some internal mechanism.

It should use pandoc (like protoc-docs-plugin does), so we are not having to roll our own. This would avoid errors like those encountered here: https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4030/commits/6efa645e253754a88eee923480680c28fd34dc0e

eoogbe:
Fixes #1587 

Java has no pandoc library that accepts strings, so the pandoc command is used directly with `InputStream`s and `OutputStream`s.

Formatting is stricter than most client libraries in googleapis might expect, and more closely follows strict markdown spec. (Though pandoc markdown is used.)

### DIFF
--- a/src/main/java/com/google/api/codegen/util/py/PythonCommentReformatter.java
+++ b/src/main/java/com/google/api/codegen/util/py/PythonCommentReformatter.java
@@ -14,54 +14,69 @@
  */
 package com.google.api.codegen.util.py;
 
-import com.google.api.codegen.util.CommentPatterns;
 import com.google.api.codegen.util.CommentReformatter;
 import com.google.api.codegen.util.CommentTransformer;
 import com.google.api.codegen.util.LinkPattern;
-import com.google.common.base.Splitter;
+import com.google.common.io.CharStreams;
+import java.io.BufferedWriter;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.OutputStream;
+import java.io.OutputStreamWriter;
+import java.io.Reader;
+import java.io.UncheckedIOException;
+import java.io.Writer;
 
+/** Reformats a proto markdown comment to the Python reST format. */
 public class PythonCommentReformatter implements CommentReformatter {
 
-  private CommentTransformer transformer =
+  private static final CommentTransformer TRANSFORMER =
       CommentTransformer.newBuilder()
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/api/codegen/util/py/PythonCommentReformatter.java
```diff
@@ -14,54 +14,69 @@
...
-    for (String line : lines) {
-      if (inCodeBlock) {
-        // Code blocks are either empty or indented
-        if (!(line.trim().isEmpty()
-            || CommentPatterns.CODE_BLOCK_PATTERN.matcher(line).matches())) {
-          inCodeBlock = false;
-          line = transformer.transform(line);
-        }
+    String transformedComment = TRANSFORMER.transform(comment);
+    ProcessBuilder builder = new ProcessBuilder("pandoc", "-f", "markdown", "-t", "rst");
```
garrettjonesgoogle: Why aren't we running pandoc from artman?

eoogbe: It seems expecting pandoc in gapic-generator breaks the tests. I'm not sure where in artman the docs should be converted. Do I need to write a separate plugin to transform the docs?

garrettjonesgoogle: `artman` doesn't really use the term "plugin". You would probably create an artman "task" and add it to the list of Gapic tasks to run in https://github.com/googleapis/artman/blob/master/artman/pipelines/gapic_generation.py#L166 (in a language-aware way).

eoogbe: I was referring to a protoc plugin that I would hook into artman as a language-aware task. Trying to find a place for the doc conversion.

Gapic generator seemed like a good place to hook in changes to the comments because it's processing the results of api-compiler. However, pandoc doesn't work for it.

Artman has pandoc, but I'm not sure whether I'd convert the comments before the descriptor generation, between the descriptor generation and the gapic generation, or between the gapic generation and the formatting. They all seem like strange places to hook up the logic.

PR #2112, opened 2018-06-25
sample: implement loops

pongad:


src/main/resources/com/google/api/codegen/py/standalone_sample.snip
```diff
@@ -84,8 +84,11 @@
         {@processLoopView(view)}
       @case "PRINT"
         {@processPrintView(view)}
+      @case "LOOP"
```
vchudnov-g: I think this case got merged in. It should be removed, since we handle it above.

src/test/java/com/google/api/codegen/testsrc/library_gapic.yaml
```diff
@@ -248,9 +248,13 @@ interfaces:
       - shelf.name=Math
       - series_uuid.series_string=xyz3141592654
       output_specs:
-      - print:
-        - "book names %s"
-        - $resp.book_names
+      - loop:
+          variable: book_name
```
vchudnov-g: itty-nitty suggestion: we could do `title` here for concision and readability, but no biggie either way.

src/test/java/com/google/api/codegen/testsrc/library_gapic.yaml
```diff
@@ -248,9 +248,13 @@ interfaces:
...
+          collection: $resp.book_names
+          body:
+            print:
+              - "book name: %s"
```
vchudnov-g: nit: Here I would have text that's slightly friendlier, like `"Title: %s"` or `"Book: %s"`

src/main/resources/com/google/api/codegen/py/standalone_sample.snip
```diff
@@ -108,7 +111,7 @@
 @end
 
 @private processLoopView(view)
-  for ({@view.variable} in {@variableField(view.collection)}):
+  for {@view.variableName} in {@variableField(view.collection)}:
```
vchudnov-g: in `processAssignmentView`, could you also make s/`variable`/`variableName`/? My mistake!

Issue #1730, opened 2017-12-01
Artman code generation: emitted test code doesn't build


kmjung:
(copied from https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2673)

I use artman to generate Java client libraries for my gRPC-based service. Specifically, I'm using the googleapis and toolkit versions from this commit: googleapis/artman@d065fbb.

The emitted production code builds and runs, but the test code doesn't compile as it's emitted. Two fixes are required to make it build:

1. Gradle dependencies

The following Gradle dependencies are emitted:
```
dependencies {
  compile project(":grpc-myproject"),
    'com.google.api:gax:1.15.0',
    'com.google.api:gax-grpc:1.15.0',
    'commons-cli:commons-cli:1.4',
    'commons-lang:commons-lang:2.6',
    // This dependency needs to be update-to-date with the version that gRPC expects.
    'io.netty:netty-tcnative-boringssl-static:1.1.33.Fork26'
}
```
In order to get the test code to build, I have to add dependencies on JUnit and on the test configuration of the gax-grpc library:
```
dependencies {
  testCompile 'junit:junit:4.11',
    'com.google.api:gax-grpc:1.15.0:testlib'
}
```
2. Mock service imports

The test code emits a mock service .java class which imports the service implementation base class from the wrong package. Specifically:
```
import mypackage.PackageName.MyPackageServiceGrpc.MyPackageServiceImplBase;
```
... has to be changed to ...
```
import mypackage.MyPackageServiceGrpc.MyPackageServiceImplBase;
```
... in order to get the test code to compile.

andreamlin:
I am able to build and run generated unit tests from artman-generated code on the Pubsub API.
Some of the dependencies that had been emitted when this issue was filed are now present, e.g. the testCompile for gax-grpc:testlib and junit.

@kmjung is this still a problem? If so, for which API?

kmjung:
@andreamlin I haven't rebuilt the client code here in months. Adding @mmladenovski who has looked at this more recently.

mmladenovski:
@andreamlin , the issue #2 (described above) still exists (I was able to reproduce it today).

I can share more details offline.

andreamlin:
@mmladenovski what's a concrete example of the second issue?

mmladenovski:
Followed up offline with @andreamlin

andreamlin:
[...]

PR #2113, opened 2018-06-26
Account for java_multiple_baselines=false in naming grpc elements

andreamlin:
Fixes https://github.com/googleapis/gapic-generator/issues/1730.

When `java_multiple_files = false` in the proto API definition, the generated proto files are put under a single outer file. However, the generated grpc class names are *not* put under a single outer file.

The gapic-generator should also match the grpc import paths by not putting the grpc classes under an outer Service file. In the baseline tests, `TaggerProto` is the service file in which proto classes are put, but grpc classes should not be put.


andreamlin:
PTAL

PR #2114, opened 2018-06-26
make java_multiple_files false

andreamlin:
Change test resources and baseline so that PR #2113's baseline diffs make more sense.


andreamlin:
PTAL

PR #2115, opened 2018-06-26
Support flattenings on streaming methods

chrisdunelm:
Fixes https://github.com/GoogleCloudPlatform/google-cloud-dotnet/issues/1791


src/test/java/com/google/api/codegen/testdata/csharp/csharp_library.baseline
```diff
@@ -8924,6 +8924,28 @@ namespace Google.Example.Library.V1
...
+        /// <param name="name">
+        /// The name of the shelf whose books we'd like to list.
+        /// </param>
+        /// <param name="callSettings">
+        /// If not null, applies overrides to this RPC call.
+        /// </param>
+        /// <returns>
+        /// The server stream.
+        /// </returns>
+        public virtual Book StreamBooks(
```
jskeet: Book seems to be an odd return type for a stream-based method. Is this really the intended behavior? Not some kind of stream?

chrisdunelm: Ah drat, thanks for noticing this, it is indeed wrong.
I'll fix it now.

chrisdunelm:
@jskeet PTAL

PR #2119, opened 2018-06-27
samplegen: implement alias, use in Python

pongad:
The specification is slightly different from the one in the design doc,
being a string instead of a dictionary. I believe this is more
intuitive.

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
@@ -25,8 +25,11 @@ import com.google.api.codegen.viewmodel.OutputView.VariableView;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import javax.annotation.Nullable;
 
 class OutputTransformer {
@@ -40,11 +43,17 @@ class OutputTransformer {
         OutputSpec.newBuilder().addPrint("%s").addPrint(RESPONSE_PLACEHOLDER).build());
   }
 
-  static OutputView toView(
-      OutputSpec config,
-      MethodContext context,
-      SampleValueSet valueSet,
-      Map<String, TypeModel> localVars) {
+  static List<OutputView> toViews(
+      List<OutputSpec> configs, MethodContext context, SampleValueSet valueSet) {
+    ScopeTable localVars = new ScopeTable();
+    return configs
+        .stream()
+        .map(s -> OutputTransformer.toView(s, context, valueSet, localVars))
+        .collect(ImmutableList.toImmutableList());
+  }
+
+  private static OutputView toView(
+      OutputSpec config, MethodContext context, SampleValueSet valueSet, ScopeTable localVars) {
     Runnable once =
         new Runnable() {
           boolean ran;
@@ -69,9 +78,9 @@ class OutputTransformer {
       once.run();
       view = printView(config.getPrintList(), context, valueSet, localVars);
     }
-    if (!config.getAssign().isEmpty()) {
+    if (!config.getDefine().isEmpty()) {
       once.run();
-      view = assignmentView(config.getAssign(), context, valueSet, localVars);
+      view = definitionView(config.getDefine(), context, valueSet, localVars);
     }
 
     return Preconditions.checkNotNull(
@@ -82,10 +91,7 @@ class OutputTransformer {
   }
 
   private static OutputView.PrintView printView(
-      List<String> config,
-      MethodContext context,
-      SampleValueSet valueSet,
-      Map<String, TypeModel> localVars) {
+      List<String> config, MethodContext context, SampleValueSet valueSet, ScopeTable localVars) {
     Preconditions.checkArgument(
         !config.isEmpty(),
         "%s:%s: print spec cannot be empty",
@@ -98,9 +104,14 @@ class OutputTransformer {
             config
                 .subList(1, config.size())
                 .stream()
-                .map(a -> accessor(a, context, valueSet, localVars, null))
+                .map(a -> accessor(a, context, valueSet, localVars))
                 .collect(ImmutableList.toImmutableList()))
         .build();
+  }
+
+  private static OutputView.VariableView accessor(
+      String config, MethodContext context, SampleValueSet valueSet, ScopeTable localVars) {
+    return accessorNewVariable(config, context, valueSet, localVars, null);
   }
 
   // accessor:
@@ -115,17 +126,17 @@ class OutputTransformer {
    * config refers to a local variable, the variable is looked up in {@code localVars}. If {@code
    * newVar} is not null, it is registered into {@code localVars}.
    */
-  private static OutputView.VariableView accessor(
+  private static OutputView.VariableView accessorNewVariable(
       String config,
       MethodContext context,
       SampleValueSet valueSet,
-      Map<String, TypeModel> localVars,
+      ScopeTable localVars,
       @Nullable String newVar) {
 
     OutputView.VariableView.Builder view = OutputView.VariableView.newBuilder();
 
     int cursor = 0;
-    int end = identifier(config, cursor);
+    int end = identifierEnd(config, cursor);
 
     String baseIdentifier = config.substring(cursor, end);
     TypeModel type;
@@ -154,7 +165,7 @@ class OutputTransformer {
             config.substring(0, end));
 
         cursor = end + 1;
-        end = identifier(config, cursor);
+        end = identifierEnd(config, cursor);
 
         String fieldName = config.substring(cursor, end);
         FieldModel field =
@@ -183,7 +194,7 @@ class OutputTransformer {
     }
 
     if (newVar != null) {
-      if (localVars.putIfAbsent(newVar, type) != null) {
+      if (!localVars.put(newVar, type)) {
         throw new IllegalStateException(
             String.format(
                 "%s:%s: duplicated variable declaration not allowed: %s",
@@ -195,7 +206,7 @@ class OutputTransformer {
   }
 
   /** Returns the largest p such that s.substring(startsFrom, p) is an identifier. */
-  private static int identifier(String s, int startFrom) {
+  private static int identifierEnd(String s, int startFrom) {
     for (int p = startFrom; p < s.length(); p++) {
       char c = s.charAt(p);
       if (!Character.isLetterOrDigit(c) && c != '_' && c != '$') {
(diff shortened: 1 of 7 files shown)


src/test/java/com/google/api/codegen/testdata/py/python_library.baseline
```diff
@@ -3666,7 +3667,8 @@ for page in client.find_related_books(names, shelves, options=CallOptions(page_t
     for element in page:
         # process element
         pass
-print('Here\'s the list of related books:\n {}'.format(response))
+books = response.names
```
pongad: @vchudnov-g This should go in the loop right?

src/test/java/com/google/api/codegen/testsrc/library_gapic.yaml
```diff
@@ -647,9 +647,11 @@ interfaces:
       - names[0]=Odyssey
       - shelves[0]=Classics
       output_specs:
+        - assign:
+            books=$resp.names
```
vchudnov-g: I like this more compact format! I'll update the doc.

My stylistic preference: I would put it in the same line when it fits, for compactness:
```
- assign: books=$resp.names
```

src/test/java/com/google/api/codegen/testsrc/library_gapic.yaml
```diff
@@ -647,9 +647,11 @@ interfaces:
       - names[0]=Odyssey
       - shelves[0]=Classics
       output_specs:
+        - assign:
```
vchudnov-g: I am thinking we should s/`assign`/`define`/g everywhere. It more accurately names what we're doing (variable definition) and it leaves the name `assign` free should we ever need to implement just a plain old assign.

Thoughts?

src/test/java/com/google/api/codegen/testdata/py/python_library.baseline
vchudnov-g: We could have something like this:

```
      output_specs:
        - assign:
            books=$resp.names
        - print:
          - 'Here''s the list of related books:\n %s'
          - books
        - print:
          - 'Here''s each book:'
        - loop
      - loop:
          variable: title
          collection: $books
          body:
            print:
              - "Title: %s"
              - title
      - print:
        - 'That''s all!'
```

(May need to tweak this to match the actual structure of the structs and fields.)

WDYT?

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -233,4 +234,28 @@ private static int identifier(String s, int startFrom) {
...
+    String ident = assign.substring(0, p);
+    Preconditions.checkArgument(
+        assign.startsWith("=", p),
+        "%s:%s invalid assign, expecting '=': %s",
+        context.getMethodModel().getSimpleName(),
+        valueSet.getId(),
+        assign);
+    OutputView.VariableView reference =
+        accessor(assign.substring(p + 1), context, valueSet, localVars, ident);
+    return OutputView.AssignmentView.newBuilder()
```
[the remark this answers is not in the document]
vchudnov-g: Actually, my last comment on this thread is not quite correct if you implement a local variable table. To put the restriction in place coherently, you would need to check that a new variable wan't defined even inside a block that is not in scope now.

---

Another thought: another way to implement block scope is, instead of cloning the variable table, having something akin to a stack of variable tables corresponding to the scopes that are on the stack. I say "akin" because you'd want to easily search for a given variable name at all levels of the stack. (If we were doing proper block scoping allowing duplicate identifiers in different scopes, you would do this search from the top of the stack down.)

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -233,4 +234,28 @@ private static int identifier(String s, int startFrom) {
...
     return ret;
   }
+
+  private static OutputView.AssignmentView assignmentView(
+      String assign,
+      MethodContext context,
+      SampleValueSet valueSet,
+      Map<String, TypeModel> localVars) {
+
+    int p = identifier(assign, 0);
```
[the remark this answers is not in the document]

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -41,11 +40,7 @@
         OutputSpec.newBuilder().addPrint("%s").addPrint(RESPONSE_PLACEHOLDER).build());
   }
 
-  static OutputView toView(OutputSpec config, MethodContext context, SampleValueSet valueSet) {
-    return toView(config, context, valueSet, new HashMap<>());
-  }
-
-  private static OutputView toView(
+  static OutputView toView(
```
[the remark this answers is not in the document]
pongad: At least right now `OutputTransformer` transforms one spec at a time. However, alias declarations affect not only itself, it affects all declarations behind it. But I agree. I'll make OutputTransformer work on list instead.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -233,4 +234,28 @@ private static int identifier(String s, int startFrom) {
...
+    int p = identifier(assign, 0);
+    String ident = assign.substring(0, p);
+    Preconditions.checkArgument(
+        assign.startsWith("=", p),
+        "%s:%s invalid assign, expecting '=': %s",
+        context.getMethodModel().getSimpleName(),
+        valueSet.getId(),
+        assign);
+    OutputView.VariableView reference =
+        accessor(assign.substring(p + 1), context, valueSet, localVars, ident);
```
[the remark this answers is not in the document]

src/test/java/com/google/api/codegen/testsrc/library_gapic.yaml
pongad: Oh right, done

src/test/java/com/google/api/codegen/testsrc/library_gapic.yaml
pongad: Fine with me :)

src/test/java/com/google/api/codegen/testdata/py/python_library.baseline
pongad: Sorry I think you misunderstood. I meant, this is a paginating method, so we have to do
```python
for element in page:
  books = response.names
  # ...
```
right?

pongad:
@vchudnov-g PTAL

src/test/java/com/google/api/codegen/testdata/py/python_library.baseline
vchudnov-g: (You mean `books = page.names`?)

OK, two things:

1. yes, I did not think about paging. We will have to deal with this when we make response handling work with paging methods. Good catch.
2. In this particular instance, the loop is part of the stub samples that we're cribbing from the incode samples. When we fill out the real standalone samples, then yes, we'll have to have the response handling in the loop.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -211,51 +222,104 @@ private static int identifier(String s, int startFrom) {
...
             .body(
                 loop.getBodyList()
                     .stream()
-                    .map(body -> toView(body, context, valueSet, localVars))
+                    .map(body -> toView(body, context, valueSet, scope))
                     .collect(ImmutableList.toImmutableList()))
             .build();
 
     // The variable is only visible within the loop, delete it from the table before we return.
-    localVars.remove(loop.getVariable());
```
vchudnov-g: Remove the comment for this deleted line, too.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -211,51 +222,104 @@ private static int identifier(String s, int startFrom) {
...
-      MethodContext context,
-      SampleValueSet valueSet,
-      Map<String, TypeModel> localVars) {
-
-    int p = identifier(assign, 0);
-    String ident = assign.substring(0, p);
+  private static OutputView.DefinitionView definitionView(
+      String definition, MethodContext context, SampleValueSet valueSet, ScopeTable localVars) {
+    int p = identifierEnd(definition, 0);
+    String ident = definition.substring(0, p);
```
vchudnov-g: what do you think about invoking `trim` here and in line 260 (`definition.substring(p+1)`, though this will need to be tweaked if we do) so that we allow whitespace around the `=`?

(Even if you agree it's worthwhile, I'm OK deferring)

vchudnov-g: I still vote for s/`ident`/`identifier`/ or s/`ident`/`newVariable`/ or some such.... Not blocking on this.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -211,51 +222,104 @@ private static int identifier(String s, int startFrom) {
...
         .reference(reference)
         .build();
   }
+
+  /**
+   * We track two scopes: universe and local.
+   *
+   * <p>Universe keeps track of all variable declared by the output specs. We need this because
+   * variables are function-scoped in many dynamic languages, and we should error if the spec
+   * declares variable with the same name twice.
```
vchudnov-g: s/variable/a variable/

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -211,51 +222,104 @@ private static int identifier(String s, int startFrom) {
...
+   *
+   * <p>Local keeps track of variable in the current block. We need this because variables are
```
vchudnov-g: s/variable/variables/

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -211,51 +222,104 @@ private static int identifier(String s, int startFrom) {
...
+   * block-scoped in many static languages, and we should error if the spec uses a variable not in
+   * the current block or its transitive parents.
```
vchudnov-g: s/transitive parents/ancestors/ or
s/the current block or its transitive parents/the nested blocks currently in scope/

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -211,51 +222,104 @@ private static int identifier(String s, int startFrom) {
...
+   */
+  private static class ScopeTable {
+    private final Set<String> universe;
+    @Nullable private final ScopeTable parent;
+    private final Map<String, TypeModel> scope = new HashMap<>();
+
+    private ScopeTable() {
```
vchudnov-g: TIL that private methods in private inner classes get package-level access: https://stackoverflow.com/questions/663059/why-do-inner-classes-make-private-methods-accessible

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -211,51 +222,104 @@ private static int identifier(String s, int startFrom) {
...
 
-  private static OutputView.AssignmentView assignmentView(
-      String assign,
-      MethodContext context,
-      SampleValueSet valueSet,
-      Map<String, TypeModel> localVars) {
-
-    int p = identifier(assign, 0);
-    String ident = assign.substring(0, p);
+  private static OutputView.DefinitionView definitionView(
```
vchudnov-g: nit: I would vote for keeping the verbs from the spec in the view names, ie `DefineView`

(This didn't come up before because "Loop" and "Comment" can also be nouns, and "Print" is also noun with a slightly different meaning)

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -98,11 +104,16 @@ public void run() {
             config
                 .subList(1, config.size())
                 .stream()
-                .map(a -> accessor(a, context, valueSet, localVars, null))
+                .map(a -> accessor(a, context, valueSet, localVars))
                 .collect(ImmutableList.toImmutableList()))
         .build();
   }
 
+  private static OutputView.VariableView accessor(
```
vchudnov-g: Let's put all the utility functions (`accessor*`, `identifierEnd`) at the end of the fileso the various `*View()` are together at the top.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -115,17 +126,17 @@ public void run() {
    * config refers to a local variable, the variable is looked up in {@code localVars}. If {@code
    * newVar} is not null, it is registered into {@code localVars}.
    */
-  private static OutputView.VariableView accessor(
+  private static OutputView.VariableView accessorNewVariable(
```
vchudnov-g: Fold the 
```
// accessor:
...
```
comment lines into the javadoc comment.

src/main/java/com/google/api/codegen/viewmodel/OutputView.java
```diff
@@ -20,28 +20,28 @@
 public interface OutputView {
 
   public enum Kind {
-    ASSIGNMENT,
     COMMENT,
+    DEFINITION,
```
vchudnov-g: WDYT about the idea above of making these all verbs that correspond to the input spec?

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
pongad: Done.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
pongad: If we're doing this, I think we should also allow spaces like `foo = $resp .bar [ 42   ]`. We can definitely do that with a lexer, but I think we should defer.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java

src/main/resources/com/google/api/codegen/py/standalone_sample.snip
```diff
@@ -78,7 +78,7 @@
     @switch view.kind
       @case "COMMENT"
         {@processCommentView(view)}
-      @case "DEFINITION"
+      @case "DEFINE"
         {@processDefinitionView(view)}
```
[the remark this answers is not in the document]
pongad: Derp sorry.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -218,64 +269,16 @@ private static int identifierEnd(String s, int startFrom) {
...
-    return OutputView.DefinitionView.newBuilder()
-        .variableType(
-            context.getNamer().getAndSaveTypeName(context.getTypeTable(), localVars.get(ident)))
-        .variableName(context.getNamer().localVarName(Name.from(ident)))
-        .reference(reference)
-        .build();
-  }
-
   /**
    * We track two scopes: universe and local.
```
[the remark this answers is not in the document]
pongad: I like "sample".

PR #2121, opened 2018-06-27
Supporting same proto_deps structure as artman config

garrettjonesgoogle:
- artman uses a structure with a key named `name` instead of just a list of dependency names
- artman uses `test_proto_deps` instead of `proto_test_deps`

### DIFF
--- a/src/test/java/com/google/api/codegen/testsrc/library_pkg2.yaml
+++ b/src/test/java/com/google/api/codegen/testsrc/library_pkg2.yaml
@@ -4,8 +4,8 @@ artifact_type: GAPIC
 organization_name: google-cloud
 proto_path: google/library
 proto_deps:
-- google-common-protos
-- google-some-other-package-v1
-proto_test_deps:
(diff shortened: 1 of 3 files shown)


src/test/java/com/google/api/codegen/testsrc/library_pkg2.yaml
```diff
@@ -4,8 +4,8 @@ artifact_type: GAPIC
 organization_name: google-cloud
 proto_path: google/library
 proto_deps:
-- google-common-protos
-- google-some-other-package-v1
-proto_test_deps:
-- google-some-test-package-v1
+- name: google-common-protos
```
pongad: Do we need to make a similar change anywhere else?

garrettjonesgoogle: I also pushed an artman change to send this format: https://github.com/googleapis/artman/pull/444

PR #2128, opened 2018-06-28
Enabling separate code-only generation for C#

garrettjonesgoogle:
This enables a command like this:

```
java -cp ../gapic-generator/build/libs/gapic-generator-latest-fatjar.jar \
  com.google.api.codegen.GeneratorMain GAPIC_CODE \
  --descriptor_set=artman-genfiles/google-cloud-pubsub-v1.desc \
  --output=artman-genfiles/csharp/google-cloud-pubsub \
  --language=csharp \
  --service_yaml=google/pubsub/pubsub.yaml \
  --gapic_yaml=google/pubsub/v1/pubsub_gapic.yaml
```

This will only output code files (*.cs), omitting *.csproj files. 

To output only *.csproj files, use GAPIC_PACKAGE instead of GAPIC_CODE.

### DIFF
--- a/src/main/java/com/google/api/codegen/gapic/GapicGeneratorFactory.java
+++ b/src/main/java/com/google/api/codegen/gapic/GapicGeneratorFactory.java
@@ -31,9 +31,13 @@ import com.google.api.codegen.common.TargetLanguage;
 import com.google.api.codegen.config.GapicProductConfig;
 import com.google.api.codegen.config.PackageMetadataConfig;
 import com.google.api.codegen.config.ProtoApiModel;
+import com.google.api.codegen.gapic.GapicGenerator.Builder;
 import com.google.api.codegen.nodejs.NodeJSCodePathMapper;
 import com.google.api.codegen.php.PhpGapicCodePathMapper;
 import com.google.api.codegen.rendering.CommonSnippetSetRunner;
+import com.google.api.codegen.transformer.ModelToViewTransformer;
+import com.google.api.codegen.transformer.csharp.CSharpBasicPackageTransformer;
+import com.google.api.codegen.transformer.csharp.CSharpGapicClientPackageTransformer;
 import com.google.api.codegen.transformer.csharp.CSharpGapicClientTransformer;
 import com.google.api.codegen.transformer.csharp.CSharpGapicSmokeTestTransformer;
 import com.google.api.codegen.transformer.csharp.CSharpGapicSnippetsTransformer;
@@ -74,24 +78,10 @@ import java.io.File;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.function.Function;
 
 /** GapicGeneratorFactory creates CodeGenerator instances based on an id. */
 public class GapicGeneratorFactory {
-  public static final String ARTIFACT_SURFACE = "surface";
-  public static final String ARTIFACT_TEST = "test";
-  public static final String ARTIFACT_SAMPLES = "samples";
-
-  public static boolean enableSurfaceGenerator(List<String> enabledArtifacts) {
-    return enabledArtifacts.isEmpty() || enabledArtifacts.contains(ARTIFACT_SURFACE);
-  }
-
-  public static boolean enableTestGenerator(List<String> enabledArtifacts) {
-    return enabledArtifacts.isEmpty() || enabledArtifacts.contains(ARTIFACT_TEST);
-  }
-
-  public static boolean enableSampleGenerator(List<String> enabledArtifacts) {
-    return enabledArtifacts.contains(ARTIFACT_SAMPLES);
-  }
 
   /** Create the GapicGenerators based on the given id */
   public static List<CodeGenerator<?>> create(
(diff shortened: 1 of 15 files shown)


src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicClientPackageTransformer.java
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  @Override
+  public List<ViewModel> transform(ProtoApiModel model, GapicProductConfig productConfig) {
+    List<ViewModel> surfaceDocs = new ArrayList<>();
+    SurfaceNamer namer = new CSharpSurfaceNamer(productConfig.getPackageName(), ALIAS_MODE);
+    CSharpFeatureConfig featureConfig = new CSharpFeatureConfig();
+
+    InterfaceModel lastApiInterface = null;
+    for (InterfaceModel apiInterface : model.getInterfaces()) {
```
eoogbe: You're exposed to an NPE if `getInterfaces` is empty. How about:
```java
List<ProtoInterfaceModel> apiInterfaces = model.getInterfaces();
if (apiInterfaces.isEmpty()) {
  return surfaceDocs;
}

GapicInterfaceContext context =
    GapicInterfaceContext.create(apiInterfaces.get(apiInterfaces.size() - 1), ...
```

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicClientPackageTransformer.java
```diff
@@ -0,0 +1,140 @@
...
+    String outputPath =
+        pathMapper.getOutputPath(context.getInterface().getFullName(), productConfig);
+    view.outputPath(outputPath + File.separator + productConfig.getPackageName() + ".csproj");
+    view.fileHeader(fileHeaderTransformer.generateFileHeader(context));
+    view.serviceTitle(model.getServiceConfig().getTitle());
+    view.serviceDescription(model.getServiceConfig().getDocumentation().getSummary().trim());
+    view.domainLayerLocation(productConfig.getDomainLayerLocation());
+    view.authScopes(new ArrayList<>()); // Unused in C#
+    view.releaseLevel(productConfig.getReleaseLevel());
+    String versionSuffix;
```
eoogbe: Nit: Replace mutable variable with private method.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicClientPackageTransformer.java
```diff
@@ -0,0 +1,140 @@
...
+        versionSuffix = "-beta01";
+        break;
+      default:
+        versionSuffix = "";
+        break;
+    }
+    view.version("1.0.0" + versionSuffix);
+    String tags = "";
+    for (String tag : Splitter.on('.').split(productConfig.getPackageName())) {
+      if (tag.matches("[vV][\\d.]+")) {
```
eoogbe: Duplicates [`VersionMatcher`](https://github.com/googleapis/gapic-generator/blob/master/src/main/java/com/google/api/codegen/util/VersionMatcher.java)

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicClientPackageTransformer.java
```diff
@@ -0,0 +1,140 @@
...
+        break;
+      }
+      tags += ";" + tag;
+    }
+    view.tags(tags.isEmpty() ? "" : tags.substring(1));
```
eoogbe: This process is a bit opaque on first read. If you filter the matching tags out of the split package name, you can just use a `Joiner` to intersperse the semicolons.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicClientTransformer.java
```diff
@@ -92,12 +83,10 @@
 
   private static final String XAPI_TEMPLATE_FILENAME = "csharp/gapic_client.snip";
```
eoogbe: Nit: Rename `XAPI`. (You know you wanna 😛 )

src/main/java/com/google/api/codegen/gapic/GapicGeneratorFactory.java
```diff
@@ -74,97 +78,93 @@
...
-        CodeGenerator snippetGenerator =
-            GapicGenerator.newBuilder()
-                .setModel(model)
-                .setProductConfig(productConfig)
-                .setSnippetSetRunner(new CommonSnippetSetRunner(new CSharpRenderingUtil()))
-                .setModelToViewTransformer(new CSharpGapicSnippetsTransformer(snippetPathMapper))
-                .build();
-        generators.add(snippetGenerator);
+
+      Function<String, GapicCodePathMapper> newCodePathMapper =
```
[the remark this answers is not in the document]
garrettjonesgoogle: They are lambda variables because that creates the most concise, non-repetitive code. If I created private methods for each of these, they would have about as many parameters as they had lines of code, and the resulting code would actually be the same size as the original, which would defeat the purpose. I feel that lambdas are highly appropriate here because they can reference the local scope, and be scoped to that local scope at the same time.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpBasicPackageTransformer.java
```diff
@@ -0,0 +1,123 @@
...
+      "csharp/gapic_snippets_csproj.snip";
+  private static final String UNITTEST_CSPROJ_TEMPLATE_FILENAME =
+      "csharp/gapic_unittest_csproj.snip";
+
+  private static final CSharpAliasMode ALIAS_MODE = CSharpAliasMode.MessagesOnly;
+
+  private final GapicCodePathMapper pathMapper;
+  private final String templateFilename;
+  private final String fileBaseSuffix;
+  private final Function<GapicInterfaceContext, Boolean> shouldGenerateFn;
```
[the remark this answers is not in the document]
garrettjonesgoogle: Oh that is indeed the right thing to use, I'll switch to that.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicClientPackageTransformer.java
garrettjonesgoogle: This method was moved from https://github.com/googleapis/gapic-generator/blob/master/src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicClientTransformer.java#L160 , and I made no modifications to it - would you be ok if I left it as-is? (applies to the two other comments you left also)

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicClientTransformer.java
garrettjonesgoogle: yes I do indeed!

src/main/java/com/google/api/codegen/transformer/csharp/CSharpBasicPackageTransformer.java
```diff
@@ -0,0 +1,123 @@
...
+      String templateFilename,
+      String fileBaseSuffix,
+      Function<GapicInterfaceContext, Boolean> shouldGenerateFn) {
+    this.pathMapper = pathMapper;
+    this.templateFilename = templateFilename;
+    this.fileBaseSuffix = fileBaseSuffix;
+    this.shouldGenerateFn = shouldGenerateFn;
+  }
+
+  public static CSharpBasicPackageTransformer forSmokeTests(GapicCodePathMapper pathMapper) {
```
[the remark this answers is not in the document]
garrettjonesgoogle: The list of things here is a very restricted set - I don't think we even have in mind any additional cases right now that will expand this, so I prefer the present simple approach. 

This was the most concise way I found to encapsulate the specific values for the different cases without dumping it on the caller (`GapicGeneratorFactory`). 

Later, if this did start expanding, I would heartily agree with a different design.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicClientPackageTransformer.java
garrettjonesgoogle: Will do.

garrettjonesgoogle:
@eoogbe agreed-to feedback addressed, PTAL

src/main/java/com/google/api/codegen/gapic/GapicGeneratorFactory.java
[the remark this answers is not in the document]
eoogbe: Good point. I didn't notice the local references in the lambdas. `newCodePathMapper` appears to only have 2 local references though. Up to you if it's worth moving to a private method.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicClientPackageTransformer.java
eoogbe: I'm a big fan of the Boy Scout Rule and depending on the tests (and code reviewers) to catch regression. (Doc forthcoming.) YMMV though. Since it doesn't make the codebase *worse* then I guess it's fair enough.

src/main/java/com/google/api/codegen/gapic/GapicGeneratorFactory.java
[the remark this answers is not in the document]
garrettjonesgoogle: newCodePathMapper has four local references; I'd prefer to keep it a lambda because the logic can be kept closer to where it is used (`create()` is a mega-function).

garrettjonesgoogle:
Both comments Ack'd.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicClientPackageTransformer.java
eoogbe: That approach only works if we make an explicit note (e.g. issue, doc) that we will indeed come back later and pick up all the cigarette butts.

PR #2138, opened 2018-07-03
Java: no ResourceName types for repeated fields

andreamlin:
Never generate ResourceName types for repeated fields to avoid type erasure problems when List<ResourceName> param method is generated alongside the overriden List<String> param method.


andreamlin:
PTAL

PR #2139, opened 2018-07-04
C# Fix various incorrect resource-name usages

chrisdunelm:


chrisdunelm:
The two changes seen in the baseline are:

* `GetShelfRequest` and `CreateBookRequest` `Name` property now set using the string directly, not a resource-name. This is correct, as the gapic_config configures these flattenings to not use resource-names.
* `GetBookFromAbsolutelyAnywhereRequest.AsResourceName` set to the correct resource-name. Previously the tests would have failed due to inconsistent resource-name usage in the expected request and the flattened method call.

jskeet:
Yup, will definitely do that :)

PR #2140, opened 2018-07-09
samplegen: print Python paging response correctly

pongad:
Previously we print the response-handling after the loop,
we correctly do that in the loop now.


src/main/resources/com/google/api/codegen/py/method_sample.snip
```diff
@@ -174,26 +174,29 @@
...
 
-@private responseStreamingSampleCode(apiMethod, initCode)
+@private responseStreamingSampleCode(apiMethod, initCode, processResp)
   for element in {@methodCallSampleCode(apiMethod, initCode)}:
-      @# process element
-      pass
+      {@processResp}
+@end
+
+@private processElement()
```
vchudnov-g: This may need a more descriptive name once we process the LRO responses, but we can deal with that then.

src/main/resources/com/google/api/codegen/py/method_sample.snip
```diff
@@ -174,26 +174,29 @@
...
 
-@snippet pagedOptionalArrayMethodSampleCodeAll(apiMethod, initCode)
+@snippet pagedOptionalArrayMethodSampleCodeAll(apiMethod, initCode, processResp)
   @# Iterate over all results
-  {@responseStreamingSampleCode(apiMethod, initCode)}
+  {@responseStreamingSampleCode(apiMethod, initCode, processResp)}
 @end
 
-@snippet pagedOptionalArrayMethodSampleCodePaged(apiMethod, initCode)
+@snippet pagedOptionalArrayMethodSampleCodePaged(apiMethod, initCode, processResp)
```
vchudnov-g: The following is giving me pause (which is why I didn't reply yesterday...):

We are re-using the template functions from this file in the temporary stubs in `standalone_sample.snip`. Those stubs will very likely get replaced by proper standalone sample-related template functions. In that case, changing the signature of these functions here that we're only using as temporary stubs there seems like adding unnecessary complexity. The alternative would be to use this as the opportunity to copy these functions into `standalone_sample.snip` and not depend on these, and those can have the new parameter.

I _think_ I'm leaning to accept what you have here because it's possible that we _may_ reuse some of these functions in the real standalone samples, not just the stubs. I'm just worried that if we don't, and wind up replacing the stubs after all, then the functions here will have crufty more-complicated signatures with the extra param that will no longer be needed.

Thoughts?

pongad:
@vchudnov-g PTAL

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -182,7 +184,8 @@ public void run() {
       MethodContext context,
       SampleValueSet valueSet,
       ScopeTable localVars,
-      @Nullable String newVar) {
+      @Nullable String newVar,
+      boolean intoScalar) {
```
vchudnov-g: Explain `intoScalar` in the function comments: if `config` is a collection, makes the new variable type be the type of the collection's elements.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -263,6 +266,9 @@ public void run() {
     }
 
     if (newVar != null) {
+      if (intoScalar) {
```
vchudnov-g: Shouldn't you verify here that `type` is a collection?

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -263,6 +266,9 @@ public void run() {
...
+        type = type.makeOptional();
```
vchudnov-g: I would add an inline comment for clarity:
```
// "optional" is how protobuf defines singular fields
```

src/main/resources/com/google/api/codegen/py/init.snip
```diff
@@ -0,0 +1,79 @@
+@extends "py/common.snip"
+
```
vchudnov-g: I would call this file `sample_init.snip`, since it is only used for samples. This will reduce confusion when looking at the template files in the directory, trying to figure out which one is used when.



[ Down the line, it may be worth moving the sample files into
* `.../codegen/py/samples/inline` for the inline samples
* `.../codegen/py/samples/standalone` for the standalone samples
* `.../codegen/py/samples/` for the files, like this one, common to all samples.

We can do this later. ]

src/main/resources/com/google/api/codegen/py/method_sample.snip
```diff
@@ -164,7 +94,7 @@
 @end
 
 @snippet optionalArrayMethodSampleCodeServerStreaming(apiMethod, initCode)
```
vchudnov-g: These `@snippet`s should go back to being `@private` as they were before we used them for the standalone stubs (cf #2046 ). This will help guarantee we don't accidentally call the wrong one. Pretty much each of the functions you coped to the standalone file should be made `@private` here too.

src/test/java/com/google/api/codegen/testdata/py/python_library.baseline
```diff
@@ -4107,7 +4107,6 @@ names = [names_element]
 shelves_element = 'Classics'
 shelves = [shelves_element]
 
-# Iterate over all results
```
vchudnov-g: Given the level of commenting in the published samples, I would keep these comments in there.

I don't feel strongly about this, though, since we'll be getting feedback on the samples anyway.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java

src/main/resources/com/google/api/codegen/py/init.snip
pongad: Done

src/main/resources/com/google/api/codegen/py/method_sample.snip
pongad: Done

src/test/java/com/google/api/codegen/testdata/py/python_library.baseline
pongad: I'll revert.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -267,6 +269,13 @@ public void run() {
 
     if (newVar != null) {
       if (intoScalar) {
+        Preconditions.checkArgument(
+            type.isRepeated() && !type.isMap(),
+            "%s:%s: %s is not a repeated field",
+            context.getMethodModel().getSimpleName(),
+            valueSet.getId(),
+            config.substring(0, end));
+        // "optional" is how protobuf defines singular fields
```
vchudnov-g: nit: I was thinking of this comment after the code on the next line. is there a stylistic reason not to?

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -267,6 +269,13 @@ public void run() {
 
     if (newVar != null) {
       if (intoScalar) {
+        Preconditions.checkArgument(
+            type.isRepeated() && !type.isMap(),
```
vchudnov-g: Don't we want to allow iterating over maps?

src/main/resources/com/google/api/codegen/py/method_sample.snip
```diff
@@ -1,5 +1,5 @@
 @extends "py/common.snip"
-@extends "py/init.snip"
+@extends "py/simple_init.snip"
```
vchudnov-g: uhm, s/`simple`/`sample`/ here and in the filename

src/test/java/com/google/api/codegen/testdata/py/python_library.baseline
```diff
@@ -4175,14 +4172,11 @@ names = [names_element]
 shelves_element = 'Classics'
 shelves = [shelves_element]
 
-# Iterate over results one page at a time
+# Iterate over all results
```
vchudnov-g: I would keep this as "one page at a time", since we are showing how to do that (and could insert code to print a page header).

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
pongad: I think comments usually describe what's about to happen instead of what just happened right? I make a line break to make this more clear.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
pongad: I think we should hold off on this. It's unclear what you're iterating over. Keys? Values? Pairs?

src/main/resources/com/google/api/codegen/py/method_sample.snip
pongad: Done

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
vchudnov-g: Oh, I meant in the same line:
```
type = type.makeOptional();  // "optional" is how protobuf defines singular fields
```

Sorry I wasn't clear.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
vchudnov-g: Agreed, but we need to decide this before we declare the loop construct to be fully implemented.....

PR #2142, opened 2018-07-10
Update gax and common-protos generated versions

andreamlin:
The version of google-common-protos has been bumped to 1.2 in google-cloud-java.

~The gax-java versions will be bumped in https://github.com/GoogleCloudPlatform/google-cloud-java/pull/3439 (should wait for this to pass first).~ CI passed


andreamlin:
PTAL

PR #2148, opened 2018-07-13
give OutputTransformer a proper lexer

pongad:


src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -359,4 +357,59 @@ private ScopeTable newChild() {
...
+      p += Character.charCount(cp);
+      return cp;
+    }
+
+    private static boolean identLead(int codePoint) {
+      return Character.isLetter(codePoint) || "$_".indexOf(codePoint) >= 0;
+    }
+
+    private static boolean identFollow(int codePoint) {
+      return identLead(codePoint) || Character.isDigit(codePoint);
```
vchudnov-g: "$" handling: We want `$` to be an identifier modifier prefix for our own purposes (eg `$resp`). Specifically, my understanding is that we were going to allow it _just_ to be a prefix. If that's not what you had in mind, let's discuss. 
If it is the case, then `$` should not be allowed in this function. As you have it, this allows identifiers such as `$$`, `$foo$bar`, and `$foo$`.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -359,4 +357,59 @@ private ScopeTable newChild() {
       return new ScopeTable(this);
     }
   }
+
+  private static class Scanner {
+    private static final int EOF = -1;
+    private static final int IDENT = -2;
+
+    private final String s;
```
vchudnov-g: `s` -> `str` or `input`

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -359,4 +357,59 @@ private ScopeTable newChild() {
...
+    private String token = "";
+    private int p;
```
vchudnov-g: s/`p`/`nextLoc`/

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -359,4 +357,59 @@ private ScopeTable newChild() {
...
+
+    private Scanner(String s) {
+      this.s = s;
+    }
+
+    private int scan() {
+      token = "";
+      int cp;
```
vchudnov-g: s/`cp`/`currCh`/ or similar

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -359,4 +357,59 @@ private ScopeTable newChild() {
       return new ScopeTable(this);
     }
   }
+
+  private static class Scanner {
```
vchudnov-g: This function should have a test.

vchudnov-g: A brief function comment would be useful, saying what characters we accept, and that `scan()` returns `EOF`, `IDENT`, or a character, and that in the case of `IDENT`, the token can be fetched via `.token`

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -181,7 +186,7 @@ public void run() {
    * </code></pre>
    */
   private static OutputView.VariableView accessorNewVariable(
```
vchudnov-g: Do we have a test for this function?

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
pongad: Changed to codePoint, cp is not a char.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
pongad: I think most of these are just tested by baselines.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
pongad: Done. Note that `$foo$bar` should then be legal. The string consists of two identifiers, `$foo` and `$bar`.

src/test/java/com/google/api/codegen/transformer/ScannerTest.java
```diff
@@ -0,0 +1,61 @@
...
+    Scanner scanner = new Scanner("$a$b$");
+
+    assertThat(scanner.scan()).isEqualTo(Scanner.IDENT);
+    assertThat(scanner.token()).isEqualTo("$a");
+
+    assertThat(scanner.scan()).isEqualTo(Scanner.IDENT);
+    assertThat(scanner.token()).isEqualTo("$b");
+
+    assertThrow(() -> scanner.scan());
+  }
```
vchudnov-g: I would also test that `"a$$b"`  and `"a$"` throw

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -360,54 +360,81 @@ private ScopeTable newChild() {
...
-    private static final int EOF = -1;
-    private static final int IDENT = -2;
+  /**
+   * Tokenizes a string. {@code scan()} returns the next token. A token is either the next unicode
+   * code point in the string or a special token. If a special token is returned, the string value
+   * of the token can be retrieved from {@code token}.
+   *
+   * <p>Special tokens:
+   *
+   * <pre>{@code
```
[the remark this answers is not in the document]
pongad: It does, but `<pre>` preserves them :)

PR #2151, opened 2018-07-14
php standalone samples

hzyi-google:


hzyi-google:
PTAL

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
```diff
@@ -119,29 +122,31 @@ public PhpGapicSurfaceTransformer(
               modelTypeTable,
               new PhpSurfaceNamer(productConfig.getPackageName()),
               new PhpFeatureConfig());
-      surfaceDocs.addAll(transform(context));
+      surfaceDocs.addAll(transform(context, model.hasMultipleServices()));
```
vchudnov-g: A cautionary question: since this new field was not being passed in before, do you have any assurance the value being passed in corresponds to what was assumed/defaulted to before? Or that this fixes previous behavior?

(For example, `PythonMethodViewGenerator` does not take this param; it's always false there.)

src/main/java/com/google/api/codegen/transformer/php/PhpMethodViewGenerator.java
```diff
@@ -0,0 +1,99 @@
...
+import com.google.api.codegen.transformer.DynamicLangApiMethodTransformer;
+import com.google.api.codegen.transformer.GapicInterfaceContext;
+import com.google.api.codegen.transformer.GapicMethodContext;
+import com.google.api.codegen.viewmodel.CallingForm;
+import com.google.api.codegen.viewmodel.OptionalArrayMethodView;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class PhpMethodViewGenerator {
```
vchudnov-g: Please add class comment as in Python

src/main/java/com/google/api/codegen/transformer/php/PhpMethodViewGenerator.java
```diff
@@ -0,0 +1,99 @@
...
+      InitCodeContext initContext,
+      boolean packageHasMultipleServices) {
+    OptionalArrayMethodView methodView = null;
+    if (methodContext.getMethodConfig().isPageStreaming()) {
+      methodView =
+          clientMethodTransformer.generatePagedStreamingMethod(
+              methodContext,
+              initContext,
+              packageHasMultipleServices,
+              Arrays.asList(CallingForm.RequestPagedAll, CallingForm.RequestPaged));
```
vchudnov-g: nit: within a call, let's put the `CallingForm` enums in alphabetical order.

src/main/java/com/google/api/codegen/viewmodel/CallingForm.java
```diff
@@ -27,15 +27,17 @@
...
+  Request, // used by: java nodejs py php
   RequestAsync,
   RequestAsyncPaged, // used by: nodejs
   RequestAsyncPagedAll, // used by: nodejs
-  RequestPaged, // used by: java py
-  RequestPagedAll, // used by: py
-  RequestStreamingBidi, // used by: nodejs py
-  RequestStreamingClient, // used by: nodejs py
-  RequestStreamingServer, // used by: nodejs py
+  RequestPaged, // used by: java py php
```
vchudnov-g: For consistency, let's keep the language comments in alphabetical order.

src/main/resources/com/google/api/codegen/php/method_sample.snip
```diff
@@ -58,7 +58,7 @@
     @end
 @end
 
-@private methodSampleCode(apiMethod)
+@snippet methodSampleCode(apiMethod)
```
vchudnov-g: In this file, the `@sampleCode` function that selects which samples to output should key off the calling form rather than what it currently does, so that both the in-code and standalone samples key off the calling form.

[Incidentally, that's part of the reason in the past I've gravitated to making these changes in 2 PRs per language: one for the generator infrastructure in Java, and a barebones stub template, and another one normalizing the standalone and in-code samples to both key off the calling form, with the in-code samples yielding the same code as before.]

src/main/resources/com/google/api/codegen/php/method_sample.snip
```diff
@@ -82,6 +82,24 @@
     }
 @end
 
+@snippet pagedMethodCallSampleCode(apiMethod)
```
vchudnov-g: If these new functions are specific to standalone samples, move them to `standalone_sample.py`. If they're used by both in-code samples and standalone samples (which may well mean they are only temporary stubs for the standalone samples until we flesh them out), leave them here.

src/main/resources/com/google/api/codegen/php/standalone_sample.snip
```diff
@@ -0,0 +1,65 @@
+@extends "php/common.snip"
+@extends "php/method_sample.snip"
+
```
vchudnov-g: nit: You have some trailing whitespace in several lines in this file. Please remove it if you can.

src/main/resources/com/google/api/codegen/php/standalone_sample.snip
```diff
@@ -0,0 +1,65 @@
...
+       // [END core_sample]
+  
+       // FIXME: Insert here clean-up code.
+         
+     @end
+  @end
+  // [END full_sample]
+@end
+
+@snippet standaloneSample(apiMethod, sample)
```
vchudnov-g: Please make the inline samples in the other file also key off the `callingForm` like you have here, and add a comment in both places to remind us the structure between the two (not necessarily the methods called) should be the same. See the corresponding Python template for an example.

src/main/java/com/google/api/codegen/gapic/GapicGeneratorFactory.java
```diff
@@ -373,7 +374,19 @@
                 .build();
         generators.add(testGenerator);
       }
-
+      if (artifactFlags.sampleGeneratorEnabled()) {
```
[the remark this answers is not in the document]
hzyi-google: I looked at google-cloud-lang repos and it seems none of them put samples in the same repo... All of them put samples in a different repo called lang-docs-samples. Can you clarify more about what "living elsewhere" means?

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
```diff
@@ -73,10 +75,13 @@
   private ServiceTransformer serviceTransformer;
   private PathTemplateTransformer pathTemplateTransformer;
   private PageStreamingTransformer pageStreamingTransformer;
-  private DynamicLangApiMethodTransformer apiMethodTransformer;
   private GrpcStubTransformer grpcStubTransformer;
+  private final DynamicLangApiMethodTransformer apiMethodTransformer =
+      new DynamicLangApiMethodTransformer(new PhpApiMethodParamTransformer());
```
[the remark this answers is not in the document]
hzyi-google: I don't get this. Can you explain more?

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
hzyi-google: Yes you are right. The boolean value is never used. I'll make the change.

src/main/resources/com/google/api/codegen/php/standalone_sample.snip
hzyi-google: Ack

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSamplesTransformer.java
```diff
@@ -0,0 +1,117 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen.transformer.php;
+
+/**
+ * A transformer to generate php standalone samples for each method in the GAPIC surface generated
```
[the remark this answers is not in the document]
hzyi-google: Done

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSamplesTransformer.java
```diff
@@ -0,0 +1,117 @@
...
+        String className =
+            namer.getApiSampleClassName(
+                method.name(), methodSample.callingForm().toString(), methodSample.valueSet().id());
+        String sampleOutputPath = subPath + File.separator + namer.getApiSampleFileName(className);
+        viewModels.add(
+            sampleClassBuilder
+                .templateFileName(STANDALONE_SAMPLE_TEMPLATE_FILENAME)
+                .outputPath(sampleOutputPath)
+                .className(className)
+                .libraryMethod(method.toBuilder().samples(Arrays.asList(methodSample)).build())
```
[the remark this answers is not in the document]
hzyi-google: Done.

src/main/java/com/google/api/codegen/transformer/php/PhpMethodViewGenerator.java
hzyi-google: Done.

src/main/java/com/google/api/codegen/transformer/php/PhpMethodViewGenerator.java
hzyi-google: Done.

src/main/java/com/google/api/codegen/viewmodel/CallingForm.java
hzyi-google: Done.

src/main/resources/com/google/api/codegen/php/method_sample.snip
hzyi-google: Ack.

src/main/resources/com/google/api/codegen/php/method_sample.snip
hzyi-google: Ack

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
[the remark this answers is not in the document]
vchudnov-g: `DynamicLangApiMethodTransformer` has a few overloads. One of them has an explicit sample type.

src/main/java/com/google/api/codegen/gapic/GapicGeneratorFactory.java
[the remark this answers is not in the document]
vchudnov-g: The plan is that for most languages (unless there is a strong language convention not to) the standalone samples will live in a separate, top-level directory `samples/` so that they are browseable with the client libraries. We have not started publishing samples yet, which is why you don't see them there yet....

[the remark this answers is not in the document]
hzyi-google: Ok. I will change the behavior to default first and check with yoshi team to see if they are OK with the default behavior.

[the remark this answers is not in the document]
vchudnov-g: We can just have the samples generated with the client library for now (development) and then as they get ready to publish we can tweak if they need it changed.

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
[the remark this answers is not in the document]
hzyi-google: Done.

hzyi-google:
PTAL @vchudnov-g

src/main/resources/com/google/api/codegen/php/method_sample.snip
```diff
@@ -32,37 +33,39 @@
...
+        {@incodeSampleCall(apiMethod, sample)}
     @end
 @end
 
-@private methodSampleCode(apiMethod)
-    @if apiMethod.hasReturnValue
-        $response = {@methodCallSampleCode(apiMethod)};
-    @else
-        {@methodCallSampleCode(apiMethod)};
+@private incodeSampleCall(apiMethod, sample)
```
vchudnov-g: Add a comment saying the structure of this should be parallel to the method in standalone_sample.snip. And add a comment there to make sure the structure should be parallel to the one here.

(See py for an example)

src/main/resources/com/google/api/codegen/php/sample_common.snip
```diff
@@ -0,0 +1,212 @@
+@snippet methodCallSampleCode(apiMethod)
+    @if apiMethod.isSingularRequestMethod
+        ${@apiMethod.apiVariableName}->{@apiMethod.name}(\
+            {@sampleMethodCallArgList(apiMethod.initCode)})
+    @else
+        ${@apiMethod.apiVariableName}->{@apiMethod.name}()
+    @end
+@end
+
+@snippet longRunningSampleCode(apiMethod)
```
vchudnov-g: The paradigm I've been using in other languages is that each of these sample functions would also get the `sample.initCode` as a param. For in-code samples, each of the configured samples would thus result in a separate function call with a different `sample`'s `initCode` being passed in. For standalone sample, we only do this for one sample,  `apiMethod.samples.get(0)`.

I was confused by how you're getting correct baselines without passing `initCode` to these functions, but I think I see it now: you are doing the init code separately, so by the time you get to these functions, all the `initCode`-related set-up is already done so the the remaining function call does not need to know it. Is that correct?

If so, then that's totally fine (and in fact, may well be the way we switch all languages to, since this is one way to enable the responsive APIs Explorer samples later).

src/main/resources/com/google/api/codegen/php/method_sample.snip
```diff
@@ -32,37 +33,39 @@
...
-          {@bidiStreamingMethodSampleCode(apiMethod)}
-      @default
-          $unhandledCase: {@apiMethod.grpcStreamingType}$
-      @end
+    {@incodeSamples(apiMethod)}
+@end
 
+@snippet incodeSamples(apiMethod)
+    @join sample : apiMethod.samples on BREAK.add(BREAK).add(BREAK).add("// Alternatively:").add(BREAK)
+        {@incodeSampleCall(apiMethod, sample)}
```
vchudnov-g: I think the `{@initCode...}` from `decorateSampleCode` should be moved here, right above {@incodeSampleCall....}

Also (IMPORTANT), that `{@initCode ... }` line should not be referencing `apiMethod.initCode`, but rather `sample.initCode`. (We're deprecating use of method-level `initCode` in favor of sample-level `initCode`. The infrastructure is in place so that if only the former is configured in the YAML, it will be available as the latter by the time we create view models.)

Now, for methods with multiple in-code samples, this would repeat the setup code. To avoid that for situations where the setup code is the same, you could implement something like `py/method_samples.snip:12:incodeSamples()`

src/main/resources/com/google/api/codegen/php/sample_common.snip
hzyi-google: Yes, that was how php snippets originally handle `initCode`. I barely changed that part.

hzyi-google:
PTAL @vchudnov-g 
(Sorry I screwed up the white spaces a bit :/)

src/test/java/com/google/api/codegen/testdata/php/php_library.baseline
```diff
@@ -1498,19 +1504,21 @@ class LibraryServiceGapicClient
      *     $name = '';
      *     $request = new DiscussBookRequest();
      *     $request->setName($name);
```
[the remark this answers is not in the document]
hzyi-google: Adding a blank line here would result in adding a blank line here: https://github.com/googleapis/gapic-generator/pull/2151/files?utf8=%E2%9C%93&diff=split#diff-36f32ff1b2217a381ca1eaeeea45dfe6R141 since it's between any initialization and sample call. Is that ok with you?

src/main/java/com/google/api/codegen/viewmodel/DynamicLangSampleView.java
```diff
@@ -16,11 +16,18 @@
...
 import com.google.api.codegen.SnippetSetRunner;
 import com.google.auto.value.AutoValue;
+import javax.annotation.Nullable;
 
 /** A ViewModel for standalone samples for dynamic languages. */
 @AutoValue
 public abstract class DynamicLangSampleView implements ViewModel {
 
+  /**
+   * a placeholder class intended for different languages to subclass and put language specific
```
vchudnov-g: nits:
- start comment with cap
- hyphenize "language-specific"

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSamplesTransformer.java
```diff
@@ -96,12 +96,74 @@ public PhpGapicSamplesTransformer(
...
+                        .hasDefaultServiceAddress(
+                            context.getInterfaceConfig().hasDefaultServiceAddress())
+                        .build())
                 .build());
       }
     }
     return viewModels.build();
   }
 
+  public static class PhpSampleExtraInfo extends DynamicLangSampleView.SampleExtraInfo {
```
vchudnov-g: Can't you use Autovalue for this class (for the `has*` methods)? I would expect  it would be possible.

Alternatively, you can forgo the Builder pattern entirely and either:
1. have a class with accessors and a param constructor instead of a builder
2. have a class with public fields instead of accessors and a param constructor instead of a builder

I'd probably consider 1 for simplicity. But if you want to leave as is, the code's correct too, obviously.

src/main/resources/com/google/api/codegen/php/method_sample.snip
```diff
@@ -37,11 +36,37 @@
...
+@private incodeSampleSetup(apiMethod, sample)
+    @if sample.initCode.lines
+        {@initCode(sample.initCode, apiMethod.apiVariableName)}
+
+    @end
+@end
+
+# This function returns true iff @incodeSampleSetup() would produce
+# different code for firstSample and secondSample.
+@private sampleSetupDiffers(firstSample, secondSample) horizontal
```
vchudnov-g: Since this is used in Python too, maybe we could factor it out of both places and into `common/`?

Additionally, I would not   be opposed to also factoring the `if` condition in both languages so that the statement in both places reads `@if sampleNeedsSetup(sample, firstSample)`

src/test/java/com/google/api/codegen/testdata/php/php_library.baseline
```diff
@@ -1112,7 +1113,6 @@ class LibraryServiceGapicClient
      * ```
      * $libraryServiceClient = new LibraryServiceClient();
      * try {
```
vchudnov-g: I wonder whether we should have a blank line before these `try`s. WDYT?

Not a big deal right now, we can tweak this later.

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSamplesTransformer.java
```diff
@@ -109,58 +110,42 @@ public PhpGapicSamplesTransformer(
     return viewModels.build();
   }
 
-  public static class PhpSampleExtraInfo extends DynamicLangSampleView.SampleExtraInfo {
+  @AutoValue
+  public abstract static class PhpSampleExtraInfo extends DynamicLangSampleView.SampleExtraInfo {
 
-    private boolean hasDefaultServiceAddress;
-    private boolean hasDefaultServiceScopes;
+    // private PhpSampleExtraInfo(Builder builder) {
```
vchudnov-g: cruft?

hzyi-google: Oh sorry about that :/

hzyi-google:
cc/ @beccca sorry forgot to cc you earlier.

PR #2167, opened 2018-07-24
Enable naive server streaming retries.

igorbernstein2:
This enables retries that were added to gax in
https://github.com/googleapis/gax-java/pull/463.
Server streaming api can be treated the same as unary apis before the
first response is received. Allowing GAPIC clients to automatically
retry RPCs. This PR exposes the retry settings that were configured in
gapic yaml file to the java stub settings.

This will affect the codegen for 3 clients:
- firestore - will enable retries for batchGetDocuments & runQuery.
  Which will bring its behavior closer to the golang client
- bigtable - the handwritten overlay will overwrite these settings
- spanner - doesn't currently use gapic client, so will ignore the new
  settings

ping @schmidt-sebastian (for firestore behavior change)

### DIFF
--- a/src/main/java/com/google/api/codegen/viewmodel/StaticLangStubSettingsView.java
+++ b/src/main/java/com/google/api/codegen/viewmodel/StaticLangStubSettingsView.java
@@ -51,15 +51,13 @@ public abstract class StaticLangStubSettingsView {
     return unaryCallSettings;
   }
 
-  /**
-   * ApiCalls that send exactly one request. This includes unary and server streaming apis.
-   */
+  /** ApiCalls that send exactly one request. This includes unary and server streaming apis. */
   public List<ApiCallSettingsView> unaryRequestCallSettings() {
     ArrayList<ApiCallSettingsView> retryableCallSettings = new ArrayList<>();
     for (ApiCallSettingsView settingsView : callSettings()) {
       if (settingsView.type().serviceMethodType() == ServiceMethodType.UnaryMethod
           || settingsView.type().serviceMethodType()
-          == ServiceMethodType.GrpcServerStreamingMethod) {
+              == ServiceMethodType.GrpcServerStreamingMethod) {
         retryableCallSettings.add(settingsView);
       }
     }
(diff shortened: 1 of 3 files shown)


schmidt-sebastian:
Ack. We are doing something similar in the Node client already and retry batchGetDocuments & runQuery is we haven't yet responded to the user.

igorbernstein2:
@schmidt-sebastian, this is default out of box behavior. In case you are interested, you can easily extend it to resume streams as well. For example you can set a custom [StreamResumptionStrategy](https://github.com/googleapis/gax-java/blob/master/gax/src/main/java/com/google/api/gax/retrying/StreamResumptionStrategy.java) that counts the number of results seen and if the stream is interrupted, modify the original request's offset and have gapic/gax seamlessly resume the result stream

igorbernstein2:
@pongad can you review?

src/main/java/com/google/api/codegen/viewmodel/StaticLangStubSettingsView.java
```diff
@@ -51,6 +51,19 @@
     return unaryCallSettings;
   }
 
+  /** ApiCalls that send exactly one request. This includes unary and server streaming apis. */
+  public List<ApiCallSettingsView> unaryRequestCallSettings() {
```
garrettjonesgoogle: What exactly does the `request` part of this mean, and how does it distinguish this case from `unaryCallSettings`? The name doesn't feel right because it has `unary` in it, and it returns settings for server streaming methods.

igorbernstein2: a. unary rpcs are 1:1 request/response
b. server streaming are 1:n
c. client streaming are n:1
d. bidi are n:m

I was trying group cases a & b "RPCs that send exactly one request before receiving N responses". Would `singleRequestRpcCallSettings` be better?

Naming is hard :(

igorbernstein2:
All feedback has been addressed

PR #2168, opened 2018-07-25
scanner: fewer escape sequences

pongad:
Some characters like backspace and form feeds seem unlikely to appear in
samples. Many languages also don't have these sequences so strings will
render uglily.


pongad:
@vchudnov-g They exist in Java, Go, and Python. I'm fine adding those back. I assume you'd rather we have them?

vchudnov-g:
That would be my preference, but i don't feel strongly. Up to you.

Issue #2090, opened 2018-06-19
Configgen: Gracefully handle page streaming resource field heuristic


eoogbe:
The config generator will fail with an error if the page streaming resources field cannot be heuristically determined. Throwing an exception breaks config generation for libraries with multiple repeating fields in the response of their page streaming methods.

Since the determination is a heuristic, it should use the best guess for Cloud conventions and add a FIXME comment to ensure the config is accurate.

eoogbe:
Found a workaround, bumping to P2

PR #2169, opened 2018-07-26
Improve page streaming heuristic

andreamlin:
Fixes https://github.com/googleapis/gapic-generator/issues/2090.

All current protos in googleapis follow a convention where the very first repeated field in a List[Resource]Response message is the page streaming object, so this PR uses that as a heuristic.

I've run this PR on all artman_*.yaml configs in googleapis that have a `gapic_config` target, and they all succeed. Therefore, this fixes https://circleci.com/gh/googleapis/googleapis/1312, where the `gapic_config` target fails on the monitoring API.
For whatever reason, this also fixes the `gapic_config` target for the bigtableadmin API that had been chronically failing in the googleapis smoke test (see it exempt from the smoke test [here](https://github.com/googleapis/artman/blob/e1fd2255ccfe8f500f2b0d4637e8f6278669590c/test/smoketest_artman.py#L42)).

### DIFF
--- a/src/main/java/com/google/api/codegen/configgen/ProtoPageStreamingTransformer.java
+++ b/src/main/java/com/google/api/codegen/configgen/ProtoPageStreamingTransformer.java
@@ -71,17 +71,14 @@ public class ProtoPageStreamingTransformer implements PageStreamingTransformer {
         continue;
       }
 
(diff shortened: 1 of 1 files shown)


andreamlin:
PTAL

PR #2172, opened 2018-07-26
"Parse" the whole init config left-to-right

pongad:
This will let us incrementally create initialization tree later.
The tree-descending code can then be used to work with sample
parameters.

The use of field name as a map key is regrettable. This can be fixed
later.

The RHS being arbitrary string is also regrettable. We'll migrate
slowly.

### DIFF
--- a/src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
+++ b/src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
@@ -63,7 +63,7 @@ public class FieldStructureParser {
     return valueConfig;
   }
 
-  // Parses `pathconfig` to construct the `InitCodeNode` it specifies. `config` must be a valid
+  // Parses `config` to construct the `InitCodeNode` it specifies. `config` must be a valid
   // config
   // satisfying the eBNF grammar below:
   //
@@ -148,8 +148,7 @@ public class FieldStructureParser {
 
       // TODO(pongad): Quote the RHS of existing configs, and once that's done use 'parseValue' here
       // (`String valueString = parseValue(scanner)`). For now we are preserving the previous
-      // behavior,
-      // where everything on the right of the equal sign is a string.
+      // behavior, where everything on the right of the equal sign is a string.
       String valueString = config.substring(scanner.pos());
 
       if (valueString.contains(InitFieldConfig.RANDOM_TOKEN)) {
--- a/src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
+++ b/src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
@@ -26,6 +26,7 @@ import java.util.Map;
  * will actually be a list.
  */
 public class FieldStructureParser {
+  private static final String PROJECT_ID_TOKEN = "$PROJECT_ID";
 
   public static InitCodeNode parse(String initFieldConfigString) {
     return parse(initFieldConfigString, ImmutableMap.<String, InitValueConfig>of());
@@ -33,9 +34,7 @@ public class FieldStructureParser {
 
   public static InitCodeNode parse(
       String initFieldConfigString, Map<String, InitValueConfig> initValueConfigMap) {
-    InitFieldConfig fieldConfig = InitFieldConfig.from(initFieldConfigString);
-    InitValueConfig valueConfig = createInitValueConfig(fieldConfig, initValueConfigMap);
-    return parsePartialDottedPathToInitCodeNode(fieldConfig.fieldPath(), valueConfig);
+    return parseConfig(initFieldConfigString, initValueConfigMap);
   }
 
   private static InitValueConfig createInitValueConfig(
@@ -65,36 +64,38 @@ public class FieldStructureParser {
   }
 
   // Parses `path` to construct the `InitCodeNode` it specifies. `path` must be a valid config
-  // satisfying the recursive grammar below:
-  // config:
-  //   ident
-  //   config '.' ident
-  //   config '[' number ']'
-  //   config '{' number '}'
-  //   config '{' string '}'
-  //   config '{' ident '}' (for compatibility, the identifier is just treated as a string)
-  private static InitCodeNode parsePartialDottedPathToInitCodeNode(
-      String path, InitValueConfig initValueConfig) {
-    Scanner scanner = new Scanner(path);
+  // satisfying the eBNF grammar below:
+  //
+  // config = path ['=' value];
+  // path = ident pathElem* ['%' ident]
+  // pathElem = ('.' ident) | ('[' int ']') | ('{' value '}');
+  // value = int | string | ident;
+  //
+  // For compatibility with the previous parser, when ident is used as a value, the value is
+  // the name of the ident.
+  private static InitCodeNode parseConfig(
+      String config, Map<String, InitValueConfig> initValueConfigMap) {
+    Scanner scanner = new Scanner(config);
 
     Preconditions.checkArgument(
-        scanner.scan() == Scanner.IDENT, "expected root identifier: %s", path);
+        scanner.scan() == Scanner.IDENT, "expected root identifier: %s", config);
     InitCodeNode root = InitCodeNode.create(scanner.token());
 
     InitCodeNode parent = root;
+    int token;
+
+    pathElem:
     while (true) {
-      int token = scanner.scan();
+      token = scanner.scan();
       switch (token) {
+        case '%':
+        case '=':
         case Scanner.EOF:
-          if (initValueConfig != null) {
-            parent.updateInitValueConfig(initValueConfig);
-            parent.setLineType(InitCodeLineType.SimpleInitLine);
-          }
-          return root;
+          break pathElem;
 
         case '.':
           Preconditions.checkArgument(
-              scanner.scan() == Scanner.IDENT, "expected identifier after '.': %s", path);
+              scanner.scan() == Scanner.IDENT, "expected identifier after '.': %s", config);
           parent.setLineType(InitCodeLineType.StructureInitLine);
           InitCodeNode child = InitCodeNode.create(scanner.token());
           parent.mergeChild(child);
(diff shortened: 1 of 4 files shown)


pongad:
cc @beccca

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
```diff
@@ -103,33 +104,92 @@ private static InitCodeNode parsePartialDottedPathToInitCodeNode(
...
+
+    InitValueConfig valueConfig =
+        createInitValueConfig(
+            InitFieldConfig.newBuilder()
+                .fieldPath(config.substring(0, fieldNamePos))
+                .entityName(entityName)
+                .value(initValue)
+                .build(),
+            initValueConfigMap);
+    if (valueConfig != null) {
```
vchudnov-g: What would `valueConfig == null` mean? Should we throw?

src/main/java/com/google/api/codegen/transformer/go/GoSurfaceNamer.java
```diff
@@ -441,7 +442,11 @@ public String getSmokeTestClassName(InterfaceConfig interfaceConfig) {
...
   @Override
   public String injectRandomStringGeneratorCode(String randomString) {
-    return randomString.replace(
-        InitFieldConfig.RANDOM_TOKEN, "\" + strconv.FormatInt(time.Now().UnixNano(), 10) + \"");
+    randomString =
+        CommonRenderingUtil.stripQuotes(randomString)
+            .replace(
+                InitFieldConfig.RANDOM_TOKEN,
+                "\" + strconv.FormatInt(time.Now().UnixNano(), 10) + \"");
+    return "\"" + randomString + "\"";
```
vchudnov-g: If the configs haven't been ported to have quotes RHSs (as I gather from your TODO in the other file), will this cause too many quotes to occur right now?

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
```diff
@@ -65,36 +64,38 @@ private static InitValueConfig createInitValueConfig(
...
+  // satisfying the eBNF grammar below:
+  //
+  // config = path ['=' value];
+  // path = ident pathElem* ['%' ident]
+  // pathElem = ('.' ident) | ('[' int ']') | ('{' value '}');
+  // value = int | string | ident;
+  //
+  // For compatibility with the previous parser, when ident is used as a value, the value is
+  // the name of the ident.
+  private static InitCodeNode parseConfig(
```
vchudnov-g: Do we have a pre-existing unit test for this that continues to pass with your changes?

I presume that since baselines aren't changed, the uses that we capture remain unaffected....but it would be good to capture the intended behavior in a unit test.

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
```diff
@@ -65,36 +64,38 @@ private static InitValueConfig createInitValueConfig(
...
-  //   ident
-  //   config '.' ident
-  //   config '[' number ']'
-  //   config '{' number '}'
-  //   config '{' string '}'
-  //   config '{' ident '}' (for compatibility, the identifier is just treated as a string)
-  private static InitCodeNode parsePartialDottedPathToInitCodeNode(
-      String path, InitValueConfig initValueConfig) {
-    Scanner scanner = new Scanner(path);
+  // satisfying the eBNF grammar below:
```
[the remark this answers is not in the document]
pongad: Thank you. Done.

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
```diff
@@ -103,33 +104,92 @@ private static InitCodeNode parsePartialDottedPathToInitCodeNode(
...
-              String.format("unexpected character '%c': %s", token, path));
+              String.format("unexpected character '%c': %s", token, config));
       }
     }
+
+    int fieldNamePos = config.length();
+
+    String entityName = null;
+    if (token == '%') {
+      fieldNamePos = scanner.pos() - 1;
```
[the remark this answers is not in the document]
pongad: You're right that the scanner points to the next token. So after lexing `%`, the position is the character after the `%`. We want the substring *before* the `%` though, so the `-1` is there to skip the `%`.

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
```diff
@@ -103,33 +104,92 @@ private static InitCodeNode parsePartialDottedPathToInitCodeNode(
...
+      token = scanner.scan();
+    }
+
+    InitValue initValue = null;
+    if (token == '=') {
+      fieldNamePos = Math.min(fieldNamePos, scanner.pos() - 1);
+
+      // TODO(pongad): Currently the RHS of equal sign is just arbitrary run of text treated as string, eg
+      //   a.b=https://foo.bar.com/zip/zap
+      // We'll quote the RHS of existing configs, then we can use parseValue here.
```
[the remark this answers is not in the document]
pongad: Yes. Thank you, English is hard.

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
```diff
@@ -103,33 +104,92 @@ private static InitCodeNode parsePartialDottedPathToInitCodeNode(
...
+      // For now, just preserve old behavior.
+      // String valueString = parseValue(scanner);
+      String valueString = config.substring(scanner.pos());
+
+      if (valueString.contains(InitFieldConfig.RANDOM_TOKEN)) {
+        System.err.println(valueString);
```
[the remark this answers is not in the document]
pongad: Derp

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
```diff
@@ -103,33 +104,92 @@ private static InitCodeNode parsePartialDottedPathToInitCodeNode(
...
+        initValue = InitValue.createRandom(valueString);
+      } else if (valueString.contains(PROJECT_ID_TOKEN)) {
+        Preconditions.checkArgument(
+            valueString.equals(PROJECT_ID_TOKEN),
+            "%s cannot be used as substring: %s",
```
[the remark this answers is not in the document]
pongad: `RANDOM_TOKEN` can be used as a substring. For example: `foo-$RANDOM` might become `fmt.Sprintf("foo-%s", randomNumber())`.

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
```diff
@@ -103,33 +104,92 @@ private static InitCodeNode parsePartialDottedPathToInitCodeNode(
...
+            PROJECT_ID_TOKEN,
+            config);
+        initValue = InitValue.createVariable(InitFieldConfig.PROJECT_ID_VARIABLE_NAME);
+      } else {
+        initValue = InitValue.createLiteral(valueString);
+      }
+      token = scanner.scan();
```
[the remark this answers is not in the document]
pongad: Done.

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
pongad: I don't completely understand this code myself. IIUC, if `valueConfig == null`, somewhere we put in the default value: zero, empty string, empty list, struct with all zero fields, etc.

src/main/java/com/google/api/codegen/transformer/go/GoSurfaceNamer.java
pongad: No because we `stripQuotes` on L446 right?

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
pongad: [`SampleInitCodeTest`](https://github.com/googleapis/gapic-generator/blob/master/src/test/java/com/google/api/codegen/metacode/SampleInitCodeTest.java) tests this. I made an intentional bug and verified that the test fails :)

pongad:
@vchudnov-g PTAL

src/main/java/com/google/api/codegen/transformer/go/GoSurfaceNamer.java
vchudnov-g: And if something didn't have quotes, adding them here should be safe, correct? OK.

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
vchudnov-g: :+1:

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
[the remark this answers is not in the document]
vchudnov-g: Oh, so `createRandom` causes the string around the token to be preserved? OK. (I jumped to conclusions and assume dit replaced the whole string.) 

Out of curiosity, do you happen to know where this occurs?

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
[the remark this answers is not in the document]
vchudnov-g: I'm still confused: the `scan()` below is checking for an identifier _after_ the `%`, right?

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
```diff
@@ -63,7 +63,8 @@ private static InitValueConfig createInitValueConfig(
     return valueConfig;
   }
 
-  // Parses `path` to construct the `InitCodeNode` it specifies. `path` must be a valid config
+  // Parses `pathconfig` to construct the `InitCodeNode` it specifies. `config` must be a valid
```
vchudnov-g: s/`pathconfig`/`config`/  :smile:

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
```diff
@@ -143,15 +146,13 @@ private static InitCodeNode parseConfig(
...
       fieldNamePos = Math.min(fieldNamePos, scanner.pos() - 1);
 
-      // TODO(pongad): Currently the RHS of equal sign is just arbitrary run of text treated as string, eg
-      //   a.b=https://foo.bar.com/zip/zap
-      // We'll quote the RHS of existing configs, then we can use parseValue here.
-      // For now, just preserve old behavior.
-      // String valueString = parseValue(scanner);
+      // TODO(pongad): Quote the RHS of existing configs, and once that's done use 'parseValue' here
+      // (`String valueString = parseValue(scanner)`). For now we are preserving the previous
+      // behavior,
```
vchudnov-g: nit: line breaks

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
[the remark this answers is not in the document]
pongad: You mean when it's used? I think it's only used to smoke test artman right now :(

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
pongad: GAH

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
pongad: Hmm, google-java-format seems to do a poor job here :(

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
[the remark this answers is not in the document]
vchudnov-g: I mean where the substitution happens. It does not seem to be in `createRandom` itself, which is just an Autovalue class.

PR #2173, opened 2018-07-27
don't use regex for stripQuotes

pongad:
It's easy enough to by hand.
This also reduces the (rounded) test time on my machine from 18s to 17s :)

### DIFF
--- a/src/main/java/com/google/api/codegen/util/CommonRenderingUtil.java
+++ b/src/main/java/com/google/api/codegen/util/CommonRenderingUtil.java
@@ -19,22 +19,18 @@ import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 /** Utility class to process text in the templates. */
 public class CommonRenderingUtil {
-  private static Pattern singleQuoteStringPattern = Pattern.compile("'([^\\\']*)'");
-  private static Pattern doubleQuoteStringPattern = Pattern.compile("\"([^\\\"]*)\"");
 
   /** Strips the surrounding quotes from the given string */
   public static String stripQuotes(String value) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/api/codegen/util/CommonRenderingUtil.java
```diff
@@ -19,22 +19,18 @@
...
-    Matcher singleQuoteMatcher = singleQuoteStringPattern.matcher(value);
-    Matcher doubleQuoteMatcher = doubleQuoteStringPattern.matcher(value);
-    if (singleQuoteMatcher.matches()) {
-      value = singleQuoteMatcher.group(1);
-    } else if (doubleQuoteMatcher.matches()) {
-      value = doubleQuoteMatcher.group(1);
+    for (char quoteChar : new char[] {'\'', '"'}) {
+      if (value.length() >= 2
+          && value.charAt(0) == quoteChar
+          && value.indexOf(quoteChar, 1) == value.length() - 1) {
```
vchudnov-g: What if the string ends with an escaped quote `\"? Do we just assume that the since there's a beginning quote, the last quote ends the string and the preceding slash is just a slash?

pongad: I think so. Our code base doesn't handle strings very well. I think it's better to keep old behavior until we understand it more.

vchudnov-g: ie something like  `"This is my string\"`

pongad: Yeah I'm 100% sure we'd return `This is my string \` in both the old and new implementations.

vchudnov-g: If it's the same behavior as the old code, sgtm.

I would love to have a test to capture this behavior. Would you be able to add one?

PR #2175, opened 2018-07-31
Circle ci smoketest (WIP)

andreamlin:

Issue #2160, opened 2018-07-20
Generated output for KMS v1 has bad import name (ends in _grpc but shouldn't)


crwilcox:
Replacing "iam_policy_pb2_grpc" with "iam_policy_pb2" resolves issues

```
--- a/kms/google/cloud/kms_v1/gapic/transports/key_management_service_grpc_transport.py
+++ b/kms/google/cloud/kms_v1/gapic/transports/key_management_service_grpc_transport.py
@@ -15,7 +15,7 @@
 import google.api_core.grpc_helpers

 from google.cloud.kms_v1.proto import service_pb2_grpc
-from google.iam.v1 import iam_policy_pb2_grpc
+from google.iam.v1 import iam_policy_pb2


 class KeyManagementServiceGrpcTransport(object):
@@ -67,7 +67,7 @@ class KeyManagementServiceGrpcTransport(object):
             'key_management_service_stub':
             service_pb2_grpc.KeyManagementServiceStub(channel),
             'iam_policy_stub':
-            iam_policy_pb2_grpc.IAMPolicyStub(channel),
+            iam_policy_pb2.IAMPolicyStub(channel),
         }

     @classmethod
diff --git a/kms/synth.py b/kms/synth.py
index d5cee177fe..56d0e0cdf6 100644
--- a/kms/synth.py
+++ b/kms/synth.py
@@ -39,3 +39,5 @@ s.replace('setup.py',
           f"\\1'{release_status}'")
 s.replace('setup.py', 'version = .*', f"version = '{client_library_version}'")

+# Wrong import name. Drop _grpc
+s.replace("**/*.py", "iam_policy_pb2_grpc", "iam_policy_pb2")
```

andreamlin:
I tried to regen python pubsub and the same thing came up.

PR #2177, opened 2018-08-01
Python unit tests

andreamlin:
Built on top of #2175.

Test generated Python clients with `nox`. 
Test generated unit tests, and skip testing system tests (#2205 will set up credentials for system tests) for later.

This only tests PubSub and Speech, because Logging generation is broken by #2209


andreamlin:
Blocked on #2197, #2160 which break the generated python client :(

andreamlin:
PTAL

gapic_smoketest.py
```diff
@@ -29,85 +29,57 @@
...
-    cmds = ("%s/gradlew build test" % os.getcwd()).split()
-    exit_code = subprocess.call(cmds, stdout=log, stderr=log)
-    if exit_code:
-        return exit_code
-
-    cmds = ("%s/gradlew -p %s build test" % (os.getcwd(), gapic_dir)).split()
-    return subprocess.call(cmds, stdout=log, stderr=log)
 
-def run_smoke_test(api_name, language, root_dir, log, user_config):
+def run_smoke_test(root_dir, log, user_config):
```
landrito: This method name seems inaccurate.

gapic_smoketest.py
```diff
@@ -29,85 +29,57 @@
...
+                msg = 'Succeded to generate %s of %s.' % (
+                    target, artman_yaml_path)
+                success.append(msg)
+            logger.info(msg)
     logger.info('================ Smoketest summary ================')
     if not warning and not failure:
         logger.info('Successes:')
-        for msg in success:
-            logger.info(msg)
+        if warning:
```
landrito: This is strange. This conditional is wrapped in its opposite.

.circleci/config.yml
```diff
@@ -34,77 +24,88 @@ jobs:
...
             export RUNNING_IN_ARTMAN_DOCKER=True
             rm -rf gapic-generator/.git/
-            gapic-generator/gradlew -p gapic-generator fatJar createToolPaths install build -x test
+            gapic-generator/gradlew -p gapic-generator fatJar createToolPaths install build -x test -x javadoc
       - run:
-          name: Make reports dir
+          name: Make reports directory
           command: mkdir -p reports
+      - run:
+          name: Generate all client libraries
```
landrito: I am leaning towards thinking it would be better for this to split into generating the client libraries per language. My rationale is that running the test for an API in python should not depend on whether the java generation failed. Thoughts?

andreamlin: That makes sense. artman requires the artman docker image, so generating the clients has to happen in seperate jobs and containers from testing the clients. I think it would be a bit much to spin up 7 containers with artman to generate all the language clients, so one optimization could be to conditionally run the test-language-client job when the generation in that language passes.

I'll try that approach out, and hopefully that fits the model where "running the test for an API in python should not depend on whether the java generation failed".

gapic_smoketest.py
andreamlin: Ok, renamed to `generate_clients`

andreamlin:
Ok, only running tests for a language if the language's client generation succeeded. There will be calls to artman to (try to) generate all language clients, without short circuiting if any generation fails.

PTAL

gapic_smoketest.py
```diff
@@ -29,85 +29,50 @@
...
             failure.append(msg)
         else:
-            msg = 'Tests passed for %s library.' % (
-                language)
+            msg = 'Succeded to generate %s of %s.' % (
+                target, artman_yaml_path)
             success.append(msg)
-        os.chdir(cwd)
         logger.info(msg)
     logger.info('================ Smoketest summary ================')
```
landrito: Does `Smoketest` still apply here? It seems like this is just now generation messages.

andreamlin: Done.

andreamlin:
Ok, changed gapic_smoketest.py to generate_clients.py

PTAL

PR #2189, opened 2018-08-03
Support custom region tags in samples

vchudnov-g:
- The default region tag is "sample"
- Custom region tags can be specified in the GAPIC YAML. The region tag specification accepts (but does not require) `%m`, `%c`, and `%v` placeholders for method name, calling form name, and value set id.
- If the sample specification would result in differently configured samples being written to the same path, an exception is thrown. This can happen if the same value set and calling form combo is specified more than once (eg with wildcards) with different region names.

### DIFF
--- a/src/main/java/com/google/api/codegen/config/SampleSpec.java
+++ b/src/main/java/com/google/api/codegen/config/SampleSpec.java
@@ -22,10 +22,8 @@ import com.google.api.codegen.util.Name;
 import com.google.api.codegen.viewmodel.CallingForm;
 import com.google.auto.value.AutoValue;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Set;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 
@@ -85,11 +83,10 @@ public class SampleSpec {
     sampleConfiguration = methodConfigProto.getSamples();
     valueSets = methodConfigProto.getSampleValueSetsList();
 
-    HashMap<String, SampleValueSet> setMap = new HashMap<>();
+    HashSet<String> ids = new HashSet<>();
     for (SampleValueSet valueSet : valueSets) {
       String id = valueSet.getId();
-      SampleValueSet oldSet = setMap.put(id, valueSet);
-      if (oldSet != null) {
+      if (!ids.add(id)) {
         throw new IllegalArgumentException(
             String.format(
                 "in method \"%s\": duplicate value set id: \"%s\"",
@@ -138,14 +135,12 @@ public class SampleSpec {
 
     // Construct a `ValueSetAndTags` for each sample specified in each element of `matchingSamples`.
     List<ValueSetAndTags> result = new ArrayList<>();
-    Set<String> generated = new HashSet<>();
     for (SampleValueSet vset : valueSets) {
       for (SampleTypeConfiguration sample : matchingSamples) {
         for (String valueSetExpression : sample.getValueSetsList()) {
           if (expressionMatchesId(valueSetExpression, vset.getId())) {
             result.add(
                 ValueSetAndTags.newBuilder().values(vset).regionTag(sample.getRegionTag()).build());
-            generated.add(vset.getId());
           }
         }
       }
--- a/src/main/java/com/google/api/codegen/config/SampleSpec.java
+++ b/src/main/java/com/google/api/codegen/config/SampleSpec.java
@@ -18,25 +18,30 @@ import com.google.api.codegen.MethodConfigProto;
 import com.google.api.codegen.SampleConfiguration;
 import com.google.api.codegen.SampleConfiguration.SampleTypeConfiguration;
 import com.google.api.codegen.SampleValueSet;
+import com.google.api.codegen.util.Name;
 import com.google.api.codegen.viewmodel.CallingForm;
-import com.google.api.codegen.viewmodel.ClientMethodType;
-import com.google.common.collect.ImmutableList;
+import com.google.auto.value.AutoValue;
+import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import java.util.stream.Collectors;
+import javax.annotation.Nullable;
 
 /**
- * Class SampleSpec stores the sample specification for a given method, and provides methods to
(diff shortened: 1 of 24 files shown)


src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
```diff
@@ -195,37 +197,67 @@ public void generateSamples(
...
       }
     }
     return methodSampleViews;
   }
 
+  /**
+   * Creates a region tag from the spec, replacing any {@code %m}, {@code %c}, and {@code %v}
+   * placeholders with the method name, calling form name, and value set id, respectively.
+   */
+  private String regionTagFromSpec(
```
hzyi-google: this method could be declared static?

src/main/java/com/google/api/codegen/transformer/SampleFileRegistry.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * {@code SampleFileRegistry} is used to verify that the samples we generate with different
+ * parameters have different paths, so they don't clobber each other.
+ */
+public class SampleFileRegistry {
+
+  Map<String, SampleInfo> files = new HashMap<>();
```
hzyi-google: Can you make this `private final`?

src/main/java/com/google/api/codegen/config/SampleSpec.java
```diff
@@ -46,16 +51,49 @@
...
+    }
+  }
+
   public SampleSpec(MethodConfigProto methodConfigProto) {
     specified = methodConfigProto.hasSamples();
     sampleConfiguration = methodConfigProto.getSamples();
     valueSets = methodConfigProto.getSampleValueSetsList();
 
-    HashSet<String> ids = new HashSet<>();
+    HashMap<String, SampleValueSet> setMap = new HashMap<>();
```
pongad: I don't think this loop has to change at all. We're not actually using the `valueSet` themselves for anything here right?

src/main/java/com/google/api/codegen/config/SampleSpec.java
```diff
@@ -101,21 +134,26 @@ private static boolean expressionMatchesId(String expression, String id) {
...
-        .stream()
-        .filter(
-            valueSet ->
-                valueSetExpressions
-                    .stream()
-                    .anyMatch(expression -> expressionMatchesId(expression, valueSet.getId())))
-        .collect(ImmutableList.toImmutableList());
+    // Construct a `ValueSetAndTags` for each sample specified in each element of `matchingSamples`.
+    List<ValueSetAndTags> result = new ArrayList<>();
+    Set<String> generated = new HashSet<>();
```
pongad: This seems unused?

src/main/java/com/google/api/codegen/transformer/SampleFileRegistry.java
```diff
@@ -0,0 +1,103 @@
...
+              .method(method)
+              .callingForm(callingForm)
+              .valueSet(valueSet)
+              .regionTag(regionTag)
+              .build());
+    } else if (!(previous.path().equals(path)
+        && previous.method().equals(method)
+        && previous.callingForm().equals(callingForm)
+        && previous.valueSet().equals(valueSet)
+        && previous.regionTag().equals(regionTag))) {
```
pongad: Just use `this.equals(previous)`, `@AutoValue` already makes one for us.

src/main/java/com/google/api/codegen/transformer/SampleFileRegistry.java
```diff
@@ -0,0 +1,103 @@
...
+      throw new IllegalArgumentException(
+          String.format(
+              "conflicting configurations for sample \"%s\":\n"
+                  + "%s\n%s\n"
+                  + "  {method:\"%s\", calling form:\"%s\", value set:\"%s\", region tag:\"%s\"}\n"
+                  + "  {method:\"%s\", calling form:\"%s\", value set:\"%s\", region tag:\"%s\"}\n",
```
pongad: Just use `toString()`, AutoValue already implements it.

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
```diff
@@ -195,37 +197,67 @@ public void generateSamples(
...
+      String spec, String methodName, CallingForm callingForm, String valueSetId) {
+    final String DEFAULT_REGION_SPEC = "sample";
+
+    if (spec == null || spec.isEmpty()) {
```
pongad: Prefer Guava's `isNullOrEmpty`

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
vchudnov-g: done

src/main/java/com/google/api/codegen/transformer/SampleFileRegistry.java
vchudnov-g: done

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
vchudnov-g: Used `Strings.isNullOrEmpty`

src/main/java/com/google/api/codegen/transformer/SampleFileRegistry.java
vchudnov-g: OK, but at the price of always building a new `SampleInfo`

src/main/java/com/google/api/codegen/transformer/SampleFileRegistry.java
vchudnov-g: OK, but at the price of always building a new `SampleInfo`

src/main/java/com/google/api/codegen/config/SampleSpec.java
vchudnov-g: Good catch! This was cruft from when I was checking for uniqueness here.

src/main/java/com/google/api/codegen/config/SampleSpec.java
vchudnov-g: Good catch! Cruft from an earlier approach

vchudnov-g:
PTAL

PR #2199, opened 2018-08-03
Add path_template import when needed.

andreamlin:
Generated python clients need a path_template import when there are formatResourceFunctions (see [main.snip](https://github.com/googleapis/gapic-generator/blob/c984cc4326f0d883ece55d20374edbe182e081f0/src/main/resources/com/google/api/codegen/py/main.snip#L65)) , and the formatResourceFunctions is indirectly populated by getSingleResourceNameConfigsUsedByInterface(context) (see [PathTemplateTransformer.java](https://github.com/googleapis/gapic-generator/blob/c984cc4326f0d883ece55d20374edbe182e081f0/src/main/java/com/google/api/codegen/transformer/PathTemplateTransformer.java#L183)).

An alternative is to just always import `path_template`, but I figure it's better to do expensive computation in the code generator than to import more than necessary on the client developer side.


andreamlin:
There aren't tests for this right now, but https://github.com/googleapis/gapic-generator/pull/2177 would ensure that generated python clients pass their unit tests.

PR #2202, opened 2018-08-04
Don't use _pb2_grpc for IAMPolicy

andreamlin:
Fixes https://github.com/googleapis/gapic-generator/issues/2160.


andreamlin:
Source of documentation for why the IAM service is special and not actually used and must be implemented by API producers would be helpful to put in the comments, if anyone knows of any.

PTAL

andreamlin:
This PR has been used to locally generate a python Pubsub client. There aren't any tests, but once https://github.com/googleapis/gapic-generator/pull/2177 is checked in, running the generated python unit tests would ensure that we don't introduce any bugs in the generated code.

PR #2203, opened 2018-08-04
Don't throw error on multiple version strings

### LINKED ISSUE #2192: Support names that look like version numbers in Ruby
The Ruby generator will throw an exception if it finds a fully qualified name that includes more than one segment that looks like a version. For example: `my.service.ServiceV1.v1`.

jbolinger:
Resolves #2192 by removing the sanity check for multiple version like strings.

This should unblock: https://github.com/googleapis/artman/issues/483

I wanted to convert this to a warning but it appears that there is no logger used for this project?

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/ruby/RubyPackageMetadataTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/ruby/RubyPackageMetadataTransformer.java
@@ -331,9 +331,10 @@ public class RubyPackageMetadataTransformer implements ModelToViewTransformer<Pr
     for (InterfaceModel apiInterface : interfaces) {
       versionNamespaces.add(namer.getNamespace(apiInterface));
     }
-    if (versionNamespaces.size() > 1) {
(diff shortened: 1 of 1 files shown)


landrito:
Hm. I don't think this change actually works for the package name that is mentioned in the issue:

`my.service.ServiceV1.v1` will return `ServiceV1` where I think the intention is to return `v1`

jbolinger:
Ah, you might be correct but it turns out the issue was wrong in the first place.

I was testing on `google/devtools/containeranalysis/artman_containeranalysis_v1beta1.yaml` and it looks like it does produce the desired result. This particular method gets two values:

`Google::Devtools::Containeranalysis::V1beta1`
`Grafeas::V1beta1`

because there are actually two different namespaces (so it has nothing to do with the example that I used) and taking the first seems to be the right thing to do.

landrito:
There is a logger that is built into the templating system / api compiler but I don't think it has been sent down to all of these transformers.

PR #2215, opened 2018-08-08
delete things we don't need

pongad:
View classes and their methods are deleted if there is no snippet files using them.


pongad:
@vchudnov-g What would make you comfortable? I regenerated Java client and found no regression

PR #2216, opened 2018-08-08
Samplegen response handling spec: s/output_spec/on_success/

vchudnov-g:
Also, better document the OutputSpec proto fields.


beccca:
👍 like the print/define comments

PR #2217, opened 2018-08-09
delete unused SurfaceNamer methods

pongad:
unused methods detected by the attached script.


testutils/unused-namer.sh
```diff
@@ -0,0 +1,28 @@
...
+#   common lines
+# -2 -3 suppresses the last two.
+#
+# arg1 is names of methods in SurfaceNamers
+# arg2 is all identifiers used outside of namers
+#
+# so we end up with methods defined in SurfaceNamer that
+# we don't use anywhere.
+#
+# NOTE(pongad): This misses overloads and helper methods used by namers themselves.
```
vchudnov-g: IIUC:

(1) I think this is correct:
"For overloads, this will flag when none are used. If only some are used, it won't determine which ones are or aren't used."

(2) I'm not sure what you mean for the helper methods. I think it might be this:
"For helper methods used only in the namer classes, this may incorrectly say they are not used at all, when in reality they are not used outside the namers but may be used in the namers"

And finally, a third case:
(3) If a namer identifier is not used anywhere, but some other class also happens to define the same identifier, this will incorrectly  flag the namer identifier as being used.

(1) and (3) are safe; it means we may not be as aggressive in deleting stuff as we could have been with a more sophisticated analysis, but we won't break anything.

(2) is not safe, but those deletions will cause a compile-time error. So as  long as we compile before merging, it should be OK.

pongad: Done

pongad:
@vchudnov-g PTAL

PR #2218, opened 2018-08-09
delete unused *View methods

pongad:


michaelbausor:
This looks great! Is there a good way we can validate that we haven't made a mistake somewhere, and one of these methods is actually used in an obscure API somewhere? Or do we need to rely on code review of the bash script?

pongad:
I guess you *could* grep for a few things I'm deleting and convince yourself that snip files don't use them. That's exactly what I did :)

I suppose it's possible that grep is broken, but we need to start trusting something somewhere :P

PR #2221, opened 2018-08-10
rewire samplegen neurons regarding sample parameters

pongad:
Python: The values to be passed in by function parameter is segregated into their own section.
Eventually they should be commented out, so that they are visible for documentation,
but won't clobber to values passed in.

Java, PHP, Node: The values are deleted for now. Eventually we'd want to render them commented out.

The function declaration itself is not implemented yet.


src/test/java/com/google/api/codegen/testsrc/library_gapic.yaml
```diff
@@ -247,6 +247,8 @@ interfaces:
       parameters:
       - shelf.name=Math
       - series_uuid.series_string=xyz3141592654
+      sample_args:
+      - shelf
```
beccca: Could we make this `shelf.name` – realistically args will never be a complex type, they'll be natives like `name`

src/test/java/com/google/api/codegen/testdata/java/java_library.baseline
```diff
@@ -7673,7 +7673,6 @@ public class MonologAboutBookCallableCallableStreamingClientProg {
 public class PublishSeriesFlattenedPiVersion {
   public static void main(String[] args) {
     // [START sample_core]
-    String name = "Math";
```
vchudnov-g: Could we keep the variables for now?

src/test/java/com/google/api/codegen/testsrc/library_gapic.yaml
```diff
@@ -247,6 +247,8 @@ interfaces:
       parameters:
       - shelf.name=Math
       - series_uuid.series_string=xyz3141592654
+      sample_args:
```
vchudnov-g: We haven't decided on this format. Let's sit down with the design doc and decide on one so we minimize backtracking.

src/main/resources/com/google/api/codegen/py/standalone_sample.snip
```diff
@@ -74,8 +74,14 @@
 
   client = {@apiMethod.apiModuleName}.{@apiMethod.apiClassName}()
 
+  @if sample.initCode.argDefaultLines
+    @# ARGS - These lines should be commented. The top-level vars are declared by function param.
```
vchudnov-g: They should be commented here, but used at the call site to set the actual parameters.

src/test/java/com/google/api/codegen/testsrc/library_gapic.yaml
pongad: Yeah I did the easiest thing to make progress. Sitting down sounds good.

src/main/java/com/google/api/codegen/metacode/InitCodeContext.java
```diff
@@ -65,6 +65,12 @@
   /** Returns init config strings. */
   public abstract List<String> initFieldConfigStrings();
 
+  /**
+   * When generating samples, the sample function accept these parts of the request object as
+   * argument.
```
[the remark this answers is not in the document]
pongad: Done

pongad:
@vchudnov-g PAL

src/main/resources/com/google/api/codegen/py/standalone_sample.snip
vchudnov-g: The sample file will be copy-paste runnable, so it will have a main that will call into the sample function. In Goish pseudocode ("Gpc"?):

```
import (
...
)


// GetBookSample retrieves bookName
//   bookName: Name of book to retrieve
// [START sample]
func GetBookSample(bookName string) {


  // [START sample_core]

  client := ...
  request := ...
 
  // bookName = "War and Peace"   
  request.title = bookName

   response := client.GetBook(request)
   fmt.Print(response.pages) 

  // [END sample_core]
}

// [END sample]

func main() {
   myBook := flagValue("book_name", "Name of book to retrieve", "War and Peace")
   invokeExample(myBook)
}

```

The whole file is runnable. The part between the `SAMPLE` tags will be displayed in DPE docs. The part between the `SAMPLE_core` tags will be displayed in APIx. 

Given that we include the function signature (as per [existing precedent](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/language/cloud-client/v1/snippets.py)) I would leave out the line with the commented-out `bookName` sample assignment, but my understanding is that at the moment DPE wants it in there so it's super clear. 

Regardless, we may wind up tweaking the location of the tags, but this code structure reflects my understanding of the agreed solution to the requirements.

pongad: Ah that makes sense. Is there any necessary change to this PR then? I'm not sure how to comment things out yet :/ I'm hoping to do that in a follow up if you're OK with it.

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -333,9 +333,11 @@ message SampleValueSet {
   //   several placeholders.
   repeated string parameters = 4;
 
+  repeated string sample_args = 5;
```
vchudnov-g: I'd ask for a comment, but we agreed on a new schema that you'll implement in a chaser PR. :-).

However, I would just add a very temporary TODO saying that this will go away in favor of a new schema shortly, just to avoid confusion for folks looking at it in the interim.

src/main/java/com/google/api/codegen/viewmodel/InitCodeView.java
```diff
@@ -19,6 +19,8 @@
 
 @AutoValue
 public abstract class InitCodeView {
+  public abstract List<InitCodeLineView> argDefaultLines();
```
vchudnov-g: Could you add comments to these getters, particularly the top three? It's not immediately clear from the names what the differences are.

src/main/resources/com/google/api/codegen/py/standalone_sample.snip
vchudnov-g: Some languages (I assume all) have template functions to comment out lines of code, so we can use that. I'm fine if you want to defer this to the chaser PR, but add a `FIXME` line here to say that change will be coming shortly.

src/main/resources/com/google/api/codegen/java/initcode.snip
```diff
@@ -1,7 +1,17 @@
 @extends "java/common.snip"
 
 @snippet initCode(initCodeSpec)
-  @join line : initCodeSpec.lines
+  @if initCodeSpec.argDefaultLines
+    // should be commented to not clobber args
```
vchudnov-g: Modify the templates for all languages so that this comment (and the closing one) have a `FIXME` prefix so we remember to address this.

(FYI, in the `.snip`s, I'm leaving `TODO` as something the sample user needs to do, and `FIXME` for things we have to do to get this feature out the door.)

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -232,22 +232,41 @@ private ClientTestAssertView createAssertView(
...
+        .sampleArgStrings()
+        .stream()
+        .flatMap(sampleArg -> root.subTree(sampleArg).listInInitializationOrder().stream())
+        .collect(Collectors.toList());
   }
 
   private InitCodeView buildInitCodeView(
-      MethodContext context, Iterable<InitCodeNode> orderedItems, Iterable<InitCodeNode> argItems) {
+      MethodContext context,
+      List<InitCodeNode> orderedItems,
```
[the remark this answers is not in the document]
pongad: Done

src/main/proto/com/google/api/codegen/config.proto

src/main/java/com/google/api/codegen/viewmodel/InitCodeView.java
pongad: Done

src/main/resources/com/google/api/codegen/java/initcode.snip
pongad: done

pongad:
@vchudnov-g PTAL

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -241,8 +241,12 @@ private InitCodeView buildInitCodeViewFlattened(
     // Remove things already initialized by the args
     orderedItems.removeAll(argDefaults);
 
-    return buildInitCodeView(
-        context, orderedItems, ImmutableList.copyOf(root.getChildren().values()), argDefaults);
+    // NOTE(pongad): To the best of my understanding, this variable is used by dynamic languages to
+    // determine the arguments to be passed to the function. We should not remove argDefaults from
```
vchudnov-g: "to the function" means to the "API client library method"? Could you clarify. It sounds like you're talking about the sample function we're parametrizing (because that's what I'm attuned to), but that's not a pre-existing thing....

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -262,6 +266,17 @@ private InitCodeView buildInitCodeViewRequestObject(
         .collect(Collectors.toList());
   }
 
+  /**
+   * Transform {@code InitCodeNode}s into {@code InitCodeView}.
+   *
+   * @param orderedItems These nodes are converted into request-initialization code. The
+   *     initialization is "shallow": children nodes are not initialized. If children nodes should
+   *     also be initialized, callers must also include them in the list.
+   * @param argItems Used by samples for flattened methods. These nodes contain values that become
```
vchudnov-g: "flattened client library methods"

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -262,6 +266,17 @@ private InitCodeView buildInitCodeViewRequestObject(
...
+   *     arguments to the method.
+   * @param argDefaults Used by standalone samples, where the sample themselves are enclosed within
```
vchudnov-g: "where each client library method call is contained within a function that's part of the sample"

vchudnov-g: We should come up with less confusing names than `argItems` and `argDefaults`. Maybe `libraryArguments` vs `sampleFunctionDefaults`?

I'm still not clear on how the two interact, since a given parameter can be in both, right? (i.e. a flattened library method argument can also be a parameter to a sample function)

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
pongad: You're right. Done.

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
pongad: Done

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
pongad: Done

pongad: Renamed.

As far as I understand, they do not actually interact, but I could be wrong.

PR #2224, opened 2018-08-13
Python samplegen: clean up sample structure

vchudnov-g:
- remove stub cruft from the standalone sample template
- also configure a canonical example (with both region tags and response handling) for all languages


pongad:
This change seems to upset the baseline checker, because it can't find the lines like `# callingForm: odyssey`. I'm OK with the change itself but we should either revert the calling form strings or change the script.

vchudnov-g:
Good point about the checker. I didn't think to run it manually. I'll fix it.

The good news: it's working as intended!

vchudnov-g:
The calling form presubmits should now be passing. PTAL....

PR #2226, opened 2018-08-14
Ruby: Introduce Showcase CI

landrito:


.circleci/config.yml
```diff
@@ -29,6 +29,9 @@ jobs:
     docker:
       - image: circleci/openjdk:8-jdk
     working_directory: /tmp/workspace
+    environment:
+      # TODO (landrito): make this a project wide environment variable.
```
andreamlin: https://circleci.com/gh/googleapis/gapic-generator/edit#env-vars

.circleci/config.yml
```diff
@@ -62,13 +75,39 @@ jobs:
...
-          name: Generate all language libraries
+          name: Generate Pubsub Libraries
+          command: |
+            cd workspace/gapic-generator
+            python generate_clients.py \
+                google/pubsub/artman_pubsub.yaml \
+                --user-config=.circleci/artman_config.yaml \
+                --log=/tmp/workspace/reports/smoketest.log
+      - run:
+          name: Generate Logging Libraries
```
andreamlin: Since this `run` task happens sequentially, if the previous `Generate Pubsub Libraries` task fails, then the `Generate Logging libraries` task will also fail. So if the python generation fails for pubsub, we don't get to see if logging fails for java.

What do you think about setting `when: always` to each of these run configs?

generate_clients.py
```diff
@@ -129,6 +109,10 @@ def parse_args(*args):
         # Default to the artman-specified default location.
         default="~/.artman/config.yaml",
         help='Specify where the artman user config lives.')
+    parser.add_argument(
+        '--languages',
```
andreamlin: Can we use `nargs` here? https://docs.python.org/3.3/library/argparse.html#nargs

generate_clients.py
```diff
@@ -129,6 +109,10 @@ def parse_args(*args):
...
+        default='java python ruby',
+        help='Space separated list of languages to generate clients for.')
```
andreamlin: Can we use `nargs` here? https://docs.python.org/3.3/library/argparse.html#nargs

generate_clients.py
```diff
@@ -147,4 +131,4 @@ def _setup_logger(log_file):
     log = os.path.abspath(flags.log)
     user_config = os.path.abspath(os.path.expanduser(flags.user_config)) if flags.user_config else None
 
-    generate_clients(root_dir, log, user_config)
+    generate_clients(root_dir, flags.languages.split(), flags.artman_config, log, user_config)
```
andreamlin: And then if nargs was used for languages, we don't have to pass in a quoted list of space-separated lang values that gets `.split()`

showcase/ruby/showcase_integration_test.rb
```diff
@@ -0,0 +1,47 @@
+require "minitest/autorun"
```
andreamlin: Maybe have someone else review the ruby files? (If it hasn't been reviewed already)

src/test/java/com/google/api/codegen/testsrc/artman_showcase.yaml
```diff
@@ -0,0 +1,34 @@
...
+  api_name: showcase
+  api_version: v1alpha1
+  organization_name: google
+  service_yaml: showcase.yaml
+  gapic_yaml: v1alpha1/showcase_gapic.yaml
+  src_proto_paths:
+  - v1alpha1
+  proto_deps:
+  - name: google-common-protos
+artifacts:
```
andreamlin: I'm 80% sure we should be able to remove the whole `artifacts` config, since it isn't used anymore.

generate_clients.py
landrito: Done.

generate_clients.py
landrito: Done.

.circleci/config.yml
landrito: I decided against doing project wide env variable since the showcase version will be changing rapidly until API is fully fleshed out.

landrito:
PTAL

@jbolinger can you take a look at the ruby code?

andreamlin:
circleci setup LGTM

showcase/ruby/showcase_integration_test.rb
```diff
@@ -0,0 +1,140 @@
...
+
+      # Call method
+      response = @client.wait(response_delay, success: success)
+
+      # Verify the response
+      assert_equal(expected_response, response)
+    end
+  end
+
+  describe 'pagination' do
```
jbolinger: Consider breaking this into two tests?

showcase/ruby/showcase_integration_test.rb
```diff
@@ -0,0 +1,140 @@
...
+      # Verify the response
+      assert_raises do
+        @client.echo(error: error)
+      end
+    end
+  end
+
+  describe 'expand' do
+    it 'invokes expand' do
+      content = "The rain in spain stays mainly on the plain!"
```
jbolinger: super nit: Spain?

showcase/ruby/showcase_integration_test.rb
landrito: Done!

showcase/ruby/showcase_integration_test.rb
landrito: Done!

jbolinger:
Thanks @landrito :)

PR #2231, opened 2018-08-15
Add gcpApiConfigPath to client default options

WeiranFang:
Add a new key-value pair 'gcpApiConfigPath' to the client default options. This value will be used by gax-php.
Related PR: https://github.com/googleapis/gax-php/pull/207


src/test/java/com/google/api/codegen/testdata/php/php_library.baseline
```diff
@@ -196,6 +196,7 @@ class LibraryServiceGapicClient
             'serviceAddress' => self::SERVICE_ADDRESS . ':' . self::DEFAULT_SERVICE_PORT,
             'clientConfig' => __DIR__ . '/../resources/library_service_client_config.json',
             'descriptorsConfigPath' => __DIR__ . '/../resources/library_service_descriptor_config.php',
+            'gcpApiConfigPath' => __DIR__ . '/../resources/{@xapiClass.clientConfigName}_grpc_config.json',
```
dwsupplee: I believe this needs to utilize the service name here (e.g. `library_service_grpc_config.json`).  This applies on the other baseline files as well.

WeiranFang: done. May I ask how these test files will be used? Do I also need to provide a library_service_grpc_config.json file? thx.

dwsupplee: I believe one of the steps in the CI process is to generate GAPIC clients and ensure the output matches the output of the baseline. Providing the actual config file shouldn't be necessary. /cc @michaelbausor to confirm.

dwsupplee: @WeiranFang expressed [we may only want to use these configs on a case by case basis](https://github.com/GoogleCloudPlatform/google-cloud-php/pull/1146#issuecomment-412598983), as they don't apply to all cloud APIs. If the generator here has the intel available to determine when to make that switch, I'd be in favor of generating the config as well.

michaelbausor: Right now the generator doesn't have that intel. So not generating the file by default may be the best way to go. That will also avoid the potentially annoying problem of generating a default file that will always overwrite a manually edited file. We just need to make sure that a missing file is not an error, which based on the gax PR is the current behaviour.

WeiranFang: Currently this only works for Spanner (And also we only did tests for Spanner), so I think it's better if we only have the config file for Spanner. Other client APIs will not enable the gcp extension because the files won't exist in their resource folder. We will slowly roll out the feature for all other storage APIs in the future, if by that time we have more client libraries involved, the we can do that way (generates a default config file and use another flag to determine when to switch on/off). Does that make sense?

PR #2233, opened 2018-08-15
Csharp unit tests (WIP)

andreamlin:
~Built on #2228, so that should go in first.~


alexander-fenster:
@chrisdunelm Asking you since this one has C# in the title :) Do you think this PR can/should be brought back to life, or should we just close it?

PR #2239, opened 2018-08-17
sample: use the correct config schema

pongad:


src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -333,13 +333,28 @@ message SampleValueSet {
...
+// `SampleInitAttribute` controls how a sample initializes the request object.
+message SampleInitAttribute {
+  // Required. The part of the request object this attribute is talking about.
+  string path = 1;
+
+  // If a sample, the call to client lib method is itself in a function.
+  // If `function_param` is true, the part of the request object specified by `path`
+  // should be accepted as a parameter to the sample function.
+  //
+  // TODO: Allow configuring the param name?
```
pongad: @vchudnov-g How important is this? I'm not sure how to do this yet, but I can prioritize investigating.

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -333,13 +333,28 @@ message SampleValueSet {
   //   several placeholders.
   repeated string parameters = 4;
 
-  repeated string sample_args = 5;
+  // Attributes governing how to init request object.
+  // Each attribute must have unique, non-overlapping, path.
+  repeated SampleInitAttribute attributes = 5;
```
vchudnov-g: As per our discussion, `SampleValueSet` should have its field `parameters` contain two parts: a `repeated string` of default key/value pairs, and a `repeated SampleInitAttribute` for their metadata. Do you have a reason for wanting to change that?

```
sample_value_set:
   …
  parameters:
    defaults:
      - alice=bob
      - carl=denise
      - …
    attributes:
      parameter: carl
      sample_argument: true
```

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -333,13 +333,28 @@ message SampleValueSet {
...
+message SampleInitAttribute {
+  // Required. The part of the request object this attribute is talking about.
+  string path = 1;
+
+  // If a sample, the call to client lib method is itself in a function.
+  // If `function_param` is true, the part of the request object specified by `path`
+  // should be accepted as a parameter to the sample function.
+  //
+  // TODO: Allow configuring the param name?
+  bool function_param = 2;
```
vchudnov-g: How about "sample_argument"? I worry we could confuse people with library function param vs sample function param

src/test/java/com/google/api/codegen/testsrc/library_gapic.yaml
```diff
@@ -247,8 +247,9 @@ interfaces:
       parameters:
       - shelf.name=Math
       - series_uuid.series_string=xyz3141592654
-      sample_args:
-      - shelf.name
+      attributes:
```
vchudnov-g: FYI, the sample labeled "canonical" is the one where I'm trying to showcase all the samplegen features, as per @beccca 's request

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
```diff
@@ -211,34 +212,45 @@ public void generateSamples(
...
+        if (thisContext == null) {
+          // TODO(pongad): Check that the attributes don't overlap.
+          // Probably do this after we implement the second kind of attributes;
+          // with only one implemented, I'm not sure what's the best way to do this yet.
+          thisContext =
+              createInitCodeContext(
+                  context,
+                  fieldConfigs,
+                  initCodeOutputType,
+                  valueSet.getParametersList(),
```
vchudnov-g: In `createInitCodeContext` below, I would now suggest for clarity s/`sampleCodeInitFields`/`sampleCodeDefaultValues`/ and s/`sampleArgs`/sampleFunctionArguments`/

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
```diff
@@ -211,34 +212,45 @@ public void generateSamples(
...
-                        initCodeOutputType,
-                        valueSet.values().getParametersList(),
-                        valueSet.values().getSampleArgsList()));
-        List<OutputSpec> outputs = valueSet.values().getOnSuccessList();
+      for (ValueSetAndTags setAndTag : matchingValueSets) {
+        // Don't overwrite the initContext in outer scope.
+        InitCodeContext thisContext = initContext;
+        SampleValueSet valueSet = setAndTag.values();
+        if (thisContext == null) {
+          // TODO(pongad): Check that the attributes don't overlap.
```
vchudnov-g: I would explain what you mean by overlap. What we discussed was not having any part of the request object described by more than one `SampleInitAttribute`

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
```diff
@@ -211,34 +212,45 @@ public void generateSamples(
...
+          // Probably do this after we implement the second kind of attributes;
+          // with only one implemented, I'm not sure what's the best way to do this yet.
```
vchudnov-g: I don't think this depends on the types of attributes, but rather whether a part of the request object is covered by more than one: ie specifying a `SampleInitAttribute` for `request.list[0]` and another one for `request.list` should be an error.

OK to leave this as a TODO for now, but we should address it before we release.

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
```diff
@@ -211,34 +212,45 @@ public void generateSamples(
...
+          thisContext =
+              createInitCodeContext(
+                  context,
```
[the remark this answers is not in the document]
pongad: Done

src/main/proto/com/google/api/codegen/config.proto
pongad: Let's loop in customers next week about this. I can see it becoming important.

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -333,13 +333,28 @@ message SampleValueSet {
...
 
   // How to process a successful response from the client library
   // function being exemplified.
   repeated OutputSpec on_success = 6;
 }
 
+// `SampleInitAttribute` controls how a sample initializes the request object.
+message SampleInitAttribute {
+  // Required. The part of the request object this attribute is talking about.
+  string path = 1;
```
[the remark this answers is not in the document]

src/main/proto/com/google/api/codegen/config.proto

src/test/java/com/google/api/codegen/testsrc/library_gapic.yaml
pongad: Ack. Since canonical is defined in the same method, I don't think we need to make any change to this PR right?

src/main/proto/com/google/api/codegen/config.proto
vchudnov-g: Yeah, let's defer this for after this PR, but we'll need a user-friendly solution before releasing. We can discuss which of the two makes more sense when.

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -323,21 +323,41 @@ message SampleValueSet {
...
 
-  // How to process a successful response from the client library
-  // function being exemplified.
-  repeated OutputSpec on_success = 6;
+// `SampleInitAttribute` controls how a sample initializes the request object.
+message SampleInitAttribute {
+  // Required. The part of the request object this attribute is talking about.
+  string parameter = 1;
+
+  // If a sample, the call to client lib method is itself in a function.
```
vchudnov-g: s/`If`/`In`/, yes?

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -323,36 +323,41 @@ message SampleValueSet {
...
   string description = 3;
 
-  // A set of parameter values to be used together in samples. Each
-  // member of `parameters` can be specified in one of two formats:
+  // How to initialize the request object for the RPC.
+  SampleParameters parameters = 4;
+
+  // How to process a successful response from the client library
+  // function being exemplified.
+  repeated OutputSpec on_success = 6;
```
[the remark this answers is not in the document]

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -323,36 +323,41 @@ message SampleValueSet {
...
+}
+
+message SampleParameters {
```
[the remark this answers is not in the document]

src/main/proto/com/google/api/codegen/config.proto
pongad: Yes :)

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
```diff
@@ -198,57 +200,59 @@ public void generateSamples(
...
       }
 
       for (ValueSetAndTags setAndTag : matchingValueSets) {
         // Don't overwrite the initContext in outer scope.
         InitCodeContext thisContext = initContext;
         SampleValueSet valueSet = setAndTag.values();
         if (thisContext == null) {
-          // TODO(pongad): Check that the attributes don't overlap.
+          // TODO(pongad): Check that the attributes don't "overlap".
+          // Eg, specifying attributes for both `list` and `list[0]` should be an error.
```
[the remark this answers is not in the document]
pongad: Done

PR #2240, opened 2018-08-17
sample: comment out arg defaults

pongad:
Since these values are passed by arguments,
we comment out the initializations so that they serve as
documentation but won't clobber the arguments.

### DIFF
--- a/src/main/resources/com/google/api/codegen/java/initcode.snip
+++ b/src/main/resources/com/google/api/codegen/java/initcode.snip
@@ -2,9 +2,10 @@
 
 @snippet initCode(initCodeSpec)
   @if initCodeSpec.argDefaultLines
-    // FIXME(pongad): should be commented to not clobber args
-    {@initCodeLines(initCodeSpec.argDefaultLines)}
-    // FIXME(pongad): args end
+    @join line : util.pretty(initCodeLines(initCodeSpec.argDefaultLines))
+      // {@line}
+    @end
+
 
   @end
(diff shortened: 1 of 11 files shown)


src/main/resources/com/google/api/codegen/java/initcode.snip
```diff
@@ -2,9 +2,10 @@
 
 @snippet initCode(initCodeSpec)
   @if initCodeSpec.argDefaultLines
-    // FIXME(pongad): should be commented to not clobber args
-    {@initCodeLines(initCodeSpec.argDefaultLines)}
-    // FIXME(pongad): args end
+    @join line : util.pretty(initCodeLines(initCodeSpec.argDefaultLines))
+      // {@line}
+    @end
+
```
pongad: I have no idea why these newlines are required, but they are. Otherwise we get really weird results like
```
// this the last line of the commentThis should be the first line but somehow it is concatenated here.
```

vchudnov-g: Did you try `@join ... on BREAK`?

pongad: I did, but I don't think it can possibly matter, if you don't have the `on` clause, it joins on BREAK by default: https://github.com/googleapis/api-compiler/blob/093283bb4bf19f3866bca5e784fe722d277afbc5/src/main/java/com/google/api/tools/framework/snippet/SnippetParser.java#L306

PR #2241, opened 2018-08-18
bump gradle version

pongad:


pongad:
@andreamlin Does changing Gradle version need to change any of the docker images we pull for Circle build?

PR #2244, opened 2018-08-19
Samplegen: java output

hzyi-google:
Augment samplegen output features to support java.


src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -344,15 +382,39 @@ private TypeModel get(String name) {
...
       if (!sample.add(name)) {
         return false;
       }
-      scope.put(name, type);
+      if (type == null) {
+        typeNames.put(name, typeName);
+        return true;
+      }
+      types.put(name, type);
+      typeNames.put(name, typeName);
```
pongad: I'd flip the logic
```java
typeNames.put(name, typeName);
if (type != null) types.put(name, type);
```

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -209,26 +203,68 @@ public void run() {
...
       if (context.getMethodConfig().getPageStreaming() != null) {
-        type =
-            context
-                .getMethodConfig()
-                .getPageStreaming()
-                .getResourcesFieldConfig()
-                .getField()
-                .getType()
-                .makeOptional();
+        FieldConfig resourceFieldConfig =
```
pongad: I think I know where this is going but the flow control is scary (I'm having a hard time matching braces). I think you can flatten this.

Something like
```java
boolean pageStreaming = context.getMethodConfig().getPageStreaming() != null;
boolean pageStreamingResourceField = ....;


if (pageStreamingResourceField) {
  // ...
  return
}
if (pageStreaming) {
  // ...
} else {
  // ...
}
```

src/main/resources/com/google/api/codegen/java/standalone_sample.snip
```diff
@@ -41,42 +41,278 @@
...
+        public void onError(Throwable t) {
+          // Add error-handling
+        }
+
+        @@Override
+        public void onCompleted() {
+          // Do something when complete.
+        }
+      };
+@end
```
pongad: new line at end of file.

hzyi-google:
PTAL

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -207,8 +200,48 @@ public void run() {
     TypeModel type;
     if (baseIdentifier.equals(RESPONSE_PLACEHOLDER)) {
       view.variable(context.getNamer().getSampleResponseVarName(context));
+      boolean pageStreaming = context.getMethodConfig().getPageStreaming() != null;
+      boolean pageStreamingAndUseResourceName =
+          pageStreaming
```
pongad: pageStreaming && context....?

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -207,8 +200,48 @@ public void run() {
...
-      if (context.getMethodConfig().getPageStreaming() != null) {
+        if (newVar != null) {
+          Preconditions.checkArgument(
+              !intoScalar,
+              "%s:%s: cannot parse a resource name into scalar",
+              context.getMethodModel().getSimpleName(),
+              valueSet.getId());
+
+          if (!localVars.put(newVar, null, typeName)) {
+            throw new IllegalStateException(
```
pongad: This should be `IllegalArgumentException` I think, because user gave us bad config. You can also use `checkArgument` here.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -220,15 +253,25 @@ public void run() {
       } else {
         type = context.getMethodModel().getOutputType();
       }
-    } else {
+    } else { // Defining and assigning the value of a local variable. Reference the type from the
```
pongad: nit: Let's start the comment on a new line.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
hzyi-google: lol. I had a brain fart and was doing pageStreaming & context... and always got a npe. Now I understand why that kept happening :/

src/main/resources/com/google/api/codegen/java/standalone_sample.snip
```diff
@@ -41,42 +41,278 @@
 
 # adapted from main.snip:apiMethods(xapiClass) and its dependencies
 # For real use, we'll want to switch on the calling form id.
-@private generateSample(apiMethod, callingForm, initCode)
-  @switch callingForm
+@private generateSample(apiMethod, sample)
```
vchudnov-g: Update the comment right above this, since we're already switching on the calling form.

src/main/resources/com/google/api/codegen/java/standalone_sample.snip
```diff
@@ -41,42 +41,278 @@
...
-    {@asyncOperationMethodSampleCode(apiMethod, initCode)}
+    {@asyncOperationMethodSampleCode(apiMethod, sample)}
+    {@processResponse(sample)}
   @default
-    $unhandledCallingForm: {@callingForm} in sample "{@apiMethod.getClass.getSimpleName}"$
+    $unhandledCallingForm: {@sample.callingForm} in sample "{@apiMethod.getClass.getSimpleName}"$
   @end
 @end
 
+# The snippets are copied and modified from method_sample.snip since standalone
```
vchudnov-g: "... are modified versions of those in method_sample.snip since standalone samples can perform more sophisticated response handling."

src/main/resources/com/google/api/codegen/java/standalone_sample.snip
```diff
@@ -41,42 +41,278 @@
...
+  @case "ServerStreaming"
+    {@serverStreamingCallableSampleCode(apiMethod, sample)}
+  @case "ClientStreaming"
+    {@clientStreamingCallableSampleCode(apiMethod, sample)}
+  @default
+    $unhandledCase: {@apiMethod.grpcStreamingType}$
+  @end
+@end
+
+@snippet bidiStreamingCallableSampleCode(apiMethod, sample)
```
vchudnov-g: If these are only used in this file, lets make all of these `@private`

In `method_sample.snip`, let's also make the corresponding functions from which you've now branched  `@private` to clarify our intended use cases and  catch errors.

src/main/resources/com/google/api/codegen/java/standalone_sample.snip
```diff
@@ -41,42 +41,278 @@
...
+  {@initCode(sample.initCode)}
+  for ({@apiMethod.listMethod.resourceTypeName} element : {@sampleSyncMethodCall(apiMethod)}.{@apiMethod.listMethod.iterateMethodName}()) {
+    {@processResponse(sample)}
+  }
+@end
+
+@snippet pagedCallableMethodSampleCode(apiMethod, sample)
+  {@initCode(sample.initCode)}
+  ApiFuture<{@apiMethod.responseTypeName}> future = {@sampleFutureMethodCall(apiMethod)};
+  // Do something
```
vchudnov-g: Let's remove all these "Do something"s in favor of the response handling.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -220,15 +253,25 @@ public void run() {
...
-      type =
-          Preconditions.checkNotNull(
-              localVars.get(baseIdentifier),
-              "%s:%s: variable not defined: %s",
-              context.getMethodModel().getSimpleName(),
-              valueSet.getId(),
-              baseIdentifier);
+      type = localVars.getType(baseIdentifier);
+      if (type == null) {
+        String typeName =
```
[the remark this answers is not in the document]
hzyi-google: Yes. Simply checking that the `typeName` exists in `localVars` would be sufficient.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -220,15 +253,25 @@ public void run() {
...
+                "%s:%s: variable not defined: %s",
+                context.getMethodModel().getSimpleName(),
+                valueSet.getId(),
+                baseIdentifier);
+        Preconditions.checkArgument(
+            config.scan() == Scanner.EOF,
+            "%s:%s: accessing fields of resource names is not supported",
+            context.getMethodModel().getSimpleName(),
+            valueSet.getId());
+        return view.accessors(ImmutableList.<String>of()).build();
```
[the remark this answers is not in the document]
hzyi-google: This means there will only be cases like `ResourceNameType resourceName = oldResourceName`, but not cases like `String a = resourceName.getFieldA()`, where `getFieldA` would live in the accessor list.

src/main/resources/com/google/api/codegen/java/standalone_sample.snip
hzyi-google: I think these `//Do something`s are to show the `future.get()` does not have to be called immediately after a `future` is created?

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -207,8 +200,47 @@ public void run() {
...
+      if (pageStreamingAndUseResourceName) {
+        String typeName =
+            context
+                .getNamer()
+                .getAndSaveElementResourceTypeName(
+                    context.getTypeTable(),
+                    context.getMethodConfig().getPageStreaming().getResourcesFieldConfig());
+
+        Preconditions.checkArgument(
+            config.scan() == Scanner.EOF,
```
vchudnov-g: We will want to support accessing members of a resource object, either directly with `.` or (most likely) indirectly via a `%` that will cause its rhs to resolve to the right accessor.

So include a TODO here for clarity, reminding us that the intent  is to allow this, but we have not yet gotten to this.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -207,8 +200,47 @@ public void run() {
...
+      // Compute the resource name format of output type and save that to local variables
+      if (pageStreamingAndUseResourceName) {
+        String typeName =
+            context
+                .getNamer()
+                .getAndSaveElementResourceTypeName(
+                    context.getTypeTable(),
+                    context.getMethodConfig().getPageStreaming().getResourcesFieldConfig());
+
+        Preconditions.checkArgument(
```
vchudnov-g: One of the questions I had is why we're doing this only at the top level and not further down the tree (for example, somebody could set the assignment `foo = RESPONSE_PLACEHOLDER`. As we discussed offline, in that case `foo` will be in the local scope table and the checks in the loop below will catch it.

So, one thing to investigate is whether we can simplify the logic here by removing these special checks and instead putting the node corresponding to `BASE_IDENTIFIER` directly in the type table. This may also apply to non-page streaming cases, I think. (Not too sure about this, rather confusing.)

@pongad for comments

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -207,8 +200,47 @@ public void run() {
     TypeModel type;
     if (baseIdentifier.equals(RESPONSE_PLACEHOLDER)) {
       view.variable(context.getNamer().getSampleResponseVarName(context));
+      boolean pageStreaming = context.getMethodConfig().getPageStreaming() != null;
+      boolean pageStreamingAndUseResourceName =
+          pageStreaming
+              && context
+                  .getFeatureConfig()
+                  .useResourceNameFormatOption(
+                      context.getMethodConfig().getPageStreaming().getResourcesFieldConfig());
```
vchudnov-g: Do resource objects only apply to the top level of a response (paged response in this case), or could they also be defined for submessages within the response? In `library_gapic.yaml`, for example, the end section contains `- message_name` entries. Can those only refer to top level responses (either pages or unpaged), or could they apply to submessages within the responses?

@pongad @garrettjonesgoogle

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -207,8 +200,47 @@ public void run() {
...
+        if (newVar != null) {
+          Preconditions.checkArgument(
+              !scalarTypeForCollection,
+              "%s:%s: resournce names can never be a repeated field",
+              context.getMethodModel().getSimpleName(),
+              valueSet.getId());
+
+          Preconditions.checkArgument(
+              localVars.put(newVar, null, typeName),
+              "%s:%s: duplicated variable declaration not allowed: %s",
```
vchudnov-g: "duplicate"

below as well

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
garrettjonesgoogle: I don't think we support submessages for within responses, but if we do, I doubt we have used it in any current APIs.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
pongad: I think resource name here is actually a special case for two reasons.

1. It's the only time we need to deal with a "synthetic type". Resource names don't have a protobuf type backing it up.

2. IIUC, resource names can only appear at the top level. GAPIC client can return resource names, or lists of resource names, but never Protobuf messages containing resource names or lists thereof. This is because resource names are known only to GAPICs, not Protobuf. When Protobuf types are generated, `protoc` doesn't know what resource names are, so response messages have no way of referring to them.

hzyi-google:
PTAL @vchudnov-g

Issue #2097, opened 2018-06-21
[Python] some py files from generator contain utf-8 chars but are missing the source encoding header


crwilcox:
*.py files that contain utf-8 characters should specify the encoding with the following header
```
# -*- coding: utf-8 -*-
```

crwilcox:
Example: https://github.com/dialogflow/dialogflow-python-client-v2-private/pull/1/commits/c370c731aa003c61eea619d4d631cbf8de108ce0

crwilcox:
I should also note I believe this can be safely added to all generated files.

timburks:
Thanks @crwilcox -- @lukesneeringer could you take care of this or find an owner?

lukesneeringer:
We can add these to our files easily, but it will be trickier to add it to the pb2 files (which is I think where the problems lie). It can be done though, and we already edit those files to alter imports.

crwilcox:
Those are the files with the issue typically, yes.

tseaver:
For those playing along at home:  note that current `synth.py` files in `google-cloud-python` still need to have the following replacement, or else the unit tests break on Python 2.7:

```python
s.replace(
    '**/proto/*_pb2.py',
    r"(^.*$\n)*",
    r"# -*- coding: utf-8 -*-\n\g<0>")
```

PR #2246, opened 2018-08-20
Add utf-8 header to generated *.py files

andreamlin:
Resolves https://github.com/googleapis/gapic-generator/issues/2097


andreamlin:
PTAL

src/main/resources/com/google/api/codegen/copyright-google.txt
```diff
@@ -1 +1 @@
-Copyright 2018 Google LLC
+Copyright 2018 Google LLC
```
landrito: I'm confused what was changed here?

andreamlin: There used to be a newline at the end of the file, now it's removed.

PR #2248, opened 2018-08-22
[Proto Annotations] Enable passing in --package flag

andreamlin:
The `--package` flag is necessary when only a descriptor set is given (no service yaml, no gapic config, no package2.yaml). It specifies the proto package e.g. `google.pubsub.v1` for the source protos that we actually want to make a library for, as opposed to imported packages like `google.api.http`.

### DIFF
--- a/src/main/java/com/google/api/codegen/GeneratorMain.java
+++ b/src/main/java/com/google/api/codegen/GeneratorMain.java
@@ -106,7 +106,7 @@ public class GeneratorMain {
           .argName("GAPIC-YAML")
           .required(true)
           .build();
-  private static final Option GAPIC_YAML_NON_REQUIRED_OPTION =
+  private static final Option GAPIC_YAML_NONREQUIRED_OPTION =
       Option.builder()
           .longOpt("gapic_yaml")
           .desc(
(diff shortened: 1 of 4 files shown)


andreamlin:
PTAL

src/main/java/com/google/api/codegen/GeneratorMain.java
```diff
@@ -200,11 +220,12 @@ public static void gapicGeneratorMain(ArtifactType artifactType, String[] args)
     Options options = new Options();
     options.addOption("h", "help", false, "show usage");
     options.addOption(DESCRIPTOR_SET_OPTION);
-    options.addOption(SERVICE_YAML_OPTION);
+    options.addOption(SERVICE_YAML_NONREQUIRED_OPTION);
     // TODO make required after artman passes this in
     options.addOption(LANGUAGE_NONREQUIRED_OPTION);
-    options.addOption(GAPIC_YAML_OPTION);
+    options.addOption(GAPIC_YAML_NON_REQUIRED_OPTION);
```
garrettjonesgoogle: Let's be consistent - either `NONREQUIRED` or `NON_REQUIRED`.

src/main/java/com/google/api/codegen/GeneratorMain.java
```diff
@@ -408,4 +447,15 @@ private static void checkFile(String filePath) {
       throw new IllegalArgumentException("File not found: " + filePath);
     }
   }
+
+  // Throws an exception if neither options were given.
```
garrettjonesgoogle: -> "if neither option was given"

andreamlin:
openjdk8 is broken because pubsub gapic fails to generate due to https://github.com/googleapis/gapic-generator/pull/2244 ([failure log](https://1687-56533838-gh.circle-artifacts.com/0/tmp/workspace/reports/generations/smoketest.log). Word on the street is that a quick fix is coming.

PR #2250, opened 2018-08-23
Test Logging PHP client

andreamlin:
Add Logging test back (common protos for PHP were refreshed so this should work)
Add system tests for PHP clients.


andreamlin:
PTAL

andreamlin:
Blocked on https://github.com/googleapis/gapic-generator/issues/2252

andreamlin:
openjdk8 fails because of https://github.com/googleapis/gapic-generator/issues/2256

PR #2251, opened 2018-08-23
Nodejs generated unit tests

andreamlin:


andreamlin:
This is built on https://github.com/googleapis/gapic-generator/pull/2283, which is a fix for node. Let that get merged in first.

andreamlin:
PTAL

.circleci/config.yml
```diff
@@ -235,6 +235,16 @@ anchor_run_php: &anchor_run_php
         name: Test PHP Logging client
         <<: *anchor_test_php
 
+# Test a NodeJS client if generation succeeded.
+anchor_test_nodejs_client: &anchor_test_nodejs_client
+  command: |
+    cd ${NODEJS_GEN_DIR}/${TEST_API}
+    npm install
+    # Copy over IAM protos.
+    cp -r node_modules/google-proto-files/google/iam/ protos/google/
```
alexander-fenster: why do we need to copy those? In other words, why we don't need to copy those for artman-generated client libraries? Does artman do some extra steps that gapic-generator does not?

andreamlin: `${NODEJS_GEN_DIR}/${TEST_API}` is an artman generated library. I don't know why artman doesn't place the necessary protos in the protos/google directory of the generated lib.

andreamlin: Looking at the existing [nodejs-pubsub/proto](https://github.com/googleapis/nodejs-pubsub/tree/master/protos/google) dir, those protos haven't been updated for a long time, so they've just always been there, so node automation doesn't have to deal with copying over the protos

alexander-fenster: Well, it makes sense. OK let's leave this copy here for now, even though I feel it's not exactly correct :)

PR #2254, opened 2018-08-23
Use api-compiler-0.0.8

andreamlin:
api-compiler is needed to support the [api-common-protos](https://github.com/googleapis/api-common-protos/tree/input-contract) (from branch input-contract) that are needed to be installed in gapic-generator in order to parse annotated protos, using [annotations.proto](https://github.com/googleapis/api-common-protos/blob/master/google/api/annotations.proto)

Edit: all other changes in src/ are due to surface changes from the new api-compiler dependency.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -59,7 +59,7 @@ ext {
   libraries = [
     // General
     apiCommon: 'com.google.api:api-common:1.1.0',
-    apiCompiler: 'com.google.api:api-compiler:0.0.7',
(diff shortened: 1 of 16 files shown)


andreamlin:
PTAL

garrettjonesgoogle:
cc @jskeet wasn't this relevant to you for some reason?

jskeet:
Yup, could make it much quicker to run on Windows :) Will try...

PR #2260, opened 2018-08-23
Add in api-common-protos from api-common-protos/input-branch

andreamlin:
Use com.google.api.grpc:{grpc,proto}-google-common-protos:1.13.0-pre1, which were created by PR https://github.com/googleapis/api-client-staging/pull/661

We need the newest forked common protos package for the proto annotations defined in [annotations.proto](https://github.com/googleapis/api-common-protos/blob/input-contract/google/api/annotations.proto) in order to parse the API proto's annotations.

The clients will have some extra imports (see baselines) caused by the addition of new files from the new libraries, but it doesn't seem to break the generated clients, even tho the generated clients use whatever version of common protos is in [dependencies.yaml](https://github.com/googleapis/gapic-generator/blob/master/src/main/resources/com/google/api/codegen/packaging/dependencies.yaml).


andreamlin:
the approval has been given!

PR #2261, opened 2018-08-23
add unit tests for OutputTransformer

hzyi-google:


src/test/java/com/google/api/codegen/transformer/OutputTransformerTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+  @Before
+  public void setUp() {
+    outer = new ScopeTable();
+    inner = new ScopeTable(outer);
+  }
+
+  @Test
+  public void testScopeTablePut() {
+    assertThat(outer.put(stringVariable, stringTypeModel, stringTypeName)).isTrue();
```
pongad: For all the tests, I think we should write out the value in-line. Declaring all the constants up top means we have to jump back and forth between test and definition.

hzyi-google: Ack

src/test/java/com/google/api/codegen/transformer/OutputTransformerTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+public class OutputTransformerTest {
+
+  private ScopeTable outer;
+  private ScopeTable inner;
+
+  @Before
+  public void setUp() {
+    outer = new ScopeTable();
+    inner = new ScopeTable(outer);
```
pongad: I just realized, I think you have this swapped right? Let's have
```java
parent = new ScopeTable();
child = new ScopeTable(parent);
```
since that's the name used in the scope table.

PR #2262, opened 2018-08-24
Proto-annotate library.protos

andreamlin:
Annotate the library.proto files for testing.

Soon, we should have two versions of this file, one with proto annotations, and one without, so we can still test using the GAPIC config on a proto file both with and without annotations.

### DIFF
--- a/src/test/java/com/google/api/codegen/packagegen/testdata/library.proto
+++ b/src/test/java/com/google/api/codegen/packagegen/testdata/library.proto
@@ -33,15 +33,35 @@ option java_package = "com.google.example.library.v1";
 //
 // - Each Shelf has a collection of [Book][google.example.library.v1.Book]
 //   resources, named `shelves/*/books/*`
+
+option (google.api.metadata) = {
+  product_name: "Library Service"
+  product_uri: "https://cloud.google.com/library/"
+  package_name: "Library"
+  package_namespace: ["Google", "Example"]
+};
+
 service LibraryService {
+  option (google.api.default_host) = "library-example.googleapis.com";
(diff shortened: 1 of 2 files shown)


andreamlin:
PTAL

src/test/java/com/google/api/codegen/packagegen/testdata/library.proto
```diff
@@ -64,40 +87,61 @@ service LibraryService {
...
+      fields: ["name"]
+    };
   }
 
   // Updates a book. Returns INVALID_ARGUMENT if the name of the book
   // is non-empty and does equal the previous name.
   rpc UpdateBook(UpdateBookRequest) returns (Book) {
     option (google.api.http) = { put: "/v1/{name=shelves/*/books/*}" body: "book" };
+    option (google.api.method_signature) = {
+      fields: ["name", "book", "update_mask"]
```
lukesneeringer: This seems wrong. I would have expected either `["book"]` or `["book", "update_mask"]` (probably the latter).

src/test/java/com/google/api/codegen/packagegen/testdata/library.proto
```diff
@@ -230,24 +297,34 @@ message ListBooksResponse {
 // Request message for LibraryService.UpdateBook.
 message UpdateBookRequest {
   // The name of the book to update.
-  string name = 1;
+  string name = 1 [
```
lukesneeringer: Oh, I see, it is because this is wrong (which is unrelated to this PR).

Meh, okay.

PR #2267, opened 2018-08-24
use `const` instead of `var` for nodejs

JustinBeckwith:


andreamlin:
(just driving by) The *.baseline files also have to be updated. An easy way to do that is to run the unit tests locally, and failing baselines will be spit out in `/tmp/com.google.api.codegen_testdata/`. Copy over the new baselines from there into their appropriate location in `/usr/local/google/home/andrealin/Documents/git/veneer/toolkit/src/test/java/com/google/api/codegen/testdata/...`

src/main/resources/com/google/api/codegen/nodejs/package.json.snip
```diff
@@ -41,7 +41,7 @@
     },
     "license": "{@metadata.licenseName}",
     "engines": {
-      "node": ">=4.0.0"
+      "node": ">=6.0.0"
```
lukesneeringer: Are you going 2.0 right now™?

JustinBeckwith: We've already made this change across all of the `nodejs-` repositories, figured we may as well here. Happy to back it out if that's a problem :)

alexander-fenster: The generated `package.json` is only used for the new libraries, it won't update the existing ones, so it's safe to put `>=6.0.0` here.

lukesneeringer: Okay, that makes sense to me.

src/main/resources/com/google/api/codegen/nodejs/main.snip
```diff
@@ -71,13 +71,13 @@
...
       opts.scopes = this.constructor.scopes;
-      var gaxGrpc = new gax.GrpcClient(opts);
+      const gaxGrpc = new gax.GrpcClient(opts);
 
       // Save the auth object to the client, for use by other methods.
       this.auth = gaxGrpc.auth;
 
       // Determine the client header string.
-      var clientHeader = [
+      const clientHeader = [
```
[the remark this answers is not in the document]
alexander-fenster: The former: In JavaScript, `const` is similar to `T * const` in C++ (as opposed to `const T *`). The variable won't be reassigned but if it's an object, its contents can be changed.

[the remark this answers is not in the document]
alexander-fenster: I don't think we had any discussions about `const` usage (not with me at least), but I see no problems using `const` as defined in the language. [link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const)

PR #2269, opened 2018-08-25
Normalize line endings when splitting for docs

### LINKED ISSUE #2268: Comment line splitting fails on Windows
Now that we're on a more recent version of api-compiler, we can run GAPIC on Windows.

It *nearly* works - but whatever splits comment lines is broken. We end up with comments like this:

```csharp
        /// <summary>
        /// Deletes all the log entries in a log.
The log reappears if it receives new entries.
Log entries written shortly before the delete operation might not be
deleted.
        /// <summary>
```

instead of

```csharp
        /// <summary>
        /// Deletes all the log entries in a log.
        /// The log reappears if it receives new entries.
        /// Log entries written shortly before the delete operation might not be
        /// deleted.
        /// <summary>
```

I've labeled this as a C# issue as it affects us most, but it would probably affect other users on Windows too. I'll have a quick look to see if I can find the issue myself, but I'm not sure how much luck I'll have...

jskeet:
The tests only cover one overload, but it basically works the same
in both cases.

Fixes #2268.

### DIFF
--- a/src/test/java/com/google/api/codegen/util/CommonRenderingUtilTest.java
+++ b/src/test/java/com/google/api/codegen/util/CommonRenderingUtilTest.java
@@ -36,4 +36,15 @@ public class CommonRenderingUtilTest {
     assertThat(CommonRenderingUtil.stripQuotes("'a'bc'")).isEqualTo("'a'bc'");
     assertThat(CommonRenderingUtil.stripQuotes("\"a\"bc\"")).isEqualTo("\"a\"bc\"");
   }
+
+  @Test
+  public void testGetDocLines() {
+    // Check that we don't care which form of line break is used.
+    assertThat(CommonRenderingUtil.getDocLines("a\nb\nc")).containsExactly("a", "b", "c").inOrder();
+    assertThat(CommonRenderingUtil.getDocLines("a\rb\rc")).containsExactly("a", "b", "c").inOrder();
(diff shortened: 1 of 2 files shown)


src/test/java/com/google/api/codegen/util/CommonRenderingUtilTest.java
```diff
@@ -36,4 +36,15 @@ public void testStripQuotes() {
     assertThat(CommonRenderingUtil.stripQuotes("'a'bc'")).isEqualTo("'a'bc'");
     assertThat(CommonRenderingUtil.stripQuotes("\"a\"bc\"")).isEqualTo("\"a\"bc\"");
   }
+
+  @Test
+  public void testGetDocLines() {
+    // Check that we don't care which form of line break is used.
+    assertThat(CommonRenderingUtil.getDocLines("a\nb\nc")).containsExactly("a", "b", "c").inOrder();
+    assertThat(CommonRenderingUtil.getDocLines("a\rb\rc")).containsExactly("a", "b", "c").inOrder();
```
pongad: Do we actually want to split on `\r`? I thought that the "newline sequence" was only `\r\n` on windows.

jskeet: It is, but my understanding was that Macs used \r. It would be a shame to get it working for Windows but not Mac. (For example, BufferedReader breaks on all three options.)

PR #2271, opened 2018-08-27
Create separate directories for different versions of library.proto

andreamlin:
There's 
- annotated library.proto that has been moved to testsrc/libraryproto/annotationsonly
- un-annotated library.proto that has been moved to testsrc/libraryproto/configonly, which also contains the library service yaml, gapic yaml, and package yaml
- all other *.proto and *.yaml files for test services, that have been moved to testsrc/common

This makes it easy to write separate tests for GapicTestBase2.java that simply use different inputs, GapicCodeGeneratorAnnotationsTest that uses the annotated proto, and GapicCodeGeneratorTest that uses config files without an annotated proto.

Update the test resource paths in all other paths to reflect the test resources directories refactoring.
The contents of *.yaml and *.proto files hasn't changed (although we readded the unannotated library.proto)


andreamlin:
PTAL

PR #2272, opened 2018-08-27
samplegen: convert binary type to str

hzyi-google:


src/test/java/com/google/api/codegen/testdata/py/python_library.baseline
```diff
@@ -4345,6 +4345,9 @@ def sample_publish_series(name):
 
   # name = 'Math'
 
+  if isinstance(name, six.binary_type):
```
vchudnov-g: We need to `import six` as well (with the other import that shows up in the sample)

hzyi-google:
PTAL @vchudnov-g

src/main/resources/com/google/api/codegen/py/standalone_sample.snip
```diff
@@ -35,6 +38,14 @@
       @end
 
 
+      # FIXME(hzyi): handle ListInitType, MapInitType and StructInitType correctly
```
vchudnov-g: Instead of having this code that duplicates the one below, why don't you make `shouldImportSix` return all these code lines, and then you include them here? Above, you can just check whether the lines are non-empty.

I'm thinking something like this, in pseudo-code:
```
@let conversionLines = convertText(sample.initCode.argDefaultParams)

@if @conversionLines.empty()
  import six
@

…

def ...
   {@conversionLines}

```

WDYT?

hzyi-google: Yes this should work. I'll make the change.

src/main/resources/com/google/api/codegen/py/standalone_sample.snip
```diff
@@ -23,6 +23,9 @@
   @# [START {@sample.regionTag}]
 
   {@importList(sample.initCode.importSection.appImports)}
+  @if convertText(sample)
```
vchudnov-g: Can you assign this to a template variable around line 8, so you're not calling the function twice?
I think I saw in some of the templates that you can assign more than one variable in  a `@let` statement by separating them with whitespace  / slash-newline

src/main/resources/com/google/api/codegen/py/standalone_sample.snip
```diff
@@ -76,6 +81,28 @@
   @end
 @end
 
+@private shouldImportSix(defaultParams)
```
vchudnov-g: This is no longer used, right? Remove,.

src/main/resources/com/google/api/codegen/py/standalone_sample.snip
```diff
@@ -23,51 +23,59 @@
   @# [START {@sample.regionTag}]
 
   {@importList(sample.initCode.importSection.appImports)}
+  @let convertTextCode = convertText(sample)
```
hzyi-google: @vchudnov-g does this look good to you? I didn't define the `convertTextCode` at line 8 because that would require adding one more parameter to `standaloneSample`, which would be slightly inconsistent with other languages. WDYT?

vchudnov-g: Right, good point, I skipped the fact that this is in a different function....

This looks good!

hzyi-google:
For the record: flake8 lint check passed.

PR #2273, opened 2018-08-27
Add a main function to python samples

vchudnov-g:
This marshals `argv` to the sample's function, without any changes.

TODO: Make sure the elements of `argv` are converted to the proper types for the sample's function.

### DIFF
--- a/src/main/resources/com/google/api/codegen/py/standalone_sample.snip
+++ b/src/main/resources/com/google/api/codegen/py/standalone_sample.snip
@@ -68,11 +68,11 @@
 
   @# [END {@sample.regionTag}]
 
-  from sys import argv
+  import sys
   
   def main():
     @# FIXME: Convert argv from strings to the correct types.
-    sample_{@apiMethod.name}(*argv[1:])
+    sample_{@apiMethod.name}(*sys.argv[1:])
 
   if __name__ == '__main__':
     main()
--- a/src/main/resources/com/google/api/codegen/py/standalone_sample.snip
+++ b/src/main/resources/com/google/api/codegen/py/standalone_sample.snip
@@ -67,7 +67,14 @@
     @# [END {@sample.regionTag}_core]
 
   @# [END {@sample.regionTag}]
-  # FIXME(pongad): add a main method
+
+  def main():
+    import sys
+    @# FIXME: Convert argv from strings to the correct types.
+    sample_{@apiMethod.name}(*sys.argv[1:])
+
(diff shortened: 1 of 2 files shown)


src/main/resources/com/google/api/codegen/py/standalone_sample.snip
```diff
@@ -67,7 +67,14 @@
     @# [END {@sample.regionTag}_core]
 
   @# [END {@sample.regionTag}]
-  # FIXME(pongad): add a main method
+
+  def main():
+    import sys
```
hzyi-google: `import sys` usually happens at the beginning of the file and outside main function. Any specific reason it was put here?

vchudnov-g: Yes, so that we don't pollute the samples on the website. What we show in the website is between the begin and end tags, and it includes the imports we need inside the sample function but explicitly excludes `main`, which is the only place we need `sys`.  Thus we have the main import separately.

I'm now uploading a commit that keeps the import separate but moves it to just above (no longer inside!) `main`. I like it there because it's clear it's only needed for `main`. Another option is to move it to the top of the file but outside the region tags, assuming that the formatter respects that ordering.

WDYT? Opinions, @lukesneeringer ?

src/main/resources/com/google/api/codegen/py/standalone_sample.snip
```diff
@@ -67,7 +67,15 @@
     @# [END {@sample.regionTag}_core]
 
   @# [END {@sample.regionTag}]
-  # FIXME(pongad): add a main method
+
+  from sys import argv
```
lukesneeringer: This is uncommon. The generally-accepted style is `import sys` and then you use `sys.argv` styled as such.

src/main/resources/com/google/api/codegen/py/standalone_sample.snip
lukesneeringer: I may not quite understand. Is the entire snippet the sample?

If you are putting `import sys` within the function because of pure printing reasons (e.g. the user will see `import sys` at what _looks like_ the module level), then I think that is reasonable.

I also think it is okay to `import sys` at the top of the file and not have the import in the snippet used on the sample website. If someone sees `sys.argv` in the code they are cutting and pasting, they know what to do. (On the other hand, if you are importing _other things_ in the sample, that is less good.)

vchudnov-g:
ptal

PR #2276, opened 2018-08-28
Update Python standalone sample template with lint fixes

beccca:
##### Update Python standalone sample template with lint fixes

1. Don't start block comments with multiple `#` (`#### No` | `# Yes`)
 2. Lines < 80 characters

Related: https://github.com/GoogleCloudPlatform/python-docs-samples/pull/1660 sample PR failing Lint ([result output](https://travis-ci.org/GoogleCloudPlatform/python-docs-samples/builds/421674903?utm_source=github_status&utm_medium=notification))

=> @vchudnov-g

### DIFF
--- a/src/main/resources/com/google/api/codegen/py/standalone_sample.snip
+++ b/src/main/resources/com/google/api/codegen/py/standalone_sample.snip
@@ -6,7 +6,7 @@
   {@""}
   @let apiMethod = sampleFile.libraryMethod
     @let sample = apiMethod.samples.get(0)
-      @#### DO NOT EDIT! This is a generated sample ("{@sample.callingForm}",  "{@sample.valueSet.id}")
+      @# DO NOT EDIT! This is a generated sample ("{@sample.callingForm}",  "{@sample.valueSet.id}")
 
       @# To run with the published packages, execute the following before running this code:
       @#   pip install {@sampleFile.gapicPackageName}
--- a/src/main/resources/com/google/api/codegen/py/standalone_sample.snip
+++ b/src/main/resources/com/google/api/codegen/py/standalone_sample.snip
@@ -8,7 +8,7 @@
     @let sample = apiMethod.samples.get(0)
       @# DO NOT EDIT! This is a generated sample ("{@sample.callingForm}",  "{@sample.valueSet.id}")
 
-      @# To run with the published packages, run the following before running code:
+      @# To install the latest published package dependency, execute the following:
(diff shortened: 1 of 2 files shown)


src/main/resources/com/google/api/codegen/py/standalone_sample.snip
```diff
@@ -6,9 +6,9 @@
   {@""}
   @let apiMethod = sampleFile.libraryMethod
     @let sample = apiMethod.samples.get(0)
-      @#### DO NOT EDIT! This is a generated sample ("{@sample.callingForm}",  "{@sample.valueSet.id}")
+      @# DO NOT EDIT! This is a generated sample ("{@sample.callingForm}",  "{@sample.valueSet.id}")
 
-      @# To run with the published packages, execute the following before running this code:
+      @# To run with the published packages, run the following before running code:
```
vchudnov-g: * We have "run" three times in this one sentence. Is that what you want?
* "before running code": it sounds better to say "before running the code" or "...the code below", don't you think?

beccca: I'm honestly down with whatever! As this language is not displayed to users, I could take or leave this.

I'll tweak this and see what you think!

beccca: "before running code": it sounds better to say "before running the code"

^--- I removed the article altogether for horizontal character space ('the' doesn't fit)

beccca: Thoughts?

(a)
```py
# To run this sample using the latest published package, install package via:
#   pip install google-cloud-language
```
(b)
```py
# To run this sample using published packages, execute the following:
#   pip install google-cloud-language
```
(c)
```py
# To install the latest published package dependency, execute the following:
#   pip install google-cloud-language
```

beccca: (d)
```py
```
😏

vchudnov-g: My preference would be a variant of (a) with s/`install package`/`install the package`/ (I'd rather split this line into two than have choppy grammar)

The reason I'm not crazy about (c) is that it leaves the user to make the leap from "I want to run this sample" to "I need to figure out how to install this package" to "oh, this is how". The other alternatives are explicit: if you want to run the sample, do this first.

(And to clarify, this comment will not show up on the website, but it will be visible to users browsing the repo.)

PR #2287, opened 2018-08-31
verify that SampleInitAttributes don't overlap

pongad:


src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -258,6 +264,25 @@ private InitCodeView buildInitCodeViewRequestObject(
     return paths.stream().map(path -> root.subTree(path)).collect(ImmutableList.toImmutableList());
   }
 
+  @VisibleForTesting
+  static void assertNoOverlap(InitCodeNode root, List<String> paths) {
```
vchudnov-g: Add a function comment explaining what this does

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -258,6 +264,25 @@ private InitCodeView buildInitCodeViewRequestObject(
...
+    HashMap<InitCodeNode, String> nodes = new HashMap<>();
```
vchudnov-g: Could you made these names mo descriptive? `nodes` `subNodes` and `node` (below) don't really described how they're being used in this algorithm. Which one is the already checked list, which one is the one currently being checked, etc.

vchudnov-g: Explain the map key/values: `nodes[node]=path`

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -234,6 +237,8 @@ private ClientTestAssertView createAssertView(
 
   private InitCodeView buildInitCodeViewFlattened(
       MethodContext context, InitCodeContext initCodeContext, InitCodeNode root) {
+    assertNoOverlap(root, initCodeContext.sampleArgStrings());
```
vchudnov-g: `sampleArgStrings` only covers the inner functiona arguments, correct? We should consider whether to also check for overlaps with the non-arg init nodes. For example, if we have a description for `images`, should we allow a description for `images[0]`. I had been thinking "no", but now I'm less convinced. Maybe we can let it be. 

I don't consider this blocking for this PR.

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
vchudnov-g: Could you make the function signature more descriptive?
eg `assertNoOverlap(newNode, previousPaths)`
or `assertArgumentDistinct(...)`

pongad: Done

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
pongad: Done

pongad: I added some comments to help. The algorithm is essentially depth-first search. I looked to textbooks for better names but textbooks use `v` and `S` :(

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
pongad: This is correct. It seems like the most natural place to put this at least until we start implementing arg description.

I also lean towards "no" but I'm not married to the idea.

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
pongad: I don't think `newNode` and `prevPaths` are good names, since the function acts on all paths at once; there's no before or after here.

pongad:
@vchudnov-g PTAL

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
vchudnov-g: OK. I guess I was thinking that since you call it every time you add a new param to the view.  (Is that right?)

pongad: I don't think so? This is called only once with all the params.

PR #2291, opened 2018-09-06
Append to CircleCI log files

andreamlin:
Currently the artman smoketest log is overwritten for each call to generate an artman_[api].yaml file for a given language, e.g. build [2053](https://circleci.com/gh/googleapis/gapic-generator/2053#artifacts/containers/0) only has ruby+showcase in it's [smoketest.log](https://2053-56533838-gh.circle-artifacts.com/0/tmp/workspace/reports/generations/smoketest.log).

Change this to collate all logs for all language generations for all artman configs.

Also write a failure log for each language that fails.


andreamlin:
PTAL

PR #2292, opened 2018-09-06
go: remove IAM resource handling

pongad:
These become partial GAPICs instead.


andreamlin:
Was IAM only generated in golang clients?

pongad:
Yup. It was there for a while then we realized only pubsub was ever going to use it :P

PR #2293, opened 2018-09-06
Order default retry codes alphabetically

andreamlin:
There's a lot of refactoring to be done for proto annotations support in the near future, and having an ordered list of values in the baseline tests will make for a cleaner baseline diff in future PRs.


andreamlin:
PTAL

PR #2294, opened 2018-09-06
test: fan out code generation

pongad:


pongad:
This seems to have problems of different jobs persisting the same files :(

PR #2295, opened 2018-09-07
[Proto annotations] Fallback to default copyright and license headers

andreamlin:
This is necessary for generating without gapic config, which defines custom copyright and license headers.
\+ unit tests


andreamlin:
PTAL

src/main/java/com/google/api/codegen/util/LicenseHeaderUtil.java
```diff
@@ -0,0 +1,110 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import javax.annotation.Nullable;
+
+public class LicenseHeaderUtil {
+  public static final String DEFAULT_LICENSE_FILE = "license-header-apache-2.0.txt";
+  public static final String DEFAULT_COPYRIGHT_FILE = "copyright-google.txt";
+
+  private LicenseHeaderProto licenseHeader;
```
michaelbausor: This is never assigned to outside the create method. Can we add it to the constructor, and make it (and diagCollector) final?

src/main/java/com/google/api/codegen/util/LicenseHeaderUtil.java
```diff
@@ -0,0 +1,110 @@
...
+
+  public ImmutableList<String> loadCopyrightLines() {
+    try {
+      String filepath;
+      if (licenseHeader == null || Strings.isNullOrEmpty(licenseHeader.getCopyrightFile())) {
+        filepath = DEFAULT_COPYRIGHT_FILE;
+      } else {
+        filepath = licenseHeader.getCopyrightFile();
+      }
+      return ImmutableList.copyOf(getResourceLines(filepath));
```
michaelbausor: `getResourceLines` returns an `ImmutableList` - is it necessary to copy it here (and above)?

src/main/java/com/google/api/codegen/util/LicenseHeaderUtil.java
```diff
@@ -0,0 +1,110 @@
...
+      return ImmutableList.copyOf(getResourceLines(licenseFile));
+
+    } catch (Exception e) {
+      diagCollector.addDiag(Diag.error(SimpleLocation.TOPLEVEL, "Exception: %s", e.getMessage()));
+      throw new RuntimeException(e);
+    }
+  }
+
+  public ImmutableList<String> loadCopyrightLines() {
+    try {
```
michaelbausor: What about pushing the try/catch block into `getResourceLines`, to reduce duplication between the methods?

src/main/java/com/google/api/codegen/util/LicenseHeaderUtil.java
```diff
@@ -0,0 +1,110 @@
...
+  private ImmutableList<String> getResourceLines(String resourceFileName) throws IOException {
+    InputStream fileStream = ConfigProto.class.getResourceAsStream(resourceFileName);
+    if (fileStream == null) {
+      throw new FileNotFoundException(resourceFileName);
+    }
+    InputStreamReader fileReader = new InputStreamReader(fileStream, Charsets.UTF_8);
+    return ImmutableList.copyOf(CharStreams.readLines(fileReader));
+  }
+
+  public DiagCollector getDiagCollector() {
```
michaelbausor: Annotate this as `publicfortesting`? (I'm not sure about this actually, do we do that elsewhere? If you prefer to leave as-is that's fine too).

src/main/java/com/google/api/codegen/util/LicenseHeaderUtil.java
```diff
@@ -0,0 +1,110 @@
...
+        return null;
+      }
+    }
+    return licenseHeaderUtil;
+  }
+
+  public ImmutableList<String> loadLicenseLines() {
+    try {
+      String licenseFile;
+      if (licenseHeader == null || Strings.isNullOrEmpty(licenseHeader.getLicenseFile())) {
```
michaelbausor: Nit: we are enforcing not-null in the `create` method.

src/main/java/com/google/api/codegen/util/LicenseHeaderUtil.java
```diff
@@ -0,0 +1,110 @@
...
+    LicenseHeaderUtil licenseHeaderUtil = new LicenseHeaderUtil(diagCollector);
+
+    if (configProto != null) {
+      licenseHeaderUtil.licenseHeader =
+          configProto
+              .getLicenseHeader()
+              .toBuilder()
+              .mergeFrom(settings.getLicenseHeaderOverride())
+              .build();
+      if (licenseHeaderUtil.licenseHeader == null) {
```
michaelbausor: To confirm the logic for construction here: if configProto is null, we set a null licenseHeader field, and don't use settings.getLicenseHeaderOverride(); if configProto is not null, we do use settings.getLicenseHeaderOverride(). Is that correct?

Also, will `.build()` ever return null?

src/main/java/com/google/api/codegen/util/LicenseHeaderUtil.java
michaelbausor: Oops, no we aren't - deleting.

src/main/java/com/google/api/codegen/util/LicenseHeaderUtil.java
andreamlin: I would rather keep `licenseHeader` inside the class because it is never used outside of the class, and `licenseHeader` is populated by a non-trivial assignment. Keeping the `config != null` checks inside this class will keep the calling code cleaner, because it just has to call the constructor, without doing the null checks.

Made `diagCollector` final.

src/main/java/com/google/api/codegen/util/LicenseHeaderUtil.java
andreamlin: Removed!

src/main/java/com/google/api/codegen/util/LicenseHeaderUtil.java
andreamlin: Done. 
Net loss of 6 LOC

src/main/java/com/google/api/codegen/util/LicenseHeaderUtil.java
andreamlin: Sure.

andreamlin:
Rerunning golang build. Failed last time because of network issues.

src/main/java/com/google/api/codegen/util/LicenseHeaderUtil.java
```diff
@@ -0,0 +1,107 @@
...
+    this.licenseHeader = licenseHeader;
+  }
+
+  public static LicenseHeaderUtil create(
+      @Nullable ConfigProto configProto,
+      @Nullable LanguageSettingsProto settings,
+      @Nullable DiagCollector diagCollector) {
+    Preconditions.checkNotNull(diagCollector);
+    Preconditions.checkArgument(
+        (configProto != null && settings != null) || configProto == null,
```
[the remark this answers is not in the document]
andreamlin: yeep

andreamlin:
ok, fixed! thanks!

PR #2296, opened 2018-09-07
sample: accept CLI arguments

pongad:
To the best of my knowledge this should work, but I haven't tried it
yet.

### DIFF
--- a/src/main/resources/com/google/api/codegen/py/standalone_sample.snip
+++ b/src/main/resources/com/google/api/codegen/py/standalone_sample.snip
@@ -90,14 +90,14 @@
     main()
 @end
 
-@private actualArgs(defaultParams)
-  @join param : defaultParams on ", "
+@private actualArgs(defaultArgs)
+  @join param : defaultArgs on ", "
     args.{@param.identifier}
   @end
 @end
 
-@private formalArgs(defaultParams)
-  @join param : defaultParams on ", "
+@private formalArgs(defaultArgs)
+  @join param : defaultArgs on ", "
     {@param.identifier}
   @end
 @end
--- a/src/main/resources/com/google/api/codegen/py/standalone_sample.snip
+++ b/src/main/resources/com/google/api/codegen/py/standalone_sample.snip
@@ -77,7 +77,8 @@
   def main():
     import argparse
 
-    @# FIXME: Convert argv from strings to the correct types.
+    # FIXME: Convert the args from strings to the correct types;
+    # default values should be correct type as well.
     parser = argparse.ArgumentParser()
     @join param : sample.initCode.argDefaultParams
       parser.add_argument('--{@param.identifier}', default='')
@@ -90,14 +91,14 @@
     main()
 @end
 
-@private actualArgs(defaultArgs)
-  @join param : defaultArgs on ", "
+@private actualArgs(sampleArgs)
+  @join param : sampleArgs on ", "
     args.{@param.identifier}
   @end
 @end
 
-@private formalArgs(defaultArgs)
-  @join param : defaultArgs on ", "
+@private formalArgs(sampleArgs)
+  @join param : sampleArgs on ", "
     {@param.identifier}
   @end
 @end
--- a/src/main/resources/com/google/api/codegen/py/standalone_sample.snip
+++ b/src/main/resources/com/google/api/codegen/py/standalone_sample.snip
@@ -31,7 +31,7 @@
 
     @end
 
-    def sample_{@apiMethod.name}({@sampleFuncParam(sample.initCode.argDefaultParams)}):
+    def sample_{@apiMethod.name}({@sampleFuncParams(sample.initCode.argDefaultParams)}):
       """{@sample.valueSet.description}"""
 
       @# [START {@sample.regionTag}_core]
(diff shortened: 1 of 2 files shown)


src/main/resources/com/google/api/codegen/py/standalone_sample.snip
```diff
@@ -80,14 +80,28 @@
...
+    args = parser.parse_args()
+
+    sample_{@apiMethod.name}({@sampleFuncArgs(sample.initCode.argDefaultParams)})
 
   if __name__ == '__main__':
     main()
 @end
 
-@private sampleFuncParam(defaultParams)
+@private sampleFuncArgs(defaultParams)
```
vchudnov-g: make the function name more descriptive. I suggest something like `sampleFuncArgsFromFlags`, `flagsToSampleFuncArgs`, or `actualArgs`

src/main/resources/com/google/api/codegen/py/standalone_sample.snip
```diff
@@ -80,14 +80,28 @@
...
+  @join param : defaultParams on ", "
+    args.{@param.identifier}
+  @end
+@end
+
+@private sampleFuncParams(defaultParams)
```
vchudnov-g: Consider having something like `formalArgs` in the name.

src/main/resources/com/google/api/codegen/py/standalone_sample.snip
```diff
@@ -80,14 +80,28 @@
   @end
 
   def main():
+    import argparse
+
     @# FIXME: Convert argv from strings to the correct types.
-    sample_{@apiMethod.name}(*sys.argv[1:])
+    parser = argparse.ArgumentParser()
+    @join param : sample.initCode.argDefaultParams
+      parser.add_argument('--{@param.identifier}')
```
vchudnov-g: We should set the default value from the view model, preferably here where we define the flag. (We could also do it when we get the flags, manually substituting the default if no flag was provided. But the flag library should take be the one doing it....)

src/main/resources/com/google/api/codegen/py/standalone_sample.snip
pongad: Done

src/main/resources/com/google/api/codegen/py/standalone_sample.snip
pongad: Done

src/main/resources/com/google/api/codegen/py/standalone_sample.snip
pongad: Done. Setting everything to empty string for now.

pongad:
@vchudnov-g PTAL

src/main/resources/com/google/api/codegen/py/standalone_sample.snip
```diff
@@ -31,7 +31,7 @@
 
     @end
 
-    def sample_{@apiMethod.name}({@sampleFuncParams(sample.initCode.argDefaultParams)}):
+    def sample_{@apiMethod.name}({@formalArgs(sample.initCode.argDefaultParams)}):
```
vchudnov-g: We can do this in a separate CL, but I think it would make things clearer to rename `sample.initCode.argDefaultParams`. As it's used here, these are just the sample param definitions, including but not limited to their default values. Maybe the terminal field should be just `.sampleParams`.

Like I said, we can do this separately.

src/main/resources/com/google/api/codegen/py/standalone_sample.snip
```diff
@@ -85,23 +85,23 @@
...
 
-    sample_{@apiMethod.name}({@sampleFuncArgs(sample.initCode.argDefaultParams)})
+    sample_{@apiMethod.name}({@actualArgs(sample.initCode.argDefaultParams)})
 
   if __name__ == '__main__':
     main()
 @end
 
-@private sampleFuncArgs(defaultParams)
+@private actualArgs(defaultParams)
```
vchudnov-g: In this function and the one below, I would s/`defaultParams`/`sampleArgs`/. 

(Incidentally, we should be consistent with "args" vs "params". We can do that lazily as we work on further changes.)

src/main/resources/com/google/api/codegen/py/standalone_sample.snip
```diff
@@ -85,23 +85,23 @@
...
   @join param : defaultParams on ", "
     args.{@param.identifier}
   @end
 @end
 
-@private sampleFuncParams(defaultParams)
+@private formalArgs(defaultParams)
```
vchudnov-g: Here, too, I suggest s/`defaultParams`/`sampleArgs`/ for clarity

src/main/resources/com/google/api/codegen/py/standalone_sample.snip
```diff
@@ -85,23 +85,23 @@
     @# FIXME: Convert argv from strings to the correct types.
     parser = argparse.ArgumentParser()
     @join param : sample.initCode.argDefaultParams
-      parser.add_argument('--{@param.identifier}')
+      parser.add_argument('--{@param.identifier}', default='')
```
vchudnov-g: Update the `FIXME` above to (a) not refer to `argv` directly and (b) remind us that we will fill in the appropriate default values.

src/main/resources/com/google/api/codegen/py/standalone_sample.snip
vchudnov-g: You changed it to `defaultArgs`, not `sampleArgs` :-)

src/main/resources/com/google/api/codegen/py/standalone_sample.snip
vchudnov-g: (not sure whether you saw this; it got accidentally "resolved" by someone)

Also, I just noticed that this is tagged inconsistently. We are using for `@# TODO` for user-visible comments of things the developer needs to do, and `# FIXME` for things we ourselves need to fix in the templates. Could you s/`@# FIXME`/`# FIXME`/ above?

src/main/resources/com/google/api/codegen/py/standalone_sample.snip
pongad: AHAHAHAH

src/main/resources/com/google/api/codegen/py/standalone_sample.snip
pongad: Done

src/main/resources/com/google/api/codegen/py/standalone_sample.snip
pongad: Oh yeah you're right. Done.

PR #2297, opened 2018-09-07
[Proto Annotations] Set default language package name

andreamlin:
Package names are configured in GAPIC config, but when no GAPIC config is provided, we should give a default package name.


andreamlin:
PTAL

src/main/java/com/google/api/codegen/configgen/mergers/ProtoConfigMerger.java
```diff
@@ -57,14 +56,10 @@ private ConfigMerger createMerger(Model model, String fileName) {
...
-      Interface apiInterface = model.getSymbolTable().lookupInterface(api.getName());
-      if (apiInterface != null) {
-        return apiInterface.getFile().getFullName();
-      }
+    String packageName = ProtoParser.getPackageName(model);
+    if (packageName != null) {
+      return packageName;
     }
-
     helper.error(model.getLocation(), "No interface found");
```
michaelbausor: Should we update this message to something like "Failed to determine package name"?

andreamlin: Done.

andreamlin:
Yes, filed an issue https://github.com/googleapis/gapic-generator/issues/2299! Thanks for pointing that out.

PR #2304, opened 2018-09-11
Add flag support for dev samples.

vchudnov-g:
- This adds --dev_samples to generate standalone samples in all implemented languages, even those that only have stubs; this is useful for development. If not set, which is the default, this will only generate the samples for the production-ready languages.

(artman PR to follow)


vchudnov-g:
Thanks for the lgtm, @hzyi-google . I need an approval to submit. ;-)

hzyi-google:
Ah sorry about that :/

PR #2307, opened 2018-09-12
Tweak output handling in baseline

hzyi-google:
- Renaming all loop variables in response streaming (server, bidi) and paged streaming to `responseItem` (or `response_item`) and the collection variable to `response` so that response handling is consistent for `$resp`.
- Restrict user from defining a variable `response` or `responseItem`.
- This PR touches standalone samples only.
- This PR touches Java and Python only. Nodejs is more complicated since node uses indices to access element rather than a foreach kind of loop. I'll have a separate PR for PHP.

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
@@ -25,6 +25,7 @@ import com.google.api.codegen.viewmodel.OutputView;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -35,6 +36,8 @@ import javax.annotation.Nullable;
 
 class OutputTransformer {
   private static final String RESPONSE_PLACEHOLDER = "$resp";
+  private static final Set<String> RESERVED_KEYWORDS =
+      ImmutableSet.<String>of("response", "response_item");
 
   static List<OutputSpec> defaultOutputSpecs(MethodModel method) {
     if (method.isOutputTypeEmpty()) {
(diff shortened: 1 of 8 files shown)


hzyi-google:
PTAL @vchudnov-g

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -143,6 +153,12 @@ public void run() {
         valueSet.getId(),
         definition.input());
     String identifier = definition.tokenStr();
+    Preconditions.checkArgument(
```
vchudnov-g: Consider factoring out these three new `Preconditions.checkArgument`s  into a helper function like `ensureIdentifierNotReserved` just so we have less repetition and a consistent error message.

src/main/resources/com/google/api/codegen/java/standalone_sample.snip
```diff
@@ -83,7 +83,7 @@
 # since standalone samples can perform more sophisticated response handling
 @private pagedIterableMethodSampleCode(apiMethod, sample)
   {@initCode(sample.initCode)}
-  for ({@apiMethod.listMethod.resourceTypeName} element : {@sampleSyncMethodCall(apiMethod)}.{@apiMethod.listMethod.iterateMethodName}()) {
+  for ({@apiMethod.listMethod.resourceTypeName} responseItem : {@sampleSyncMethodCall(apiMethod)}.{@apiMethod.listMethod.iterateMethodName}()) {
```
vchudnov-g: Hmm, if we already have `getSampleResponseVarName`, then maybe we shouldn't hard-code these variable names here. WDYT, @pongad ?

Let's fix this in a separate PR. For now, add a `#FIXME` near the top of this file.

src/test/java/com/google/api/codegen/transformer/OutputTransformerTest.java
```diff
@@ -68,6 +68,23 @@ public void setUp() {
...
 
+  @Test
+  public void testAccessorNewVariableFailWithReservedKeyword() {
+    Scanner scanner = new Scanner("$resp");
+    when(config.getPageStreaming()).thenReturn(pageStreamingConfig);
+    when(pageStreamingConfig.getResourcesFieldConfig()).thenReturn(resourceFieldConfig);
+    when(namer.getAndSaveElementResourceTypeName(typeTable, resourceFieldConfig))
+        .thenReturn("ShelfBookName");
+    when(featureConfig.useResourceNameFormatOption(resourceFieldConfig)).thenReturn(true);
+    try {
```
vchudnov-g: repeat try/catch with `responseItem`

src/main/resources/com/google/api/codegen/py/standalone_sample.snip
```diff
@@ -60,14 +60,12 @@
       @case "RequestPagedAll"
         {@pagedOptionalArrayMethodSampleCodeAll(apiMethod, sample)}
       @case "RequestStreamingBidi"
-        {@optionalArrayMethodSampleCodeBidiStreaming(apiMethod, sample.initCode)}
-        {@processResponse(sample)}
+        {@optionalArrayMethodSampleCodeBidiStreaming(apiMethod, sample)}
```
vchudnov-g: For clarity/consistency, I would suggest moving the remaining `processResponse` calls out of the `switch` and into the functions being called.

src/main/resources/com/google/api/codegen/java/standalone_sample.snip
hzyi-google: IIUC `getSampleResponseVarName` is only a utility function, and we don't actually save this somewhere in the model or view-model. And therefore hard-coding in snippets is the only way with current infrastructure. But I agree with you that we should fix this.

PR #2308, opened 2018-09-12
Move around test resources, add GapicCodeGeneratorAnnotationsTest

andreamlin:
- Move testdata to a `testdata` directory as is conventional in this codebase.
- Add an empty GapicCodeGeneratorAnnotationsTest.
- Add `protoPackage` to the GapicProductConfig to represent the `--package` option for the proto package; this is a required option if no gapic config is given, and it represents the proto package, out of the protofiles in the descriptor set (which can contain common protos), that we want to generate a client for (e.g. `google.example.library.v1`).


.travis.yml
```diff
@@ -9,7 +9,7 @@ script:
   - ./gradlew check --info
   - ./testutils/check-baselines.sh
   - go test testutils/check-calling-form-in-baseline/*.go
-  - go run $(find testutils/check-calling-form-in-baseline -type f | grep -v _test) -yaml src/test/java/com/google/api/codegen/testsrc/libraryproto/config_not_annotated/library_gapic.yaml src/test
+  - go run $(find testutils/check-calling-form-in-baseline -type f | grep -v _test) -yaml src/test/java/com/google/api/codegen/testsrc/libraryproto/config_not_annotated/testdata/library_gapic.yaml src/test
```
michaelbausor: I'm not sure this is consistent with what we do elsewhere in the codebase. We have a testsrc directory (which this is already under) for files that are inputs to tests, and a testdata directory for baseline files. So I'm not sure that introducing a new testdata folder here makes sense. Is there an example in the existing codebase that this is matching?

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -112,17 +112,26 @@ public GapicProductConfig withPackageName(String packageName) {
    * Creates an instance of GapicProductConfig based on ConfigProto, linking up API interface
    * configurations with specified interfaces in interfaceConfigMap. On errors, null will be
    * returned, and diagnostics are reported to the model.
+   *
+   * @param configProto The parsed set of config files from input
```
michaelbausor: nit: if we are adding javadoc, should we add `@param` takes for all the parameters?

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -112,17 +112,26 @@ public GapicProductConfig withPackageName(String packageName) {
...
+      @Nullable String protoPackage,
+      TargetLanguage language) {
 
     // Get the proto file containing the first interface listed in the config proto, and use it as
     // the assigned file for generated resource names, and to get the default message namespace
     ProtoFile file =
         model.getSymbolTable().lookupInterface(configProto.getInterfaces(0).getName()).getFile();
     String defaultPackage = file.getProto().getPackage();
 
+    // TODO(andrealin): Use protoPackage to determine source protofiles.
```
michaelbausor: Is this intentionally left to a follow-up PR? Right now `protoPackage` is unused in this function, so maybe we should also leave adding it to the signature for the followup PR as well.

src/test/java/com/google/api/codegen/GapicCodeGeneratorAnnotationsTest.java
```diff
@@ -36,20 +36,20 @@ public GapicCodeGeneratorAnnotationsTest(
       String packageConfigFileName,
       List<String> snippetName,
       String apiName,
-      String baseline) {
-    super(language, null, null, snippetName, baseline);
+      String baseline,
+      String protoPackage) {
+    //    super(language, null, null, snippetName, baseline);
+    super(language, null, null, snippetName, baseline, "google.example.library.v1");
```
michaelbausor: Should "google.example.library.v1" instead be `protoPackage`? Otherwise protoPackage is unused.

.travis.yml
andreamlin: I had seen `packagegen/testdata`, `metacode/testdata`, `go/testdata`, `advising/testdata`, and `config/testdata` and assumed it was a thing.

This is a nice setup because there are multiple library.protos in this repo, and say for packagegen tests, we can just test the specific library.proto located at `packagegen/testdata/library.proto`, and have the test class PackageTest.java be in `packagegen`

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
andreamlin: Sure

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
andreamlin: Yes! I have that all in a local branch; and I just wanted to have `protoPackage` be a param here, even though it's unused, to show how the test classes would pass in that param.

src/test/java/com/google/api/codegen/GapicCodeGeneratorAnnotationsTest.java
andreamlin: Yes, thanks :)

andreamlin:
PTAL

.travis.yml
andreamlin: Ok, moved GapicCodeGeneratorAnnotationsTest and its test resources into a new package as described above (except calling the package `protonannotations` instead of `config_not_annotated`.

I can move the other test resources for the gapic tests in a different PR.

andreamlin:
Moved the new *AnnotationsTest.java into its own package. PTAL

src/test/java/com/google/api/codegen/transformer/go/GoGapicSurfaceTransformerTest.java
```diff
@@ -66,7 +66,9 @@ public static void setupClass() {
             locator,
             new String[] {"myproto_gapic.yaml"});
 
-    productConfig = GapicProductConfig.create(model, configProto, TargetLanguage.GO);
+    productConfig =
+        GapicProductConfig.create(
+            model, configProto, "google.example.myproto.v1", TargetLanguage.GO);
```
michaelbausor: Pass in null here, instead of a value? Maaaaybe we should add an overload for `GapicProductConfig.create` that matches the original signature and passes through null? I don't feel strongly about that though, so happy to leave asis if you prefer.

andreamlin: Done.

PR #2312, opened 2018-09-14
Samplegen: PHP Output Handling

hzyi-google:
- PHP output handling
- Also renaming `$result` and `$element` to `$response` and `$responseItem` to make return values consistent with Java and Python.

### DIFF
--- a/src/main/resources/com/google/api/codegen/php/standalone_sample.snip
+++ b/src/main/resources/com/google/api/codegen/php/standalone_sample.snip
@@ -331,7 +331,7 @@
 
 @private processResponse(sample)
     @if sample.outputs.size == 0
-        printf($response . PHP_EOL);
+        print_r($response);
     @else
         {@processOutputViews(sample.outputs)}
     @end
--- a/src/main/resources/com/google/api/codegen/php/standalone_sample.snip
+++ b/src/main/resources/com/google/api/codegen/php/standalone_sample.snip
@@ -330,9 +330,8 @@
 @end
 
 @private processResponse(sample)
-    #{@sample.outputs}
     @if sample.outputs.size == 0
-        print($response);
+        printf($response . PHP_EOL);
     @else
         {@processOutputViews(sample.outputs)}
     @end
@@ -356,15 +355,10 @@
 @end
 
 @private processPrintView(view)
-    # Since numeric comparison seems to not work, we use explicit zero tests
     @if not(view.args.size)
-        print("{@view.format}");
+        printf("{@view.format}" . PHP_EOL);
     @else
-        @if view.format == "%s"
-            print({@variableField(view.args.get(0))});
-        @else
-            printf("{@view.format}", {@concatenatedArgs(view.args)});
-        @end
+        printf("{@view.format}" . PHP_EOL, {@concatenatedArgs(view.args)});
     @end
 @end
 
--- a/src/main/resources/com/google/api/codegen/php/standalone_sample.snip
+++ b/src/main/resources/com/google/api/codegen/php/standalone_sample.snip
@@ -75,7 +75,6 @@
             {@longRunningAsyncSampleCode(sample, apiMethod)}
         @case "Request"
             {@requestSampleCode(sample, apiMethod)}
-            {@processResponse(sample)}
         @case "RequestPaged"
             {@pagedSampleCode(sample, apiMethod)}
         @case "RequestPagedAll"
@@ -86,10 +85,8 @@
             {@bidiStreamingAsyncSampleCode(sample, apiMethod)}
         @case "RequestStreamingClient"
             {@clientStreamingSampleCode(sample, apiMethod)}
-            {@processResponse(sample)}
         @case "RequestStreamingClientAsync"
             {@clientStreamingAsyncSampleCode(sample, apiMethod)}
-            {@processResponse(sample)}
         @case "RequestStreamingServer"
             {@serverStreamingSampleCode(sample, apiMethod)}
         @default
@@ -122,7 +119,7 @@
         @if apiMethod.longRunningView.isEmptyOperation
             // operation succeeded and returns no value
         @else
-            $result = $operationResponse->getResult();
+            $response = $operationResponse->getResult();
             {@processResponse(sample)}
         @end
     } else {
@@ -145,7 +142,7 @@
       @if apiMethod.longRunningView.isEmptyOperation
           // operation succeeded and returns no value
       @else
-          $result = $newOperationResponse->getResult();
+          $response = $newOperationResponse->getResult();
           {@processResponse(sample)}
       @end
     } else {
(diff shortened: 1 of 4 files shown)


src/main/resources/com/google/api/codegen/php/sample_common.snip
```diff
@@ -1,223 +0,0 @@
-@snippet methodCallSampleCode(apiMethod)
```
hzyi-google: Since standalone samples and method samples have pretty much diverged (And may be more different in the future), it seems unnecessary to try to reuse the snippets in `sample_common`. Deleted `sample_common` and let `standalone_sample` and `method_sample` define their own templates. Java and Python follow this pattern too.

src/test/java/com/google/api/codegen/testdata/php/php_library.baseline
```diff
@@ -2762,14 +2762,14 @@ try {
...
+        print($responseItem);
     }
     $stream->closeWrite();
-    $element = $stream->read();
-    while (!is_null($element)) {
-        // doSomethingWith($element)
-        $element = $stream->read();
+    $responseItem = $stream->read();
+    while (!is_null($responseItem)) {
+        print($responseItem);
```
michaelbausor: `print($x)` works for primitive types, but produces an error when called on arrays or protobuf objects.

hzyi-google: Thanks for the quick response! I'll have a PR to move code from sample_common to standalone_sample first to not affect the in method samples. 

For the `print($x)` part, would changing all of them to `var_dump($x)` make sense here?

hzyi-google: Yes. This is a feature requested by DPEs. The idea is here to allow post-processing the response from an API call (printing, defining the return value to a local variable or adding extra comments, etc.) by adding configuration in gapic.yaml. In a normal case, `$x` would never be an array (arrays are handled separately using loops(https://github.com/googleapis/gapic-generator/pull/2312/files/08083d209fdca8da222ef61fd87881e92bcd6da4#diff-36f32ff1b2217a381ca1eaeeea45dfe6R2865)), but it's possible `$x` is a protobuf object.

vchudnov-g:
@hzyi-google Should we hold off on further reviews of this PR, if you're going to split this into two PRs?

hzyi-google:
@vchudnov-g The other PR(#2313) is already submitted, PTAL and I'll rebase this after #2313 is merged.

src/main/resources/com/google/api/codegen/php/method_sample.snip
```diff
@@ -304,4 +304,4 @@
     @default
         $unhandledCase: {@initValue.type}$
     @end
-@end
+@end
```
michaelbausor: Nit: restore newline

src/main/resources/com/google/api/codegen/php/standalone_sample.snip
```diff
@@ -142,80 +142,82 @@
...
       $error = $newOperationResponse->getError();
       // handleError($error)
     }
 @end
 
-@private requestSampleCode(apiMethod)
+@private requestSampleCode(sample, apiMethod)
     @if apiMethod.hasReturnValue
         $response = {@methodCallSampleCode(apiMethod)};
+
```
michaelbausor: Is this newline required?

src/main/resources/com/google/api/codegen/php/standalone_sample.snip
```diff
@@ -326,3 +328,77 @@
         $unhandledCase: {@initValue.type}$
     @end
 @end
+
+@private processResponse(sample)
+    #{@sample.outputs}
```
michaelbausor: Is this line supposed to be in the output? Won't the `#` character make it into a comment in the snippet?

src/main/resources/com/google/api/codegen/php/standalone_sample.snip
```diff
@@ -326,3 +328,77 @@
...
+        @case "PRINT"
+            {@processPrintView(view)}
+        @default
+            $unhandledResponseForm: {@view}
+        @end
+    @end
+@end
+
+@private processPrintView(view)
+    # Since numeric comparison seems to not work, we use explicit zero tests
```
michaelbausor: Should this apply to line 334?

src/test/java/com/google/api/codegen/testdata/php/php_library.baseline
michaelbausor: OK cool. I think `print_r` is preferable to `var_dump`, although I don't have a strong opinion, and the real solution is to make the protobuf objects behave correctly when passed to `print`. For now I would use `print_r`.

michaelbausor: (If you need to convert a protobuf object to a string when working with printf statements, then you could use print_r($x, true), which will return a string instead of printing.)

hzyi-google:
Everything up-to-date! PTAL @michaelbausor @pongad

pongad:
LGTM but @michaelbausor should sign off

beccasaurus:
Samples look great in the .baseline! 🙌

src/main/resources/com/google/api/codegen/php/standalone_sample.snip
```diff
@@ -326,3 +328,71 @@
         $unhandledCase: {@initValue.type}$
     @end
 @end
+
+@private processResponse(sample)
+    @if sample.outputs.size == 0
+        printf($response . PHP_EOL);
```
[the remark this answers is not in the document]
pongad: I think printing proto object should definitely be supported. From reading `print_r` manual, it's possible to pass `$return` argument so that it returns a string. In which case is it possible to do something like this?

```php
printf("foo %s bar", print_r($obj, $return=true));
```
(I don't actually know how PHP syntax works)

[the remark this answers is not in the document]
beccasaurus: Here's my proposal:

 1. For this PR, I'm alright if trying to print a protobuf message object blows up with a **runtime error** 👍
 2. As @michaelbausor said, let's definitely make sure this is on the protobuf teams' radar to fix!
 3. As a follow-up to this PR, to fix the runtime error, let's render PHP that prints *something* about the object *without blowing up* -- let's roll with what you put, @pongad, using `print_r($obj, $return=true)`

/cc @tmatsuo @bshaffer – thoughts?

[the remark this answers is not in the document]
michaelbausor: Yes, and that would be a sensible way to do it I think. Syntax would be:
```php
printf("foo %s bar", print_r($obj, true));
```

(I have just been looking at other options, and can't find anything better. The output of var_dump is much nicer that print_r, but can't be captured in a string. var_export looked promising, but fails because it cannot handle circular references (which all protobuf objects contain in their internals)).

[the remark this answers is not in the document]
michaelbausor: +1 to proposal above from @beccasaurus

PR #2313, opened 2018-09-17
split templates in php/sample_common to standalone_sample and method_sample

hzyi-google:
Split templates in php/sample_common to implement output handling in standalone samples while preserving the current behaviors of in method samples.


hzyi-google:
PTAL

src/main/resources/com/google/api/codegen/php/method_sample.snip
```diff
@@ -94,3 +93,215 @@
         {@optionalField(initCode.optionalFieldSettings)}
     @end
 @end
+
+@snippet methodCallSampleCode(apiMethod)
```
vchudnov-g: Can the various `@snippet`s in the two files be made into `@private`s, or are they used elsewhere? I assume if they're no longer in `sample_common.snip`, they're probably only used in one place?

hzyi-google: It is used by [`smoke_test.snip`](https://github.com/googleapis/gapic-generator/blob/a9d1d05f31e4993be79fc4c60a92354f57f2c671/src/main/resources/com/google/api/codegen/php/smoke_test.snip#L27)

hzyi-google: And yes it is a straight copy operation. I deleted `sample_common.snip` in the next patch.

hzyi-google: PTAL @vchudnov-g

PR #2314, opened 2018-09-17
py: only convert param strings

pongad:
Previously we convert all strings, even the ones that are not function
parameters.


vchudnov-g:
Where is `convertText` used? It's `@private`, but I don't see a call.

pongad:
It's a little hard to find: https://github.com/googleapis/gapic-generator/blob/4e17fb2a3f1bd5c138a14ab549843ebd96c6dcd8/src/main/resources/com/google/api/codegen/py/standalone_sample.snip#L28

PR #2318, opened 2018-09-18
Reorganize testdata

andreamlin:
Create `testdata` subpackages for that test package's test classes to use.
- Move `testdata/discovery` to `discovery/testdata`, and move over the DiscoGapic tests into `discovery`
- Move the Gapic code gen tests into `gapic`, and move its test data into `gapic.testdata`
- Move test classes in package `com.google.api.codegen` to their relevant subpackages.
- Move `discoverytestdata` into `com.google.api.codegen.discovery.testdata`

The existing `testdata` package was getting bloated, and this will make it easier to narrow down which test resource file is used by which test class, when they are siblings in the same package. Later, I want to add baseline files for testing proto annotations, so there will be multiple files called `[language]_library.baseline`, each of which should go in separate directories.


andreamlin:
PTAL

src/main/java/com/google/api/codegen/configgen/DiscoConfigGeneratorApp.java
```diff
@@ -72,22 +74,19 @@ protected void process() throws Exception {
   }
 
   /** Initializes the Discovery document document. */
-  private Document setupDocument() {
+  public static Document setupDocument(String discoveryDocPath, DiagCollector diagCollector) {
```
michaelbausor: Why does this need to be public?

andreamlin: oh yes this can be reverted back to `private static`. thanks

PR #2322, opened 2018-09-19
Parse protofile for Resource Names

andreamlin:


andreamlin:
PTAL

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -429,8 +471,15 @@ public static GapicProductConfig createDummyInstance(
...
       DiagCollector diagCollector,
       ConfigProto configProto,
-      ProtoFile file,
-      TargetLanguage language) {
+      @Nullable List<ProtoFile> sourceProtos,
+      TargetLanguage language,
+      ProtoParser protoParser) {
+    ProtoFile file = null;
+    if (sourceProtos == null) {
+      sourceProtos = new LinkedList<>();
```
pongad: `ImmutableList.of()`? I don't think we're mutating this list right?

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -429,8 +471,15 @@ public static GapicProductConfig createDummyInstance(
   private static ImmutableMap<String, SingleResourceNameConfig> createSingleResourceNameConfigs(
       DiagCollector diagCollector,
       ConfigProto configProto,
-      ProtoFile file,
-      TargetLanguage language) {
+      @Nullable List<ProtoFile> sourceProtos,
+      TargetLanguage language,
+      ProtoParser protoParser) {
+    ProtoFile file = null;
```
pongad: Is this variable used?

src/main/java/com/google/api/codegen/config/ResourceNameMessageConfig.java
```diff
@@ -41,6 +42,17 @@ public static ResourceNameMessageConfig createResourceNameMessageConfig(
     return new AutoValue_ResourceNameMessageConfig(fullyQualifiedMessageName, fieldEntityMap);
   }
 
+  @Nullable
+  public static ResourceNameMessageConfig createResourceNameMessageConfig(Field field) {
```
pongad: This never returns null right?

I'm not very familiar with resource names. Why do we make the assumption on L49?

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
andreamlin: Done.

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
andreamlin: It's used several call stacks deeper, in `FixedResourceNameConfig.createFixedResourceNameConfig()`, `SingleResourceNameConfig.createSingleResourceName()`, `ResourceNameOneofConfig.createResourceNameOneof()`, and possibly something else.

src/main/java/com/google/api/codegen/config/ResourceNameMessageConfig.java
andreamlin: Removed the `@Nullable`. That was some copypasta.

Hmm thinking about it now, we don't have to make that assumption. Removed the comment.

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
pongad: Oh I somehow missed those lines.

andreamlin:
Fixed w.r.t. comments. PTAL

PR #2326, opened 2018-09-20
Bazel-based codegen #4 (MAIN, Java)

vam-google:
0. Bazel version required `0.16.0+`. The rules are expected to be used from `googleapis` workspace (will be comitted soon as a separate PR in googleapis).
1. Introduce `gapic_*` and `java_gapic_*` bazel rules, which allow to generate gapic clients and packages from bazel (**no** `artman` or `docker` involved on any stage).
2. The introduced feature somewhat change from the original design document. The reasons are:
    1. The change in design to also build and run tests on the generated output within bazel itself (i.e. bazel **generates** the gapic code and also **builds** and **runs** the generated output after that, which provides the necessary validation and naturally integrates the built artifacts into any external bazel-based project). The build files (`build.gradle`) for third_party build tools (`gradle`) and src package are generated as "**post-build**" steps.
    2. Various things added/changed as result of bazel-related discoveries/issues met during implementation.
    3. Decision to follow as strict as possible the bazel's best practices and recomendations:
         [Best practices for Bazel](https://docs.bazel.build/versions/master/best-practices.html)
         [BUILD.bazel Style Guide](https://docs.bazel.build/versions/master/skylark/build-style.html)
         [.bzl (extensions) Style Guide](https://docs.bazel.build/versions/master/skylark/bzl-style.html)
         [Java and Bazel](https://docs.bazel.build/versions/master/bazel-and-java.html)
        One of the fundamental best-practices principles followed is reusing existing rules everywhere when possible instead creating custom rules.
    4. Some of the best practices are violated (bazel's documentation itself admits that it is impossible to follow all best practices in one project):
        1. [Deploying Rules](https://docs.bazel.build/versions/master/skylark/deploying.html) - currently rules are in same repository as gapic-generator (not in a separate as suggested). This will change in future, once bazel-based codegen starts replacing artman for releases.
         2. Default public visibility is used for many packages (this is all about API, having stuff public makes perfect sense)
         3. There are no separate bazel package per source package (as suggested for java). This can be changed if we choose bazel as primary build system for gapic-generator (currently it is gradle)

3. Provided public rules and macroses (please note that the distinction between rule and macros is **essential**):
    1. **`gapic_srcjar`** - cross-language (main) **rule** which calls gapic-generator, supplies necessary input files to it (like `gapic_<service>.yaml`, `<service>.yaml` or `discovery_doc.json` in case of discogapic generation). Notice, that `artman_<service>.yaml` (googleapis), `dependencies.yaml` and `api_defaults.yaml` (gapic-genrator) and `package_yaml2` (generated by artman from artman and service yamls) become obsolete (not required at any stage). The chosen name tries to follow bazel best practices, when a rule must be a noun, which represents the output of the rule.
    2. **`proto_custom_library`** - cross-language **rule**, which allows to: 1) call `protoc` with custom plugins; 2) specify the output files format and file extension (essential for consuming (down the chain) rules); 3) accept `proto_library` targets as inputs (to adhere to bazel protobuf best practices and stay consistent with everything else). Unfortunatelly bazel does not have native support for the features above and a similar rule in protobuf repo [proto_gen](https://github.com/protocolbuffers/protobuf/blob/master/protobuf.bzl#L173) satisfies the 1st requirement/feature but not the other two.  Notice that this rule does not replace `proto_library`, it is additive to it (it uses `proto_library`'s output as input, while `proto_library` has raw proto files as input).
    3. **`proto_library_with_info`** - cross-language **macros**, which wraps `proto_custom_library` and provides "fat" proto descriptor. This is not supported by native `proto_library` rule but is required by gapic-generator to include documentation from protos and to overcome the limitation of "only one" desc file as an input in gapic-generator (dependencies' proto descriptors must be supplied somehow, here they are embedded in one big descriptor which then goes to gapic-gen as input).
    4. **`java_gapic_srcjar`** - java **rule**, which does all the java-specific post-processing of the java gapic-generator output (produced by `gapic_srcjar` rule). This includes running formatter and splitting output into main and test packages (a design decision/assumption made, but it is essential (though not strictly required) for proper building of the artifacts within bazel).
    5. **`java_gapic_library`** - java **macros**, which subsequently calls `gapic_srcjar` to generate sources, then `java_gapic_srcjar` to post-process them, then  native `java_library` (twise, one for main and one for tests) to build java binary lib from the generated output within bazel. This is **the rule**, which intends to put gapic-generator on same level of tools integration as protoc (`java_proto_library`) and grpc (`java_grpc_library`).
    6. **`java_discogapic_library`** - java **macros**, which is very similar to `java_gapic_library` and does same, but for discogapic libraries.
    7. **`java_resource_name_proto_library`** - java **macros**, which uses `proto_custom_library` and `java_library` rules to generate resource names sources and then calls `java_library` to build the corresponding library from the generated sources. Notice that resource names can be generated and compiled as a standalone library (do not have to be  a part of `proto-<service>` library).
    8. **`_java_gapic_gradle_pkg`** - **private** java **rule**, used by most other `java*_pkg` (see below) macroses to package stuff (i.e. generate artman-like output).
    9. **`java_gapic_build_configs_pkg`** - java **rule**, which is responsible for generating (from templates) build-specific resources for third_party build systems (like gradle). For example it is used to generate `build.bradle` and `settings.gradle` from templates. Do not confuse this with resource names, it has nothing to do with them.
    10. **`java_gapic_srcs_pkg`** - java **rule**, which organizes generated sources in a "custom" form (expected/defined by third_party build system package, gradle in our case).
    11. **`java_gapic_proto_gradle_pkg`** - java **macros**, which accepts previously built `java_library` (proto classes, and, optionally, resource names classes) and uses other rules to generate java `proto-google-cloud-<service>` packages (identical to the output provided by artman, archived in `.tar.gz` format). The `.tar.gz` format is chosen because it allows to reuse existing `pkg_tar` bazel rule (unfortunately bazel does not have native support for packaging into `.zip` yet).
    12. **`java_gapic_grpc_gradle_pkg`** - java **macros**, which accepts previously built `java_library` (grpc classes) and uses other rules to generate java `grpc-google-cloud-<service>` packages (identical to the output provided by artman, archived in `.tar.gz` format).
    13. **`java_gapic_gapic_gradle_pkg`** - java **macros**, which accepts previously built `java_library` (gapic classes) uses other rules to generate java `gapic-google-cloud-<service>` packages (identical to the output provided by artman, archived in `.tar.gz` format).
    14. **`java_gapic_assembly_gradle_pkg`** - java **macros**, which accept other `*_pkg` targets as input and generates complete self-contained assembly of java clients (identical to the output of batch generation in artman, archived in `.tar.gz` format). The assembly then can be built (this time by a third_party tool `gradle` (not `bazel` itself)) by simply calling `./gradlew clean test`.

4.  Other essential bazel files:
    1. **`rules_gapic/java/java_gapic_pkg_deps_resolution.bzl`** - contains private functions, used by `*_pkg` (see above) rules to "reverse engineer" dependencies and their versions of bazel-built java libraries (the gapic libraries compiled by bazel in `java_gapic_library`). Those dependencies and their versions are then put into `build.gradle` files generated by `java_gapic_build_configs_pkg` rule. This ensures that the dependencies put in `bazel.build` files are the proper ones (the actual dependencies used to build and test actual library by bazel, where bazel, as a proper built tool, does the dependencies resolution). This eliminates the posibility of versions/dependencies mismatch, makes `dependencies.yaml` and `api_defaults.yaml` obsolete, and allows gapic-generator to be a proper code-generating tool, which does not (and should not) manage dependencies (like `protoc`, `javac`, etc which do only generation, the dependencies management is done outside).
    2. **`rules_gapic/java/java_gapic_pkg_repos.bzl`** - this file essentially replaces `artman_<service>.yaml`, `dependencies.yaml` and `api_defaults.yaml` by using bazel itself for dependencies resolution. Previously the dependencies were handled in a form of yaml config values, when they are not valideted to: 1) be correct/exist; 2) match generated code; 3) be sufficient/redundant. *Very important* thing to realize is the difference between `repositories.bzl` and `gapic/java/java_gapic_pkg_repos.bzl`. `repositories.bzl` define dependencies used to compile, run and test gapic-generator itself, `java_gapic_pkg_repos.bzl` define dependencies, used by the code, generated by gapic-generator. Those dependencies intersect, but may have differences (for example because gapic-generator is Java 8 and gapic coutput is Java 7, so they have different guava versions dependencies). When gapic-generator is used from googleapis to build clients, both repositories files are imported (`repositories.bzl` and `gapic/java/java_gapic_pkg_repos.bzl`) in googleapi's workspace. The first one is used to compile gapic-generator, the second one is used to compile and test gapic-generator's output (gapic clients). To deal with dependencies versions mismatch, the `repo_mapping` feature of bazel is supposed to be used (enabled by `--experimental_enable_repo_mapping` bazel command line argument).
    3. **`rules_gapic/java/resources/gradle/*`** - templates for build files (third_party build tool packages generation). Their counterparts inside gapic-generator, like `build_grpc.gradle.snip` are to be removed during later cleanup.
    4. **`BUILD.bazel`** - gapic-generator build file. This is the file used to build gapic-generator itself, and has **nothing** to do with building gapic clients (outputs of gapic-generator).
    5. **`repositories.bzl`** - file, which incapsulates gapic-generator's dependencies. This file is supposed to be used to import gapi-generator dependencies when building it from other workspace (specifically from `googleapis` workspace). The repositories.bzl file seems to be a convention in bazel world (please see [grpc-java/repositories.bzl](https://github.com/grpc/grpc-java/blob/master/repositories.bzl)).


gapic/java/resources/gradle/grpc.gradle.tmpl
```diff
@@ -0,0 +1,47 @@
...
+  classifier = 'sources'
+  from sourceSets.main.allSource
+}
+
+artifacts {
+  archives javadocJar, sourcesJar
+}
+
+compileJava.options.encoding = 'UTF-8'
+javadoc.options.encoding = 'UTF-8'
```
garrettjonesgoogle: add a newline to all of your files missing one at the end

gapic/java/java_gapic_pkg_repos.bzl
```diff
@@ -0,0 +1,237 @@
+# The versions which are most likely to be updated frequently
+_GAX_VERSION = "1.31.0"
+_GRPC_VERSION = "1.13.1"
+_COMMON_PROTOS_VERSION = "1.12.0"
+_AUTH_AERSION = "0.11.0"
```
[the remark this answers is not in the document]
vam-google: Corrected

WORKSPACE
```diff
@@ -1,206 +1,30 @@
...
-maven_jar(
-    name = "com_google_truth_truth",
-    artifact = "com.google.truth:truth:0.42",
+#
+# protoc-java-resource-names-plugin repository dependencies (required to support resource names
+# feature in gapic generator)
+#
+local_repository(
+    name = "com_google_protoc_java_resource_names_plugin_repositories",
+    path = "/usr/local/google/home/vam/_/projects/github/vam-google/protoc-java-resource-names-plugin",
```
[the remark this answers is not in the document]
vam-google: Will do (most probably just update to git_repository(), and commit this after https://github.com/googleapis/protoc-java-resource-names-plugin/pull/38)

gapic/gapic.bzl
```diff
@@ -0,0 +1,152 @@
+def _gapic_srcjar_impl(ctx):
+    output = ctx.outputs.output
+    optional_arguments = []
+    if ctx.attr.package_yaml2:
+        optional_arguments.append("--package_yaml2=%s" % ctx.file.package_yaml2.path)
```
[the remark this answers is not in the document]
vam-google: To support generation of other languages (except java) by monolithic gapic-generator (untill the micgrogens are a thing and the old one is completelly deprecated, which will take a while). The other languages still may require package_yaml2. Note, this is a cross-langauge rule

gapic/gapic.bzl
```diff
@@ -0,0 +1,152 @@
...
+
+gapic_srcjar = rule(
+    attrs = {
+        # src is used instead of srcs, because of the limitation of gapic-generator
+        # (more specifically the api-compiler, which is a dependency of gapic-generator), which
+        # accepts only single descriptor (a fat one, with embedded imports)
+        "src": attr.label(
+            allow_single_file = True,
+            mandatory = True,
+#            providers = [["custom_proto", "proto"]],
```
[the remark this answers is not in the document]
vam-google: Deleted

gapic/BUILD.bazel
```diff
@@ -0,0 +1 @@
+package(default_visibility = ["//visibility:public"])
```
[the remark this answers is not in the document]
vam-google: No real reason. Those files used to be in just root directory, but it was growing and growing so I moved it to a separate directory and defined a structure, which splits common-lang rules/resources and lang-specific resources. If you have good ideas for the name I'll be happy to rename it.

gapic/java/java_gapic.bzl
```diff
@@ -0,0 +1,142 @@
...
+def _java_gapic_srcjar_impl(ctx):
+    gapic_srcjar = ctx.file.gapic_srcjar
+    output_main = ctx.outputs.main
+    output_test = ctx.outputs.test
+    formatter = ctx.executable.formatter
+
+    output_dir_name = ctx.label.name
+    output_dir_path = "%s/%s" % (output_main.dirname, output_dir_name)
+
+    # Note the script is more complicated thant it intuitively should be because of limitations
```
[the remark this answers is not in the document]
vam-google: Corrected

gapic/java/java_gapic.bzl
```diff
@@ -0,0 +1,142 @@
...
+    # inherent to bazel execution environment: no absolute paths allowed, the generated artifacts
+    # must ensure uniquness within a build.
```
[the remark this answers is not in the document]
vam-google: Corrected

gapic/java/java_gapic.bzl
```diff
@@ -0,0 +1,142 @@
...
+    script = """
+    unzip -q {gapic_srcjar} -d {output_dir_path}
+    echo '--replace' > {output_dir_path}/{output_dir_name}.txt
+    find {output_dir_path} -type f >> {output_dir_path}/{output_dir_name}.txt
+    {formatter} @{output_dir_path}/{output_dir_name}.txt
+    pushd .
+    cd {output_dir_path}
+    cd src/main/java && zip -q -r ../../../{output_dir_name}.srcjar ./* && cd ../../..
```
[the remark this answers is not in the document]
vam-google: Changed to `cd -`

gapic/java/java_gapic.bzl
```diff
@@ -0,0 +1,142 @@
...
+
+    java_gapic_srcjar(
+        name = srcjar_target_name,
+        gapic_srcjar = srcjar_raw_target_name,
+        visibility = visibility,
+    )
+
+    lib_deps = [
+        "@com_google_protobuf_protobuf_java//jar",
+#        "@com_google_api_api_common//jar",
```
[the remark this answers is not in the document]
vam-google: So currently artman generation includes api-common library for every library, which is not in `JavaPackageTransformer.SERVICES_WITH_NO_RESOURCE_NAMES`, but that list hasn't been properly updated, thus artman includes api-common in many libs, which do not really need it. Not including it by default, means people need to include it manually for every lib with resource name. In other words, it is confusing and I did not know what to do with api_common, so just left it commented to indicate it...

gapic/java/java_gapic_pkg.bzl
```diff
@@ -0,0 +1,276 @@
...
+        expanded_templates.append(expanded_template)
+        ctx.actions.expand_template(
+            template = template[0].files.to_list()[0],
+            substitutions = substitutions,
+            output = expanded_template,
+        )
+
+    # Note the script is more complicated than it intuitively should be because of the limitations
+    # inherent to bazel execution environment: no absolute paths allowed, the generated artifacts
+    # must ensure uniquness within a build. The template output directory manipulations are
```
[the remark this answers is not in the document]
vam-google: Corrected

gapic/java/java_gapic_pkg.bzl
```diff
@@ -0,0 +1,276 @@
...
+    # to modify defaul 555 file permissions on any generated by bazel file (exectuable read-only,
```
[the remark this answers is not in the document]
vam-google: Corrected

gapic/java/java_gapic_pkg.bzl
```diff
@@ -0,0 +1,276 @@
...
+    # which is not at all what we need for build files). There is no bazel built-in way to change
+    # the generated files pemissions also the actual files accessible by the script are symlinks and
```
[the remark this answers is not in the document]
vam-google: Corrected

gapic/java/java_gapic_pkg.bzl
```diff
@@ -0,0 +1,276 @@
...
+    test_srcs = []
+    for test_src_dep in ctx.attr.test_deps:
+        if is_source_dependency(test_src_dep):
+            test_srcs.extend(test_src_dep.java.source_jars.to_list())
+
+    paths = _construct_package_dir_paths(ctx.attr.package_dir, ctx.outputs.pkg, ctx.label.name)
+
+    # Note the script is more complicated thant it intuitively should be because of limitations
+    # inherent to bazel execution environment: no absolute paths allowed, the generated artifacts
+    # must ensure uniquness within a build.
```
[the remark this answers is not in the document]
vam-google: Corrected

gapic/java/java_gapic_pkg.bzl
```diff
@@ -0,0 +1,276 @@
...
+def _construct_package_dir_paths(attr_package_dir, out_pkg, label_name):
+    if attr_package_dir:
+        package_dir = attr_package_dir
+        package_dir_expr = "../{}/*".format(package_dir)
+    else:
+        package_dir = label_name
+        package_dir_expr = "./*"
+
+    # We need to include label in the path to eliminate possible output files duplicates
+    # (labes are guaranteed to be unique by bazel itself)
```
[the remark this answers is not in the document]
vam-google: Corrected

gapic/java/java_gapic_pkg_deps_resolution.bzl
```diff
@@ -0,0 +1,158 @@
...
+def _reconstruct_artifact_id(file, group_overrides = {}):
+    """
+    This method reconstruct artifact id in the form of `group:name:version:classifier` from the
+    corresponding jar dependency file's full path. This implies that maven_jar targets are named
+    following the convention documented in bazel's official documentation (a relatively safe
+    assumption).
+
+    The aforementioned bazel documentation link: https://docs.bazel.build/versions/master/be/workspace.html#maven_jar
+
+    The implementation apporach is dictated by the fact that maven_jar rule and corresponding java_*
```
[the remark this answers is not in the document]
vam-google: Corrected

gapic/java/java_gapic_pkg_deps_resolution.bzl
```diff
@@ -0,0 +1,158 @@
...
+    bazel rules drop the original artifact information (after downloading the jar is completelly
```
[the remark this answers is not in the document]
vam-google: Corrected

gapic/java/java_gapic_pkg_deps_resolution.bzl
```diff
@@ -0,0 +1,158 @@
...
+    "demavenized"). Also there is really no good way of retreiving the value of the "artifact"
+    attribute value (passed to maven_jar rule) during analysis phase (note, native.existing_rule()
+    is not available durin analysis phase).
```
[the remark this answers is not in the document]
vam-google: Corrected

gapic/java/java_gapic_pkg_deps_resolution.bzl
```diff
@@ -0,0 +1,158 @@
...
+
+    There is no support for regular expressions in Starlark, so all the string shenanigans are done
```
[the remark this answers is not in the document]
vam-google: Skylark reached the higher limits and reached the stars ([proof link](https://docs.bazel.build/versions/master/skylark/language.html)). When you were making the design it was Skylark, then changed the name right in the middle of me working on this =)

vam-google:
@garrettjonesgoogle PTAL

gapic/java/java_gapic_pkg_deps_resolution.bzl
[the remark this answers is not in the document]
garrettjonesgoogle: *shakes head and sighs*

gapic/java/java_gapic_pkg.bzl
```diff
@@ -0,0 +1,302 @@
...
+        )
+
+    # Note the script is more complicated than it intuitively should be because of the limitations
+    # inherent to bazel execution environment: no absolute paths allowed, the generated artifacts
+    # must ensure uniqueness within a build. The template output directory manipulations are
+    # to modify default 555 file permissions on any generated by bazel file (exectuable read-only,
+    # which is not at all what we need for build files). There is no bazel built-in way to change
+    # the generated files permissions, also the actual files accessible by the script are symlinks
+    # and `chmod`, when applied to a directory, does not change the attributes of symlink targets
+    # inside the directory. Chaning the symlink targets permissions is also not an option, because
```
garrettjonesgoogle: spelling (Chaning); s/targets/target's/

gapic/java/java_gapic_pkg.bzl
```diff
@@ -0,0 +1,302 @@
...
+    # they are on read-only file system.
```
garrettjonesgoogle: s/on read-only/on a read-only/

gapic/java/java_gapic_pkg.bzl
```diff
@@ -0,0 +1,302 @@
...
+    outputs = {"pkg": "%{name}.tar.gz"},
+    implementation = _java_gapic_srcs_pkg_impl,
+)
+
+def java_gapic_proto_gradle_pkg(
+        name,
+        deps,
+        test_deps = None,
+        visibility = None,
+        group = "com.google.api.grpc",
```
garrettjonesgoogle: I'm uneasy about using a google namespace as a default in a build rule that could be used by anyone

vam-google: It is not a build rule, it is a pkg rule, which creates gradle packages. It is easy to not have it there, but it will mean that everybody in googleapis will have to repeatedly set the group. 

Note, that artman also hardcodes that and much deeper: it is hardcoded in .snip files for build gradle files themselves (this was the main reason, why I decided that it is ok to have it like this for now).

gapic/java/java_gapic_pkg.bzl
vam-google: done

gapic/java/java_gapic_pkg.bzl
vam-google: done

vam-google:
@garrettjonesgoogle PTAL (added support for discogapic)

gapic/java/java_gapic_pkg_repos.bzl
```diff
@@ -0,0 +1,276 @@
+# The versions which are most likely to be updated frequently
+_PROTOBUF_VERSION = "3.6.1"
+_GAX_VERSION = "1.32.0"
+_GAX_BETA_VERSION = "0.49.0"
+_GRPC_VERSION = "1.13.1"
+_COMMON_PROTOS_VERSION = "1.13.0-pre1"
```
garrettjonesgoogle: I don't think we want to depend on the pre version here

src/main/java/com/google/api/codegen/ArtifactType.java
```diff
@@ -28,9 +28,13 @@
 
   // This will be split into GAPIC and GAPIC_PACKAGE
   LEGACY_GAPIC_AND_PACKAGE,
-  // This will be split into DISCOGAPIC and DISCOGAPIC_PACKAGE
+  // This will be split into DISCOGAPIC_CODE and DISCOGAPIC_PACKAGE
```
garrettjonesgoogle: The comment for `LEGACY_GAPIC_AND_PACKAGE` should be corrected too

gapic/java/java_gapic_pkg_deps_resolution.bzl
```diff
@@ -0,0 +1,190 @@
...
+                artifact_id[3] = "-".join(chunks[i: len(chunks)])
+                break
+            else:
+                artifact_id[2] = "-".join([artifact_id[2], chunks[i]])
+
+    # Reconstructing `group` portion from dirname (full path to dependency jar). This assumes that
+    # maven_jar's target names follow official naming best practices.
+    if artifact_id[1]:
+        if group_overrides.get(artifact_id[1]):
+            # Simply use overriden (manually specified) group name for the artifact name, if provided
```
garrettjonesgoogle: spelling (overriden)

gapic/java/java_gapic_pkg_deps_resolution.bzl
```diff
@@ -0,0 +1,190 @@
...
+        elif not artifact_id[3]:
+            # Everything, which comes after version chunk and does not have digits in it is
+            # recognized as a classifier chunk, otherwise (if there is at least one digit) it is
+            # recognized as a continuation of the version and appended to the artifact version
+            # portion accordingly.
+            if _possibly_artifact_id_classifier_chunk(chunks[i]):
+                artifact_id[3] = "-".join(chunks[i: len(chunks)])
+                break
+            else:
+                artifact_id[2] = "-".join([artifact_id[2], chunks[i]])
```
garrettjonesgoogle: Is there any way to write unit tests for starlark code? This seems heavy on raw logic, which is something I feel has the most value for being tested.

gapic/java/java_gapic_pkg.bzl
```diff
@@ -0,0 +1,313 @@
...
+            proto_srcs.extend(src_dep.proto.check_deps_sources.to_list())
+
+    test_srcs = []
+    for test_src_dep in ctx.attr.test_deps:
+        if is_source_dependency(test_src_dep):
+            test_srcs.extend(test_src_dep.java.source_jars.to_list())
+
+    paths = _construct_package_dir_paths(ctx.attr.package_dir, ctx.outputs.pkg, ctx.label.name)
+
+    # Note the script is more complicated thant it intuitively should be because of limitations
```
garrettjonesgoogle: spelling (thant)

gapic/java/java_gapic_pkg.bzl
garrettjonesgoogle: Well, it's much harder to customize artman. Using Bazel should allow us to dig out these assumptions. I had in mind a Cloud-specific package rule that defaulted this package.

vam-google: Ok, I moved the group/version to the required attributes and googleapis/discovery-artifact-manager repositories will specify them for each artifact

gapic/java/java_gapic_pkg_repos.bzl
vam-google: Yes, I went too far with upgrading the protos version (`1.13.0-pre1` is the dependency of the generator, but the generated artifacts depend on `1.12.0`)

src/main/java/com/google/api/codegen/ArtifactType.java
vam-google: Updated, also changed "will be - is" (but not for discogapic, because there is no discogapic_package)

gapic/java/java_gapic_pkg_deps_resolution.bzl
vam-google: Corrected

gapic/java/java_gapic_pkg_deps_resolution.bzl
vam-google: There is a way to test starlark rules: https://docs.bazel.build/versions/master/skylark/testing.html, but please lets not do it now. As for this particular logic, I believe the proper way of reducing complexity here would be to request from bazel team to have `maven_jar` rule to return a provider with the actual artifact_id value in t (i.e. to not drop maven-specific info too early). This would allow to get rid from 90% of the `java_gapic_pkg_deps_resolution.bzl` file and also remove all the heuristics. Unless i'm missing something, this should be easy to implement (the only potential problem is that bazel team may not want to keep that information for some reason).

gapic/java/resources/gradle/client.gradle.tmpl
```diff
@@ -0,0 +1,55 @@
+buildscript {
+  repositories {
+    mavenCentral()
+  }
+}
+
+apply plugin: 'java'
+
+description = 'PROTO library for {{name}}'
```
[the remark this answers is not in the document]
vam-google: Oops =) Nice catch. It should be GAPIC library (to stay consistent with artman's output)

gapic/java/java_gapic_pkg_deps_resolution.bzl
garrettjonesgoogle: Alright - let's do at least one of those things (add tests or get Bazel to return the data from maven_jar that we want); I don't want to have complex, untested rule code indefinitely.

gapic/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+#### Java
+1. **`java_gapic_srcjar`** - java **rule**, which does all the java-specific post-processing of the java gapic-generator output (produced by `gapic_srcjar` rule). This includes running formatter and splitting output into main and test packages (a design decision/assumption made, but it is essential (though not strictly required) for proper building of the artifacts within bazel).
+2. **`java_gapic_library`** - java **macros**, which subsequently calls `gapic_srcjar` to generate sources, then `java_gapic_srcjar` to post-process them, then  native `java_library` (twise, one for main and one for tests) to build java binary lib from the generated output within bazel. This is **the rule**, which intends to put gapic-generator on same level of tools integration as protoc (`java_proto_library`) and grpc (`java_grpc_library`).
```
eoogbe: s/twise/twice

gapic/README.md
```diff
@@ -0,0 +1,30 @@
...
+3. **`proto_library_with_info`** - cross-language **macros**, which wraps `proto_custom_library` and provides "fat" proto descriptor. This is not supported by native `proto_library` rule but is required by gapic-generator to include documentation from protos and to overcome the limitation of "only one" desc file as an input in gapic-generator (dependencies' proto descriptors must be supplied somehow, here they are embedded in one big descriptor which then goes to gapic-gen as input).
```
eoogbe: s/macros/macro here and below

gapic/README.md
```diff
@@ -0,0 +1,30 @@
...
+3. **`java_discogapic_library`** - java **macros**, which is very similar to `java_gapic_library` and does same, but for discogapic libraries.
+4. **`java_resource_name_proto_library`** - java **macros**, which uses `proto_custom_library` and `java_library` rules to generate resource names sources and then calls `java_library` to build the corresponding library from the generated sources. Notice that resource names can be generated and compiled as a standalone library (do not have to be  a part of `proto-<service>` library).
+5. **`_java_gapic_gradle_pkg`** - **private** java **rule**, used by most other `java*_pkg` (see below) macroses to package stuff (i.e. generate artman-like output).
+6. **`java_gapic_build_configs_pkg`** - java **rule**, which is responsible for generating (from templates) build-specific resources for third_party build systems (like gradle). For example it is used to generate `build.bradle` and `settings.gradle` from templates. Do not confuse this with resource names, it has nothing to do with them.
```
eoogbe: s/bradle/gradle

gapic/README.md
```diff
@@ -0,0 +1,30 @@
...
+2. **`java_gapic_library`** - java **macros**, which subsequently calls `gapic_srcjar` to generate sources, then `java_gapic_srcjar` to post-process them, then  native `java_library` (twise, one for main and one for tests) to build java binary lib from the generated output within bazel. This is **the rule**, which intends to put gapic-generator on same level of tools integration as protoc (`java_proto_library`) and grpc (`java_grpc_library`).
+3. **`java_discogapic_library`** - java **macros**, which is very similar to `java_gapic_library` and does same, but for discogapic libraries.
+4. **`java_resource_name_proto_library`** - java **macros**, which uses `proto_custom_library` and `java_library` rules to generate resource names sources and then calls `java_library` to build the corresponding library from the generated sources. Notice that resource names can be generated and compiled as a standalone library (do not have to be  a part of `proto-<service>` library).
+5. **`_java_gapic_gradle_pkg`** - **private** java **rule**, used by most other `java*_pkg` (see below) macroses to package stuff (i.e. generate artman-like output).
```
eoogbe: s/macroses/macros

gapic/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+### Requirements
+
+- Bazel version `0.16.0+`.
+- Linux (may work on other platforms, but haven't been tested)
+
+### Rules and Macroses
+
+#### Cross-language
+1. **`gapic_srcjar`** - cross-language (main) **rule** which calls gapic-generator, supplies necessary input files to it (like `gapic_<service>.yaml`, `<service>.yaml` or `discovery_doc.json` in case of discogapic generation). Notice, that `artman_<service>.yaml` (googleapis), `dependencies.yaml` and `api_defaults.yaml` (gapic-genrator) and `package_yaml2` (generated by artman from artman and service yamls) become obsolete (not required at any stage). The chosen name tries to follow bazel best practices, when a rule must be a noun, which represents the output of the rule.
```
eoogbe: s/Notice,/Notice

gapic/README.md
```diff
@@ -0,0 +1,30 @@
...
...encies resolution. Previously the dependencies were handled in a form of yaml config values, when they are not valideted to: 1) be correct/exist; 2) match generated code; 3) be sufficient/redundant. *Very important* thing to realize is the difference between `repositories.bzl` and `gapic/java/java_gapic_pkg_repos.bzl`. `repositories.bzl` define dependencies used to compile, run and test gapic-generator itself, `java_gapic_pkg_repos.bzl` define dependencies, used by the code, generated by gapic-generator. Those dependencies intersect, but may have differences (for example because gapic-generator is Java 8 and gapic coutput is Java 7, so they have different guava versions dependencies). When gapic-generator is used from googleapis to build clients, both repositories files are imported (`repositories.bzl` and `gapic/java/java_gapic_pkg_repos.bzl`) in googleapi's workspace. The first one is used to compile gapic-generator, the second one is used to compile and test gapic-generator's output (gapic clients). To deal with dependencies versions mismatch, the `repo_mapping` feature of bazel is supposed to be used (enabled by `--experimental_enable_repo_mapping` bazel command line argument).
```
eoogbe: s/valideted/validated
s/both repositories/both repository's
s/googleapi's/googleapis's

gapic/README.md
```diff
@@ -0,0 +1,30 @@
...
+7. **`java_gapic_srcs_pkg`** - java **rule**, which organizes generated sources in a "custom" form (expected/defined by third_party build system package, gradle in our case).
+8. **`java_gapic_proto_gradle_pkg`** - java **macros**, which accepts previously built `java_library` (proto classes, and, optionally, resource names classes) and uses other rules to generate java `proto-google-cloud-<service>` packages (identical to the output provided by artman, archived in `.tar.gz` format). The `.tar.gz` format is chosen because it allows to reuse existing `pkg_tar` bazel rule (unfortunately bazel does not have native support for packaging into `.zip` yet).
+9. **`java_gapic_grpc_gradle_pkg`** - java **macros**, which accepts previously built `java_library` (grpc classes) and uses other rules to generate java `grpc-google-cloud-<service>` packages (identical to the output provided by artman, archived in `.tar.gz` format).
+10. **`java_gapic_gapic_gradle_pkg`** - java **macros**, which accepts previously built `java_library` (gapic classes) uses other rules to generate java `gapic-google-cloud-<service>` packages (identical to the output provided by artman, archived in `.tar.gz` format).
```
eoogbe: s/java_gapic_gapic_gradle_pkg/java_gapic_client_gradle_pkg

vam-google:
@vchudnov-g:
Thanks for the feedback!
1. Added README.md (also renamed to rules_gapic, to be more consistent with how bazel extensions are [structured](https://github.com/bazelbuild?utf8=%E2%9C%93&q=rules&type=&language=)). 
2. In `proto_custom_library` the most important feature it adds on top of `proto_library` is an ability to call custom plugins. Also, since it can be any plugin (including a microgenerator plugin) `proto_custom_library` can actually generate anything, so I tried to keep the name as generic as possible. Specifically, in case of microgenerators, implemented as protoc plugins, `proto_custom_library` is supposed to be called instead of `gapic_srcjar` to generate the sources.
3. Renamed `java_gapic_build_resources_pkg` to `java_gapic_build_configs_pkg`
4. `java_gapic_build_configs_pkg` and `java_gapic_srcs_pkg` are not gradle-specific, that is why they do not have `gradle` in their names. The actual template (the gradle-specific portion) is a parameter to `java_gapic_build_configs_pkg`. `java_gapic_srcs_pkg` packages sources and proto files. The folder structure it create is specific to java (that is why it has `java` in its name) but independent from build tool used (both gradle and maven will be happy with it). In the README.md I changed the order of rules and a little bit changed the doc structure to make this less confusing.
5. Corrected the `java_gapic_assembly_gradle_pkg` documentation issue (wrong name used).
6. Yes, `java_gapic_library`accepts a descriptor. The descriptor can be a file on a disc or an output of another rule. In reality that parameter is always an output of `proto_library` (it looks like exactly what you are asking for). There is no restriction on accepting only rules with "proto" provider (similar to how java_grpc_library is implemented, which makes it more restrictive and allows to fail fast on analysis phase). The reason for that is that the underlying rule used for the actual generation (`gapic_srcjar`) supports both gapic and discogapic clients generation, and in case of discogapic the `src` parameter is discovery `.json` file instead of protobuf descriptor.
7. Making the `_pkg` depending on success of the tests sounds really good. I think it is a big and isolated enough piece of work to be accomplished in another PR. Also it is not that easy in bazel, because it is "too smart". Bazel thinks in terms of the "actual product" of the build, which is a file or a set of files. When build is executed bazel tries to do as **minimal** job as possible to achieve the goal: output requested **set of files**. Bazel **aggressively** cuts all build DAG (direct acyclic graph) branches which do not have impact on the output of the generated files even if tey are explicitly specified as "dependency" of the target "file". Specifically, since the generated package has exactly **same content regardless** of the success of the tests, bazel will not even try to call them (bazel is able to detect that tests are "irrelevant" as early as analysis phase is completed). Another problem is that tests are exected as part of `bazel test` command and package is built as part of `bazel build`. I'm not saying it is impossible to somehow ensure that tests pass before generating a package, but it is more tricky than in other build tools, where adding dependency would be just sufficient. Bazel is smart... sometimes too smart. In Russian there is a common phrase "grief from mind" to describe this situation =).

vam-google:
@vchudnov-g I've addressed the README.md comments. I'm not opposed to providing the documentation about dependencies (we can add it later), but pushing it now so I can unblock Eva.

PR #2328, opened 2018-09-20
Parse service hostname from proto annotations

andreamlin:
Adding testing using mocks, and testing of the ProtoParser itself.


andreamlin:
PTAL

src/test/java/com/google/api/codegen/config/ServiceConfigTest.java
```diff
@@ -0,0 +1,87 @@
...
+import com.google.api.codegen.util.ProtoParser;
+import com.google.api.tools.framework.model.Interface;
+import com.google.api.tools.framework.model.Model;
+import java.util.Arrays;
+import java.util.List;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+public class ServiceConfigTest {
```
vam-google: Unfortunately this test is not properly testing the added functionality. It consists of:
- Mocks
- Protobuf objects with some values set (`Service serviceYaml`), plus `ProtoInterfaceModel` (which is also just a set of properties) -  i.e. **test data**.
- But there is no a non-mock functional object (the one, which logic we want to test (protobuf and other POJO objects do not fit, because they are just data, but not the functionality tested by this test)), or a static method, which functionality we want to test. 

Specifically, this PR introduces the following new methods (some of them have actual logic, which is worth to be tested):
- `ProtoApiModel#getAuthScopes()`
- `ProtoApiModel#getAuthScopes(ProtoParser, List<ProtoInterfaceModel>)`
- `DiscoGapicInterfaceContext#getDefaultHost()`
- `GapicInterfaceContext.java#getDefaultHost()`
- `ProtoParser#getDefaultHost()`
- `ProtoParser#getAuthScopes()`

It looks like none of these methods will be called as part of this test execution (the mocks will be, but not the actual methods, i.e. if we check test coverage, the corresponding lines will be reported as untested).

src/test/java/com/google/api/codegen/config/ServiceConfigTest.java
```diff
@@ -0,0 +1,87 @@
...
+                "https://www.googleapis.com/auth/food", "https://www.googleapis.com/auth/default"));
+    Mockito.when(protoModel.getServiceConfig()).thenReturn(serviceYaml);
+  }
+
+  @Test
+  public void testServiceConfigFromProtoFile() {
+    List<ProtoInterfaceModel> interfaces =
+        Arrays.asList(
+            new ProtoInterfaceModel(libraryService), new ProtoInterfaceModel(foodService));
+    List<String> scopes = apiModel.getAuthScopes(protoParser, interfaces);
```
vam-google: Note, `apiModel` is a mock object, which depends only on other mock objects (except `ProtoInterfaceModel`, but that class was not changed in this PR, and it cannot be the subject of testing here). Thus the following assert statements check that mocks return proper values, but do not check that the actual thing.

src/test/java/com/google/api/codegen/config/ServiceConfigTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+public class ServiceConfigTest {
+  private static final Model protoModel = Mockito.mock(Model.class);
+  private static final ProtoApiModel apiModel = new ProtoApiModel(protoModel);
+  private static final ProtoParser protoParser = Mockito.mock(ProtoParser.class);
+  private static final Interface libraryService = Mockito.mock(Interface.class);
+  private static final Interface foodService = Mockito.mock(Interface.class);
+
+  private static final Service serviceYaml =
+      Service.newBuilder()
```
vam-google: Please use intermediate variables in cases like this, so the final initialization block looks more readable and has less indentations (here it has 9 levels of indentation, ideally there should be no more than 3-4 levels of indentation in any code block).

src/main/java/com/google/api/codegen/transformer/DiscoGapicInterfaceContext.java
```diff
@@ -328,4 +328,9 @@ public DiscoGapicMethodContext asFlattenedMethodContext(
   public String serviceTitle() {
     return getDocument().name();
   }
+
+  @Override
+  public String getDefaultHost() {
+    return getDocument().baseUrl();
```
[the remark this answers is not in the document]
andreamlin: Ok, renamed `getDefaultHost` to `getServiceAddress`

src/test/java/com/google/api/codegen/config/ServiceConfigTest.java
andreamlin: Done.

src/test/java/com/google/api/codegen/config/ServiceConfigTest.java
andreamlin: Correct, this test tests `ProtoApiModel#getAuthScopes()` on the surface, and only verifies that the resulting `List<String>` contains the Strings that we expect, given whatever the mocks return.

I will soon have out a PR for baseline tests for testing the ProtoParser functionality.

andreamlin: `apiModel` is not a mock object because it was created using the `ProtoApiModel` constructor.

ProtoParserTest has been created to test ProtoParser.

vam-google: My apologies for misreading that apiModel is created using `ProtoApiModel` constructor. I "double checked" multiple times before making such a strong statement, but somehow I "saw" all being mocks in the initialization block. (maybe because ProtoApiModel has Model in its name and I read it  as "Mock" (looks similar)). With ProtoApiModel being initialized properly this test makes much more sense, but then it tests specifically `ProtoApiModel` class and `getAuthScopes`. Why isn't this test called `ProtoApiModelTest`, and the test method is not called `testGetAuthScopes()`?

src/test/java/com/google/api/codegen/util/ProtoParserTest.java
```diff
@@ -0,0 +1,156 @@
...
+import com.google.api.tools.framework.model.testing.TestDataLocator;
+import com.google.longrunning.OperationTypes;
+import com.google.rpc.Code;
+import java.util.List;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+public class ProtoParserTest {
```
vam-google: Why does `ProtoParser` class uses a mix of static/non-static methods in a way it does (I can't understand a pattern)? Currently `ProtoParser` looks like a mix of a singleton, stateless helper class (class with public constructor but not data/fields/state in it) and a utility class with static methods only:
- it has `getProtoParser()` (i.e. `getInstance()`) - an attribute of a singleton (though constructor is public);
- it has non-static methods with some functionality in them (not just getters/setters), a public constructor and no state/fields - an attribute of a stateless helper class;
- it has static methods with some functionality, not related to instantiating the class itself (or doing something specific, which forces the method to be static) - an attribute a utilities class with static-only methods;

 Specifically, why `getResourceEntityName(Field)` is static and `getLongRunningOperation(Method)` is not, even though it does not need any internal state? In general, having stateless helper class is much more flexible, than having a utility class with only-static methods or a singleton. For example mocking utility class with static methods is more difficult. 

In case if there is no very concrete reason to keep some methods static and some not, please convert `ProtoParser` to a stateless helper class (no static methods,  public constructor, no need to have `getProtoParser()`, no reason to be afraid of "overinstantiating" the class, everywhere when needed, it is done super fast and jvm is optimized for this kind of thing extremely well (unless you do it in ridiculous quantities, of course, but it is not the case)), this will allow keep things consistent and flexible (easy to mock/substitute with different implementation, if needed in certain cases).

src/main/java/com/google/api/codegen/util/ProtoParser.java
```diff
@@ -56,8 +57,14 @@ public static String getPackageName(Model model) {
       if (apiInterface != null) {
         return apiInterface.getFile().getFullName();
       }
+      return null;
+    } else {
+      Iterator<ProtoElement> rootsIterator = model.getRoots().iterator();
+      if (rootsIterator.hasNext()) {
+        return rootsIterator.next().getFullName();
```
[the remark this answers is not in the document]
andreamlin: Hmm, actually I don't anticipate the `else` block being called, so I'll remove the block and the test for this.

andreamlin:
Nodejs test CI test failing because of https://github.com/googleapis/gapic-showcase/issues/25.

PR #2329, opened 2018-09-20
Add tests for printing nested messages

hzyi-google:
Currently we don't have tests for printing nested messages in output handling. Adding these would give us more confidence in generating samples.

### DIFF
--- a/src/test/java/com/google/api/codegen/testsrc/libraryproto/library_gapic.yaml
+++ b/src/test/java/com/google/api/codegen/testsrc/libraryproto/library_gapic.yaml
@@ -262,7 +262,7 @@ interfaces:
         - "That's all!"
       - print:
         - "series_uuid: %s"
-        - $resp.series_uuid
+        - $resp.series_uuid.series_bytes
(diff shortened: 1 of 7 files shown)


src/test/java/com/google/api/codegen/testsrc/libraryproto/library_gapic.yaml
```diff
@@ -260,6 +260,9 @@ interfaces:
               - title
       - print:
         - "That's all!"
+      - print:
+        - "series_uuid: %s"
+        - $resp.series_uuid
```
pongad: Didn't we want to test nested fields? Should this be `$resp.series_uuid.something`?

hzyi-google: `series_uuid` is already a message nested in `PublishSeriesResponse`? But I updated the yaml to print `$resp.series_uuid.series_bytes`

hzyi-google:
PTAL

PR #2331, opened 2018-09-21
[Proto annotations] Parse method signatures (flattenings)

andreamlin:
Combine flattenings from proto annotation `MethodSignature` and GAPIC config. 

Adding the condition that resource object methods must be generated (and public) if there is a flattening, according to the API config spec, "Code generators implementing overloads must provide them in addition to the implementation taking the request object outright (to satisfy the requirement that adding an overload is always non-breaking)." This results in the changes in java_library.baseline.

Including unit testing, since baseline testing isn't set up yet.


andreamlin:
PTAL

garrettjonesgoogle:
I think we should completely remove the `request_object_method` config feature because of the new proto annotation requirement. It should be a non-breaking change since there is no hierarchical relation between flattened types and the request types, so having the new method appear shouldn't create any clashes. This will simplify gapic-generator.

andreamlin:
Merged in blocker PR.

PTAL

garrettjonesgoogle:
@eoogbe would you be able to review this?

src/main/java/com/google/api/codegen/config/GapicInterfaceConfig.java
```diff
@@ -128,14 +132,19 @@ static GapicInterfaceConfig createInterfaceConfig(
               resourceNameConfigs,
               retryCodesConfig,
               retrySettingsDefinition.keySet());
-      methodConfigs = createMethodConfigs(methodConfigMap, interfaceConfigProto);
```
eoogbe: What keeps this from being nullable?

src/main/java/com/google/api/codegen/util/ProtoParser.java
```diff
@@ -81,6 +122,11 @@ public Retry getRetry(Method method) {
     return method.getDescriptor().getMethodAnnotation(AnnotationsProto.retry);
   }
 
+  /** Return the extra retry codes for the given method. */
+  public String getResourceType(Field field) {
+    return (String) field.getOptionFields().get(AnnotationsProto.resourceType.getDescriptor());
```
eoogbe: Converting null to empty here will make checks easier upstream.

src/test/java/com/google/api/codegen/config/ResourceNameMessageConfigsTest.java
```diff
@@ -221,4 +240,130 @@ public void testCreateResourceNameConfigs() {
 
     assertThat(diagCollector.getErrorCount()).isEqualTo(0);
   }
+
+  @Test
+  public void testCreateFlattenings() {
```
eoogbe: This method appears to be testing multiple things from multiple classes. Please split into multiple test case classes.

src/main/java/com/google/api/codegen/config/FlatteningConfig.java
```diff
@@ -122,6 +210,69 @@ static FlatteningConfig createFlattening(
...
+      DiagCollector diagCollector,
+      ResourceNameMessageConfigs messageConfigs,
+      ImmutableMap<String, ResourceNameConfig> resourceNameConfigs,
+      MethodSignature methodSignature,
+      ProtoMethodModel method,
+      ProtoParser protoParser) {
+
+    ImmutableMap.Builder<String, FieldConfig> flattenedFieldConfigBuilder = ImmutableMap.builder();
+    Set<String> oneofNames = new HashSet<>();
+    if (methodSignature == null || methodSignature.getFieldsCount() == 0) return null;
```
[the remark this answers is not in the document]
andreamlin: indeed

src/main/java/com/google/api/codegen/config/FlatteningConfig.java
```diff
@@ -34,16 +40,98 @@
...
+  @VisibleForTesting
+  @Nullable
+  static ImmutableList<FlatteningConfig> createFlatteningConfigs(
+      DiagCollector diagCollector,
+      ResourceNameMessageConfigs messageConfigs,
+      ImmutableMap<String, ResourceNameConfig> resourceNameConfigs,
+      MethodConfigProto methodConfigProto,
+      MethodModel methodModel,
+      ProtoParser protoParser) {
+    boolean missing = false;
```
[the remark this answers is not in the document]
andreamlin: oh neat method!

src/main/java/com/google/api/codegen/config/FlatteningConfig.java
```diff
@@ -122,6 +210,69 @@ static FlatteningConfig createFlattening(
         flattenedFieldConfigBuilder.build(), flatteningGroup.getFlatteningGroupName());
   }
 
+  /**
+   * Creates an instance of FlatteningConfig based on a FlatteningGroupProto, linking it up with the
+   * provided method.
+   */
+  @Nullable
+  private static FlatteningConfig createFlatteningFromProtoFile(
```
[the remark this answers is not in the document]
andreamlin: If this is about `createFlatteningFromProtoFile` and `createFlatteningsFromGapicConfig`, the methods are different enough that it's tough to abstract out any common functionality. Put in a TODO for refactoring later, once the dust settles on everything.

src/main/java/com/google/api/codegen/configgen/ProtoMethodTransformer.java
```diff
@@ -17,14 +17,18 @@
...
 import com.google.api.codegen.config.ProtoMethodModel;
 import com.google.api.tools.framework.model.Model;
+import java.util.List;
+import java.util.stream.Collectors;
 
 /** MethodTransformer implementation for proto Methods. */
 public class ProtoMethodTransformer implements MethodTransformer {
   private static final PagingParameters PAGING_PARAMETERS = new ProtoPagingParameters();
   private static final int MILLIS_PER_SECOND = 1000;
+  private static final int REQUEST_OBJECT_METHOD_THRESHOLD = 1;
```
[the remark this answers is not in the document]
andreamlin: Removed!

src/main/java/com/google/api/codegen/configgen/ProtoMethodTransformer.java
```diff
@@ -33,7 +37,7 @@ public boolean isIgnoredParameter(String parameter) {
 
   @Override
   public String getTimeoutMillis(MethodModel method) {
-    return String.valueOf(getTimeoutMillis((ProtoMethodModel) method));
+    return String.valueOf(getTimeoutMillisOrDefault((ProtoMethodModel) method));
   }
 
   public static long getTimeoutMillis(ProtoMethodModel method) {
```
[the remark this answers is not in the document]
andreamlin: done

src/main/java/com/google/api/codegen/config/GapicInterfaceConfig.java
andreamlin: I'm not sure I understand the question. 
`createMethodConfigs` will return a non-null list. The way that the parameters to that function are created ensures that `null` won't be put into the map.

src/main/java/com/google/api/codegen/util/ProtoParser.java
andreamlin: Hmm, that would hide null values behind empty strings, and potentially hide NPEs.

Also, I would prefer to have the calling code do the `Strings.isNullOrEmpty()` check. That way, if I forget to convert empty to null in one of the ProtoParser methods, the calling method is still fine.

src/test/java/com/google/api/codegen/config/ResourceNameMessageConfigsTest.java
andreamlin: The scope of the test is large and covers multiple classes. 

The main reason this test is in this test class is because it uses the the mocks of this class, and I'd rather keep this unit test method here than to copy over the mocks from this class into another test class.

src/main/java/com/google/api/codegen/util/ProtoParser.java
```diff
@@ -69,6 +76,40 @@ public OperationTypes getLongRunningOperation(Method method) {
     return method.getDescriptor().getMethodAnnotation(OperationsProto.operationTypes);
   }
 
+  /* Return a list of method signatures, aka flattenings, specified on a given method.
+   * This flattens the repeated additionalSignatures into the returned list of MethodSignatures. */
+  public List<MethodSignature> getMethodSignatures(ProtoMethodModel method) {
```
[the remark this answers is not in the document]
andreamlin: Done!

andreamlin:
Made changes.
PTAL

src/main/java/com/google/api/codegen/config/GapicInterfaceConfig.java
eoogbe: Ok didn't notice that at first. Thanks

src/main/java/com/google/api/codegen/util/ProtoParser.java
eoogbe: Good point

src/test/java/com/google/api/codegen/config/ResourceNameMessageConfigsTest.java
eoogbe: Ok that's reasonable. I think it might be nice to change the name of the test class, but probably not important in this PR.

src/main/java/com/google/api/codegen/config/GapicInterfaceConfig.java
```diff
@@ -216,45 +221,44 @@ private static SmokeTestConfig createSmokeTestConfig(
...
-      methodConfigMapBuilder.put(methodConfigProto.getName(), methodConfig);
+      // Re-insertion of the same method name doesn't affect the existing order
+      methodNames.add(methodConfigProto.getName());
     }
 
-    for (Method method : apiInterface.getMethods()) {
-      // TODO(andrealin): Reroute to grpc interface.
-      MethodConfigProto methodConfigProto = methodConfigProtoMap.get(method.getSimpleName());
+    for (String methodName : methodNames) {
+      // TODO(andrealin): Handle reroute to grpc interface.
```
eoogbe: I think this TODO is done now

PR #2334, opened 2018-09-24
Use showcase v0.0.7

landrito:


.circleci/config.yml
```diff
@@ -109,10 +109,10 @@ anchor_run_ruby: &anchor_run_ruby
   working_directory: /tmp/
   docker:
     - image: circleci/ruby:2.4-node
+    - image: gcr.io/gapic-showcase/gapic-showcase:0.0.6
```
andreamlin: Which commands use this image?

landrito: No commands use this but rather this spins up this image to run in the background. The way that these image configurations work in circleci is a bit confusing. The first listed image is the image that the subsequent steps are run in. The following images are images that are run in the background and are networked together. Essentially this is causing gapic-showcase to run in the background which causes the showcase integration tests to be able to make calls to it.

landrito:
Ah it seems that the ruby test command was lost somewhere in the mix.

.circleci/config.yml
andreamlin: oh neato

PR #2336, opened 2018-09-25
NodeJS: Add showcase integration tests

landrito:


landrito:
@fenster, there was an issue with client streaming methods that was caught by the showcase integration test. 😄 that showcase found it but 😞 that the error was present (for probably months).

alexander-fenster:
@landrito That's because we have no real-life `CLIENT_STREAMING` in any library (we do have `SERVER_STREAMING` and `BIDI_STREAMING` though). It's great that showcase found this!

PR #2338, opened 2018-09-25
Samplegen: PHP parameter handling

hzyi-google:
Make PHP standalone samples accept CLI arguments.

### DIFF
--- a/src/main/resources/com/google/api/codegen/php/standalone_sample.snip
+++ b/src/main/resources/com/google/api/codegen/php/standalone_sample.snip
@@ -43,27 +43,32 @@
 
 
     @end
-    @if sampleFile.extraInfo.hasMissingDefaultOptions
-        $options = [
-            @if sampleFile.extraInfo.missingDefaultServiceAddress
-                'serviceAddress' => 'my-service-address',
-            @end
-            @if sampleFile.extraInfo.missingDefaultServiceScopes
-                'credentialsConfig' => [
-                    'scopes' => ['my-service-scope'],
-                ],
-            @end
-            ];
-        ${@apiMethod.apiVariableName} = new {@apiMethod.apiClassName}($options);
+    {@apiMethod.name}
+    {@sample.initCode.argDefaultParams}
+    {@sample.initCode.argDefaultLines}
+    function sample_{@apiMethod.name}({@formalArgs(sample.initCode.argDefaultParams)})
+    {
+        @if sampleFile.extraInfo.hasMissingDefaultOptions
+            $options = [
+                @if sampleFile.extraInfo.missingDefaultServiceAddress
+                    'serviceAddress' => 'my-service-address',
+                @end
+                @if sampleFile.extraInfo.missingDefaultServiceScopes
+                    'credentialsConfig' => [
+                        'scopes' => ['my-service-scope'],
+                    ],
+                @end
+                ];
+            ${@apiMethod.apiVariableName} = new {@apiMethod.apiClassName}($options);
 
 
-    @else
-        ${@apiMethod.apiVariableName} = new {@apiMethod.apiClassName}();
+        @else
+            ${@apiMethod.apiVariableName} = new {@apiMethod.apiClassName}();
 
 
-    @end
-    @if sample.initCode.lines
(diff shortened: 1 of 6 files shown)


src/test/java/com/google/api/codegen/gapic/testdata/php/php_library.baseline
```diff
@@ -3144,27 +3216,42 @@ use \Google\Example\Library\V1\Book;
...
 }
+
+$opts = [
+    'name::',
+    'edition::',
+];
+
+$options = getopt('', $opts);
+
+$name == $options['name'] ?? 'Math';
```
beccasaurus: Awesome 👍

src/main/resources/com/google/api/codegen/php/standalone_sample.snip
```diff
@@ -43,64 +43,95 @@
...
 
+    {@optsArray(sample.initCode.argDefaultParams)}
 
-    @else
-        ${@apiMethod.apiVariableName} = new {@apiMethod.apiClassName}();
+    $options = getopt('', $opts);
+
+    @if sample.initCode.argDefaultParams
+        @join param : sample.initCode.argDefaultParams
+            {@variable(param.identifier)} == $options['{@param.identifier}'] ?? {@renderInitValue(param.initValue, param.identifier)};
```
michaelbausor: `??` is not available in PHP 5.*, which we want to support :-(

You may need to use the much wordier:
```
isset($options['{@param.identifier}']) ? $options['{@param.identifier}'] : {@renderInitValue(param.initValue, param.identifier)}
```

Also, should this be `=` instead of `==`?

michaelbausor: Following up on the structure here - a pattern than we use elsewhere for setting default arguments is:
- Create a `$defaultOptions` array with all defaults set
- Create the `$options` array using whatever method (in this case getopts)
- `$options += $defaultOptions`

That may work here and avoids the `isset...` construction. It is very sad that we don't have the `??` operator.

src/test/java/com/google/api/codegen/gapic/testdata/php/php_library.baseline
```diff
@@ -2747,33 +2747,42 @@ return [
...
+            $responseItem = $stream->read();
+        }
+    } finally {
+        $libraryServiceClient->close();
     }
-} finally {
-    $libraryServiceClient->close();
 }
+
+$opts = [];
```
michaelbausor: Maybe when there are no args, we can skip the whole process of calling getopt?

src/main/resources/com/google/api/codegen/php/standalone_sample.snip
beccasaurus: **^--- Note:** from the _Developer Experience_ perspective: I have no qualms with taking a wordier (more idiomatic & correct) approach 🙂

The CLI parsing will generally not be shown to developers, ie. it's usually not included when we embed samples in the website. Sometimes it may be, so...

So long as the generated CLI parsing & main() invoking code is _**idiomatic**_ to the language & _looks familiar_ to anyone who's used this option parser – I'm +1 from my side!

hzyi-google:
PTAL @michaelbausor

src/test/java/com/google/api/codegen/gapic/testdata/php/php_library.baseline
```diff
@@ -3144,27 +3184,48 @@ use \Google\Example\Library\V1\Book;
...
+    'edition' => 123,
+];
+
+$options = getopt('', $opts);
+$options += $defaultOptions;
+
+$name = $options['name'];
+$edition = $options['edition'];
+
+samplePublishSeries($name, $edition);
```
beccasaurus: Purty 😍

beccasaurus: <3 the other samples w/o the empty options – so clean! simple function() calls – love it

hzyi-google: :smile: Thanks!

PR #2339, opened 2018-09-25
Set up test once in static @BeforeClass

andreamlin:
Setting up the test resources once decreases the run time of this test from ~800ms to ~50ms on Intellij.


andreamlin:
PTAL

PR #2340, opened 2018-09-26
Deprecate request_object_method from config

andreamlin:
By default, we will generate a public request object method.


garrettjonesgoogle:
@lukesneeringer FYI

PR #2342, opened 2018-09-26
php standalone samples: remove boilerplate

hzyi-google:


hzyi-google:
PTAL

src/main/resources/com/google/api/codegen/php/standalone_sample.snip
```diff
@@ -1,50 +1,31 @@
...
-
-        /*
-        {@standaloneSample(sampleFile, apiMethod, sample)}
-        */
-        // [END {@sample.regionTag}_core]
-
-        // FIXME: Insert here clean-up code.
-             
-        // [END {@sample.regionTag}]
+            {@standaloneSample(sampleFile, apiMethod, sample)}
```
michaelbausor: Since this contains just a single function call, we can probably get rid of the let statements as well.

src/main/resources/com/google/api/codegen/php/standalone_sample.snip
```diff
@@ -440,3 +424,9 @@
         ->{@field}()
     @end
 @end
+
+@private generatedSampleWarning(sample)
```
michaelbausor: Is this used in just one place? If so, lets inline it.

pongad:
LGTM :)

PR #2343, opened 2018-09-26
Python: Add showcase integration tests

landrito:
The integration tests are expected to be failing currently. They should be passing when @lukesneeringer cuts a new release of googleapis-common-protos that includes the annotations in longrunning_pb2.py.


.circleci/config.yml
```diff
@@ -77,16 +78,34 @@ anchor_verify_and_test_python: &anchor_verify_and_test_python
...
+              --noxfile /tmp/workspace/gapic-generator/showcase/python/nox.py
+        when: always
+    - run:
+        name: Run Showcase Integration Tests.
+        command: |
+          nox --session "showcase(py='2.7')" \
+              --noxfile /tmp/workspace/gapic-generator/showcase/python/nox.py
+          nox --session "showcase(py='3.5')" \
+              --noxfile /tmp/workspace/gapic-generator/showcase/python/nox.py
+          nox --session "showcase(py='3.6')" \
```
lukesneeringer: You can restrict showcase to 2.7 and 3.7.

showcase/python/tests/showcase_integration_test.py
```diff
@@ -0,0 +1,75 @@
...
+    transport = echo_grpc_transport.EchoGrpcTransport(channel=channel)
+    client = showcase_v1alpha2.EchoClient(transport=transport)
+
+    def test_echo(self):
+        content = 'hello world'
+        response = self.client.echo(content=content)
+        assert content == response.content
+
+    def test_echo_exception(self):
+        with pytest.raises(Exception):
```
lukesneeringer: Can this catch the correct exception?

.circleci/config.yml
landrito: Done!

showcase/python/tests/showcase_integration_test.py
landrito: Done!

.circleci/config.yml
landrito: I think we can only use 3.6 because the latest image of nox is not compatible some stuff we are using in circle and that image only has python 3.6.

PR #2344, opened 2018-09-27
[Proto Annotations] Add baseline files

andreamlin:
Copy over Java, PHP, Ruby, and NodeJS baselines for the library API clients. These are exact copies of the files located in `src/test/java/com/google/api/codegen/gapic/testdata/`. 

Once these are checked in, we can start doing baseline testing.


andreamlin:
PTAL

PR #2349, opened 2018-09-28
Java: add showcase integration tests

landrito:


.circleci/config.yml
```diff
@@ -434,6 +471,16 @@ jobs:
...
+      - run:
+          name: Install the generated showcase library
+          command: |
+            cd workspace/gapic-generator/artman-genfiles/java
+            ./gradlew install -p gapic-google-cloud-showcase-v1alpha2
+      - run:
+          name: Run Showcase Integration Tests
+          command: |
+            cd workspace/gapic-generator/showcase/java
+            ./gradlew build test
```
jbolinger: `./gradlew cleanTest check` might be better just in case... I'm not sure if the tests results would ever get cached in CI, but it can't hurt to clean it. Check would also verify the test code is formatted.

showcase/java/src/test/java/com/google/api/showcase/ShowcaseTransportChannelProvider.java
```diff
@@ -0,0 +1,110 @@
...
+  public TransportChannelProvider withPoolSize(int size) {
+    return this;
+  }
+
+  @Override
+  public TransportChannel getTransportChannel() {
+    GrpcHeaderInterceptor headerInterceptor =
+        new GrpcHeaderInterceptor(headerProvider.getHeaders());
+
+    // not exposed - can't test
```
jbolinger: I left this here so that I wouldn't forget to bring it up, but we should probably remove this comment for now.

This metadata handling is a feature, and we might want to consider if it should be part of showcase. We can address it later, and if we do add it I think we'll have to make a few other changes to enable testing on an insecure channel.

landrito: Ah gotcha. I'll remove.

.circleci/config.yml
landrito: Done!

PR #2350, opened 2018-09-28
Add port 7469 to showcase.yaml

andreamlin:
Currently the service address is pulled from the `name` field in each service config. Set the service address to include the port, and parse the port from the service address.


andreamlin:
PTAL

PR #2352, opened 2018-10-02
allow indexing into repeated fields

pongad:


src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -295,15 +296,32 @@ public void run() {
...
 
         type = field.getType();
-        accessors.add(context.getNamer().getFieldGetFunctionName(field));
+        accessors.add(
+            AccessorView.MemberView.newBuilder()
+                .member(context.getNamer().getFieldGetFunctionName(field))
+                .build());
       } else if (token == '[') {
         Preconditions.checkArgument(
             type.isRepeated() && !type.isMap(),
```
hzyi-google: Would type be a map in any case? I went though the design doc but did not find an example like that. Let's add a todo if we need to handle map types in the future.

pongad: Ah  yes done.

src/main/java/com/google/api/codegen/viewmodel/AccessorView.java
```diff
@@ -0,0 +1,66 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen.viewmodel;
+
+import com.google.auto.value.AutoValue;
+
+public interface AccessorView {
+  enum Kind {
+    MEMBER,
```
vchudnov-g: nit, feel free to ignore: `FIELD` would be clearer and more specific IMO

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -295,15 +296,34 @@ public void run() {
...
+
+        accessors.add(AccessorView.IndexView.newBuilder().index(config.tokenStr()).build());
+
+        Preconditions.checkArgument(
+            config.scan() == ']',
+            "%s:%s: expected ']': %s",
+            context.getMethodModel().getSimpleName(),
+            valueSet.getId(),
+            config.input());
+      } else if (token == '{') {
```
vchudnov-g: since `.index` is a string, extending to maps should be pretty easy over here, correct?
(when we get to it, I suspect we'll want to s/`index`/`key`/ for generality)

src/main/resources/com/google/api/codegen/java/standalone_sample.snip
```diff
@@ -284,7 +284,12 @@
 
 @private accessor(accessors)
   @join field : accessors on ""
-    .{@field}()
+    @switch field.kind
+    @case "MEMBER"
+      .{@field.member}()
+    @case "INDEX"
```
vchudnov-g: Maybe add a comment in the template acknowledging that this will throw if `field.index` is out of bounds, but we explicitly don't check.

In the other langs as well.

src/main/resources/com/google/api/codegen/java/standalone_sample.snip
```diff
@@ -284,7 +284,12 @@
 
 @private accessor(accessors)
   @join field : accessors on ""
-    .{@field}()
+    @switch field.kind
```
vchudnov-g: nit: Above I would s/`concatenatedArgs`/`commaSeparatedFields`/g

In the other langs as well.

src/main/java/com/google/api/codegen/viewmodel/AccessorView.java
pongad: Done.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
pongad: Yup it should be. I know of languages that treat array lookup and map lookup differently though (Perl comes to mind), I think we should treat them differently at this stage.

src/main/resources/com/google/api/codegen/java/standalone_sample.snip
pongad: Done

src/main/resources/com/google/api/codegen/java/standalone_sample.snip
pongad: Done

pongad:
@vchudnov-g PTAL

PR #2353, opened 2018-10-04
use artman v0.16.2 in CI

alexander-fenster:
Bumping artman version in CircleCI config to v0.15.9.


andreamlin:
We can bump this further?

alexander-fenster:
We can!

andreamlin:
LGTM when all tests (except for the long artman_smoke_tests job, which isn't tested here b/c it uses the latest version of artman) pass.

PR #2357, opened 2018-10-08
Order resource names

andreamlin:
Enforce an ordering on the map/list of resource name configs used by the transformers. This will be useful later on when resource names from protofile annotations are introduced, and we want to only compare baseline changes that result from that. This doesn't result in any semantic changes in the resulting code.


src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
```diff
@@ -1321,7 +1321,7 @@ namespace Google.Example.Library.V1.Snippets
             // Create client
             LibraryServiceClient libraryServiceClient = await LibraryServiceClient.CreateAsync();
             // Initialize request argument(s)
-            IResourceName name = new ShelfName("[SHELF_ID]");
+            IResourceName name = new ArchivedBookName("[ARCHIVE_PATH]", "[BOOK_ID]");
```
andreamlin: library_gapic.yaml specifies the `ListStringsResponse` to be an `AnyResourceNameConfig` via the `*` value, so it's ok that the instantiating type for the `name` parameter changes.

src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
```diff
@@ -12167,91 +12167,6 @@ using linq = System.Linq;
 
 namespace Google.Example.Library.V1
 {
-    /// <summary>
```
andreamlin: This `class ShelfName` declaration is merely declared later in the class.

src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
```diff
@@ -12436,105 +12351,6 @@ namespace Google.Example.Library.V1
         public static bool operator !=(BookName a, BookName b) => !(a == b);
     }
 
-    /// <summary>
-    /// Resource name for the 'return' resource.
-    /// </summary>
-    public sealed partial class ReturnName : gax::IResourceName, sys::IEquatable<ReturnName>
```
andreamlin: This `class ReturnName` is also declared later in the class.

src/test/java/com/google/api/codegen/gapic/testdata/java/java_library.baseline
```diff
@@ -1637,7 +1637,7 @@ public class LibraryClient implements BackgroundResource {
    * Sample code:
    * <pre><code>
    * try (LibraryClient libraryClient = LibraryClient.create()) {
-   *   ResourceName name = ShelfName.of("[SHELF_ID]");
+   *   ResourceName name = ArchivedBookName.of("[ARCHIVE_PATH]", "[BOOK_ID]");
```
andreamlin: Here too, `ListStringsResponse.name` is listed as an AnyResourceNameConfig in the library_gapic.yaml, so this change is semantically alright.

andreamlin:
PTAL

PR #2359, opened 2018-10-08
Samplegen: print python enums and php messages

hzyi-google:


hzyi-google:
PTAL

michaelbausor:
PHP changes LGTM

hzyi-google:
cc/ @beccasaurus somehow I can't tag you as an assignee or reviewer

src/main/resources/com/google/api/codegen/py/standalone_sample.snip
```diff
@@ -259,7 +259,11 @@
 
 @private commaSeparatedFields(args)
   @join arg : args on ", "
-    {@variableField(arg)}
+    @if arg.type.isEnum
```
pongad: Can you check if this compiles? I *think* we need to import the `enums` package right?

src/main/resources/com/google/api/codegen/php/standalone_sample.snip
```diff
@@ -405,7 +405,11 @@
 
 @private commaSeparatedFields(args)
     @join arg : args on ", "
-        {@variableField(arg)}
+        @if or(arg.type.isMessage, arg.type.isEnum)
+            print_r({@variableField(arg)}), true)
```
vchudnov-g: I think you may have an extra right paren here....

src/test/java/com/google/api/codegen/gapic/testdata/py/python_library.baseline
```diff
@@ -4360,7 +4360,7 @@ def sample_monolog_about_book():
 
   requests = [request]
   response = client.monolog_about_book(requests)
-  print(response)
+  print('The stage of the comment is: {}'.format(enums.EnumName(response.stage).name))
```
beccasaurus: _No way! Wow, that was a quick PR!_

**Awesome!** 😎

src/test/java/com/google/api/codegen/gapic/testdata/php/php_library.baseline
```diff
@@ -3103,7 +3103,7 @@ function sampleMonologAboutBook()
             $stream->write($request);
         }
         $response = $stream->readResponse();
-        printf("%s" . PHP_EOL, $response);
+        printf("The stage of the comment is: %s" . PHP_EOL, print_r($response->getStage()), true));
```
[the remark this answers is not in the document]
michaelbausor: Enums are represented as ints, see e.g. here: 

So we could just print them directly. I followup problem is that users are just going to get an int :-(

This PR was opened months ago to add string-to-int conversions and vice-versa https://github.com/protocolbuffers/protobuf/pull/4534 - seems like reviving it might be the best way to print something more useful.

src/main/resources/com/google/api/codegen/php/standalone_sample.snip
hzyi-google: Yes! Good catch

src/test/java/com/google/api/codegen/gapic/testdata/php/php_library.baseline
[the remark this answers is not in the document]
hzyi-google: Awesome. Would it be OK to just print enums 'normally' for now and let's update the snippets when the pending PR is finished?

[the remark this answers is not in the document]
vchudnov-g: @hzyi-google You're suggesting we print the ints for now, and then use the new protobuf functionality when https://github.com/protocolbuffers/protobuf/pull/4534 is merged?

If we see activity on that PR, then that's fine; let's not reinvent the wheel. If that doesn't seem to be moving along, then we might need to implement the functionality ourselves in the meantime.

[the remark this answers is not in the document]
beccasaurus: > But I guess in both cases we need to make efforts outside gapic-generator in order to print something meaningful

I'd love to track this ^ to make sure we follow-up 😄 

Up until recently, Python was "solving" this problem by creating a List of the enum display fields (from the proto) and printing via `list_of_display_names[response.enum_field]`. If the data model object snippets have access to includes a full representation of the proto, it's _technically_ possible for us to do the same thing (until https://github.com/protocolbuffers/protobuf/pull/4534 is merged). I'd love to see this help push the PR thru instead tho

src/main/resources/com/google/api/codegen/py/standalone_sample.snip
hzyi-google: Added.

hzyi-google:
PTAL @vchudnov-g @pongad

pongad:
I think this is fine, but Victor should sign off.

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
```diff
@@ -243,12 +250,19 @@ public void generateSamples(
           outputs = OutputTransformer.defaultOutputSpecs(methodContext.getMethodModel());
         }
 
+        ImmutableList<OutputView> outputView =
```
vchudnov-g: nit: pluralize  name

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -35,11 +36,33 @@
...
+
+  public OutputTransformer() {
+    this.importTransformer = new OutputImportTransformer() {};
+  }
+
+  public OutputTransformer(OutputImportTransformer importTransformer) {
+    this.importTransformer = importTransformer;
+  }
+
+  public static interface OutputImportTransformer {
```
vchudnov-g: Add doc to this interface. In particular, clarify that this computes any imports necessary for the sample output to execute correctly, and which must be passed to the `MethodSampleView`

src/main/resources/com/google/api/codegen/php/standalone_sample.snip
```diff
@@ -405,7 +405,14 @@
 
 @private commaSeparatedFields(args)
     @join arg : args on ", "
-        {@variableField(arg)}
+        # FIXME(hzyi): print enums directly will just get us an int
```
vchudnov-g: Thanks for adding this FIXME!

PR #2361, opened 2018-10-10
set up for sample resource name testing

pongad:
This commit just adds the field so we can test later. No change in
behavior.

### DIFF
--- a/src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
+++ b/src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
@@ -682,8 +682,8 @@ namespace Google.Example.Library.V1.Snippets
         /// <summary>Snippet for PublishSeriesAsync</summary>
         public async Task PublishSeriesAsync()
         {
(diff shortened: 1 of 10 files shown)


src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
```diff
@@ -7291,6 +7302,9 @@ namespace Google.Example.Library.V1
         /// <param name="seriesUuid">
         /// Uniquely identifies the series to the publishing house.
         /// </param>
+        /// <param name="parent">
+        ///
```
vchudnov-g: There should be some content in here, yes?

src/test/java/com/google/api/codegen/testsrc/libraryproto/library.proto
```diff
@@ -423,6 +423,10 @@ message PublishSeriesRequest {
 
   // Uniquely identifies the series to the publishing house.
   SeriesUuid series_uuid = 5;
+
+  // TODO(pongad): Come up with a reason for this path to be here.
```
vchudnov-g: Yes, please do. This does not provide a clear mental model. :(

Maybe the `parent` in an `AddCommentsRequest` could be a `Book`?

PR #2363, opened 2018-10-11
Minimize semantic differences between library.proto and library_gapic.yaml

andreamlin:
This is so we can later test using only the library.proto (annotated) to generate a client, and it should have a similar surface to the client generated using the library_gapic.yaml.

### DIFF
--- a/src/test/java/com/google/api/codegen/protoannotations/testdata/library.proto
+++ b/src/test/java/com/google/api/codegen/protoannotations/testdata/library.proto
@@ -47,7 +47,7 @@ option (google.api.metadata) = {
 // Service comment may include special characters: <>&"`'@.
 //
 service LibraryService {
-  option (google.api.default_host) = "library-example.googleapis.com";
+  option (google.api.default_host) = "library-example.googleapis.com:1234";
   option (google.api.oauth) = {
     scopes: ["https://www.googleapis.com/auth/library",
       "https://www.googleapis.com/auth/cloud-platform"]
@@ -67,15 +67,19 @@ service LibraryService {
     option (google.api.http) = { get: "/v1/{name=bookShelves/*}" };
     option (google.api.method_signature) = {
       fields: ["name"]
-      additional_signatures: [{
-        fields: ["name", "message"]
-      }]
+      additional_signatures: [
+        {fields: ["name", "message"]},
+        {fields: ["name", "message", "string_builder"]}
+      ]
     };
   }
 
   // Lists shelves.
   rpc ListShelves(ListShelvesRequest) returns (ListShelvesResponse) {
     option (google.api.http) = { get: "/v1/bookShelves" };
+    option (google.api.method_signature) = {
+      fields: []
+    };
   }
 
   // Deletes a shelf.
@@ -111,7 +115,7 @@ service LibraryService {
   rpc PublishSeries(PublishSeriesRequest) returns (PublishSeriesResponse) {
     option (google.api.http) = { post: "/v1:publish" body: "*" };
     option (google.api.method_signature) = {
-      fields: ["shelf", "books"]
+      fields: ["shelf", "books", "edition", "series_uuid"]
     };
   }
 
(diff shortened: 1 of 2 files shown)


andreamlin:
PTAL

src/test/java/com/google/api/codegen/protoannotations/testdata/library.proto
```diff
@@ -47,7 +47,7 @@ option (google.api.metadata) = {
 // Service comment may include special characters: <>&"`'@.
 //
 service LibraryService {
-  option (google.api.default_host) = "library-example.googleapis.com";
+  option (google.api.default_host) = "library-example.googleapis.com:1234";
```
andreamlin: the service yaml at library.yaml has this as the service address

src/test/java/com/google/api/codegen/protoannotations/testdata/library.proto
```diff
@@ -427,8 +519,7 @@ message Shelf {
 message CreateShelfRequest {
   // The shelf to create.
   Shelf shelf = 1 [
-    (google.api.required) = true,
-    (google.api.resource_type) = "google.example.library.v1.Shelf"];
```
pongad: There's probably something I don't understand. Does the current GAPIC YAML not use resource names for these?

src/test/java/com/google/api/codegen/util/ProtoParserTest.java
```diff
@@ -132,33 +144,34 @@ public void testIsHttpGetMethod() {
...
   public void testGetRequiredFields() {
     Method publishSeriesMethod = libraryService.lookupMethod("PublishSeries");
     List<String> requiredFields = protoParser.getRequiredFields(publishSeriesMethod);
     Collections.sort(requiredFields);
-    assertThat(requiredFields.size()).isEqualTo(2);
+    assertThat(requiredFields.size()).isEqualTo(3);
     assertThat(requiredFields.get(0)).isEqualTo("books");
-    assertThat(requiredFields.get(1)).isEqualTo("shelf");
+    assertThat(requiredFields.get(1)).isEqualTo("series_uuid");
+    assertThat(requiredFields.get(2)).isEqualTo("shelf");
```
pongad: I think you can leave `requiredFields` unsorted then `assertThat(requiredFields).containsExactly("books", "series_uuid", "shelf");`

src/test/java/com/google/api/codegen/util/ProtoParserTest.java
```diff
@@ -186,6 +199,12 @@ public void testGetMethodSignatures() {
     assertThat(additionalSignature.getFieldsList().size()).isEqualTo(2);
     assertThat(additionalSignature.getFieldsList().get(0)).isEqualTo("name");
     assertThat(additionalSignature.getFieldsList().get(1)).isEqualTo("message");
+
+    MethodSignature additionalSignature2 = getShelfFlattenings.get(2);
+    assertThat(additionalSignature2.getFieldsList().size()).isEqualTo(3);
+    assertThat(additionalSignature2.getFieldsList().get(0)).isEqualTo("name");
+    assertThat(additionalSignature2.getFieldsList().get(1)).isEqualTo("message");
+    assertThat(additionalSignature2.getFieldsList().get(2)).isEqualTo("string_builder");
```
pongad: ditto

src/test/java/com/google/api/codegen/protoannotations/testdata/library.proto
andreamlin: The `resource_type` annotation only makes sense for `string` type fields. This is a weird circular reference, where the `Shelf shelf` refers to a `resource_type = "google.example.library.v1.Shelf"`, which doesn't provide any new information. I had previously put this annotation in to match what the annotated pubsub.proto used to look like, but those changes were recently reverted in https://github.com/googleapis/googleapis/pull/520

src/test/java/com/google/api/codegen/util/ProtoParserTest.java
andreamlin: neat

src/test/java/com/google/api/codegen/util/ProtoParserTest.java
andreamlin: done

andreamlin:
PTAL
fixed tests

src/test/java/com/google/api/codegen/protoannotations/testdata/library.proto
pongad: Oh right I get it now.

PR #2364, opened 2018-10-11
Cleanup

andreamlin:


src/main/java/com/google/api/codegen/config/GapicInterfaceConfig.java
```diff
@@ -162,7 +162,7 @@ static GapicInterfaceConfig createInterfaceConfig(
     for (CollectionConfigProto collectionConfigProto : interfaceConfigProto.getCollectionsList()) {
       String entityName = collectionConfigProto.getEntityName();
       ResourceNameConfig resourceName = resourceNameConfigs.get(entityName);
-      if (resourceName == null || !(resourceName instanceof SingleResourceNameConfig)) {
```
andreamlin: intellij said this was redundant

src/main/java/com/google/api/codegen/config/ProtoMethodModel.java
```diff
@@ -208,11 +207,6 @@ public String getScopedDescription() {
     return outputFields;
   }
 
-  @Override
-  public List<ProtoField> getResourceNameInputFields() {
```
andreamlin: this was never called

andreamlin:
PTAL

src/main/java/com/google/api/codegen/config/ProtoField.java
```diff
@@ -41,7 +42,11 @@
   public ProtoField(Field protoField) {
     Preconditions.checkNotNull(protoField);
     this.protoField = protoField;
-    this.protoTypeRef = ProtoTypeRef.create(protoField.getType());
+    TypeRef typeRef = protoField.getType();
+    if (typeRef == null) {
```
michaelbausor: What case is this new condition handling? Could (should) protoField.getType() ever by null?

src/main/java/com/google/api/codegen/util/ProtoParser.java
```diff
@@ -72,8 +73,19 @@ public OperationTypes getLongRunningOperation(Method method) {
...
+            .getOptionFields()
+            .keySet()
+            .stream()
+            .filter(o -> o.getFullName().equals("google.api.method_signature"))
+            .findAny();
+    if (!signatureDescriptor.isPresent()) {
+      return ImmutableList.of();
+    }
     MethodSignature methodSignature =
-        method.getDescriptor().getMethodAnnotation(AnnotationsProto.methodSignature);
```
michaelbausor: Does this NPE, or return null? If it returns null, it seems simpler than using streams, and doesn't hard-code the "google.api.method_signature" string.

src/main/java/com/google/api/codegen/config/SingleResourceNameConfig.java
```diff
@@ -83,7 +83,7 @@ public static SingleResourceNameConfig createSingleResourceName(
   // e.g. createPathTemplate("bookShelves/*/books/{book}") returns
   // PathTemplate.create("bookShelves/{bookShelf}/books/{book}")
   @VisibleForTesting
-  static String escapePathTemplate(String template) {
+  public static String escapePathTemplate(String template) {
```
michaelbausor: Why does this need to be public?

src/main/java/com/google/api/codegen/config/ProtoField.java
andreamlin: Hmm I thought that this was necessary some time ago, but I just ran tests without this change and it's fine. Reverted.

src/main/java/com/google/api/codegen/util/ProtoParser.java
andreamlin: The problem with the previous implementation is that `MethodSignature methodSignature` would always return a non-null object, which makes it impossible to differentiate between an empty method signature (with 0 parameters) and the non-presence of a method signature. 

We need to do the stream thing to see if the method_signature annotation is even present.

Added a code comment to explain.

src/main/java/com/google/api/codegen/config/SingleResourceNameConfig.java
andreamlin: Removed.

andreamlin:
PTAL

src/main/java/com/google/api/codegen/util/ProtoParser.java
michaelbausor: Hmm, it is a bit annoying that we can't access the underlying MethodDescriptorProto, as then we could call getOptions().hasExtension(...), similar to how the implementation of getMethodAnnotation calls getOptions().getExtension(...).

Does getOptionsFields().contains(AnnotationsProto.methodSignature.defaultInstance()) work?

andreamlin: Oh that does work! Thanks! It's a lot cleaner.
Also added a test for empty method signatures.

andreamlin:
PTAL

src/test/java/com/google/api/codegen/util/ProtoParserTest.java
```diff
@@ -201,6 +201,24 @@ public void testGetMethodSignatures() {
         .containsExactly("name", "message", "string_builder");
   }
 
+  @Test
+  public void testEmptySignature() {
+    // Test that we can detect empty method signatures.
```
michaelbausor: Can we also test the case when there is no method signature? (or is that already tested?)

andreamlin: Sure! Done.

PR #2365, opened 2018-10-12
[Proto annotations]: For LRO, use gapic config retry settings if they exist

andreamlin:
The LRO proto annotation has only two parameters: return_type and metadata_type.
The LRO GAPIC config contains the above two parameters plus a bunch of retry params. 

If the LRO proto annotation is the same as the LRO gapic config type params, then use the GAPIC config retry params.

Once we start annotating live API protos, this allows the surface for LRO client methods to stay the same.

### DIFF
--- a/src/test/java/com/google/api/codegen/config/LongRunningConfigTest.java
+++ b/src/test/java/com/google/api/codegen/config/LongRunningConfigTest.java
@@ -35,12 +35,12 @@ public class LongRunningConfigTest {
   private static final String GAPIC_CONFIG_METADATA_TYPE = "HeaderType";
   private static final String ANNOTATIONS_RETURN_TYPE_NAME = "BookType";
   private static final String ANNOTATIONS_METADATA_TYPE = "FooterType";
-  private static boolean TEST_IMPLEMENTS_DELETE = false;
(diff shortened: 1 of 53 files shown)


andreamlin:
PTAL

src/test/java/com/google/api/codegen/config/LongRunningConfigTest.java
```diff
@@ -35,8 +35,8 @@
   private static final String GAPIC_CONFIG_METADATA_TYPE = "HeaderType";
   private static final String ANNOTATIONS_RETURN_TYPE_NAME = "BookType";
   private static final String ANNOTATIONS_METADATA_TYPE = "FooterType";
-  private static final boolean TEST_IMPLEMENTS_DELETE = false;
-  private static final boolean TEST_IMPLEMENTS_CANCEL = false;
+  private static boolean TEST_IMPLEMENTS_DELETE = false;
```
garrettjonesgoogle: Why is the `final` being removed here?

andreamlin: Oops. Making all those private static primitives final now.

andreamlin:
PTAL

Issue #2366, opened 2018-10-12
honor `sample_argument_name` config


pongad:
Currently we still let the init code logic choose a name.

PR #2367, opened 2018-10-12
allow user to specify sample argument name

pongad:
We don't honor this config yet. This is tracked by #2366.


src/test/java/com/google/api/codegen/testsrc/libraryproto/library_gapic.yaml
```diff
@@ -240,9 +240,9 @@ interfaces:
         - edition=123
         attributes:
         - parameter: shelf.name
-          sample_argument: true
+          sample_argument_name: name
```
vchudnov-g: I think we could s/`sample_argument_name`/`sample_argument`/ , iejust changing the type of the field

PR #2380, opened 2018-10-15
Remove additional retry codes, added due to the change in retry when …

chrisdunelm:
…using proto annotations

(Yes, this is a bit of a hack, but this code won't be used for much longer)


src/main/java/com/google/api/codegen/transformer/csharp/CSharpGapicClientTransformer.java
```diff
@@ -377,6 +388,7 @@ private StaticLangSettingsView generateSettingsClass(GapicInterfaceContext conte
               context.asRequestMethodContext(method));
       settingsMembers.addAll(calls);
     }
+    // settingsMembers.stream().filter(s -> s.memberName() != )
```
jskeet: Remove comment?

chrisdunelm: drat

PR #2381, opened 2018-10-15
Don't generate methods that aren't present in GAPIC config

andreamlin:
Also don't generate retry codes for methods that aren't defined in the GAPIC config.

Add a method to the test library.proto but not to its test GAPIC config to test that it won't be present in the generated surface.


andreamlin:
PTAL

src/main/java/com/google/api/codegen/config/GapicInterfaceConfig.java
```diff
@@ -223,15 +223,15 @@ private static SmokeTestConfig createSmokeTestConfig(
       ProtoParser protoParser) {
     Map<String, GapicMethodConfig> methodConfigMapBuilder = new TreeMap<>();
 
-    // The order in which to create GapicMethodConfigs; first in the order of methods listed in
-    // the protofile, and then any remaining methods in the Gapic config.
+    // The order in which to create GapicMethodConfigs; only use .
```
eoogbe: incomplete sentence

PR #2392, opened 2018-10-29
Samplegen: print python enums

### LINKED ISSUE #2391: Use actual enum name instead of EnumName
In the Python code (eg [here](https://github.com/googleapis/gapic-generator/blob/b4c1fb4c4e24f9cbb32c9c91894c7de17d6ca47b/src/test/java/com/google/api/codegen/gapic/testdata/py/python_library.baseline#L4379)), the `EnumName` in `enums.EnumName` should be the name of the enum defined in the `enums.py` file.

(We talked about this earlier, but I'm filing this issue so we don't lose track of it.)

hzyi-google:
Fix #2391.


hzyi-google:
PTAL @pongad @beccasaurus @vchudnov-g

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -42,13 +43,17 @@
       ImmutableSet.<String>of("response", "response_item");
 
   private final OutputImportTransformer importTransformer;
+  private final PrintArgTransformer printArgTransformer;
```
hzyi-google: I feel it's a bit unnecessary to have two nested transformer both of which only have one public method. Instead, we can probably move these two methods (`generateOutputImports` and `generatePrintArg`) to `OutputTransformer`, and have each language subclass `OutputTransformer`. Thoughts? @pongad @vchudnov-g

src/main/java/com/google/api/codegen/transformer/py/PythonSampleOutputImportTransformer.java
```diff
@@ -60,21 +61,24 @@ private static void addEnumImports(
...
-                .build();
-        imports.add(
-            ImportFileView.newBuilder()
-                .moduleName(context.getNamer().getVersionedDirectoryNamespace())
-                .types(Collections.singletonList(importTypeView))
-                .build());
-      }
+    for (PrintArgView arg : view.args()) {
+      for (PrintArgView.ArgSegmentView segment : arg.segments())
+        if (segment instanceof PrintArgView.VariableSegmentView) {
```
pongad: FYI: I think stream can be appropriate here. Something like this might work.

```java
boolean addEnumImport = view
  .args()
  .stream()
  .flatMap(arg -> arg.segments().stream())
  .filter(seg -> seg instanceof PrintArgView.VariableSegmentView)
  .map(seg -> ((PrintArgView.VariableSegmentView) seg).variable().type())
  .anyMatch(type -> type != null && type.isEnum());
if (addEnumImport) {
  imports.add(...)
}
```

src/main/java/com/google/api/codegen/transformer/py/PythonSamplePrintArgTransformer.java
```diff
@@ -0,0 +1,95 @@
...
+   * <p>When the type of the enum value {@code x} is a top-level enum {@code EnumType}, we can call
+   * {@code print(enums.EnumType(x).name)} to print the name of {@code x}.
+   *
+   * <p>When the type of the enum value {@code x} is a nested enum {@code MessageType.EnumType}, we
+   * can call {@code print(enums.MessageType.EnumType(x).name)} to print the name of {@code x}.
+   */
+  @Override
+  public PrintArgView generatePrintArg(
+      MethodContext context, OutputView.VariableView variableView) {
+    if (variableView.type() != null && variableView.type().isEnum()) {
```
pongad: I think we should invert this, checking for variable not being enum first. In general, I think it's good practice to handle easy cases earlier.

src/main/java/com/google/api/codegen/transformer/py/PythonSamplePrintArgTransformer.java
```diff
@@ -0,0 +1,95 @@
...
+    TypeRef protoType = ((ProtoTypeRef) type).getProtoType();
+    ProtoElement t = protoType.getEnumType();
+    List<String> names = new ArrayList<>();
+    while (!(t instanceof ProtoFile)) {
+      names.add(t.getSimpleName());
+      t = t.getParent();
+    }
+    names.add("enums");
+    StringBuilder builder = new StringBuilder();
+    for (int i = names.size() - 1; i >= 1; i--) {
```
pongad: Consider using [Lists.reverse](https://google.github.io/guava/releases/23.0/api/docs/com/google/common/collect/Lists.html#reverse-java.util.List-).

```java
for (String name : Lists.reverse(names)) {
  builder.append(".").append(name);
}
return builder.substring(1);
```

hzyi-google:
@pongad PTAL

src/test/java/com/google/api/codegen/gapic/testdata/py/python_library.baseline
```diff
@@ -4376,7 +4376,7 @@ def sample_monolog_about_book():
 
   requests = [request]
   response = client.monolog_about_book(requests)
-  print('The stage of the comment is: {}'.format(enums.EnumName(response.stage).name))
+  print('The stage of the comment is: {}'.format(enums.Comment.Stage(response.stage).name))
```
beccasaurus: This is the bit I care about! Big ol' 👍 for the baseline

src/main/java/com/google/api/codegen/transformer/py/PythonSamplePrintArgTransformer.java
```diff
@@ -51,17 +52,17 @@
   @Override
   public PrintArgView generatePrintArg(
       MethodContext context, OutputView.VariableView variableView) {
-    if (variableView.type() != null && variableView.type().isEnum()) {
+    if (variableView.type() == null || !variableView.type().isEnum()) {
+      return PrintArgView.newBuilder()
+          .segments(ImmutableList.<ArgSegmentView>of(VariableSegmentView.of(variableView)))
+          .build();
+    } else {
       ImmutableList<ArgSegmentView> segments =
```
pongad: I don't think you need an `else` here.

hzyi-google: Ah, yes

PR #2395, opened 2018-10-31
Remove spurious 'CallOptions' fossils from Python examples.

### LINKED ISSUE #2388: Python:  'CallOptions' not imported in examples (nor is it needed for page iteration)
From https://github.com/googleapis/google-cloud-python/issues/5560.

Examples in generated docstrings for methods returning iterators use `CallOptions` without importing it, e.g.:

```python

    # Service calls
    def list_monitored_resource_descriptors(
            self,
            name,
            filter_=None,
            page_size=None,
            retry=google.api_core.gapic_v1.method.DEFAULT,
            timeout=google.api_core.gapic_v1.method.DEFAULT,
            metadata=None):
        """
        Lists monitored resource descriptors that match a filter. This method does not require a Stackdriver account.

        Example:
            >>> from google.cloud import monitoring_v3
            >>>
            >>> client = monitoring_v3.MetricServiceClient()
            >>>
            >>> name = client.project_path('[PROJECT]')
            >>>
            >>> # Iterate over all results
            >>> for element in client.list_monitored_resource_descriptors(name):
            ...     # process element
            ...     pass
            >>>
            >>>
            >>> # Alternatively:
            >>>
            >>> # Iterate over results one page at a time
            >>> for page in client.list_monitored_resource_descriptors(name, options=CallOptions(page_token=INITIAL_PAGE)):
            ...     for element in page:
            ...         # process element
            ...         pass
   ...
   """
```

Worse, that usage is unidiomatic:  the iterators returned from GAPIC methods have a `pages` attribute, which allows users to iterate pages without managing the token at all.  E.g., for the example above, the more correct / idiomatic usage would be:

```python
            >>> # Iterate over results one page at a time
            >>> for page in client.list_monitored_resource_descriptors(name).pages:
            ...     for element in page:
            ...         # process element
            ...         pass
```

tseaver:
Closes #2388.

Note to reviewers:  I'm confident that the expanded versions in the `,baseline` file are correct, but I'm guessing here about how to make the change in the `.snip` files.


andreamlin:
Can you update the .baseline files too? The unit tests will fail if the output client code has changed, but the baseline test resource files haven't.

andreamlin:
It looks like one of the python baseline tests is still failing. Can you run the test suite locally and make sure it's fine on your end?

tseaver:
@andreamlin Whew, getting the test environment set up was painful!  But d65faea gets the one failing test to pass:  I needed to suppress the trailing comma in the argument list.

PR #2397, opened 2018-11-01
Samplegen: import the client and other fixes in PHP standalone samples

hzyi-google:
In PHP standalone samples:
- import the client
- remove namespace
- require autoload


hzyi-google:
PTAL

hzyi-google:
cc/ @beccasaurus

src/main/resources/com/google/api/codegen/php/standalone_sample.snip
```diff
@@ -14,7 +14,7 @@
      */
 
     // [START {@sample.regionTag}]
-    namespace {@sampleFile.fileHeader.packageName};
+    require __DIR__ . '/../../../../vendor/autoload.php';
```
michaelbausor: We shouldn't hardcode the depth of the directories here. The 'vendor' directory will be alongside the generated composer.json file, so we should calculate this based on the output directory of composer.json vs the sample file.

hzyi-google:
PTAL @michaelbausor

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSamplesTransformer.java
```diff
@@ -99,6 +102,12 @@ public PhpGapicSamplesTransformer(
                 callingForm,
                 valueSet);
         String sampleOutputPath = subPath + File.separator + namer.getApiSampleFileName(className);
+        String autoloadPath =
```
michaelbausor: Maybe put a comment explaining this value? Otherwise I imagine people reading this code in future may have trouble working out why it is being done this way.

src/test/java/com/google/api/codegen/gapic/testdata/php/php_library.baseline
```diff
@@ -3265,11 +3278,12 @@ samplePublishSeries($name, $edition);
  */
 
 // [START canonical]
-namespace Google\Cloud\Example\Library\V1;
+require __DIR__ . '/../../../../vendor/autoload.php';
 
-use \Google\Example\Library\V1\Book;
-use \Google\Example\Library\V1\SeriesUuid;
-use \Google\Example\Library\V1\Shelf;
+use Google\Cloud\Example\Library\V1\LibraryServiceClient;
```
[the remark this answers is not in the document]
hzyi-google: :)

PR #2402, opened 2018-11-02
Less magic pathtemplate strings

andreamlin:
Partially addresses https://github.com/googleapis/google-cloud-java/issues/3604.

Don't force pathtemplate strings to start with a literal string.

The status quo, for path templates, is to magically append a literal string to templated paths that begin with a bracketed string, e.g. the path `{project}/zones/{zone}` is rendered as `projects/{project}/zones/{zone}`. I don't remember exactly why we decided to do this, though it may have been related to purity of usage of PathTemplates, where all existing Google grpc path templates start with a `projects` (or possibly `organizations`) substring.

More importantly, this PR allows us to parse a resource string returned from the server (see issue https://github.com/googleapis/google-cloud-java/issues/3604), which often looks like `https://www.googleapis.com/compute/v1/projects/{project}/zones/{zone}`. The [compute.v1.json](https://raw.githubusercontent.com/googleapis/discovery-artifact-manager/master/discoveries/compute.v1.json) Discovery API specifies that the `baseUrl` is exactly `https://www.googleapis.com/compute/v1/projects`. We can then take a resource string from the server, remove the given `baseUrl` prefix, and then parse the remaining suffix `{project}/zones/{zone}` into a ProjectsZoneName. 

See https://github.com/googleapis/google-cloud-java/pull/3899 and https://github.com/googleapis/discovery-artifact-manager/pull/95 for the resulting surface diffs from this generator change.


andreamlin:
PTAL

andreamlin:
I agree that there would be an inconsistency of project name patterns across different Google Cloud APIs, but this change would restore consistency in 
- how a method endpoint is simply `getBaseUrl() + resourceName.toString()`
- how prior users of Compute would have created a serialized resource name for Compute resources, because this is how it's been defined in the Discovery doc
- (this hasn't been implemented yet) what generated code comments would say, which would be constructed from relevant fields like `method.path` and `method.description`.

I'm generally wary of any kind of magic that happens after configgen; all guessed-at logic should be put in configgen. Removing `/projects/` from the end of `baseUrl` is more magic; how would we know to remove that from the baseUrl? Endpoint URLs are defined in the service yaml, which isn't generally an input for discogapic. We get lucky in this API because `projects` is the plural of `{project}`, but we could also encounter a `projects/{project_id}` where we can't figure out if we should remove the ending suffix from the baseurl.

andreamlin:
The circleCI compute-discovery-test fails because this PR would require a new version of the compute_gapic.yaml to be generated using this PR.

PTAL

andreamlin:
Merging into a separate branch, since this would be a breaking change for the Compute client.

PR #2406, opened 2018-11-05
Only grpc-related classes need LRO java import

andreamlin:
HTTP-based clients don't need the LRO import. This had been causing problems when trying to compile the discogapic output before removing unused imports (of which LRO was one).


andreamlin:
PTAL

PR #2407, opened 2018-11-06
Remove pending deprecation spew from unit tests.

### LINKED ISSUE #2389: PendingDeprecationWarning spew from generated tests
Running unit tests for almost any GAPIC-generated package yields a spew of the following:

```python
...firestore/google/cloud/firestore_v1beta1/gapic/firestore_client.py:182: PendingDeprecationWarning: The `client_config` argument is deprecated.
  PendingDeprecationWarning)
```

and:

```
.../firestore/google/cloud/firestore_v1beta1/gapic/firestore_client.py:186: PendingDeprecationWarning: The `channel` argument is deprecated; use `transport` instead.
  '`transport` instead.', PendingDeprecationWarning)
```

I added a `stacklevel=2` argument to both of those warnings (which should be in the generated code anyway).
It turns out that the test for the first one is bogus:  literally every invocation triggers it!

```python
    def __init__(self,
                 transport=None,
                 channel=None,
                 credentials=None,
                 client_config=firestore_client_config.config,
                 client_info=None):
        ...
        # Raise deprecation warnings for things we want to go away.
        if client_config:
            warnings.warn('The `client_config` argument is deprecated.',
                          PendingDeprecationWarning, stacklevel=2)
```

For the second, most the warnings were coming from generated tests!

```python
.../firestore/tests/unit/gapic/v1beta1/test_firestore_client_v1beta1.py:541: PendingDeprecationWarning: The `channel` argument is deprecated; use `transport` instead.
  client = firestore_client.FirestoreClient(channel=channel)
```

The generator certainly shouldn't be emitting tests which use the deprecated pattern (unless maybe there is one which uses `warnings.catch_warnings` to test that it is emitted).

tseaver:
- Add `stacklevel=2` to each emitted `PendingDeprecationWarning`.
- Fix detection of non-default `client_config` argument (make the default `None`, and assign in the body).
- Replace deprecation-emitting `channel=channel` arguments to client constructors with mocks of the `create_channel` utility function.

Closes #2389.

### DIFF
--- a/showcase/python/nox.py
+++ b/showcase/python/nox.py
@@ -36,7 +36,7 @@ def showcase(session, py):
     session.virtualenv_dirname = 'showcase-' + py
 
     # Install all test dependencies, then install this package in-place.
-    session.install('pytest', 'mock')
+    session.install('pytest')
     session.install('--pre', 'googleapis-common-protos')
     session.install(
         '-e',
@@ -60,7 +60,7 @@ def unit(session, py):
     session.virtualenv_dirname = 'unit-' + py
 
     # Install all test dependencies, then install this package in-place.
-    session.install('pytest')
+    session.install('pytest', 'mock')
     session.install('--pre', 'googleapis-common-protos')
     session.install(
         '-e',
--- a/showcase/python/nox.py
+++ b/showcase/python/nox.py
@@ -36,7 +36,7 @@ def showcase(session, py):
     session.virtualenv_dirname = 'showcase-' + py
 
     # Install all test dependencies, then install this package in-place.
-    session.install('pytest')
+    session.install('pytest', 'mock')
     session.install('--pre', 'googleapis-common-protos')
     session.install(
         '-e',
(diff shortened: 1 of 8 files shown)


showcase/python/nox.py
```diff
@@ -36,7 +36,7 @@ def showcase(session, py):
     session.virtualenv_dirname = 'showcase-' + py
 
     # Install all test dependencies, then install this package in-place.
-    session.install('pytest')
+    session.install('pytest', 'mock')
```
lukesneeringer: Do we need `mock` on 3.x?

tseaver: Yes, because we import it as `mock` everywhere, rather than `unittest.mock`, for compatibility with 2.7.

tseaver:
@lukesneeringer I'm not sure how to debug the [`python-test` failure](https://circleci.com/gh/googleapis/gapic-generator/7802), given that I can't see the generated output.  The `gradlew googleJavaFormat build` tests pass on my local machine:  is there a way to run them such that they don't remove the generated code?

PR #2409, opened 2018-11-06
Discovery gapic CircleCI integration test (WIP)

andreamlin:


andreamlin:
We'll just use artman for now

PR #2412, opened 2018-11-07
CircleCI test for discogapic

andreamlin:
Run the gapic-generator jar on the Compute API Discovery doc to generate a client, and then test it.

Use compute.v1.json and compute_gapic.yaml from a fetched versioned commit of discovery-artifact-manager.

Add a build.gradle to put in the generated Compute client so that we can use Gradle to test the *Test.java files.


andreamlin:
PTAL

PR #2414, opened 2018-11-07
Samplegen: read from files

hzyi-google:
Set up infrastructure for standalone samples to be able to read from files and assign the file content to a local variable. 

- Does not yet work for sample function parameters.
- IO utils are not yet imported in any languages.

### DIFF
--- a/src/main/java/com/google/api/codegen/metacode/InitCodeContext.java
+++ b/src/main/java/com/google/api/codegen/metacode/InitCodeContext.java
@@ -102,7 +102,7 @@ public abstract class InitCodeContext {
         .outputType(InitCodeOutputType.SingleObject)
         .additionalInitCodeNodes(ImmutableList.of())
         .initFieldConfigStrings(ImmutableList.of())
-        .sampleParamConfigMap(ImmutableMap.<String, SampleParameterConfig>of());
+        .sampleParamConfigMap(ImmutableMap.of());
   }
 
   @AutoValue.Builder
--- a/src/main/java/com/google/api/codegen/metacode/InitCodeContext.java
+++ b/src/main/java/com/google/api/codegen/metacode/InitCodeContext.java
@@ -16,6 +16,7 @@ package com.google.api.codegen.metacode;
 
 import com.google.api.codegen.config.FieldConfig;
 import com.google.api.codegen.config.FieldModel;
+import com.google.api.codegen.config.SampleParameterConfig;
 import com.google.api.codegen.config.TypeModel;
 import com.google.api.codegen.util.Name;
 import com.google.api.codegen.util.SymbolTable;
@@ -65,11 +66,21 @@ public abstract class InitCodeContext {
   /** Returns init config strings. */
   public abstract List<String> initFieldConfigStrings();
 
+  /** Configuration for the sample arguments. */
+  public abstract ImmutableMap<String, SampleParameterConfig> sampleParamConfigMap();
+
   /**
    * When generating samples, the sample function accept these parts of the request object as
    * arguments.
    */
-  public abstract ImmutableList<String> sampleArgStrings();
+  public ImmutableList<String> sampleArgStrings() {
+    return sampleParamConfigMap()
+        .values()
+        .stream()
+        .filter(SampleParameterConfig::isSampleArgument)
+        .map(SampleParameterConfig::identifier)
+        .collect(ImmutableList.toImmutableList());
+  }
 
   /**
    * Allows additional InitCodeNode objects which will be placed into the generated subtrees. This
@@ -91,7 +102,7 @@ public abstract class InitCodeContext {
         .outputType(InitCodeOutputType.SingleObject)
         .additionalInitCodeNodes(ImmutableList.of())
         .initFieldConfigStrings(ImmutableList.of())
-        .sampleArgStrings(ImmutableList.of());
+        .sampleParamConfigMap(ImmutableMap.<String, SampleParameterConfig>of());
   }
 
   @AutoValue.Builder
@@ -110,7 +121,8 @@ public abstract class InitCodeContext {
 
     public abstract Builder initFieldConfigStrings(List<String> configStrings);
 
-    public abstract Builder sampleArgStrings(ImmutableList<String> val);
+    public abstract Builder sampleParamConfigMap(
+        ImmutableMap<String, SampleParameterConfig> configs);
 
     public abstract Builder initValueConfigMap(Map<String, InitValueConfig> configMap);
 
--- a/src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
+++ b/src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
@@ -18,6 +18,7 @@ import com.google.api.codegen.config.FieldConfig;
 import com.google.api.codegen.config.FieldModel;
 import com.google.api.codegen.config.OneofConfig;
 import com.google.api.codegen.config.ProtoTypeRef;
+import com.google.api.codegen.config.SampleParameterConfig;
 import com.google.api.codegen.config.TypeModel;
 import com.google.api.codegen.metacode.InitCodeContext.InitCodeOutputType;
 import com.google.api.codegen.util.Name;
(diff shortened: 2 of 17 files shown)


hzyi-google:
@beccasaurus @pongad @vchudnov-g PTAL

src/main/java/com/google/api/codegen/metacode/InitCodeContext.java
```diff
@@ -91,7 +102,7 @@ public static Builder newBuilder() {
         .outputType(InitCodeOutputType.SingleObject)
         .additionalInitCodeNodes(ImmutableList.of())
         .initFieldConfigStrings(ImmutableList.of())
-        .sampleArgStrings(ImmutableList.of());
+        .sampleParamConfigMap(ImmutableMap.<String, SampleParameterConfig>of());
```
pongad: I could be wrong, but I don't think you need type parameters here?

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -478,6 +485,39 @@ private InitCodeLineView generateMapInitCodeLine(MethodContext context, InitCode
...
+        // File names are always strings
+        value =
+            typeTable.renderPrimitiveValue(
+                ProtoTypeRef.create(TypeRef.fromPrimitiveName("string")), value);
+        break;
+      case Variable:
+        value = namer.localVarReference(Name.anyLower(value));
+        break;
+      default:
+        throw new IllegalArgumentException("Unhandled init value type");
```
pongad: You can also do `checkArgument(false, "unhandled init value type")` :)

src/main/java/com/google/api/codegen/viewmodel/ReadFileInitCodeLineView.java
```diff
@@ -0,0 +1,47 @@
...
+import com.google.api.codegen.metacode.InitCodeLineType;
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+public abstract class ReadFileInitCodeLineView implements InitCodeLineView {
+  public InitCodeLineType lineType() {
+    return InitCodeLineType.ReadFileInitLine;
+  }
+
+  public abstract String typeName();
```
pongad: Isn't the type name already known to be `ByteString` etc?

src/test/java/com/google/api/codegen/gapic/testdata/py/python_library.baseline
```diff
@@ -4153,7 +4153,11 @@ def sample_discuss_book():
   client = library_v1.LibraryServiceClient()
 
   name = 'BASIC'
-  request = {'name': name}
+  file_name = os.path.join(os.path.dirname(__file__), 'comment_file')
```
beccasaurus: Let's presume that the path provided is always absolute

src/main/java/com/google/api/codegen/viewmodel/ReadFileInitCodeLineView.java
hzyi-google: The protobuf type is definitely `bytes`, but the type name may be different in different languages?

hzyi-google: Oh I think I see what you mean... are you saying that we could render the correct type in snippets?

src/test/java/com/google/api/codegen/gapic/testdata/nodejs/nodejs_library.baseline
```diff
@@ -275,8 +275,14 @@ const stream = client.discussBook().on('data', response => {
   // doThingsWith(response)
 });
 const name = 'BASIC';
+const fileName = 'comment_file';
+const comment = fs.readFileSync(fileName).toString('base64');
```
beccasaurus: 👍

src/test/java/com/google/api/codegen/gapic/testdata/php/php_library.baseline
```diff
@@ -2757,8 +2758,12 @@ function sampleDiscussBook()
     $libraryServiceClient = new LibraryServiceClient();
 
     $name = 'BASIC';
+    $comment = file_get_contents('comment_file');
```
beccasaurus: 👍

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
```diff
@@ -364,6 +367,34 @@ private static void validateType(
...
+    if (ROOT_KEY.equals(key)) {
+      fieldPath = "";
+    } else if (parentFieldPath.isEmpty()) {
+      fieldPath = key;
+    } else {
+      fieldPath = parentFieldPath + "." + key;
+    }
+    this.sampleParamConfig = sampleParamConfigMap.get(fieldPath);
+    if (sampleParamConfig != null && sampleParamConfig.readFromFile()) {
+      setLineType(InitCodeLineType.ReadFileInitLine);
```
vchudnov-g: If the condition is false, do you need to `setLineType` to something else?

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -478,6 +485,39 @@ private InitCodeLineView generateMapInitCodeLine(MethodContext context, InitCode
...
   }
 
+  // TODO(hzyi): generate necessary imports
+  private InitCodeLineView generateReadFileInitCodeLine(MethodContext context, InitCodeNode item) {
+    ReadFileInitCodeLineView.Builder surfaceLine = ReadFileInitCodeLineView.newBuilder();
+    SurfaceNamer namer = context.getNamer();
+    ImportTypeTable typeTable = context.getTypeTable();
+    checkState(
+        item.getType().isBytesType(),
+        "Error setting init code node %s to reading from file. "
```
vchudnov-g: A user may not node about "init code node", so I suggest rephrasing slightly:
"Error setting \"%s\" to be read from file."

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -354,6 +354,11 @@ message SampleInitAttribute {
   // If `sample_argument_name` is not empty, the part of the request object specified by `path`
   // should be accepted as a parameter, with the given name, to the sample function.
   string sample_argument_name = 2;
+
+  // If `read_file` is true, in the initialization part of the sample, the value of this
```
vchudnov-g: rephrase for clarity:
```
If `read_file` is true, then during sample initialization, the value...
```

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -354,6 +354,11 @@ message SampleInitAttribute {
   // If `sample_argument_name` is not empty, the part of the request object specified by `path`
   // should be accepted as a parameter, with the given name, to the sample function.
   string sample_argument_name = 2;
+
+  // If `read_file` is true, in the initialization part of the sample, the value of this 
+  // attribute gets replaced by the contents of the local file with the given name.
```
vchudnov-g: Add a note about comma separated files if the field is a repeated bytes.

src/test/java/com/google/api/codegen/gapic/testdata/py/python_library.baseline
vchudnov-g: as clarified at the meeting, don't require `dirname(__file__)`. Assume a dir relative to the cwd, or an absolute path if it starts with `/`

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
hzyi-google: It would be an no-op. The `lineType` is handled at different places, e.g. `simpleLineType` [here](https://github.com/googleapis/gapic-generator/blob/c12a7da29e0571a940664c4fa17080cf096b7dae/src/main/java/com/google/api/codegen/metacode/InitCodeNode.java#L308).

hzyi-google:
PTAL @vchudnov-g @pongad

PR #2415, opened 2018-11-08
Update test resources to use new Resource definitions

andreamlin:
- use the new resource_definition annotation
- make Resource.name and ResourceSet.name be CamelCase
- expand all paths to include resource identifiers as wildcards, e.g. `projects/{project}` instead of `projects/*`
- use resource_references in the ResourceSet
- temporarily remove references to `google.api.Project`, which has been deleted in https://github.com/googleapis/api-common-protos/pull/20/files

### DIFF
--- a/src/test/java/com/google/api/codegen/protoannotations/testdata/library_gapic.yaml
+++ b/src/test/java/com/google/api/codegen/protoannotations/testdata/library_gapic.yaml
@@ -2,7 +2,7 @@ type: com.google.api.codegen.ConfigProto
 config_schema_version: 1.0.0
 language_settings:
   java:
-    package_name: com.google.gcloud.pubsub.v1
+    package_name: com.google.example.library.v1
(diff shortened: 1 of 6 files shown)


andreamlin:
PTAL

src/test/java/com/google/api/codegen/protoannotations/testdata/library_gapic.yaml
```diff
@@ -2,7 +2,7 @@ type: com.google.api.codegen.ConfigProto
 config_schema_version: 1.0.0
 language_settings:
   java:
-    package_name: com.google.gcloud.pubsub.v1
+    package_name: com.google.example.library.v1
```
andreamlin: All changes in this file are just to make this file the same as the other library_gapic.yaml test resource that is used to generate old-style GAPICs.

PR #2416, opened 2018-11-08
Account for placeholders in additional bindings

dwsupplee:
For more background: https://github.com/googleapis/google-cloud-php/pull/1411

/cc @tmatsuo @jdpedrie


michaelbausor:
Can we also update our tests so that this change is visible in the baseline file? I presume we will need to modify an existing method in our test proto (e.g. https://github.com/googleapis/gapic-generator/blob/master/src/test/java/com/google/api/codegen/testsrc/libraryproto/library.proto#L130) so that it has an additional binding with a different placeholder structure.

PR #2418, opened 2018-11-09
Python: use 'ProtoFiles.getProtoFiles' to sort enums by filename.

### LINKED ISSUE #2319: Order of entries in generated 'enums' is not stable
E.g., see: https://github.com/GoogleCloudPlatform/google-cloud-python/pull/6015/files.

The order should match the order in the `.proto` file, except that they might be collected from more than one:  if that case in general, then sort them.

tseaver:
Closes #2319.

I hope. :)

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/py/PythonGapicSurfaceTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/py/PythonGapicSurfaceTransformer.java
@@ -27,6 +27,7 @@ import com.google.api.codegen.config.ProtoApiModel;
 import com.google.api.codegen.config.SampleSpec.SampleType;
 import com.google.api.codegen.gapic.GapicCodePathMapper;
 import com.google.api.codegen.gapic.GapicParser;
+import com.google.api.codegen.gapic.ProtoFiles;
 import com.google.api.codegen.transformer.DefaultFeatureConfig;
 import com.google.api.codegen.transformer.DynamicLangApiMethodTransformer;
 import com.google.api.codegen.transformer.FeatureConfig;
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/api/codegen/transformer/py/PythonGapicSurfaceTransformer.java
```diff
@@ -170,13 +171,21 @@ public PythonGapicSurfaceTransformer(
...
+        generateEnumView(productConfig, modelTypeTable, namer, getProtoFiles(productConfig));
     if (!enumFile.elementDocs().isEmpty()) {
       serviceSurfaces.add(enumFile);
     }
     return serviceSurfaces.build();
   }
 
+  private List<ProtoFile> getProtoFiles(GapicProductConfig productConfig) {
+    ImmutableList.Builder<ProtoFile> files = ImmutableList.builder();
+    for (ProtoFile protoFile : ProtoFiles.getProtoFiles(productConfig)) {
```
andreamlin: Can this method just be `files = ImmutableList.copyOf(ProtoFiles.getProtoFiles(productConfig))`?

andreamlin:
Are there any baseline changes, or are there any changes in the existing generated clients once this goes in?

src/main/java/com/google/api/codegen/transformer/py/PythonGapicSurfaceTransformer.java
tseaver: If that is more idiomatic, sure.  My familiarity with Java is some years out of date at this point.

PR #2419, opened 2018-11-09
[Proto Annotations] Parse out all Resources and ResourceSets

andreamlin:


src/main/java/com/google/api/codegen/util/ProtoParser.java
```diff
@@ -46,6 +61,24 @@ public String getResourcePath(Field element) {
     return null;
   }
 
+  @SuppressWarnings("unchecked")
+  @Nullable
+  private <T, O extends Message, E extends ProtoElement> T getProtoExtension(
```
pongad: that's a gnarly signature!

src/main/java/com/google/api/codegen/util/ProtoParser.java
```diff
@@ -61,21 +94,102 @@ public String getPackageName(Model model) {
...
+        if (Strings.isNullOrEmpty(baseName)) {
+          diagCollector.addDiag(
+              Diag.error(
+                  SimpleLocation.TOPLEVEL,
+                  "There is a %s option with"
+                      + " no name defined in proto file %s. %s.name is required.",
+                  fileExtension.getDescriptor().getFullName(),
+                  protoFile.getFullName(),
+                  fileExtension.getDescriptor().getFullName()));
+        }
```
pongad: should we `continue` here?

src/main/java/com/google/api/codegen/util/ProtoParser.java
andreamlin: action required? i just wanted to cram the internal proto extension parsing into a tiny private method that the outside world won't have to look at

src/main/java/com/google/api/codegen/util/ProtoParser.java
andreamlin: ooh good call

andreamlin:
wait don't review yet; gotta fix CI first

src/main/java/com/google/api/codegen/util/ProtoParser.java
pongad: no action required, just took me a while decipher :)

andreamlin:
Fixed CI
PTAL

PR #2420, opened 2018-11-09
Samplegen: honor sample argument name

### LINKED ISSUE #2366: honor `sample_argument_name` config

hzyi-google:
Fix #2366


hzyi-google:
PTAL @pongad @vchudnov-g @beccasaurus

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
```diff
@@ -387,12 +387,31 @@ private void resolveSampleParamConfigs(
...
+      }
+
+      // If sample_argument_name is specified in config, set identifier to this name if the name has
+      // not be used yet and error out otherwise.
+      if (sampleParamConfig.isSampleArgument()) {
+        Name argName = Name.anyLower(sampleParamConfig.sampleArgumentName());
+        if (!argName.equals(identifier)) {
+          Preconditions.checkArgument(
+              !context.symbolTable().isSymbolUsed(argName),
+              "sample_argument_name \"%s\" is not valid.",
```
vchudnov-g: it might be more helpful to say " is already in use"

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
```diff
@@ -387,12 +387,31 @@ private void resolveSampleParamConfigs(
...
+
+    if (sampleParamConfig != null) {
+
+      // If read_file is set to true in config, change the line type to ReadFileInitLine.
+      if (sampleParamConfig.readFromFile()) {
+        setLineType(InitCodeLineType.ReadFileInitLine);
+      }
+
+      // If sample_argument_name is specified in config, set identifier to this name if the name has
+      // not be used yet and error out otherwise.
```
vchudnov-g: s/be/been/

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
```diff
@@ -387,12 +387,31 @@ private void resolveSampleParamConfigs(
...
+              context
+                  .symbolTable()
+                  .getNewSymbol(Name.anyLower(sampleParamConfig.sampleArgumentName()));
+        }
+      }
     }
-    children
-        .values()
-        .forEach(child -> child.resolveSampleParamConfigs(sampleParamConfigMap, fieldPath));
+    children.values().forEach(child -> child.resolveSampleParamConfigs(context, fieldPath));
```
vchudnov-g: Can it ever happen that `sampleParamConfig != null` above and then we recurse here? I would expect that we only recurse if `sampleParamConfig == null`. Is that correct? If so, an assertion would be good. If not, I'd like to understand this better.

src/test/java/com/google/api/codegen/testsrc/libraryproto/library_gapic.yaml
```diff
@@ -240,7 +240,7 @@ interfaces:
         - edition=123
         attributes:
         - parameter: shelf.name
-          sample_argument_name: name
+          sample_argument_name: shelf_name
```
vchudnov-g: For brevity, I slightly lean toward s/`sample_argument_name`/`sample_argument`/, but with the same semantics. Thoughts? @beccasaurus

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
hzyi-google: It might be possible as we might specify the `sample_argument_name` (or `sample_argument`) for both a parent node and one of it's children. It's never used for now but should be a valid use case. For example:
```
atrributes:
- parameter: foo
  sample_argument_name: foo_name
- parameter: foo.bar
  sample_argument_name: foo_bar_name
```

vchudnov-g: But does that make sense? They are overlapping specifications. What are the semantics? Should the latter one override the former one? Should the more specific one override the more general one?

If our code doesn't have set semantics for this situation, or if we don't have a way of dealing with such overlap (including tests!), I think it's safer to explicitly disallow it; we can deliberately add it later if needed.

hzyi-google: Oh I was wrong! I confused `sample_argument_name` with `variable_name` :/ Yeah I think you are right that it would never happen. I'll add a check there. And actually we do have [assertions](https://github.com/googleapis/gapic-generator/blob/e052651a59cf6e5a12c1f4c77dcc58d17ecdd670/src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java#L246).

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
pongad: Why is it an error? Shouldn't we just rename it like we do other variables?

src/main/java/com/google/api/codegen/util/SymbolTable.java
```diff
@@ -87,6 +87,11 @@ public String getNewSymbol(String desiredName) {
     return desiredName + suffix;
   }
 
+  /** Returns true if a {@code symbol} is already used. */
+  public boolean isSymbolUsed(Name symbol) {
```
pongad: The `SymbolTable` deals only with symbols, I think we can just call this `isUsed` here.

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
vchudnov-g: Hmm, I would lean towards erroring rather than automatically changing a variable name the user has explicitly specified

src/main/java/com/google/api/codegen/util/SymbolTable.java
vchudnov-g: Maybe so that it reads better in English it should be `SymbolTable.uses(symbol)` or `SymbolTable.contains(symbol)'?

pongad: Good idea!

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
pongad: That makes sense to me too.

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
hzyi-google: @vchudnov-g It seems repetitive to do what `assertNoOverlap` again here, since we are going to do this check after the tree is built. Are you OK if I just skip the recursion if `sampleParamConfig != null`? I've moved the recursion into an `else` block.

hzyi-google:
PTAL @pongad @vchudnov-g

src/main/java/com/google/api/codegen/metacode/InitCodeNode.java
vchudnov-g: sgtm. I'd recommend adding a brief comment making that we're purposefully not recursing since overlaps are not allowed.

vchudnov-g:
Oh, I'm signed off. My approval remains in place. ;-)

PR #2421, opened 2018-11-13
fix longrunning issue

hzyi-google:


src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceNamer.java
```diff
@@ -413,9 +413,9 @@ public String getAndSaveTypeName(ImportTypeTable typeTable, TypeModel type) {
    * <p>All Google cloud java libraries have package names like "com.google.cloud.library.v1" and
    * the respective examples have package names like
    * "com.google.cloud.examples.library.v1.snippets". This method assumes {@code packageName} has
-   * the format of "com.google.(.+).artifact.name". This works for both existing libraries and the
-   * baseline test. We will need to adjust this if in the future there are libraries that do not
-   * follow the package name format assumed here.
+   * the format of "com.google.orgname.artifact.name", where "artifact.name" could be empty. This
```
michaelbausor: This description doesn't seem quite right for the longrunning case - `longrunning` isn't an orgname.

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceNamer.java
```diff
@@ -424,11 +424,14 @@ public String getExamplePackageName() {
...
-    return "com.google."
-        + simpleOrgName
-        + ".examples"
-        + packageName.replaceFirst(simpleOrgName, "")
-        + ".snippets";
+    checkArgument(
+        !packageName.isEmpty(),
+        "package name should have at least one more component than 'com.google'");
+    String simpleOrgName =
+        packageName.indexOf('.') < 0
```
michaelbausor: Can we do this calculation just once?

hzyi-google:
@michaelbausor PTAL. I was having a hard time coming up descriptive variable names so any suggestions are welcomed :)

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceNamer.java
```diff
@@ -427,11 +425,9 @@ public String getExamplePackageName() {
...
         !packageName.isEmpty(),
         "package name should have at least one more component than 'com.google'");
-    String simpleOrgName =
-        packageName.indexOf('.') < 0
-            ? packageName
-            : packageName.substring(0, packageName.indexOf('.'));
-    String artifactName = packageName.replaceFirst(simpleOrgName, "");
-    return "com.google." + simpleOrgName + ".examples" + artifactName + ".snippets";
+    int index = packageName.indexOf('.');
+    String component = index < 0 ? packageName : packageName.substring(0, index);
```
michaelbausor: What about firstComponent and remainingComponents?

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceNamer.java
```diff
@@ -410,12 +410,10 @@ public String getAndSaveTypeName(ImportTypeTable typeTable, TypeModel type) {
...
-   * <p>All Google cloud java libraries have package names like "com.google.cloud.library.v1" and
-   * the respective examples have package names like
-   * "com.google.cloud.examples.library.v1.snippets". This method assumes {@code packageName} has
-   * the format of "com.google.orgname.artifact.name", where "artifact.name" could be empty. This
-   * works for both existing libraries and the baseline test. We will need to adjust this if in the
-   * future there are libraries that do not follow the package name format assumed here.
+   * <p>Currently we assume that package names always start with "com.google.". For example, if
+   * package name is "com.google.foo", the sample package name returned by this method will be
+   * "com.google.foo.examples.snippets". If package name is "com.google.foo.bar", the sample package
+   * name returned by this method will be "com.google.foo.examples.bar.snippets".
```
michaelbausor: This looks good. Perhaps add this afterwards:

We structure the example package name in this way because in the case of a package named "com.google.foo.bar", 'foo' is very often the organization name, and this lets us group examples from the same org into a common package. E.g. "com.google.cloud.library.v1"

PR #2424, opened 2018-11-13
[Proto Annotations] Better parsing of ResourceNameConfigs

andreamlin:
Prioritize GAPIC config values over those from proto annotations.


andreamlin:
PTAL

andreamlin:
Breaking up this PR into smaller ones.

PR #2425, opened 2018-11-13
[Proto Annotations] Update test resources

andreamlin:
- removing the protoannotations version of library_gapic.yaml; we can just use one source of truth for library_gapic.yaml
- library.proto is updated to reflect the singular  library_gapic.yaml
- copy over gapic/testdata/java/java_library.baseline into the protoannotations dir version

After https://github.com/googleapis/gapic-generator/pull/2424/ is checked in, we can write the baseline test using proto annotations with a gapic.


andreamlin:
PTAL

src/test/java/com/google/api/codegen/protoannotations/testdata/java_library.baseline
```diff
@@ -7240,786 +7161,820 @@ public class LibraryServiceStubSettings extends StubSettings<LibraryServiceStubS
     }
   }
 }
-============== file: src/main/java/samples/com/google/gcloud/pubsub/v1/discussbookcallable/DiscussBookCallableCallableStreamingBidiProg.java ==============
-//// [ This is an auto-generated sample file produced by the gapic-generator. Sample name: "DiscussBookCallableCallableStreamingBidiProg" ]
-//// STUB standalone sample "DiscussBookCallableCallableStreamingBidiProg" /////
-
-// FIXME: Insert here set-up comments that we never want to display in cloudsite. These are seen by users perusing the samples directly in the repository.
+============== file: src/main/java/samples/com/google/example/library/v1/discussbookcallable/DiscussBookCallableCallableStreamingBidiProg.java ==============
```
hzyi-google: I think we are not yet able to generate samples using protoannotations. Is the baseline copied here for testing generating libraries using protoannotations? If so the samples shouldn't be here.

src/test/java/com/google/api/codegen/protoannotations/testdata/library.proto
```diff
@@ -810,6 +809,7 @@ message GetBookFromAnywhereRequest {
   // An alternate book name, used to test restricting flattened field to a
   // single resource name type in a oneof.
   string alt_book_name = 2 [
+    (google.api.field_behavior) = REQUIRED,
```
hzyi-google: I don't have enough knowledge about the protoannotation migration work so probably have someone else review this part?

src/test/java/com/google/api/codegen/protoannotations/testdata/java_library.baseline
andreamlin: This baseline would be generated from using both the GAPIC config and the annotated proto file, so the samples would exist.

This test is to make sure there aren't regressions once we introduce annotations in the proto files (even with the GAPIC config).

src/test/java/com/google/api/codegen/protoannotations/testdata/library.proto
andreamlin: @lukesneeringer

lukesneeringer: Assuming it is actually a required field (the API would send INVALID_ARGUMENT if it was unset, or something similar) this seems reasonable.

andreamlin: Yep, this is to agree with the corresponding library_gapic.yaml, which marks this field as required (see [here](https://github.com/googleapis/gapic-generator/blob/e72072aea215d87b00e6c931b869d53a4e062f11/src/test/java/com/google/api/codegen/testsrc/libraryproto/library_gapic.yaml#L458))

andreamlin:
@vchudnov-g would you mind reviewing this? (hzyi is OOO)

src/test/java/com/google/api/codegen/protoannotations/testdata/java_library.baseline
```diff
@@ -76,7 +76,7 @@ task allJars(type: Copy) {
   // Replace with `from configurations.testRuntime, jar` to include test dependencies
   from configurations.runtime, jar
 }
-============== file: src/main/java/com/google/gcloud/pubsub/v1/LibraryClient.java ==============
+============== file: src/main/java/com/google/example/library/v1/LibraryClient.java ==============
```
vchudnov-g: Why is `pubsub` not part of the baseline? I know we had talked about making the baselines smaller, but I assumed that would be done separately from the annotations. Could you bring me up to speed?

src/test/java/com/google/api/codegen/protoannotations/testdata/java_library.baseline
vchudnov-g: I'm confused, because the library_gapic.yaml is deleted below.....

andreamlin: Yes, there was a library_gapic.yaml in a `protoannotations` test folder, and and another file of the same name in the `gapic` test resource folder. This PR removes the former, so there is only one `library_gapic.yaml` for all tests.

src/test/java/com/google/api/codegen/protoannotations/testdata/java_library.baseline
andreamlin: This is just copying over the java_library.baseline from the gapic test source folder. That change was made inhttps://github.com/googleapis/gapic-generator/pull/2399

andreamlin: This java_library.baseline isn't yet used, but it will be soon.

vchudnov-g: To recap our discussion: this is a separate baseline that (for now) you're keeping manually in sync with the original baseline hzyi updated in #2399---you're just copying those changes into here.

src/test/java/com/google/api/codegen/protoannotations/testdata/java_library.baseline
vchudnov-g: Ah, different directory. Gotcha.

PR #2432, opened 2018-11-14
[Proto Annotations] Change getAllResource[Set]s to return Map

andreamlin:
This allows us to fully qualify Resources and ResourceSets.


src/main/java/com/google/api/codegen/config/FlatteningConfig.java
```diff
@@ -305,7 +305,7 @@ private static FlatteningConfig createFlatteningFromProtoFile(
       }
 
       ResourceNameTreatment resourceNameTreatment = ResourceNameTreatment.NONE;
-      String resourceNameType = protoParser.getResourceType(parameterField.getProtoField());
```
andreamlin: This is just a renaming of the function getResourceType() to getResourceReference()

andreamlin:
PTAL

src/main/java/com/google/api/codegen/util/ProtoParser.java
```diff
@@ -129,62 +134,72 @@ public OperationData getLongRunningOperation(Method method) {
...
-                  fieldExtension.getDescriptor().getFullName(),
-                  baseName,
-                  protoFile.getFullName(),
-                  fieldExtension.getDescriptor().getFullName()));
-          continue;
+      // Maps base names to Resource[Sets].
+      Map<String, T> localDefs = new LinkedHashMap<>();
+      // Get definitions from protofile options.
+      List<T> resourcesAtFileLevel = getProtoExtension(protoFile, fileExtension);
+      if (resourcesAtFileLevel != null) {
```
pongad: Here and below, you can `if (resourcesAtFileLevel == null) continue;` to reduce nesting.

src/main/java/com/google/api/codegen/util/ProtoParser.java
```diff
@@ -129,62 +134,72 @@ public OperationData getLongRunningOperation(Method method) {
...
+            diagCollector.addDiag(
+                Diag.error(
+                    SimpleLocation.TOPLEVEL,
+                    "There is a %s option with"
+                        + " no name defined in proto file %s. %s.name is required.",
+                    fileExtension.getDescriptor().getFullName(),
+                    protoFile.getFullName(),
+                    fileExtension.getDescriptor().getFullName()));
+          }
+          if (localDefs.containsKey(baseName)) {
```
pongad: I don't have a strong opinion here, but
```
if (localDefs.put(basename, definition) != null) {
  Diag.error(...)
}
```
would also work because `put` returns the previously-mapped element.

src/main/java/com/google/api/codegen/util/ProtoParser.java
```diff
@@ -129,62 +134,72 @@ public OperationData getLongRunningOperation(Method method) {
...
+      // Get Resource[Set] definitions from fields in message types.
+      for (MessageType message : protoFile.getMessages()) {
+        for (Field field : message.getFields()) {
+          T definition = getProtoExtension(field, fieldExtension);
+          if (definition != null) {
+            if (Strings.isNullOrEmpty(getNameFunc.apply(definition))) {
+              String baseName = getResourceEntityName(field);
+              definition = setNameFunc.apply(definition, baseName);
+            }
+            localDefs.put(getNameFunc.apply(definition), definition);
```
pongad: Does a name conflict here also error?

src/main/java/com/google/api/codegen/util/ProtoParser.java
andreamlin: Hmm, but I don't want to continue if `(resourcesAtFileLevel == null)`? I still want to parse the protofile for more Resources and ResourceSets down below.

pongad: Derp you're right. I got confused by the nesting :P

src/main/java/com/google/api/codegen/util/ProtoParser.java
andreamlin: haha ok done

src/main/java/com/google/api/codegen/util/ProtoParser.java
andreamlin: Ok, also logging an error there too!

andreamlin:
addressed comments
PTAL

andreamlin:
Hmm the bazel test is failing b/c it failed JavaDiscoGapicGeneratorTest, which passed on the test-gapic-generator job. I'm guessing this is a problem with bazel-on-circleci, a resource issue that's bothered us recently.

PR #2433, opened 2018-11-14
[Proto Annotations] Flattening Changes

andreamlin:
- Favor GAPIC config over proto annotations.
- Internally set resource_name_treatment to STATIC_TYPES if there is a field in the method that refers to a Resource or ResourceSet.


andreamlin:
PTAL

andreamlin:
Opened https://github.com/googleapis/gapic-generator/pull/2467 instead.

PR #2435, opened 2018-11-15
Cleanup

andreamlin:
Remove unused methods. Some more lambdas b/c Intellij said so.


andreamlin:
PTAL

src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceDocTransformer.java
```diff
@@ -113,17 +112,11 @@ private boolean isSourceApiInterfaceFile(ProtoFile file) {
...
-      GapicProductConfig productConfig,
-      SurfaceNamer namer,
-      ProtoFile file,
-      boolean hasOverview) {
+      ApiModel model, GapicProductConfig productConfig, SurfaceNamer namer, ProtoFile file) {
     ImmutableList.Builder<ModuleView> moduleViews = ImmutableList.builder();
     for (String moduleName : namer.getApiModules()) {
       if (moduleName.equals(namer.getApiWrapperModuleVersion()) && file != null) {
         moduleViews.add(generateTocModuleView(model, file, productConfig, namer, moduleName));
-      } else if (moduleName.equals(namer.getModuleServiceName()) && hasOverview) {
```
pongad: For my own understanding, what was this doing?

andreamlin:
addressed comments
PTAL

src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceDocTransformer.java
pongad: Hmm yeah you're right. @jbolinger can you confirm that this is safe to do?

pongad:
This LGTM but I think we should have someone familiar with Ruby to sign off.

andreamlin:
@landrito would you mind taking a look at this PR?

src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceDocTransformer.java
landrito: IIRC this had to do with adding some meta documentation to some generated files. I assume there would have been some feedback saying that it was unnecessary causing the documentation to be removed and this wasn't cleaned up.

landrito:
LGTM assuming that overview view is not being used in templates.

PR #2436, opened 2018-11-16
[Proto Annotations] ResourceName one ofs from ResourceSets

andreamlin:
Actually parse the path templates from a Resource to create FieldConfigs.

When a ResourceSet refers to resource definitions by name, find the referent Resource and make a ResourceNameOneofConfig from that.

### DIFF
--- a/src/main/java/com/google/api/codegen/config/GapicProductConfig.java
+++ b/src/main/java/com/google/api/codegen/config/GapicProductConfig.java
@@ -14,6 +14,8 @@
  */
 package com.google.api.codegen.config;
 
+import com.google.api.Resource;
+import com.google.api.ResourceSet;
 import com.google.api.codegen.CollectionConfigProto;
 import com.google.api.codegen.CollectionOneofProto;
 import com.google.api.codegen.ConfigProto;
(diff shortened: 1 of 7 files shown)


andreamlin:
PTAL

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -436,52 +455,183 @@ public static GapicProductConfig createDummyInstance(
...
+    ImmutableMap<String, ResourceNameOneofConfig> resourceNameOneofConfigsFromProtoFile =
+        createResourceNameOneofConfigsFromProtoFile(
+            diagCollector,
+            fullyQualifiedSingleResourceNameConfigsFromProtoFile,
+            resourceSetDefs,
+            protoParser);
+
+    // Populate a SingleResourceNameConfigs map, using just the unqualified names.
+    Map<String, SingleResourceNameConfig> singleResourceConfigsFromProtoFile =
+        new LinkedHashMap<>();
```
pongad: Just use hash map, I assume the `resourceNameConfigs` below will reorder things anyway?

andreamlin: ah true

PR #2438, opened 2018-11-16
Remove escapePathTemplates

andreamlin:
We no longer have to guess at how to escape something like `projects/*/resources/*` into `projects/{project}/resources/{resource}` because the API config spec has been updated to require resource identifiers within the resource paths.


andreamlin:
PTAL

PR #2439, opened 2018-11-17
static import assertThat, like the documentation suggests

pongad:


pongad:
See section 2 of https://google.github.io/truth/ . I think one reason is that we have both `Truth.assertThat` and `Truth8.assertThat` for Java 8 stuff. `static import`s, surprisingly, lets you import both and picks the right one on each call site.

Issue #2429, opened 2018-11-14
Python: bigtable enum generation missing 'StorageType' from 'google/bigtable/admin/v2/common.proto'


tseaver:
See today's [Bigtable autosynth PR](https://github.com/googleapis/google-cloud-python/pull/6492).

I'm worried that somehow the change is due to the merge last week of #2418.

andreamlin:
Are able to use this change in a local fork of gapic-generator to generate the desired Bigtable's python client?

andreamlin:
@tseaver Are you able to pull the latest gapic-generator and successfully generate the Bigtable python client?

tseaver:
@andreamlin 

> Are you able to pull the latest gapic-generator and successfully generate the Bigtable python client?

The only way I know to do that is via `synthtool`, which pulls down the latest `artman` docker image.  Unfortunately, it does not generate the `StorageType` enum:

```bash
$ ../../gcp-tools/bin/synthtool --version
2018.11.8
$ ../../gcp-tools/bin/synthtool 
synthtool > Executing /home/tseaver/projects/agendaless/Google/src/gcp/bigtable/synth.py.
synthtool > Cloning googleapis.
synthtool > Running generator for google/bigtable/artman_bigtable.yaml.
synthtool > Ensuring dependencies.
synthtool > Pulling artman image.
latest: Pulling from googleapis/artman
Digest: sha256:90f9d15e9bad675aeecd586725bce48f5667ffe7d5fc4d1e96d51ff34304815b
Status: Image is up to date for googleapis/artman:latest
synthtool > Generated code into /home/tseaver/.cache/synthtool/googleapis/artman-genfiles/python/bigtable-v2.
synthtool > Running generator for google/bigtable/admin/artman_bigtableadmin.yaml.
synthtool > Ensuring dependencies.
synthtool > Pulling artman image.
latest: Pulling from googleapis/artman
Digest: sha256:90f9d15e9bad675aeecd586725bce48f5667ffe7d5fc4d1e96d51ff34304815b
Status: Image is up to date for googleapis/artman:latest
synthtool > Generated code into /home/tseaver/.cache/synthtool/googleapis/artman-genfiles/python/bigtable-admin-v2.
synthtool > Replaced "'google-cloud-bigtable-admin'" in google/cloud/bigtable_admin_v2/gapic/bigtable_instance_admin_client.py.
[...]

PR #2440, opened 2018-11-19
Revert "Fix enum imports (#2430)"

andreamlin:
This reverts commit e72072aea215d87b00e6c931b869d53a4e062f11. (PR https://github.com/googleapis/gapic-generator/pull/2430)

It broke  artman CircleCI: https://circleci.com/gh/googleapis/artman/8422#artifacts/containers/0

https://github.com/googleapis/gapic-generator/pull/2441 is also a fix for the same problem.

PR #2442, opened 2018-11-19
Revert "Python: use 'ProtoFiles.getProtoFiles' to sort enums by filename."

tseaver:
Reverts googleapis/gapic-generator#2418

See: #2429.


andreamlin:
Are able to use this change in a local fork of gapic-generator to generate the desired Bigtable's python client?

tseaver:
@andreamlin 

> Are able to use this change in a local fork of gapic-generator to generate the desired Bigtable's python client?

I don't have enough knowledge of the toolchain to figure that out:  I'm just backing the change out because the bug appeared to occur after the change from #2418 was propagated to the artman image.

tseaver:
@andreamlin Should I close this PR, given your change in #2440?

andreamlin:
@tseaver right now i'll generate the python client with and without my change, and see if it the fix fixes this

andreamlin:
Verified locally that without this PR (the reverting PR), the `StorageType` class (from the linked bigtable synth PR) is not generated in the Python client, and with this PR, artman will generate the `StorageType` class. Therefore, we should check in this PR.

Thanks a lot for opening this reverting PR!

PR #2443, opened 2018-11-19
Run Artman smoke tests as part of CircleCI

andreamlin:
Run the full suite of artman smoke tests. This does even more tightly couple gapic-generator to artman for now, but it will save us time down the road when we go to release artman and gapic-generator has broken artman.

Soon after this, I can put in a CircleCI job filter so these smoke tests only run on the master branch, given that they take a long time (~40 minutes).


andreamlin:
PTAL

Check out https://circleci.com/gh/googleapis/gapic-generator/9989 for the artman smoke test run!

andreamlin:
Oh wow it caught some actual failures that are locally reproducible, namely the containeranalysis_v1beta1 artman failures.

andreamlin:
This is blocked by https://github.com/googleapis/gapic-generator/issues/2454, which is a regression in gapic-generator that prevents some artman calls from succeeding.

andreamlin:
https://github.com/circleci/circleci-images/issues/275 causes the Ruby failure

PR #2444, opened 2018-11-20
Samplegen: import io utilities for Java and Python

hzyi-google:
Import what we need to read files for standalone samples in Java and Python. PHP does not need extra imports.


hzyi-google:
PTAL @pongad @vchudnov-g @beccasaurus

src/main/java/com/google/api/codegen/transformer/StandardImportSectionTransformer.java
```diff
@@ -32,7 +36,21 @@ public ImportSectionView generateImportSection(TransformationContext context, St
   @Override
   public ImportSectionView generateImportSection(
       MethodContext context, Iterable<InitCodeNode> specItemNodes) {
-    return generateImportSection(context.getTypeTable().getImports(), null);
+    boolean importIOUtility =
+        Streams.stream(specItemNodes)
+            .anyMatch(node -> node.getLineType() == InitCodeLineType.ReadFileInitLine);
+    ImportSectionView view = generateImportSection(context.getTypeTable().getImports(), null);
```
pongad: Here and below: I think it'd be cleaner to do something like
```
if (importIOUtility) {
  context.getTypeTable().addImports(...);
}
return generateImportSection(...)
```
Was there a specific reason you did it this way?

hzyi-google: `ImportTypeTable` does not have a method called `addImports`?

hzyi-google: I'm calling `context.getTypeTable().getAndSaveNicknameFor(...)` instead, which works in Java, but errored out [here](https://github.com/googleapis/gapic-generator/blob/2c0686b5c51fec23542c5bf8a6abcbd7914ddcc6/src/main/java/com/google/api/codegen/util/py/PythonTypeTable.java#L73) in Python and Node. It seems a dot separated string is expected but we want to import `io` and `fs`. Anything else I can do to make this work?

pongad: That check seems to be there to make sure you don't import unqualified names. I think that function should itself change to allow stdlib imports like this. @lukesneeringer WDYT?

src/main/java/com/google/api/codegen/transformer/StandardImportSectionTransformer.java
```diff
@@ -32,7 +36,21 @@ public ImportSectionView generateImportSection(TransformationContext context, St
   @Override
   public ImportSectionView generateImportSection(
       MethodContext context, Iterable<InitCodeNode> specItemNodes) {
-    return generateImportSection(context.getTypeTable().getImports(), null);
+    boolean importIOUtility =
```
vchudnov-g: Just a nit: in these files, you might communicate your intent more clearly if you s/`importIOUtility`/`needIOUtility`/

Also cosmetic: consider s/`IOUtility`/`FileUtility`/g to avoid the 3 caps in a row (but  totally up to you)

src/main/java/com/google/api/codegen/transformer/StandardImportSectionTransformer.java
vchudnov-g: I agree with the general sentiment of avoiding creating something and then making a builder for it so we can build it again with the final values.

hzyi-google: Actually, this might not work for python at all since the respective transformers are not using the `typeTable` when generating import sections for initcode: [python](https://github.com/googleapis/gapic-generator/blob/1add5a8ba75e29a80613fb2a187a46c08887434a/src/main/java/com/google/api/codegen/transformer/py/PythonImportSectionTransformer.java#L194).

I'm still investigating node, but it seems it should work for node.

hzyi-google:
PTAL @pongad @vchudnov-g

PR #2447, opened 2018-11-20
Calculate defaultResourceNameTreatment

andreamlin:
If the GAPIC method config doesn't provide a ResourceNameTreatment, look at the the proto annotations.
If the proto annotations say that a method's input message has a field that references some Resource[Set], then use `ResourceNameTreatment.STATIC`.

### DIFF
--- a/src/main/java/com/google/api/codegen/config/GapicMethodConfig.java
+++ b/src/main/java/com/google/api/codegen/config/GapicMethodConfig.java
@@ -257,7 +257,10 @@ public abstract class GapicMethodConfig extends MethodConfig {
             .getInputMessage()
             .getFields()
             .stream()
-            .anyMatch(f -> !Strings.isNullOrEmpty(protoParser.getResourceReference(f)))) {
+            .anyMatch(
+                f ->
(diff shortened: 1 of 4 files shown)


andreamlin:
PTAL

src/main/java/com/google/api/codegen/config/GapicMethodConfig.java
```diff
@@ -243,6 +243,40 @@ static GapicMethodConfig createMethodConfig(
...
+            .getInputMessage()
+            .getFields()
+            .stream()
+            .anyMatch(
+                f ->
+                    !Strings.isNullOrEmpty(protoParser.getResourceReference(f))
+                        || !Strings.isNullOrEmpty(protoParser.getResourceOrSetEntityName(f)))) {
+      String methodInputPackageName =
+          protoParser.getProtoPackage(((ProtoFile) method.getInputMessage().getParent()));
+      if (!defaultPackageName.equals(methodInputPackageName)) {
```
pongad: I think it's cleaner to flip the logic so equals comes first.

andreamlin: for sure

PR #2450, opened 2018-11-21
[Proto Annotations] Allow for null configProto; small changes

andreamlin:
- Pass in list of ProtoFiles to parse interfaces from

### DIFF
--- a/src/main/java/com/google/api/codegen/config/GapicProductConfig.java
+++ b/src/main/java/com/google/api/codegen/config/GapicProductConfig.java
@@ -163,20 +163,18 @@ public abstract class GapicProductConfig implements ProductConfig {
             .filter(f -> f.getProto().getPackage().equals(defaultPackage))
             .collect(Collectors.toList());
 
-    if (protoPackage != null) {
-      if (configProto == null) {
-        if (sourceProtos.size() == 0) {
-          model
-              .getDiagReporter()
-              .getDiagCollector()
-              .addDiag(
-                  Diag.error(
-                      SimpleLocation.TOPLEVEL,
-                      "There are no source proto files with package %s",
-                      defaultPackage));
-        }
-        sourceProtos.forEach(model::addRoot);
+    if (protoPackage != null && configProto == null) {
+      if (sourceProtos.isEmpty()) {
+        model
+            .getDiagReporter()
+            .getDiagCollector()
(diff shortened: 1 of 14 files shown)


src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
```diff
@@ -160,7 +160,7 @@ private ViewModel buildGapicClientViewModel(GapicInterfaceContext context) {
     apiImplClass.protoFilename(context.getInterface().getFile().getSimpleName());
     String implName = namer.getApiWrapperClassImplName(context.getInterfaceConfig());
     apiImplClass.name(implName);
-    String serviceAddress = context.getModel().getServiceConfig().getName();
+    String serviceAddress = context.getServiceAddress();
```
andreamlin: This change is to follow the convention set by https://github.com/googleapis/gapic-generator/pull/2328

andreamlin:
PTAL

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -139,31 +140,49 @@ public static GapicProductConfig create(
...
-          .getDiagCollector()
-          .addDiag(
-              Diag.error(
-                  SimpleLocation.TOPLEVEL,
-                  "There are no source proto files with package %s",
-                  defaultPackage));
+
+    if (protoPackage != null) {
+      if (configProto == null) {
+        if (sourceProtos.size() == 0) {
```
pongad: Let's use `isEmpty` instead.

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -139,31 +140,49 @@ public static GapicProductConfig create(
...
-          .getDiagReporter()
-          .getDiagCollector()
-          .addDiag(
-              Diag.error(
-                  SimpleLocation.TOPLEVEL,
-                  "There are no source proto files with package %s",
-                  defaultPackage));
+
+    if (protoPackage != null) {
+      if (configProto == null) {
```
pongad: `if (protoPackage != null && configProto == null)`?

PR #2451, opened 2018-11-22
[Proto Annotations] Generate Java client without a GAPIC config

andreamlin:
Add a baseline test to generate the packaged Java Library test API client. See new `java_library_no_gapic_config.baseline` file for the output.

It seems there is still a ways to go to get to surface parity...


src/main/java/com/google/api/codegen/config/GapicInterfaceConfig.java
```diff
@@ -286,31 +298,6 @@ private static SmokeTestConfig createSmokeTestConfig(
     }
   }
 
-  private static <T> List<T> createMethodConfigs(
```
andreamlin: This method has been moved to a static method in GapicProductConfig, as it is used by both GapicInterfaceConfig and DiscoGapicInterfaceConfig

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -854,4 +867,31 @@ public SingleResourceNameConfig getSingleResourceNameConfig(String entityName) {
     }
     return null;
   }
+
+  public enum GapicConfigPresence {
+    NOT_PROVIDED,
+    PROVIDED
+  }
+
+  static <T> List<T> createMethodConfigs(
```
andreamlin: This method has been moved here from GapicInterfaceConfig

src/test/java/com/google/api/codegen/gapic/GapicTestBase2.java
```diff
@@ -154,7 +154,13 @@ protected boolean suppressDiagnosis() {
       snippetNames.addAll(generator.getInputFileNames());
     }
 
-    String baseline = language.toString().toLowerCase() + "_" + apiName + ".baseline";
+    StringBuilder gapic_config_missing = new StringBuilder();
+    if (gapicConfigFileNames == null || gapicConfigFileNames.length == 0) {
+      gapic_config_missing.append("_no_gapic_config");
+    }
+
+    String baseline =
```
andreamlin: This will create the java_library_no_gapic_config.baseline file!

andreamlin:
PTAL

src/main/java/com/google/api/codegen/config/FlatteningConfig.java
```diff
@@ -124,9 +124,11 @@
     // flattening.
     Map<String, FlatteningConfig> flatteningConfigs = new LinkedHashMap<>();
 
-    flatteningConfigs.putAll(flatteningConfigsFromGapicConfig);
-    // Let flattenings from proto annotations override flattenings from GAPIC config.
+    // Let flattenings from GAPIC config override flattenings from proto annotations.
     flatteningConfigs.putAll(flatteningConfigsFromProtoFile);
+    flatteningConfigs.putAll(flatteningConfigsFromGapicConfig);
```
michaelbausor: Why duplicate this on line 131?

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -90,6 +91,9 @@
   /** Returns the type of transport for the generated client. Defaults to Grpc. */
   public abstract TransportProtocol getTransportProtocol();
 
+  /** Returns the package name. */
```
michaelbausor: Update comment

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -90,6 +91,9 @@
...
+  public abstract GapicConfigPresence isGapicConfigProvided();
```
michaelbausor: Not sure if having the method name start with "is" is appropriate if the return type isn't boolean. Should this return a boolean?

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -843,15 +856,44 @@ public InterfaceConfig getInterfaceConfig(String fullName) {
...
       if (Iterables.size(oneofConfig.getSingleResourceNameConfigs()) > 0) {
         return Iterables.get(oneofConfig.getSingleResourceNameConfigs(), 0);
       }
     }
     return null;
   }
+
+  // Describes if a GAPIC config was given as input. This is a wrapper around checking
+  // `if (ConfigProto == null)`, and letting ConfigProto be non-null internally avoids NPEs.
+  public enum GapicConfigPresence {
```
michaelbausor: Two questions, not sure of correct answer for either:
- Should this have its own file?
- Is an enum necessary? Or could/should this be a boolean?

andreamlin:
PTAL

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -91,8 +91,8 @@
   /** Returns the type of transport for the generated client. Defaults to Grpc. */
   public abstract TransportProtocol getTransportProtocol();
 
-  /** Returns the package name. */
-  public abstract GapicConfigPresence isGapicConfigProvided();
+  /** Returns whether or not the GAPIC config was provided as an input to the generator. */
```
michaelbausor: Nit: "Returns whether or not ..." suggests a boolean. Perhaps "Returns a GapicConfigPresence whose value is determined by whether or not a GAPIC config was provided as an input to the generator."?

src/main/java/com/google/api/codegen/config/FlatteningConfig.java
```diff
@@ -124,9 +124,9 @@
     // flattening.
     Map<String, FlatteningConfig> flatteningConfigs = new LinkedHashMap<>();
 
-    flatteningConfigs.putAll(flatteningConfigsFromGapicConfig);
-    // Let flattenings from proto annotations override flattenings from GAPIC config.
+    // Let flattenings from GAPIC config override flattenings from proto annotations.
```
michaelbausor: Here we are merging the configs from proto annotations and gapic config. But in other places in this PR we are using the new enum to pick one or the other, right? Or is that not the correct way to think about this?

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
andreamlin: Done

src/main/java/com/google/api/codegen/config/FlatteningConfig.java
andreamlin: There's a TODO in [ProtoParser.java](https://github.com/googleapis/gapic-generator/blob/1add5a8ba75e29a80613fb2a187a46c08887434a/src/main/java/com/google/api/codegen/util/ProtoParser.java#L56) to toggle on and off parsing the proto annotations. 

After that is implemented, then `createFlatteningConfigsFromProtoFile()` should return an empty list because the ProtoParser was blind to any MethodSignature annotations in the proto file.

andreamlin:
PTAL

src/main/java/com/google/api/codegen/config/FlatteningConfig.java
michaelbausor: I guess what I am saying is - it isn't clear to me why in some places we have threaded in the GapicConfigPresence enum value, while here we have not. If the reason is because we can be sure that createFlatteningConfigsFromProtoFile will return empty when GapicConfigPresence==PROVIDED, then can we apply that logic elsewhere and not thread the new enum through method signatures? We already have the protoParser available I think.

andreamlin:
@michaelbausor The GapicConfigPresence is merely a wrapper around checking `if (configProto == null)`. When `configProto` is first parsed from the yaml, and if it is null, we turn it into a non-null `ConfigProto.defaultInstance()`. This allows us to not have to do null checks on the `configProto` and its containing `MethodConfigProto`, `InterfaceConfigProto`, etc; we can pass around this empty but non-null object and operate on its empty list of elements and be OK without throwing an NPE. 
This enum for non-presence is required in instances where it _does_ matter if a GAPIC config was given, such as when we should only generate methods that are defined in the GAPIC config, and if an empty InterfaceConfigProto with an empty list of MethodConfigProtos was given, then we would not generate any methods because that's what the GAPIC config said, however this empty InterfaceConfigProto was just really just a shortcut to represent a null InterfaceConfigProto.

Regarding turning off processing proto annotations, that would be a separate concept covered by a separate enum/flag, perhaps `AnnotationsProcessing`.

src/main/java/com/google/api/codegen/config/GapicInterfaceConfig.java
andreamlin: Update: removing this method, but keeping its method overload below.

src/main/java/com/google/api/codegen/config/FlatteningConfig.java
andreamlin: Ok, removed the GapicConfigPresence!

andreamlin:
Ok, the pipeline should now create the list of Methods and Interfaces (clients) for generation in one single place, so we don't have to pass around a isGapicConfigPresent flag everywhere. Thanks a tooon for that suggestion! This is a much cleaner draft, plus it's got a smaller LOC diff for the same semantic changes :)

src/main/java/com/google/api/codegen/config/FlatteningConfig.java
andreamlin: Ok, removed the GapicConfigPresence!

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -224,16 +228,23 @@ public static GapicProductConfig create(
       clientPackageName = settings.getPackageName();
     }
 
+    // Collect the interfaces (clients) and methods that we will generate on the surface.
+    // Not all methods defined in the protofiles will be generated on the surface.
+    ImmutableMap<Interface, InterfaceConfigProto> interfacesMap =
+        createInterfacesMap(
```
andreamlin: Here we determine the ordered set of clients and methods to generate. These sets are then passed down further into the pipeline, so we don't have to pass around a isGapicConfigPresent flag any more.

The invoked methods `createInterfacesMap` and `createMethodsMap` should only ever be called here.

andreamlin:
hmm i broke something in the smoke tests

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -371,7 +386,7 @@ public static GapicProductConfig createDummyInstance(
 
   /** Creates an GapicProductConfig with fixed content. Exposed for testing. */
   @VisibleForTesting
-  public static GapicProductConfig createDummyInstance(
+  private static GapicProductConfig createDummyInstance(
```
michaelbausor: If this can be private, should we remove the `@VisibleForTesting` annotation?

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -442,22 +451,96 @@ public static GapicProductConfig createDummyInstance(
...
+        return null;
+      }
+      interfaceInputs.add(interfaceInput);
+    }
+
+    return interfaceInputs.build();
+  }
+
+  /** Return a data object containing the client entities to be generated. */
+  private static GapicInterfaceInput createInterfaceInput(
```
michaelbausor: Rather than have one method with a bool param, can this be two methods? There is arguably not really any shared code between the two cases, except for creating the builders.

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -392,19 +407,13 @@ public static GapicProductConfig createDummyInstance(
         configSchemaVersion);
   }
 
-  private static ImmutableMap<String, InterfaceConfig> createInterfaceConfigMap(
+  /** Return the list of information about clients to be generated. */
+  private static ImmutableList<GapicInterfaceInput> createInterfaceInputs(
```
michaelbausor: If you agree with my comment below about the `createInterfaceInput` method, then perhaps also this should be two methods? The only shared code is the construction of the `protoInterfaces` map - but that could be pulled out, and passed into each method instead of sourceProtos.

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
andreamlin: hokay

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
andreamlin: done.

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
andreamlin: Hmm, so at this point in the refactoring, the if/else block on `gapicConfigPresent` is one method call per each branch, so there's like 60 LOC that's shared, and only 1 that differs depending on the value of the flag. I'm wary of extracting out the if/else block to the parent caling method because the if/else block is inside a medium-size for-loop, and then that for-loop would have to be moved into it's calling method `GapicProduct.create()` which is already very long.

michaelbausor: Fair enough, although I would argue that lines 436-451, and all of the lines inside the for loop that reference `interfaceConfigProto`, are not really "shared", since they only affect one branch of the if statement. So I wouldn't pull the for loop out, I would create a `createInterfaceInputsWithConfigProto` method that handled the case with gapicConfigPresent==true, and modify `createInterfaceInputs` to handle only the false case, and have the shared `protoInterfaces` variable be an input instead of `sourceProtos`. But I also feel like this is less important that other comments, so feel free to disregard if you disagree.

michaelbausor: (or, keep passing in `sourceProtos` but move creation of `protoInterfaces` variable into its own method and share it between the two methods)

andreamlin: Ok, moved the switching logic into GapicProductConfig.create(), so we can remove the `configPresence` flag altogether, assuming an empty GAPIC config is the same thing as a null one.

andreamlin:
PTAL

michaelbausor:
I'm looking at the tests you added, particularly the new baseline file. I don't really know how to assess it (it is >5000 lines...) - should I be comparing it to the existing baselines in Java? Should it be the same or very similar to them?

andreamlin:
@michaelbausor gooood question. I opened https://github.com/googleapis/gapic-generator/pull/2462 to just copy over the java_library.baseline that's generated from the GAPIC-yaml-only pipeline into this new baseline file.

After that gets merged in, this PR will show the diff between what happens when we use GAPIC-yaml-only and when we use proto-annotations-only.

andreamlin:
Ok, the diff on the new baseline file is now visible. From what I see, there's two major differences that have to be fixed later on: 
- The order of methods in the annotation-only output is sorted alphabetically, whereas the GAPIC config-only surface mostly follows the order specified in the GAPIC config, which happens to follow the order in the proto files.
- The package and client names of the annotations-only output hasn't been overriden by config values yet.

PTAL

andreamlin:
No, it doesn't have to be sorted alphabetically, and I intend to fix the surface ordering to reflect the proto file order in the near future.

src/main/java/com/google/api/codegen/config/GapicInterfaceConfig.java
```diff
@@ -217,50 +220,17 @@ private static SmokeTestConfig createSmokeTestConfig(
...
       String defaultPackageName,
-      InterfaceConfigProto interfaceConfigProto,
-      Interface apiInterface,
+      Map<Method, MethodConfigProto> methodsToGenerate,
       ResourceNameMessageConfigs messageConfigs,
       ImmutableMap<String, ResourceNameConfig> resourceNameConfigs,
       RetryCodesConfig retryCodesConfig,
       ImmutableSet<String> retryParamsConfigNames,
       ProtoParser protoParser) {
-    Map<String, GapicMethodConfig> methodConfigMapBuilder = new TreeMap<>();
```
andreamlin: This structure, which contains the ordered list of methods, was accidentally a TreeMap (alphabetically ordered by name) instead of a LinkedHashMap (insertion order).

Fixing that, the method order (which follows the declarations in the protofile) is now as expected.

andreamlin:
Fixed the surface order of methods. (The baseline diff is still quite large...)
PTAL

Issue #2422, opened 2018-11-13
[Python] Replacement not occurring in docstring for apiMethod.name


crwilcox:
```
scheduler/google/cloud/scheduler_v1beta1/gapic/transports/cloud_scheduler_grpc_transport.py
     @property
    def list_jobs(self):
        """Return the gRPC stub for {$apiMethod.name}.
```

We noticed an issue with replacements in python docstrings.

andreamlin:
@crwilcox Would you mind linking to a sample PR for google-cloud-python where this happens?

andreamlin:
@crwilcox Actually I realized there is linked PR, thanks :)

andreamlin:
fix on the way

PR #2452, opened 2018-11-26
Fix Python snippet bug

andreamlin:
Fixes https://github.com/googleapis/gapic-generator/issues/2422.


andreamlin:
PTAL

landrito:
I think the correct way to document this is like: 

```
:meth:`method_name`
```

andreamlin:
@landrito ok, using the Sphinx syntax for method referencing. Also qualifying the method reference by the class that contains the method
PTAL

andreamlin:
@crwilcox would you mind reviewing this too?

andreamlin:
The Ruby test has been chronically failing https://github.com/googleapis/gapic-generator/issues/2458

andreamlin:
PTAL

landrito:
Bah sorry, since the transport should be agnostic to the client that is using it, I think simply using the method name without the fancy linking syntax is best.

PR #2459, opened 2018-11-28
go: use standard notation for marking generated files

pongad:
See: https://golang.org/s/generatedcode

Fixes googleapis/artman#533.


pongad:
No probs!

PR #2462, opened 2018-11-29
Copy over java_library.baseline to protoannotations testdata

andreamlin:
Copy over the existing java_library.baseline (baseline file for legacy GAPIC yaml only generation) into the java_library_no_gapic_config.baseline file that will be generated by https://github.com/googleapis/gapic-generator/pull/2451/.

Also update the annotated library.proto file to reflect the recent changes in the un-annotated library.proto.  (Note: We should stop needing to copy over changes from the un-annotated library.proto to the annotated.proto once there's a toggle for turning off parsing proto annotations.)


andreamlin:
PTAL

PR #2466, opened 2018-12-04
Remove most of LanguageTransformer

andreamlin:
There's a lot of duplicate logic in LanguageTransformer and LanguageSettingsMerger. Rip out everything in LanguageTransformer that already exists in LanguageSettingsMerger.

Ideally we'd remove all duplicate logic between the old Transformers and the ConfigNodeMerger implementations, but this is an easy first step.


andreamlin:
PTAL

PR #2467, opened 2018-12-04
[Proto Annotations] Better heuristics for defaultResourceNameTreatment

andreamlin:
For fields in flattened methods, set the default ResourceNameTreatment according to the logic in `GapicMethodConfig.defaultResourceNameTreatment`.

See changes in the java_library_no_gapic_config.baseline file, which now has more of the flattened methods that are present in the GAPIC-config-generated java_library.baseline.

### DIFF
--- a/src/main/java/com/google/api/codegen/config/FlatteningConfig.java
+++ b/src/main/java/com/google/api/codegen/config/FlatteningConfig.java
@@ -26,15 +26,18 @@ import com.google.api.tools.framework.model.Oneof;
 import com.google.api.tools.framework.model.SimpleLocation;
 import com.google.auto.value.AutoValue;
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 
 /** FlatteningConfig represents a specific flattening configuration for a method. */
@@ -70,7 +73,8 @@ public abstract class FlatteningConfig {
               flatteningGroup,
               methodModel);
       if (groupConfig != null) {
-        flatteningConfigs.put(flatteningConfigToString(groupConfig), groupConfig);
+        String key = flatteningConfigToString(groupConfig);
+        flatteningConfigs.put(key, groupConfig);
       }
     }
     if (diagCollector.hasErrors()) {
@@ -99,6 +103,7 @@ public abstract class FlatteningConfig {
   @Nullable
   static ImmutableList<FlatteningConfig> createFlatteningConfigs(
       DiagCollector diagCollector,
+      String defaultPackageName,
       ResourceNameMessageConfigs messageConfigs,
       ImmutableMap<String, ResourceNameConfig> resourceNameConfigs,
       MethodConfigProto methodConfigProto,
@@ -115,7 +120,13 @@ public abstract class FlatteningConfig {
     // Get flattenings from protofile annotations
     Map<String, FlatteningConfig> flatteningConfigsFromProtoFile =
         createFlatteningConfigsFromProtoFile(
-            diagCollector, messageConfigs, resourceNameConfigs, methodModel, protoParser);
+            diagCollector,
+            defaultPackageName,
+            messageConfigs,
+            resourceNameConfigs,
+            methodConfigProto,
+            methodModel,
+            protoParser);
     if (flatteningConfigsFromProtoFile == null) {
       return null;
     }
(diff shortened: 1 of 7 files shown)


src/test/java/com/google/api/codegen/protoannotations/testdata/java_library_no_gapic_config.baseline
```diff
@@ -318,6 +341,31 @@ public class LibraryServiceClient implements BackgroundResource {
...
+   * try (LibraryServiceClient libraryServiceClient = LibraryServiceClient.create()) {
+   *   Shelf shelf = Shelf.newBuilder().build();
+   *   Shelf response = libraryServiceClient.createShelf(shelf);
+   * }
+   * </code></pre>
+   *
+   * @param shelf The shelf to create.
+   * @throws com.google.api.gax.rpc.ApiException if the remote call fails
+   */
+  public final Shelf createShelf(Shelf shelf) {
```
andreamlin: This generated method, along with other flattened methods that use a STATIC_TYPES param, also exists in the GAPIC-config-generated java_library.baseline.

src/test/java/com/google/api/codegen/protoannotations/testdata/java_library_no_gapic_config.baseline
```diff
@@ -410,6 +620,28 @@ public class LibraryServiceClient implements BackgroundResource {
...
+   * Sample code:
+   * <pre><code>
+   * try (LibraryServiceClient libraryServiceClient = LibraryServiceClient.create()) {
+   *
+   *   ListShelvesResponse response = libraryServiceClient.listShelves();
+   * }
+   * </code></pre>
+   * @throws com.google.api.gax.rpc.ApiException if the remote call fails
+   */
+  public final ListShelvesResponse listShelves() {
```
andreamlin: This method is now generated, but it *should* be a paged method. I'll do paged method generation in a separate PR.

src/test/java/com/google/api/codegen/protoannotations/testdata/library.proto
```diff
@@ -1017,39 +1014,39 @@ message TestOptionalRequiredFlatteningParamsRequest {
...
-    repeated google.protobuf.Int32Value repeated_int32_value = 127;
-    repeated google.protobuf.UInt32Value repeated_uint32_value = 128;
-    repeated google.protobuf.Int64Value repeated_int64_value = 129;
-    repeated google.protobuf.UInt64Value repeated_uint64_value = 130;
-    repeated google.protobuf.FloatValue repeated_float_value = 131;
-    repeated google.protobuf.DoubleValue repeated_double_value = 132;
-    repeated google.protobuf.StringValue repeated_string_value = 133;
-    repeated google.protobuf.BoolValue repeated_bool_value = 134;
-    repeated google.protobuf.BytesValue repeated_bytes_value = 135;
+  google.protobuf.Any any_value = 100;
```
andreamlin: removed extra indent here.

src/test/java/com/google/api/codegen/protoannotations/testdata/library.proto
```diff
@@ -251,9 +251,6 @@ service LibraryService {
   // Test bidi-streaming.
   // gRPC streaming methods don't have an HTTP equivalent and don't need to have the google.api.http option.
   rpc DiscussBook(stream DiscussBookRequest) returns (stream Comment) {
-    option (google.api.method_signature) = {
```
andreamlin: This method signature didn't exist in the library_gapic.yaml, so it's getting removed.

src/test/java/com/google/api/codegen/protoannotations/testdata/library.yaml
```diff
@@ -1,6 +1,6 @@
 type: google.api.Service
 config_version: 1
-name: library-example.googleapis.com
+name: library-example.googleapis.com:1234
```
andreamlin: This reflects the main library.yaml service yaml used by all the other baseline tests.

andreamlin:
PTAL

src/main/java/com/google/api/codegen/config/GapicMethodConfig.java
```diff
@@ -277,6 +271,17 @@ static ResourceNameTreatment defaultResourceNameTreatment(
     return defaultResourceNameTreatment;
   }
 
+  @VisibleForTesting
```
michaelbausor: Are you testing this directly? If not, we can remove this annotation.

src/main/java/com/google/api/codegen/config/FlatteningConfig.java
```diff
@@ -138,31 +149,49 @@
   @Nullable
   private static Map<String, FlatteningConfig> createFlatteningConfigsFromProtoFile(
```
michaelbausor: The name of this method suggests it only looks at the proto file, but now we are mixing in the gapic config. Can we rename? And also, can we clarify (either in method comments or inline comments somewhere) exactly what behaviour we expect from the different cases of gapic config being either present or missing? E.g. if we have a gapic config, should we ignore all proto annotation flattenings? If not, why not?

src/main/java/com/google/api/codegen/config/GapicMethodConfig.java
andreamlin: Yes, this is being tested a couple times in [ResourceNameMessageConfigsTest.java](https://github.com/googleapis/gapic-generator/blob/3f92b3c11fe4017e819a262a66ce4d6120dba94b/src/test/java/com/google/api/codegen/config/ResourceNameMessageConfigsTest.java#L469)

src/main/java/com/google/api/codegen/config/FlatteningConfig.java
andreamlin: There is no difference in behavior, for this method, if the GAPIC config is missing or not.

Ok, updated the javadoc comment to answer those questions. Hows' that look?

andreamlin:
PTAL

src/main/java/com/google/api/codegen/config/GapicMethodConfig.java
michaelbausor: Ah ok. So that means that the new overload (on line 244) is not being tested, and can be private?

src/main/java/com/google/api/codegen/config/FlatteningConfig.java
michaelbausor: That looks good. But I also wanted to clarify that the list of flattenings generated will include flattenings from *both* the gapic config and the proto annotations. Is that right? And is that what we want to happen?

andreamlin: ok this is the last straw, i'll draft up a proposal for handling both config specs :)

PR #2468, opened 2018-12-04
[Proto Annotations] Expose generator option for client package name

andreamlin:
Allow client package names to be customizable through a command-line option.

Custom package names were previously only definable in the [language-specific settings](https://github.com/googleapis/gapic-generator/blob/49c2677021a2447cf3a68ad1dcdfce8adbb134d5/src/main/proto/com/google/api/codegen/config.proto#L103) of the GAPIC config. When we don't have a GAPIC config, we have to be able to override generated package names some other way.

The baseline file `java_library_no_gapic_config.baseline`, generated _without_ a GAPIC config, now uses the `com.google.example.library.v1` package as given in the [test method call](https://github.com/googleapis/gapic-generator/pull/2468/files#diff-8eb27f70b7cf7e6151268a56b248ba1dR81), parallel to the library_gapic.yaml's [java package override](https://github.com/googleapis/gapic-generator/blob/49c2677021a2447cf3a68ad1dcdfce8adbb134d5/src/test/java/com/google/api/codegen/testsrc/libraryproto/library_gapic.yaml#L5).

### DIFF
--- a/src/main/java/com/google/api/codegen/config/GapicProductConfig.java
+++ b/src/main/java/com/google/api/codegen/config/GapicProductConfig.java
@@ -135,7 +135,7 @@ public abstract class GapicProductConfig implements ProductConfig {
   @Nullable
   public static GapicProductConfig create(
       Model model, ConfigProto configProto, TargetLanguage language) {
-    return create(model, configProto, null, language);
+    return create(model, configProto, null, null, language);
   }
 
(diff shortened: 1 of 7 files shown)


andreamlin:
PTAL

andreamlin:
@lukesneeringer friendly ping

PR #2472, opened 2018-12-10
Update PHP gapic class import to correct incorrect name

jdpedrie:
The GAPIC Generator is referencing a class `Google\ApiCore\FetchAuthTokenInterface` which doesn't exist. It should be referencing `Google\Auth\FetchAuthTokenInterface`.

This change updates all the references to that class and reorders imports to maintain alphabetization.


andreamlin:
@michaelbausor can you PTAL?

jdpedrie:
Yeah, it's needed for doc generation. The alternative is referencing the fully-qualified name inline, but that is messier.

jdpedrie:
@andreamlin yep!

PR #2475, opened 2018-12-10
Order schema fields

andreamlin:
Discovery doc messages' fields should preserve the order of fields that are defined in by the Discovery Document.


andreamlin:
PTAL

PR #2476, opened 2018-12-10
[Discogapic] Better configgen package names

andreamlin:
Make Discovery-based bootstrapped GAPIC config have better generated package names.

See https://github.com/googleapis/googleapis/blob/master/google/cloud/language/v1/language_gapic.yaml for an example of standard generated package names.

Impetus for this: whenever I regen compute_gapic.yaml with configgen ([example](https://github.com/googleapis/discovery-artifact-manager/pull/101/files)), the java package name always [needs to be manually-reverted](https://github.com/googleapis/discovery-artifact-manager/pull/101/files#diff-f707e72f8271e8454e8c6e9388f8bee5R11).

### DIFF
--- a/src/test/java/com/google/api/codegen/discogapic/testdata/simplecompute_config.baseline
+++ b/src/test/java/com/google/api/codegen/discogapic/testdata/simplecompute_config.baseline
@@ -8,17 +8,17 @@ config_schema_version: 1.0.0
 # The settings of generated code in a specific language.
 language_settings:
   java:
-    package_name: com.google.simplecompute.v1
+    package_name: com.google.cloud.simplecompute.v1
   python:
-    package_name: com.google.simplecompute_v1.gapic
+    package_name: google.cloud.simplecompute_v1.gapic
   go:
-    package_name: google.golang.org/com/google/simplecompute/v1
+    package_name: cloud.google.com/go/simplecompute/apiv1
   csharp:
-    package_name: Com.Google.Simplecompute.V1
+    package_name: Google.Simplecompute.V1
   ruby:
-    package_name: Com::Google::Simplecompute::V1
+    package_name: Google::Cloud::Simplecompute::V1
   php:
-    package_name: Com\Google\Simplecompute\V1
+    package_name: Google\Cloud\Simplecompute\V1
   nodejs:
     package_name: simplecompute.v1
 # The configuration for the license header to put on generated files.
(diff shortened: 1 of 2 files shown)


andreamlin:
PTAL

src/test/java/com/google/api/codegen/discogapic/testdata/simplecompute_config.baseline
```diff
@@ -8,17 +8,17 @@ config_schema_version: 1.0.0
...
 language_settings:
   java:
-    package_name: com.google.simplecompute.v1
+    package_name: com.google.cloud.simplecompute.v1
   python:
-    package_name: com.google.simplecompute_v1.gapic
+    package_name: google.cloud.simplecompute_v1.gapic
   go:
-    package_name: google.golang.org/com/google/simplecompute/v1
+    package_name: cloud.google.com/go/simplecompute/apiv1
```
vchudnov-g: For my own understanding, could you clarify which substrings in these changed lines correspond to `org`, `model.name()` and `model.version()`?

I ask because some of these (like for Go) seem like pretty drastic changes. Are these all the correct, intended outputs?

vchudnov-g: How can `org` contain a period, given your changes to `getPackageName`?

I assume there's  language-specific tooling to add the missing substrings, like `google.com` and `api` in `apiv1`. Is that correct, and am I correct in concluding this tooling assumes google-specifc libraries? (If so, eventually we'll want to change this to allow third-party APIs to generate gapics)

andreamlin: oh you're right, I guess `org` would just be `"google"`.

The `"cloud"` string gets automatically inserted into language-specific package namers because this is GCP-centric. Practically speaking, I don't envision any non-Google API producers to use Discovery Doc, so supporting non-Google here is not a pressing concern.

andreamlin:
PTAL

andreamlin:
PTAL

src/test/java/com/google/api/codegen/discogapic/testdata/simplecompute_config.baseline
vchudnov-g: yeah, you're right we don't need to worry about non-Google Discovery. We will have to worry about non-Google gRPC, but this PR is Discovery-specific.

PR #2477, opened 2018-12-10
[Discogapic] Parse ResourceName when it contains a service address

andreamlin:
This would fix googleapis/google-cloud-java#3604.

_Problem_: When a resource comes back from the server, its resource path is something like 
```
instance.zone = https://www.googleapis.com/compute/v1/projects/project-123/zones/europe-west3-c
```
And that string won't be parseable against a PathTemplate of form `{project}/zones/{zone}`.

_Proposed solution_: Optionally ignore the leading [service address](https://github.com/googleapis/google-cloud-java/blob/969bbeef18f004fd51fd46c5def1ae5c644cae3c/google-cloud-clients/google-cloud-compute/src/main/java/com/google/cloud/compute/v1/stub/ProjectStubSettings.java#L237), which for Compute is `https://www.googleapis.com/compute/v1/projects/`.

See https://github.com/googleapis/google-cloud-java/pull/4208 for resulting changes.


andreamlin:
`ci/circleci: compute-discovery-test` is going to fail because  https://github.com/googleapis/discovery-artifact-manager/pull/102 isn't merged yet.

andreamlin:
PTAL

PR #2478, opened 2018-12-11
Fix compute resource parsing

andreamlin:


andreamlin:
PTAL

andreamlin:
PTAL

andreamlin:
@pongad  PTAL

PR #2479, opened 2018-12-12
[golang] Mark "beta" APIs as beta

andreamlin:
Fixes https://github.com/googleapis/artman/issues/534.

Infer that APIs with a "beta" in their version in their go package are actually Beta version, for the purposes of documentation.

See https://github.com/googleapis/google-cloud-go/pull/1255 for an example of generating cloudtasks v2beta3 with this PR.


andreamlin:
PTAL

andreamlin:
PTAL @pongad 

cc/ @jadekler

src/main/java/com/google/api/codegen/transformer/go/GoGapicSurfaceTransformer.java
```diff
@@ -79,6 +80,8 @@
 
   private static final int COMMENT_LINE_LENGTH = 75;
 
+  private static final Pattern versionPattern = Pattern.compile("v\\d+(p\\d)?beta");
```
jadekler: Would it make sense to just do something like strings.Contains(str, "beta")? Since the \d+ is optional, it's probably OK to just remove.

WDYT?

jadekler:
Thanks Andrea! I'll leave Michael to Approve since I'm clueless about a lot of what's happening in this repo :)

src/main/java/com/google/api/codegen/transformer/go/GoGapicSurfaceTransformer.java
andreamlin: Since google APIs follow the versioning scheme defined in https://goto.google.com/aip/0201, I figured we could be extra certain that a package name contains a version string representing beta if we matched against those guidelines.

jadekler: Gotcha, thanks for the context. I'll leave that call to y'all!

pongad: I'd just do `v\\d.*beta`. I think the current pattern will have trouble with `v1p11beta1`.

andreamlin: haha I never get these regex patterns right the first time
Done.

andreamlin:
PTAL

PR #2482, opened 2018-12-13
Samplegen: print enums in PHP

hzyi-google:


hzyi-google:
PTAL @vchudnov-g @michaelbausor @beccasaurus

michaelbausor:
Using the new `::name` methods on enums is definitely what we want to do - but there isn't a released version of protobuf that includes them yet, so we will need to wait for protobuf 3.7 I think.

hzyi-google:
@michaelbausor Can we just leave this PR open until protobuf 3.7 is out?

michaelbausor:
Yes, we can. I don't know what time frame is expected for protobuf 3.7, but I think it is coming fairly soon.

vchudnov-g:
@michaelbausor Do you have any sense on what "soon" is? Early January? EOQ1?

michaelbausor:
The current latest I heard from them is ~early January.

beccasaurus:
From the Protobuf issue:

> 3.7.0rc2 should already have this [`*`](https://github.com/protocolbuffers/protobuf/pull/5342#issuecomment-464773459)

Do our client libraries (PHP) use `rc` versions of Protobuf? I'm guessing not, but want to check before asking Protobuf if they know when `3.7` stable will be released.

Looks like this'll work now, but only using the latest `rc2`

michaelbausor:
@beccasaurus Protobuf 3.7 is out! https://github.com/protocolbuffers/protobuf/releases/tag/v3.7.0

hzyi-google:
Ah thanks! This PR has become a bit outdated though, I'm gonna close it here and create a new one.

PR #2486, opened 2018-12-18
Remove grpc_streaming

andreamlin:
What happens if we remove `grpc_streaming` config?

### DIFF
--- a/src/test/java/com/google/api/codegen/gapic/testdata/php/php_library.baseline
+++ b/src/test/java/com/google/api/codegen/gapic/testdata/php/php_library.baseline
@@ -5222,20 +5222,11 @@ class LibraryServiceClientTest extends GeneratedTest
         $this->assertTrue($transport->isExhausted());
 
         // Mock response
-        $shelvesElement = new Shelf();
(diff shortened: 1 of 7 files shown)


src/test/java/com/google/api/codegen/gapic/testdata/php/php_library.baseline
```diff
@@ -2598,7 +2598,6 @@ return [
             'StreamShelves' => [
                 'grpcStreaming' => [
                     'grpcStreamingType' => 'ServerStreaming',
-                    'resourcesGetMethod' => 'getShelves',
```
andreamlin: @michaelbausor Is this a breaking change?

michaelbausor: Yes. When this setting is present, we use the specified method (in this case `getShelves`) to give the user a stream of `Shelf` objects. When it is not present, we just give them a stream of `StreamShelvesResponse` objects.

andreamlin: ok, thanks!

Issue #2481, opened 2018-12-12
Use timeout value in config


andreamlin:
We never use the value of [`timeout_millis`](https://github.com/googleapis/gapic-generator/blob/6f888379edfdd5735fe7cbf65880dcf7d015f884/src/main/proto/com/google/api/codegen/config.proto#L251) from the GAPIC config. Nothing ever calls the derived [`MethodConfig#getTimeout`](https://github.com/googleapis/gapic-generator/blob/6f888379edfdd5735fe7cbf65880dcf7d015f884/src/main/java/com/google/api/codegen/config/MethodConfig.java#L53)

We should use this value in the generated surface.

hzyi-google:
Obsolete as Gapic config is deprecated.

PR #2487, opened 2018-12-18
timeout_millis is unsupported

andreamlin:
We never actually process the `timeout_millis` config value. Note that in the comments.

See https://github.com/googleapis/gapic-generator/issues/2481.


andreamlin:
PTAL

PR #2491, opened 2018-12-27
[Proto Annotations] Parse routing headers

andreamlin:
Parse out the header request parameter from the google.api.http annotation.


src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
```diff
@@ -11410,40 +11410,53 @@ namespace Google.Example.Library.V1
...
-                GrpcClient.DeleteShelfAsync, GrpcClient.DeleteShelf, effectiveSettings.DeleteShelfSettings);
+                GrpcClient.DeleteShelfAsync, GrpcClient.DeleteShelf, effectiveSettings.DeleteShelfSettings)
+                .WithCallSettingsOverlay(request => gaxgrpc::CallSettings.FromHeader("x-goog-request-params", $"name={request.Name}"));
             _callMergeShelves = clientHelper.BuildApiCall<MergeShelvesRequest, Shelf>(
-                GrpcClient.MergeShelvesAsync, GrpcClient.MergeShelves, effectiveSettings.MergeShelvesSettings);
+                GrpcClient.MergeShelvesAsync, GrpcClient.MergeShelves, effectiveSettings.MergeShelvesSettings)
+                .WithCallSettingsOverlay(request => gaxgrpc::CallSettings.FromHeader("x-goog-request-params", $"name={request.Name}"));
             _callCreateBook = clientHelper.BuildApiCall<CreateBookRequest, Book>(
                 GrpcClient.CreateBookAsync, GrpcClient.CreateBook, effectiveSettings.CreateBookSettings)
                 .WithCallSettingsOverlay(request => gaxgrpc::CallSettings.FromHeader("x-goog-request-params", $"name={request.Name}&book.read={request.Book.Read}"));
```
andreamlin: This was configured using the GAPIC yaml.

andreamlin:
@lukesneeringer Are we fine with all the additional surface changes resulting from the default behavior?

andreamlin:
Blocked on https://github.com/googleapis/gapic-generator/pull/2507

andreamlin:
We should check in https://github.com/googleapis/gapic-generator/pull/2524 first, to decrease the baseline diff of this PR.

andreamlin:
@vam-google PTAL

src/main/java/com/google/api/codegen/util/ProtoParser.java
```diff
@@ -86,6 +88,57 @@ public ProtoParser(boolean enableProtoAnnotations) {
...
+    // No need to recurse on additional bindings' additional bindings.
+    for (HttpRule rule : topRule.getAdditionalBindingsList()) {
+      getHeaderParam(rule, allParams);
+    }
+
+    return allParams.build().asList();
+  }
+
+  // Finds the header param from a HttpRule and add the non-null value to the running set.
+  private void getHeaderParam(HttpRule httpRule, ImmutableSet.Builder<String> runningList) {
```
vam-google: `runningList` is a set. We should be rigorous here: `Set` and `List` are different data structures:
- Set: ensures uniqueness of its elements (calls `equals()` and in case of HashSet `hashCode()` of each inserted object); does not define iteration order;
- List: allows duplicates (`equals` and `hashCode` are not called); defines order of elements

Looking into how this argument is initialized, pupulated and returned (using `asList()`) I'm really confused what we really need here. Do we need just a collection, a set, a list or a list with unique elements in it (then we need to define what happens in case if a duplicate is added: the new item replaces the old one or the old one is kept)?

src/main/java/com/google/api/codegen/util/ProtoParser.java
```diff
@@ -86,6 +88,57 @@ public ProtoParser(boolean enableProtoAnnotations) {
     }
   }
 
+  @Nullable
+  /* Return the name of the field representing the header parameter. */
+  public List<String> getHeaderParams(Method method) {
```
vam-google: Please check the commend on L117. But in general, do we really need a list here? 

Also, can this method return just an empty list instead of `null` if nothing is found? If we do not need to distinguish between emtpy list and `null` it makes usage of this method easier (since caller does not have to handle `null`s, like it is done [here](https://github.com/googleapis/gapic-generator/pull/2491/files#diff-a81b001e7c99214072d01c43e97184bbR285).

src/main/java/com/google/api/codegen/util/ProtoParser.java
andreamlin: We need 
- uniqueness of elements (a header param name can be repeated across different HttpRules, but we only need one of each name)
- a List return object, because that is what the calling code requires. Also, having a deterministic and easy-to-reason-about order is nice for refreshing clients and testing, so we don't have diffs with no semantic changes.

Converting an ImmutableSet to a List satisfies all those conditions.

src/main/java/com/google/api/codegen/util/ProtoParser.java
andreamlin: Ok, returning an empty collection (changed to explicit ImmutableSet)

src/main/java/com/google/api/codegen/util/ProtoParser.java
```diff
@@ -86,6 +88,57 @@ public ProtoParser(boolean enableProtoAnnotations) {
...
+      urlVar = httpRule.getPut();
+    } else {
+      return;
+    }
+
+    PathTemplate pathTemplate = PathTemplate.create(urlVar);
+    String var = pathTemplate.singleVar();
+
+    if (!Strings.isNullOrEmpty(var)) {
+      runningList.add(var);
```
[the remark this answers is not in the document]
andreamlin: Thanks for calling this out :)

src/main/java/com/google/api/codegen/config/GapicMethodConfig.java
```diff
@@ -272,6 +272,23 @@ static GapicMethodConfig createMethodConfig(
...
+  private static List<String> findHeaderRequestParams(
+      MethodConfigProto methodConfigProto, Method method, ProtoParser protoParser) {
+    List<String> headerRequestParams =
+        ImmutableList.copyOf(methodConfigProto.getHeaderRequestParamsList());
+
+    if (!headerRequestParams.isEmpty()) {
+      return headerRequestParams;
+    } else {
+      // Fetch header params from the proto annotations.
+      List<String> headerParams = protoParser.getHeaderParams(method);
```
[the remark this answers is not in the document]
andreamlin: Done.

src/main/java/com/google/api/codegen/config/GapicMethodConfig.java
```diff
@@ -272,6 +272,23 @@ static GapicMethodConfig createMethodConfig(
...
   }
 
+  private static List<String> findHeaderRequestParams(
+      MethodConfigProto methodConfigProto, Method method, ProtoParser protoParser) {
+    List<String> headerRequestParams =
+        ImmutableList.copyOf(methodConfigProto.getHeaderRequestParamsList());
+
+    if (!headerRequestParams.isEmpty()) {
+      return headerRequestParams;
+    } else {
```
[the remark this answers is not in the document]
andreamlin: yep

andreamlin:
Fixed code according to comments. PTAL

src/main/java/com/google/api/codegen/util/ProtoParser.java
```diff
@@ -86,6 +89,51 @@ public ProtoParser(boolean enableProtoAnnotations) {
...
+    // No need to recurse on additional bindings' additional bindings.
+    for (HttpRule rule : topRule.getAdditionalBindingsList()) {
+      getHeaderParam(rule).ifPresent(allParams::add);
+    }
+
+    return allParams.build();
+  }
+
+  // Finds the header param from a HttpRule and add the non-null value to the running set.
+  private Optional<String> getHeaderParam(HttpRule httpRule) {
```
vam-google: Minor, but there is actually nothing wrong in returning `null` especially for the private method. For public method, especially defined on the surface of the library, using `Optional` maybe the right choice, but here `null` is just fine (and technically is slightly more efficient).

src/main/java/com/google/api/codegen/util/ProtoParser.java
```diff
@@ -86,6 +89,51 @@ public ProtoParser(boolean enableProtoAnnotations) {
     }
   }
 
+  /* Return the name of the field representing the header parameter. */
+  public ImmutableSet<String> getHeaderParams(Method method) {
```
vam-google: Minor: would just specifying method signature as returning `Set<String>` (when the actual returned object will still be `ImmutableSet<String>`) be enought here? The best practice is to always try to declare in the method signature the most generic ("generic" in its regular meaning, I do not mean java generics here) input/return type(s) (but the one which still "fit"). The more generic the declaration is the less commitment the implementor/maintainer of the method makes, thus it becomes easier to modify/update the method in the future (though that should not affect the usability of the method, otherwise everythign will be returning and accepting `Object` type).

src/main/java/com/google/api/codegen/config/GapicMethodConfig.java
```diff
@@ -272,6 +272,18 @@ static GapicMethodConfig createMethodConfig(
     }
   }
 
+  private static List<String> findHeaderRequestParams(
```
vam-google: This  method returns set converted to list. Maybe it should just return set? Or maybe just a `Collection` is good enough?

src/main/java/com/google/api/codegen/util/ProtoParser.java
andreamlin: Ahh, I just didn't feel like doing a null check on a `@Nullable` return value.

I just read https://homes.cs.washington.edu/~mernst/advice/nothing-is-better-than-optional.html, which says that `Optional` is worse than using `@Nullable`, so I'll revert.

src/main/java/com/google/api/codegen/util/ProtoParser.java
andreamlin: `Set<>` isn't guaranteed to any order; I want to assure the method callers that the return type has insertion-order, which is what the `ImmutableSet` type conveys.

src/main/java/com/google/api/codegen/config/GapicMethodConfig.java
andreamlin: I want to guarantee method callers that the Collection object has a defined order.

PR #2494, opened 2019-01-03
Properly handle quoted strings in the sample default configuration values

hzyi-google:
If RHS is a string literal, return the unquoted string. Otherwise, return the RHS as it is for backward compatibility.

### DIFF
--- a/src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
+++ b/src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
@@ -99,13 +99,12 @@ public class FieldStructureParser {
 
     InitValue initValue = null;
     if (token == '=') {
-      fieldNamePos = Math.min(fieldNamePos, scanner.pos() - 1);
-
+      // fieldNamePos = Math.min(fieldNamePos, scanner.pos() - 1);
       // TODO(pongad): Quote the RHS of existing configs, and once that's done use 'parseValue' here
       // (`String valueString = parseValue(scanner)`). For now we are preserving the previous
       // behavior, where everything on the right of the equal sign is a string.
-      String valueString = config.substring(scanner.pos());
-
+      // String valueString = config.substring(scanner.pos());
+      String valueString = parseValue(scanner);
       if (valueString.contains(InitFieldConfig.RANDOM_TOKEN)) {
         initValue = InitValue.createRandom(valueString);
       } else if (valueString.contains(PROJECT_ID_TOKEN)) {
--- a/src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
+++ b/src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
@@ -191,9 +191,9 @@ public class FieldStructureParser {
   }
 
   /**
-   * Parses the value of configs. If the value is a double-quoted string literal we return it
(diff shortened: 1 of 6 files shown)


hzyi-google:
PTAL @vchudnov-g

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
```diff
@@ -190,12 +181,28 @@ public static InitCodeNode parsePath(InitCodeNode root, Scanner scanner) {
...
     int token = scanner.scan();
     Preconditions.checkArgument(
         token == Scanner.INT || token == Scanner.IDENT || token == Scanner.STRING,
         "invalid value: %s",
         scanner.input());
     return scanner.tokenStr();
   }
+
+  /**
+   * Parses the RHS of configs. If the RHS is a string literal we return the unquoted string.
```
vchudnov-g: I would clarify this comment: `... If the RHS is a double-quoted string literal, we return it unquoted. Otherwise, ....`

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
```diff
@@ -190,12 +181,28 @@ public static InitCodeNode parsePath(InitCodeNode root, Scanner scanner) {
...
+   * Otherwise we strip leading spaces and return the rest of RHS as a string for backward
+   * compatibility.
+   */
+  private static String parseValue(Scanner scanner) {
+    int token = scanner.scan();
+    if (token == Scanner.STRING) {
```
vchudnov-g: I would add a comment here reminding readers that `Scanner.STRING` means a double-quoted string literal.

src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
vchudnov-g: It may just be me, but I find "RHS" somewhat confusing because in the config, this is already inside a yaml RHS. We need to clarify that this is the RHS relative to the `=`, not relative to the yaml `:`.

How about: `Parses the value of the InitCodeNode configuration (ie the RHS of the '=')`

src/test/java/com/google/api/codegen/testsrc/libraryproto/library_gapic.yaml
```diff
@@ -577,8 +577,8 @@ interfaces:
       parameters:
         defaults:
         - name=BASIC
-        - comment.comment=comment_file
-        - image=image_file.jpg
+        - comment.comment = comment_file
+        - image="image_file.jpg"
```
vchudnov-g: Could you change this to `image= "image_file.jpg` (space before the quoted string) to make sure that works? Testing that gives me more confidence that it will work without the space than vice-versa. Or you could include both in configs.

hzyi-google: you mean `image= "image_file.jpg"` right?

hzyi-google:
@vchudnov-g PTAL

src/test/java/com/google/api/codegen/testsrc/libraryproto/library_gapic.yaml
vchudnov-g: yep

PR #2496, opened 2019-01-04
 Check interface presence in gapic config

### LINKED ISSUE #1232: Allow service config with multiple APIs
The service config allows exposure of multiple APIs in one yaml file. Toolkit does not handle this case, resulting in hacks where what would normally be one service config is split into multiple. (See https://github.com/googleapis/googleapis/tree/master/google/cloud/language which has a language_v1.yaml and a language_v1beta2.yaml instead of just a language.yaml). Let's get rid of the hacks, and allow services to be defined normally.

eoogbe:
Generates clients only for API interfaces present in the gapic config,
ignoring other API interfaces in the service config. This check allows
the service config to include interfaces for other versions of the API
without affecting gapic client generation for a single version.

Closes #1232


andreamlin:
Can you look into the go CircleCI test?

src/test/java/com/google/api/codegen/gapic/testdata/py/python_no_path_templates.baseline
```diff
@@ -1155,22 +1155,10 @@ import sys
 from google.api_core.protobuf_helpers import get_messages
 
 from example.proto import no_path_templates_messages_pb2
-from google.api import http_pb2
```
andreamlin: Are these changes safe?

src/test/java/com/google/api/codegen/gapic/testdata/py/python_multiple_services.baseline
```diff
@@ -1555,27 +1555,11 @@ import sys
 
 from google.api_core.protobuf_helpers import get_messages
 
-from google.api import http_pb2
```
andreamlin: Are these changes safe?

andreamlin:
@eoogbe Would it be possible to split this into two PRs, one that takes care of the Java 8 lambda refactoring and another once that concisely fixes that bug?

src/test/java/com/google/api/codegen/gapic/testdata/py/python_no_path_templates.baseline
eoogbe: They seemed to reference unreachable code, but I can double check with someone who knows the `types.py` files better. Perhaps @lukesneeringer ?

src/test/java/com/google/api/codegen/gapic/testdata/py/python_multiple_services.baseline
andreamlin: Ok, looking at google-cloud-python, I agree that longrunning, metadata, resources, signature, and descriptor *pb2 imports are unused. I'm not sure about the the API proto imports.

src/test/java/com/google/api/codegen/gapic/testdata/py/python_no_path_templates.baseline
andreamlin: I think you're right.

andreamlin:
How will this work when we don't have a GAPIC config interface at all? Is there a single point in the pipeline that we can later rework so that we do get to generate interfaces not present in the GAPIC config?

andreamlin:
I would imagine that during the migration period, we would be generating clients without a GAPIC config, say for Ads, while we are still generating clients with a GAPIC config.

Does this PR support that situation?

eoogbe:
As discussed, checking key presence in `GapicProductConfig#getInterfaceConfigMap` handles all the cases of: gapic config support, proto annotation support, and transitional support.

PR #2498, opened 2019-01-08
Qualify LRO type if not qualified

andreamlin:
Per API config specification, we should assume that LRO types of the proto package they are used in if the LRO type is not fully qualified.


andreamlin:
PTAL

PR #2499, opened 2019-01-09
Samplegen: generate build.gradle for the sample package

hzyi-google:


hzyi-google:
@beccasaurus @andreamlin @vchudnov-g PTAL

andreamlin:
And you've run the generated build.gradle to use gradle to run the samples?

If you wanted, you could also add executing samples into the CircleCI tests (separate PR maybe).

hzyi-google:
@andreamlin this and #2497 will make samples runnable. The other PR would shuffle the java baseline file completely so I split the changes into two.

We are still looking into tests for generated samples. We will definitely integrate with CircleCI when we are confident enough with the quality of generated samples.

PR #2501, opened 2019-01-09
Fix bash script

andreamlin:
https://circleci.com/gh/googleapis/gapic-generator/13088 failing log says
```
#!/bin/bash -eo pipefail
if [ -z $NOX_COMMANDS]
then
  echo "NOX_COMMANDS was not configured. Check CircleCI config."
  exit 1
fi
if [ -e ${PYTHON_GEN_DIR}/${TEST_API} ]
then
  cd ${PYTHON_GEN_DIR}/${TEST_API}
  echo "Running test commands: $NOX_COMMANDS"
  eval "$NOX_COMMANDS"
else
  echo "Python ${TEST_API} library was not generated."
  exit 1
fi


/bin/bash: line 0: [: missing `]'
```

This is because the first line should instead be ` [ -z "$NOX_COMMANDS" ]`, with `$NOX_COMMANDS` in quotes and a space between it and the closing bracket.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,7 +46,7 @@ anchor_check_generation: &anchor_check_generation
 # Test a Ruby client if generation succeeded.
 anchor_test_python_client: &anchor_test_python_client
   command: |
-    if [ -z $NOX_COMMANDS]
+    if [ -z "$NOX_COMMANDS" ]
     then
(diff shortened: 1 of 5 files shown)


andreamlin:
Internal fix for CircleCI python failure in cl/228612975.

PR #2502, opened 2019-01-09
Samplegen: avoid incorrect resource name configuration collisions

hzyi-google:
This PR fixes the incorrect collision between `resource%a` and `resource%b`.

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
@@ -309,9 +309,8 @@ public class InitCodeTransformer {
     HashMap<InitCodeNode, String> refFrom = new HashMap<>();
 
     // Keep track of the resource name entities. Configuring an entity twice or configuring an
-    // entity
-    // and the parent node at the same time will cause collision. Configuring two different entities
-    // will not.
+    // entity and the parent node at the same time will cause collision. Configuring two different
+    // entities will not.
     Multimap<InitCodeNode, String> nodeEntities = HashMultimap.create();
 
     // Below we'll perform depth-first search, keep a list of nodes we've seen but have not
@@ -326,12 +325,17 @@ public class InitCodeTransformer {
         InitCodeNode node = subNodes.pollLast();
         String oldPath = refFrom.put(node, path);
         if (oldPath == null) {
+          // The node has not been specified before, thus check if entity has been specified
           checkArgument(
               entity == null || nodeEntities.put(node, entity),
-              "SampleInitAttribute %s overlaps with %s",
-              path,
+              "Entity %s in path %s specified multiple types",
+              entity,
               path);
         } else {
+          // The node has been specified before. The will be no overlap if and only if:
+          // All previous paths are configuring entities
+          // This path is configuraing an entity
+          // The same entity is never specified before
           checkArgument(
               entity != null && nodeEntities.containsKey(node),
               "SampleInitAttribute %s overlaps with %s",
@@ -339,8 +343,8 @@ public class InitCodeTransformer {
               path);
           checkArgument(
               nodeEntities.put(node, entity),
-              "SampleInitAttribute %s overlaps with %s",
-              path,
+              "Entity %s in path %s specified multiple types",
+              entity,
               path);
         }
         subNodes.addAll(node.getChildren().values());
--- a/src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
@@ -14,6 +14,7 @@
  */
 package com.google.api.codegen.transformer;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkState;
 
 import com.google.api.codegen.config.FieldConfig;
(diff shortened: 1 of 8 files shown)


hzyi-google:
@beccasaurus @vchudnov-g PTAL

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -305,19 +308,40 @@ static void assertNoOverlap(InitCodeNode root, List<String> paths) {
...
-              String.format("SampleInitAttribute %s overlaps with %s", oldPath, path));
+        if (oldPath == null) {
+          checkArgument(
+              entity == null || nodeEntities.put(node, entity),
+              "SampleInitAttribute %s overlaps with %s",
+              path,
+              path);
+        } else {
+          checkArgument(
+              entity != null && nodeEntities.containsKey(node),
```
vchudnov-g: Am I  understanding this correctly?: Because oldPath was not null, the node was previously inserted. Because the node is being referenced again, it must be because it has an entity. And if it has an entity now, it means that the previous reference must have referenced an entity. The assertion checks these two "must"s.

If this is right, then the error message is not helpful. Also, adding in a comment the brief explanation I put above would help me in the future (and others not deeply immersed in this part of the code to begin with)

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -305,19 +308,40 @@ static void assertNoOverlap(InitCodeNode root, List<String> paths) {
...
+              "SampleInitAttribute %s overlaps with %s",
+              oldPath,
+              path);
+          checkArgument(
+              nodeEntities.put(node, entity),
+              "SampleInitAttribute %s overlaps with %s",
```
vchudnov-g: how about something ~`"entity '%s' in parameter '%s' specified multiple times", entity, node`

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -305,19 +308,40 @@ static void assertNoOverlap(InitCodeNode root, List<String> paths) {
...
+              path,
```
vchudnov-g: `path, path` again

src/test/java/com/google/api/codegen/gapic/testdata/py/python_library.baseline
```diff
@@ -4536,6 +4536,73 @@ def main():
...
+  client = library_v1.LibraryServiceClient()
+
+  # shelf = 'Novel'
+  # big_book_name = 'War and Peace'
+
+  if isinstance(shelf, six.binary_type):
+    shelf = shelf.decode('utf-8')
+  if isinstance(big_book_name, six.binary_type):
+    big_book_name = big_book_name.decode('utf-8')
+  name = client.book_path(shelf, big_book_name)
```
beccasaurus: 👍

src/test/java/com/google/api/codegen/gapic/testdata/php/php_library.baseline
```diff
@@ -3296,6 +3375,78 @@ $options += $defaultOptions;
...
+
+function sampleGetBigBook($shelf, $bigBookName)
+{
+    // [START hopper_core]
+
+    $libraryServiceClient = new LibraryServiceClient();
+
+    // $shelf = 'Novel';
+    // $bigBookName = 'War and Peace';
+    $formattedName = $libraryServiceClient->bookName($shelf, $bigBookName);
```
beccasaurus: 👍

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -305,19 +308,40 @@ static void assertNoOverlap(InitCodeNode root, List<String> paths) {
...
         InitCodeNode node = subNodes.pollLast();
         String oldPath = refFrom.put(node, path);
-        if (oldPath != null) {
-          throw new IllegalArgumentException(
-              String.format("SampleInitAttribute %s overlaps with %s", oldPath, path));
+        if (oldPath == null) {
+          checkArgument(
+              entity == null || nodeEntities.put(node, entity),
+              "SampleInitAttribute %s overlaps with %s",
+              path,
```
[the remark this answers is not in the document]
hzyi-google: Yeah, in this case it means the same entity has been specified twice - i.e. `path` overlaps with `path` :)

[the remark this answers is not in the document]
vchudnov-g: That's not clear from the error message. Let's just say it's been specified twice.

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -305,19 +308,44 @@ static void assertNoOverlap(InitCodeNode root, List<String> paths) {
...
+          // This path is configuraing an entity
+          // The same entity is never specified before
+          checkArgument(
+              entity != null && nodeEntities.containsKey(node),
+              "SampleInitAttribute %s overlaps with %s",
+              oldPath,
+              path);
+          checkArgument(
+              nodeEntities.put(node, entity),
+              "Entity %s in path %s specified multiple types",
```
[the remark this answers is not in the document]
hzyi-google: Oops... I'll fix it in a chaser PR.

Issue #2503, opened 2019-01-10
CircleCI Ruby Pubsub test/generation is failing


andreamlin:
Possibly due to https://github.com/googleapis/google-cloud-ruby/pull/2782#discussion_r246550700.

Sample failure log: https://circleci.com/gh/googleapis/gapic-generator/13089.

andreamlin:
cc @blowmage

eoogbe:
@jbolinger What's the status of this?

jbolinger:
This is blocked by https://github.com/protocolbuffers/protobuf/issues/5584

Yoshi, et al have post-processing workarounds but this is a good test for the namespace feature so I don't think we should workaround the root cause for our tests.

blowmage:
I somehow missed the notification for this over the weekend. My apologies.

As Joe says, we are working around this for the time being, including releasing a google-cloud-pubsub gem with the workaround last week.

andreamlin:
(Before this issue gets closed, we should re-instate the Ruby test in CircleCI)

jbolinger:
I sent an upstream patch to fix this: https://github.com/protocolbuffers/protobuf/pull/5735

However, even if they take it soon, it's probably going to be awhile before it gets released.

blowmage:
That's awesome! Thanks!

noahdietz:
@jbolinger can you provide an update here? I see there are some recent comments on the PR you opened.

1. Have the tests been re-instated following the workaround release?
2. Can we close this issue once your aforementioned PR is merged? Rather than wait for a release.

jbolinger:
We cannot reinstate the tests unless we apply a similar workaround or wait for a release with the fix. I suspect it's going to be awhile before a fix is released.

I think it may be better to lower the priority of this rather than close it, so we know the test isn't running. We're not making any changes to this Ruby generator here so fixing this test is not that important IMHO.

noahdietz:
Lowering to P2 as workarounds have unblocked the affected GAPIC and the upstream fix is in flight.

vchudnov-g:
@jbolinger Any updates on this? We're out of SLO.

jbolinger:
I think we should close it at this point given the progress on https://github.com/googleapis/gapic-generator-ruby. If it's still an issue it'll get raised during the migration and we've been able to work around it here fairly easily.

PR #2504, opened 2019-01-10
Remove Pubsub Ruby from testing

andreamlin:
https://github.com/googleapis/gapic-generator/issues/2503: Ruby Pubsub client generation is broken.

Temporarily remove Pubsub Ruby testing until the underlying issue is resolved.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -143,6 +143,10 @@ anchor_run_ruby: &anchor_run_ruby
           echo 'export TEST_API="pubsub"' >> $BASH_ENV
     # https://github.com/googleapis/gapic-generator/issues/2503 blocks Pubsub testing.
     - run:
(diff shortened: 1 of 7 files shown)


andreamlin:
PTAL

.circleci/config.yml
```diff
@@ -141,13 +141,7 @@ anchor_run_ruby: &anchor_run_ruby
         name: Prepare for Pubsub testing.
         command: |
           echo 'export TEST_API="pubsub"' >> $BASH_ENV
-    - run:
-        name: Test Pubsub.
-        <<: *anchor_test_ruby_client
-    - run:
-        name: Prepare for Speech testing.
```
jbolinger: I don't think we need to remove Speech.

andreamlin: whoops!
fixed.

PR #2506, opened 2019-01-10
Rename initView 

andreamlin:
There's so many methods named `initView` and it makes it hard to debug baseline files when snippet files' `initView` calls could refer to so many different methods.

Make different `initView` methods in different classes have different names.


src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
```diff
@@ -152,7 +152,7 @@ public void generateSamples(
     // to
     // use is the first one on the list, since we set it in the overload of generateSamples.
     if (methodSampleViews.size() > 0) {
-      methodViewBuilder.initCode(methodSampleViews.get(0).initCode());
+      methodViewBuilder.initCode(methodSampleViews.get(0).methodInitCode());
```
hzyi-google: Can you change this to `sampleInitCode`?

PR #2507, opened 2019-01-10
Parse proto annotations by default if no GAPIC config is given

andreamlin:
Also remove the library.proto, library.yaml, and library_pkg2.yaml files that were in a separate directory. Add the proto annotations to the main library.proto, and all baseline tests are just to use one source-of-truth library.proto.


src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -197,7 +197,14 @@ public static GapicProductConfig create(
     }
 
     // Toggle on/off proto annotations parsing.
-    ProtoParser protoParser = new ProtoParser(getDefaultLanguageFeatureConfig(language, null));
+    ProtoParser protoParser;
+    // TODO(andrealin): Expose command-line option for toggling proto annotations parsing.
+    if (configProto == null) {
```
michaelbausor: Nit: Maybe merge this with the configProto check on line 208? Don't feel that strongly though, please keep separate if you prefer.

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -197,7 +197,14 @@ public static GapicProductConfig create(
...
+      // By default, enable proto annotations parsing when no GAPIC config is given.
+      protoParser = new ProtoParser(true);
+    } else {
+      protoParser = new ProtoParser(getDefaultLanguageFeatureConfig(language, null));
```
michaelbausor: Isn't the purpose of Proposal 2 from your doc that this should always be `false`? If so, can we just make it `false`? Or is there benefit to retaining the `getDefaultLanguageFeatureConfig` method?

andreamlin: Hmm yes I don't think there will be any language-specific exemptions anytime soon.

andreamlin:
PTAL
Removed cruft.

PR #2508, opened 2019-01-10
Remove methodModel from GapicMethodContext because redundant

andreamlin:
cleanup

GapicMethodContext already contains a GapicMethodConfig, from which we can fetch a MethodModel. We don't need to carry around an extra MethodModel, which is sometimes different from the GapicMethodConfig.getMethodModel (this is more accurate because it has the reroute_to_grpc_interface config applied).


andreamlin:
PTAL

andreamlin:
It was just bothering me while I was debugging, because GapicMethodContext.getMethodModel() is sometimes different from GapicMethodContext.getMethodConfig().getMethodModel(), and the latter is more accurate.

PR #2510, opened 2019-01-11
Remove request_object_method

andreamlin:
There shall be no trace left of `request_object_method`.

All existing uses of that config param have been removed in googleapis and googleapis_private.


andreamlin:
PTAL

PR #2512, opened 2019-01-11
Preserve JSON field names

andreamlin:
Fixes https://github.com/googleapis/google-cloud-java/issues/4292.

The Gson util serializes a class by mapping the class's field names to the field's values. However, the class field names are not always the names used in the Discovery doc (names are sometimes escaped or in lowerCamelCase), and this makes Gson render the wrong field name.

This PR annotates fields that have a different name from the Discovery doc field with `@SerializedName`, which Gson will use to override the class's field name.

### DIFF
--- a/src/main/java/com/google/api/codegen/config/DiscoGapicInterfaceConfig.java
+++ b/src/main/java/com/google/api/codegen/config/DiscoGapicInterfaceConfig.java
@@ -154,7 +154,6 @@ public abstract class DiscoGapicInterfaceConfig implements InterfaceConfig {
       return null;
     } else {
       return new AutoValue_DiscoGapicInterfaceConfig(
-          methodConfigs,
           retryCodesConfig,
           retrySettingsDefinition,
           requiredConstructorParams,
@@ -163,10 +162,14 @@ public abstract class DiscoGapicInterfaceConfig implements InterfaceConfig {
           new DiscoInterfaceModel(interfaceName, model),
           smokeTestConfig,
           methodToSingleResourceNameMap.build(),
+          methodConfigs,
           methodConfigMap,
           singleResourceNames);
     }
   }
+
+  @Override
+  public abstract List<DiscoGapicMethodConfig> getMethodConfigs();
 
   private static Method lookupMethod(Document source, String lookupMethod) {
     for (com.google.api.codegen.discovery.Method method : source.methods()) {
--- a/src/main/java/com/google/api/codegen/config/DiscoveryField.java
+++ b/src/main/java/com/google/api/codegen/config/DiscoveryField.java
@@ -111,6 +111,10 @@ public class DiscoveryField implements FieldModel, TypeModel {
     return new DiscoveryField(schema, rootApiModel);
   }
 
+  public String getRawName() {
+    return originalSchema.getIdentifier();
+  }
+
   /** @return the underlying dereferenced Discovery Schema. */
   public Schema getDiscoveryField() {
     return schema;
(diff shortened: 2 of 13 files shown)


andreamlin:
PTAL

src/main/java/com/google/api/codegen/transformer/go/GoGapicSurfaceTransformer.java
```diff
@@ -255,7 +255,7 @@ static ModelTypeTable createTypeTable() {
 
   @VisibleForTesting
   List<StaticLangApiMethodView> generateApiMethods(
-      InterfaceContext context, Iterable<MethodModel> methods) {
+      InterfaceContext context, Iterable<? extends MethodModel> methods) {
```
hzyi-google: Does this `GoGapicSufaceTransformer` apply to both `Gapic` and `Discovery` APIs? If so this method LGTM. Otherwise (if GapicSurfaceTransformer works for Gapics only), it might be helpful to restrict the type of `methods` to `Iterable<ProtoMethodModel>` or `Iterable<? extends ProtoMethodModel>`

src/main/java/com/google/api/codegen/config/DiscoGapicInterfaceConfig.java
```diff
@@ -163,11 +162,15 @@ static DiscoGapicInterfaceConfig createInterfaceConfig(
...
           smokeTestConfig,
           methodToSingleResourceNameMap.build(),
+          methodConfigs,
           methodConfigMap,
           singleResourceNames);
     }
   }
 
+  @Override
+  public abstract List<DiscoGapicMethodConfig> getMethodConfigs();
```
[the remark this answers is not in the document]
hzyi-google: I assume the returned value would be popped by autovalue?

src/main/java/com/google/api/codegen/config/DiscoveryField.java
```diff
@@ -111,6 +111,10 @@ public static synchronized DiscoveryField create(Schema schema, DiscoApiModel ro
     return new DiscoveryField(schema, rootApiModel);
   }
 
+  public String getRawName() {
```
hzyi-google: Comments for public method?

src/main/java/com/google/api/codegen/transformer/go/GoGapicSurfaceTransformer.java
andreamlin: What is the added value of providing the concrete class implementation over the base abstract class?

src/main/java/com/google/api/codegen/config/DiscoveryField.java
andreamlin: Oh yes!

andreamlin:
PTAL; responded to comments

src/main/java/com/google/api/codegen/transformer/go/GoGapicSurfaceTransformer.java
hzyi-google: It may help to error out earlier and help debugging. (E.g., if a DiscoMethodModel is passed in accidentally, the code will fail at compile time rather than runtime)

I mean if discogapics can use the same concrete `SurfaceTransformer` as gapics, this is fine with me. Otherwise, keeping accessibility as strict as possible should be a good practice for Java.

andreamlin: Ideally the discogapics would use the same concrete SurfaceTransformer class as gapics, in the way JavaSurfaceTransformer supports both formats.

I agree that keeping accessibility strict helps with debugging and preventing bugs, but as long as we're having discussions about discogapics for Go, I would like to prevent devs from using `Gapic*` specific methods and entities in this SurfaceTransformer class, so that it makes it easier to reuse the existing abstractions.

PR #2515, opened 2019-01-17
Samplegen: Slightly refactor samplegen infrastructure and correctly import everything Java needs

hzyi-google:
- Add a default `SampleImportTransformer` to take care of import section for samples. It keeps track of types to import throughout init code, sample body and output part.
- Implement `JavaSampleImportTransformer` to make the import sections for Java complete. Those for Python and PHP will come later.
- Remove `sampleGenerator` functional interface which used to generate the initcode part for samples on the fly. Instead, we store an `initCodeTransformer` as a field of `SampleTransformer` since we need use it not only to generate `initCode`, but also to generate `sampleImportSection`.
- Convert `SampleTransformer` into an `AutoValue` to prevent the number of constructors from exploding.
- Other refactors to make `SampleTransformer` cleaner.

### DIFF
--- a/src/main/java/com/google/api/codegen/metacode/InitCodeContext.java
+++ b/src/main/java/com/google/api/codegen/metacode/InitCodeContext.java
@@ -82,6 +82,7 @@ public abstract class InitCodeContext {
         .collect(ImmutableList.toImmutableList());
   }
 
+  /** Returns a copy of this initCodeContext, but with an empty symbolTable. */
   public InitCodeContext cloneWithEmptySymbolTable() {
     return newBuilder()
         .initObjectType(initObjectType())
--- a/src/main/java/com/google/api/codegen/metacode/InitCodeContext.java
+++ b/src/main/java/com/google/api/codegen/metacode/InitCodeContext.java
@@ -82,6 +82,22 @@ public abstract class InitCodeContext {
         .collect(ImmutableList.toImmutableList());
   }
 
+  public InitCodeContext cloneWithEmptySymbolTable() {
+    return newBuilder()
+        .initObjectType(initObjectType())
+        .suggestedName(suggestedName())
+        .symbolTable(new SymbolTable())
+        .initFields(initFields())
+        .initValueConfigMap(initValueConfigMap())
+        .fieldConfigMap(fieldConfigMap())
+        .outputType(outputType())
+        .additionalInitCodeNodes(additionalInitCodeNodes())
+        .initFieldConfigStrings(initFieldConfigStrings())
+        .sampleParamConfigMap(sampleParamConfigMap())
+        .valueGenerator(valueGenerator())
+        .build();
+  }
+
   /**
    * Allows additional InitCodeNode objects which will be placed into the generated subtrees. This
    * is currently used by smoke testing only.
--- a/src/main/java/com/google/api/codegen/metacode/InitCodeContext.java
+++ b/src/main/java/com/google/api/codegen/metacode/InitCodeContext.java
@@ -84,19 +84,7 @@ public abstract class InitCodeContext {
 
   /** Returns a copy of this initCodeContext, but with an empty symbolTable. */
   public InitCodeContext cloneWithEmptySymbolTable() {
-    return newBuilder()
-        .initObjectType(initObjectType())
-        .suggestedName(suggestedName())
-        .symbolTable(new SymbolTable())
-        .initFields(initFields())
-        .initValueConfigMap(initValueConfigMap())
-        .fieldConfigMap(fieldConfigMap())
-        .outputType(outputType())
-        .additionalInitCodeNodes(additionalInitCodeNodes())
-        .initFieldConfigStrings(initFieldConfigStrings())
-        .sampleParamConfigMap(sampleParamConfigMap())
-        .valueGenerator(valueGenerator())
-        .build();
+    return toBuilder().symbolTable(new SymbolTable()).build();
   }
 
   /**
@@ -121,6 +109,8 @@ public abstract class InitCodeContext {
         .initFieldConfigStrings(ImmutableList.of())
         .sampleParamConfigMap(ImmutableMap.of());
   }
+
+  abstract Builder toBuilder();
 
   @AutoValue.Builder
   public abstract static class Builder {
--- a/src/main/java/com/google/api/codegen/transformer/DynamicLangApiMethodTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/DynamicLangApiMethodTransformer.java
@@ -63,7 +63,11 @@ public class DynamicLangApiMethodTransformer {
       SampleType sampleType) {
     this.apiMethodParamTransformer = apiMethodParamTransformer;
     this.initCodeTransformer = initCodeTransformer;
-    this.sampleTransformer = new SampleTransformer(sampleType);
+    this.sampleTransformer =
+        SampleTransformer.newBuilder()
+            .sampleType(sampleType)
+            .initCodeTransformer(initCodeTransformer)
+            .build();
   }
 
   public DynamicLangApiMethodTransformer(
@@ -73,7 +77,12 @@ public class DynamicLangApiMethodTransformer {
       OutputTransformer outputTransformer) {
     this.apiMethodParamTransformer = apiMethodParamTransformer;
     this.initCodeTransformer = initCodeTransformer;
-    this.sampleTransformer = new SampleTransformer(sampleType, outputTransformer);
+    this.sampleTransformer =
+        SampleTransformer.newBuilder()
+            .sampleType(sampleType)
+            .outputTransformer(outputTransformer)
+            .initCodeTransformer(initCodeTransformer)
+            .build();
   }
 
   public OptionalArrayMethodView generateMethod(GapicMethodContext context) {
@@ -236,9 +245,6 @@ public class DynamicLangApiMethodTransformer {
         initContext,
         context.getMethodConfig().getRequiredFieldConfigs(),
         initCodeOutputType,
-        initCodeContext ->
-            initCodeTransformer.generateInitCode(
-                context.cloneWithEmptyTypeTable(), initCodeContext),
         callingForms);
   }
 
--- a/src/main/java/com/google/api/codegen/transformer/DynamicLangApiMethodTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/DynamicLangApiMethodTransformer.java
@@ -40,7 +40,6 @@ import java.util.List;
  */
 public class DynamicLangApiMethodTransformer {
   private final ApiMethodParamTransformer apiMethodParamTransformer;
-  private final InitCodeTransformer initCodeTransformer;
   private final LongRunningTransformer lroTransformer = new LongRunningTransformer();
   private final HeaderRequestParamTransformer headerRequestParamTransformer =
       new HeaderRequestParamTransformer();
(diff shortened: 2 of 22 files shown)


src/main/java/com/google/api/codegen/metacode/InitCodeContext.java
```diff
@@ -82,6 +82,23 @@
         .collect(ImmutableList.toImmutableList());
   }
 
+  /** Returns a copy of this initCodeContext, but with an empty symbolTable. */
```
michaelbausor: What do you think about using AutoValue to generate a toBuilder method, and then just modifying the symbol table? (See https://github.com/google/auto/blob/master/value/userguide/builders-howto.md#-initialize-a-builder-to-the-same-property-values-as-an-existing-value-instance)

src/main/java/com/google/api/codegen/transformer/StaticLangApiMethodTransformer.java
```diff
@@ -61,7 +61,36 @@
   private final SampleTransformer sampleTransformer;
 
   public StaticLangApiMethodTransformer(SampleType sampleType) {
-    this.sampleTransformer = new SampleTransformer(sampleType);
+    this.initCodeTransformer = new InitCodeTransformer();
+    this.sampleTransformer = SampleTransformer.create(sampleType);
+  }
+
+  public StaticLangApiMethodTransformer(
```
michaelbausor: A possible refactoring here would be to have only one new constructor, that accepted a `SampleTransformer`, instead of multiple constructors that take combinations of parameters used to create the `SampleTransformer`. WDYT? The same applies to the DynamicLangApiMethodTransformer constructors too.

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
```diff
@@ -301,13 +307,25 @@ private InitCodeContext createInitCodeContext(
...
-      List<String> sampleCodeDefaultValues,
-      ImmutableMap<String, SampleParameterConfig> sampleParamConfigMap) {
+      SampleValueSet valueSet) {
     return InitCodeContext.newBuilder()
         .initObjectType(context.getMethodModel().getInputType())
         .suggestedName(Name.from("request"))
-        .initFieldConfigStrings(sampleCodeDefaultValues)
-        .sampleParamConfigMap(sampleParamConfigMap)
+        .initFieldConfigStrings(valueSet.getParameters().getDefaultsList())
+        .sampleParamConfigMap(
```
michaelbausor: It is fairly hard to understand what this large expression is doing. Maybe break it up instead of placing it all inline? Is the conversion from `SampleValueSet` to sampleParamConfigMap deserving of its own method?

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
```diff
@@ -183,44 +188,41 @@ public void generateSamples(
...
+                .build());
+
+    for (CallingForm form : callingForms) {
+      List<ValueSetAndTags> matchingValueSets =
+          methodConfig.getSampleSpec().getMatchingValueSets(form, sampleType());
+
+      if (sampleType() == SampleType.IN_CODE
+          || !methodConfig.getSampleSpec().isConfigured()
+          || matchingValueSets.isEmpty()) {
+        matchingValueSets = defaultValueSets;
```
michaelbausor: Is the intention here to overwrite the contents of `matchingValueSets` when `sampleType() == SampleType.IN_CODE`?

src/main/java/com/google/api/codegen/transformer/DynamicLangApiMethodTransformer.java
```diff
@@ -63,17 +63,30 @@ public DynamicLangApiMethodTransformer(
...
+    this.sampleTransformer =
+        SampleTransformer.newBuilder()
+            .sampleType(sampleType)
+            .initCodeTransformer(initCodeTransformer)
+            .sampleImportTransformer(
+                new SampleImportTransformer(initCodeTransformer.getImportSectionTransformer()))
+            .build();
   }
 
   public DynamicLangApiMethodTransformer(
```
michaelbausor: Is this constructor used anywhere? I couldn't see where this change is used.

src/main/java/com/google/api/codegen/metacode/InitCodeContext.java
hzyi-google: That's a nice feature!

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
hzyi-google: done.

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
```diff
@@ -44,18 +47,38 @@
  * A class that performs the transformations needed to generate the MethodSampleView for the
  * applicable sample types.
  */
-public class SampleTransformer {
+@AutoValue
+public abstract class SampleTransformer {
```
michaelbausor: It is a little bit odd to me that we have an autovalue class that also contains a lot of complicated logic. I'm not sure whether we do this in other places in gapic-generator. But I don't have any concrete suggestions for change... :-/

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
```diff
@@ -183,44 +190,41 @@ public void generateSamples(
...
 
     List<MethodSampleView> methodSampleViews = new ArrayList<>();
+    MethodConfig methodConfig = methodContext.getMethodConfig();
 
-    for (CallingForm form : callingForms) {
-      MethodConfig methodConfig = methodContext.getMethodConfig();
+    // For backwards compatibility in the configs, we need to use sample_code_init_fields instead
+    // to generate the samples in various scenarios. Once all the configs have been migrated to
+    // use the SampleSpec, we can delete the code below as well as sample_code_init_fields.
+    String defaultId =
```
michaelbausor: Can we refactor all of the logic for constructing `defaultValueSets` into its own method?

src/main/java/com/google/api/codegen/transformer/SampleImportTransformer.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+package com.google.api.codegen.transformer;
+
+import com.google.api.codegen.metacode.InitCodeNode;
+import com.google.api.codegen.viewmodel.CallingForm;
+import com.google.api.codegen.viewmodel.ImportSectionView;
+import com.google.api.codegen.viewmodel.OutputView;
+import java.util.List;
+
+public class SampleImportTransformer {
```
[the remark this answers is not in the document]
hzyi-google: It makes sense to make it an interface. I am not sure, however, if it should extend `ImportSectionTransformer`, since the concrete classes of this interface won't need to implement methods defined in `ImportSectionTransformer`. For example, they wouldn't be able to generate an `ImportSection` from a collection of `InitCodeNodes`.

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
hzyi-google: How about handwriting the fields and a builder?

src/main/java/com/google/api/codegen/transformer/SampleImportTransformer.java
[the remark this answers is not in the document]
hzyi-google: It would work, but I have a few concerns:
- There are two methods in `ImportSectionTransformer`:
```java
ImportSectionView generateImportSection(TransformationContext context, String className);
ImportSectionView generateImportSection(MethodContext context, Iterable<InitCodeNode> specItemNodes);
```
And if we have `SampleImportTransformer` extend `ImportSectionTransformer`, we will need to disallow users from calling the first method, probably by throwing an `UnsupportedOperationException` in `StandardSampleImportTransformer`, since that method wouldn't make sense for the import section for samples. So is it still worth extending `ImportSectionTransformer`?

- I am fine with the idea of having callers calling whatever `addXXX` to add imports in certain parts of the samples (that actually provides a lot of flexibility for in code samples in the future potentially). But if we make this decision on the surface, it seems better to make `generateImportSection` like:
```java
ImportSectionView generateImportSection(MethodContext context);
// or
ImportSectionView generateImportSection(ImportTypeTable typeTable);
```
Because all we need if all the types added to that `typeTable`. Reusing the method `mportSectionView generateImportSection(MethodContext context, Iterable<InitCodeNode> specItemNodes)` seems to indicated that the `SampleImportTransformer` generates an `ImportSection` from the `methodContext` and the `specItemNodes`, while the latter should have been taken care of by `addInitCodeImports`.

I am also not sure if these concerns are actually serious enough, but happy to know your thoughts.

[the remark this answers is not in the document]
michaelbausor: I think those concerns are serious enough. Let's try making it an interface but not extending `ImportSectionTransformer`, as you suggested previously.

hzyi-google:
PTAL

src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSGapicSamplesTransformer.java
```diff
@@ -55,8 +56,10 @@
   private final DynamicLangApiMethodTransformer apiMethodTransformer =
       new DynamicLangApiMethodTransformer(
           new NodeJSApiMethodParamTransformer(),
-          new InitCodeTransformer(new NodeJSImportSectionTransformer()),
-          SampleType.STANDALONE);
+          SampleTransformer.newBuilder()
+              .initCodeTransformer(new InitCodeTransformer(new NodeJSImportSectionTransformer()))
```
vchudnov-g: Can we reuse the transformer we instantiated above?

src/main/java/com/google/api/codegen/transformer/py/PythonGapicSurfaceTransformer.java
```diff
@@ -87,8 +90,13 @@
   private final DynamicLangApiMethodTransformer apiMethodTransformer =
       new DynamicLangApiMethodTransformer(
           new PythonApiMethodParamTransformer(),
-          new InitCodeTransformer(importSectionTransformer),
-          sampleType);
+          SampleTransformer.newBuilder()
+              .sampleType(sampleType)
+              .initCodeTransformer(new InitCodeTransformer(new PythonImportSectionTransformer()))
```
vchudnov-g: Here and in 98, can we reuse `importSectionTransformer`?

src/main/java/com/google/api/codegen/transformer/py/PythonPackageMetadataTransformer.java
```diff
@@ -321,7 +322,10 @@ private OptionalArrayMethodView createExampleApiMethodView(GapicMethodContext co
         new PythonMethodViewGenerator(
             new DynamicLangApiMethodTransformer(
                 new PythonApiMethodParamTransformer(),
-                new InitCodeTransformer(new PythonImportSectionTransformer())));
+                SampleTransformer.newBuilder()
+                    .initCodeTransformer(
+                        new InitCodeTransformer(new PythonImportSectionTransformer()))
```
vchudnov-g: can we reuse `importSectionTransformer`?

src/test/java/com/google/api/codegen/gapic/testdata/java/java_library.baseline
```diff
@@ -195,9 +196,12 @@ package com.google.example.examples.library.v1;
 
 // [START sample]
 import com.google.example.library.v1.FindRelatedBooksRequest;
+import com.google.example.library.v1.FindRelatedBooksResponse;
```
vchudnov-g: Unrelated to this PR, but we should follow up separately: this particular value set defines a default value of "Odyssey", which I don't see mentioned anywhere in this sample. Other samples do have it.

Could you file an issue and we pursue this separately from this PR?

src/main/java/com/google/api/codegen/transformer/py/PythonGapicSamplesTransformer.java
```diff
@@ -61,10 +63,16 @@
...
-              new PythonSampleOutputImportTransformer(), new PythonSamplePrintArgTransformer()));
+          SampleTransformer.newBuilder()
+              .initCodeTransformer(new InitCodeTransformer(importSectionTransformer))
+              .sampleType(sampleType)
+              .outputTransformer(
+                  new OutputTransformer(
+                      new PythonSampleOutputImportTransformer(),
+                      new PythonSamplePrintArgTransformer()))
+              .sampleImportTransformer(
+                  new StandardSampleImportTransformer(new PythonImportSectionTransformer()))
```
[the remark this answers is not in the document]
hzyi-google: Yes you are right!

src/test/java/com/google/api/codegen/gapic/testdata/java/java_library.baseline
hzyi-google: Sure.

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
```diff
@@ -44,18 +47,38 @@
...
+  public abstract InitCodeTransformer initCodeTransformer();
+
+  public abstract OutputTransformer outputTransformer();
+
+  public abstract SampleImportTransformer sampleImportTransformer();
+
+  public static Builder newBuilder() {
+    return new AutoValue_SampleTransformer.Builder()
+        .sampleType(SampleType.IN_CODE)
+        .initCodeTransformer(new InitCodeTransformer())
```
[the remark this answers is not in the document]
hzyi-google: It's mostly for backward compatibility. E.g., For now the `SampleTransformer` in `GapicSurfaceTransformers` are used to generate incode samples, and they don't have to worry about `OutputTransformer`. It'd be confusing if they are all required to give an `OutputTransformer` which isn't gonna to anything.
I instantiated those default instances outside the `newBuilder` method, so now they are only instantiated once. How about that?

src/main/java/com/google/api/codegen/transformer/java/JavaSampleImportTransformer.java
```diff
@@ -0,0 +1,178 @@
...
+      case Flattened:
+        saveResponseTypeName(context);
+        break;
+      case RequestPaged:
+        saveResourceTypeName(context);
+        break;
+      case FlattenedPaged:
+      case CallablePaged:
+        typeTable.saveNicknameFor(API_FUTURE);
+        saveResponseTypeName(context);
```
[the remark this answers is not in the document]
vchudnov-g: d'oh!

src/main/java/com/google/api/codegen/transformer/java/JavaSampleImportTransformer.java
```diff
@@ -0,0 +1,178 @@
...
+        saveResourceTypeName(context);
+        break;
+      case Callable:
+        typeTable.saveNicknameFor(API_FUTURE);
+        typeTable.getAndSaveNicknameFor(namer.getGenericAwareResponseTypeName(context));
+        break;
+      case CallableList:
+        saveResponseTypeName(context);
```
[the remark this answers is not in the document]
vchudnov-g: d'oh!

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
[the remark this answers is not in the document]
vchudnov-g: None of them have state, so that sounds good.

PR #2516, opened 2019-01-17
Validate method signature [WIP]

andreamlin:
Implement this part of the API Client Configuration doc:

> If an RPC lists a nested field in (google.api.method_signature).fields, none of the referenced fields may be repeated except for the last one. Code generators implementing this feature must error with a descriptive error message if encountering a non-terminal repeated field as a field name.
> 
> If any fields are required arguments, all required arguments are expected to appear before any optional ones. Code generators implementing this feature should error with a descriptive error message if encountering a required field after an optional one, and must do so if the resulting client library would not be valid.

Blocked on implementing parsing of method signatures with nested fields.


hzyi-google:
Based on the description I think this is something gapic-generator needs. I'll rebase off of it when I get some free cycles.

chrisdunelm:
Old PR. Closing.

PR #2517, opened 2019-01-17
Samplegen: Java template small fixes

hzyi-google:


src/main/resources/com/google/api/codegen/java/method_sample.snip
```diff
@@ -80,9 +85,15 @@
 
 @snippet asyncOperationCallableMethodSampleCode(apiMethod, initCode)
   {@initCode(initCode)}
-  OperationFuture<{@apiMethod.callableMethod.genericAwareResponseType}> future = {@sampleFutureMethodCall(apiMethod)};
```
andreamlin: Do you know why we used `genericAwareResponseType` in the first place?

hzyi-google: It seems it's converting `Empty` to `Void`. For exmaple, the callable methods for those rpcs who return `Empty` would return `ApiFuture<Void>` instead to be more Java idiomatic. But it seems we do not do this conversion for `LongRunningCallable`s.

andreamlin:
I think @lukesneeringer is working on expanding the `run_generated_tests` CI target. After that, we can merge if the only thing failing is the artman_smoke_tests (this has a preexisting problem).

hzyi-google:
Cool. They seem to be running now. Thanks for the heads up.

hzyi-google:
@andreamlin @lukesneeringer Would there be any permission issue to be sorted out for CI? The test has been <i>in progress<i> for more than an hour.

andreamlin:
@hzyi-google You can merge whenever the button says you can!

Issue #2521, opened 2019-01-18
String formatting functions are removed after introducing resource_name_generation


andreamlin:
This blocks release of some Java clients, such as the below Vision v1 library.

See comments of https://github.com/googleapis/google-cloud-java/pull/4161 for details.

It might be common for an API to only configure the resource path entities without specifying the `resource_name_generation` section that lists all the message objects that refer to those entities. In that case, `enableStringFormatFunctions` will be true, because it is true iff `resource_name_generation` is empty. However, once the `resource_name_generation` section is introduced, then we no longer generate the String format functions, and removing those functions will break some clients.

andreamlin:
@chingor13 Besides Vision, what other GA'd APIs are affected by this?

andreamlin:
still need to 
release artman with latest gapic-generator
add the `enableStringFormatFunctions` config param to the affected clients

vchudnov-g:
@andreamlin Could you provide an update?

andreamlin:
Done! Thanks for the reminder

PR #2522, opened 2019-01-18
Override enableStringFormatFunctions

andreamlin:
Fixes https://github.com/googleapis/gapic-generator/issues/2521.

Blocked on https://github.com/googleapis/gapic-generator/pull/2507


src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -88,6 +88,13 @@ message ConfigProto {
 
   repeated FixedResourceNameValueProto fixed_resource_name_values = 17;
 
+  // Set this value to True to enable String format functions for resource name
+  // entities. If this value is False, String format functions may or may not
+  // be enabled. This option is exposed for backward compatibility of Java clients.
+  // TODO(andrealin): Migrate this to a command-line option after migrating
+  // off artman.
+  bool enableStringFormatFunctionsOverride = 22;
```
michaelbausor: WDYT about making this a `BoolValue` instead of `bool`? That way, it can be true, false, or null - and can function as an override in both the true and false cases, instead of having the "may or may not" behaviour. I'm not sure how this works in the yaml files though.

andreamlin: Sure, that makes sense. Done.

andreamlin:
@michaelbausor PTAL once more?

src/main/java/com/google/api/codegen/transformer/java/JavaFeatureConfig.java
```diff
@@ -68,17 +69,27 @@ public static Builder newBuilder() {
   }
 
   @AutoValue.Builder
-  public abstract static class Builder {
+  abstract static class Builder {
```
michaelbausor: Why are these becoming package private? Just because they can be?

src/test/java/com/google/api/codegen/config/FeatureConfigTest.java
```diff
@@ -0,0 +1,52 @@
...
+  private ResourceNameMessageConfigs resourceNameMessageConfigs =
+      Mockito.mock(ResourceNameMessageConfigs.class);
+
+  @Test
+  public void testEnableStringFormatFunctionsOverride() {
+    Mockito.doReturn(resourceNameMessageConfigs)
+        .when(productConfig)
+        .getResourceNameMessageConfigs();
+
+    Mockito.doReturn(null).when(productConfig).enableStringFormattingFunctionsOverride();
```
michaelbausor: The other `Mockito.doReturn` statements are paired with an `assertThat` - is this one used?

src/main/java/com/google/api/codegen/transformer/java/JavaFeatureConfig.java
andreamlin: Yeah, figured I would adhere to the principle where we restrict visibility if possible; we can always make it public if need be.

src/test/java/com/google/api/codegen/config/FeatureConfigTest.java
andreamlin: Ok, added another assert statement.

PR #2523, opened 2019-01-18
Samplegen: make Node.js sample templates nicer

hzyi-google:


hzyi-google:
@alexander-fenster PTAL
@beccasaurus Just an FYI, nothing new here :)

PR #2524, opened 2019-01-19
Add test resources for routing headers

andreamlin:
Purely test resource changes.

Add a `shelf.name` header_request_param to the library_gapic.yaml, and add a corresponding `google.api.http` annotation to library.proto with the semantically same header request param.

Checking this in first will decrease the baseline diff for https://github.com/googleapis/gapic-generator/pull/2491.


andreamlin:
PTAL

andreamlin:
I wanted to test with a nested field, like `shelf.name`, that is representable by a resource path. In the code you linked, there is a nested field for a header param, but I didn't know how to set it to a semantically reasonable resource path in the http path.

PR #2531, opened 2019-01-31
Samplegen: response - comment

hzyi-google:


hzyi-google:
PTAL @beccasaurus @vchudnov-g

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -709,4 +709,13 @@ message OutputSpec {
   // bracketed list subscripts or map keys to dereference a descendant
   // of `var`.
   string define = 3;
+
+  // The elements of `comment` consists of a printf-like format string
```
vchudnov-g: I would reword the first sentence slightly:

```
The first element of `comment` is a printf-like format string, and subsequent elements are variable names (not names) to be interpolated in the format string....
```

src/test/java/com/google/api/codegen/testsrc/libraryproto/library_gapic.yaml
```diff
@@ -246,6 +246,11 @@ interfaces:
         - parameter: edition
           sample_argument_name: edition
       on_success:
+      - comment:
+        - "%% %% %% output handling %% %% %% %%"
+      - comment:
+        - "Do something with %s one by one"
+        - book_names
```
beccasaurus: I was struggling to figure out why this was separated out into `%s` but then I read the baseline's and saw how it helps with consistent casing/etc!

AWESOME.

beccasaurus: Guessing this will work within a loop?

I'm guessing we process the object exactly the same, but it might be worth adding a comment into a loop block to verify!

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -218,6 +222,19 @@ public void run() {
...
+  private OutputView.CommentView commentView(List<String> configs, MethodContext context) {
+    String comment = configs.get(0);
+    Object[] args =
+        configs
+            .subList(1, configs.size())
+            .stream()
+            .map(c -> context.getNamer().localVarName(Name.anyLower(c)))
+            .toArray(Object[]::new);
+    return OutputView.CommentView.newBuilder()
+        .lines(ImmutableList.<String>of(String.format(comment, args)))
```
vchudnov-g: This only generates one line in `OutputView.CommentView`, right?  How do we deal with user-specified `\n`s in the comment? Seems to me we would want to split on those so we have separate lines in the comment.

Also, how do we deal with extra-long lines specified by the user? Do we rely on a post-formatter to adjust them?

src/test/java/com/google/api/codegen/testsrc/libraryproto/library_gapic.yaml
```diff
@@ -246,6 +246,11 @@ interfaces:
         - parameter: edition
           sample_argument_name: edition
       on_success:
+      - comment:
+        - "%% %% %% output handling %% %% %% %%"
+      - comment:
```
vchudnov-g: This looks great!

Two edge cases that occur to me that we should include in these tests:
- comments with embedded `\n` provided by the user: let's make sure we  display those correctly
- comments that are longer than the idiomatic line length for a language: do we rely on a formatter to shorten them (my preference, though I'm wary of unintended consequences) or emit them as provided (less idiomatic, but easier to predict?...but then again, for the user to get this right they would have to know which level of indentation the comment will appear in, so maybe not so predictable after all)

beccasaurus: +1 to both

Note: I don't think we need to support conventions for multi-line comments in languages which support it (`/* */`), multiple `//` lines in those languages SGTM.

For newlines, in general, this would be helpful! (although also OK if we just wanted to avoid using newlines and, instead, use multiple comment array elements (?)

If you do decide to support newlines in the comment string, I recommend writing it using YAML's multiline string support _(of which there are a ridiculous number of [types](https://stackoverflow.com/questions/3790454/in-yaml-how-do-i-break-a-string-over-multiple-lines))_

```yaml
- comment:
  - |
      Hello there
      I have new lines
      which are preserved
```

src/test/java/com/google/api/codegen/testsrc/libraryproto/library_gapic.yaml
beccasaurus: Can `book_names` be any string?

– or does it refer to messages/fields/variable tables in any way?

If any string, does this always presume the string_will_be_in_snake_case? (which I like, just clarifying)

hzyi-google: `book_names` can be any string, and has to be in_snake_case :)

src/test/java/com/google/api/codegen/testsrc/libraryproto/library_gapic.yaml
beccasaurus: > RE: line length: formatters in artman will handle it

Awesome!

> RE: new line: I didn't know we have this feature :(

It's a feature of YAML, so it's handled by the YAML deserializer. Your code doesn't need to know about the YAML, it will simply see `\n` newlines :)

**#misc** **#yaml-syntax** **#multiline**
 - Using `|` keeps the newlines and includes a trailing `\n`.
 - Using `|-` keeps the newlines and does not add a trailing `\n`.
 - _someone even made a [small app](https://yaml-multiline.info/) to help choose between the various YAML multiline options_ 😂

So, thinking about `\n` ...

We should _not ignore empty newlines_, which help us add spacing into comments as desired.

Examples of my expectations, thinking about newlines...

```php
# Basic multiline example
"Hello\nworld\n"

// Hello
// world

# Added newline
"Hello\nworld\n\n"

// Hello
// world
//

# Another example
"Hello\nworld\n\nMore\n"

// Hello
// world
//
// More

# also, we don't need to care if there's no trailing \n (edge-case)
# e.g. if someone writes this string in the YAML: "Hello\nworld"
"Hello\nworld"

// Hello
// world
```

vchudnov-g: I agree we should not ignore empty new lines.

Being able to parse `|`-specified newlines sounds good. IIUC, we don't need to worry explicitly about that, they'll show up as `"...\n..."` when we parse  the yaml.

src/test/java/com/google/api/codegen/testsrc/libraryproto/library_gapic.yaml
vchudnov-g: Good idea about capturing the in-loop comment in a test!

src/test/java/com/google/api/codegen/testsrc/libraryproto/library_gapic.yaml
vchudnov-g: @hzyi-google On line length: will the formatters concatenate user-specified lines A and B if A gets reformatted so that it ends in a short visual (wrapped) line? I know emacs does this when it wraps code comments, and I imagine this may happen frequently with formatters. Example:

we specify:
`"this is a really long line which goes on and on and includes many words because we want people to really understand what we're trying to communicate.\nNo big deal\n"`

intending to get:

```
this is a really long line which goes on and on and includes many
words because we want people to really understand what we're
trying to communicate.
No big deal
```

but formatters could instead produce:

```
this is a really long line which goes on and on and includes many
words because we want people to really understand what we're
trying to communicate. No big deal
```

In those cases, we may need to have an explicit empty line between user-specified lines . (And hopefully this particular behavior is uniform across languages....)

Maybe we do this automatically, but the more I think about it, the more I lean towards just informing sample producers that if they have a single newline between lines of text, the lines may get concatenated after word-wrapping. They should explicitly insert a double newline `\n\n`. This would be less discoverable (I bet everyone would fail once before they get it right) but more predictable.

What do you two think? Automatic blank newline or user-specified?

hzyi-google: I don't know the exact behavior of formatters that artman uses regarding word-wrapping. 

We are able to transform a double newline into a blank line of comment. In my opinion this is fine for now since this is more consistent with most programming languages. We can fine tune this later if needed.

Yet I somehow can't get it work for trailing newlines. In fact, any number of trailing newline characters will be removed somewhere by gapic-generator :( I am still investigating the trailing newline problem.

vchudnov-g: Separately from the line concatenation issue, @hzyi-google, we should also add tests for the cases @beccasaurus pointed out above—specifically, that zero or one trailing newline should result in exactly one trailing newline, and two or more result in as many as are specified. Ie
`"foo"` and `"foo\n"` both result in

```
// foo
```

while `"foo\n\n"` results in
 
```
// foo
//
```

beccasaurus: > Yet I somehow can't get it work for trailing newlines

I'm less concerned about making this work (probably not major use-case)

```
//
// Hello there
//
```

Than I am making this work (pretty important)

```
// Hello
//
// There
```

i.e. If it's non-trivial to support trailing newlines, that's probably OK 😄

hzyi-google:
PTAL

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -224,14 +224,18 @@ public void run() {
 
   private OutputView.CommentView commentView(List<String> configs, MethodContext context) {
     String comment = configs.get(0);
+    System.out.println("comment:");
```
vchudnov-g: cruft

hzyi-google: Whoops. Definitely.

PR #2532, opened 2019-02-01
Use protoc plugin input/outputs

andreamlin:
In order to run the generator as a protoc plugin, we must accept CodeGeneratorRequest as the input object to the generator, and we must emit a CodeGeneratorResponse.

Saving the final hook-up of std-in/out to the generator for a different PR.


andreamlin:
PTAL

andreamlin:
PTAL

andreamlin:
Not sure why the bazel test is failing. The bazel test command works locally.

src/main/java/com/google/api/codegen/ProtocGeneratorMain.java
```diff
@@ -0,0 +1,125 @@
...
+/**
+ * Entrypoint for protoc-plugin invoked generation. Protoc passes input via std.in as a serialized
+ * CodeGeneratorRequest, and expects to read a CodeGeneratorResponse from std.out.
+ */
+public class ProtocGeneratorMain {
+
+  private static final ArtifactType DEFAULT_ARTIFACT_TYPE = GAPIC_CODE;
+
+  public static void main(String[] args) {
+    // TODO(andrealin): Implement. Handle I/O interfacing with protoc.
```
michaelbausor: Is the intention to submit this PR with this TODO here?

src/main/java/com/google/api/codegen/ProtocGeneratorMain.java
```diff
@@ -0,0 +1,125 @@
...
+    FileOutputStream fileoutput = new FileOutputStream(descriptorSetFile);
+    descriptorSet.writeTo(fileoutput);
+    fileoutput.close();
+    descriptorSetFile.deleteOnExit();
+
+    List<String> parsedArgs = new LinkedList<>();
+    parsedArgs.add("--descriptor_set");
+    parsedArgs.add(descriptorSetFile.getAbsolutePath());
+
+    // For now, assume there will only be one proto package to be generated.
```
michaelbausor: If we are making this assumption for now, we should error out in the case that there are additional files.

src/main/java/com/google/api/codegen/ProtocGeneratorMain.java
```diff
@@ -0,0 +1,125 @@
...
+    String firstFiletoGenerate = request.getFileToGenerate(0);
+    String protoPackage =
```
michaelbausor: What about reading the metadata proto annotation? Is that something to be added later?

src/main/java/com/google/api/codegen/ProtocGeneratorMain.java
```diff
@@ -0,0 +1,125 @@
...
+    // Parse plugin params, ignoring unknown params.
+    String[] requestArgs = request.getParameter().split(",");
+    for (String arg : requestArgs) {
+      if (Strings.isNullOrEmpty(arg)) continue;
+      parsedArgs.add("--" + arg);
+    }
+
+    String[] argsArray = parsedArgs.toArray(new String[] {});
+
+    return GeneratorMain.createCodeGeneratorOptionsFromProtoc(argsArray);
```
michaelbausor: This parseOptions function seems deeply tied together with the createCodeGeneratorOptionsFromProtoc function. Why does the createCodeGeneratorOptionsFromProtoc function belong in a different class?

src/main/java/com/google/api/codegen/gapic/GapicGeneratorApp.java
```diff
@@ -99,10 +101,33 @@
...
+
+  /**
+   * Constructs a code generator api based on given options.
+   *
+   * @param writeAsProtoResponse : If the output should be returned as a CodeGeneratorResponse,
+   *     instead of being written to file. The response object will be populated in
+   *     getCodeGeneratorProtoResponse().
+   */
+  public GapicGeneratorApp(
+      ToolOptions options, ArtifactType artifactType, boolean writeAsProtoResponse) {
```
michaelbausor: Rather than add a boolean flag and a codeGeneratorProtoResponse property to extract the data, what about abstracting out the writer concept. So accept a GapicWriter (or something like that) interface with a default implementation that writes to files, and an alternative impl that creates the proto response object. That way, the GapicGeneratorApp doesn't need to know anything about the proto response format, and writing to any new format (e.g. zip files, from the existing TODO) is easy in future. It could also be used for testing GapicGeneratorApp, if that's useful.

src/main/java/com/google/api/codegen/gapic/ProtoModels.java
```diff
@@ -29,6 +29,10 @@ private ProtoModels() {}
 
   /** Gets the interfaces for the apis in the service config. */
   public static List<Interface> getInterfaces(Model model) {
+    if (model.getServiceConfig().getApisCount() == 0) {
+      // A valid service config was not given.
```
michaelbausor: What is this checking for? Why is returning the list of interfaces the correct behaviour in this case?

src/test/java/com/google/api/codegen/gapic/ProtocGapicPluginGeneratorTest.java
```diff
@@ -0,0 +1,83 @@
...
+            .addAllProtoFile(
+                model.getFiles().stream().map(ProtoFile::getProto).collect(Collectors.toList()))
+            // Only the file to generate a client for (don't generate dependencies)
+            .addFileToGenerate("multiple_services.proto")
+            .setParameter("language=java")
+            .build();
+
+    CodeGeneratorResponse response = ProtocGeneratorMain.generate(codeGeneratorRequest);
+
+    // TODO(andrealin): Look into setting these up as baseline files.
```
michaelbausor: Is the intention to leave this for a future PR?

src/main/java/com/google/api/codegen/ProtocGeneratorMain.java
andreamlin: I have another PR right on top of this one that will implement the main method. So yes, if that's ok.

src/main/java/com/google/api/codegen/ProtocGeneratorMain.java
andreamlin: AFAIK, the metadata proto annotation will specify the *client* package name, not the input proto package name. Also yes, the final draft of the metadata annotations hasn't gone into the public repo for api-common-protos yet, so I'll wait for that to become SOT before implementing.

src/main/java/com/google/api/codegen/ProtocGeneratorMain.java
andreamlin: The `createCodeGeneratorOptionsFromProtoc` looks like all the other command-line-options-processing methods in `GeneratorMain`, plus it uses the `static final Option` fields defined in `GeneratorMain`. I figure it's easier to figure out and compare/contrast what that method is doing when placed next to other very similar methods.

src/main/java/com/google/api/codegen/gapic/GapicGeneratorApp.java
andreamlin: Sure, see https://github.com/googleapis/gapic-generator/pull/2546. That should be reviewed first before going any farther in this review.

src/test/java/com/google/api/codegen/gapic/ProtocGapicPluginGeneratorTest.java
andreamlin: Yes.

src/main/java/com/google/api/codegen/gapic/ProtoModels.java
andreamlin: This is checking for a valid service config. `model.getServiceConfig()` returns a non-null object even if a service config was not given. Also, `!model.getServiceConfig().equals(Service.getDefaultInstance()` even when a service config was not given (at the very least, the `config_version` field is set to 3.

One empirical observation is that empty service config will have `model.getServiceConfig().getApisCount() == 0`. Additionally, looking at the original logic that looks at the `apis` to find the interfaces, a valid service config must have non-empty `apis` in order to have a non-empty set of `Interface`s.

michaelbausor: That makes sense as a check for the service config. But why shouldn't we just throw an error if no service config is provided? Is it because we want to support the proto annotations? If so, then it is not very clear that this is what is going on - is it possible to move this logic "up", closer to where we make the decision about whether or not to parse the proto annotations?

andreamlin: See https://github.com/googleapis/gapic-generator/pull/2551.

There is a check in the main method to ensure that a service config option is provided.

src/main/java/com/google/api/codegen/gapic/GapicGeneratorApp.java
andreamlin: Resolved.

src/main/java/com/google/api/codegen/gapic/ProtoModels.java
andreamlin: Resolved.

andreamlin:
Ok, the existing blockers have been cleared up.

PTAL

src/main/java/com/google/api/codegen/ProtocGeneratorMain.java
michaelbausor: OK, sgtm. Perhaps we could consider a dedicated parsing class that both ProtocGeneratorMain and GeneratorMain share? But I haven't thought that through enough, and doesn't need to be in this PR.

PR #2535, opened 2019-02-01
[Discogapic] Generate comments for field methods

andreamlin:
Generate comments for setters and getters.
Generate comments for resource classes.


andreamlin:
PTAL

PR #2538, opened 2019-02-01
Samplegen: Nodejs used variable names in samples

hzyi-google:


hzyi-google:
@vchudnov-g PTAL

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -1565,6 +1568,13 @@ public String getSampleResponseVarName(MethodContext context) {
     return "response";
   }
 
+  public Set<String> getSampleUsedVarNames(MethodContext context, CallingForm form) {
+    // To not break PHP, Java and Python. We need to change this to an empty set
```
vchudnov-g: This comment isn't clear. Is this a TODO? Do you mean to set it to an empty set here and assign the values in a NodeJS-specific file?

I think what you're saying is
"TODO: Change this to an empty set once the PHP, Java, and Python implementations of this method are in place to define these identifiers. These values are here for now so we don't break those languages in the interim." Is this correct?

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -442,14 +458,19 @@ public void run() {
...
+      String identifier,
+      String methodName,
+      String valueSetId,
+      MethodContext context,
+      CallingForm form) {
     Preconditions.checkArgument(
-        !RESERVED_KEYWORDS.contains(identifier),
-        "%s:%s cannot define variable %s: it is a reserved keyword",
+        !context.getNamer().getSampleUsedVarNames(context, form).contains(identifier),
+        "%s: %s cannot define variable %s: it is used by the sample template for calling form %s.",
```
vchudnov-g: quotes around the variable name and the calling form name

src/test/java/com/google/api/codegen/transformer/OutputTransformerTest.java
```diff
@@ -77,22 +80,15 @@ public void testAccessorNewVariableFailWithReservedKeyword() {
     when(namer.getAndSaveElementResourceTypeName(typeTable, resourceFieldConfig))
         .thenReturn("ShelfBookName");
     when(featureConfig.useResourceNameFormatOption(resourceFieldConfig)).thenReturn(true);
+    when(namer.getSampleUsedVarNames(context, form)).thenReturn(ImmutableSet.of("response"));
```
vchudnov-g: suggestion: call this "fooresponse"  or something so it's obvious this is something that's only defined here and not in the code, and we an see this identifier clearly appear in the lines below.

PR #2542, opened 2019-02-04
go: example file uses aliased gapic import

### LINKED ISSUE #2541: Go: {gapic}_example_test.go files should use aliased gapic import
This issue was moved from [artman #547](https://github.com/googleapis/artman/issues/547).

The generated `{gapic}_example_test.go` file should import the GAPIC using a named/aliased package.

example:
```go
package library_test

import (
    "context"
    "io"

    library "cloud.google.com/go/library/apiv1"
    "google.golang.org/api/iterator"
    librarypb "google.golang.org/genproto/googleapis/example/library/v1"
)
```

Here, the GAPIC pkg import is named `library` explicitly in the generated code.

noahdietz:
Fix #2541

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/go/GoGapicSurfaceTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/go/GoGapicSurfaceTransformer.java
@@ -338,7 +338,11 @@ public class GoGapicSurfaceTransformer implements ModelToViewTransformer<ProtoAp
   void addXExampleImports(InterfaceContext context, Iterable<? extends MethodModel> methods) {
     ImportTypeTable typeTable = context.getImportTypeTable();
     typeTable.saveNicknameFor("context;;;");
-    typeTable.saveNicknameFor(context.getProductConfig().getPackageName() + ";;;");
+    typeTable.saveNicknameFor(
+        context.getProductConfig().getPackageName()
+            + ";"
+            + context.getNamer().getLocalPackageName()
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/api/codegen/transformer/go/GoGapicSurfaceTransformer.java
```diff
@@ -338,7 +338,11 @@ void addXApiImports(InterfaceContext context, Collection<MethodModel> methods) {
   void addXExampleImports(InterfaceContext context, Iterable<? extends MethodModel> methods) {
     ImportTypeTable typeTable = context.getImportTypeTable();
     typeTable.saveNicknameFor("context;;;");
-    typeTable.saveNicknameFor(context.getProductConfig().getPackageName() + ";;;");
+    typeTable.saveNicknameFor(
+        context.getProductConfig().getPackageName()
```
andreamlin: Woah, how does this change, which includes semicolons, result in the baseline change? Are the semicolon chars just part of the internal representation of the import type?

noahdietz: > Are the semicolon chars just part of the internal representation of the import type

That's my guess as well. I don't know 100%, I'd let @pongad confirm.

The `gapic-generator-go`, expects the gapic pkg path to be in a similar the form `path/to/impot;alias` which is then split into the path & alias name.

PR #2548, opened 2019-02-07
Uses snake_case for proto variable naming of enable_string_format_functions_override

andreamlin:
This field  hasn't been used at all in the wild so it's ok to change it right now


andreamlin:
PTAL

PR #2551, opened 2019-02-07
Get interfaces

andreamlin:
When service config is not given, just look at `model.getInterfaces()`. When service config is given, look at `model.getApis()`.

This blocks https://github.com/googleapis/gapic-generator/pull/2532

### DIFF
--- a/src/main/java/com/google/api/codegen/gapic/ProtoModels.java
+++ b/src/main/java/com/google/api/codegen/gapic/ProtoModels.java
@@ -36,7 +36,7 @@ public class ProtoModels {
       return getInterfacesFromServiceConfig(model);
     }
 
-    // Assume valid service config if proto annotations parsing is not enabled.
+    // No service config was given, or we shouldn't parse it.
     return model.getSymbolTable().getInterfaces().asList();
   }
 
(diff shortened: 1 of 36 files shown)


andreamlin:
PTAL

src/main/java/com/google/api/codegen/gapic/ProtoModels.java
```diff
@@ -27,8 +29,19 @@
...
   private ProtoModels() {}
 
+  /** Gets the interfaces for the apis in the model. */
+  public static List<Interface> getInterfaces(
+      Model model, @Nullable GapicProductConfig productConfig) {
+    if (productConfig == null || !productConfig.getProtoParser().isProtoAnnotationsEnabled()) {
+      return getInterfacesFromServiceConfig(model);
+    }
+
+    // Assume valid service config if proto annotations parsing is not enabled.
```
michaelbausor: Won't we end up here if proto annotation parsing *is* enabled?

andreamlin: Oh yes, fixed comment.

andreamlin:
Fixed wrong comment; PTAL

PR #2552, opened 2019-02-07
Change SampleGen CLI to snake_case

### LINKED ISSUE #2550: SampleGen: Update the generated CLI to use consistent `--param-name=""` flags for all languages
Today, we vary the CLI flags used to pass parameters to generated samples (when run) across the languages:
 - Python: `--file_path="image.png"`
 - PHP: `--filePath="image.png"`

I advocate that these are all changed to follow a `--file-path="image.png"` dash-es convention

Reasoning:
 - Almost all common *nix tools follow a standard `--foo-bar=` naming convention, so this does not conflict with any particular programming language's convention
 - It will be somewhat useful for any users who happen to run samples in many languages
 - It will be very, very useful for any sample developers who are generating samples in many languages
 - It will allow tools which test samples in many languages to use a common interface across the CLIs for executing samples for testing

hzyi-google:
Fix #2550

### DIFF
--- a/src/main/java/com/google/api/codegen/viewmodel/SampleFunctionParameterView.java
+++ b/src/main/java/com/google/api/codegen/viewmodel/SampleFunctionParameterView.java
@@ -1,0 +1,48 @@
+/* Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen.viewmodel;
+
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+public abstract class SampleFunctionParameterView {
+
+  public abstract String typeName();
+
+  public abstract String identifier();
+
+  public abstract InitValueView initValue();
+
+  /** We need to keep the cli flag names in the same case (snake_case) across all languages. */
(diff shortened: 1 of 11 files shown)


hzyi-google:
PTAL @vchudnov-g @beccasaurus

src/main/java/com/google/api/codegen/viewmodel/SampleFunctionParameterView.java
```diff
@@ -0,0 +1,48 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen.viewmodel;
+
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+public abstract class SampleFunctionParameterView {
```
vchudnov-g: Please add a class comment explaining why/what for. There are so many interlocking views in this generator!

src/main/java/com/google/api/codegen/viewmodel/SampleFunctionParameterView.java
```diff
@@ -0,0 +1,48 @@
...
+
+  public abstract String typeName();
+
+  public abstract String identifier();
+
+  public abstract InitValueView initValue();
+
+  /** We need to keep the cli flag names in the same case (snake_case) across all languages. */
+  public abstract String cliFlagName();
```
vchudnov-g: If one of the jobs of this class is to enforce snake_case across all languages, then maybe instead of passing the snake_cased cli flag name to the builder (so it's the responsibility of the caller), you derive it from `identifier` in this class (so this class is responsible for the uniformity). I think you can take `identifier` and convert it to lower_camel in this function, rather than having this function be AutoValue'd.

hzyi-google: I wanted to approach the problem that way too, but found no easy way to do it. The main issue is that the `identifier` is a string, and is either `camelCase` or `snake_case` depending on the language we are generating samples for, while this class does not know this information, so it doesn't know how to convert `identifier` to a `snake_case` string.

PR #2553, opened 2019-02-07
Only run artman_smoke_tests on master branch

andreamlin:
There's a very low chance that the artman_smoke_tests will fail, so let's not run them so we can free up circleCI machines.


andreamlin:
PTAL

PR #2561, opened 2019-02-11
BoolValue presence checking

andreamlin:
Presence of an object in proto is given by the `has[Object]` method.


andreamlin:
PTAL

andreamlin:
Ran this new jar locally on firestore and it did restore the string formatting methods.

PR #2563, opened 2019-02-12
Use Multiyaml reader everywhere

andreamlin:
Currently the baseline tests and Discogapic generator use a different config parsing pipeline than the actual  Gapic pipeline, which means that baseline tests don't actually show what would be produced by the Gapic generator.

This PR was motivated by https://github.com/googleapis/gapic-generator/pull/2564/, which adds a BoolValue field to a yaml file; this currently can't be parsed by the existing setup.


andreamlin:
@hzyi-google Can you review the baseline changes? Can you confirm that the new baseline changes are what would be expected given the existing gapic config ([on_success config](https://github.com/googleapis/gapic-generator/blob/de217ce99dcf059b8b6634b472b83c3b8c506084/src/test/java/com/google/api/codegen/testsrc/libraryproto/library_gapic.yaml#L253))

If this seems right, I'll send this to another person to review the implementation logic changes.

hzyi-google:
LGTM. This is indeed what we wanted to achieve. Thanks for doing this.

andreamlin:
PTAL @eoogbe

PR #2566, opened 2019-02-13
SampleGen: escape special characters in print statement

### LINKED ISSUE #2513: ensure string literals properly escape characters as needed
At the moment, this means ensuring that double quotes `"` inside the string literal (eg in a samplegen `print` statement) get escaped if needed. I encountered this with a config that contained something like this:
```
- print:
  - "alpha\"beta\"gamma"

```

which resulted in an incorrect  PHP string literal in the generated sample, `"alpha"beta"gamma"`, instead of the correct `"alpha\"beta\"gamma"`

This is not a problem in Python, since we use single quotes `'` to delimit string literals. But in PHP we aren't escaping the doubles, and it causes an error. Don't know about other languages.

We should ensure the characters in {`"` `'` `%` `\n`} are escaped as needed  in each language in `- print`:

- [x] Python
- [x] Java
- [x] PHP
- [x] NodeJS

We should ensure the characters in {`"` `'` `%` `\n`} are escaped as needed  in each language in `- defaults`:

- [ ] Python
- [ ] Java
- [ ] PHP
- [ ] NodeJS

hzyi-google:
`\"` -> `"`
`%%` -> `%%` for Java and PHP, or `%` for Node.js and Python
`%s` -> `%s`
`%x` -> error (x is not `%` or `s`)
`\\t` -> `\t`
Fix #2513


hzyi-google:
@vchudnov-g A comment about `\n` (and `\t`, whichi is quite similar) you mentioned in the issue comment. Users have to explicitly specify `\\t` or `\\n`, because `\t` and `\n` will be escaped by yaml (moreover, leading \t and trailing \n are left out by yaml) and gapic-generator won't know what character was there.
This PR will correctly escape `\"` in existing languages, but introduce a new one -- if a user specifies `\\"`, an error during sample compilation might occur. It seems to be a rare case and a case that's hard to solve nicely though.

vchudnov-g:
"Users have to explicitly specify `\\t` or `\\n`, because `\t` and `\n` will be escaped by yaml"

Could you expand on this comment? What does our YAML parser do? Does `\t` in the yaml become a literal string, ie in Pyton notation `r'\t'`, that is `"\\\t"`? If so, What happens to '\\\\' in the YAML?  Does it become `r'\\'`, ie `"\\\\"`?

hzyi-google:
I think I was wrong saying "`\t` will be escaped".
- `\t` is still `\t`.
e.g., `"ab\tcd"` in yaml will give us a string of five characters, `'a'`, `'b'`, '`\t`', `'c'` and `'d'`. When rendered by `.snip` files, the string will look like `"ab    cd"`. 
while leading tabs are left out, so `"\tabc"` only gives us a string of four characters, `'a'`, `'b'`, `'c'`.
- `\\` is still `\\`.
e.g., `"\\tabc"` in yaml will give us a string of five characters, `'\\'`, `'t'`, `'a'`, `'b'` and `'c'`. When rendered by `.snip` files, the string will look like `"\tabc"`.

vchudnov-g:
Ok, so it sounds like if we allow normal escape sequences in the yaml, they will be rendered correctly, except for leading `\t` and trailing `\n`, correct?
yaml:
`name='a\tb\nc'`
will result in a Java variable with characters `a` `\t` `b` `\n` `c`

If that is so, then the only issue is accommodating leading `\t` and trailing `\n`. Am I correct in understanding that these are issues even if the string is quoted in the yaml file? That is very surprising behavior for the parser.
ie. yaml:
```
- name: "\tfoo\n"
```

hzyi-google:
Yes. see: https://yaml-multiline.info/

vchudnov-g:
That link talks about multiline strings. Is that what we're talking about here?

Is this a problem for in-line strings?

See my (Python) gist: https://gist.github.com/vchudnov-g/f8f150e7ff55140949d67693028f931a
It suggests that for multi-line strings characters are treated as having already been escaped:
ie a yaml `\t\n` becomes the sequence `\\` `t` `\\` `n`, but for inline strings, _as long as they're quoted_, trailing `\n` and leading `\t` seem to be recognized. Is this what we're seeing in Java?

I'm trying to determine whether this problem is universal for all yaml parsers or just the one we're using in this generator. I want to make sure the config files will work with various correct yaml parsers in whatever language they're in.

(It sounds from what you're saying that doing normal escapes works in all cases except possibly trailing `\n` and leading `\t`. I want to see if there's an easy workaround for these two cases while retaining the simpler escaping for everything else. Hopefully the workaround is just having users just use the right input indicators, but if not, maybe we can just have the longer escapes for only these cases.)

hzyi-google:
Oh thanks for the sample!

Your comment about different behaviors reminded me of the previous issue we had with `comment` elements. Looking at commit history, this PR indeed happened before Andrea's fix of the YamlReader that gapic-generator uses. I merged in her commit to this PR and it indeed has the same behavior with your Python sample. I'll fix up the rest of the PR tomorrow. My apologies for causing the confusion :/

But I have a few follow up questions:
- How do we treat `\\`? Following the logic of `\t` and `\n`, it seems we should render the character backslash, according to the escaping rule of each language.
- How do we treat the trailing `\n`? For `print`, we always assume that there is a trailing newline at the end of the string. So if there is an explicit `\n` at the end of a `print` element, do we render two newline characters in the generated sample?
- Are there any other special characters that we need to deal with?

vchudnov-g:
Oh, cool if this becomes a non-issue.

Here are my thoughts on your questions. Happy to discuss!  :thinking: :

* When we read/escape the YAML config and wind up with an `\` character, if we then output it to a language where it needs to be escaped, then we should escape it. In fact, this goes for all special characters in all supported languages: eg if `$` needs to be escaped in a given language, then we should.

* trailing `\n`: The high-order bit is that we have the same behavior in all languages.Two alternatives
  - A: we can continue the current convention of always adding a `\n` at the end of a string. This is predictable, so if people have an extra new line,  they know what to do. In particular, with multiline yaml strings, they should be careful about using `|+` and `|-`
  - B: we can change the convention so that there isn't a trailing newline. This would mean suppressing the newline in Python (which it sounds like we can do easily in 2.6+, ([ref](https://stackoverflow.com/a/493399)) if we can distinguish the 2.x from 3.y)

   I have a _slight_ preference for B as the ideal end state, so we should get to it soon if we decide to do that. However, I'm totally fine with A, it's easier to do, and arguably makes configuration easier in the most common cases, which are probably single-line strings in the yaml file that will want trailing newlines.  @beccasaurus , thoughts?

- Other special characters? 
  The only special character we _accept_ and interpret specially is `%` and its escape sequence `%%` (double-check this sequence). Other escaped characters (in quoted yaml single-line strings) are assumed to be already escaped by the time they get to us. 

  For anything we _emit_, we need to make sure it's properly escaped, if needed, in the target languages/contexts. In various cases, these may include, for example, some of the following: `'` `"` `\` `$` `\t` `\n` (and in fact, any non-printable ASCII (or UTF-8?) character that could have been included in the YAML and we don't forbid. We could certainly enforce a restriction that we reject yaml with some unprintable characters)

hzyi-google:
@vchudnov-g PTAL. I did not change the behavior of trailing `\n`s for now. I kinda feel that could go in a different PR since that's more like a behavior change than a bug.

src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSSurfaceNamer.java
```diff
@@ -576,6 +576,7 @@ public String getFieldAccessorName(FieldModel field) {
 
   @Override
   public List<String> getPrintSpecs(String spec, List<String> args) {
+    spec = spec.replace("\\", "\\\\").replace("\t", "\\t").replace("\n", "\\n");
```
vchudnov-g: Since we output strings within backticks, we should escape the backticks, too.

Also, some of our strings (eg defaults) have single quotes, not backticks. So in that case we should escape _those_.

src/test/java/com/google/api/codegen/gapic/testdata/nodejs/nodejs_library.baseline
```diff
@@ -778,7 +778,7 @@ function samplePublishSeries(shelfName, edition) {
       const bookNames = response.bookNames;
       for (const title of response.bookNames) {
         // Now we deal with thisSingleBook!
-        console.log(`Title: {$title}`);
+        console.log(`\t% The book's title: "{$title}", \\\nand the book costs $50.00 %`);
```
vchudnov-g: We use backticks? Then we need to escape any backticks, too.

And a few lines below, we have single-quoted default values. We should makes sure to escape single quotes in that case.

src/test/java/com/google/api/codegen/testsrc/libraryproto/library_gapic.yaml
```diff
@@ -270,7 +270,7 @@ interfaces:
               - "Now we deal with %s!"
               - this_single_book
           - print:
-              - "Title: %s"
+              - "\t%% The book's title: \"%s\", \\\nand the book costs $50.00 %%"
```
vchudnov-g: Let's also have single quotes and backticks in this test string

Also, let's have a similar string for a default value, including double quotes, single quotes, backticks, `\t` `\n` to make sure we handle that correctly, since in NodeJS at least we use two different quoting conventions.

hzyi-google: - The single quote is already there => `book's`, I'll add a backstick :)
- Are the default values referring to the default values of parameters whose values are strings?

hzyi-google:
@vchudnov-g Do you mind if we handle special characters in `defaults` separately? The values are parsed at totally different places and it's hard to squash them in one PR.

vchudnov-g:
On the `defaults` escape chars: Sure, we can do that in a separate PR. Could you file a GitHub issue so we're sure not to forget, though?

PR #2571, opened 2019-02-13
Use the correct collection type in output handling in Java

### LINKED ISSUE #2570: SampleGen: Java output section does not handle repeated types correctly
Currently, if we assign a repeated type to a local variable in the output part, Java samples incorrectly render the inner type instead of the collection type. E.g.:
- proto:
```
message FooResponse {
  repeated string resources = 1;
  ...
}
```
- yaml:
```
- on_success:
  - define: my_resources=$resp.resources
  ...
```
The generated sample would be:
```java
String myResources = response.getResources();
```
Instead, the correct sample should be:
```java
List<String> myResources = response.getResources();
```

hzyi-google:
Fix #2570


src/test/java/com/google/api/codegen/testsrc/libraryproto/library_gapic.yaml
```diff
@@ -262,6 +262,7 @@ interfaces:
       - comment:
         - "Do something with %s one by one"
         - book_names
+      - define: book_names = $resp.book_names
```
vchudnov-g: Suggest you add a `- comment` saying you're just defining this variable for baseline testing purposes, even though it's not subsequently used.

PR #2573, opened 2019-02-13
Sanitize Node.js standalone sample templates

hzyi-google:
- add `'use strict';`
- remove placeholders
- correctly use `yargs`
- initialize the client
- move `require` outside sample function

### DIFF
--- a/src/main/resources/com/google/api/codegen/nodejs/standalone_sample.snip
+++ b/src/main/resources/com/google/api/codegen/nodejs/standalone_sample.snip
@@ -30,15 +30,15 @@
 # FIXME: Replace the following function calls with calls to functions that emit full standalone samples. These stubs have been adapted from method_sample.snip
 @snippet standaloneSample(apiMethod, sample)
 
-  const {@apiMethod.localPackageName} = require('{@apiMethod.packageName}').{@apiMethod.apiVersion};
+  @if apiMethod.hasApiVersion
+    const {@apiMethod.localPackageName} = require('{@apiMethod.packageName}').{@apiMethod.apiVersion};
 
+  @else
+    const {@apiMethod.localPackageName} = require('{@apiMethod.packageName}');
 
+  @end
   function {@sample.sampleFunctionName}({@formalArgs(sample.sampleInitCode.argDefaultParams)}) {
-    @if apiMethod.hasApiVersion
-      const client = new {@apiMethod.localPackageName}.{@apiMethod.apiVersion}.{@apiMethod.apiClassName}();
-    @else
-      const client = new {@apiMethod.localPackageName}.{@apiClassName}({})
-    @end
+    const client = new {@apiMethod.localPackageName}.{@apiMethod.apiClassName}();
     @switch sample.callingForm
     @case "Request"
       @if apiMethod.hasReturnValue
--- a/src/main/resources/com/google/api/codegen/nodejs/standalone_sample.snip
+++ b/src/main/resources/com/google/api/codegen/nodejs/standalone_sample.snip
@@ -39,7 +39,7 @@
   @end
   function {@sample.sampleFunctionName}({@formalArgs(sample.sampleInitCode.argDefaultParams)}) {
     @if apiMethod.hasApiVersion
-      const client = new {@apiMethod.localPackageName}.{@apiMethod.apiVersion}.{@apiMethod.apiClassName}({})
+      const client = new {@apiMethod.localPackageName}.{@apiMethod.apiVersion}.{@apiMethod.apiClassName}()
     @else
       const client = new {@apiMethod.localPackageName}.{@apiClassName}({})
     @end
--- a/src/main/resources/com/google/api/codegen/nodejs/standalone_sample.snip
+++ b/src/main/resources/com/google/api/codegen/nodejs/standalone_sample.snip
@@ -2,16 +2,21 @@
 @extends "nodejs/init_code.snip"
 
 @snippet generate(sampleFile)
-   //// [ This is an auto-generated sample file produced by the gapic-generator. Sample name: "{@sampleFile.className}" ]
-  'use strict';
   @let apiMethod = sampleFile.libraryMethod
     @let sample = apiMethod.samples.get(0)
+      // DO NOT EDIT! This is a generated sample ("{@sample.callingForm}",  "{@sample.valueSet.id}")
+      'use strict';
       // [START {@sample.regionTag}]
       // [START {@sample.regionTag}_core]
       {@standaloneSample(apiMethod, sample)}
       // [END {@sample.regionTag}_core]
       // tslint:disable-next-line:no-any
       // [END {@sample.regionTag}]
+      @if sample.sampleInitCode.argDefaultParams
+        {@processCliArguments(sample)}
+
+      @end
+      {@sample.sampleFunctionName}({@argvArgs(sample.sampleInitCode.argDefaultParams)});
     @end
   @end
 @end
@@ -60,11 +65,6 @@
     @end
   }
 
-  @if sample.sampleInitCode.argDefaultParams
-    {@processCliArguments(sample)}
-
-  @end
-  {@sample.sampleFunctionName}({@argvArgs(sample.sampleInitCode.argDefaultParams)});
 @end
 
 @private processResponse(sample)
--- a/src/main/resources/com/google/api/codegen/nodejs/standalone_sample.snip
+++ b/src/main/resources/com/google/api/codegen/nodejs/standalone_sample.snip
@@ -29,16 +29,19 @@
 #
 # FIXME: Replace the following function calls with calls to functions that emit full standalone samples. These stubs have been adapted from method_sample.snip
 @snippet standaloneSample(apiMethod, sample)
-  function {@sample.sampleFunctionName}({@formalArgs(sample.sampleInitCode.argDefaultParams)}) {
+
+  @if apiMethod.hasApiVersion
+    const {@apiMethod.localPackageName} = require('{@apiMethod.packageName}').{@apiMethod.apiVersion};
+
+  @else
     const {@apiMethod.localPackageName} = require('{@apiMethod.packageName}');
+
+  @end
+  function {@sample.sampleFunctionName}({@formalArgs(sample.sampleInitCode.argDefaultParams)}) {
     @if apiMethod.hasApiVersion
-      const client = new {@apiMethod.localPackageName}.{@apiMethod.apiVersion}.{@apiMethod.apiClassName}({
-        // optional auth parameters.
-      })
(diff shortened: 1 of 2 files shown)


src/test/java/com/google/api/codegen/gapic/testdata/nodejs/nodejs_library.baseline
```diff
@@ -250,28 +250,14 @@ module.exports.default = Object.assign({}, module.exports);
 
 ============== file: src/samples/v1/discuss_book_request_streaming_bidi_prog.js ==============
```
hzyi-google: @alexander-fenster 
Currently the samples live in the gapic source code. Do you think it makes sense to pull them out to the same level as `src`? i.e., `artman-genfiles/language/samples/v1/...`

src/test/java/com/google/api/codegen/gapic/testdata/nodejs/nodejs_library.baseline
```diff
@@ -292,41 +278,24 @@ function sampleDiscussBook(imageFileName) {
   stream.write(request);
 }
 
-require(`yargs`)
+argv = require(`yargs`)
```
hzyi-google: @beccasaurus @alexander-fenster 
To use `yargs` in standalone samples it seems that we need to add `yargs` to `package.json` of the gapic package. Is it OK?

src/test/java/com/google/api/codegen/gapic/testdata/nodejs/nodejs_library.baseline
```diff
@@ -351,36 +320,19 @@ function sampleFindRelatedBooks() {
...
-//        [names[0]=Odyssey, shelves[0]=Classics]
-//      apiMethod "findRelatedBooks" of type "PagedOptionalArrayMethod"
-
 // [START sample_core]
-
-// FIXME: Insert here code to prepare the request fields, make the call, process the response.
-
-/*
 function sampleFindRelatedBooks() {
+  const library = require('@google-cloud/library');
```
hzyi-google: @alexander-fenster @beccasaurus 
Currently this `require('@google-cloud/library')` is not compiling, since the samples and the client live in the same package, and they are sharing a common collection dependencies (`package.json`).
To make it work, we can:
1. Change `require('@google-cloud/library')` to `require('../../src')` similar to smoke test, but that would be very confusing for people who read the samples.
2. Move the samples out to a separate package which has it's own `package.json`. In this case we can keep `require('@google-cloud/library')`, and with some post-processing, it would be fairly easy to have the samples depend on either the generated client or a previous version that has already been published. However this would take fairly longer time for us to implement, as we might need to change artman too to make this happen.

hzyi-google:
@vchudnov-g @beccasaurus @alexander-fenster  PTAL

src/test/java/com/google/api/codegen/gapic/testdata/nodejs/nodejs_library.baseline
alexander-fenster: Can we follow the pattern established by existing libraries? That is: https://github.com/googleapis/nodejs-text-to-speech/tree/master/samples

Here `samples` folder is on the top level, it's a sibling to `src`.

src/test/java/com/google/api/codegen/gapic/testdata/nodejs/nodejs_library.baseline
alexander-fenster: Option 2 is how it works now. We generate a fake `package.json` for samples and put it inside `samples/` folder. No need for post-processing, all the logic you're talking about is being taken care of by `npm link` if needed.

src/test/java/com/google/api/codegen/gapic/testdata/nodejs/nodejs_library.baseline
hzyi-google: Sounds good!

src/test/java/com/google/api/codegen/gapic/testdata/nodejs/nodejs_library.baseline
hzyi-google: Cool!

src/main/resources/com/google/api/codegen/nodejs/standalone_sample.snip
```diff
@@ -3,45 +3,34 @@
 
 @snippet generate(sampleFile)
    //// [ This is an auto-generated sample file produced by the gapic-generator. Sample name: "{@sampleFile.className}" ]
-   //// STUB standalone sample "{@sampleFile.className}" /////
-
-   // FIXME: Insert here set-up comments that we never want to display in cloudsite. These are seen by users perusing the samples directly in the repository.
-
+  'use strict';
```
vchudnov-g: Use the metadata line like we do in the other languages:
```
@// DO NOT EDIT! This is a generated sample ("{@sample.callingForm}",  "{@sample.valueSet.id}")
```

We should have this be consistent to facilitate any tooling that may need to be built on top. If we decide to change the format, we should do uniformly for all langs.

src/test/java/com/google/api/codegen/gapic/testdata/nodejs/nodejs_library.baseline
vchudnov-g: The "main" code, including argument processing, should be outside the region tags.

hzyi-google:
@alexander-fenster I captured the restructure of the samples directory in #2574 and it might be better we do it in a separate PR, since the work will shuffle the files in the baseline and would be hard to review.
@vchudnov-g  PTAL

src/test/java/com/google/api/codegen/gapic/testdata/nodejs/nodejs_library.baseline
```diff
@@ -413,16 +413,16 @@ function sampleGetBigBook(shelf) {
     });
 }
 
+// [END hopper_core]
+// tslint:disable-next-line:no-any
+// [END hopper]
```
vchudnov-g: I'd suggest a blank line between the closing tag and the "main" lines.

src/main/resources/com/google/api/codegen/nodejs/standalone_sample.snip
```diff
@@ -2,46 +2,40 @@
...
-       // FIXME: Insert here code to prepare the request fields, make the call, process the response.
-
-       /*
-       {@standaloneSample(apiMethod, sample)}
-       */
-       // [END {@sample.regionTag}_core]
+    @let sample = apiMethod.samples.get(0)
+      // DO NOT EDIT! This is a generated sample ("{@sample.callingForm}",  "{@sample.valueSet.id}")
+      'use strict';
+      // [START {@sample.regionTag}]
```
vchudnov-g: I'd suggest some more vertical whitespace: a blank line above and below the `START` set and above and below the `END` set.

src/test/java/com/google/api/codegen/gapic/testdata/nodejs/nodejs_library.baseline
alexander-fenster: No, we should not add `yargs` to the main `package.json` of gapic library. Samples must have its own `package.json` that should have this dependency.

src/test/java/com/google/api/codegen/gapic/testdata/nodejs/nodejs_library.baseline
```diff
@@ -292,41 +278,24 @@ function sampleDiscussBook(imageFileName) {
...
-//      apiMethod "findRelatedBooks" of type "PagedOptionalArrayMethod"
-
 // [START sample_core]
-
-// FIXME: Insert here code to prepare the request fields, make the call, process the response.
-
-/*
 function sampleFindRelatedBooks() {
+  const library = require('@google-cloud/library');
+  const client = new library.v1.LibraryServiceClient({
```
alexander-fenster: One concern I have about this is that in JavaScript creating a new client is an expensive operation that involves reading and parsing proto files. By creating a client here we have a risk that users will copy this function as it goes and will create a lot of clients in their code instead of creating just one. I have some feeling that it might be better if a function accepted a client as a parameter. @beccasaurus what do you think?

beccasaurus:
@hzyi-google is this ready to be merged?

src/test/java/com/google/api/codegen/gapic/testdata/nodejs/nodejs_library.baseline
beccasaurus: > creating a new client is an expensive operation

Our code samples are isolated snippets which should be copy/paste runnable.

In cases where creating a new client object would hurt the user, we can call those out with custom hand-written code samples, e.g. to demonstrate re-using service objects to preserve streams/etc.

This follows the existing conventions of all code samples – 👍 we should instantiate client object

src/test/java/com/google/api/codegen/gapic/testdata/nodejs/nodejs_library.baseline
```diff
@@ -249,29 +249,17 @@ module.exports.v1 = gapic.v1;
...
-//        [name=BASIC, comment.comment = comment_file, image= "image_file.jpg"]
-//      apiMethod "discussBook" of type "OptionalArrayMethod"
-
 // [START turing_prog_request_streaming_bidi_core]
 
-// FIXME: Insert here code to prepare the request fields, make the call, process the response.
-
-/*
 function sampleDiscussBook(imageFileName) {
+  const library = require('@google-cloud/library');
```
beccasaurus: Move `require` outside of the function.

```js
const library = require('@google-cloud/library');

function sampleDiscussBook(imageFileName) {
  const client = new library.v1.LibraryServiceClient();
```

src/test/java/com/google/api/codegen/gapic/testdata/nodejs/nodejs_library.baseline
```diff
@@ -249,29 +249,17 @@ module.exports.v1 = gapic.v1;
...
+  const client = new library.v1.LibraryServiceClient({
+    // optional auth parameters.
```
beccasaurus: Remove this

src/test/java/com/google/api/codegen/gapic/testdata/nodejs/nodejs_library.baseline
```diff
@@ -478,41 +427,28 @@ function sampleGetBigBook(shelf) {
...
+'use strict';
 
+// [START hopper]
 // [START hopper_core]
 
-// FIXME: Insert here code to prepare the request fields, make the call, process the response.
-
-/*
 function sampleGetBigBook(shelf, bigBookName) {
+  const library = require('@google-cloud/library');
```
beccasaurus: Add the version of the API being required:

```js
```

hzyi-google:
@beccasaurus PTAL

PR #2576, opened 2019-02-16
LRO for Discovery clients [WIP]

andreamlin:


hzyi-google:
LRO in discogapic seems to already have been supported.

Issue #2580, opened 2019-02-20
SampleGen: Java CLI failure


hzyi-google:
For Java standalone samples, checked exception thrown by `CommandLineParser` is not caught or declared to throw, making the samples uncompilable.

- short-term solution
  - Add `throws Exception` to main method if we need to parse CLIs in a sample.
- long-term solution
  - Do not use `commons-cli`. Manually parse CLIs. This will drop the dependency of `commons-cli` altogether and avoid potential diamond dependency problems.

hzyi-google:
cc/ @vchudnov-g

vchudnov-g:
I agree with the short-term solution.

Long-term solution: are the diamond dependency problems an issue here with samples? I would think not, since the samples are meant to be standalone-runnable. I think we are better off relying on a standard library for CLI parsing rather than doing it ourselves, which is more work for us and bloat/distraction in the samples.

hzyi-google:
@vam-google may have a stronger opinion on this. It _might_ be a problem for samples in the sense that the dependencies of samples might depend on a different version of `common-cli`. And since we need to make the samples standalone-runnable we need to pull in all the dependencies of dependencies, which may introduce a risk of diamond dependency there.

However, while this might be a potential problem for us, it wouldn't cause any problem for others in the near feature as no one should depend on the samples. So IMO we can push the problem back until we really have to solve this.

Also, I wouldn't expect manually parsing the CLIs to be too hard since we just need to parse `--flags` with default values, i.e., we don't need any other fancy features `common-cli` supports.

vchudnov-g:
Yeah, parsing ourselves isn't the end of the world, but it could be extra cognitive load for users and friction for us in the future if we do need fancier features or if we need to tweak something.

I wonder whether dependencies of samples could ever depend on `common-cli`....

Curious to hear @vam-google 's thoughts. I'm OK with staying the course for now and revisiting if it does indeed become a problem or imminent problem. If we're all in agreement, maybe we can drop this to P3 and then bump it back if/when we determine we need to deal with this.

vam-google:
@vchudnov-g, @hzyi-google  Using `common-cli` would not necessarily break something (see "Samples specifics" section below for details), I was just trying to emphasize the importance of dependencies and potential harm of extra dependencies. 

**Some context**
[...]

PR #2581, opened 2019-02-20
SampleGen: handle exceptions thrown by CommandLineParser

hzyi-google:
Short-term fix for #2580


hzyi-google:
PTAL @andreamlin @vchudnov-g @beccasaurus

PR #2583, opened 2019-02-21
SampleGen: Fix Node templates

hzyi-google:


hzyi-google:
@beccasaurus @vchudnov-g PTAL

PR #2585, opened 2019-02-21
Use EmptyMessage instead of Void for HTTP response wrapper type

andreamlin:
This allows all interfaces that use `ApiMessage` to enforce that response objects have to be of type `ApiMessage`, which includes `EmptyMessage`. This replaces `java.lang.Void` as a response type.

Goes along with https://github.com/googleapis/gax-java/pull/678.

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/java/JavaSchemaTypeNameConverter.java
+++ b/src/main/java/com/google/api/codegen/transformer/java/JavaSchemaTypeNameConverter.java
@@ -52,7 +52,7 @@ public class JavaSchemaTypeNameConverter extends SchemaTypeNameConverter {
 
   private static String getPrimitiveTypeName(Schema schema) {
     if (schema == null) {
-      return "java.lang.Void";
+      return "com.google.api.gax.httpjson.EmptyMessage";
     }
     switch (schema.type()) {
       case INTEGER:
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/api/codegen/transformer/java/JavaSchemaTypeNameConverter.java
```diff
@@ -146,7 +146,7 @@ public TypeName getTypeNameForElementType(TypeModel type) {
   private TypeName getTypeNameForElementType(
       DiscoveryField fieldModel, BoxingBehavior boxingBehavior) {
     if (fieldModel == null) {
-      return new TypeName("java.lang.Void", "Void");
+      return new TypeName("com.google.api.gax.httpjson.EmptyMessage", "Void");
```
vam-google: Should the second argument be replaced with `"EmptyMessage"` (now it is still `"Void"`)?

andreamlin: Oh yes. Thanks.

andreamlin:
Fixed a thing. PTAL

andreamlin:
The compute-discovery-test won't pass until gax-java has been released with the actual `EmptyMessage` class (and google-cloud-java uses the newest release of gax-java). We can hold off merging until that happens.

andreamlin:
Actually merging this PR into a separate branch, from which all compute LRO work will happen on. This is because merging this into master will break immediate future google-cloud-compute clients.

PR #2587, opened 2019-02-22
Make default EmptyMessage value a non-null object

andreamlin:
This goes along with https://github.com/googleapis/gax-java/pull/685, which allows EmptyMessage to be constructed like a proto message.

This allows EmptyMessage to have a non-null test value, just like com.google.protobuf.Empty is constructed in https://github.com/googleapis/gapic-generator/blob/12e27eda15d2823684bba0a070252016bea9334e/src/test/java/com/google/api/codegen/gapic/testdata/java/java_library.baseline#L8368

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/java/JavaSchemaTypeNameConverter.java
+++ b/src/main/java/com/google/api/codegen/transformer/java/JavaSchemaTypeNameConverter.java
@@ -51,9 +51,6 @@ public class JavaSchemaTypeNameConverter extends SchemaTypeNameConverter {
   }
 
   private static String getPrimitiveTypeName(Schema schema) {
-    if (schema == null) {
-      return "com.google.api.gax.httpjson.EmptyMessage";
-    }
     switch (schema.type()) {
       case INTEGER:
         return "int";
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/api/codegen/transformer/java/JavaSchemaTypeNameConverter.java
```diff
@@ -51,9 +51,6 @@ public JavaSchemaTypeNameConverter(String implicitPackageName, JavaNameFormatter
   }
 
   private static String getPrimitiveTypeName(Schema schema) {
-    if (schema == null) {
```
andreamlin: this was never called (also not called in production, when generating the compute API)

src/main/java/com/google/api/codegen/transformer/java/JavaSchemaTypeNameConverter.java
```diff
@@ -94,9 +91,6 @@ private static String getPrimitiveZeroValue(Schema schema) {
     if (primitiveType.equals("java.lang.String")) {
       return "\"\"";
     }
-    if (primitiveType.equals("com.google.api.gax.httpjson.EmptyMessage")) {
```
andreamlin: this was never called (also not called in production, when generating the compute API)

andreamlin:
Note: this will be checked into a different branch (origin_lro) that isn't master. The compute-discovery-test will fail because gax-java hasn't yet been released with the proposed changes.

hzyi-google:
Please do remember to release gax.

Issue #2584, opened 2019-02-21
SampleGen: iterate over `map` fields in response


beccasaurus:
Samples cannot currently access `map` fields in responses.

⚠️ `java.lang.IllegalArgumentException: AnalyzeEntities:language_entities_text: entity.metadata is not a repeated field`

![screen shot 2019-02-21 at 1 06 13 pm](https://user-images.githubusercontent.com/280/53201705-80683880-35d9-11e9-9400-990bbf6df7e2.png)

```yaml
- id: language_entities_text
  description: "Analyze entities in text"
  parameters:
    defaults:
    - document.type=PLAIN_TEXT
    - document.content="California is a state."
  on_success:
  - loop:
      collection: $resp.entities
      variable: entity
      body:
      #
      # 🤔 How can we access metadata?
      #
      - loop:
          collection: entity.metadata
          variable: item
          body:
          - print:
            - "What is item? %s"
            - item
```

beccasaurus:
=> @hzyi-google 

/cc @tswast @vchudnov-g

vchudnov-g:
Thanks for the report. We had decided to punt on maps until there was a use case for it. Now we'll tackle it!

beccasaurus:
Here's an example GAPIC config you can use to test:
https://github.com/beccasaurus/gapic-docs-samples/blob/master/google/cloud/language/v1/language_gapic-with-iterating-over-map.yaml#L138-L149

hzyi-google:
Fixed by #2593

PR #2588, opened 2019-02-22
SampleGen: Access map entries in output handling

hzyi-google:
Allow accessing the value of a map entry by the key using `{}`. e.g., `$resp.map{key}`

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
@@ -398,7 +398,22 @@ public class OutputTransformer {
             valueSet.getId(),
             config.input());
       } else if (token == '{') {
-        throw new UnsupportedOperationException("map indexing not supported yet");
+        Preconditions.checkArgument(
+            type.isMap(),
+            "%s:%s: %s is not a map field",
+            context.getMethodModel().getSimpleName(),
+            valueSet.getId(),
+            config.input());
+        TypeModel keyType = type.getMapKeyType();
+        config.scan();
+        accessors.add(context.getNamer().getMapKeyAccessorName(keyType, config.tokenStr()));
+        type = type.getMapValueType();
+        Preconditions.checkArgument(
+            config.scan() == '}',
+            "%s:%s: expected '}': %s",
+            context.getMethodModel().getSimpleName(),
+            valueSet.getId(),
+            config.input());
       } else {
         throw new IllegalArgumentException(
             String.format(
--- a/src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
@@ -406,6 +406,7 @@ public class OutputTransformer {
             config.input());
         TypeModel keyType = type.getMapKeyType();
         config.scan();
+        keyType.validateValue(config.tokenStr());
         accessors.add(context.getNamer().getMapKeyAccessorName(keyType, config.tokenStr()));
         type = type.getMapValueType();
         Preconditions.checkArgument(
--- a/src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
@@ -398,6 +398,7 @@ public class OutputTransformer {
             valueSet.getId(),
             config.input());
       } else if (token == '{') {
+        // TODO: honor https://github.com/googleapis/gapic-generator/issues/2600
         Preconditions.checkArgument(
             type.isMap(),
             "%s:%s: %s is not a map field",
--- a/src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
@@ -405,7 +405,31 @@ public class OutputTransformer {
             valueSet.getId(),
             config.input());
         TypeModel keyType = type.getMapKeyType();
-        config.scan();
+        int keyToken = config.scan();
+        if (keyType.isStringType()) {
+          Preconditions.checkArgument(
+              keyToken == Scanner.STRING,
+              "%s:%s: expected string type for map key: %s",
+              context.getMethodModel().getSimpleName(),
+              valueSet.getId(),
+              config.input());
+        } else if (keyType.isBooleanType()) {
+          // `true` and `false` are the only valid literals here
+          Preconditions.checkArgument(
+              keyToken == Scanner.IDENT,
+              "%s:%s: expected boolean type for map key: %s",
+              context.getMethodModel().getSimpleName(),
+              valueSet.getId(),
+              config.input());
+        } else {
+          // Protobuf map keys can only be strings, booleans or integers
+          Preconditions.checkArgument(
+              keyToken == Scanner.INT,
+              "%s:%s: expected integral type for map key: %s",
+              context.getMethodModel().getSimpleName(),
+              valueSet.getId(),
+              config.input());
+        }
         keyType.validateValue(config.tokenStr());
         accessors.add(context.getNamer().getMapKeyAccessorName(keyType, config.tokenStr()));
         type = type.getMapValueType();
--- a/src/test/java/com/google/api/codegen/testsrc/libraryproto/library.proto
+++ b/src/test/java/com/google/api/codegen/testsrc/libraryproto/library.proto
@@ -499,6 +499,9 @@ message Book {
   // Tests Python doc generation: should generate a dummy file for shared_type
   // resource, but *not* its import, other_shared_type
   google.test.shared.data.Used resource = 29;
+
+  // For testing accessing map fields in samplegen
+  map<bool, string>  map_bool_key = 30;
 }
 
 // A single book in the archives.
--- a/src/test/java/com/google/api/codegen/testsrc/libraryproto/library_gapic.yaml
+++ b/src/test/java/com/google/api/codegen/testsrc/libraryproto/library_gapic.yaml
@@ -334,20 +334,15 @@ interfaces:
         defaults:
         - name="test_on_success_map"
       on_success:
-        - define: string_key_val = $resp.map_list_value_value{key}
-        - define: another_string_key_val = $resp.map_list_value_value{123}
         - define: int_key_val = $resp.map_string_value{123}
         - define: boolean_key_val = $resp.map_bool_key{true}
-        - define: map_value_field = $resp.map_message_value{key}.field
-        - print:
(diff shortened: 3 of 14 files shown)


hzyi-google:
PTAL @beccasaurus @vchudnov-g

beccasaurus:
Will there be a follow-up PR (at some point) for iterating through Map values? 😄 

For accessing via key, I'm 👍

I love all of the test cases in the YAML and baselines look good to me!!

Thank you for this _super quick_ follow-up to #2584 

/cc @tswast

hzyi-google:
Ah, I didn't read #2584 carefully enough. This PR clearly doesn't solve iterating through map values :(
But we still need this PR :) So I'll leave it here and work on #2584 in a separate PR.
Edited the PR comment as well.

beccasaurus:
Np, we need this too! Both halves of the Map support equation. 😎

I'm guessing you'll need to play around for a sec to figure out whether we do something like

```yaml
- loop:
    collection: $resp.some_map
    key: key_variable
    value: key_variable
```

... or something like ...

```yaml
- loop:
    collection: $resp.some_map
    variable: key_value_pair
    body:
    - print:
      - "Hi from %s and %s"
      - key_value_pair.key
      - key_value_pair.value
```

... or? 🤷🏼‍♀️

src/test/java/com/google/api/codegen/testsrc/libraryproto/library_gapic.yaml
```diff
@@ -328,10 +328,33 @@ interfaces:
       parameters:
         defaults:
         - name="Unreal book"
+    - id: test_on_success_map
+      title: "test maps in response handling"
+      parameters:
+        defaults:
+        - name="test_on_success_map"
+      on_success:
+        - define: string_key_val = $resp.map_list_value_value{key}
```
vchudnov-g: Where is `key` defined? Either I'm missing it, or this should fail.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -398,7 +398,23 @@ public void run() {
             valueSet.getId(),
             config.input());
       } else if (token == '{') {
-        throw new UnsupportedOperationException("map indexing not supported yet");
+        Preconditions.checkArgument(
```
vchudnov-g: Is curly braces the syntax we want to have? I know it's (only slightly) easier for us to code here, but is it better for users?

In NodeJS, PHP, Python, and Go, (and I think C# and Ruby) we use square brackets to index both lists and maps. Is there a reason to break that convention here? 

@beccasaurus, what do you think?

hzyi-google: Incode samples already use curly braces to access map entries: https://github.com/googleapis/gapic-generator/blob/12e27eda15d2823684bba0a070252016bea9334e/src/test/java/com/google/api/codegen/testsrc/libraryproto/library_gapic.yaml#L522

src/test/java/com/google/api/codegen/testsrc/libraryproto/library_gapic.yaml
hzyi-google: `key` is the literal of the actual name of the key in the map. Gapic-generator will try to parse  whatever is between the curly braces based on the key type. If [it can't](https://github.com/googleapis/gapic-generator/pull/2588/files#diff-1099a27774a5a8cfd19623070e23e508R409), it will fail.

For this specific behavior, the type of the key of the map is `string`, so gapic-generator will try to parse `key` as a string.

IMO gapic-generator is pretty liberal when it comes to parse literals from yaml configs. I think it's OK though, as long as the literals are rendered correctly according to the language-idiomatic syntax and their types.

That said, I am also OK to make the configuration more conservative (e.g., strings have to be quoted, etc) if y'all think this will benefit us more.

vchudnov-g: I think in this instance we should require quotes here. Otherwise it's easier to confuse it for a variable name (which is what I assumed).

My thinking is this:
- an entire yaml  value can be quoted or not, as per yaml rules, etc.
- we should generally be stricter about subparts of a yaml value that we parse, so that strings should be quoted, variable names should be lower_camel, etc.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
vchudnov-g: Sure, but the intent has always been to get rid of `samplecode_init_fields`, and in fact, it is going away with the microgenerators. I really don't think what we did here should have much, if any, weight on what we decide to do for samplegen.

The question, IMO, should be: what's most intuitive for our users? Given that it seems like the languages that we generate that support map subscripting use square brackets, I think there's a strong case for adhering to that widespread convention.

WDYT?

src/test/java/com/google/api/codegen/testsrc/libraryproto/library_gapic.yaml
hzyi-google: SG. I'll check it here.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
hzyi-google: I prefer curly braces. Happy to discuss further :)

First, yes, I agree we are gonna remove `sample_code_init_fields`, but my point was accessing a map field using `{}` is the pattern already adopted by gapic_config.yamls (In fact, wherever `InitCode` is used this pattern is adopted, so that includes smoke tests and the setup part for standalone samples too). I generally prefer not to break existing patterns unless we really need to.

Second, for explicitness. We have four ways to access nested elements all together: 1) access a field of a message, 2) access an entity of a resource name, 3) access a repeated field by giving an index, 4) access a map field by giving a key. The existing pattern has four different syntax for these four cases: `.` for 1), `%` for 2), `[]` for 3) and `{}` for 4) which is explicit and less error-prone. If we change 4) to use `[]`, same as 3), yes, it would be consistent with lots of real programming languages, but we will lose some explicitness.

Second point five (this is a bit of a personal philosophy so don't actually count as an argument), I think programming languages surfaces tend to lean towards being more general (as most of them use `[]` to access an array element and a map entry) to make themselves easier to use for developers, while configurations (yaml in our case) should lean towards being as explicit as possible to catch mistakes earlier.

vchudnov-g: On point 1: if this affects other parts of the config after moving to annotations, then yes, that's an argument for `{ }`. But in the annotation world, will the brackets for maps continue to be used?

On points 2 and 2.5: I'm not sure that `{ }` is less error prone than `[ ]`: you could make the same errors in both when _writing_ the config, and it's only when running the generator that you get notified. What `{ }` arguably buys us is that when you _read_ the config, you're immediately reminded by the syntax that this is a map.

On point 2.5: I agree with your intent, but I'm wary of creating special-case syntax to replace common-use syntax in just our situation.

I'm still personally leaning towards `[ ]` simply because it's more in use in programming languages, though I could accept the alternative. My main point is that I want to make sure that whichever option we go with, we actually do so _consciously_, making an argument for it (and I say this as the person who was fine with `{ }` in the stub/TODO code).

@beccasaurus, as a user of this, do you have an opinion?

If not, let's not block on this, but try get additional opinions/feedback quickly so if we need to change it we can do so before it's in widespread use.

beccasaurus: > What `{ }` arguably buys us is that when you _read_ the config, you're immediately reminded by the syntax that this is a map.

This is what I like about `[ ]`

But I could go either way. No strong preference as `[ ]` is the most common syntax across languages for accessing both array items by index and map items by key.

hzyi-google: Hmmmm, it seems `[ ]` wins.
I guess we need to change to use `[ ]` everywhere within samples? Here's what I think we should do:
1. Not block on this PR
2. We reach consensus that only `[ ]` is accepted to access a map item by key everywhere within sample config. This includes `print`, `define`, `defaults` and `attributes`, in both gapic-generator and the micro generators.
3. Specifically for gapic-generator languages, we accept both `{}` and `[]` for `defaults` and `attributes` to not break exiting incode samples and smoke tests because these configurations share a same parser. 
4. Have a PR to implement 2 and 3 for gapic-generator languages
5. When we completely get rid of the smoke tests and the old way of configuring incode samples we will deprecate `{ }` altogether.
@vchudnov-g WDYT?

hzyi-google: Issue filed: #2600

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -398,7 +398,47 @@ public void run() {
...
```
vchudnov-g: Could you add a comment/TODO referencing #2600 , that the subscript syntax may change?

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
vchudnov-g: Cool, thank you!

src/test/java/com/google/api/codegen/testsrc/libraryproto/library.proto
```diff
@@ -499,6 +499,9 @@ message Book {
   // Tests Python doc generation: should generate a dummy file for shared_type
   // resource, but *not* its import, other_shared_type
   google.test.shared.data.Used resource = 29;
+
+  // For testing accessing map fields in samplegen
+  map<bool, string>  map_bool_key = 30;
```
vchudnov-g: No need to do this now, or maybe at all, but it sure would be nice if we had consistent and informative naming for these maps: wg `map_bool_string`, `map_string_message`, `map_int_string`, etc.....

src/test/java/com/google/api/codegen/transformer/OutputTransformerTest.java
```diff
@@ -255,6 +255,63 @@ public void testAccessorNewVariableWithIndex() {
     assertThat(parent.getTypeModel("newVar")).isEqualTo(optionalPropertyTypeModel);
   }
 
+  @Test
+  public void testAccessorNewVariableMapKeyInvalidBooleanFail() {
```
vchudnov-g: Thanks for testing this!

src/test/java/com/google/api/codegen/testsrc/libraryproto/library.proto
hzyi-google: Hmmmm, good point :)

PR #2590, opened 2019-02-25
Added MoreExecutors

elisheva-qlogic:
Fixing [Remove warnings #4384](https://github.com/googleapis/google-cloud-java/issues/4384)

`ApiFutures.transform()` without an executor is deprecated.  Adding an explicit `MoreExecutors.directExecutor()` to use the non-deprecated variant of the method.


ajaaym:
@sduskis can you please review?

sduskis:
@andreamlin, I'm good with this change.  PTAL to see if there's anything obvious that's missing.

PR #2593, opened 2019-02-26
SampleGen: iterate over a map field in response handling

hzyi-google:


hzyi-google:
PTAL @vchudnov-g @beccasaurus

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -686,13 +686,29 @@ message LongRunningConfigProto {
 message OutputSpec {
   message LoopStatement {
     // The collection over which to iterate.
+    // Exactly one of `map` and `collection` should be specified.
     string collection = 1;
 
     // The iteration variable.
+    // Required only should only be specified if `collection` is specified.
```
vchudnov-g: nit: "only"x2

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -163,32 +163,98 @@ public void run() {
...
       SampleValueSet valueSet,
       ScopeTable localVars,
       CallingForm form) {
+    if (!loop.getCollection().isEmpty() && loop.getMap().isEmpty()) {
+      Preconditions.checkArgument(
+          !loop.getVariable().isEmpty(),
+          "Bad format: `variable` must be specified if `collection` is specified.");
+      Preconditions.checkArgument(
+          loop.getKey().isEmpty() && loop.getValue().isEmpty(),
+          "Bad format: `key` and `value` can't be specified if `collection is specified.");
```
vchudnov-g: nit: "... neither `key` nor `value` can be ..."
nit: end quote after `collection`

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -163,32 +163,98 @@ public void run() {
...
+      return arrayLoopView(loop, context, valueSet, localVars, form);
+    } else if (!loop.getMap().isEmpty() && loop.getCollection().isEmpty()) {
+      Preconditions.checkArgument(
+          loop.getVariable().isEmpty(),
+          "Bad format: `variable` can't be specified if `map` is specified.");
+      Preconditions.checkArgument(
+          !loop.getKey().isEmpty() || !loop.getValue().isEmpty(),
+          "Bad format: at least one of `key` and `value` must be specified if `collection is specified.");
```
vchudnov-g: nit: end quote after `collection`

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -686,13 +686,29 @@ message LongRunningConfigProto {
...
 
     // The contents of the loop.
     repeated OutputSpec body = 3;
+
+    // The map over which to iterate.
+    // Exactly one of `map` and `collection` should be specified. 
+    string map = 4;
+
+    // The iteration variable for key.
+    // Should only be specified when `map` is specified.
```
vchudnov-g: nit: g/Should/Can/

below as well

src/test/java/com/google/api/codegen/testsrc/libraryproto/library_gapic.yaml
```diff
@@ -776,6 +776,36 @@ interfaces:
         - parameter: name%shelf_id
           sample_argument_name: shelf
       on_success:
+      - comment:
+        - "Testing interating over map fields when both key and value are specified."
```
vchudnov-g: nit: s/interating/iterating/g
here and in 2 places below

src/test/java/com/google/api/codegen/gapic/testdata/nodejs/nodejs_library.baseline
```diff
@@ -519,6 +534,21 @@ function sampleGetBigBook(shelf) {
...
+        console.log(`key: ${myKey}, value: ${myValue}`);
+      }
+
+      // Testing interating over map fields when only key is specified.
+      for (const anotherKey of Object.keys(result.mapListValueValue)) {
+        console.log(`key: ${anotherKey}`);
+      }
+
+      // Testing interating over map fields when only value is specified.
+      for (const anotherValue of Object.values(result.mapListValueValue)) {
```
vchudnov-g: Is this preferable to `for...in`?
(Just asking what our style preference is)

hzyi-google: Yep, I did check with Alex in the design doc. 
However, @beccasaurus `Object.entries()`, `Object.keys()` and `Object.values()` will break Node 11(?), but it will go EOL this April. Is it OK we do it this way already (it's still two months away)?

vchudnov-g: For my own edification: is that Node11 EOL as a whole, or our support for it?
[This page](https://github.com/nodejs/Release) suggests Node 11 EOL is 2019-07 (but Node 10 EOL is in 2021??)

hzyi-google: Oh, it's Node 6 that does not support this feature.

I think Node 10 EOL is in 2021 because it's an LTS version?

beccasaurus: If I'm reading this correctly, it's supported in pretty much all Node.js versions:
https://node.green/#ES2015-misc-Object-static-methods-accept-primitives

beccasaurus: I vote that this is OK. We can revisit later if there are Node.js version issues. I like this thought

Although I would do `Object.values(result.mapListValueValue).forEach(anotherValue => { ... })`

beccasaurus: Hmm. I haven't seen `for (x of y)` usage yet.

@alexander-fenster is this the most idiomatic nowadays?

Samples today all use `forEach` but if the world is headed in the direction of `for... of` _and_ it's supported across versions of Node.js nowadays, I'm +1 `for... of`

alexander-fenster: I personally like `for (x of y)` but it's a personal preference rather than an observation of where is the world going (I really have no idea of that :) )

beccasaurus: Also confirmed with another Node person, beckwith.

`for (x of y)` +1

hzyi-google: Sounds good!

PR #2595, opened 2019-02-27
Cleanup

andreamlin:
Remove unused variables.


andreamlin:
PTAL

src/main/java/com/google/api/codegen/transformer/StaticLangApiMethodTransformer.java
```diff
@@ -465,15 +464,11 @@ public StaticLangApiMethodView generateAsyncOperationFlattenedMethod(MethodConte
...
-        ClientMethodType.AsyncOperationFlattenedMethod,
-        false);
+        ClientMethodType.AsyncOperationFlattenedMethod);
   }
 
   public StaticLangApiMethodView generateAsyncOperationFlattenedMethod(
-      MethodContext context,
-      List<ParamWithSimpleDoc> additionalParams,
-      ClientMethodType type,
-      boolean requiresOperationMethod) {
```
andreamlin: This var is removed because it is only referenced in line 493, which toggles line 494, which is already repeated in line 498. If line 498 is removed, then a java baseline test fails because the `operationMethod` field is referenced during runtime by one of the java snippet files.

PR #2598, opened 2019-02-28
SampleGen: Import Longrunning Metadata in Java and and dependency of common-cli build.gradle

### LINKED ISSUE #2592: SampleGen: Java LRO missing import for LongRunningXMetadata [API Speech V1]
I generated a sample for [Speech V1 `LongRunningRecognize`](https://cloud.google.com/speech-to-text/docs/reference/rpc/google.cloud.speech.v1#google.cloud.speech.v1.Speech.LongRunningRecognize)

The sample generated includes the following:

```java
 64       LongRunningRecognizeRequest request =
 65           LongRunningRecognizeRequest.newBuilder().setConfig(config).setAudio(audio).build();
 66       OperationFuture<LongRunningRecognizeResponse, LongRunningRecognizeMetadata> future =
 67           speechClient.longRunningRecognizeOperationCallable().futureCall(request);
```

([sample code](https://github.com/beccasaurus/gapic-docs-samples/blob/master/code-samples/speech/v1/java/src/main/java/com/google/cloud/examples/speech/v1/LongRunningRecognizeOperationCallableLongRunningCallableSpeechTranscribeAsync.java) – _lol it has lots of 'Metadata' imports because a script of mine kept adding them when I regenerated_)

I got an error because `LongRunningRecognizeMetadata` is not imported.

By default, the imports included:

```java
import com.google.cloud.speech.v1.LongRunningRecognizeRequest;
import com.google.cloud.speech.v1.LongRunningRecognizeResponse;
import com.google.cloud.speech.v1.LongRunningRecognizeMetadata;
```

And it worked when I added Metadata

```java
import com.google.cloud.speech.v1.LongRunningRecognizeRequest;
import com.google.cloud.speech.v1.LongRunningRecognizeResponse;
```

Not sure if this is generic to LROs or something Speech specific :)

Poking around other LROs I don't see Metadata messages, e.g. a [Video Intelligence LRO](https://cloud.google.com/video-intelligence/docs/reference/rpc/google.cloud.videointelligence.v1#google.cloud.videointelligence.v1.VideoIntelligenceService.AnnotateVideo)

/cc @hzyi-google

hzyi-google:
Fix #2592


src/main/java/com/google/api/codegen/transformer/java/JavaGapicSamplesPackageTransformer.java
```diff
@@ -66,11 +67,18 @@ public JavaGapicSamplesPackageTransformer(PackageMetadataConfig packageConfig) {
...
             .group("com.google.api")
             .name("api-common")
             .versionBound(packageConfig.apiCommonVersionBound(TargetLanguage.JAVA))
-            .build());
+            .build();
+    PackageDependencyView commonsCliDep =
+        PackageDependencyView.newBuilder()
+            .group("commons-cli")
+            .name("commons-cli")
+            .versionBound(VersionBound.create("1.4", "1.4"))
```
andreamlin: Not that it has to be done now, but is there a plan to fetch these hard-coded values from some other file?

hzyi-google: I thought about that too, but didn't come up with a better way to deal with it :/ This dependency is very specific to Java so it doesn't make sense to configure it from yaml.
Also, commons-cli doesn't have new release very frequently: 1.4 was release two years ago and the previous version before that (1.3.1) was released four years ago. So it wouldn't cause much maintenance friction for us to hardcode the version here.

andreamlin: There is the dependencies.yaml file that might be able to be referenced here. We could add another line to that file for the commons-cli version and then fetch from the package config?

hzyi-google: Oh OK. I'll fix it.

hzyi-google:
PTAL

hzyi-google:
Import Longrunning Metadata in Java and and dependency of common-cli build.gradle

PR #2601, opened 2019-02-28
Refactor MethodConfig

andreamlin:
Hide MethodConfig#getLongRunningConfig inside of MethodContext. This allows us to have, for the same API method that is a LRO:
- a MethodContext to generate a LRO method,
- and another MethodContext where it is not generated as an LRO method

This is needed because in Discovery LRO clients, the LRO methods are determined by the GAPIC config. If a method is not configured to be an LRO, and then later it is toggled as an LRO method in the GAPIC config, the previous non-LRO method will be replaced by the generated LRO method. This will cause a breaking change. Instead, we should always generate the non-LRO flavor for Discovery clients, where the heuristics to determine LRO are less standardized than the gRPC clients.


andreamlin:
PTAL

src/main/java/com/google/api/codegen/config/MethodConfig.java
```diff
@@ -74,8 +74,14 @@
 
   public abstract ReleaseLevel getReleaseLevel();
 
+  /**
+   * package-private for internal use.
+   *
+   * @see MethodContext#getLongRunningConfig wraps this call, because there are times where we want
```
vam-google: This will make the whole `wraps this call, because there are times where we want to generate the same API method as an LRO method and also as a non-LRO method.` string a one huge hyperlink in the generated javadoc. (See [getStreamWaitTimeout](http://googleapis.github.io/gax-java/1.40.0/apidocs/) funtion in gax-java documentation for an example of what the `@see` construct is converted into in a javadoc html). You can use `Ctrl+Q` key combination to see rendered javadoc of a method directly from `IntelliJ`.

Did you mean `{@link }` instead?

src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSPackageMetadataTransformer.java
```diff
@@ -232,15 +232,6 @@ private boolean hasBatching(ApiModel model, GapicProductConfig productConfig) {
         .anyMatch(InterfaceConfig::hasBatchingMethods);
   }
 
-  private boolean hasMixinApis(ApiModel model, GapicProductConfig productConfig) {
```
vam-google: Why was this removed?

src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceTestTransformer.java
```diff
@@ -203,13 +202,13 @@ private InitCodeContext createUnitTestCaseInitCodeContext(
         .build();
   }
 
-  private ClientMethodType getMethodType(MethodConfig config) {
+  private ClientMethodType getMethodType(MethodContext context) {
     ClientMethodType clientMethodType = ClientMethodType.RequestObjectMethod;
-    if (config.isPageStreaming()) {
+    if (context.getMethodConfig().isPageStreaming()) {
```
vam-google: Minor: For cases like this it is usually worth it to create a temporary variable `MethodConfig config = context.getMethodConfig()` first, because it is going to be reused multiple times.

src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSPackageMetadataTransformer.java
andreamlin: it was never referenced.

src/main/java/com/google/api/codegen/transformer/ruby/RubyGapicSurfaceTestTransformer.java
andreamlin: sure.

src/main/java/com/google/api/codegen/config/MethodConfig.java
andreamlin: oh thanks!

src/main/java/com/google/api/codegen/config/GapicInterfaceContext.java
```diff
@@ -301,7 +279,7 @@ private boolean isSupported(MethodModel method) {
   public Iterable<MethodModel> getLongRunningMethods() {
     List<MethodModel> methods = new ArrayList<>();
     for (MethodModel method : getSupportedMethods()) {
-      if (getMethodConfig(method).isLongRunningOperation()) {
+      if (getMethodConfig(method).getLroConfig() != null) {
```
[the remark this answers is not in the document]
andreamlin: done

PR #2602, opened 2019-02-28
Update generator to handle newest API Config Spec

andreamlin:
Using the [input-contract branch of api-common-protos](https://github.com/googleapis/api-common-protos/blob/input-contract/), because it still contains ResourceSets.

### DIFF
--- a/src/main/java/com/google/api/codegen/util/ProtoParser.java
+++ b/src/main/java/com/google/api/codegen/util/ProtoParser.java
@@ -44,6 +44,7 @@ import com.google.longrunning.OperationsProto;
 import com.google.protobuf.DescriptorProtos.FieldOptions;
 import com.google.protobuf.DescriptorProtos.FileOptions;
 import com.google.protobuf.Descriptors.EnumValueDescriptor;
+import com.google.protobuf.ExtensionRegistry;
 import com.google.protobuf.GeneratedMessage.GeneratedExtension;
 import com.google.protobuf.Message;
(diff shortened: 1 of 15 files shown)


andreamlin:
PTAL

src/main/java/com/google/api/codegen/util/ProtoParser.java
```diff
@@ -378,14 +391,15 @@ public boolean isHttpGetMethod(Method method) {
...
   }
 
   /** The OAuth scopes for this service (e.g. "https://cloud.google.com/auth/cloud-platform"). */
   public List<String> getAuthScopes(Interface service) {
-    OAuth oAuth = getProtoExtension(service, AnnotationsProto.oauth);
+    String oAuth = getProtoExtension(service, ClientProto.oauthScopes);
     if (oAuth != null) {
-      return ImmutableList.copyOf(oAuth.getScopesList());
+      return ImmutableList.copyOf(
+          Arrays.stream(oAuth.split(",")).map(String::trim).collect(Collectors.toList()));
```
michaelbausor: Do we need to check for isNullOrEmpty? And if not - can we remove the check on line 357?

andreamlin: 1. Oh yes.
2. For line 357, it needs the `isNullOrEmpty` check to catch an instance where some method signature is explicitly set to the non-null empty string `""`. Without the check, the method would return one of the method signatures being a List<String> containing the element `""`, when really we want to return an empty list.

Issue #937, opened 2017-01-18
Manual interface documentation cannot be un-escaped


pongad:
See: https://github.com/googleapis/googleapis/pull/241#discussion_r96564470

eoogbe:
@pongad Can you add a priority to this?

noahdietz:
@pongad can you give some more detail on this issue? Not sure there is enough here to start a fix. Thanks.

pongad:
I'm not sure how relevant this is anymore. At one point, we had `lang_doc` GAPIC config that lets us add free form string to our client. At least in Java, this string is also always escaped so it's impossible to make Javadoc link like `@See` etc.

Does this make sense?

noahdietz:
I'm just poking issue for SLO triage. If you think this is obsolete, feel free to close the issue. Else, lets assign to someone else?

pongad:
Ah sorry. I don't think this issue should be closed yet: `lang_doc` is [still being use](https://github.com/googleapis/googleapis/search?q=lang_doc&unscoped_q=lang_doc) and we still can't link is Javadoc. Who's the Java lead now?

noahdietz:
@andreamlin ?

andreamlin:
@pongad feel free to reassign this to me.

As a P2+, this is a low-level FR.

hkdevandla:
Obsolete.

PR #2604, opened 2019-02-28
Example of *not* escaping lang_doc string

andreamlin:
We can un-escape YAML special chars by putting them in single-quotes.

By following this example, we should be able to fix pubsub_gapic.yaml and close out https://github.com/googleapis/gapic-generator/issues/937

The resulting `@ literal @` is kinda weird but it's in line with the rendering in https://github.com/googleapis/gapic-generator/blob/580c79d8aecadd5f401b9a333c76af7c7f381907/src/test/java/com/google/api/codegen/gapic/testdata/java/java_library.baseline#L3918


andreamlin:
PTAL

pongad:
I think you got the issue backwards. I think we should be able to *not* escape. Eg, if we actually want to insert a `@See This.Other.Class` in the Javadoc, we should be able to.

andreamlin:
@pongad whoops yes fixed title. 
The content of the PR is not backwards tho is it?

pongad:
I think it is right? It's a little confusing to explain, so let me know if you want to follow up on IM.

But imagine what you might do if you actually want to insert a link in the Javadoc. If you're editing the Javadoc directly, you might write `@see the.Class`. But writing that in `lang_doc` doesn't create a link, it writes `{@literal @}see the.Class` which renders `@see the.Class`. Am I making this more confusing? :rofl:

andreamlin:
@pongad I am confused, yes. How is this different from 
```
 *         {@literal @}Override
```
in the generated method comment?

vchudnov-g:
I'm lacking context on this PR. Why do you have a PR exemplifying how **not** to escape?

Not an expert on javadoc, but it seems to me the output in the baselines is not what we want: we do want the `{@see everywhere}` just copied from the config, right? Is there a reason we cant just copy it?

Again, I'm probably missing the context/motivation for this PR.

vchudnov-g:
No action on this, not sure we need it. Closing for now.

fyi: @hzyi-google

PR #2605, opened 2019-03-01
Move around test resources for protoc

andreamlin:
protoc is really finicky when it comes to finding and linking dependent proto files. The way we have it set up right now, where the `testsrc/libraryproto/library.proto` is located in a different directory (testsrc/common) from the files it references (`"book_from_anywhere.proto"`, etc) without qualification, doesn't please protoc.

This PR allows the call
```
protoc --java_out /tmp/target/generated-sources \
-I=src/test/java/com/google/api/codegen/testsrc/common \
-I=/usr/local/bin/protoc-3.6.0/include/ \
-I ~/api-common-protos/ \
library.proto
``` 
to successfully compile library.proto.


andreamlin:
PTAL

PR #2606, opened 2019-03-01
Implement ProtoGeneratorMain.main()

andreamlin:
Also add a gradle target `protocPluginJar` to build the binary for the plugin,


src/main/java/com/google/api/codegen/ProtocGeneratorMain.java
```diff
@@ -45,7 +46,40 @@
   private static final ArtifactType DEFAULT_ARTIFACT_TYPE = GAPIC_CODE;
 
   public static void main(String[] args) {
-    // TODO(andrealin): Implement. Handle I/O interfacing with protoc.
+    System.err.println("Parsing System.in for CodeGeneratorRequest.");
```
michaelbausor: Do we want to always output this to System.err?

src/main/java/com/google/api/codegen/ProtocGeneratorMain.java
```diff
@@ -45,7 +46,40 @@
...
+      PrintWriter pw = new PrintWriter(sw);
+      e.printStackTrace(pw);
+      pw.flush();
+      response = PluginProtos.CodeGeneratorResponse.newBuilder().setError(sw.toString()).build();
+      exitCode = 1;
+    }
+
+    try {
+      response.writeTo(System.out);
+    } catch (IOException e) {
```
michaelbausor: Re error handling in this function, it looks like the structure is:
 - for IO errors, write to System.err and exit immediately
 - for errors in the generate function, don't write to System.error but set the error field on the response

Is that right? If so, should we call System.exit when we get an error calling response.writeTo?

src/main/java/com/google/api/codegen/ProtocGeneratorMain.java
andreamlin: ahh it was a debugging statement i suppose we can remove

src/main/java/com/google/api/codegen/ProtocGeneratorMain.java
andreamlin: That makes sense. Done.

PR #2608, opened 2019-03-01
Refactor gapic-generator Bazel build

vam-google:
The changes:
1) Minimum bazel version is now 0.23.0
2) Break the output of `java_gapic_assembly_gradle_pkg` :(. The target still can be built normally, but `./gradlew clean test` on the result of the target will fail because the generated `build.gradle` fille will be missing `gax` and `grpc` maven artifact dependencies. This is so because now those dependencies are pulled as Bazel workspaces and maven versions are not available during bazel build (so we do not know what to put in the generated build.gradle for those dependencies, because gradle would not be able to consume bazel dependency). This is relatively minor issue for now (the status of `java_gapic_assembly_gradle_pkg` is not clear). It will be fixed soon, but as a separate PR.
3) To simplify convergence between Gradle and Bazel builds, add `dependencies.properties` file, which lists the dependencies. This file is shared between gradle and bazel builds to pull actual dependencies and their versions.
4) Depend on gax ang grpc as bazel workspaces (not maven jars). This is what caused the issue described in item 1.
5) Update circle-ci bazel version from `0.19.0` to `0.23.1`.
6) Make `java_gapic_srcjar` rule compatible with `google-java-format-1.1` (before it was using features only available in `google-java-formatter-1.6`).


build.gradle
```diff
@@ -257,7 +234,7 @@ configurations {
 }
 
 dependencies {
-  codeGeneration libraries.autovalue, libraries.jsr305
+  codeGeneration libraries['maven.com_google_auto_value_auto_value'], libraries['maven.com_google_code_findbugs_jsr305']
```
michaelbausor: Nit: split line?

build.gradle
```diff
@@ -55,74 +55,51 @@ def protoVersion = '3.6.0'
...
+
+  // Gradle-specific build script dependencies
+  libraries.putAll([
+    protobuf: 'com.google.protobuf:protobuf-java:' + libraries['version.com_google_protobuf'],
+    protoc:  'com.google.protobuf:protoc:' + libraries['version.com_google_protobuf'],
+    javaFormatter: 'com.google.googlejavaformat:google-java-format:' + libraries['version.google_java_format'],
+  ])
 }
 
+
```
michaelbausor: Nit: remove

repositories.bzl
```diff
@@ -14,300 +14,65 @@
...
+        )
+
+    # Import Bazel-only dependencies (Gradle version will import maven artifacts of same
+    # version, while Bazel will depend on Bazel workspaces). The versions are shared in the
+    # properties file.
+
+    # A temporary hack, to fix incompatibility of protobuf 3.6.1 with new Bazel versions.
+    # We have to do it, because version 3.6.1.3 exists only as a Bazel workspace, but not as
+    # a maven artifact.
+    # Remove once upgraded to 3.7.0
```
michaelbausor: 3.7.0 is out - can we do this now?

rules_gapic/java/java_gapic_pkg.bzl
```diff
@@ -171,8 +171,11 @@ def java_gapic_proto_gradle_pkg(
     _java_gapic_gradle_pkg(
         name = name,
         pkg_type = "proto",
+        # TODO: the output of `java_gapic_assembly_gradle_pkg` is broken now
+        #  (because of missing the maven artifacts for the Bazel dependencies
+        #  we now depend on (gax, grpc). Fix this.
         deps = deps + [
-            "@com_google_protobuf_protobuf_java//jar",
+            #            "@com_google_protobuf_protobuf_java//jar",
```
michaelbausor: This formatting is a bit odd... why not just remove these lines?

rules_gapic/java/java_gapic_pkg_deps_resolution.bzl
```diff
@@ -197,6 +197,8 @@ def _possibly_artifact_id_version_chunk(chunk):
     return True
 
 def _possibly_artifact_id_classifier_chunk(chunk):
+    if chunk == "android" or chunk == "jre":
```
michaelbausor: Can you comment on why this is necessary?

rules_gapic/java/java_gapic_repositories.bzl
```diff
@@ -12,279 +12,28 @@
...
+#        name = "com_google_api_gax_java",
+#        urls = ["https://github.com/googleapis/gax-java/archive/v1.38.0.zip"],
+#        strip_prefix = "gax-java-1.38.0" % _gax_version,
+#    )
+
+    # TODO: switch to http_archive once the gax change is pushed
+    _maybe(
+        native.local_repository,
+        name = "com_google_api_gax_java",
+        path = "/usr/local/google/home/vam/_/projects/github/vam-google/gax-java"
```
michaelbausor: Can you make this an http_archive rule to your fork on github? Or is it not worth it, since it will change once the other PR is merged?

build.gradle
vam-google: done

build.gradle
vam-google: done

repositories.bzl
vam-google: Maybe, but this would mean diverging from what artman is doing. Also piloting gprc 3.7.0 on all googleapis. I think the chance of catching something is above 50%, I really do not want to mix this work with grpc dependency upgrade.

rules_gapic/java/java_gapic_pkg.bzl
vam-google: Reformatted. I'm not removing those, because those deps should be put back eventually.

rules_gapic/java/java_gapic_pkg_deps_resolution.bzl
vam-google: Added a comment

rules_gapic/java/java_gapic_repositories.bzl
vam-google: I think it is not worth it. I will change to point to real stuff, once gax PR is is submitted.

rules_gapic/java/java_gapic_pkg_deps_resolution.bzl
```diff
@@ -197,6 +197,8 @@ def _possibly_artifact_id_version_chunk(chunk):
     return True
 
 def _possibly_artifact_id_classifier_chunk(chunk):
+    # A hack to support guava-androind and guava-jre (androind and jre belong to a version chunk,
```
michaelbausor: Nit: s/androind/android

vam-google:
@michaelbausor PTAL

PR #2609, opened 2019-03-02
Remove extraneous @*Api annotations in snippet

andreamlin:
If there is a release annotation set on the method, it will conflict with the immediately following explicit `@BetaApi` annotation.


andreamlin:
PTAL

PR #2610, opened 2019-03-05
Fix Http Callables

andreamlin:
gax-java has changed `HttpJsonStubCallableFactory` to now be parameterized on a couple types: https://github.com/googleapis/gax-java/pull/684/files

Autogenerate a null-returning method and basic types to fulfill the updated interface's wants.

This PR would allow the generated Compute clients to work with the latest gax 1.42.0.


andreamlin:
This is still going to break the compute-discovery-test because gax-java can only be updated simultaneously with this PR in artman.

andreamlin:
PTAL

PR #2613, opened 2019-03-06
SampleGen: generate sample function documentation

### LINKED ISSUE #2577: SampleGen: add attribute for parameter description
I would like to update the snippets to look like this (_in the future – out of scope for this issue_ [`*`](https://github.com/beccasaurus/gapic-docs-samples/blob/master/code-samples/canonical-samples/python/synchronous-request/language_sentiment_text.py))

```py
# [START language_sentiment_text]
from google.cloud import language
from google.cloud.language import enums
from google.cloud.language import types

def analyze_sentiment_text(text_content='I am happy and joyful'):
    """Analyze sentiment of text
    Args:
        text: Text to analyze, e.g. 'Hello, world!'
    """

    client = language.LanguageServiceClient()

    document = types.Document(
        content=text_content,
        type=enums.Document.Type.PLAIN_TEXT)

    # ....
```

In this example, you'll notice a few things about this method parameter

 - (a) it's mapped to a request field _(**implemented!**)_
 - (b) it has a custom parameter name name _(**implemented!**)_
 - (c) it has a default value _(**implemented!**)_
 - (d) it has a docstring parameter description

I was thinking we could add this to the structural schema of samples:

```yaml
parameters:
  defaults:
  - document.type=PLAIN_TEXT
  - document.content="I am so happy and joyful."
  attributes:
  - parameter: document.content
    sample_argument_name: text_content
    description: "Text to analyze, e.g. 'Hello, world!'"  # <~~~ <~~~
```

`=>` @hzyi-google

hzyi-google:
Configuration:
```yaml
- parameters:
    defaults:
    - ...
    attributes:
    - parameter: this.is.a.param
    - description: "The description of this parameter."
```
Output: see baseline changes.

Fix #2577


src/test/java/com/google/api/codegen/gapic/testdata/py/python_library.baseline
```diff
@@ -4319,6 +4319,7 @@ import six
 def sample_discuss_book(image_file_name):
   """Testing calling forms"""
```
hzyi-google: This comes from `SampleValueSet.Description`. It seems the field was used as the description of the sample function for Python, but ignored by other languages. I added it to other languages as well. I _think_ this is the intended behavior?

hzyi-google:
PTAL @beccasaurus @vchudnov-g

src/test/java/com/google/api/codegen/gapic/testdata/py/python_library.baseline
```diff
@@ -4567,8 +4567,10 @@ from google.cloud.example import library_v1
 import six
 
 def sample_get_big_book(shelf, big_book_name):
-  """Testing resource name overlap"""
-
+  """ Testing resource name overlap
+  Args:
+  big_book_name: The name of the book.
```
beccasaurus: 💖 I am so happy right now.

src/test/java/com/google/api/codegen/gapic/testdata/php/php_library.baseline
```diff
@@ -3298,6 +3303,11 @@ require __DIR__ . '/../../../../vendor/autoload.php';
 
 use Google\Cloud\Example\Library\V1\LibraryServiceClient;
 
+/**
+ * Testing resource name overlap
+ *
+ * @param string bigBookName The name of the book.
```
beccasaurus: 🌷 Thank you, sir.

src/test/java/com/google/api/codegen/gapic/testdata/nodejs/nodejs_library.baseline
```diff
@@ -459,9 +459,13 @@ sampleGetBigBook(argv.shelf);
 // [START hopper]
 // [START hopper_core]
 
-
 const library = require('@google-cloud/library').v1;
 
+/**
+ * Testing resource name overlap
+ *
+ * @param bigBookName {string} The name of the book.
```
beccasaurus: 💕 This is a very good day for code samples.

src/test/java/com/google/api/codegen/gapic/testdata/java/java_library.baseline
```diff
@@ -520,6 +526,11 @@ import com.google.example.library.v1.ShelfBookName;
  * import com.google.example.library.v1.ShelfBookName;
  */
 public class GetBigBookAsyncLongRunningFlattenedAsyncWap2 {
+  /**
+   * Testing resource name overlap
+   *
+   * @param bigBookName The name of the book.
```
beccasaurus: 🤩 Simply beautiful.

src/test/java/com/google/api/codegen/gapic/testdata/py/python_library.baseline
beccasaurus: How is line wrapping handled? Could you add a longer description to one of the params to test wrapping > 80 chars?

hzyi-google: Line wrapping is not handled :( I _think_ artman format can handle it correctly? Actually let me test it real quick.

hzyi-google: Hmmmmm, clearly I am wrong :( Only Java formats the long line correctly. I'm gonna check with artman experts, and if that doesn't work out we may need to do it in gapic-generator.

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -300,18 +305,24 @@ private InitCodeView buildInitCodeViewRequestObject(
    * <li>a ReadFile node, returns the child node of that node.
    * <li>a resource path, returns the child node whose key equals the entity name in the path.
    */
-  private List<InitCodeNode> sampleFuncParams(InitCodeNode root, List<String> paths) {
+  private List<InitCodeNode> sampleFuncParams(
```
vchudnov-g: Could you add the new param to the javadoc?

src/main/resources/com/google/api/codegen/java/standalone_sample.snip
```diff
@@ -290,6 +293,23 @@
   @end
 @end
 
+@private processDoc(doc)
+  # If there are no parameters, we generate the method description in one line.
+  @if doc.parameters.isEmpty
+    /** {@doc.description} */
```
vchudnov-g: The formatter takes care of long lines, correct? (In all languages?)

src/test/java/com/google/api/codegen/gapic/testdata/py/python_library.baseline
```diff
@@ -4567,8 +4567,10 @@ from google.cloud.example import library_v1
 import six
 
 def sample_get_big_book(shelf, big_book_name):
-  """Testing resource name overlap"""
-
+  """ Testing resource name overlap
+  Args:
```
vchudnov-g: Blank line before `Args`

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
```diff
@@ -251,6 +252,17 @@ private MethodSampleView generateSample(
...
         sampleImportTransformer().generateImportSection(methodContext);
+    SampleFunctionDocView docView =
+        SampleFunctionDocView.newBuilder()
+            .parameters(
+                initCodeView
+                    .argDefaultParams()
+                    .stream()
+                    .filter(p -> !p.description().isEmpty())
+                    .collect(ImmutableList.toImmutableList()))
+            .description(valueSet.getDescription())
```
vchudnov-g: The value set description was not originally intended to be the method description. It was intended as a description visible in the UI of APIx samples. However, I think it can double as a method description: the information both descriptions convey is the same, as I think about it, so this field can be the source of truth for both.

src/test/java/com/google/api/codegen/gapic/testdata/py/python_library.baseline
vchudnov-g: Not exactly, but this is fine. (See comment in my review.)

src/main/resources/com/google/api/codegen/java/standalone_sample.snip
hzyi-google: Not really. we'll have to handle long lines in gapic-generator. I'm gonna push some new changes to handle them.

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
hzyi-google: SGTM

hzyi-google:
Added word wrapping. Do y'all want to take a look again?

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
```diff
@@ -251,6 +255,18 @@ private MethodSampleView generateSample(
                     initCodeContext.cloneWithEmptySymbolTable())); // to avoid symbol collision
     ImportSectionView sampleImportSectionView =
         sampleImportTransformer().generateImportSection(methodContext);
+    SampleFunctionDocView sampleFunctionDocView =
+        SampleFunctionDocView.newBuilder()
+            .paramDocLines(paramDocLines(methodContext, initCodeView))
```
vchudnov-g: nit: I suggest passing `DOC_LINE_MAX_WIDTH` to `paramDocLines()` so both uses of the constant are within the same function. I find the asymmetry troubling :man_shrugging:

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
```diff
@@ -51,6 +53,8 @@
 @AutoValue
 public abstract class SampleTransformer {
 
+  private static final int DOC_LINE_MAX_WIDTH = 80;
```
vchudnov-g: Should this be overridable by language? IIUC, the Java convention is 100 chars. Go isn't strict about line length, so we could also probably do 100 chars. Don't know about other languages.

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -1375,11 +1379,27 @@ public String getParamDocName(String var) {
...
+  }
+
   /** Converts the given text to doc lines in the format of the current language. */
   public List<String> getDocLines(String text) {
     return CommonRenderingUtil.getDocLines(commentReformatter.reformat(text));
   }
 
+  /**
+   * Converts the given text to doc lines and split at maxWidth. If reformat is true, text is also
+   * reformatted in the format of the current language.
```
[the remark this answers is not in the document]
vchudnov-g: Looking at https://github.com/googleapis/gapic-generator/blob/master/src/main/java/com/google/api/codegen/util/java/JavaCommentReformatter.java, I see reformatting  also includes, eg, escaping  special characters.

Are we guaranteed this reformatting only happens once, total, for each language? (eg all the `getParamDocText` already reformat....) If we reformat twice, the Java escape codes will be incorrect: `<` → `&lt;`→`&amp;lt;`, which would render as `&lt;` rather than `<` in JavaDoc. (tl;dr: reformatting is not idempotent)

Naïvely, it seems to me we would always format or never format..... It seems like in your usages, it's always false, so do we ever need it to be true? If not, I suggest removing the parameter, and then we don't have to worry about the idempotency issue.

src/test/java/com/google/api/codegen/gapic/testdata/nodejs/nodejs_library.baseline
```diff
@@ -459,9 +459,15 @@ sampleGetBigBook(argv.shelf);
...
 // [START hopper_core]
 
-
 const library = require('@google-cloud/library').v1;
 
+/**
+ * Testing resource name overlap
+ *
+ * @param shelf {string} Test word wrapping for long lines. This is a long
+ * comment. The name of the shelf to retrieve the big book from.
```
vchudnov-g: IMO, it would be nice to indent lines after the first in param descriptions (open to guidance from DevRel). I guess for Java we post-run the formatter, but what about other languages? And to properly account for line length with the indentation, we should probably indent the lines in the generator before wrapping them, even if we later run a formatter.

If it's not simple to do this now, let's file an issue and get back to it separately. I think this is a cosmetic issue we'll need to fix, but no need to block this PR.

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
```diff
@@ -340,9 +357,28 @@ private InitCodeContext createInitCodeContext(
...
               .readFromFile(attr.getReadFile())
               .sampleArgumentName(attr.getSampleArgumentName())
+              .description(attr.getDescription())
               .build();
       builder.put(identifier, config);
     }
     return builder.build();
   }
+
+  private ImmutableList<List<String>> paramDocLines(
```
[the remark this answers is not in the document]
hzyi-google: Yes.

src/main/java/com/google/api/codegen/viewmodel/SampleFunctionDocView.java
```diff
@@ -0,0 +1,46 @@
...
+import com.google.common.collect.ImmutableList;
+import java.util.List;
+
+@AutoValue
+public abstract class SampleFunctionDocView {
+
+  public abstract ImmutableList<List<String>> paramDocLines();
+
+  /** Description of the sample function. */
+  public abstract ImmutableList<String> mainDocLines();
```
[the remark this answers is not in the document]
hzyi-google: The naming is consistent with Api method documentations though: https://github.com/googleapis/gapic-generator/blob/96c3c5a4c8397d4bd29a6abce861547a271383e1/src/main/java/com/google/api/codegen/viewmodel/ApiMethodDocView.java#L24

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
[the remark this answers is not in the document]
hzyi-google: I mainly added `boolean reformat` because the other overload (a few lines above) always reformats the text, and if we don't pass in a flag here, it'd be confusing that these two methods have different behaviors.

src/test/java/com/google/api/codegen/gapic/testdata/nodejs/nodejs_library.baseline
hzyi-google: It's not simple to do, and also our clients do not have indentations when param descriptions take up more than one line: https://github.com/googleapis/gapic-generator/blob/96c3c5a4c8397d4bd29a6abce861547a271383e1/src/test/java/com/google/api/codegen/gapic/testdata/php/php_library.baseline#L1738.

So I thought it'd be OK.

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
hzyi-google: I didn't find any requirements by other languages. Let's just do 100 here for now? And we can always change it if some language has a different requirement in the future.
@beccasaurus WDYT?

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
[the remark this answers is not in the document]
vchudnov-g: I think then we're mixing semantics. All the other `getDocLines` unconditionally format and don't worry about `maxWidth`. Why does this method need to be an overload? It seems to me it's not used in the same situations as the others (is that right? you always call it with `reformat==false`). How about  calling `wrapDocLines(String doc, int maxWidth)` and saying `// Wraps the given formatted doc text at maxWidth columns`?

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
vchudnov-g: FWIW, Google style guides (internal and external)  for other languages, last time I checked, seem to go for an 80 char limit (eg https://github.com/google/styleguide/blob/gh-pages/pyguide.md#32-line-length)

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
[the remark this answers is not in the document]
vchudnov-g: Could you add the function comment?
And as per the comment above, could you pass the line length as a param here, so that the constant is only used in one function rather than two?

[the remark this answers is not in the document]
hzyi-google: Sorry forgot to push the local changes after commenting above :/ It will come right away.

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
hzyi-google: OMG this is so annoying :( Yeah I'll make this value override-able

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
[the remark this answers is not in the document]
hzyi-google: Ah I see.

I _think_ this is commonly used style in Java. Since overloading methods in Java is fairly easy and Java is verbose enough, it's very common to reuse method names (to describe the major task) with different parameters (to describe details), as the parameters are most of the time self-documenting. So it's common to have a method with only the most crucial parameters take care of the default / most common behavior, and have overloads take care of customized behaviors.

But it's just a style problem, not something that needs to be strictly enforced. In fact I think changing to `wrapDocLines` makes sense too and I'll just do that.

src/test/java/com/google/api/codegen/gapic/testdata/nodejs/nodejs_library.baseline
beccasaurus: Sooo.... you're going to hate me for this.

But I looked across the interwebs and it looks like pretty much, across the board, the wrapping lines for method documentation are indented (2 spaces should do)

e.g.
```js
/**
 * Testing resource name overlap
 *
 * @param shelf {string} Test word wrapping for long lines. This is a long
 *   comment. The name of the shelf to retrieve the big book from.
 * @param another
```

Don't hate me. But yeah. This ^ for all the langs.

beccasaurus: Per our chat: out of scope for this PR. This PR only needs to take one-liners into consideration.

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
```diff
@@ -251,6 +255,18 @@ private MethodSampleView generateSample(
...
         sampleImportTransformer().generateImportSection(methodContext);
+    SampleFunctionDocView sampleFunctionDocView =
+        SampleFunctionDocView.newBuilder()
+            .paramDocLines(paramDocLines(methodContext, initCodeView))
+            .mainDocLines(
+                ImmutableList.<String>builder()
+                    .addAll(
+                        methodContext
+                            .getNamer()
+                            .getDocLines(valueSet.getDescription(), DOC_LINE_MAX_WIDTH, false))
```
hzyi-google: @vchudnov-g Actually sorry, we do need the boolean flag `reformat` here. The `false` here should have been `true`, i.e.,  we need to reformat the function description here.

src/test/java/com/google/api/codegen/gapic/testdata/nodejs/nodejs_library.baseline
vchudnov-g: Oh, cool that they're indented (I think it's more readable, as per my earlier comment above). Weird that the one or two places I looked this morning don't have the continuation line indented.

I'm +1 for doing this, and I agree that doing it in a chaser makes sense. No need to block this one.

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
vchudnov-g: OK, then in the test cases, please add special characters to the appropriate sample config to ensure they get escaped correctly (ie once).  eg `// do foo & bar if a < 5 `

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
vchudnov-g: Humans are so picky. "What do you mean, 81 columns? Can't read!" Sigh....

Thanks for doing this! Feel free to do it in a chaser PR if it's simpler.

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
hzyi-google: Sure!

PR #2614, opened 2019-03-06
go: add pkg level doc for context use

### LINKED ISSUE #2612: Go: context should be documented at the package level of each gapic
Something like,

```
Context


NewClient and all client methods required a context. NewClient context is
used for authentication requests as well as gRPC transport setup, but not
for any future calls. Context passed to client methods are used for the
lifetime of the request.

We recommend using deadlines on requests. See more information at
https://golang.org/pkg/context/.
```

noahdietz:
Adds a package-level comment regarding the use of `context` in client initialization & method invocation.

Fixes #2612 

@jadekler can you give a +1 on the language. I copied it entirely, but tweaked things a tad.

### DIFF
--- a/src/main/resources/com/google/api/codegen/go/doc.snip
+++ b/src/main/resources/com/google/api/codegen/go/doc.snip
@@ -13,12 +13,12 @@
     //
     // Use of Context
     //
-    // The ctx passed to NewClient is used for authentication requests and gRPC
-    // transport setup, but not for future calls performed using the client.
-    // These calls use the ctx passed to individual methods on the client.
+    // The ctx passed to NewClient is used for authentication requests and
+    // for the underlying connection, which is shared between requests.
+    // Individual methods on the client use the ctx given to them.
     //
-    // Using deadlines on requests is recommended. See more information at
-    // https://golang.org/pkg/context/.
+    // For information about setting deadlines, reusing contexts, and more
+    // please visit godoc.org/cloud.google.com/go
     @if view.domainLayerLocation
         //
         // Use the client at {@view.domainLayerLocation} in preference to this.
--- a/src/main/resources/com/google/api/codegen/go/doc.snip
+++ b/src/main/resources/com/google/api/codegen/go/doc.snip
@@ -20,7 +20,7 @@
     // To close the open connection, use the Close() method.
     //
     // For information about setting deadlines, reusing contexts, and more
-    // please visit godoc.org/cloud.google.com/go
+    // please visit godoc.org/cloud.google.com/go.
     @if view.domainLayerLocation
         //
         // Use the client at {@view.domainLayerLocation} in preference to this.
--- a/src/main/resources/com/google/api/codegen/go/doc.snip
+++ b/src/main/resources/com/google/api/codegen/go/doc.snip
@@ -14,7 +14,7 @@
     // Use of Context
     //
     // The ctx passed to NewClient is used for authentication requests and
-    // for creating the underlying connection, but is not used across methods.
+    // for creating the underlying connection, but is not used for subsequent calls.
     // Individual methods on the client use the ctx given to them.
     //
     // To close the open connection, use the Close() method.
--- a/src/main/resources/com/google/api/codegen/go/doc.snip
+++ b/src/main/resources/com/google/api/codegen/go/doc.snip
@@ -14,8 +14,10 @@
     // Use of Context
     //
     // The ctx passed to NewClient is used for authentication requests and
-    // for the underlying connection, which is shared between requests.
+    // for creating the underlying connection, but is not used across methods.
     // Individual methods on the client use the ctx given to them.
+    //
+    // To close the open connection, use the Close() method.
     //
     // For information about setting deadlines, reusing contexts, and more
     // please visit godoc.org/cloud.google.com/go
--- a/src/main/resources/com/google/api/codegen/go/doc.snip
+++ b/src/main/resources/com/google/api/codegen/go/doc.snip
@@ -10,6 +10,16 @@
     @join line : view.packageDoc
         // {@line}
     @end
+    //
+    // Use of Context
+    //
+    // NewClient and all client methods require a context. NewClient context is
+    // used for authentication requests as well as gRPC transport setup, but not
+    // for any future calls. Context passed to client methods are used for the
+    // lifetime of the request.
+    //
+    // Using deadlines on requests is recommended. See more information at
+    // https://golang.org/pkg/context/.
     @if view.domainLayerLocation
         //
         // Use the client at {@view.domainLayerLocation} in preference to this.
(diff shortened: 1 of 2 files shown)


noahdietz:
@andreamlin not really sure what the circleci `compute-discovery-test` failure is. Any ideas? Or can it be ignored?

src/main/resources/com/google/api/codegen/go/doc.snip
```diff
@@ -10,6 +10,16 @@
     @join line : view.packageDoc
         // {@line}
     @end
+    //
+    // Use of Context
+    //
+    // NewClient and all client methods require a context. NewClient context is
```
jadekler: The ctx passed to NewClient is used for authentication requests and gRPC transport setup, but not for future calls performed using the client. These calls use the ctx passed to individual methods on the client.

src/main/resources/com/google/api/codegen/go/doc.snip
```diff
@@ -10,6 +10,15 @@
...
         // {@line}
     @end
+    //
+    // Use of Context
+    //
+    // The ctx passed to NewClient is used for authentication requests and gRPC
+    // transport setup, but not for future calls performed using the client.
+    // These calls use the ctx passed to individual methods on the client.
+    //
+    // Using deadlines on requests is recommended. See more information at
```
broady: Another common pattern is to use the `r.Context()` from HTTP handlers.

The context is canceled if the client goes away, and so with it any in-flight gRPC requests.

I think I'd prefer us link off to some other more detailed docs than provide a one-liner that might be taken to extremes (e.g., setting deadlines on every request, rather than a deadline for a series of requests).

src/main/resources/com/google/api/codegen/go/doc.snip
```diff
@@ -10,6 +10,15 @@
     @join line : view.packageDoc
         // {@line}
     @end
+    //
+    // Use of Context
+    //
+    // The ctx passed to NewClient is used for authentication requests and gRPC
+    // transport setup, but not for future calls performed using the client.
```
broady: Can we skip the "gRPC" and say something like "for the underlying connection, which is shared between requests".

src/main/resources/com/google/api/codegen/go/doc.snip
jadekler: > Another common pattern is to use the r.Context() from HTTP handlers.

I'm not grokking this. People use req.Context() as the input to our requests?

> The context is canceled if the client goes away, and so with it any in-flight gRPC requests.

What client? The http client? What does "go away" mean?

jadekler: > I think I'd prefer us link off to some other more detailed docs than provide a one-liner that might be taken to extremes (e.g., setting deadlines on every request, rather than a deadline for a series of requests).

This makes sense to me. It's definitely a nuanced topic.

broady: > I'm not grokking this. People use req.Context() as the input to our requests?

Yeah!

```
func handler(w http.ResponseWriter, r *http.Request) {
   ctx := r.Context()

   datastoreClient.Get(ctx, ...)
}
```

> What client? The http client? What does "go away" mean?

yep. e.g., if someone closed the tab or navigated away.

https://golang.org/pkg/net/http/#Request.Context

"For incoming server requests, the context is canceled when the client's connection closes, the request is canceled (with HTTP/2), or when the ServeHTTP method returns."

jadekler: Gotcha. Thanks! I don't think any of this precludes advice telling people to use deadlines. Folks should ideally do,

```
func handler(w http.ResponseWriter, r *http.Request) {
   ctx := context.WithDeadline(r.Context(), <something reasonable>)

   datastoreClient.Get(ctx, ...)
}
```

Apart from generally being good to set reasonable limits on requests, it's also good because of our retry policies (which have no timeout).

noahdietz: Can we take the more detailed advice/scenarios and include them in the repo-level documentation, then link back to that?

`For information about setting deadlines, reusing contexts, and more please visit godoc.org/cloud.google.com/go#hdr-Context`

jadekler: SGTM!

noahdietz: Actually, it would be better to link to documentation that is more general, and not cloud specific. This generator supports other Google APIs, so hard coding Cloud doc links into the comments might be undesirable.

noahdietz: Any suggestions on good sources to link to on this topic or should we be writing something ourselves? Might make for a good blog post...

andreamlin:
LGTM from me from the code repo perspective.

src/main/resources/com/google/api/codegen/go/doc.snip
```diff
@@ -10,6 +10,15 @@
     @join line : view.packageDoc
         // {@line}
     @end
+    //
+    // Use of Context
+    //
+    // The ctx passed to NewClient is used for authentication requests and
+    // for the underlying connection, which is shared between requests.
```
jadekler: // The ctx passed to NewClient is used for authentication requests and
    // for the underlying connection, which is shared between requests.

Wait - we need to be careful here. The context is passed to gRPC is _only used for setup_. https://godoc.org/google.golang.org/grpc#DialContext

We should not describe it as being used for the whole connection and all subsequent requests.

However, this might be the behavior for http. On the bus so bit unable to dig. cc @broady

broady: good catch. yes, all gapics are gRPC.

so yes, we should say it's only used for NewClient and they need to close the client if they want to shut connections.

noahdietz: PTAL

broady: "methods -> "subsequent calls" or "API calls" I think?
"methods" is a gRPC concept (and a poorly defined Go concept)

otherwise LGTM

jadekler: +1 chris' comments, otherwise LGTM.

noahdietz: Done. I will merge once CI finishes.

src/main/resources/com/google/api/codegen/go/doc.snip
```diff
@@ -10,6 +10,17 @@
...
+    // Use of Context
+    //
+    // The ctx passed to NewClient is used for authentication requests and
+    // for creating the underlying connection, but is not used for subsequent calls.
+    // Individual methods on the client use the ctx given to them.
+    //
+    // To close the open connection, use the Close() method.
+    //
+    // For information about setting deadlines, reusing contexts, and more
+    // please visit godoc.org/cloud.google.com/go
```
tbpg: Period at the end.

noahdietz:
@broady @jadekler FYI this change will be released today in artman 0.16.16

Issue #2519, opened 2019-01-17
Samplegen: we should turn off default output handling for RPCs that returns `Empty`


hzyi-google:
cc/ @vchudnov-g

eoogbe:
@hzyi-google What the status of this?

hzyi-google:
I haven't worked on it yet. Is it causing you any troubles?

vchudnov-g:
Lowering to P2, though I hope we get to this soon.

hzyi-google:
Fixed by #2617

PR #2617, opened 2019-03-08
SampleGen: Add missing response handling for methods without return value

### LINKED ISSUE #2616: SampleGen: Node.js delete sample doesn't show response handling section
Example:
 - [Sample definition](https://github.com/beccasaurus/gapic-docs-samples/blob/48ba35b04c9b915f5fb9bd4d9d2112db0875a601/google/cloud/talent/v4beta1/talent_gapic.yaml#L198-L216)
 - [Python sample](https://github.com/beccasaurus/gapic-docs-samples/blob/48ba35b04c9b915f5fb9bd4d9d2112db0875a601/code-samples/talent/v4beta1/python/delete_company_request_talent_delete_company.py#L46-L49) - OK
 - [PHP sample](https://github.com/beccasaurus/gapic-docs-samples/blob/48ba35b04c9b915f5fb9bd4d9d2112db0875a601/code-samples/talent/v4beta1/php/DeleteCompanyRequestTalentDeleteCompany.php#L37-L43) - OK _except there's a bug with the comment, take a look!_
 - [Java sample](https://github.com/beccasaurus/gapic-docs-samples/blob/48ba35b04c9b915f5fb9bd4d9d2112db0875a601/code-samples/talent/v4beta1/java/src/main/java/com/google/cloud/examples/talent/v4beta1/DeleteCompanyRequestTalentDeleteCompany.java#L30-L35) - OK
 - [Node.js sample](https://github.com/beccasaurus/gapic-docs-samples/blob/48ba35b04c9b915f5fb9bd4d9d2112db0875a601/code-samples/talent/v4beta1/nodejs/delete_company_request_talent_delete_company.js#L10-L18) - `<= the response body doesn't show up!!!`

#### Sample Definition

```yaml
- id: talent_delete_company
  description: Delete Company
  parameters:
    defaults:
    - name%project="Your Google Cloud Project ID CHANGED"
    - name%company="ID of the company to delete"
    attributes:
    - parameter: name%project
      sample_argument_name: project_id_changed
    - parameter: name%company
      sample_argument_name: company_id_changed
  on_success:
  - comment:
    - "What is going on?"
  - print:
    - "Hello???"
  - print:
    - "Deleted company"
```

#### Correct Output (Python)

```py
def sample_delete_company(project_id_changed, company_id_changed):
    """Delete Company"""

    # [START talent_delete_company_core]

    client = talent_v4beta1.CompanyServiceClient()

    # project_id_changed = 'Your Google Cloud Project ID CHANGED'
    # company_id_changed = 'ID of the company to delete'

    if isinstance(project_id_changed, six.binary_type):
        project_id_changed = project_id_changed.decode('utf-8')
    if isinstance(company_id_changed, six.binary_type):
        company_id_changed = company_id_changed.decode('utf-8')
    name = client.company_path(project_id_changed, company_id_changed)

    client.delete_company(name)
    # What is going on?
    print('Hello???')
    print('Deleted company')
```

#### Incorrect Output (Node.js - response handling missing)

```js
const talent = require('@google-cloud/talent').v4beta1;

function sampleDeleteCompany(projectIdChanged, companyIdChanged) {
  const client = new talent.CompanyServiceClient();
  // const projectIdChanged = 'Your Google Cloud Project ID CHANGED';
  // const companyIdChanged = 'ID of the company to delete';
  const formattedName = client.companyPath(projectIdChanged, companyIdChanged);
  client.deleteCompany({name: formattedName}).catch(err => {
    console.error(err);
  });
}
```

/cc @googleapis/samplegen @hzyi-google

hzyi-google:
Fix #2616

### DIFF
--- a/src/main/resources/com/google/api/codegen/java/standalone_sample.snip
+++ b/src/main/resources/com/google/api/codegen/java/standalone_sample.snip
@@ -59,17 +59,23 @@
   @switch sample.callingForm
   @case "Request"
     {@syncMethodSampleCode(apiMethod, sample)}
-    {@processResponse(sample)}
+    @if sample.outputs
+      {@processResponse(sample)}
+    @end
   @case "RequestPaged"
     {@pagedIterableMethodSampleCode(apiMethod, sample)}
   @case "Flattened"
     {@syncMethodSampleCode(apiMethod, sample)}
-    {@processResponse(sample)}
+    @if sample.outputs
+      {@processResponse(sample)}
+    @end
   @case "FlattenedPaged"
     {@pagedCallableMethodSampleCode(apiMethod, sample)}
   @case "Callable"
     {@callableMethodSampleCode(apiMethod, sample)}
-    {@processResponse(sample)}
+    @if sample.outputs
+      {@processResponse(sample)}
+    @end
   @case "CallableList"
     {@unpagedListCallableMethodSampleCode(apiMethod, sample)}
   @case "CallablePaged"
@@ -85,10 +91,14 @@
     {@processResponse(sample)}
   @case "LongRunningFlattenedAsync"
     {@asyncOperationMethodSampleCode(apiMethod, sample)}
-    {@processResponse(sample)}
+    @if sample.outputs
+      {@processResponse(sample)}
+    @end
   @case "LongRunningRequestAsync"
     {@asyncOperationMethodSampleCode(apiMethod, sample)}
-    {@processResponse(sample)}
+    @if sample.outputs
+      {@processResponse(sample)}
+    @end
   @default
     $unhandledCallingForm: {@sample.callingForm} in sample "{@apiMethod.getClass.getSimpleName}"$
   @end
@@ -158,7 +168,7 @@
         {@sampleMethodCallArgList(sample.sampleInitCode.fieldSettings)}).get();
   @else
     {@apiMethod.apiVariableName}.{@apiMethod.name}(\
-        {@sampleMethodCallArgList(initCode.fieldSettings)}).get();
+        {@sampleMethodCallArgList(sample.sampleInitCode.fieldSettings)}).get();
   @end
 @end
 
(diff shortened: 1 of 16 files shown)


hzyi-google:
cc/ @googleapis/samplegen

hzyi-google:
Oh wait a minute, didn't see the PHP comment bug. Working on that...

beccasaurus:
Could you add minimal repro cases so that each of these features has a corresponding test?

The way that gapic-generator tests work, I realize it's kind of annoying to have to add 1-2 samples on Library rpcs & then copy/paste the resulting samples in 4 languages into their respective .baseline's...

But I can't LGTM without a test (which in the case of gapic-generator is a .baseline)

Actually.... let me test this change locally and verify it fixes the Node.js missing response and the PHP command...

hzyi-google:
@vchudnov-g since we need to add tests, I decided to fix #2519 as well.

hzyi-google:
PTAL

src/main/resources/com/google/api/codegen/java/standalone_sample.snip
```diff
@@ -291,11 +301,7 @@
 @end
 
 @private processResponse(sample)
-  @if sample.outputs.size == 0
-    System.out.println(response);
-  @else
-    {@processOutputViews(sample.outputs)}
-  @end
+  {@processOutputViews(sample.outputs)}
```
vchudnov-g: For concision, consider always calling a variant of this, eg `printResponse` from up above and having the `if sample.outputs` inside it. That will get rid of a lot of repeated code that, right now, the caller is responsible for. (I realize you still need `processResponse` as is for exceptional cases like the one in l416.

Similarly for the other languages, where most calls are currently wrapped in an `if`. Make the caller have less responsibility ;-)

hzyi-google: Yes that's definitely a better approach. However, if we put the empty check inside a wrapper function, it will result in an empty line in the generated code if the `sample.outputs` is empty (idk why the `snip` files are evaluated like that, which totally does not make sense to me). Sometimes it's OK, but sometimes it's really annoying (like, a blank line between two closed curly braces :()

PR #2618, opened 2019-03-08
Add service level samples for Callable methods

andreamlin:
Fixes b/124763830.

For videointelligence v1p3beta1, the only method in one of the proto services is a [streaming RPC](https://github.com/googleapis/googleapis/blob/c986e1d9618ac41343962b353d136201d72626ae/google/cloud/videointelligence/v1p3beta1/video_intelligence.proto#L53). We just have to generate the service level sample for this kind of Callable generated method.


andreamlin:
PTAL

andreamlin:
Added a test case. There is code coverage on the source file diff.

PR #2619, opened 2019-03-08
Update python docs configuration file

busunkim96:
Updates template to match top-level conf.py in google-cloud-python


src/main/resources/com/google/api/codegen/py/docs/conf.py.snip
```diff
@@ -123,12 +128,20 @@
...
   
   @# Theme options are theme-specific and customize the look and feel of a theme
   @# further.  For a list of options available for each theme, see the
   @# documentation.
-  @#html_theme_options = {}
+  html_theme_options = {
+      @#'description': 'Google Cloud Client Libraries for Python',
+      @#'github_user': 'GoogleCloudPlatform',
+      @#'github_repo': 'google-cloud-python',
+      @#'github_banner': True,
```
tseaver: Seems like we could omit those four lines:  some are stale, and they get dropped anyway.

busunkim96: It looks like these options are used in the top-level conf.py, so I re-added them. The fields will need to be modified post repo split.

PR #2622, opened 2019-03-11
SampleGen: Print names of enum fields in PHP using Protobuf 3.7 helper method

hzyi-google:


hzyi-google:
PTAL

hzyi-google:
cc/ @googleapis/samplegen

beccasaurus:
Recommend adding [Do Not Merge] label (doesn't look like y'all have one) or add [DO NOT MERGE] to the title or... something :)

Having this in a branch is really nice so I can test this locally. Very very very very very much looking forward to this.

src/main/java/com/google/api/codegen/transformer/php/PhpSurfaceNamer.java
```diff
@@ -292,11 +293,22 @@ public String getMapKeyAccessorName(TypeModel keyType, String key) {
   }
 
   @Override
-  public String getFormattedPrintArgName(TypeModel type, String variable, List<String> accessors) {
+  public String getFormattedPrintArgName(
+      ImportTypeTable typeTable, TypeModel type, String variable, List<String> accessors) {
+    Preconditions.checkArgument(
+        !type.isRepeated() && !type.isMap(),
+        "Expected non-repeated fields in print statement, found %s",
```
vchudnov-g: nit: would s/non-repeated/scalar/ be more accurate?

hzyi-google: It seems protobuf only refers to string, numbers, bytes and bools as scalar fields.
But the field could be a nested message or an enum here.

vchudnov-g:
@hzyi-google Is this unblocked now?

hzyi-google:
@vchudnov-g Yes!

beccasaurus:
YAAAASSSSSSSSSS 🎉🎉🎉🎉🎉

beccasaurus:
Will we need to bump the protobuf requirement in PHP libraries to pick this up?
I'm guessing a regen will carry this over?

I'll try to play with this soon~ish!

hzyi-google:
I don't know actually... @michaelbausor Do you think we need to bump up protobuf runtime at all?

michaelbausor:
A regeneration with the latest artman will pick up the changes in the generated protobuf classes. Each library will also need to depend on the latest protobuf - but actually I think that is already done, as the dependency on gax-php has been bumped to 1.0 (here: https://github.com/googleapis/google-cloud-php/blob/master/composer.json#L53) and that requires protobuf 3.7.1 or above.

PR #2628, opened 2019-03-13
SampleGen: Ruby groundwork

hzyi-google:
- Set up infrastructure to generate Ruby standalone samples. Templates are not filled out yet.
- Refactor `DynamicLangApiMethodTransformer` since after Ruby, SampleGen for all dynamic languages are done. This way we don't need the `XLangMethodViewGenerators`, which are mostly same logic.
- Split the logic of getting matching calling forms for different type of RPC methods from `XLangMethodViewGenerator`. This will make implementing default calling forms easier too.
- Remove `PythonMethodViewGenerator` as well and use `DynamicLangApiMethodTransformer` directly.
A separate PR will take care of PHP and Node.js.


hzyi-google:
@vchudnov-g PTAL
@beccasaurus FYI, you don't have to review this PR since the baselines are still full of TODOs.
cc/ @googleapis/samplegen

src/test/java/com/google/api/codegen/gapic/testdata/ruby/ruby_library.baseline
```diff
@@ -4969,6 +4969,312 @@ Gem::Specification.new do |gem|
...
+//FIXME: license section
+
+# DO NOT EDIT! This is a generated sample ("RequestStreamingBidi",  "prog")
+# FIXME: import section
+# FIXME: sample function declaration
+# FIXME: initialize client
+
+name = "BASIC"
+file_name = "comment_file"
+comment = File.binread file_name
```
beccasaurus: Hehe, I love that... while 90%+ of the Ruby code isn't present... the reading local file feature is present and working.

Thanks for the ping, keep me in the loop!

src/main/java/com/google/api/codegen/viewmodel/CallingForm.java
```diff
@@ -60,12 +74,90 @@
...
 
+  private static enum RpcType {
+    UNARY,
+    LRO,
+    CLIENT_STREAMING,
+    SERVER_STREAMING,
+    BIDI_STREAMING,
+    PAGED_STREAMING;
+
+    static RpcType fromMethodContext(MethodContext context) {
```
vchudnov-g: nice!

src/main/java/com/google/api/codegen/viewmodel/CallingForm.java
```diff
@@ -60,12 +74,90 @@
...
+            throw new IllegalArgumentException(
+                "Illegal MethodContext: unhandled streaming type: " + streamingType);
+        }
+      }
+      return UNARY;
+    }
+  }
+
+  // TODO: Factor this out to a yaml file
+  private static final Table<TargetLanguage, RpcType, List<CallingForm>> CALLING_FORM_TABLE =
```
vchudnov-g: woo-hoo!

src/main/java/com/google/api/codegen/viewmodel/CallingForm.java
```diff
@@ -60,12 +74,90 @@
...
+          default:
+            throw new IllegalArgumentException(
+                "Illegal MethodContext: unhandled streaming type: " + streamingType);
+        }
+      }
+      return UNARY;
+    }
+  }
+
+  // TODO: Factor this out to a yaml file
```
vchudnov-g: +1

src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java
```diff
@@ -1694,6 +1694,13 @@ public String getSampleFunctionName(MethodModel method) {
     return getApiMethodName(Name.from("sample").join(method.asName()), VisibilityConfig.PRIVATE);
   }
 
+  // TODO(hzyi): this method doesn't fit very well in SurfaceNamer. So far we don't have a
+  // better place for this logic but consider move this out when we implement default
```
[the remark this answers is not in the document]
hzyi-google: The reason I don't think calling forms fit well here is that calling forms are not necessarily a `Name` on the `surface`, they are more like a `Context` for sample generation.

I don't have any concrete plans now unfortunately :(. But probably we can have something called `SampleContext` (something similar to `MethodContext` and `InterfaceContext`), and put them there?

hzyi-google:
@vchudnov-g Thanks for reviewing!

PR #2629, opened 2019-03-14
api-common-java dep set to 1.7.0

andreamlin:
Latest released version is [1.7.0.](https://github.com/googleapis/api-common-java/releases/tag/v1.7.0)


andreamlin:
PTAL

PR #2631, opened 2019-03-18
Generate unit tests for LRO discovery

andreamlin:


andreamlin:
PTAL

src/main/resources/com/google/api/codegen/java/http_test.snip
```diff
@@ -90,6 +90,46 @@
...
+    {@addException(test)}
+
+    try {
+      {@initCode(test.testCaseInitCode)}
+
+      client.{@test.clientMethodName}(\
+          {@sampleMethodCallArgList(test.testCaseInitCode.fieldSettings)}).get();
+      Assert.fail("No exception raised");
+    } catch (ExecutionException e) {
+      Assert.assertEquals(InvalidArgumentException.class, e.getCause().getClass());
```
vam-google: Maybe
```java
Assert.assertTrue(e.getCause() instanceof InvalidArgumentException);
```
is cleaner, and also allows subclasses of `InvalidArgumentException` as valid cause. This seems lika a desired behavior. In real world `catch(InvalidArgumentException e)` will also catch a subclass of `InvalidArgumentException` which is idiomatic for java exceptions, so the tests should treat `InvalidArgumentException` and its potential subclasses as equally correct.

hzyi-google:
@vam-google Do you think this PR is still needed?

chrisdunelm:
Old PR. Closing.

PR #2632, opened 2019-03-18
Generate LRO operation snapshot callable

andreamlin:
Analogous to gax's [GrpcOperationSnapshotCallable](https://github.com/googleapis/gax-java/blob/master/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcOperationSnapshotCallable.java)

The Compute Operation type is hard-coded right now, and will later be put in the GAPIC config.

### DIFF
--- a/src/main/resources/com/google/api/codegen/java/operation_callable.snip
+++ b/src/main/resources/com/google/api/codegen/java/operation_callable.snip
@@ -1,0 +1,49 @@
+@extends "java/common.snip"
+
+@snippet renderStubFileHeader(fileHeader)
+  {@license(fileHeader)}
+  package {@fileHeader.packageName};
+
+  {@importList(fileHeader.importSection.appImports)}
+@end
+
+@snippet generate(classFile)
+  {@renderStubFileHeader(classFile.fileHeader)}
+
+  {@classDoc()}
+  @@Generated("by gapic-generator")
+  @@BetaApi("The surface for long-running operations is not stable yet and may change in the future.")
+  {@renderClass(classFile.classView)}
+@end
+
+@private classDoc()
+  // AUTO-GENERATED DOCUMENTATION AND CLASS
+  /**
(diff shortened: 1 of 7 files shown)


andreamlin:
PTAL

src/main/resources/com/google/api/codegen/java/operation_callable.snip
```diff
@@ -0,0 +1,49 @@
...
+@private classDoc()
+  // AUTO-GENERATED DOCUMENTATION AND CLASS
+  /**
+   * A {@@code UnaryCallable} that wraps a UnaryCallable returning a
+   * long-running operation API resource and returns an OperationSnapshot instead.
+   */
+@end
+
+@private renderClass(classView)
+  public class OperationSnapshotCallable<RequestT>
```
vam-google: Do discogapic client classes have prefixes, indicating that they belong to discogapic client? For example, the analogous class in grpc is called `GrpcOperationSnapshotCallable`. There is a fundamental difference between the classes here of course (that `GrpcOperationSnapshotCallable` is api-independent, and shared by all clients, while this one is generated per discogapic client). Should we name this class `<ServiceName>OperationSnapshotCallable`? With current name, the class seems like a class, that `GrpcOperationSnapshotCallable` would want to extend (has more generic name)

hzyi-google:
@vam-google Do you think this PR is still needed?

vam-google:
Something like this will be needed if we want to support more discovery-based clients as @AlanGasperini suggests. 

I'm not sure how outdated it is (maybe not too much, because it is discovery-related and there hasn't been active development in this area recently). 

So honestly, I don't know. I believe this PR has useful non-trivial  logic that we will want to reuse if discogapic gets greater adoption. Can this PR be merged as is? Probably no. Can we close it? Probabl, yes, but we should not loose this work, because it has high chance of becoming valuable in mid-term future.

chrisdunelm:
Old PR. Closing.

PR #2635, opened 2019-03-19
Remove flattening_group_name

andreamlin:
This is not used anywhere in googleapis or googleapis-private or google3/


andreamlin:
PTAL

PR #2637, opened 2019-03-19
[configgen] Handle singular resource entities

### LINKED ISSUE #2633: Handle singular resources in gapic config generation
Gapic config generation does not name singular resources appropriately. For example, with `"/v1/{resource_name=customers/*/hotelPerformanceView}"`, we expect the entity_name to be `customer_hotel_performance_view`. Instead the entity_name is `customer`.

eoogbe:
Generates a different entity name for singular resources than for their 
parent collections in gapic config generation. A singular resource is a 
different resource entirely from the parent collection.

For example, "users/*" is a collection resource for users, while 
"users/*/profile" is a singular resource for a user profile. The entity 
name of the collection resource would be `user`, and the entity name of 
the singular resource is `user_profile`.

Prefixes the singular resource entity name with its parent collection. 
This runs the risk of redundancy, but is clearer when the singular 
resource depends on its parent context.

Maintains the simple entity names for the path templates. These names 
only refer to collection resources since singular resources don't have a 
corresponding wildcard value.

Fixes #2633


src/main/java/com/google/api/codegen/configgen/CollectionPattern.java
```diff
@@ -34,45 +38,136 @@ public static CollectionPattern create(FieldSegment fieldSegment) {
...
+
+  /**
+   * Builds an entity name for a wildcard path segment.
+   *
+   * @param pathSegments the full path containing the wildcard segment at wildcardIndex
+   * @param wildcardIndex the index of the wildcard segment
+   * @param prefix the prefix to the entity name. Should be the resource name.
+   * @return the entity name that corresponds to the wildcard segment
+   */
+  private static String buildEntityName(
```
michaelbausor: I found it quite hard to parse this section of code, because there are multiple functions with the same name that call each other. I have suggested some renamings that might clarify things a bit, let me know what you think.

Suggested renaming: `appendUnboundWildcardSuffix`

src/main/java/com/google/api/codegen/configgen/CollectionPattern.java
```diff
@@ -34,45 +38,136 @@ public static CollectionPattern create(FieldSegment fieldSegment) {
...
+   * Builds the entity prefix for a wildcard path segment.
+   *
+   * <p>The entity prefix is just the entity name for the collection resource without handling
+   * singular resources or unbounded wildcards.
+   *
+   * @param pathSegments the full path containing the wildcard segment at wildcardIndex
+   * @param wildcardIndex the index of the wildcard segment
+   * @return the entity prefix that corresponds to the wildcard segment
+   */
+  private static String buildSimpleEntityPrefix(List<PathSegment> pathSegments, int wildcardIndex) {
```
michaelbausor: Suggested renaming: `buildSimpleWildcardEntityName`

src/main/java/com/google/api/codegen/configgen/CollectionPattern.java
```diff
@@ -34,45 +38,136 @@ public static CollectionPattern create(FieldSegment fieldSegment) {
...
+  /**
+   * Builds an entity name for a wildcard path segment, handling singular resources.
+   *
+   * <p>The entity name of a singular resource will be prefixed by its parent collection resource.
+   *
+   * @param pathSegments the full path containing the wildcard segment at wildcardIndex
+   * @param wildcardIndex the index of the wildcard segment
+   * @return the entity name that corresponds to the wildcard segment
+   */
+  private static String buildEntityName(List<PathSegment> pathSegments, int wildcardIndex) {
```
michaelbausor: Suggested renaming: `buildLastWildcardEntityName` (since this is only called on the last wildcard in a template)

src/main/java/com/google/api/codegen/configgen/CollectionPattern.java
```diff
@@ -34,45 +38,136 @@ public static CollectionPattern create(FieldSegment fieldSegment) {
...
+  /**
+   * Builds the entity prefix for a wildcard path segment, handling singular resources.
+   *
+   * <p>The entity prefix for singular resources contains the entity name of the parent collection.
+   *
+   * @param pathSegments the full path containing the wildcard segment at wildcardIndex
+   * @param wildcardIndex the index of the wildcard segment
+   * @return the entity prefix that corresponds to the wildcard segment
+   */
+  private static String buildEntityPrefix(List<PathSegment> pathSegments, int wildcardIndex) {
```
michaelbausor: Can we roll this function into buildEntityName on line 95? It is only called from that one place.

src/main/java/com/google/api/codegen/configgen/CollectionPattern.java
```diff
@@ -34,45 +38,136 @@ public static CollectionPattern create(FieldSegment fieldSegment) {
...
+    String collectionEntityName = buildSimpleEntityPrefix(pathSegments, wildcardIndex);
+
+    // If there are an odd number of remaining segments, then the last segment is for a singular
```
michaelbausor: We only ever call this on the last wildcard in a template. So what about checking for `remainingSegments > 0` instead of for an even number of segments?

src/main/java/com/google/api/codegen/configgen/CollectionPattern.java
```diff
@@ -34,45 +38,136 @@ public static CollectionPattern create(FieldSegment fieldSegment) {
...
+      PathSegment segment, int index, List<PathSegment> pathSegments) {
+    if (!(segment instanceof WildcardSegment)) {
+      return segment;
+    }
 
-  private CollectionPattern(FieldSegment fieldSegment) {
-    this.fieldSegment = fieldSegment;
+    String prefix = buildSimpleEntityPrefix(pathSegments, index);
+    String entityName = buildEntityName(pathSegments, index, prefix);
+    if (((WildcardSegment) segment).isUnbounded()) {
```
michaelbausor: The prior function call to `buildEntityName` has only one purpose, to check for isUnbounded and append `_path`. WDYT about removing the call to `buildEntityName` from here and appending `_path` before constructing the LiteralSegment? This results in duplicating the logic to append `_path`, but unduplicates the checking for isUnbounded, and means that the function `buildEntityName` function could likely be completely removed. I'm not that committed to this change though, if you prefer as-is then that's fine.

src/main/java/com/google/api/codegen/configgen/CollectionPattern.java
```diff
@@ -34,45 +38,136 @@ public static CollectionPattern create(FieldSegment fieldSegment) {
...
+            .filter(index -> index >= 0)
+            .map(index -> buildEntityName(fieldSegment.getSubPath(), index))
+            .orElseThrow(() -> new IllegalStateException("Field segment contained no wildcards."));
+
+    List<PathSegment> templatizedSubpath =
+        Streams.mapWithIndex(
+                fieldSegment.getSubPath().stream(),
+                (segment, index) ->
+                    buildTemplatizedSegment(segment, (int) index, fieldSegment.getSubPath()))
+            .collect(toImmutableList());
```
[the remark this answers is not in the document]
eoogbe: Indentation is normal for lambda parameters

src/main/java/com/google/api/codegen/configgen/CollectionPattern.java
eoogbe: Done

src/main/java/com/google/api/codegen/configgen/CollectionPattern.java
eoogbe: Done

src/main/java/com/google/api/codegen/configgen/CollectionPattern.java
eoogbe: Done

src/main/java/com/google/api/codegen/configgen/CollectionPattern.java
eoogbe: Done

src/main/java/com/google/api/codegen/configgen/CollectionPattern.java
eoogbe: Done

src/main/java/com/google/api/codegen/configgen/CollectionPattern.java
eoogbe: Done

PR #2640, opened 2019-03-20
Generate all proto methods

andreamlin:
Generate all methods as defined in the API protos. 

The only (existing) exception to this is to not generate methods that have language-specific method visibilities (see `AddTag`).

Note: This is blocked on a release of artman and updates to the affected API configs that currently exclude methods from generation via exclusion of the GAPIC method config. The artman release build is [underway right now.](https://circleci.com/workflow-run/33aa7415-e644-402f-95cc-9edd0dd92d67)

Edit: All prerequisite tasks (release artman, update GAPIC configs) have been completed, so this is ready to be submitted.


andreamlin:
This PR cannot be merged until all of googleapis/googleapis is updated to exclude (set GAPIC method visibility to `DISABLED`) the proto methods that currently don't exist in their GAPIC configs. That itself must also wait for artman to be updated with the newest gapic-generator with the change that allows fully disabling methods in the GAPIC config.

andreamlin:
PTAL

PR #2641, opened 2019-03-21
SampleGen: generate comments on fields of request objects

hzyi-google:
- Configuration:
```yaml
- parameters:
  - defaults:
    ...
  - attributes:
    - parameter: some_parameter
      description: "some comments on the parameter"
    - parameter: another_parameter
      description: "this comment goes to sample function documentation"
      sample_argument_name: other_param
```
- Comments on resource name entities are not supported, since they are currently inlined. We can fix this when we start working on generating inline parameters.
- Also correctly generates default lines in Ruby.


hzyi-google:
PTAL @googleapis/samplegen

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -86,8 +86,14 @@
...
-  // false if the initialization code is being used to generate tests, rather than code samples.
+  // Whether the initialization code should include hardcoded comments like TODOs. This
+  // should only be true when generating incode samples.
+  //
+  // This field should be set to false when generating tests because tests do not need to include
+  // user facing comments.
+  //
+  // This field should be set to false when generating standalone samples because comments in
+  // standalone samples are derived from user configurations, not hardcoded.
   private final boolean generateUserFacingComments;
```
vchudnov-g: minor: Since we're talking about a distinction between these non-configurable comments and the ones that the user specifies, consider renaming this field and associated params. WDYT? Maybe `generateStandardComments`?

(I wonder what we meant originally by "user facing" comments. In the gapic-gen context, all comments we generate are user-facing....)

src/main/resources/com/google/api/codegen/java/initcode.snip
```diff
@@ -30,6 +30,30 @@
   @end
 @end
 
+@private initCodeLinesWithDescriptions(lines)
```
vchudnov-g: I suggest factoring out the `@switch` statement so that both `initCodeLines` and `initCodeLinesWithDescription` can call a single source of truth. Maybe `emitInitCodeLine` can be the function that holds the `@switch`?

Alternatively, we could have a single `initCodeLines(lines, printDescription)` with an extra arg. I don't like this quite as much because it's less informative at the call site.

This applies to all the languages.

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -86,8 +86,14 @@
 
   private final ImportSectionTransformer importSectionTransformer;
 
-  // Whether the initialization code should include user-facing comments like TODOs. This should be
-  // false if the initialization code is being used to generate tests, rather than code samples.
+  // Whether the initialization code should include hardcoded comments like TODOs. This
```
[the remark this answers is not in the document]
hzyi-google: Yes this makes sense.

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
hzyi-google: Sounds good.

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -86,8 +86,14 @@
...
+  // should only be true when generating incode samples.
+  //
+  // This field should be set to false when generating tests because tests do not need to include
```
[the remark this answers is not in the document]
hzyi-google: I think it's more like "may be set to true, but setting to false is better since comments in unit tests are unnecessary"?

hzyi-google:
@vchudnov-g PTAL

src/test/java/com/google/api/codegen/testsrc/common/library_gapic.yaml
```diff
@@ -307,6 +310,35 @@ interfaces:
...
+        - edition=123
+        attributes:
+        - parameter: shelf.internal_theme
+          description: "Comment on a primitive field"
+        - parameter: series_uuid.series_bytes
+          read_file: true
+          description: "Comment on a file input field"
+        - parameter: series_uuid
+          description: "Comment on a message"
+          # TODO(hzyi): this comment is currently not generated because resource name
```
beccasaurus: 👍

src/main/resources/com/google/api/codegen/ruby/initcode.snip
```diff
@@ -2,20 +2,7 @@
...
-    @case "MapInitLine"
-      {@initLineMap(line)}
-    @case "SimpleInitLine"
-      {@initLineSimple(line)}
-    @case "ReadFileInitLine"
-      {@initLineReadFile(line)}
-    @default
-      $unhandledCase: {@line.lineType.toString}$
-    @end
+    {@initCodeLine(line)}
```
vchudnov-g: Why does Ruby have this logic in two files (I just noticed!): `initcode.snip` and `standalone_sample.snip`? Is there anyway we can combine them into a SPOT?

(I'm fine doing this as a chaser PR, but if you're not doing it in this PR please file a issue so we remember)

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -86,9 +86,14 @@
 
   private final ImportSectionTransformer importSectionTransformer;
 
-  // Whether the initialization code should include user-facing comments like TODOs. This should be
-  // false if the initialization code is being used to generate tests, rather than code samples.
-  private final boolean generateUserFacingComments;
+  // Whether the initialization code should include non-configurable comments like TODOs. This
+  // should only be true when generating in-code samples.
+  //
+  // This field should be set to false when generating tests comments in unit tests are unnecessary.
```
vchudnov-g: s/tests comments/tests since comments/

src/main/resources/com/google/api/codegen/ruby/initcode.snip
hzyi-google: There are two helper methods to get a resource name in Ruby, which are handled by `initcode.snip`. Standalone samples prefer a different helper methods from in-code samples, and it's very hard to reuse the same `init_code` to generate them :(

Hopefully this will go away in the micro-generators, when we have a more idiomatic way to create resource names (from Earnest's design review)

PR #2642, opened 2019-03-21
Use "idempotent" and "non-idempotent" as default retry code def names 

andreamlin:
This will reduce the diff in client libraries once we migrate off of GAPIC config.

The default strings "idempotent" and "non-idempotent", used to refer to retry code settings for HTTP-GET and non-HTTP-GET methods respectively, will only be escaped if GAPIC config uses one of those strings to refer to a set of codes that differs from the corresponding default retry codes set.

This also preserves the mandate that C# will only generate retry code def types for those defined in the GAPIC config.


src/main/java/com/google/api/codegen/config/RetryCodesConfig.java
```diff
@@ -218,18 +227,39 @@ private void populateRetryCodesDefinitionWithProtoFile(
...
 
       String retryCodesName;
+      ImmutableList<String> retryCodesList;
       if (protoParser.isHttpGetMethod(method)) {
+        retryCodesList = RETRY_CODES_FOR_HTTP_GET;
         // It is a common case to have an HTTP GET method with no extra codes to retry on,
         // so let's put them all under the same retry code name.
-        retryCodesName = httpGetRetryName;
-        retryCodesDefinition.put(httpGetRetryName, RETRY_CODES_FOR_HTTP_GET);
+        if (!new HashSet<>(RETRY_CODES_FOR_HTTP_GET)
```
michaelbausor: Won't this check return the same result for every method? If so, can we move it outside the loop?

src/main/java/com/google/api/codegen/config/RetryCodesConfig.java
```diff
@@ -218,18 +227,39 @@ private void populateRetryCodesDefinitionWithProtoFile(
...
+          retryCodesName = RETRY_CODES_IDEMPOTENT_NAME;
+        }
+
       } else {
-        // It is a common case to have a method with no codes to retry on,
-        // so let's put these methods all under the same retry code name.
-        retryCodesName = noRetryName;
-        retryCodesDefinition.put(noRetryName, ImmutableList.of());
+        retryCodesList = ImmutableList.of();
+        if (retryCodesDefinition
```
michaelbausor: Same for this check?

src/main/java/com/google/api/codegen/config/RetryCodesConfig.java
andreamlin: :D

src/main/java/com/google/api/codegen/config/RetryCodesConfig.java
andreamlin: yep

andreamlin:
PTAL

src/main/java/com/google/api/codegen/config/RetryCodesConfig.java
```diff
@@ -203,10 +212,30 @@ private void populateRetryCodesDefinitionWithProtoFile(
 
     // Unite all HTTP GET methods that have no additional retry codes under one retry code name to
     // reduce duplication.
-    String httpGetRetryName = symbolTable.getNewSymbol(HTTP_RETRY_CODE_DEF_NAME);
+    String httpGetRetryName;
+    if (new HashSet<>(RETRY_CODES_FOR_HTTP_GET)
```
michaelbausor: Nit: personal preference here, but I think breaking this into multiple statements (i.e. declare hashsets, then compare them) is quite a lot clearer.

andreamlin: done

PR #2643, opened 2019-03-22
Paging methods must contain a page_size token to be a paged method

andreamlin:
See the List Pagination section under https://cloud.google.com/apis/design/design_patterns.


andreamlin:
PTAL

hzyi-google:
This shouldn't affect any production API right? All of them should have had all the required fields in their protos.
It seems to because baseline library API missed this field. How about adding `page_size` to relative protos and make it a correct paged streaming methods?

andreamlin:
This doesn't affect any production API right now. However, we currently don't generate `ListClusters` or `ListInstances` of Bigtable API because those methods aren't in the GAPIC config. If we did start generating them, the Bigtable team says they should _not_ be paginated methods, and if we use the critieria for pagination as proposed by this PR, then those methods will not be paginated, as expected.

I prefer leaving out `page_size` from the Library API method that is missing it because that's my test case haha

hzyi-google:
Ah that makes sense.

PR #2644, opened 2019-03-22
Generate header request params only from google.api.http

andreamlin:
And deprecate the `header_request_params` feature in config.proto. That feature can be removed after we remove all existing values in googleapis/googleapis, which can only be done after this PR is merged.

This has been tested on all APIs that use the `header_request_params` value in their GAPIC configs. The Java diffs come back clean; I don't imagine any other language being different.


andreamlin:
PTAL

src/main/java/com/google/api/codegen/config/GapicMethodConfig.java
```diff
@@ -272,15 +271,10 @@ static GapicMethodConfig createMethodConfig(
...
-  private static List<String> findHeaderRequestParams(
-      MethodConfigProto methodConfigProto, Method method, ProtoParser protoParser) {
-    List<String> headerRequestParams =
-        ImmutableList.copyOf(methodConfigProto.getHeaderRequestParamsList());
-
-    if (!headerRequestParams.isEmpty()) {
-      return headerRequestParams;
-    }
-    // Fetch header params from the proto annotations.
+  private static List<String> findHeaderRequestParams(Method method) {
```
michaelbausor: Suggestion - just remove this function? OK if you prefer to keep it though.

andreamlin:
@michaelbausor I'll bring it up in the team meeting.

PR #2645, opened 2019-03-22
Change test resources for header_request_params

andreamlin:
All APIs* have header_request_params in their GAPIC configs that are a subset of the header request params defined in the `google.api.http` annotations in the API proto files. 

This makes for a little smaller and mostly additive diff for PR https://github.com/googleapis/gapic-generator/pull/2644, where the implementation will change and pull from google.api.http instead of the GAPIC config.

* ~All APIs except for Bigquery Storage, which is being addressed in cl/239839135~ this is checked into googleapis/googleapis


andreamlin:
PTAL

PR #2646, opened 2019-03-22
use the latest showcase

alexander-fenster:
Let's use the latest showcase (v0.0.16) in gapic-generator!


alexander-fenster:
Current status: it now fails because of https://github.com/googleapis/gapic-showcase/pull/95, I will try with the new showcase when it is released.

alexander-fenster:
Attn: language leads

The updated showcase had some changes in `echo.proto` which makes the CI fail for all languages. Can I ask language leads to fix the CI for their corresponding language? Feel free to update this PR with the fixes, that way we'll be able to combine this showcase update.

I feel it's important to keep showcase up to date for several reasons: (a) using it as test (in both ways: running gapic-generator helps spotting problems in showcase, and vice versa), and (b) having a showcase baseline for micro-generators.

Cc: @andreamlin for Java, @michaelbausor for PHP, @lukesneeringer for Python, @jbolinger for Ruby, @noahdietz for Go.

andreamlin:
Java's tests are failing because of dependency conflicts in proto-google-common-protos, in which the latest gapic-generator needs exactly 1.13.0-pre3, but gapic-generator also needs the latest gax-java in order to work, and the latest gax-java says it wants 1.15.0. 

I _could_ release a new 1.15.0-pre1 common-protos with ResourceSets patched in, but it sounds like ResourceSets will finally make it into google3, and we can stop will all this patching business. I hear that will take only some days to happen, in which case I'll just hold off fix this until that happens.

andreamlin:
@alexander-fenster there are new Google common protos! Some languages, Java at least, will have to update this PR again.

landrito:
I'll take a look at this over the weekend.

PR #2647, opened 2019-03-23
 Remove GeneratedMessageV3 reffernces for the generated Java GAPIC clients

vam-google:
Also put back mistakenly removed javax_inject dependency.

This PR depends on https://github.com/googleapis/gax-java/pull/694


michaelbausor:
Why is this PR deleting 30,000 lines of baseline files?

andreamlin:
Vadym had PM'd me about the baseline files. I think four of them are not
currently being used, so we can delete them for now

PR #2650, opened 2019-03-25
Allow disabling method generation in GAPIC config

andreamlin:
If a method's GAPIC config visibility is marked as `DISABLED` for a specific language, treat that method as if it were not present in the GAPIC config at all:
- turn off all client generation for that method, including turning off method retry config generation, unit tests, boilerplate code, etc
- the only remaining artifacts related to that method will be the docs pages for the method's request and response types in Ruby and Nodejs (these are generated from the proto files directly, and do not factor in GAPIC config as an input). Because these are already generated for proto methods that aren't included in the GAPIC config, it's ok to keep generating them.

This PR is tested by marking the method visibility for mixin method `AddTag` from tagger.proto as `DISABLED` for all languages.

This PR is important because it allows us to by default generate all API methods from a protofile, and if we need to keep excluding methods from client generation, we can do so via the GAPIC config.


andreamlin:
PTAL

src/test/java/com/google/api/codegen/testsrc/common/library_gapic.yaml
```diff
@@ -760,6 +760,12 @@ interfaces:
     surface_treatments:
     - include_languages:
       - go
+      - java
```
michaelbausor: Do we still test, somewhere, a setting of DISABLED that does *not* include all languages in this list?

src/main/java/com/google/api/codegen/config/GapicInterfaceConfig.java
```diff
@@ -265,6 +265,9 @@ private static SmokeTestConfig createSmokeTestConfig(
     }
     List<T> methodConfigs = new ArrayList<>();
     for (MethodConfigProto methodConfigProto : interfaceConfigProto.getMethodsList()) {
+      if (methodConfigMap.get(methodConfigProto.getName()) == null) {
```
michaelbausor: Can we check for containsKey here?

src/test/java/com/google/api/codegen/testsrc/common/library_gapic.yaml
andreamlin: Done. Set only GO to be disabled for the `GetBookFromArchive` method (that method is primary there for testing resource names, for which Go doesn't care about).

src/main/java/com/google/api/codegen/config/GapicInterfaceConfig.java
andreamlin: yep.

andreamlin:
@michaelbausor Great question! 
So `DISABLED` is only ever used by Go, Java, and C#. 
- C#: the disabled method currently has no GAPIC artifacts (it does have protoc- and grpc- generated artifacts), so this PR wouldn't remove any further GAPIC artifacts
- Java: the disabled method does have a mock test method generated, but @igorbernstein2 has said that we can remove test resources from the generated clients
- Go: like Java, the disabled methods do have mock methods generated. Sent an email to @jadekler and @noahdietz  to see if it's ok to remove those test methods.

andreamlin:
Update: We should be able to remove test resources from the existing Go clients that have DISABLED methods, so this PR won't terribly affect any existing clients.

andreamlin:
PTAL

PR #2651, opened 2019-03-26
Add @Deprecated to formatString[ResourceName] functions in Java

andreamlin:
These formatString functions are causing us pain when we try to refresh clients.

Statically-typed resource name classes are superior to the `formatString*` functions, and it is more cluttering than helpful to have both the classes and the formatting functions. We might as well just always use the resource name classes.


andreamlin:
PTAL

@michaelbausor What are your thoughts on formatting string functions for resource names?

andreamlin:
@michaelbausor yes

PR #2656, opened 2019-03-27
Treat UNSET_TREATMENT resource_name_type as STATIC_TYPES

andreamlin:
And if not STATIC_TYPES, then VALIDATE if possible.

In a proto-options-only world, we will treat all resource_name_types as STATIC_TYPES if possible. This PR will remove the need to explicitly give the STATIC_TYPES value in the GAPIC config during the migration process.

After this, we can remove all `resource_name_type` values in existing GAPIC configs if we want.

### DIFF
--- a/src/test/java/com/google/api/codegen/discogapic/testdata/java/java_simplecompute.v1.json.baseline
+++ b/src/test/java/com/google/api/codegen/discogapic/testdata/java/java_simplecompute.v1.json.baseline
@@ -8928,9 +8928,9 @@ public class AddressClient implements BackgroundResource {
    * Sample code:
    * <pre><code>
    * try (AddressClient addressClient = AddressClient.create()) {
-   *   ProjectName project = ProjectName.of("[PROJECT]");
+   *   String formattedProject = ProjectName.format("[PROJECT]");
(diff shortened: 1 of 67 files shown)


andreamlin:
PTAL

src/test/java/com/google/api/codegen/discogapic/testdata/java/java_simplecompute.v1.json.baseline
```diff
@@ -8928,9 +8928,9 @@ public class AddressClient implements BackgroundResource {
    * Sample code:
    * <pre><code>
    * try (AddressClient addressClient = AddressClient.create()) {
-   *   ProjectName project = ProjectName.of("[PROJECT]");
+   *   String formattedProject = ProjectName.format("[PROJECT]");
```
michaelbausor: Are we expecting these changes in the compute baselines? Will this affect the actual compute client?

andreamlin: Yes, these changes will be in the actual compute client too.

No, these don't affect the client because (1) they are only comment changes and (2) they have equivalent behaviors.

PR #2657, opened 2019-03-27
Deprecate copyright_file in GAPIC config

andreamlin:
Don't generate `copyright_file` in configgen either.

We still have to be able to handle GAPIC configs that contain the `copyright_file` field, and we can't remove it while we're in a prod freeze.


andreamlin:
PTAL

PR #2662, opened 2019-03-28
SampleGen: update yargs to specify explicit types in Node.js

### LINKED ISSUE #2661: SampleGen: Node.js update yargs to specify explicit types
Please update this:

```js
const argv = require(`yargs`)
  .default('project_id', 'Your Google Cloud Project ID')
  .option('job_id', 'Job ID')
  .argv
```

To this:

```js
const argv = require(`yargs`)
  .option('project_id', {
    default: 'Your Google Cloud Project ID',
    string: true
  })
  .option('job_id', {
    default: 'Job ID',
    string: true
  })
  .argv
```

For numeric fields, set `number: true` instead.

----

How I discovered this is a problem –

I recently had a problem where I was constructing a Job Path (`projects/*/jobs/*`) and, instead of the Job ID being a UUID, it was actually a number.

So `node get_job.js --project_id=$GCLOUD_PROJECT --job_id=1234567890` resulted in:  
`const formattedName = client.jobPath("...", 1234567890);`

And the path formatter was very angry. Because 1234567890 is not a string.
`TypeError: input.charCodeAt is not a function`

@googleapis/samplegen

hzyi-google:
fix #2661.

### DIFF
--- a/src/main/resources/com/google/api/codegen/nodejs/standalone_sample.snip
+++ b/src/main/resources/com/google/api/codegen/nodejs/standalone_sample.snip
@@ -113,7 +113,10 @@
 @private processCliArguments(sample)
   const argv = require(`yargs`)
     @join param : sample.sampleInitCode.argDefaultParams
-      .default('{@param.cliFlagName}', {@renderInitValue(param.initValue)})
+      .option('{@param.cliFlagName}', {
+        default: {@renderInitValue(param.initValue)},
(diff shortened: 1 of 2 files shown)


hzyi-google:
PTAL @googleapis/samplegen @alexander-fenster

hzyi-google:
@beccasaurus That was fast!

PR #2664, opened 2019-03-28
Remove proto_version from build.gradle; update protobuf-gradle plugin

andreamlin:
`proto_version` doesn't seem to be used anywhere. It has been replaced by the value in dependencies.properties.

Update protobuf gradle plugin to latest release. This does require a minimum Java source version of 1.8.


andreamlin:
PTAL

andreamlin:
@vam will it be ok for gapic-generator's gradle build to rely on Java 8?

showcase/ruby/Gemfile.lock
```diff
@@ -11,7 +11,7 @@ GEM
       googleauth (~> 0.6.2)
       grpc (>= 1.7.2, < 2.0)
       rly (~> 0.2.3)
-    google-protobuf (3.6.1-universal-darwin)
+    google-protobuf (3.7.1-universal-darwin)
```
andreamlin: Pretty sure this is a legit version: https://rubygems.org/gems/google-protobuf/versions/3.7.1-universal-darwin

dependencies.properties
```diff
@@ -7,7 +7,7 @@
 # Target workspace name: com_google_api_codegen
 
 # Versions only, for dependencies which actual artifacts differ between Bazel and Gradle
-version.com_google_protobuf=3.6.1
+version.com_google_protobuf=3.7.0
```
vam-google: Can we depend on 3.7.1 instead? Protobuf 3.7.0 had some bugs for PHP. Even if gapic-generator does not really care about this it would be good to have the protobuf version consistent across our repositories. Also this version will be used by bazel build not only to build gapic-genrator itself but also to generate the actual stubs in each language (this is where PHP problems become relevant).

andreamlin: oh yes

vam-google:
@andreamlin Oh, the build fails, because there is no 3.7.1 maven artifact released for some reason (https://mvnrepository.com/artifact/com.google.protobuf/protobuf-java/3.7.0)

andreamlin:
@vam-google yeah i think yoshi had the same problem, so google-cloud-java uses 3.7.0 instead of 3.7.1.
So, should we revert to 3.7.0? or like have different versions for the gapic-generator gradle vs bazel?

andreamlin:
asking protobuf team to make a java release too protocolbuffers/protobuf#5972

andreamlin:
@vam-google the bazel build definitely fails using this PR. in fact, `bazel test ...` fails locally if i checkout master branch and then change the protobuf version to [3.7.1 in dependencies.properties](https://github.com/googleapis/gapic-generator/blob/b05468c1958cd804f47f53915b200d7e02f7dde9/dependencies.properties#L10). 

What should we do?

PR #2665, opened 2019-03-28
Add Gapic config v2

andreamlin:
Copy config.proto into config_v2.proto, in a new v2 package.

The target branch, origin/gapic_config_v2, is currently just a copy of the master branch. This PR will be checked in to origin/gapic_config_v2 while we bootstrap.


andreamlin:
PTAL

andreamlin:
@michaelbausor Correct, it's the same exact file, just different name and different location.

PR #2666, opened 2019-03-28
Whittling down config_v2

andreamlin:
Redefining config_v2.proto to be a subset of config.proto.

Note: this will _not_ be checked into `origin/master`.


andreamlin:
PTAL

andreamlin:
@vchudnov-g PTAL too!

andreamlin:
I'm going to merge this as it doesn't remove anything that we haven't collectively decided to remove. There are things that we can discuss removing, and we can do that later.

Issue #2668, opened 2019-03-28
Toggle off header request parameters for client/bidi streaming


andreamlin:
We hadn't supported header request parameters for client/bidi streaming in the first place. Therefore, trying to parse the header request parameters from the `google.api.http` annotation for client/bidi streaming methods breaks at least PHP, Golang, and maybe nodejs.

andreamlin:
In PHP, Golang, and nodejs, the generated code expects a method parameter called `req` (or similar) to exist, from which the parameters are parsed from.

PR #2669, opened 2019-03-29
Don't use header request parameters for client/bidi streaming

andreamlin:
Fixes https://github.com/googleapis/gapic-generator/issues/2668.

### DIFF
--- a/src/test/java/com/google/api/codegen/gapic/testdata/go/go_library.baseline
+++ b/src/test/java/com/google/api/codegen/gapic/testdata/go/go_library.baseline
@@ -760,8 +760,7 @@ func (c *LibClient) StreamBooks(ctx context.Context, req *librarypb.StreamBooksR
 
 // DiscussBook test bidi-streaming.
 func (c *LibClient) DiscussBook(ctx context.Context, opts ...gax.CallOption) (librarypb.LibraryService_DiscussBookClient, error) {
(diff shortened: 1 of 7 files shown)


src/test/java/com/google/api/codegen/gapic/testdata/go/go_library.baseline
```diff
@@ -777,8 +776,7 @@ func (c *LibClient) DiscussBook(ctx context.Context, opts ...gax.CallOption) (li
 
 // MonologAboutBook test client streaming.
 func (c *LibClient) MonologAboutBook(ctx context.Context, opts ...gax.CallOption) (librarypb.LibraryService_MonologAboutBookClient, error) {
-    md := metadata.Pairs("x-goog-request-params", fmt.Sprintf("%s=%v", "name", req.GetName()))
```
andreamlin: This code had been broken because it expects a variable named `req`, which is not in the method arguments.

src/test/java/com/google/api/codegen/gapic/testdata/php/php_library.baseline
```diff
@@ -1671,13 +1671,6 @@ class LibraryServiceGapicClient
      */
     public function discussBook(array $optionalArgs = [])
     {
-        $requestParams = new RequestParamsHeaderDescriptor([
-          'name' => $request->getName(),
```
andreamlin: I'm guessing that this had also been broken because there is no var `$request`

PR #2671, opened 2019-03-29
Tag artman to 0.16.21

hzyi-google:


michaelbausor:
Don't we need to release artman 0.16.22 with the latest gapic-generator fix?

andreamlin:
@michaelbausor Yah i'll take care of the 0.16.22 release

andreamlin:
Hmm this might be failing because we're using protobuf-java:3.7 now. Yoshi-java team had to change some stuff to make protobuf 3.7 work

andreamlin:
@hzyi-google Do think that changing the generated protobuf dep in the dependencies.yaml file, line 69, to 3.7.0 (note 3.7.1 doesn't exist for java) will fix this?

hzyi-google:
I just changed the version here in this PR but not sure whether it would work :/ but we'll see :)
We probably need a new artman first?

PR #2672, opened 2019-03-29
Add ConfigV2 Validator

andreamlin:
+ tests

### DIFF
--- a/src/main/java/com/google/api/codegen/util/ConfigNextVersionValidator.java
+++ b/src/main/java/com/google/api/codegen/util/ConfigNextVersionValidator.java
@@ -1,0 +1,75 @@
+/* Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.codegen.util;
+
+import com.google.protobuf.InvalidProtocolBufferException;
+import com.google.protobuf.util.JsonFormat;
+import com.google.protobuf.util.JsonFormat.Parser;
+import com.google.protobuf.util.JsonFormat.Printer;
+import com.google.protobuf.util.JsonFormat.TypeRegistry;
+import javax.annotation.Nonnull;
+
+public class ConfigNextVersionValidator {
+
+  public static String CONFIG_V2_MAJOR_VERSION = "2";
+  public static String CONFIG_V2_VERSION = CONFIG_V2_MAJOR_VERSION + ".0.0"; // "2.0.0"
+
+  public void checkIsNextVersionConfig(@Nonnull com.google.api.codegen.ConfigProto configV1Proto)
+      throws IllegalStateException {
+    if (!configV1Proto.getConfigSchemaVersion().startsWith(CONFIG_V2_MAJOR_VERSION + ".")
+        && !configV1Proto.getConfigSchemaVersion().equals(CONFIG_V2_MAJOR_VERSION)) {
+      throw new IllegalStateException(
+          String.format(
+              "Provided ConfigProto version is %s but should be >= %s",
+              configV1Proto.getConfigSchemaVersion(), CONFIG_V2_VERSION));
+    }
+
+    try {
+      // Serializing utils for Config V2.
+      TypeRegistry typeRegistryV2 =
+          JsonFormat.TypeRegistry.newBuilder()
+              .add(com.google.api.codegen.v2.ConfigProto.getDescriptor())
+              .build();
+      Parser parserV2 = JsonFormat.parser().usingTypeRegistry(typeRegistryV2);
+      Printer printerV2 = JsonFormat.printer().usingTypeRegistry(typeRegistryV2);
+
+      // Serialize and deserialize the Config v1 proto under the Config v2 schema to remove fields
+      // unknown to Config v2 schema.
+      com.google.api.codegen.v2.ConfigProto.Builder v2ProtoBuilder =
+          com.google.api.codegen.v2.ConfigProto.newBuilder();
+      parserV2.merge(printerV2.print(configV1Proto), v2ProtoBuilder);
+
+      // Serializing utils for Config V2.
+      TypeRegistry typeRegistryV1 =
+          JsonFormat.TypeRegistry.newBuilder()
+              .add(com.google.api.codegen.ConfigProto.getDescriptor())
+              .build();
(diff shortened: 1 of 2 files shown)


andreamlin:
PTAL

src/main/java/com/google/api/codegen/util/ConfigNextVersionValidator.java
```diff
@@ -0,0 +1,79 @@
...
+package com.google.api.codegen.util;
+
+import com.google.protobuf.InvalidProtocolBufferException;
+import com.google.protobuf.util.JsonFormat;
+import com.google.protobuf.util.JsonFormat.Parser;
+import com.google.protobuf.util.JsonFormat.Printer;
+import com.google.protobuf.util.JsonFormat.TypeRegistry;
+import javax.annotation.Nonnull;
+
+public class ConfigNextVersionValidator {
```
michaelbausor: This name is a bit odd. What about just ConfigVersionValidator?

src/main/java/com/google/api/codegen/util/ConfigNextVersionValidator.java
```diff
@@ -0,0 +1,79 @@
...
+        && !configV1Proto.getConfigSchemaVersion().equals(CONFIG_V2_MAJOR_VERSION)) {
+      throw new IllegalStateException(
+          String.format(
+              "Provided ConfigProto version is %s but should be >= %s",
+              configV1Proto.getConfigSchemaVersion(), CONFIG_V2_VERSION));
+    }
+
+    try {
+      // Serializing utils for Config V2.
+      TypeRegistry typeRegistryV2 =
```
michaelbausor: Is it necessary to use a TypeRegistry and json formatting functions? Can we not just use toByteArray, parseFrom(byte[]), and check for unknown fields?

src/main/java/com/google/api/codegen/util/ConfigNextVersionValidator.java
```diff
@@ -0,0 +1,79 @@
...
+public class ConfigNextVersionValidator {
+
+  public static String CONFIG_V2_MAJOR_VERSION = "2";
+  public static String CONFIG_V2_VERSION = CONFIG_V2_MAJOR_VERSION + ".0.0"; // "2.0.0"
+
+  /**
+   * Throw {@link IllegalStateException} iff the given input contains fields unknown to the {@link
+   * com.google.api.codegen.v2.ConfigProto} schema.
+   */
+  public void checkIsNextVersionConfig(@Nonnull com.google.api.codegen.ConfigProto configV1Proto)
```
michaelbausor: What about something like `validateV2Config`? I don't think it is desirable to make the function name version-agnostic (unless the implementation is also version-agnostic)

src/main/java/com/google/api/codegen/util/ConfigNextVersionValidator.java
andreamlin: sure

src/main/java/com/google/api/codegen/util/ConfigNextVersionValidator.java
andreamlin: ok!

src/main/java/com/google/api/codegen/util/ConfigNextVersionValidator.java
andreamlin: If we use the toByteArray and parseFrom functions, those will still [preserve the unknown fields](https://github.com/protocolbuffers/protobuf/issues/272#issuecomment-405729394). The thing with the unknown fields is that they can exist at every single level of the protobuf object tree, and I don't feel like writing a tree search function to parse through every child message type and looking for unknown types (i've already been burned trying to implement a BFS haha)

andreamlin:
asking protobuf team to make a java release too https://github.com/protocolbuffers/protobuf/issues/5972

src/main/java/com/google/api/codegen/util/ConfigNextVersionValidator.java
michaelbausor: What about using a [DiscardUnknownFieldsParser](https://github.com/protocolbuffers/protobuf/blob/master/java/core/src/main/java/com/google/protobuf/DiscardUnknownFieldsParser.java#L43)? I don't know, but my intuition is that there is an almost-one-liner that does this, something like:
```
configBytes = config.toByteArray();
configV2Bytes = DiscardUnknownFieldsParser.wrap(ConfigV2.parser())
  .parseFrom(configBytes)
  .toByteArray();
// compare byte arrays ...
```

andreamlin:
renamed things
PTAL

src/main/java/com/google/api/codegen/util/ConfigNextVersionValidator.java
michaelbausor: But - maybe I'm just wrong? Like, does the above fail because of unknown fields other than v1/v2 differences?

andreamlin: oh i didn't see that! i heard that Go had the same thing, but i didn't look hard enough for a java version

andreamlin: much shorter now :) i must have been looking at stackoverflow posts that are outdated now for help on this

andreamlin:
PTAL

src/main/java/com/google/api/codegen/util/ConfigVersionValidator.java
```diff
@@ -0,0 +1,62 @@
...
+    if (!configV1Proto.getConfigSchemaVersion().startsWith(CONFIG_V2_MAJOR_VERSION + ".")
+        && !configV1Proto.getConfigSchemaVersion().equals(CONFIG_V2_MAJOR_VERSION)) {
+      throw new IllegalStateException(
+          String.format(
+              "Provided ConfigProto version is %s but should be >= %s",
+              configV1Proto.getConfigSchemaVersion(), CONFIG_V2_VERSION));
+    }
+
+    try {
+
```
michaelbausor: Nit: remove extra line

src/main/java/com/google/api/codegen/util/ConfigVersionValidator.java
```diff
@@ -0,0 +1,62 @@
...
+      Parser<com.google.api.codegen.v2.ConfigProto> parser =
+          DiscardUnknownFieldsParser.wrap(com.google.api.codegen.v2.ConfigProto.parser());
+      com.google.api.codegen.v2.ConfigProto configV2 =
+          parser.parseFrom(configV1Proto.toByteString());
+
+      // Compare the v1-serialized and v2-serialized strings of the same config proto object.
+      if (!Arrays.equals(configV2.toByteArray(), configV1Proto.toByteArray())) {
+        throw new IllegalStateException(
+            String.format(
+                "Unknown fields in to ConfigProto v2 in configProto: %s",
```
michaelbausor: s/in to ConfigProto/to ConfigProto/

andreamlin: done

src/main/java/com/google/api/codegen/util/ConfigVersionValidator.java
andreamlin: done

PR #2674, opened 2019-03-29
Deprecate license_file and license_header

andreamlin:
We got OK to make all generated artifacts Apache now.


andreamlin:
PTAL

src/main/java/com/google/api/codegen/util/LicenseHeaderUtil.java
```diff
@@ -15,88 +15,35 @@
...
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
-import javax.annotation.Nullable;
 
 public class LicenseHeaderUtil {
-  public static final String DEFAULT_LICENSE_FILE = "license-header-apache-2.0.txt";
-  public static final String DEFAULT_COPYRIGHT_FILE = "copyright-google.txt";
+  @VisibleForTesting static final String DEFAULT_LICENSE_FILE = "license-header-apache-2.0.txt";
+  static final String DEFAULT_COPYRIGHT_FILE = "copyright-google.txt";
```
michaelbausor: Should both constants be annotated? And, should the annotation be on the line above?

andreamlin: will annotate both! google-java-format put the annotation in front instead of above

PR #2675, opened 2019-04-01
Remove resource_name_treatment from testing and configgen

andreamlin:
`resource_name_treatment` in the wild is always `STATIC_TYPES` or not set; gapic-generator sets the default value is `STATIC_TYPES` if it is possible.

With proto options, there will not be an option to turn off resource name generation, assuming the resource names are already annotated.

Also configV2 doesn't contain this field, and it needs to be removed for us to do the config party.


andreamlin:
PTAL

PR #2677, opened 2019-04-01
Parse fixed resource names from regular CollectionConfigProto

andreamlin:
Deprecate FixedResourceNameValueProto; it will still be supported because Pubsub needs it right now before the change is made.

Any resource name entities, defined in CollectionConfigProto, that do not have params in their pattern will be treated as Fixed resource names.

### DIFF
--- a/src/main/java/com/google/api/codegen/config/FixedResourceNameConfig.java
+++ b/src/main/java/com/google/api/codegen/config/FixedResourceNameConfig.java
@@ -61,4 +61,12 @@ public abstract class FixedResourceNameConfig implements ResourceNameConfig {
 
     return new AutoValue_FixedResourceNameConfig(entityName, entityName, fixedValue, file);
   }
+
+  /**
+   * Returns if the pathPattern is a fixed name resource. This primitively returns true iff the
+   * pathPattern contains a '{' char.
+   */
+  public static boolean isFixedResourceNameConfig(String pathPattern) {
+    return !pathPattern.contains("{");
+  }
 }
--- a/src/main/java/com/google/api/codegen/config/FixedResourceNameConfig.java
+++ b/src/main/java/com/google/api/codegen/config/FixedResourceNameConfig.java
@@ -14,6 +14,7 @@
  */
 package com.google.api.codegen.config;
 
+import com.google.api.codegen.FixedResourceNameValueProto;
 import com.google.api.tools.framework.model.Diag;
 import com.google.api.tools.framework.model.DiagCollector;
 import com.google.api.tools.framework.model.ProtoFile;
@@ -47,6 +48,30 @@ public abstract class FixedResourceNameConfig implements ResourceNameConfig {
   @Nullable
   public static FixedResourceNameConfig createFixedResourceNameConfig(
       DiagCollector diagCollector, String entityName, String fixedValue, @Nullable ProtoFile file) {
+
+    if (entityName == null || fixedValue == null) {
+      diagCollector.addDiag(
+          Diag.error(
+              SimpleLocation.TOPLEVEL,
+              "incorrectly configured FixedResourceNameConfig: name: "
+                  + entityName
+                  + ", value: "
+                  + fixedValue));
+      return null;
+    }
+
+    return new AutoValue_FixedResourceNameConfig(entityName, entityName, fixedValue, file);
+  }
+
+  // TODO(andrealin): Remove this method once all existing fixed resource names are removed.
+  @Nullable
+  public static FixedResourceNameConfig createFixedResourceNameConfig(
+      DiagCollector diagCollector,
+      FixedResourceNameValueProto fixedResourceNameValueProto,
+      @Nullable ProtoFile file) {
+
+    String entityName = fixedResourceNameValueProto.getEntityName();
+    String fixedValue = fixedResourceNameValueProto.getFixedValue();
 
     if (entityName == null || fixedValue == null) {
       diagCollector.addDiag(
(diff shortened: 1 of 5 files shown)


andreamlin:
PTAL

andreamlin:
PTAL

src/main/java/com/google/api/codegen/config/FixedResourceNameConfig.java
```diff
@@ -45,6 +45,25 @@ public ResourceNameType getResourceNameType() {
     return ResourceNameType.FIXED;
   }
 
+  @Nullable
+  public static FixedResourceNameConfig createFixedResourceNameConfig(
+      DiagCollector diagCollector, String entityName, String fixedValue, @Nullable ProtoFile file) {
+
+    if (entityName == null || fixedValue == null) {
```
michaelbausor: Rather than having this code be duplicated, can we have the original createFixedResourceNameConfig function call this one?

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -860,34 +875,53 @@ private static GapicProductConfig createDummyInstance(
...
-    }
-    for (InterfaceConfigProto interfaceConfigProto : configProto.getInterfacesList()) {
-      for (CollectionConfigProto collectionConfigProto :
-          interfaceConfigProto.getCollectionsList()) {
+    for (CollectionConfigProto collectionConfigProto : allCollectionConfigProtos) {
+      PathTemplate pathTemplate = PathTemplate.create(collectionConfigProto.getNamePattern());
+      try {
+        pathTemplate.instantiate(new HashMap<>());
+      } catch (ValidationException e) {
+        // This must be a single resource name, because it has binding vars in the path pattern.
```
michaelbausor: I don't think I'm a fan of using exceptions like this if we don't need to. Why not check the name pattern for `{` or `*` chars?

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -860,34 +875,53 @@ private static GapicProductConfig createDummyInstance(
...
         createSingleResourceNameConfig(
             diagCollector, collectionConfigProto, singleResourceNameConfigsMap, file, language);
+        continue;
       }
-    }
 
-    if (diagCollector.getErrorCount() > 0) {
```
michaelbausor: Looks like we are dropping this check. Should we keep it, since the `createSingleResourceNameConfig` function accepts the diagCollector?

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -860,34 +875,53 @@ private static GapicProductConfig createDummyInstance(
     return ImmutableMap.copyOf(mergedResourceNameConfigs);
   }
 
-  private static ImmutableMap<String, SingleResourceNameConfig> createSingleResourceNameConfigs(
+  private static void createSingleAndFixedResourceNameConfigs(
```
michaelbausor: Rather than having a single function that accepts builders, WDYT about creating two lists of CollectionConfigProto objects, one for single resources and one for fixed resources, and then having `createSingleAndFixedResourceNameConfigs` take that list instead of a `ConfigProto`, and define a new function for fixed resources? I think that will be cleaner and clearer than putting both into a single function.

src/main/java/com/google/api/codegen/config/FixedResourceNameConfig.java
andreamlin: even better

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
andreamlin: i suppose if we aren't terribly worried about accuracy, plus there's only one fixed resource name in existence
done.

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
andreamlin: I'll bring it back, but I would rather it throw an error at that point. (If there are errors at this point, there's been enough time to collect a fair point, and generation won't succeed anyway. It's kinda annoying when the generator just spits out literally nothing instead of giving a useful error message.)

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
andreamlin: sure
done

andreamlin:
PTAL

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
michaelbausor: I think maybe I didn't articulate this very well. I like the approach you took in the plugin, that separated the CollectionConfigProto objects first, then handled them. So what about something like:
```java
List<CollectionConfigProto> single = allCollectionConfigProtos
  .stream()
  .filter(i -> !FixedResourceNameConfig.isFixedResourceNameConfig(i.getNamePattern()))
  .toList();
List<CollectionConfigProto> fixed = allCollectionConfigProtos
  .stream()
  .filter(i -> FixedResourceNameConfig.isFixedResourceNameConfig(i.getNamePattern()))
  .toList();
```
Then the existing `createSingleResourceNameConfigs` and `createFixedResourceNameConfigs` can be virtually unchanged (just taking List<CollectionConfigProto> instead of ConfigProto. My intuition is that this will be a smaller change. But let me know if I'm wrong about that or you don't like this idea.

andreamlin: I've removed `createSingleResourceNameConfigs` because all it did was collect all the CollectionConfigProtos (which we will do now beforehand) and then iterate over them and call `createSingleResourceNameConfig` on each CollectionConfigProto. I figure it's not worth having a separate function that would now just call `createSingleResourceNameConfig` on a list of CollectionConfigProto

andreamlin:
responded to comments.

PR #2678, opened 2019-04-02
Generate flattenings even if GAPIC config is not present

andreamlin:
Fixes a bug where method signatures were not parsed from proto options if a GAPIC config was not present.


andreamlin:
PTAL

PR #2679, opened 2019-04-02
Parse field name patterns from proto options

andreamlin:


src/test/java/com/google/api/codegen/protoannotations/testdata/java_library_no_gapic_config.baseline
```diff
@@ -822,7 +822,7 @@ public class LibraryServiceClient implements BackgroundResource {
    * Sample code:
    * <pre><code>
    * try (LibraryServiceClient libraryServiceClient = LibraryServiceClient.create()) {
-   *   BookOneOfName name = DeletedBookName.of();
+   *   BookName name = BookName.of("[SHELF_ID]", "[BOOK_ID]");
```
andreamlin: This is more accurate because there is a file-level Resource defined as `Book`, and the request object `MoveBookRequest` has a 
```
(google.api.resource_reference) = "google.example.library.v1.Book"
```

src/main/java/com/google/api/codegen/util/ProtoParser.java
```diff
@@ -422,6 +422,57 @@ private String getResourceSetFullName(ResourceSet resource, ProtoFile file) {
     return String.format("%s.%s", resource.getSymbol(), getProtoPackage(file));
   }
 
+  public ImmutableMap<String, String> getFieldNamePatterns(Method method) {
+    // Only look two levels deep in the request object, so fields of fields of the request object.
+    return ImmutableMap.copyOf(getFieldNamePatterns(method.getInputMessage(), 2));
+  }
+
+  private Map<String, String> getFieldNamePatterns(MessageType messageType, int depth) {
```
michaelbausor: Can we add a method-level comment to describe what this method is trying to do?

src/main/java/com/google/api/codegen/util/ProtoParser.java
```diff
@@ -422,6 +422,57 @@ private String getResourceSetFullName(ResourceSet resource, ProtoFile file) {
...
+    LinkedHashMap<String, String> builder = new LinkedHashMap<>();
+    for (Field field : messageType.getFields()) {
+      String fieldName = field.getSimpleName();
+
+      if (field.getType().isMessage() && depth > 0) {
+        Map<String, String> children =
+            getFieldNamePatterns(field.getType().getMessageType(), depth - 1);
+        for (Map.Entry<String, String> child : children.entrySet()) {
+          // Dot-separate the nested fields, e.g. "shelf.name"
+          builder.put(String.format("%s.%s", fieldName, child.getKey()), child.getValue());
```
michaelbausor: Suggestion - can we add a `builder` and a `prefix` argument to this method? That way we could avoid having to re-copy the children each time (adding the prefix) and also avoid the ImmutableMap.copyOf at the call site.

src/main/java/com/google/api/codegen/util/ProtoParser.java
andreamlin: Done.

src/main/java/com/google/api/codegen/util/ProtoParser.java
andreamlin: yeah!

andreamlin:
PTAL

src/main/java/com/google/api/codegen/util/ProtoParser.java
```diff
@@ -422,6 +422,72 @@ private String getResourceSetFullName(ResourceSet resource, ProtoFile file) {
     return String.format("%s.%s", resource.getSymbol(), getProtoPackage(file));
   }
 
+  public ImmutableMap<String, String> getFieldNamePatterns(Method method) {
+    ImmutableMap.Builder<String, String> resultCollector = ImmutableMap.builder();
+    // Only look two levels deep in the request object, so fields of fields of the request object.
+    int depth = 2;
```
michaelbausor: Nit: Do we need to declare this?

andreamlin: removed

PR #2685, opened 2019-04-03
Update protoc-java-resource-names-plugin

andreamlin:
java-resources-names-plugin just got released to v0.0.13.


andreamlin:
PTAL

PR #2686, opened 2019-04-03
SampleGen: File output

hzyi-google:
Configuration:
```yaml
- on_success:
  - write_file:
      file_name:
      - "I am a format_%s_.txt"
      - i.am.a.variable
      contents: $resp.raw_bytes
```


hzyi-google:
PTAL @googleapis/samplegen

src/main/java/com/google/api/codegen/transformer/StandardSampleImportTransformer.java
```diff
@@ -15,13 +15,17 @@
...
 import com.google.api.codegen.viewmodel.CallingForm;
 import com.google.api.codegen.viewmodel.ImportSectionView;
-import com.google.api.codegen.viewmodel.OutputView;
 import java.util.Collections;
-import java.util.List;
 
+/**
+ * An implementation of SampleImportTransformer. Subclasses of this class can choose to override
+ * `addSampleBodyImports`, `addOutputImports` and `addInitCodeImports` to save language-specific
+ * types to the type table. Currently used by Java.
```
vchudnov-g: Does this transform only the `MethodContext`, or also the `OutputContext`? (Looking at the other files, I gather it's the latter.) Might be helpful to mention that here.

src/main/java/com/google/api/codegen/viewmodel/OutputView.java
```diff
@@ -170,33 +172,62 @@ public static Builder newBuilder() {
...
 
-      public abstract PrintArgView build();
+      public abstract Builder isFirst(boolean val);
+
+      public abstract WriteFileView build();
+    }
+  }
+
+  @AutoValue
+  abstract class StringFormatView {
```
vchudnov-g: It might be helpful to either have a comment or change the name to be more descriptive: `InterpolatedStringView`? `StringWithArgsView`?

src/main/proto/com/google/api/codegen/config.proto
```diff
@@ -714,6 +714,26 @@ message OutputSpec {
...
+    // The content to write to the local file, of the pattern
+    // `var[.field...]`. Similar to the right hand side of `define`.
+    // `var` is either `$resp` or a variable that was
+    // previously defined by `loop` or `define`, and everything to the
+    // right of the period is any valid sequence of subfield or
+    // bracketed list subscripts or map keys to dereference a descendant
+    // of `var`.
+    string contents = 2;
+  }
+
```
vchudnov-g: :+1:

src/main/resources/com/google/api/codegen/java/standalone_sample.snip
```diff
@@ -354,21 +356,17 @@
...
 
 @private processPrintView(view)
-  @if not(view.args.size)
-    System.out.println("{@view.format}");
+  @if view.formattedString.args.isEmpty
+    System.out.println("{@view.formattedString.format}");
   @else
-    @if view.format == ""
-      System.out.println({@view.args.get(0).formattedName});
+    @if view.formattedString.format == "%s"
```
vchudnov-g: we don't have `@elsif`, huh? Bummer.

src/test/java/com/google/api/codegen/gapic/testdata/java/java_library.baseline
```diff
@@ -2549,6 +2549,73 @@ public class PublishSeriesRequestTestRequestObjectFieldComments {
...
+import com.google.example.library.v1.Shelf;
+import java.io.FileOutputStream;
+import java.io.FileWriter;
+import java.io.OutputStream;
+import java.util.ArrayList;
+import java.util.List;
+
+// [START sample]
+/*
+ * Please include the following imports to run this sample.
```
vchudnov-g: nit: We should probably end this line with a colon for all samples.

src/test/java/com/google/api/codegen/gapic/testdata/nodejs/nodejs_library.baseline
```diff
@@ -244,7 +244,6 @@ module.exports.default = Object.assign({}, module.exports);
 // [START sample_core]
 
 const library = require('@google-cloud/library').v1;
-
```
vchudnov-g: IMO, this and the other samples read better with some empty lines thrown in. Is this something the user can configure via empty comment declarations? Let's think about this for a subsequent, lower priority PR.

In the meantime, I would suggest hard coding some if it's easy, ideally between sections (eg above each read file or write file that has a comment, and in NodeJS after all the `require`s). WDYT?

src/test/java/com/google/api/codegen/gapic/testdata/nodejs/nodejs_library.baseline
```diff
@@ -1319,6 +1299,49 @@ function samplePublishSeries() {
...
+  };
+  client.publishSeries(request)
+    .then(responses => {
+      const response = responses[0];
+      // testing writing bytes field.
+      var writeFile = util.promisify(fs.writeFile);
+      writeFile(`uuid_of_series_with_book_${response.bookNames[0]}.raw`, response.seriesUuid.seriesBytes, 'binary');
+      // testing writing string field.
+
+      writeFile('series_uuid_in_plain_text.txt', response.seriesUuid.seriesString);
```
vchudnov-g: Expanding on my comment above about vertical whitespace: for example, here, I personally would find it more readable to have:

```
const response = ...;

// testing ...
var writeFile = ...
writeFile(...)

// testing ...
writeFile(...)
```

src/test/java/com/google/api/codegen/gapic/testdata/py/python_library.baseline
```diff
@@ -4643,7 +4642,6 @@ import sys
 # [START sample]
 
 from google.cloud.example import library_v1
-
```
vchudnov-g: definitely blank like after imports!

src/test/java/com/google/api/codegen/gapic/testdata/nodejs/nodejs_library.baseline
beccasaurus: I would just see if this diff can be fixed, I see these removed lines as a regression.

I wouldn't make any changes to the newline spacing in the scope of this PR, if it can be helped.

vchudnov-g: I agree. Let's fix this regression and restore the blank lines to at least what they were.
Then let's have a separate PR if needed to either hard-code or allow configuring extra blank lines if needed.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -139,41 +119,78 @@ public void run() {
...
+            new Scanner(config.getContents()), context, valueSet, outputContext.scopeTable(), form);
+    Preconditions.checkArgument(
+        contents.type().isStringType() || contents.type().isBytesType(),
+        "Output to file: expected string or bytes, found %s",
+        contents.type().getTypeName());
+    outputContext.fileOutputTypes().add(contents.type());
+    return OutputView.WriteFileView.newBuilder()
+        .fileName(fileName)
+        .contents(contents)
+        .isFirst(!outputContext.hasMultipleFileOutputs())
```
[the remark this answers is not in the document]
hzyi-google: So in line 149, the current write-file type is already added to `OutputContext`, so there should be at least one file output when reaching line 153. If there is just one file output, then this one is the first file output; if there are multiple, then this one is not the first file output.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -139,41 +119,78 @@ public void run() {
...
     }
-    ImmutableList<OutputView.PrintArgView> args = argsBuilder.build();
-    List<String> specs =
-        context
-            .getNamer()
-            .getPrintSpecs(
-                format, args.stream().map(arg -> arg.formattedName()).collect(Collectors.toList()));
-    return OutputView.PrintView.newBuilder().format(specs.get(0)).args(args).build();
+    ImmutableList<String> args = builder.build();
+    ImmutableList<String> formattedFormatAndArgs =
```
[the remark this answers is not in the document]
hzyi-google: Done.

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
```diff
@@ -564,7 +584,8 @@ private static void assertIdentifierNotUsed(
       CallingForm form) {
     Preconditions.checkArgument(
         !context.getNamer().getSampleUsedVarNames(context, form).contains(identifier),
-        "%s: %s cannot define variable \"%s\": it is used by the sample template for calling form \"%s\".",
+        "%s: %s cannot define variable \"%s\": it is used by the sample template for calling form"
```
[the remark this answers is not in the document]
hzyi-google: Done.

src/main/java/com/google/api/codegen/transformer/StandardSampleImportTransformer.java
hzyi-google: Updated this comment and the comment of `addOutputImports`.

src/main/java/com/google/api/codegen/viewmodel/OutputView.java
hzyi-google: Changed to `StringInterpolationView`. Also changed a few local variable names to `interpolatedString`.

src/main/resources/com/google/api/codegen/java/standalone_sample.snip
hzyi-google: Unfortunately searching for both `elsif` and `elseif` gives nothing

src/test/java/com/google/api/codegen/gapic/testdata/java/java_library.baseline
hzyi-google: I'll do that in a separate PR to not pollute the baseline change.

src/test/java/com/google/api/codegen/gapic/testdata/nodejs/nodejs_library.baseline
hzyi-google: The first blank line is hard to predict. Sometimes the output handling part is just a print statement, in which case it seems a blank line between the rpc call and the print statement is redundant.

For now, I change the templates so that every file output has an implicit newline.

hzyi-google:
PTAL

src/main/java/com/google/api/codegen/transformer/OutputTransformer.java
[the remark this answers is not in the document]
vchudnov-g: Oh, I see. `hasMultipleFileOutputs()` will return different things because the context changes every time `writeFileView()` is called, once for each file. :+1:

src/test/java/com/google/api/codegen/gapic/testdata/java/java_library.baseline
vchudnov-g: sg

src/test/java/com/google/api/codegen/gapic/testdata/nodejs/nodejs_library.baseline
vchudnov-g: (a) However we deal with this, I'm fine doing it in a chaser PR if it's not trivial. But we should remember to do it (file an issue, etc)

(b) I think relying on the user to specify newlines is easier for us, and I can imagine in some situations it's what the user wants. However, I'm guessing (just a guess!) most of the time it would be a drag for the user and make their sample config more verbose. Let's think a bit more whether there's an easy way to automate the common case. (There may not be and we may indeed wind up relying on user-specified blank lines.)

Spitballing one idea: maybe in general, if a config statement gets translated to more than one line of code (like file output), those lines get an automatic blank line after. This is what you said you tried and were concerned with the extra newline before the closing brace. How about making the logic for the newline be like a join (so we avoid the trailing blank line)? In snippet pseudo-code:
```
@file_write_1 = 'doSomething()\ndoSomethingElse()\n' # etc....
@multi_line_blocks = [ file_write_1, file_write_2, ...] 
function mySample() {
  @'\n'.join(multi_line_blocks)
}
```

src/main/java/com/google/api/codegen/viewmodel/OutputView.java
```diff
@@ -170,33 +172,62 @@ public static Builder newBuilder() {
...
 
-      public abstract PrintArgView build();
+      public abstract Builder isFirst(boolean val);
+
+      public abstract WriteFileView build();
+    }
+  }
+
+  @AutoValue
+  abstract class StringInterpolationView {
```
vchudnov-g: Maybe a brief comment saying this contains the format string as needed in the language, and the arguments to be used in that string?

src/test/java/com/google/api/codegen/gapic/testdata/nodejs/nodejs_library.baseline
hzyi-google: It sounds reasonable to me. I am a little worried that the snip templates are not intelligent enough to accomplish this mission :(. Issue filed: https://github.com/googleapis/gapic-generator/issues/2706

PR #2688, opened 2019-04-04
Treat ResourceNameConfig.entityName as Name

andreamlin:
Methods that call ResourceNameConfig.getEntityName() pretty much immediately convert it to a Name anyways. 

By making it a `Name`, we can take care of parsing the CamelCasing/snake_casing in one place, during construction, rather than trying to convert a weird String entityName into a Name in many other places.

Also make SingleResourceNameConfig an AutoValue.Builder built object so we can alter it via `toBuilder()` later on.

### DIFF
--- a/src/main/java/com/google/api/codegen/config/SingleResourceNameConfig.java
+++ b/src/main/java/com/google/api/codegen/config/SingleResourceNameConfig.java
@@ -103,7 +103,6 @@ public abstract class SingleResourceNameConfig implements ResourceNameConfig {
         .setAssignedProtoFile(file)
         .setEntityId(resource.getSymbol())
         .setEntityName(ResourceNameMessageConfig.entityNameToName(resource.getSymbol()))
-        .setCommonResourceName(null)
         .build();
   }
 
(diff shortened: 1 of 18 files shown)


andreamlin:
PTAL

src/main/java/com/google/api/codegen/config/SingleResourceNameConfig.java
```diff
@@ -85,8 +97,14 @@ static SingleResourceNameConfig createSingleResourceName(
...
 
-    return new AutoValue_SingleResourceNameConfig(
-        pathTemplate, nameTemplate, resource.getSymbol(), resource.getSymbol(), null, file);
+    return newBuilder()
+        .setNamePattern(pathTemplate)
+        .setNameTemplate(nameTemplate)
+        .setAssignedProtoFile(file)
+        .setEntityId(resource.getSymbol())
+        .setEntityName(ResourceNameMessageConfig.entityNameToName(resource.getSymbol()))
+        .setCommonResourceName(null)
```
hzyi-google: nit: this line is probably not needed?

PR #2689, opened 2019-04-04
Bug fixes for resource names

andreamlin:


src/main/java/com/google/api/codegen/util/ProtoParser.java
```diff
@@ -166,18 +165,6 @@ public String getResourceReferenceName(
     String resourceName = getResourceReference(field);
     if (!Strings.isNullOrEmpty(resourceName)) {
 
-      TypeRef resourceType = field.getModel().getSymbolTable().lookupType(resourceName);
```
andreamlin: This code block ends up being unnecessary, and a bug, because the `allResources` and `allResourceSets` objects already contain the message-defined Resource[Set]s.

src/main/java/com/google/api/codegen/util/ProtoParser.java
```diff
@@ -415,11 +403,11 @@ public String getProtoPackage(Method method) {
   }
 
   private String getResourceFullName(Resource resource, ProtoFile file) {
-    return String.format("%s.%s", resource.getSymbol(), getProtoPackage(file));
+    return String.format("%s.%s", getProtoPackage(file), resource.getSymbol());
```
andreamlin: params here were accidentally swapped.

src/test/java/com/google/api/codegen/testsrc/common/library.proto
```diff
@@ -816,7 +816,7 @@ message GetBookFromAnywhereRequest {
   // The name of the book to retrieve.
   string name = 1 [
     (google.api.field_behavior) = REQUIRED,
-    (google.api.resource_reference) = "google.example.library.v1.Book"];
+    (google.api.resource_reference) = "google.example.library.v1.BookOneOf"];
```
andreamlin: This is to reflect library_gapic.yaml, which has the `GetBookFromAnywhere` method use the `book_oneof` resource set instead of the single `book` resource name.

src/test/java/com/google/api/codegen/protoannotations/testdata/java_library_no_gapic_config.baseline
```diff
@@ -1028,7 +1028,7 @@ public class LibraryServiceClient implements BackgroundResource {
    * Sample code:
    * <pre><code>
    * try (LibraryServiceClient libraryServiceClient = LibraryServiceClient.create()) {
-   *   BookName name = BookName.of("[SHELF_ID]", "[BOOK_ID]");
+   *   BookOneOfName name = DeletedBookName.of();
```
andreamlin: changes directly resulting from the changes in library.proto

andreamlin:
PTAL

andreamlin:
They are being tested by the change in library.proto to use a qualified and unqualified ResouceSet name. The GapicCodeGeneratorAnnotationsTest wouldn't pass with the changes in library.proto, and after half-way tweaking it would read `Book` as a messageType containing a resourceSet and use the containing ResourceSet within `Book`.

PR #2694, opened 2019-04-04
Parse ResourceNames separately when annotations parsing is toggled on/off

andreamlin:
```
if (parsing_proto_annotations) {
 rn1 = parse(proto_annotations)
 rn2 = parse(gapic_config)
 rn = apply_gapic_overrides(rn1, rn2) 
} else {
 rn = parse(gapic_config)
}
```

This sets us up for introducing a GAPIC config v2; trying to parse a GAPIC config v2 right now is very touchy with all the bugs in resource name parsing.

For merging _single_ resource name entities from protofile/gapic-config, use the `pathPattern` from the protofile `Resource` object, and use the possibly-language-specific overrides for `entityName` and `common_resource_name`.

### DIFF
--- a/src/main/java/com/google/api/codegen/config/GapicProductConfig.java
+++ b/src/main/java/com/google/api/codegen/config/GapicProductConfig.java
@@ -748,8 +748,7 @@ public abstract class GapicProductConfig implements ProductConfig {
         insertFixedResourceNameConfig(
             diagCollector,
             fixedResourceNameConfig,
-            protoFile,
-            protoParser,
+            protoParser.getProtoPackage(file) + ".",
             fullyQualifiedFixedResourcesFromProtoFileCollector);
       } else {
         createSingleResourceNameConfigFromProtoFile(
@@ -933,7 +932,13 @@ public abstract class GapicProductConfig implements ProductConfig {
     for (CollectionConfigProto c : allCollectionConfigs) {
       if (!Strings.isNullOrEmpty(c.getNamePattern())
           && FixedResourceNameConfig.isFixedResourceNameConfig(c.getNamePattern())) {
-        createFixedResourceNameConfig(diagCollector, c, fixedResourceNameConfigMap, protoFile);
+        FixedResourceNameConfig fixedResourceNameConfig =
+            FixedResourceNameConfig.createFixedResourceNameConfig(
+                diagCollector,
+                c.getEntityName(),
+                c.getNamePattern(),
+                protoFile);
+        insertFixedResourceNameConfig(diagCollector, fixedResourceNameConfig, "", fixedResourceNameConfigMap);
       }
     }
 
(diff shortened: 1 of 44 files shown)


src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -702,89 +719,122 @@ private static GapicProductConfig createDummyInstance(
...
+
+    List<CollectionConfigProto> allCollectionConfigProtos =
+        new ArrayList<>(configProto.getCollectionsList());
+    configProto
+        .getInterfacesList()
+        .stream()
+        .forEach(i -> allCollectionConfigProtos.addAll(i.getCollectionsList()));
+
+    ImmutableMap<String, SingleResourceNameConfig> singleResourceNameConfigsFromGapicConfig =
+        createSingleResourceNamesFromGapicConfigOnly(
```
andreamlin: I did end up making this method, as you had suggested in a prior PR haha
it does keep it cleaner, plus it's reusable in the two situations (parsing proto annotations and not parsing them)

andreamlin:
PTAL

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -692,99 +709,131 @@ private static GapicProductConfig createDummyInstance(
...
-        .forEach(i -> allCollectionConfigProtos.addAll(i.getCollectionsList()));
-    LinkedHashMap<String, SingleResourceNameConfig> singleResourceNameProtos =
-        new LinkedHashMap<>();
-    LinkedHashMap<String, FixedResourceNameConfig> fixedResourceNameProtos = new LinkedHashMap<>();
-    for (CollectionConfigProto c : allCollectionConfigProtos) {
-      if (FixedResourceNameConfig.isFixedResourceNameConfig(c.getNamePattern())) {
-        createFixedResourceNameConfig(diagCollector, c, fixedResourceNameProtos, file);
+    // Maps of fully qualified Resource names to derived configs.
+    LinkedHashMap<String, SingleResourceNameConfig>
+        fullyQualifiedSingleResourcesFromProtoFileBuilder = new LinkedHashMap<>();
```
michaelbausor: I don't we should call this a builder when it is a map, even though we are using it a bit like one.

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -945,38 +1126,38 @@ private static void createSingleResourceNameConfig(
...
+      configKey = StringUtils.prependIfMissing(configKey, prefixForMap);
+      singleResourceNameConfigsMap.put(configKey, singleResourceNameConfig);
     }
   }
 
-  // Construct a new SingleResourceNameConfig from the given Resource, and add the newly
+  // Construct a new FixedResourceNameConfig from the given Resource, and add the newly
   // created config as a value to the map param, keyed on the package-qualified entity_id.
-  private static void createSingleResourceNameConfig(
+  private static void createFixedResourceNameConfigFromProtoFile(
```
michaelbausor: I think this function is slightly misnamed, and could be simplified. The actual creation is just a one-liner. Most of the logic here is logic about _insertion_ into the map, and doesn't use the Resource or ProtoFile. So WDYT about pulling out the creation line, and making this an `insertFixedResourceNameConfig` method?

michaelbausor: I think then the insertion logic can actually be shared between "fromprotofile" and "fromgapic" cases (although not 100% sure).

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
andreamlin: ok, calling it a `*Collector`

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
andreamlin: oh yes! done

andreamlin:
PTAL

andreamlin:
PTAL

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -794,39 +845,120 @@ private static GapicProductConfig createDummyInstance(
...
+  /**
+   * Create all the ResourceNameOneofConfig from the protofile and GAPIC config, and let the GAPIC
+   * config resourceNames override the protofile resourceNames in event of clashes.
+   */
+  @Nullable
+  private static ImmutableMap<String, ResourceNameConfig>
+      createResourceNameConfigsForGapicConfigOnly(
+          DiagCollector diagCollector,
+          ConfigProto configProto,
+          @Nullable List<ProtoFile> protoFiles,
```
michaelbausor: These functions all take a protoFiles parameter, but (I think?) all of them use it to extract a single proto file, and nothing else. Should this parameter be a nullable single proto file?

PR #2695, opened 2019-04-04
Fix empty method signature bug

andreamlin:
There was a bug where if an empty method signature was detected, we would immediately stop processing `method_signature`s. However, empty method signatures are valid, and we should treat them as any other.

([ProtoParserTest.java](https://github.com/googleapis/gapic-generator/blob/2c7caffd5dc6a813a51640e60944fa2d02326486/src/test/java/com/google/api/codegen/util/ProtoParserTest.java#L315) already verifies that we can tell the difference between an empty method signature and the absence of any method signatures).

### DIFF
--- a/src/test/java/com/google/api/codegen/protoannotations/testdata/java_library_no_gapic_config.baseline
+++ b/src/test/java/com/google/api/codegen/protoannotations/testdata/java_library_no_gapic_config.baseline
@@ -116,12 +116,30 @@ import com.google.api.pathtemplate.PathTemplate;
 import com.google.common.base.Function;
 import com.google.common.collect.Iterables;
 import com.google.common.util.concurrent.MoreExecutors;
+import com.google.example.library.v1.TestOptionalRequiredFlatteningParamsRequest;
+import com.google.example.library.v1.TestOptionalRequiredFlatteningParamsRequest.InnerMessage;
 import com.google.example.library.v1.stub.LibraryServiceStub;
 import com.google.example.library.v1.stub.LibraryServiceStubSettings;
 import com.google.longrunning.Operation;
 import com.google.longrunning.OperationsClient;
+import com.google.protobuf.Any;
+import com.google.protobuf.BoolValue;
+import com.google.protobuf.ByteString;
+import com.google.protobuf.BytesValue;
+import com.google.protobuf.DoubleValue;
+import com.google.protobuf.Duration;
 import com.google.protobuf.Empty;
 import com.google.protobuf.FieldMask;
+import com.google.protobuf.FloatValue;
+import com.google.protobuf.Int32Value;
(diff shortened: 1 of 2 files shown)


src/test/java/com/google/api/codegen/protoannotations/testdata/java_library_no_gapic_config.baseline
```diff
@@ -561,6 +579,28 @@ public class LibraryServiceClient implements BackgroundResource {
...
+   * Sample code:
+   * <pre><code>
+   * try (LibraryServiceClient libraryServiceClient = LibraryServiceClient.create()) {
+   *
+   *   ListShelvesResponse response = libraryServiceClient.listShelves();
+   * }
+   * </code></pre>
+   * @throws com.google.api.gax.rpc.ApiException if the remote call fails
+   */
+  public final ListShelvesResponse listShelves() {
```
andreamlin: this is the missing empty method signature!

src/test/java/com/google/api/codegen/protoannotations/testdata/java_library_no_gapic_config.baseline
```diff
@@ -2682,6 +2770,610 @@ public class LibraryServiceClient implements BackgroundResource {
...
..., List<String> optionalRepeatedResourceName, List<String> optionalRepeatedResourceNameOneof, List<String> optionalRepeatedResourceNameCommon, List<Integer> optionalRepeatedFixed32, List<Long> optionalRepeatedFixed64, Map<Integer, String> optionalMap, Any anyValue, Struct structValue, Value valueValue, ListValue listValueValue, Timestamp timeValue, Duration durationValue, FieldMask fieldMaskValue, Int32Value int32Value, UInt32Value uint32Value, Int64Value int64Value, UInt64Value uint64Value, FloatValue floatValue, DoubleValue doubleValue, StringValue stringValue, BoolValue boolValue, BytesValue bytesValue, List<Any> repeatedAnyValue, List<Struct> repeatedStructValue, List<Value> repeatedValueValue, List<ListValue> repeatedListValueValue, List<Timestamp> repeatedTimeValue, List<Duration> repeatedDurationValue, List<FieldMask> repeatedFieldMaskValue, List<Int32Value> repeatedInt32Value, List<UInt32Value> repeatedUint32Value, List<Int64Value> repeatedInt64Value, List<UInt64Value> repeatedUint64Value, List<FloatValue> repeatedFloatValue, List<DoubleValue> repeatedDoubleValue, List<StringValue> repeatedStringValue, List<BoolValue> repeatedBoolValue, List<BytesValue> repeatedBytesValue) {
```
andreamlin: this is a method signature that we missed out on because we had bailed out early upon finding an empty method signature

andreamlin:
PTAL

PR #2696, opened 2019-04-05
Update GAPIC test resources to match proto files

andreamlin:
Changing library_gapic.yaml and library.proto to better match each other.
Baseline and *Test.java changes are just a result of those test resource changes.

By making the library_gapic.yaml and its library.proto consistent with each other, it greatly reduces the diff between the Library clients generated with GAPIC config v1 and config v2.


src/test/java/com/google/api/codegen/testsrc/common/library.proto
```diff
@@ -44,6 +44,13 @@ option (google.api.resource_definition) = {
   pattern : "shelves/{shelf_id}/books/{book_id}"
 };
 
+option (google.api.resource_definition) = {
```
andreamlin: This resource definition currently exists in library_gapic.yaml

src/test/java/com/google/api/codegen/testsrc/common/library_gapic.yaml
```diff
@@ -129,12 +129,6 @@ interfaces:
     flattening:
       groups:
       - parameters: []
-    page_streaming:
```
andreamlin: This method doesn't follow the OnePlatform standards on paged methods, so it wouldn't count as a paged method if we solely looked at the protofile.

src/test/java/com/google/api/codegen/testsrc/common/library_gapic.yaml
```diff
@@ -572,7 +566,7 @@ interfaces:
     retry_params_name: default
     timeout_millis: 30000
     field_name_patterns:
-      name: book
+      name: shelf
```
andreamlin: This is what is defined in the proto file.

andreamlin:
PTAL

andreamlin:
things are broken

PR #2697, opened 2019-04-06
Update artman docker image

andreamlin:
The updated pubsub_gapic.yaml only works with artman v0.16.25+

Fixes CircleCI build.


andreamlin:
PTAL

PR #2698, opened 2019-04-06
AutoValue LongRunningConfig; always use gapic config's polling settings

andreamlin:
If a non-default GAPIC LRO config is given for a method, always use those polling settings. The gapic config is the only place where LRO polling settings can be defined

metadata and return type are defined by _either_ the protofile's annotations _or_ the GAPIC config.

this reduces the diff between API clients on gapic config v1 and gapic config v2

### DIFF
--- a/src/main/java/com/google/api/codegen/config/LongRunningConfig.java
+++ b/src/main/java/com/google/api/codegen/config/LongRunningConfig.java
@@ -178,7 +178,7 @@ public abstract class LongRunningConfig {
     }
 
     double pollDelayMultiplier = longRunningConfigProto.getPollDelayMultiplier();
-    if (pollDelayMultiplier <= 1.0) {
+    if (pollDelayMultiplier < 1.0) {
       diagCollector.addDiag(
           Diag.error(
               SimpleLocation.TOPLEVEL,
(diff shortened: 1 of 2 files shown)


andreamlin:
PTAL

src/main/java/com/google/api/codegen/config/LongRunningConfig.java
```diff
@@ -152,18 +144,71 @@ private static LongRunningConfig createLongRunningConfigFromProtoFile(
...
+    if (initialPollDelay.compareTo(Duration.ZERO) < 0) {
+      diagCollector.addDiag(
+          Diag.error(
+              SimpleLocation.TOPLEVEL,
+              "Initial poll delay must be provided and set to a positive number: '%s'",
+              longRunningConfigProto.getInitialPollDelayMillis()));
+    }
+
+    double pollDelayMultiplier = longRunningConfigProto.getPollDelayMultiplier();
+    if (pollDelayMultiplier <= 1.0) {
```
michaelbausor: This causes a failure if pollDelayMultiplier == 1.0, but the comment in the error says that 1.0 is valid - they should be consistent. (I realize this was just taken from below, but let's fix it!)

PR #2700, opened 2019-04-08
Oneof resource names in samples

andreamlin:
- Fix naming in test resources for `BookOneOf` -> `BookOneof`, which is how it would be represented
- Extract the `InitValueView` creation into a method
- Allow samples to refer to resource oneof name configs too, in addition to single resource name configs

### DIFF
--- a/src/test/java/com/google/api/codegen/protoannotations/testdata/java_library_no_gapic_config.baseline
+++ b/src/test/java/com/google/api/codegen/protoannotations/testdata/java_library_no_gapic_config.baseline
@@ -1929,7 +1929,7 @@ public class LibraryServiceClient implements BackgroundResource {
    * Sample code:
    * <pre><code>
    * try (LibraryServiceClient libraryServiceClient = LibraryServiceClient.create()) {
-   *   BookOneOfName name = DeletedBookName.of();
+   *   BookName name = DeletedBookName.of();
    *   BookName altBookName = BookName.of("[SHELF_ID]", "[BOOK_ID]");
    *   BookFromAnywhere response = libraryServiceClient.getBookFromAnywhere(name, altBookName);
    * }
@@ -1940,7 +1940,7 @@ public class LibraryServiceClient implements BackgroundResource {
    * single resource name type in a oneof.
    * @throws com.google.api.gax.rpc.ApiException if the remote call fails
    */
-  public final BookFromAnywhere getBookFromAnywhere(BookOneOfName name, BookName altBookName) {
+  public final BookFromAnywhere getBookFromAnywhere(BookName name, BookName altBookName) {
 
     GetBookFromAnywhereRequest request =
         GetBookFromAnywhereRequest.newBuilder()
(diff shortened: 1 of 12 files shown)


andreamlin:
PTAL

src/test/java/com/google/api/codegen/protoannotations/testdata/java_library_no_gapic_config.baseline
```diff
@@ -1940,7 +1940,7 @@ public class LibraryServiceClient implements BackgroundResource {
    * single resource name type in a oneof.
    * @throws com.google.api.gax.rpc.ApiException if the remote call fails
    */
-  public final BookFromAnywhere getBookFromAnywhere(BookOneOfName name, BookName altBookName) {
+  public final BookFromAnywhere getBookFromAnywhere(BookName name, BookName altBookName) {
```
michaelbausor: Do we want this to become BookName? Shouldn't it be BookNameOneof?

andreamlin: ok sooo previously, this was BookOne**O**f in the protofile, and now it is BookOne**o**f, which is equivalent to `book_oneof` in the GAPIC config. In Java, we take off the `_oneof` suffix from the resource name, so none of the resource name classes have "oneof"  in their names. 

Therefore the name of the resource set class is also `BookName`, which I know conflicts with the single resource entity `BookName`.

I was thinking of putting a rule in the config validator that it should be a warning if a proto package has a ResourceSet called FooOneof if there already exists a Resource called Foo. This lets gapic-generator maintain existing behavior for Java generation. Because these resource entity/set names can be overriden by the GAPIC config, the config validator would treat this as a warning instead of an error.

Thoughts?

michaelbausor: Hmm, I'm not sure I understand. What is the name of the oneof resource name classes that we generate currently in Java? I thought they were mostly named things like TopicOneof. How would we configure a name of TopicOneof in the annotations?

andreamlin: There are no classes named `*Oneof` in google-cloud-java; there is `TopicNames` for the collection_oneof and `TopicName` for the single resource entity. The GAPIC config will list the collection_oneof names as `topic_oneof`, but we take off that "oneof" suffix ([code](https://github.com/googleapis/gapic-generator/blob/bc13b57a36471bd2af2be4739c60442d8c300fb4/src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java#L889); the comments make this sound like it was intended to have the collection oneof share the same name with a single resource name entity).

michaelbausor: But then a few lines below, we return the stripped name but with "name_oneof" appended: https://github.com/googleapis/gapic-generator/blob/bc13b57a36471bd2af2be4739c60442d8c300fb4/src/main/java/com/google/api/codegen/transformer/SurfaceNamer.java#L892

That said you are completely right, we don't generate any `*Oneof` classes (we used to! my knowledge is clearly out of date). I'll take another look with this in mind. Thanks! :-)

PR #2702, opened 2019-04-08
Page streaming fix; optional fields fix

andreamlin:
- Move the method for PageStreamingConfig creation logic from GapicMethodConfig to PageStreamingConfig class
- Toggle parsing annotations vs parsing GAPIC config based on `protoParser.isProtoAnnotationsEnabled()`, instead of presence of the GAPIC config paging config
- Also fix how we calculate optional fields when GAPIC config not given

### DIFF
--- a/src/main/java/com/google/api/codegen/config/GapicMethodConfig.java
+++ b/src/main/java/com/google/api/codegen/config/GapicMethodConfig.java
@@ -24,9 +24,7 @@ import com.google.api.codegen.ResourceNameTreatment;
 import com.google.api.codegen.SurfaceTreatmentProto;
 import com.google.api.codegen.VisibilityProto;
 import com.google.api.codegen.common.TargetLanguage;
-import com.google.api.codegen.config.PageStreamingConfig.PagingFields;
 import com.google.api.codegen.configgen.ProtoMethodTransformer;
-import com.google.api.codegen.configgen.ProtoPagingParameters;
 import com.google.api.codegen.transformer.RetryDefinitionsTransformer;
 import com.google.api.codegen.transformer.SurfaceNamer;
 import com.google.api.codegen.util.ProtoParser;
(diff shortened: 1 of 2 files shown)


andreamlin:
PTAL

src/main/java/com/google/api/codegen/config/GapicMethodConfig.java
```diff
@@ -148,17 +115,13 @@ static GapicMethodConfig createMethodConfig(
       batching =
           BatchingConfig.createBatching(
               diagCollector, methodConfigProto.getBatching(), methodModel);
-      if (batching == null) {
```
michaelbausor: Why are we removing this check? Do we want createBatching to be called and return null?

src/main/java/com/google/api/codegen/config/PageStreamingConfig.java
```diff
@@ -218,4 +221,45 @@ static PageStreamingConfig createPageStreamingFromProtoFile(
     }
     return new AutoValue_PageStreamingConfig(pagingFields, resourcesFieldConfig);
   }
+
+  static PageStreamingConfig createPageStreamingConfig(
```
michaelbausor: Why package-private?

src/main/java/com/google/api/codegen/config/GapicMethodConfig.java
andreamlin: If `createBatching()` returns null, it also adds an error to the `diagCollector`. I figured it was extraneous to have these null checks on everything, when we could just ask the `diagCollector` for # of errors logged at the end. 

tbf, this does require that any method that returns null inside the calling function to also log an error (which is true right now).

src/main/java/com/google/api/codegen/config/PageStreamingConfig.java
andreamlin: oh this was just under the principle that we should limit visibility everywhere where possible; we can expand scope later if needed

andreamlin: also intellij waves a little yellow flag

src/main/java/com/google/api/codegen/config/GapicMethodConfig.java
michaelbausor: OK sounds good - I thought we were now treating this situation as "not an error", but handling the error later SGTM

src/main/java/com/google/api/codegen/config/PageStreamingConfig.java
michaelbausor: Fair enough :-) I don't know if we have a convention throughout gapic-generator, but one thing to consider is adding a comment indicating why it is package-private (something like "Package-private to allow access from GapicMethodConfig") because otherwise it can look like an omission of visibility rather than a choice. Totally optional though.

PR #2703, opened 2019-04-08
Start parsing GAPIC config v2

andreamlin:
Toggle on parsing of a GAPIC config v2.


andreamlin:
PTAL

BUILD.bazel
```diff
@@ -13,6 +13,7 @@ _COMPILE_DEPS = [
     "@com_fasterxml_jackson_core_jackson_databind//jar",
     "@com_google_api_api_common//jar",
     "@com_google_api_api_compiler//jar",
+    "@com_google_api_api_compiler_testing//jar",
```
michaelbausor: It seems odd to have a testing jar in the main deps. This is needed for the baseline differ? Is that to provide more user-friendly error output?

andreamlin: yah it's pretty and I figure our team is familiar with the look of the output.

PR #2704, opened 2019-04-08
ResourceNameOneofConfig fixes

andreamlin:
When creating a ResourceNameOneofConfig from protofile annotations, use the already-created collections of SingleResourceNameConfigs and FixedResourceNameConfigs derived from the protofile annotations.

This will come in handy when we merge in GAPIC config overrides for resource names, because we should create the OneofConfigs from the overriden resource name configs, instead of making the child resource name configs from scratch from the protofile.

I know that there will be big changes to the ResourceSet design, but these bug fixes should still be relevant after the redesign too.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -415,7 +415,7 @@ jobs:
           path: reports/gapic-generator
   generate-clients:
     docker:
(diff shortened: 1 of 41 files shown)


andreamlin:
PTAL

PR #2705, opened 2019-04-10
Test resources for multiple services across different protofiles

andreamlin:
This will let us test having resource names in different services, in different protofiles. For some languages, like ruby, a path template entity will only be generated in the interface/client file it was defined in.

We could perhaps remove the `multiple_services` protos/gapics/tests after this is checked in, because that covers the same case this does.

This PR only contains test resources changes.


andreamlin:
PTAL

src/test/java/com/google/api/codegen/testsrc/common/library_gapic.yaml
```diff
@@ -77,10 +95,6 @@ interfaces:
     language_overrides:
     - language: java
       entity_name: shelf_book
-  # To test the case that entity name / variable which happens to be
-  # a language keyword. See: https://github.com/googleapis/toolkit/issues/618
-  - name_pattern: shelves/{shelf}/books/{book}/returns/{return}
-    entity_name: return
```
andreamlin: Moving this to the new `another_service.proto` file

PR #2708, opened 2019-04-11
Bring back timeout millis in GAPIC config v2

andreamlin:
Because https://github.com/googleapis/gapic-generator/pull/2707


andreamlin:
PTAL

PR #2711, opened 2019-04-11
Resource names across different protofiles

andreamlin:
GAPIC config v1 has a way of limiting use of a resource entity to a [particular interface](https://github.com/googleapis/gapic-generator/blob/13a77c4334160ce220984b3941c65f407e41e65f/src/main/proto/com/google/api/codegen/config.proto#L161), i.e. `service` in protobuf. API config spec can define a resource name in a protofile, though it cannot specify a particular service for the resource name.

For all services in a protofile, generate the resource name artifacts that had resources defined in that protofile.

This is so APIs like Vision, which is composed of multiple protofiles, of which only some protos files have defined resource names, to only generate resource name utils in the classes where they are used. See the updated ruby baseline for an example.


src/test/java/com/google/api/codegen/protoannotations/testdata/ruby_library_no_gapic_config.baseline
```diff
@@ -4981,6 +4967,25 @@ module Library
...
+
+      private_constant :RETURN_PATH_TEMPLATE
+
+      # Returns a fully-qualified return resource name string.
+      # @param shelf [String]
+      # @param book [String]
+      # @param return_ [String]
+      # @return [String]
+      def self.return_path shelf, book, return_
+        RETURN_PATH_TEMPLATE.render(
```
andreamlin: This RETURN_PATH_TEMPLATE comes from the pre-existing resource entity defined in another_service.proto, and now it shows up in the MyProtoClient generated class!

Note that there were no changes to the proto test files, only the gapic-generator sources.

src/test/java/com/google/api/codegen/protoannotations/testdata/ruby_library_no_gapic_config.baseline
```diff
@@ -2772,50 +2772,38 @@ module Library
...
 
-      SHELF_PATH_TEMPLATE = Google::Gax::PathTemplate.new(
-        "shelves/{shelf_id}"
-      )
-
-      private_constant :SHELF_PATH_TEMPLATE
-
-      DELETED_BOOK_PATH_TEMPLATE = Google::Gax::PathTemplate.new(
-        "_deleted-book_"
+      ARCHIVED_BOOK_PATH_TEMPLATE = Google::Gax::PathTemplate.new(
```
andreamlin: This is just re-ordering of the resource name artifacts. i thought it would make more sense to order them alphabetically, as it's deterministic and easy to reason about.

andreamlin:
PTAL

src/test/java/com/google/api/codegen/protoannotations/testdata/ruby_library_no_gapic_config.baseline
michaelbausor: How were they ordered previously? Is this going to cause churn when we switch a client from v1 to v2 config?

andreamlin: Pretty sure it was ordered by the order in GAPIC interface's `collections`, followed by any top-level non-duplicate API `collections`

So if there's no `collections` field in the GAPIC config v2 to order the resource names, yeah we'll most likely get churn

michaelbausor: OK. Not sure whether there is a coherent alternative, but I think that avoiding churn is a worthwhile goal. If it's not possible, we may just need to do a regeneration ahead of time to get the churn out of the way.

andreamlin: I think we would have to do the regeneration pre-party, yeah. We could either draw ire now by changing the current generator to order resource name artifacts alphabetically (a billion changes at once for Yoshi), or API-by-API inside GAPIC configs (manual work on our part, but diffs are spread over time for Yoshi).

andreamlin:
@michaelbausor any more comments?

PR #2712, opened 2019-04-11
Split proto parsing in GapicMethodConfig

michaelbausor:
- Fixes parsing for flattenings
- Fixes parsing for LRO
- Pushes `if (protoParser.isProtoAnnotationsEnabled())` logic into GapicMethodConfig for page streaming

The intention of pushing the `if (protoParser.isProtoAnnotationsEnabled())` up into GapicMethodConfig is to enable a followup that gathers the many `if (protoParser.isProtoAnnotationsEnabled())` statements in GapicMethodConfig into a single statement. This will involve a reordering in GapicMethodConfig but no semantic changes, but hopefully will improve clarify about when proto parsing is active.


andreamlin:
Thanks!

PR #2714, opened 2019-04-12
Allow for "*" in resource_references

andreamlin:
This does the same thing as "*" in the GAPIC config in the `field_name_patterns` and `field_entity_map`.

The updated baseline file (generated without a GAPIC config) more closely reflects the analogous java_library.baseline (generated with the GAPIC config).

### DIFF
--- a/src/test/java/com/google/api/codegen/testsrc/common/library.proto
+++ b/src/test/java/com/google/api/codegen/testsrc/common/library.proto
@@ -773,8 +773,7 @@ message ListStringsRequest {
 message ListStringsResponse {
   repeated string strings = 1 [
     // Test setting the "*" on a field in a response message
-    (google.api.resource_reference) = "*"
-  ];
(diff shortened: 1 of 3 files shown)


andreamlin:
PTAL

src/test/java/com/google/api/codegen/testsrc/common/library.proto
```diff
@@ -763,13 +763,18 @@ message MoveBookRequest {
...
   int32 page_size = 2;
   string page_token = 3;
 }
 
 message ListStringsResponse {
-  repeated string strings = 1;
+  repeated string strings = 1 [
+    // Test setting the "*" on a field in a response message
+    (google.api.resource_reference) = "*"
+  ];
```
michaelbausor: Nit: make this layout match layout on line 768

andreamlin: done

PR #2715, opened 2019-04-12
Refactor gapic method config

michaelbausor:
- Convert GapicMethodConfig to use an AutoValue Builder
- Create two GapicMethodConfig create methods: 1 for proto annotation parsing, one without

Please let me know what you think of this change - is it an improvement? Or just churn but not really better.


andreamlin:
I think this is what I would have wanted to do! haha 

Make sure it passes the build locally?

michaelbausor:
@eoogbe @alexander-fenster can one of you PTAL at this since @andreamlin is OOO this week?

PR #2716, opened 2019-04-17
go: fix ignored page token bug in iterator

noahdietz:
A mirror of the fix submitted in [gapic-generator-go#128](https://github.com/googleapis/gapic-generator-go/pull/128), the `PageToken` was not being respected at iterator creation. It was being ignored entirely if supplied when creating a new iterator.

added @jadekler for approval of changes to generated output. This is technically a bug in all Go GAPICs today. We will need to regen once the change is released.


pongad:
I'll let @jadekler sign off, but this LGTM.

PR #2722, opened 2019-04-22
Alphabetize SingleResourceNameConfigs

andreamlin:
In order to reduce churn during the config party, we should alphabetize all single resource name entities. Alphabetizing creates a deterministic, simple ordering for clients generated without a GAPIC config.


andreamlin:
PTAL

andreamlin:
It only affects the order of the surface's string formatting/parsing functions, and the baseline changes should completely capture the differences.

PR #2724, opened 2019-04-23
Config v2 tests

andreamlin:
- Update config_v2.proto with the new fields in config.proto
- Add the reduced library_gapic_v2.yaml
- Add a baseline Java test for GAPIC config v2; the goal is to iteratively reduce the diff between this file and the other java_library.baseline created from GAPIC config v1

### DIFF
--- a/src/test/java/com/google/api/codegen/protoannotations/testdata/library_v2_gapic.yaml
+++ b/src/test/java/com/google/api/codegen/protoannotations/testdata/library_v2_gapic.yaml
@@ -1,0 +1,693 @@
+type: com.google.api.codegen.ConfigProto
+config_schema_version: 2.0.0
+language_settings:
+  java:
+    package_name: com.google.example.library.v1
+    interface_names:
+      google.example.library.v1.LibraryService: Library
+    release_level: GA
+  python:
+    package_name: google.cloud.example.library_v1.gapic
+  go:
+    # Intentionally changed from the gRPC package name.
+    # See: https://github.com/googleapis/toolkit/issues/320
+    package_name: cloud.google.com/go/library/apiv1
+    domain_layer_location: cloud.google.com/go/library
+    interface_names:
+      google.example.library.v1.LibraryService: LibClient
+  ruby:
+    # Note that package_name for ruby is intentionally changed
+    # from the gRPC package name of the API to test the generated
+    # code. See: https://github.com/googleapis/toolkit/issues/309
+    package_name: Library::V1
+  # Use default PHP package name.
+  nodejs:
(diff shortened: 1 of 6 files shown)


andreamlin:
PTAL

andreamlin:
More tests are coming in another PR

src/test/java/com/google/api/codegen/protoannotations/testdata/library_v2_gapic.yaml
```diff
@@ -0,0 +1,693 @@
...
+    package_name: Library::V1
+  # Use default PHP package name.
+  nodejs:
+    package_name: library.v1
+    domain_layer_location: google-cloud
+  csharp:
+    package_name: Google.Example.Library.V1
+interfaces:
+  - name: google.example.library.v1.MyProto
+    retry_codes_def:
```
michaelbausor: Is this necessary? We should be able to omit the retry codes from v2 config if they just specify the defaults.

src/test/java/com/google/api/codegen/protoannotations/testdata/library_v2_gapic.yaml
```diff
@@ -0,0 +1,693 @@
...
+            common_resource_name: com.google.cloud.ProjectName
+    smoke_test:
+      method: UpdateBook
+      init_fields:
+        - name%shelf_id="testShelf-$RANDOM"
+        - name%book_id=$PROJECT_ID
+        - book . rating=GOOD
+    methods:
+      - name: CreateShelf
+        retry_codes_name: non_idempotent
```
michaelbausor: Can these be inferred from http annotation? We should have a mix of most being inferred, but some being overridden. WDYT?

src/test/java/com/google/api/codegen/protoannotations/testdata/library_v2_gapic.yaml
andreamlin: Done.

src/test/java/com/google/api/codegen/protoannotations/testdata/library_v2_gapic.yaml
andreamlin: Ahh I got lazy
Removed the retry codes names that are the same as the inferred value.

andreamlin:
removed redundant config values
PTAL

andreamlin:
omg i put the requested changes in another branch instead of this one
will push those v soon

michaelbausor:
Heh, my fault also for bad LGTM, whoops :-)

PR #2726, opened 2019-04-23
Really alphabetize the SingleResourceNameConfigs

andreamlin:
Emphasize that InterfaceConfig.getSingleResourceNameConfigs is an unordered collection, and that it's up to the transformers to order the configs as they see fit.

This reduces the diff between a GAPIC config v2 -generated client and one generated with GAPIC config v1.


andreamlin:
PTAL

PR #2727, opened 2019-04-23
Support new resource annotations

michaelbausor:
@andreamlin PTAL. Still some work to do:
 - <s>Fix ResourceNameMessageConfigsTest</s>
 - <s>Documentation of new ResourceDescriptorConfig class</s>
 - Release common protos package


michaelbausor:
@andreamlin PTAL

src/test/java/com/google/api/codegen/testsrc/common/library.proto
```diff
@@ -453,11 +429,15 @@ message Book {
 
 // A single book in the archives.
 message BookFromArchive {
+  option (google.api.resource) = {
+    type: "library.googleapis.com/BookFromArchive",
```
andreamlin: Is it possible to rename this back to `ArchivedBook` to match (ignoring casing) what the library_gapic.yaml does?

src/test/java/com/google/api/codegen/testsrc/common/tagger.proto
```diff
@@ -38,7 +38,7 @@ message AddTagRequest {
   // projects/{project}/zones/{zone}/disks/{disk}.
   string resource = 1 [
     (google.api.field_behavior) = REQUIRED,
-    (google.api.resource_reference) = "google.example.library.v1.Book"];
+    (google.api.resource_reference).type = "library.googleapis.com/Book"];
```
andreamlin: Do all of these resource_reference.types have to be fully qualified like this?

src/test/java/com/google/api/codegen/testsrc/common/library.proto
```diff
@@ -754,33 +739,29 @@ message MoveBookRequest {
...
 
   // The name of the destination shelf.
   string other_shelf_name = 2 [
     (google.api.field_behavior) = REQUIRED,
-    (google.api.resource_reference) = "google.example.library.v1.Shelf"];
+    (google.api.resource_reference).type = "library.googleapis.com/Shelf"];
 }
 
 message ListStringsRequest {
-  string name = 1 [
```
andreamlin: It would be very helpful to leave in some kind of test for `"*"`, preferably on this and the `ListStringsResponse` message types to match the library_gapic.yaml

src/test/java/com/google/api/codegen/testsrc/common/library.proto
```diff
@@ -453,11 +429,15 @@ message Book {
...
+  option (google.api.resource) = {
+    type: "library.googleapis.com/BookFromArchive",
+    pattern: "archives/{archive}/books/{book}",
+  };
   // The resource name of the book.
   // Book names have the form `archives/{archive_id}/books/{book_id}`.
   string name = 1 [
-    (google.api.field_behavior) = REQUIRED,
-    (google.api.resource_reference) = "ArchivedBook"];
+    (google.api.field_behavior) = REQUIRED
```
andreamlin: Can we keep the resource annotation on this field? Or is like magically implied because the message type has a resource declaration?

src/test/java/com/google/api/codegen/testsrc/common/library.proto
```diff
@@ -674,15 +659,15 @@ message GetBookRequest {
   // The name of the book to retrieve.
   string name = 1 [
     (google.api.field_behavior) = REQUIRED,
-    (google.api.resource_reference) = "Book"];
+    (google.api.resource_reference).type = "library.googleapis.com/Book"];
```
andreamlin: How are we differentiating between fields that are of the "resource set" variety and those that are just a single resource? 

Whether a field is a resource set or a singular resource mighttt change the field name in C# and be a breaking change. The latest master branch has the C# baseline file that we could see if there's a diff in.

src/test/java/com/google/api/codegen/testsrc/common/library.proto
```diff
@@ -921,14 +902,12 @@ message TestOptionalRequiredFlatteningParamsRequest {
   bytes required_singular_bytes = 8 [(google.api.field_behavior) = REQUIRED];
   InnerMessage required_singular_message = 9 [(google.api.field_behavior) = REQUIRED];
   string required_singular_resource_name = 10 [(google.api.field_behavior) = REQUIRED,
-    (google.api.resource_reference) = "Book"
+    (google.api.resource_reference).type = "library.googleapis.com/Book"
   ];
   string required_singular_resource_name_oneof = 11 [(google.api.field_behavior) = REQUIRED,
-    (google.api.resource_reference) = "google.example.library.v1.Book"
-  ];
-  string required_singular_resource_name_common = 14 [(google.api.field_behavior) = REQUIRED,
```
andreamlin: What's the status of common resource names like `Project` and `Organization`?

src/test/java/com/google/api/codegen/testsrc/common/library.proto
michaelbausor: Yes will do

src/test/java/com/google/api/codegen/testsrc/common/tagger.proto
michaelbausor: As I understand it, yes.

src/test/java/com/google/api/codegen/testsrc/common/library.proto
michaelbausor: Ack, will do

src/test/java/com/google/api/codegen/testsrc/common/library.proto
michaelbausor: Removing this was a mistake, will add back

src/test/java/com/google/api/codegen/testsrc/common/library.proto
michaelbausor: Fields with a single pattern are single, with multiple patterns are sets. This behaviour can also be modified using the History field in ResourceDescriptor.

src/test/java/com/google/api/codegen/testsrc/common/library.proto
michaelbausor: Punting on this in annotations.

src/test/java/com/google/api/codegen/testsrc/common/library.proto
michaelbausor: References to these types need to be through the ResourceDescriptor.child_type field. I don't know how to translate the xamples in library.proto into such references, since there isn't a pattern with format `projects/{project}/foos/{foo}`

src/test/java/com/google/api/codegen/testsrc/common/library.proto
andreamlin: Does `library.googleapis.com/Book` refer to the resource set? This field `name` should only refer to a single resource.

src/test/java/com/google/api/codegen/testsrc/common/library.proto
andreamlin: With this new spec, there is no way to just define a `projects/{project}`? This is necessary for many APIs (30ish)

andreamlin:
We can use proto-google-common-protos v1.16.0 now! :)

src/test/java/com/google/api/codegen/testsrc/common/library.proto
michaelbausor: The `projects/{project}` name is implied if you specify a `.child_type` reference to a `projects/{project}/foos/{foo}` pattern.

andreamlin:
Maybe the version for proto-google-common-protos should be changed everywhere to the newest? Including in the circle ci config for the showcase stuff

andreamlin:
Oh no I bet that the newest google common protos will break showcase, and the showcase generation won't work until https://github.com/googleapis/gapic-generator/pull/2646 is updated in the same manner it had been.

I would be ok with disabling the showcase generation for now, with an issue/TODO filed to restore it. We've whitelisted tests before.

src/test/java/com/google/api/codegen/testsrc/common/library.proto
michaelbausor: It is not possible to refer to only a single resource from a set under the new annotation. To get that behavior, would need to use a field_name_pattern in the v2 yaml

.circleci/config.yml
```diff
@@ -77,22 +77,22 @@ anchor_verify_and_test_python: &anchor_verify_and_test_python
...
-        when: always
-    - run:
-        name: Run Showcase Integration Tests.
-        command: |
-          nox --session "showcase(py='2.7')" \
-              --noxfile /tmp/workspace/gapic-generator/showcase/python/nox.py
-          nox --session "showcase(py='3.6')" \
-              --noxfile /tmp/workspace/gapic-generator/showcase/python/nox.py
-        when: always
+    #- run:
```
andreamlin: Can you file an issue and add a comment as to why this is commented out?

andreamlin: And everywhere else that the Showcase tests are commented out?

src/test/java/com/google/api/codegen/testsrc/common/library.proto
andreamlin: How would a single resource be referred to by the v2 yaml?

src/test/java/com/google/api/codegen/protoannotations/testdata/csharp_library.baseline
```diff
@@ -48,20 +48,11 @@ namespace Google.Example.Library.V1.SmokeTests
...
-                return 1;
-            }
-            string projectId = args[0];
-
             // Create client
             LibraryServiceClient client = LibraryServiceClient.Create();
 
             // Initialize request argument(s)
-            BookName name = new BookName($"testShelf-{Guid.NewGuid()}", projectId);
+            BookNameOneof name = BookNameOneof.From(new BookName("[BOOK_SHELF]", "[BOOK]"));
```
andreamlin: I think that changing this type to `BookNameOneof` will break the C# client, so we should do field_name_patterns map parsing after this PR. Just filed https://github.com/googleapis/gapic-generator/issues/2734

src/test/java/com/google/api/codegen/protoannotations/testdata/java_library.baseline
```diff
@@ -7653,55 +7659,7 @@ public class MyProtoClient implements BackgroundResource {
...
-      PathTemplate.createWithoutUrlEncoding("shelves/{shelf}/books/{book}/returns/{return}");
 
-  /**
-   * Formats a string containing the fully-qualified path to represent
-   * a return resource.
-   *
-   * @deprecated Use the {@link ReturnName} class instead.
-   */
-  @Deprecated
-  public static final String formatReturnName(String shelf, String book, String return_) {
```
andreamlin: Is there any way to define a resource name outside of a message type?

andreamlin: Hmm, doesn't look like the resource.proto gives us room for this, probably deliberately in order to enforce API guidelines.

andreamlin:
LGTM after showcase issue comment is resolved.

andreamlin:
I'm a little sad about the loss of functionality, but it sounds like we can re-add most of it via the gapic config v2 later. This was as good as could be done given the design constraints, thanks!

.circleci/config.yml
michaelbausor: Done

src/test/java/com/google/api/codegen/protoannotations/testdata/csharp_library.baseline
michaelbausor: Cool, thanks!

PR #2729, opened 2019-04-24
SampleGen: sample package for Node.js

hzyi-google:
- pull samples out from `src` directory
- generate a `package.json` for the samples
It should be possible to run the samples this way:
```shell
artman generate nodejs_gapic ...
cd artman-output/js/[api]
npm install
cd samples
npm install
node v1/some_sample.js
```


hzyi-google:
PTAL

PR #2740, opened 2019-04-26
Cleanup

andreamlin:


andreamlin:
PTAL

andreamlin:
PTAL

src/main/java/com/google/api/codegen/config/RetryCodesConfig.java
```diff
@@ -81,6 +82,7 @@ public static RetryCodesConfig create(InterfaceConfigProto interfaceConfigProto)
     return retryCodesConfig;
   }
 
+  @Nonnull
```
hzyi-google: I think we don't need the `@Nonnull` annotation. Java Readability discourages the use of `@Nonnull`, as `@Nonnull` is always implied unless marked `@Nullable`. See https://engdoc.corp.google.com/eng/doc/devguide/java/practices/null.md?cl=head.

Issue #2739, opened 2019-04-26
Add string methods for the flattened resource name methods for C#


andreamlin:
If there is a flattened method where one of the flattening params is a resource name type, then we should always generate an overloaded method with the same parameters but using a string type instead of the resource name type.

This allows adding resource names to configs to be a non-breaking change. Otherwise, adding a resource name removes the string param method with the resource name param method.

PR #2741, opened 2019-04-29
[C#] Always generate the string param flattening method for resource-name methods

andreamlin:
Fixes https://github.com/googleapis/gapic-generator/issues/2739

For C#, for _all_ flattened method configurations (LRO/streaming/etc), if one of the params in the method signature is a _resource entity_, then _also_ create a flattened method context in which the resource entity param is instead a _string_ field.

This allows an API to not have resource names and then later introduce resource names for existing methods without breaking the previous string-based methods.

--- Addendum ---
For transparency, this does allow for a bug where this new method overload conflicts with another flattening, e.g.
Let's say there is an RPC request message
```
message GetShelfRequest {
   Shelf shelf = 1;
   String id = 2;
   String genre = 3;
}
```
And the RPC has method signatures `"shelf, id"` and `"id, genre"`, then we will generate methods
```
public virtual Shelf GetShelf(Shelf shelf, String id);
public virtual Shelf GetShelf(String shelf, String id);
public virtual Shelf GetShelf(String id, String genre);
```
where the last two method overloads have the same type signature and (I think) will not compile.

But this is very similar to the possibility of someone defining two different method signatures with the same list of parameter types, which would result in the same compile-time error.

### DIFF
--- a/src/main/java/com/google/api/codegen/config/FlatteningConfig.java
+++ b/src/main/java/com/google/api/codegen/config/FlatteningConfig.java
@@ -329,4 +329,13 @@ public abstract class FlatteningConfig {
     paramList.forEach(p -> paramsAsString.append(p.getSimpleName()).append(", "));
     return paramsAsString.toString();
   }
+
+  /** Return if the flattening config contains a parameter that is a resource name. */
+  public static boolean hasAnyResourceNameParameter(FlatteningConfig flatteningGroup) {
+    return flatteningGroup
+        .getFlattenedFieldConfigs()
+        .values()
+        .stream()
+        .anyMatch(FieldConfig::useResourceNameType);
+  }
 }
--- a/src/main/java/com/google/api/codegen/transformer/csharp/CSharpApiMethodTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/csharp/CSharpApiMethodTransformer.java
@@ -151,9 +151,7 @@ public class CSharpApiMethodTransformer extends StaticLangApiMethodTransformer {
             if (FlatteningConfig.hasAnyResourceNameParameter(flatteningGroup)) {
               apiMethods.add(
                   generateGrpcStreamingFlattenedMethod(
-                      methodContext
-                          .withResourceNamesInSamplesOnly()
-                          .withCallingForms(Collections.singletonList(CallingForm.Flattened)),
+                      methodContext.withResourceNamesInSamplesOnly(),
                       csharpCommonTransformer.callSettingsParam()));
             }
           }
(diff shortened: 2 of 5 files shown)


andreamlin:
PTAL

src/main/java/com/google/api/codegen/transformer/csharp/CSharpApiMethodTransformer.java
```diff
@@ -226,24 +207,15 @@ protected void setServiceResponseTypeName(
         // Paged streaming methods.
         if (methodConfig.isFlattening()) {
           for (FlatteningConfig flatteningGroup : methodConfig.getFlatteningConfigs()) {
-            GapicMethodContext methodContext =
+            MethodContext methodContext =
                 context.asFlattenedMethodContext(requestMethodContext, flatteningGroup);
-            apiMethods.add(
```
andreamlin: This code block has been moved, unchanged, to the body of `generatePageStreamingFlattenedMethods()`

src/main/java/com/google/api/codegen/transformer/csharp/CSharpApiMethodTransformer.java
```diff
@@ -269,23 +241,11 @@ protected void setServiceResponseTypeName(
           for (FlatteningConfig flatteningGroup : methodConfig.getFlatteningConfigs()) {
             GapicMethodContext methodContext =
                 context.asFlattenedMethodContext(requestMethodContext, flatteningGroup);
-            apiMethods.add(
```
andreamlin: This code block has been moved, unchanged, to the body of `generateFlattenedAsyncMethods()`

src/main/java/com/google/api/codegen/transformer/csharp/CSharpApiMethodTransformer.java
```diff
@@ -166,36 +172,11 @@ protected void setServiceResponseTypeName(
...
```
andreamlin: This code block has been moved, unchanged, to the body of `generateFlattenedLroMethods()`

src/main/java/com/google/api/codegen/config/FlatteningConfig.java
```diff
@@ -329,4 +329,13 @@ private static String flatteningConfigToString(FlatteningConfig flatteningConfig
     paramList.forEach(p -> paramsAsString.append(p.getSimpleName()).append(", "));
     return paramsAsString.toString();
   }
+
+  /** Return if the flattening config contains a parameter that is a resource name. */
```
andreamlin: This was copied from JavaMethodViewGenerator; only the `public static` method modifiers were added.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpApiMethodTransformer.java
```diff
@@ -310,4 +270,80 @@ protected void setServiceResponseTypeName(
...
+            methodContext.withCallingForms(
+                ImmutableList.of(
+                    CallingForm.FlattenedPaged,
+                    CallingForm.FlattenedPagedAll,
+                    CallingForm.FlattenedPagedPageSize)),
+            pagedMethodAdditionalParams));
+    return apiMethods;
+  }
+
+  private List<StaticLangApiMethodView> generateFlattenedAsyncMethods(MethodContext methodContext) {
```
chrisdunelm: Minor point, but this probably shouldn't be named `generateFlattenedAsyncMethods`, as it's generating a sync method as well as two async methods.
Possibly `generateNormalFlattenedMethods`?

andreamlin: Ok!

andreamlin:
The diff on github does look a little weird. The diffs on the baseline are purely additive in the Intellij UI.

PR #2743, opened 2019-04-30
Make gapic.bzl rules compliant with the standalone microgenerator spec

vam-google:
Specifically `gapic_srcjar` rule now supports `-descriptor`, `--package` and `--output` parameters.


rules_gapic/gapic.bzl
```diff
@@ -56,19 +57,21 @@ gapic_srcjar = rule(
...
-        "gapic_yaml": attr.label(mandatory = True, allow_single_file = True),
-        "artifact_type": attr.string(mandatory = False, default = "GAPIC_CODE"),
-        "language": attr.string(mandatory = True),
+        "artifact_type": attr.string(mandatory = False), #default = "GAPIC_CODE"
+        "gapic_yaml": attr.label(mandatory = False, allow_single_file = True),
+        "language": attr.string(mandatory = False),
         "service_yaml": attr.label(mandatory = False, allow_single_file = True),
         "package_yaml2": attr.label(mandatory = False),
+        "package": attr.string(mandatory = False),
+        "output_suffix": attr.string(mandatory = False, default = ".srcjar"),
```
andreamlin: nit: My intuition is that someone will specify a suffix without the period, like "jar" instead of ".jar". Feel free to challenge that

rules_gapic/gapic.bzl
```diff
@@ -56,19 +57,21 @@ gapic_srcjar = rule(
             allow_single_file = True,
             mandatory = True,
         ),
-        "gapic_yaml": attr.label(mandatory = True, allow_single_file = True),
-        "artifact_type": attr.string(mandatory = False, default = "GAPIC_CODE"),
-        "language": attr.string(mandatory = True),
+        "artifact_type": attr.string(mandatory = False), #default = "GAPIC_CODE"
```
andreamlin: Why was the default removed for this now optional option?

rules_gapic/gapic.bzl
```diff
@@ -56,19 +57,21 @@ gapic_srcjar = rule(
...
+        "gapic_yaml": attr.label(mandatory = False, allow_single_file = True),
+        "language": attr.string(mandatory = False),
```
andreamlin: Why is `language` now optional?

rules_gapic/gapic.bzl
```diff
@@ -12,37 +12,38 @@
...
-        optional_arguments.append("--package_yaml2=%s" % ctx.file.package_yaml2.path)
-        optional_action_inputs.append(ctx.file.package_yaml2)
-
-    if ctx.attr.artifact_type.find("DISCOGAPIC") >= 0:
-        optional_arguments.append("--discovery_doc=%s" % ctx.file.src.path)
+    arguments = []
+    inputs = []
+
+    attr = ctx.attr
+    if attr.artifact_type:
```
eoogbe: This branch requires the `gapic_yaml` and `language`. Add a `fail` call to ensure its presence.

rules_gapic/gapic.bzl
vam-google: The suffix is not always <dot>.<someting>. It could be anythin, sometimes empty, sometimes sometnhing like `"-set.proto.bin"` (check `proto_library_with_info` rule definition). Keeping it like this is necessary and makes the rule much more flexible.

andreamlin: Totally makes sense. Are there docs for the calling interface?

rules_gapic/gapic.bzl
vam-google: Because it makes sense only for the monolithic generator and can break (as unexpected argument) microgenerators.

rules_gapic/gapic.bzl
vam-google: Because monolithic generators do not need the language argument, since they generate only one language.

rules_gapic/gapic.bzl
```diff
@@ -12,37 +12,38 @@
...
+        _set_args(attr.artifact_type, "", arguments)
+        if ctx.attr.artifact_type.find("DISCOGAPIC") >= 0:
+            _set_args(attr.src, "--discovery_doc=", arguments, inputs)
+        else:
```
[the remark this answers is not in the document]
vam-google: `service_yaml` will go away soon as it is expected to be fully freplaced with new annotations. Removing it here (maybe a little bit to early, but it feels like an ok thing to do). WDYT?

rules_gapic/gapic.bzl
vam-google: Added for language. For `gapic_yaml` it is to support cases when we don't need any gapic_yaml (even v2).

vam-google:
@eoogbe, @andreamlin  PTAL

rules_gapic/gapic.bzl
andreamlin: microgenerators?

rules_gapic/gapic.bzl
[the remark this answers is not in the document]
andreamlin: I think we could this validation logic to the generator to ensure that if a GAPIC v1 is given, that a service yaml is also given. Feel free to do that in this PR or file an issue.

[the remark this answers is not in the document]
eoogbe: The service yaml can't be fully replaced by the annotations because of LRO http bindings. Annotations for those aren't even in the works. It will still be mandatory in gapic-generator for a while.

rules_gapic/gapic.bzl
vam-google: Yes, sorry, I meant microgenerators.

rules_gapic/gapic.bzl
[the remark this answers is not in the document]
vam-google: @eoogbe but LRO is needed only by a subset of clients. Note, if at least one client can be generated without service.yaml, it becomes optional.

vam-google:
@eoogbe PTAL

PR #2744, opened 2019-04-30
SampleGen: C# main method and CLIs

hzyi-google:


hzyi-google:
=> @googleapis/samplegen @chrisdunelm PTAL

src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
```diff
@@ -14419,23 +14419,20 @@ class DeleteShelfAsyncFlattenedAsyncEmptyResponseTypeWithResponseHandling
...
+   // [END sample_core]
    }
 
    // [END sample]
-   public static void Main(String[] args)
+   public static Task Main(String[] args)
    {
-   	// FIXME: pass in commandline arguments
-   	// FIXME: call the sample function
+   	SampleDeleteShelf();
```
chrisdunelm: This needs to use async:

```
public static async Task Main(...)
{
    await SampleDeleteShelf();
}
```

Or alternatively pass through the returned task, although I would favour the above code.

```
public static Task Main(...)
{
    return SampleDeleteShelf();
}
```

src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
```diff
@@ -14419,23 +14419,20 @@ class DeleteShelfAsyncFlattenedAsyncEmptyResponseTypeWithResponseHandling
...
+       ShelfName name = new ShelfName("[SHELF_ID]");
+       await libraryServiceClient.DeleteShelfAsync(name);
+       // FIXME: inspect the results
+   }
+   // [END sample_core]
    }
 
    // [END sample]
-   public static void Main(String[] args)
+   public static Task Main(String[] args)
```
chrisdunelm: `String` should really be `string`.
Using `String` will work fine, but `string` is more idiomatic :)

src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
```diff
@@ -14799,34 +14775,31 @@ class FindRelatedBooksAsyncFlattenedAsyncPagedAllOdyssey
...
-           {
-               // FIXME: inspect the results
-           }
+           new BookName("[SHELF_ID]", "[BOOK_ID]"),
+       };
+       IEnumerable<ShelfName> shelves = new[]
+       {
+           new ShelfName("[SHELF_ID]"),
+       };
+       await gax::PagedAsyncEnumerable<FindRelatedBooksResponse, BookName> response = libraryServiceClient.FindRelatedBooksAsync(names, shelves);
```
chrisdunelm: The `await` is in the wrong place ;)
It should be `... response = await libraryServiceClient...`

src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
```diff
@@ -14799,34 +14775,31 @@ class FindRelatedBooksAsyncFlattenedAsyncPagedAllOdyssey
...
+       // Iterate over pages (of server-defined size), performing one RPC per page
+       foreach (BookName item in response)
```
chrisdunelm: It is not possible to use a `foreach` loop over an asynchronous enumerable (unfortunately).

```
await response.ForEachAsync((BookNam item) =>
{
    // FIXME: inspect the results
});
```

hzyi-google:
PTAL @chrisdunelm

src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
```diff
@@ -14419,23 +14419,20 @@ class DeleteShelfAsyncFlattenedAsyncEmptyResponseTypeWithResponseHandling
...
+       ShelfName name = new ShelfName("[SHELF_ID]");
+       await libraryServiceClient.DeleteShelfAsync(name);
+       // FIXME: inspect the results
+   }
+   // [END sample_core]
    }
 
    // [END sample]
-   public static void Main(String[] args)
+   public static Task Main(string[] args)
```
chrisdunelm: Sorry, this still needs an `async` before the `Task`:
`public static async Task Main(string[] args)`

src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
```diff
@@ -14799,34 +14775,31 @@ class FindRelatedBooksAsyncFlattenedAsyncPagedAllOdyssey
...
    {
-   	// FIXME: pass in commandline arguments
-   	// FIXME: call the sample function
+   	await SampleFindRelatedBooks();
```
chrisdunelm: Just noticed that this should probably be calling `SampleFindRelatedBooksAsync()`.

src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
```diff
@@ -14977,35 +14944,32 @@ class FindRelatedBooksAsyncRequestAsyncPagedAllOdyssey
...
-           };
-           gax::PagedAsyncEnumerable<FindRelatedBooksResponse, BookName> response = libraryServiceClient.FindRelatedBooksAsync(request);
-           // Iterate over pages (of server-defined size), performing one RPC per page
-           foreach (BookName item in response)
-           {
-               // FIXME: inspect the results
-           }
-       }
-       // [END sample_core]
+           BookNames = {
```
chrisdunelm: This isn't idiomatic C# formatting. Although I know it's sometimes hard to get this right using snippets; and this may well be being formatted by existing snippet code that's been incorrect for ages ;)

```
FindRelatedBooksRequest request = new FindRelatedBooksRequest
{
    BookNames =
    {
        new BookName(...),
    },
    ShelvesAsShelfNames =
    {
        new ShelfName(...),
    },
};
```

src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
```diff
@@ -15516,23 +15456,20 @@ class FindRelatedBooksRequestPagedPageSizeOdyssey
...
-       {
-           LibraryServiceClient libraryServiceClient = await LibraryServiceClient.CreateAsync();
-           BookName name = new BookName("[SHELF_ID]", "[BOOK_ID]");
-           stt::Task<Book> response = libraryServiceClient.GetBookAsync(name);
-           // FIXME: inspect the results
-       }
-       // [END sample_core]
+       LibraryServiceClient libraryServiceClient = await LibraryServiceClient.CreateAsync();
+       BookName name = new BookName("[SHELF_ID]", "[BOOK_ID]");
+       await stt::Task<Book> response = libraryServiceClient.GetBookAsync(name);
```
chrisdunelm: This `await` is incorrectly positioned. It should be after the `=`; and the type of `response` should not be a `Task`
`Book response = await libraryServiceClient.GetBookAsync(name);`

There are various other cases of this below.

src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
```diff
@@ -15699,34 +15627,42 @@ class PublishSeriesAsyncFlattenedAsyncPiVersion
...
 
    // [END canonical]
-   public static void Main(String[] args)
+   public static Task Main(string[] args)
    {
-   	// FIXME: pass in commandline arguments
-   	// FIXME: call the sample function
+   	CommandLine.Parser.Default.ParseArguments<Options>(args)
+    	.withParsed<Options>(async opts =>
+    	    await SamplePublishSeries(opts.ShelfName, opts.Edition));
```
chrisdunelm: I expect this should be calling `SamplePublishSeriesAsync(...)`.

src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
```diff
@@ -15699,34 +15627,42 @@ class PublishSeriesAsyncFlattenedAsyncPiVersion
...
+   }
+
+   public class Options
+   {
+   	[Option("shelf_name", DEFAULT = "Math")]
+        public string ShelfName { get; set; }
```
chrisdunelm: nit: Looks like there's an extra indent character here.

hzyi-google:
@chrisdunelm Thanks for the detailed review! PTAL again :)

src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
hzyi-google: Fighting snippets is sometimes like fighting css :(

src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
```diff
@@ -15691,43 +15649,51 @@ class GetBookRequestTestOnSuccessMap
...
+    public static async Task SamplePublishSeriesAsync(string shelfName, uint edition)
+    {
+        LibraryServiceClient libraryServiceClient = await LibraryServiceClient.CreateAsync();
+        // string shelfName = "Math"
+        // uint edition = 123
+        Shelf shelf = new Shelf
+        {
+            Name = "Math",
+        };
+        scg::IEnumerable<Book> books = new scg::List<Book>();
```
beccasaurus:
```suggestion
        IEnumerable<Book> books = new List<Book>();
```

Yuck, kill any of `foo::` that you see :D

src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
```diff
@@ -15913,43 +15881,51 @@ class PublishSeriesAsyncRequestAsyncSecondEdition
...
+        // FIXME: inspect the results
+    }
+    // [END canonical_core]
+    }
+
+    // [END canonical]
+    public static void Main(string[] args)
+    {
+        CommandLine.Parser.Default.ParseArguments<Options>(args)
+        \t.withParsed<Options>(opts =>
```
beccasaurus: Bug!

`\t`

src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
chrisdunelm: Yes, good point. In a sample this namespace aliasing is not desirable.

@hzyi-google Changing `CSharpAliasMode.Global` to `CSharpAliasMode.MessagesOnly` when instantiating the `CSharpSurfaceNamer` for sample-generation will fix this for all sample code.

chrisdunelm: (Or the `scg::` may just be a literal within the snippit? I'm not sure...)

src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
chrisdunelm: Sad but true!

src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
chrisdunelm: C# source doesn't generally use tabs, so this should be replaced with 4 spaces.

hzyi-google:
@chrisdunelm I am able to generate [samples](https://github.com/hzyi-google/sample-staging/blob/master/UnarySyncSample.cs) for language. Both of the sync and async ones are compilable (I had to manually add the imports though). It's hard to do more thorough tests at this moment  because some key features are not in yet (imports, packaging metadata, inspecting results etc).

There do seem to be some runtime issues with both samples. 

The sync sample seems to be stuck at the API call forever and I never and a response back.
The async sample exits without actually waiting for the `await languageServiceClient.AnalyzeSyntaxAsync(request);` to finish (something wrong with the way CLI parser is used?)

Do you happen to know what's going on here?

hzyi-google:
@chrisdunelm @beccasaurus 

A bit update on this hanging PR...

I solved the problem that sync samples are stuck forever. It turns out to be an unsupported openssl config in dotnet core on Linux. Changing the config on my workstation solved the issue :)

For async samples, I changed the async sample function to block (`.Wait()`) in Main method. My reasons are: 1) The thread will exit after sample function finishes anyway, 2) the main method will not be shown on devsite, it's just meant to help us invoke the sample function. 3) I did not find a way to make async methods work with commandl line parser...

WDYT? PTAL again when you have time!

Also @beccasaurus mentions that cloud DPEs don't actually write samples for async methods... So this became a YAGNI feature. So IMHO unless the current generated samples are wrong or misleading, we probably don't have to spend too much time on them.

Issue #2738, opened 2019-04-26
Allow paging methods to lack a page_size field, for GAPIC config v2


andreamlin:
GAPIC config v1 allows for not specifying a page_size field. 

GAPIC config v2 requires any method that wants to be represented as a paged method to have a page_size field.

This is needed to support **Bigquery v2**, which doesn't have a `page_size` field in their paged methods. Instead they use a wrapper type `google.protobuf.UInt32Value max_results` , see [models.proto](https://github.com/googleapis/googleapis/blob/808110e242c682d7ac2bab6d9c49fc3bf72d7604/google/cloud/bigquery/v2/model.proto#L498).

andreamlin:
Another option is to add or keep the toggle to GAPIC config v2 for a method to be paging.

PR #2745, opened 2019-05-01
Bring paging back to GAPIC config v2

andreamlin:
For https://github.com/googleapis/gapic-generator/issues/2738.

This allows us to maintain existing behavior, which allows for `page_size` to [not be specified](https://github.com/googleapis/gapic-generator/blob/4277f45eac1cbb131284094d26776ac3b7eba392/src/main/java/com/google/api/codegen/config/PageStreamingConfig.java#L41) in a `List[Resources]` method. This differs from the guidelines in https://cloud.google.com/apis/design/design_patterns#list_pagination, but is necessary to support Bigquery v2 in the short term.

Test resources: Make `ListShelves` a paginated method again; the request object lacks a `page_size` field.


andreamlin:
PTAL

PR #2748, opened 2019-05-03
SampleGen: put PHP samples in the correct directory

hzyi-google:
- Samples now live in `.../google-cloud-language-v1/samples/V1/`.
- Change sample file names to `UpperCamelCase.php`.


hzyi-google:
=> @googleapis/samplegen 

@beccasaurus as you don't seem to have preference yet whether `samples/` should be under `src/`, I did a wild guess and pulled them out. It makes more sense as they now follow the same pattern as tests :)

We can definitely change if this isn't ideal. Shouldn't be hard at all.

hzyi-google:
@andreamlin Do you mind taking a look as well, as both Michael and Victor are out ;)

andreamlin:
I think the php CircleCI test will sort itself out once the other PR to update the php-gax version is merged in.

hzyi-google:
@andreamlin Yes the diff is just caused by reordering.

beccasaurus:
<3 👍

PR #2750, opened 2019-05-05
PHP: Update gax version to 1.0.2

hzyi-google:


hzyi-google:
@dwsupplee I tried your suggestions and updated gax to 0.38.0 but that didn't actually work. However updating gax to 1.0.2 made the tests pass.

PR #2752, opened 2019-05-06
Remove all references of DEADLINE_EXCEEDED

andreamlin:
Stop generating DEADLINE_EXCEEDED as a code to retry on.

DEADLINE_EXCEEDED was likely an incorrect error code to retry on.


andreamlin:
PTAL

cc @jadekler

jadekler:
LG :+1:

PR #2754, opened 2019-05-06
SampleGen: Update Java template

### LINKED ISSUE #2753: Java sample region END tag should appear after sample method
Currently, the END tag shows up at the very bottom of the file, so it shows the main() method and CLI parsing/etc.

We don't want this.

Please move to be right after the method, so the START/END wraps the method:

```
START
// commented out imports
method () {

}
END
```

/cc @googleapis/samplegen

hzyi-google:
Fix #2753

### DIFF
--- a/src/test/java/com/google/api/codegen/protoannotations/testdata/java_library.baseline
+++ b/src/test/java/com/google/api/codegen/protoannotations/testdata/java_library.baseline
@@ -994,7 +994,7 @@ public class GetBigBookAsyncLongRunningFlattenedAsyncWap {
   public static void sampleGetBigBook(String shelf) {
     // [START hopper_core]
     try (LibraryClient libraryClient = LibraryClient.create()) {
-      // String shelf = "Novel";
+      // shelf = "Novel";
       BookName name = ShelfBookName.of("[BOOK_SHELF]", "[BOOK]");
       Book response = libraryClient.getBigBookAsync(name.toString()).get();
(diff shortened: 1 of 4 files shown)


hzyi-google:
PTAL

src/test/java/com/google/api/codegen/protoannotations/testdata/java_library.baseline
```diff
@@ -3443,17 +3442,15 @@ public class TuringProgCallableStreamingBidi {
...
-      byte[] data = Files.readAllBytes(path);
-      ByteString comment = ByteString.copyFrom(data);
+      byte[] data = Files.readAllBytes(path);ByteString comment = ByteString.copyFrom(data);
       Comment comment2 = Comment.newBuilder()
         .setComment(comment)
         .build();
       path = Paths.get(imageFileName);
-      data = Files.readAllBytes(path);
-      ByteString image = ByteString.copyFrom(data);
+      data = Files.readAllBytes(path);ByteString image = ByteString.copyFrom(data);
```
andreamlin: It might look prettier if we kept the `data = ...` and `ByteString ...` lines separate

hzyi-google: Thanks for the catch! It's really strange that whenever I wrapper something with an `@if` block I have to add a new blank line to preserve the original output.

hzyi-google:
PTAL

Issue #2141, opened 2018-07-10
[PHP] Namespace inconsistency during deprecation


tmatsuo:
We're trying to update the KMS client lib for PHP. The newly generated code is using a nicer namespace like `Google\Cloud\Kms\V1\CryptoKey\CryptoKeyPurpose` instead of `Google\Cloud\Kms\V1\CryptoKey_CryptoKeyPurpose`. Also the generator keeps the old namespaced class with a deprecation warning.

This is a great feature, but we found there are some usages of old classes in various places.

1. [`Google\Cloud\Kms\V1\Gapic\KeyManagementServiceGapicClient`](https://github.com/tmatsuo/google-cloud-php/blob/95d2dd7215af1fd20892b573fcdea8244939c843/Kms/src/V1/Gapic/KeyManagementServiceGapicClient.php#L890)
2. [`\Google\Cloud\Kms\V1\CryptoKey.php`](https://github.com/tmatsuo/google-cloud-php/blob/95d2dd7215af1fd20892b573fcdea8244939c843/Kms/src/V1/CryptoKey.php#L198)
3. Some occurrences in the unit tests [`\Google\Cloud\Kms\Tests\Unit\V1\KeyManagementServiceClientTest`](https://github.com/tmatsuo/google-cloud-php/blob/95d2dd7215af1fd20892b573fcdea8244939c843/Kms/tests/Unit/V1/KeyManagementServiceClientTest.php#L34)

For 1, this is the sample code in the doc comment, so it is basically telling users to use the deprecated class which will be removed with the next release. I think the sample code should use the newer class name.

For 2 and 3, I still think it is cleaner to use the newer namespace.

vchudnov-g:
@michaelbausor Updates?

andreamlin:
@michaelbausor Updates?

jbolinger:
@michaelbausor any updates?

landrito:
@michaelbausor any updates?

jbolinger:
@michaelbausor any updates?

michaelbausor:
No update at this time.

vchudnov-g:
@michaelbausor Could you provide an update and/or lower the priority if appropriate?

michaelbausor:
This will be fixed in https://github.com/googleapis/gapic-generator/pull/2757

vchudnov-g:
@hzyi-google Did #2757 fix this? If so could you close this issue?

hzyi-google:
[...]

PR #2757, opened 2019-05-07
Sample: Use slashes instead of underscores in PHP namespaces for nested enums and messages

hzyi-google:


hzyi-google:
PTAL

hzyi-google:
We don't usually enforce that... but it does seem the changes in this PR could use some more testing. I'll see if I can reach the target code coverage.

hzyi-google:
PTAL

hzyi-google:
In this example, the second `Hello` will be used by its full name.
Added a test to show this logic too.

src/main/java/com/google/api/codegen/transformer/php/PhpModelTypeNameConverter.java
```diff
@@ -153,67 +152,48 @@ public TypeName getTypeName(ProtoElement elem) {
...
+        builder.append('\\').append(CharMatcher.is('\\').trimFrom(namespace));
+        for (String component : components) {
           builder
-              .append('\\')
-              .append(components[index].substring(0, 1).toUpperCase())
-              .append(components[index].substring(1));
+              .append("\\")
+              .append(component.substring(0, 1).toUpperCase())
+              .append(component.substring(1));
+          return typeNameConverter.getTypeName(builder.toString());
```
michaelbausor: What is the point of this `for` loop if we always return on the first element?

michaelbausor:
Fixes https://github.com/googleapis/gapic-generator/issues/2141

hzyi-google:
@michaelbausor Thanks for the catch! Fixed and also added a test. PTAL

src/main/java/com/google/api/codegen/transformer/php/PhpModelTypeNameConverter.java
```diff
@@ -153,67 +152,48 @@ public TypeName getTypeName(ProtoElement elem) {
...
-   *   <li>\Example\Top_Nested
-   *   <li>\Example\Top_Nested_DeepNested
+   *   <li>\Example\Top\Nested
+   *   <li>\Example\Top\Nested\DeepNested
    * </ul>
-   *
-   * <p>To correctly output these type names, we need to check whether the parent of a proto element
-   * is a message, and if so use '_' as a separator.
    */
   private TypeName getTypeName(ProtoElement elem, int maxDepth) {
```
michaelbausor: maxDepth is unused now - can we remove it?

src/main/java/com/google/api/codegen/transformer/php/PhpModelTypeNameConverter.java
```diff
@@ -153,67 +152,48 @@ public TypeName getTypeName(ProtoElement elem) {
...
+        return typeNameConverter.getTypeName(builder.toString());
       }
     }
 
-    builder.append('\\').append(shortName);
-    return builder.toString();
+    // Create the type name based on the element's full name
+    for (String component : fullName.split("\\.")) {
+      builder
+          .append("\\")
```
michaelbausor: This step of splitting, appending `\`, capitalizing, and rejoining is duplicated. Consider the following refactor: create a helper function:

```java
private static String makeNamespacePiece(String piece) {
  return "\\" + piece.substring(0, 1).toUpperCase() + piece.substring(1);
}
```

Then use streams:

```java
String s = fullName.split("\\.").stream().transform(makeNamespacePiece).joinStrings();
```

(I don't think joinStrings() is a thing, but use a collector to join the strings in some way).

Issue #2763, opened 2019-05-10
[C#] Partial classes not generated for Resource classes with proto annotations


andreamlin:
e.g. for Vision API, there is a message 
```// A Product contains ReferenceImages.
message Product {
  option (google.api.resource) = {
    pattern: "projects/{project}/locations/{location}/products/{product}",
    type: "vision.googleapis.com/Product"
  }; ...
}
```

But there is no generated 
```
public partial class Product
    {
        /// <summary>
        /// <see cref="Google.Cloud.Vision.V1.ProductName"/>-typed view over the <see cref="Name"/> resource name property.
        /// </summary>
        public Google.Cloud.Vision.V1.ProductName ProductName
        {
            get { return string.IsNullOrEmpty(Name) ? null : Google.Cloud.Vision.V1.ProductName.Parse(Name); }
            set { Name = value != null ? value.ToString() : ""; }
        }

    }
```

PR #2764, opened 2019-05-10
Add partial class resource names for C for resource types

andreamlin:
Fixes https://github.com/googleapis/gapic-generator/issues/2763.

Currently, we only generate the C# resource name partial classes from `resource_reference`s. However, if there is a message type that represents a resource, it is unnecessary to annotate the `name` field with a resource_reference (see issue). We should still generate the partial class for this resource message type.

### DIFF
--- a/src/test/java/com/google/api/codegen/testsrc/common/library.proto
+++ b/src/test/java/com/google/api/codegen/testsrc/common/library.proto
@@ -366,8 +366,7 @@ message Book {
   // Book names have the form `bookShelves/{shelf_id}/books/{book_id}`.
   // Message field comment may include special characters: <>&"`'@.
   string name = 1 [
-    (google.api.field_behavior) = REQUIRED,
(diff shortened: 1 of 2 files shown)


src/test/java/com/google/api/codegen/testsrc/common/library.proto
```diff
@@ -366,8 +366,7 @@ message Book {
   // Book names have the form `bookShelves/{shelf_id}/books/{book_id}`.
   // Message field comment may include special characters: <>&"`'@.
   string name = 1 [
-    (google.api.field_behavior) = REQUIRED,
-    (google.api.resource_reference).type = "library.googleapis.com/Book"
```
andreamlin: Without the source files change, this proto change would result in C# baseline not having a partial class for the resource name in the Book class.

andreamlin:
PTAL

PR #2765, opened 2019-05-13
C# SampleGen: correct the output path of samples and add support for default calling forms for unary and paging

hzyi-google:
`Samples` -> `Google.Cloud.Language.V1.Samples`.

Again, apologies for the big diff in baselines. The contents of the samples generated shouldn't change at all, all that changed are the file names.

Generated samples for Language using this branch: https://github.com/hzyi-google/sample-staging/tree/master/20190513/csharp

Note: I realized that the location of the region tags is wrong, but let's fix it in a separate PR since this PR is already causing huge changes in the baseline.


hzyi-google:
PTAL @chrisdunelm @googleapis/samplegen

PR #2767, opened 2019-05-15
node: do not use lodash.merge, use latest gax

alexander-fenster:
For merging protos, let's just pass the list of protos to `gaxGrpc.loadProto` since it actually supports it.


alexander-fenster:
Nope - we still need something better than `Object.assign`. To be updated...

alexander-fenster:
@JustinBeckwith one more look at the generated code?

PR #2769, opened 2019-05-16
remove the condition that is always false

xuejiangtao:
The condition `type == null` is always false. If this condition was true,  this line ` Preconditions.checkArgument(
        !type.isRepeated() && !type.isMap(),
        "Expected non-repeated fields in print statement, found %s",
        type.getTypeName());
` would throw exception.
So to improve the code quality, consider to remove the condition that is always false.

[https://rules.sonarsource.com/java/type/Bug/RSPEC-1145](url)


xuejiangtao:
> I signed it!

andreamlin:
Thanks!

xuejiangtao:
> Thanks!

thanks for merging! I would like do some contribution for this project to improve the Code quality using my tool to automatically fix the CQI(Code Quality Issue).

PR #2775, opened 2019-05-17
Avoid coverage failure when no `_shared_modules` are present.

tseaver:
See: https://github.com/googleapis/google-cloud-python/pull/7998/files

### DIFF
--- a/src/main/resources/com/google/api/codegen/py/types.snip
+++ b/src/main/resources/com/google/api/codegen/py/types.snip
@@ -24,7 +24,7 @@
 
   names = []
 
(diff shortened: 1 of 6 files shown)


busunkim96:
I think the baseline files also need to be updated: https://github.com/googleapis/gapic-generator/tree/6c5016e6f6c4cda471b498fb65377e741a4c4be6/src/test/java/com/google/api/codegen/gapic/testdata/py

tseaver:
@busunkim96 I should've thought of that:  editing the file in the browser meant I wasn't thinking about tests.

PR #2781, opened 2019-05-20
SampleGen: generate a Gemfile for ruby sample package

hzyi-google:


src/main/java/com/google/api/codegen/transformer/ruby/RubySamplePackageMetadataTransformer.java
```diff
@@ -0,0 +1,94 @@
+/* Copyright 2016 Google LLC
+ *
```
vchudnov-g: 2019

src/test/java/com/google/api/codegen/gapic/testdata/ruby/ruby_library.baseline
```diff
@@ -5382,6 +5382,10 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency "yard", "~> 0.9"
 end
 
+============== file: samples/Gemfile ==============
+source "https://rubygems.org"
+
+gem "google-cloud-library", "../../"
```
beccasaurus: This should be:

```rb
gem "google-cloud-library", path: "../../"
```

A string argument following `gem` is the version string:

```rb
gem "google-cloud-library", ">=1.0"
```

PR #2782, opened 2019-05-21
SampleGen: generate sample metadata

### LINKED ISSUE #2771: feat: generated samples should include metadata for README generation
When generating samples, samples should be populated with a meta-information header.

As an example, for JavaScript, a header might look something like this:

```js
// sample-metadata:
//   title: Create Job
//   description: Create a job that posts to /log_payload on an App Engine service.
//   usage: node createJob.js [project-id] [location-id] [app-engine-service-id]
```

These headers can be parsed by tools generating READMEs and other documentation.

hzyi-google:
Generate sample metadata per https://docs.google.com/document/d/1lGvj1U-FfiY2Z_bR4lq6Jcs0BwPMn9bm-hzTyw6T3no

Fix #2771.


src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
```diff
@@ -17,6 +17,11 @@
 
 // This is a generated sample ("FlattenedAsync", "empty_response_type_with_response_handling")
 
+// sample-metadata
+//   title: Test response handling for methods that return empty
+//   description: Test response handling for methods that return empty
+//   usage: dotnet run
```
hzyi-google: Is it necessary to call out the main class here as well, similar to Java?
e.g., `dotnet run /p:StartupObject=Namespace.MainClass`?

src/test/java/com/google/api/codegen/protoannotations/testdata/ruby_library.baseline
```diff
@@ -6228,6 +6308,11 @@ end
 
 # DO NOT EDIT! This is a generated sample ("Request",  "test_write_to_file")
 
+# sample-metadata
+#   title: test writing fields to files
+#   description: Testing write fields to files.
+#   bundle exec ruby samples/v1/publish_series_request_test_write_to_file.rb
```
hzyi-google: `ruby some_file.rb` should be fine, but I think most of the time people invoke ruby code through `bundle exec`

src/main/resources/com/google/api/codegen/py/standalone_sample.snip
```diff
@@ -11,6 +12,10 @@
       @# To install the latest published package dependency, execute the following:
       @#   pip install {@sampleFile.gapicPackageName}
 
+      @# sample-metadata
+      @#   title: {@sample.title}
+      @#   description: {@sample.description}
+      @#   usage: python3 {@sampleFile.outputPath} {@commandlineArgmentList(sample.sampleInitCode.argDefaultParams)}
```
hzyi-google: Is it necessary to explicitly call out `python3` here?

hzyi-google:
PTAL @googleapis/samplegen

src/main/resources/com/google/api/codegen/metadatagen/samplegen.snip
```diff
@@ -0,0 +1,5 @@
+@snippet commandlineArgmentList(args)
```
vchudnov-g: s/Argment/Argument/

src/main/resources/com/google/api/codegen/py/standalone_sample.snip
vchudnov-g: Our code assumes python3, right? So if we don't have a `#!/usr/bin/env python3` at the top, we should. If we do have that hash-bang, I don't have a strong opinion (slightly leaning towards leaving it in for explicitness)

src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
vchudnov-g: Also, is this binary name the same in all platforms (eg on dotnet on Windows vs mono on Linux)

src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
```diff
@@ -59,6 +64,11 @@ public class DeleteShelfAsyncFlattenedAsyncEmptyResponseTypeWithResponseHandling
 
 // This is a generated sample ("FlattenedAsync", "empty_response_type_without_response_handling")
```
vchudnov-g: Missing the `// DO NOT EDIT!` prefix

src/test/java/com/google/api/codegen/gapic/testdata/java/java_library.baseline
```diff
@@ -1593,6 +1693,11 @@ public class GetBigBookOperationCallableLongRunningCallableWap {
  * limitations under the License.
  */
 // DO NOT EDIT! This is a generated sample ("LongRunningCallable",  "wap2")
+// sample-metadata:
+//   title: GetBigBook: 'War and Peace'
+//   description: Testing resource name overlap
+//   usage: gradle run -PmainClass=com.google.example.examples.library.v1.GetBigBookOperationCallableLongRunningCallableWap2 [--args='[--shelf Novel] [--big_book_name "War and Peace"]']
```
vchudnov-g: Do we want the usage to refer to `gradle run` or simply to `java`? It seems like the latter is more general and also how we're setting the invocation for the sample tests.

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
```diff
@@ -936,4 +937,25 @@ private static String getVariableName(MethodContext context, InitCodeNode item)
...
     return context.getNamer().localVarName(item.getIdentifier());
   }
+
+  private static String getCliFlagDefaultValue(InitCodeNode item) {
+    checkArgument(
+        !item.getType().isMessage(), "Only enums and primitive types are supported for now.");
+
+    String value = item.getInitValueConfig().getInitialValue().getValue();
+
+    if (item.getType().isStringType()) {
```
vchudnov-g: Are these two conditions captured in a test?

src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
hzyi-google: Based on: https://stackoverflow.com/questions/37738106/net-core-vs-mono

`dotnet` is the binary name for dotnet core which is developed by Microsoft and works on Linux. So I think we should use that?

src/test/java/com/google/api/codegen/gapic/testdata/java/java_library.baseline
hzyi-google: I have this concern too.

The good thing about `gradle run` is that we generate a `build.gradle` for the sample package, so the command would just work on the generated samples.

The main concern about `java` is that depending on how people compiles the java code (`javac` or mvn/gradle`) the compiled jar might be in different directories, so it's hard to provide a command that just works.

Probably this is a question for @beccasaurus ?

src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
hzyi-google: No. I'll add a few tests.

src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
vchudnov-g: Looking at https://docs.microsoft.com/en-us/dotnet/core/linux-prerequisites?tabs=netcore2x, that seems to be correct. And it's also the name on Windows, right? OK.

src/test/java/com/google/api/codegen/gapic/testdata/java/java_library.baseline
vchudnov-g: Hmm, I see. Given that we generate the gradle files, maybe leaving it to refer to gradle makes sense. (I'd still prefer to have it more generic, but it seems that's likely to work less than simply referring to our gradle config)

beccasaurus:
Ping. Looks like, although it's Approved, there are outstanding comments. @hzyi-google

src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
hzyi-google: It's there two lines above this line.

(C# has `DO NOT EDIT` in the license header, so that is removed from this line)

hzyi-google:
@vchudnov-g PTAL

beccasaurus:
Note, for reference, this fixes #2771

src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
vchudnov-g: Not to keen about changing the format of our boilerplate line, but since that will likely go away real soon, it doesn't matter....

PR #2784, opened 2019-05-22
SampleGen: namespace and imports in C#

hzyi-google:


hzyi-google:
PTAL @googleapis/samplegen

src/main/java/com/google/api/codegen/transformer/csharp/CSharpSampleImportTransformer.java
```diff
@@ -0,0 +1,81 @@
+/* Copyright 2018 Google LLC
```
vchudnov-g: 2019

src/main/java/com/google/api/codegen/transformer/csharp/CSharpSampleImportTransformer.java
```diff
@@ -0,0 +1,81 @@
...
+      typeTable.getAndSaveNicknameFor(node.getType());
+      switch (node.getLineType()) {
+        case ReadFileInitLine:
+          typeTable.saveNicknameFor("System.IO.File");
+          break;
+        case SimpleInitLine:
+        case StructureInitLine:
+        case ListInitLine:
+        case MapInitLine:
+          break; // fall through
```
vchudnov-g: `break; // fall through` is confusing

You probably want each of the lines above to have the comment:
```
case SimpleInitLine:  // fall-through
...
caseMapInitLine:
  break;
```

chrisdunelm:
Sorry, I managed to overlook this PR. I'll review as soon as I can.

src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
```diff
@@ -17,27 +17,33 @@
...
-    {
-        LibraryServiceClient libraryServiceClient = await LibraryServiceClient.CreateAsync();
-        ShelfName name = new ShelfName("[SHELF_ID]");
-        await libraryServiceClient.DeleteShelfAsync(name);
-        // FIXME: inspect the results
-    }
-    // [END sample_core]
+    // [START sample]
+    using Google.Example.Library;
+    using apis = Google.Example.Library.V1;
```
beccasaurus: Where did you copy this from?

[Should be](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/8665b94b10f8ad99b09004a180f21209688a05f9/speech/api/Recognize/Recognize.cs#L19):
```cs
using Google.Example.Library.V1;
```

src/main/java/com/google/api/codegen/transformer/csharp/CSharpSampleImportTransformer.java
hzyi-google: Actually, found a better way using `Stream`

hzyi-google:
PTAL @beccasaurus @vchudnov-g

PR #2785, opened 2019-05-22
Java: Move samples outside of the gapic artifact

hzyi-google:
Works together with: https://github.com/googleapis/artman/pull/624

- Previous to this PR, generated gapics are put in `[artman-output-path]/src/main/java/...`, where [artman-output-path] is determined by artman, and looks like for example `artman-output/java/gapic-google-cloud-language-v1`. Since the entry point of gapic-generator is already within the gapic directory, generated samples have been entangled with gapics.

- This PR will change the output directory of gapic source code to `[artman-output-path]/gapic/src/main/java` and `[artman-output-path]/samples/src/main/java`. In the companion PR, artman will change the output directory to `artman-output/java`, and when the generation finishes, rename `artman-output/java/gapic` to `artman-output/java/gapic-google-cloud-api-version`, and `artman-output/java/samples` to `artman-output/java/samples-google-cloud-api-version`.

- The two changes together will pull samples out into a separate directory, `samples-google-cloud-api-version` and will make testing and shipping Java samples easier and less error-prone.

For More details see: https://docs.google.com/document/d/1XfuXoLQ8tePFtf77sDFxVUrq84m7b1JXXUQpbTXqYE0/edit

Note the Java tests will unavoidably fail due to changes in artman not committed.


hzyi-google:
PTAL @andreamlin @vam-google 
FYI @googleapis/samplegen

hzyi-google:
Also @michaelbausor, do you think this change would by any chance break Ads packaging?

michaelbausor:
@hzyi-google I don't have access to the linked doc

michaelbausor:
Also - why is there such a massive change in the Java baseline?

hzyi-google:
The massive baseline change is caused by reordering of files. Because this PR changed the output path of a lot of files and baseline puts files in alphabetic order. For example, 
gapic files: `src/main/java/...` -> `gapic-google-cloud-library-v1/src/main/java...`
sample files: `samples/src/main/java/...` -> `sample-google-cloud-library-v1/src/main/java...`

So all the samples will be moved up and precede the gapics.

michaelbausor:
Re Ads - I don't think so, but it will need to be tested for sure before we release a new g3artman that uses this update.

vchudnov-g:
@hzyi-google Do you think we'll need anything similar in the other languages?

hzyi-google:
@vchudnov-g I don't think so. This should be specific to Java.

hzyi-google:
Updated both PRs as per https://github.com/googleapis/artman/pull/624#issuecomment-495818562.

PTAL

hzyi-google:
Friendly ping. @andreamlin @vam-google

hzyi-google:
Talked offline: this PR is no longer needed.

PR #2796, opened 2019-05-29
C# Add ClientBuilder generation

chrisdunelm:


chrisdunelm:
(The PHP-7.2 test failure is unrelated to this change)

src/main/resources/com/google/api/codegen/csharp/gapic_client.snip
```diff
@@ -261,6 +302,8 @@
 
         private static readonly gaxgrpc::ChannelPool s_channelPool = new gaxgrpc::ChannelPool(DefaultScopes);
 
+        internal static gaxgrpc::ChannelPool ChannelPool => s_channelPool;
```
jskeet: At some point we could remove the field and make this an autoimplemented property, of course. One step at a time though...

chrisdunelm: ack :)

Issue #2511, opened 2019-01-11
cryptic null ptr error when `parameters.attributes.parameter` does not refer to an item listed in `parameters.defaults`


vchudnov-g:
In the `sample_value_sets` config, if `parameters.attributes.parameter` specifies a parameter not in `parameters.defaults`, then we get non-useful error message:

```
java.lang.NullPointerException
	at com.google.api.codegen.transformer.SurfaceNamer.getFieldSetFunctionName(SurfaceNamer.java:308)
```

This config segment seems to illustrate the problem:
```
      parameters:
        defaults:
        - document.content="Mary had a little lamb"
        - document.type=PLAIN_TEXT
        attributes:
        - parameter: document.gcs_content_uri
          sample_argument_name: foo
```

Issue #2793, opened 2019-05-29
SampleGen: improve error message on invalid parameter name in attributes


beccasaurus:
Doing this:

```yaml
parameters:
  attributes:
  - parameter: this_has_a_typo
    description: "..."
```

Results in this:

```
WARNING: toplevel: (lint) control-presence: Service speech.googleapis.com does not have control environment configured.
ERROR: toplevel: Unexpected exception:
java.lang.NullPointerException
```

Followed by a chunk of time trying to figure out what's undefined 😎

/cc @googleapis/samplegen

hzyi-google:
Fixed by ^^^^^ #2799.

PR #2799, opened 2019-05-29
SampleGen: Clearer error message when parameter default values are not set

hzyi-google:
It's hard to add a test case for this PR.

Manual test result:

When I comment out this in googleapis/googleapis/language/v1/gapic.yaml:
```yaml
parameters:
  defaults:
    # - document.content="I am so happy and joyful."
```

I am seeing the following error:
```
toplevel: Unexpected exception:
java.lang.IllegalArgumentException: Cannot configure attributes for parameter "document.content", default value not set.
```

Fix: #2511 #2793


hzyi-google:
PTAL

PR #2800, opened 2019-05-29
node: make apiEndpoint an alias for servicePath

### LINKED ISSUE #2787: [nodejs] Add support for `apiEndpoint` override
Today, we allow developers to override the service endpoint using the `servicePath` variable in the service client constructor:

```js
const client = new ImageAnnotatorClient({
  servicePath: 'vision.not-googleapis.com'
});
```

In go/extensible-options-in-nodejs, we discuss changing this to `apiEndpoint`, as it's consistent with existing hand written libs, and feel that it's a little more obvious:

```js
const client = new ImageAnnotatorClient({
  apiEndpoint: 'vision.not-googleapis.com'
});
```

Ideally, this wouldn't be a breaking change.  In the generated code, it should look something like this:

```js
const servicePath = 
  opts.servicePath || 
  opts.apiEndpoint || 
  this.constructor.servicePath;

opts = Object.assign({
  clientConfig: {},
  port: this.constructor.port
}, opts);
```

If we're feeling 🌶 , we could also add the `apiEndpoint` getter to match the contents of `servicePath`.

alexander-fenster:
Fixes #2787. Also, adds some test coverage for the new generated static method.

### DIFF
--- a/src/main/resources/com/google/api/codegen/nodejs/main.snip
+++ b/src/main/resources/com/google/api/codegen/nodejs/main.snip
@@ -68,7 +68,7 @@
         {
           clientConfig: {},
           port: this.constructor.port,
-          servicePath: servicePath,
+          servicePath,
         },
         opts
       );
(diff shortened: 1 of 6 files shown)


alexander-fenster:
The Ruby test failure is obviously unrelated and is @hzyi-google and @michaelbausor are looking into it.

PR #2801, opened 2019-05-30
Add GAPIC_CODE artifact type support for PHP

vam-google:
This is needed to support bazel generation and avoid generating `composer.json`, since it is already broken and should not be generated by the generator. `Gapic-generator` generates code, but not packaging-specific files, because same generated code can be packaged differently depending on now it is deployed/used.

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTestTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTestTransformer.java
@@ -22,7 +22,6 @@ import com.google.api.codegen.config.GapicProductConfig;
 import com.google.api.codegen.config.GrpcStreamingConfig.GrpcStreamingType;
 import com.google.api.codegen.config.InterfaceModel;
 import com.google.api.codegen.config.MethodModel;
-import com.google.api.codegen.config.PackageMetadataConfig;
 import com.google.api.codegen.config.ProtoApiModel;
 import com.google.api.codegen.metacode.InitCodeContext;
 import com.google.api.codegen.metacode.InitCodeContext.InitCodeOutputType;
@@ -55,11 +54,15 @@ import com.google.common.collect.ImmutableList;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /** Responsible for producing testing related views for PHP. */
 public class PhpGapicSurfaceTestTransformer implements ModelToViewTransformer<ProtoApiModel> {
   private static final String SMOKE_TEST_TEMPLATE_FILE = "php/smoke_test.snip";
   private static final String UNIT_TEST_TEMPLATE_FILE = "php/test.snip";
+  private static final Pattern API_NAME_PATTERN =
+      Pattern.compile("(?<apiName>\\w+)(\\\\[Vv]\\d+\\w*)*$");
 
   private final PhpImportSectionTransformer importSectionTransformer =
       new PhpImportSectionTransformer();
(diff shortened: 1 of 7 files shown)


src/test/java/com/google/api/codegen/gapic/testdata/php/php_no_path_templates.baseline
```diff
@@ -378,7 +378,7 @@ use PHPUnit\Framework\TestCase;
 use stdClass;
 
 /**
- * @group library
+ * @group example
```
andreamlin: How did this change happen?

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTestTransformer.java
```diff
@@ -73,11 +76,7 @@
   private final PhpFeatureConfig featureConfig = new PhpFeatureConfig();
   private final MockServiceTransformer mockServiceTransformer = new MockServiceTransformer();
 
-  private final PackageMetadataConfig packageConfig;
-
-  public PhpGapicSurfaceTestTransformer(PackageMetadataConfig packageConfig) {
-    this.packageConfig = packageConfig;
-  }
+  public PhpGapicSurfaceTestTransformer() {}
```
hzyi-google: Is it necessary to remove `PackageMetadataConfig` from the constructor of `PhpGapicSamplesTransformer` as well?

hzyi-google: We could remove this constructor right? Or is it here for readability?

vam-google: @hzyi-google Yes, this is the topic I wanted to discuss with @vchudnov-g  today. It does not have to be done in this PR, as it is a bigger change. 

The idea is that samplegen does not rely on `PackageMetadataConfig` for anything in any language. That object is created from a `.yaml` file, which is passed as `--package_yaml2` argument to the generator. That argument is a legacy thing (Garrett started removing dependency on it long time ago, removing most of the stuff from it and renaming it from `--package_yaml` to `--package_yaml2`). The problem with it is that it is generated by artman from artman yaml and it contains only packaging-related stuff (actually it contains almost no useful information at this point).

Since artman will go away, the artman_yaml will go too.

src/test/java/com/google/api/codegen/gapic/testdata/php/php_no_path_templates.baseline
vam-google: This is so because we stopped using artman_yaml's `api_name` value for this value but generate it from service package name instead. 

The package name for this service is actually `Google\Example\V1`. For example for language service (real one form `googleapis`) is `Google\Cloud\Language\V1`, so it would produce `@group langauge` here (same as the old behavior).

In general I believe the "example" is more correct name here, the fact that it was using "library" before was caused by us reusing same artman_yaml in tests for multiple gapics (which does not correspond to the real world scenario, where each service has its own artman yaml).

I confirmed with @michaelbausor and @dwsupplee that it is ok and is not going to break any tooling.

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTestTransformer.java
vam-google: There was no particular reason (using automatically generated one is equally fine, I believe). I can remove it.

hzyi-google: That sounds good to me. IIRC it should be very easy to decouple `PackageMetadataConfig` and `LangGapicSamplesTransformer`, at least in most languages.

PR #2802, opened 2019-05-30
Ruby: check both custom and wrapped exception types in unit tests

hzyi-google:
Similar to https://github.com/googleapis/google-cloud-ruby/pull/3427/files. To fix failed unit tests in CI.


hzyi-google:
PTAL

PR #2808, opened 2019-05-31
SampleGen: Choose the resource name that matches binding values for Oneof resource names

### LINKED ISSUE #2719: Additional Bindings support: choose resource path helper based on arguments given
> _This feature may already be supported ~ I just need to know how to call it :)_

Many API methods support `additional_bindings`*

### » Example use case

I want to **Create a Company** for a job board.

The job board supports an _optional feature_ for multi-tenancy. If I'm a big company, I can add N tenants under my GCP project and manage multiple separately scoped sets of Companies. Cool, eh? Totally. But I don't want to use that. I just want to create a Company.

#### 🔍 Relevant Excerpt from .proto

```proto
service CompanyService {
  // Creates a new company entity.
  rpc CreateCompany(CreateCompanyRequest) returns (Company) {
    option (google.api.http) = {
      post: "/v4beta1/{parent=projects/*/tenants/*}/companies"
      body: "*"
      additional_bindings {
        post: "/v4beta1/{parent=projects/*}/companies"
        body: "*"
      }
    };
  }
  // The Request of the CreateCompany method.
  message CreateCompanyRequest {
    // Required.
    //
    // Resource name of the tenant under which the company is created.
    //
    // The format is "projects/{project_id}/tenants/{tenant_id}", for example,
    // "projects/api-test-project/tenant/foo".
    //
    // Tenant id is optional and a default tenant is created if unspecified, for
    // example, "projects/api-test-project".
    string parent = 1;

    // Required.
    //
    // The company to be created.
    Company company = 2;
  }
  // A Company resource represents a company in the service.
  message Company {
    // Required during company update.
    //
    // The resource name for a company. This is generated by the service when a
    // company is created.
    //
    // The format is
    // "projects/{project_id}/tenants/{tenant_id}/companies/{company_id}", for
    // example, "projects/api-test-project/tenants/foo/companies/bar".
    //
    // Tenant id is optional and the default tenant is used if unspecified, for
    // example, "projects/api-test-project/companies/bar".
    string name = 1;

    // Required.
    //
    // The display name of the company, for example, "Google, LLC".
    string display_name = 2;
  }
}
```
#### 👩🏼‍💻 My code to create a company

```yaml
- id: create_company
  description: Create Company
  parameters:
    defaults:
    - parent%project="Your Google Cloud Project ID" # <--
    - company.display_name="My Company Name"
    attributes:
    - parameter: parent%project # <--
      sample_argument_name: project_id
    - parameter: company.display_name
      sample_argument_name: display_name
  on_success:
  - print: ["Created company: %s", $resp.name]
```

### » My expected result

```js
'use strict';

// [START job_search_create_company]
const talent = require('@google-cloud/talent').v4beta1;

/** Create Company */
function sampleCreateCompany(projectId, displayName) {
  const client = new talent.CompanyServiceClient();
  // const projectId = 'Your Google Cloud Project ID';
  // const displayName = 'My Company Name';
  const formattedParent = client.projectPath(projectId); // <== <== <==
  const company = {
    displayName: displayName,
  };
  const request = {
    parent: formattedParent,
    company: company,
  };
  client.createCompany(request)
    .then(responses => {
      const response = responses[0];
      console.log(`Created company: ${response.name}`);
    })
    .catch(err => {
      console.error(err);
    });
}
// [END job_search_create_company]
```

### » The actual result

```js
'use strict';

// [START job_search_create_company]
const talent = require('@google-cloud/talent').v4beta1;

/** Create Company */
function sampleCreateCompany(projectId, displayName) {
  const client = new talent.CompanyServiceClient();
  // const projectId = 'Your Google Cloud Project ID';
  // const displayName = 'My Company Name';
  const formattedParent = client.tenantPath(projectId, '[TENANT]');  // <== <== <==
  const company = {
    displayName: displayName,
  };
  const request = {
    parent: formattedParent,
    company: company,
  };
  client.createCompany(request)
    .then(responses => {
      const response = responses[0];
      console.log(`Created company: ${response.name}`);
    })
    .catch(err => {
      console.error(err);
    });
}
// [END job_search_create_company]
```

The generated code is:
 1. Choosing to use the `/projects/x/tenants/y` path
 2. Providing its OWN `[TENANT]` default which we didn't define anywhere!

Please do not use a binding path with path parts that have not been populated.

If we have to _specify_ additional bindings _explicitly_ THAT'S OK! That would be great :)

=> @googleapis/samplegen @hzyi-google 

----
> \* Search googleapis/googleapis for [additional_bindings][search] in .proto files. At the time of writing there are 230 instances of additional bindings used in [31 files][search]. For this reason, this should be prioritized no lower than `P1` and should be a supported feature of micro generators. Monolith support nice but not required.
> 
> As these bindings are "additional" there is always theoretically a way to call the API using the primary binding. HOWEVER please consider that the customer USE-CASES for using various bindings may be quite different. Ability to specify a particular binding is important.

[search]: https://github.com/googleapis/googleapis/search?l=Protocol+Buffer&q=additional_bindings

hzyi-google:
Fix #2719.

Try to find a resource name that matches exactly the initial values configured in gapic.yaml for Oneof resource names. If none is found, pick a random one and use boilerplate values.

Generated talent samples: [here](https://github.com/hzyi-google/sample-staging/tree/master/20190531).

### DIFF
--- a/src/main/java/com/google/api/codegen/config/ResourceNameOneofConfig.java
+++ b/src/main/java/com/google/api/codegen/config/ResourceNameOneofConfig.java
@@ -22,9 +22,9 @@ import com.google.api.tools.framework.model.ProtoFile;
 import com.google.api.tools.framework.model.SimpleLocation;
 import com.google.auto.value.AutoValue;
 import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.Iterables;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 
 /** ResourceNameOneofConfig represents the configuration for a oneof set of resource names. */
@@ -45,8 +45,12 @@ public abstract class ResourceNameOneofConfig implements ResourceNameConfig {
     return null;
   }
 
-  public Iterable<SingleResourceNameConfig> getSingleResourceNameConfigs() {
-    return Iterables.filter(getResourceNameConfigs(), SingleResourceNameConfig.class);
+  public List<SingleResourceNameConfig> getSingleResourceNameConfigs() {
+    return getResourceNameConfigs()
+        .stream()
+        .filter(c -> c.getResourceNameType() == ResourceNameType.SINGLE)
+        .map(c -> (SingleResourceNameConfig) c)
+        .collect(Collectors.toList());
   }
 
   @Nullable
--- a/src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/InitCodeTransformer.java
@@ -668,36 +668,7 @@ public class InitCodeTransformer {
     } else if (initValueConfig.hasFormattingConfig() && !item.getType().isRepeated()) {
       if (context.getFeatureConfig().enableStringFormatFunctions()
           || fieldConfig.getResourceNameConfig() == null) {
-        FormattedInitValueView.Builder formattedInitValue = FormattedInitValueView.newBuilder();
-        formattedInitValue.apiVariableName(
-            context.getNamer().getApiWrapperVariableName(context.getInterfaceConfig()));
-        formattedInitValue.apiWrapperName(
-            context.getNamer().getApiWrapperClassName(context.getInterfaceConfig()));
-        formattedInitValue.fullyQualifiedApiWrapperName(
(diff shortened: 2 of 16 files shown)


hzyi-google:
PTAL

hzyi-google:
@vchudnov-g @andreamlin Do you want to take a look as well?

Issue #2806, opened 2019-05-31
SampleGen: remove six from Python samples


beccasaurus:
### git revert #2272 – samplegen: convert binary type to str

We can remove six :)

> Not part of the current [Python Sample Authoring Guide](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/AUTHORING_GUIDE.md)
>
> Also, I asked. We can remove it.

@googleapis/samplegen

hzyi-google:
Fixed in ^^^^^^^ #2811

PR #2811, opened 2019-06-03
SampleGen: Hello unicode, bye bye six

### LINKED ISSUE #2807: SampleGen: print(u'all output strings {}'.format('as unicode strings'))
There was a recent bug filed against out samples _(hand-written but same syntax as our generated ones)_.

If you're using Python 3 and you do this...

```py
response = client.get_company(name)
print('Company name: {}'.format(response.name)) # <==  if response.name has   🔥
                                                #  🔥  unicode characters     🔥
                                                #  🔥  this explodes terribly 🔥
```
```
🐞 UnicodeEncodeError: 'ascii' codec can't encode character
    u'\xea' in position 1: ordinal not in range(128)
```

We could fix this for Python 2 users in our generated code samples by simply rendering the following instead:

```py
response = client.get_company(name)
print(u'Company name: {}'.format(response.name)) # <==  if response.name has   🌈
      ^                                          #  🦄  unicode characters     🌈
      ^                                          #  🦄  this works fine :)     🌈
```

**Note:**
 - Python 2 support is a non-goal for generated samples now, afaik
 - e.g. I would like to [remove](https://github.com/googleapis/gapic-generator/issues/2806) the `six` code we added _(it adds distracting cruft to the samples)_

But, presuming this is always OK and safe to add to every printed string _(I'm not a Python expert)_, it seems like a trivially small change that will help protect some folks from experiencing the lovely world of string encoding errors :)

**Let's print everything** like this: `print(u'whatever {}'.format(22))` _(if using .format)_

=> @crwilcox incase it's bad to always `print(u'the things'.format('ピザ'))`

/cc @dizcology @googleapis/samplegen 

#inspiration https://github.com/GoogleCloudPlatform/python-docs-samples/pull/2183

hzyi-google:
Fix #2807 #2806

### DIFF
--- a/src/main/resources/com/google/api/codegen/py/standalone_sample.snip
+++ b/src/main/resources/com/google/api/codegen/py/standalone_sample.snip
@@ -27,58 +27,46 @@
   @if sample.sampleImports.appImports.size
     {@importList(sample.sampleImports.appImports)}
   @end
-  @let convertTextCode = convertText(sample)
-    @if convertTextCode
-      import six
 
+  def {@sample.sampleFunctionName}({@formalArgs(sample.sampleInitCode.argDefaultParams)}):
+    @if not(sample.sampleFunctionDoc.isEmpty)
+      {@processDoc(sample.sampleFunctionDoc)}
+    @end
+    @# [START {@sample.regionTag}_core]
 
-    @else
+    client = {@apiMethod.apiModuleName}.{@apiMethod.apiClassName}()
 
+    @if sample.sampleInitCode.argDefaultLines
+      @join line : util.pretty(initCode(sample.sampleInitCode.argDefaultLines))
+        @# {@line}
+      @end
 
     @end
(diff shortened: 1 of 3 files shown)


hzyi-google:
=> @googleapis/samplegen

src/main/resources/com/google/api/codegen/py/standalone_sample.snip
```diff
@@ -187,7 +166,7 @@
 @private lroSampleCode(apiMethod, sample)
   operation = {@methodCallSampleCode(apiMethod, sample.sampleInitCode)}
 
-  print('Waiting for operation to complete...')
+  print(u'Waiting for operation to complete...')
```
vchudnov-g: Python3 has Unicode strings by default, it seems, so the `u` prefix is unnecessary.

https://stackoverflow.com/a/2464968
https://docs.python.org/3/howto/unicode.html

hzyi-google: @beccasaurus It seems Python 3 does not like the `'u'`?

PR #2813, opened 2019-06-04
node: fix failure when no options are passed

alexander-fenster:
Fixing the problem introduced in #2800, and even adding auto-generated tests to cover this.

### DIFF
--- a/src/main/resources/com/google/api/codegen/nodejs/main.snip
+++ b/src/main/resources/com/google/api/codegen/nodejs/main.snip
@@ -59,9 +59,10 @@
       opts = opts || {};
       this._descriptors = {};
 
-      const servicePath = opts.servicePath ||
-                          opts.apiEndpoint ||
-                          this.constructor.servicePath;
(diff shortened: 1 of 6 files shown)


src/main/resources/com/google/api/codegen/nodejs/main.snip
```diff
@@ -58,10 +58,12 @@
     constructor(opts) {
       this._descriptors = {};
 
-      const servicePath =
-        opts.servicePath ||
-        opts.apiEndpoint ||
-        this.constructor.servicePath;
+      let servicePath = this.constructor.servicePath;
```
bcoe: I'll sometimes start with:

```js
opts = opts || {}
```

to eliminate boiler plate, might be worth considering; definitely a nit, if you had a good reason for the if/else.

alexander-fenster: No good reason! Changing to what you suggested.

PR #2817, opened 2019-06-05
Add option to disable service config in Go & Python

### LINKED ISSUE #2778: all: disable gRPC service config lookup
We need to disable gRPC service config resolution in GAPICs until the gRPC service config is fully implemented/adopted. All languages need to have their equivalent of [DisableServiceConfig](https://godoc.org/google.golang.org/grpc#WithDisableServiceConfig) set on GAPICs. Without this, if a service owner adopts and defines a gRPC service config, the clients will immediately consume it, resulting in duplicated retry implementation with the gax-based retries.

- [x] Go
- [x] Python
- [x] Java
- [x] PHP
- [x] Ruby
- [x] NodeJS
- [x] C#

noahdietz:
Disable the gRPC ServiceConfig DNS resolution in Go & Python.

Fixes #2778


noahdietz:
Bump @lukesneeringer if you get a second, please review the changes to the python snippets to disable service config resolution in gRPC core

PR #2823, opened 2019-06-07
SampleGen: add SampleConfigProto

hzyi-google:
```
cp go/actools-samplegen-config-decoupling src/proto/google/api/codegen/samplegen/v1/sample_config_v1.proto
```

I realized now that calling this config `v1` might make more sense than `v1p2`, because it looks a bit strange to me if we have a `v1p2` but no `v1` or `v1p1`.

### DIFF
--- a/src/main/proto/com/google/api/codegen/samplegen/v1/sample_config_v1.proto
+++ b/src/main/proto/com/google/api/codegen/samplegen/v1/sample_config_v1.proto
@@ -7,8 +7,8 @@ option java_outer_classname = "ConfigProtoDesc";
 option java_package = "com.google.api.codegen.samplegen.v1";
 
 message SampleConfigProto {
-  string schema_version = 1;  // current version is 1.0
-  
+  string schema_version = 1;  // current version is "1.2"
+
   repeated SampleSpec samples = 2;
 }
 
@@ -30,8 +30,8 @@ message SampleSpec {
   // the value of `id`.
   string region_tag = 2;
 
-  // A description of this sample. Used to generate the documentation
-  // above the sample function.
+  // A description of this sample. This description will appear in the
+  // generated sample files as documentation for the sample functions.
   string description = 3;
 
   // The service name (as specified in the library proto that is being
@@ -58,7 +58,7 @@ message SampleSpec {
   // calling pattern name.
   //
   // Generators SHOULD name their language-specific calling patterns
-  // with a language snake-case prefix, such as "java_".
+  // with a snake-case language prefix, such as "java_".
   //
   // Generators MUST recognize the special expression "default" as
   // explicitly indicating that the default calling pattern must be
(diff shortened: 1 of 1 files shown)


src/main/proto/com/google/api/codegen/samplegen/v1/sample_config_v1.proto
```diff
@@ -0,0 +1,246 @@
+syntax = "proto3";
+
+package com.google.api.codegen.samplegen.v1;
```
vchudnov-g: I'm not completely sure about the `.v1` here, because we're already specifying the version in `schema_version`-- the repetition troubles me slightly. I guess it's alright if we have the major  version number as part of the package, and we then require `schema_version` to be consistent with this major version and to further specify the minor version.

src/main/proto/com/google/api/codegen/samplegen/v1/sample_config_v1.proto
```diff
@@ -0,0 +1,246 @@
...
+
+  // Information for constructing the request object passed to
+  // `rpc`. Each `request` item specifies a field within the RPC
+  // request object.
+  repeated RequestField request = 7;
+
+  // Definition of how to handle the RPC response, specified as an
+  // ordered list of statements.
+  repeated ResponseStatement response = 8;
+
```
vchudnov-g: delete blank line

src/main/proto/com/google/api/codegen/samplegen/v1/sample_config_v1.proto
```diff
@@ -0,0 +1,246 @@
...
+    string key = 5;
+
+    // The iteration variable for value.
+    //
+    // At least one of `key` and `value` must be specified if `map` is
+    // specified.
+    string value = 6;
+
+    // The contents of the loop.
+    repeated ResponseStatement body = 3;
```
vchudnov-g: I'd number these sequentially---not that it matters.

(I realize I neglected to do this when I wrote this proto for the design doc)

src/main/proto/com/google/api/codegen/samplegen/v1/sample_config_v1.proto
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  // A statement that writes to a local file.
+  message WriteFileStatement {
+    // The name of the file to write to.
+    //
+    // Similar to `print`, the elements of `file_name` consists of a
```
vchudnov-g: s/consists/consist/ (my bad)

src/main/proto/com/google/api/codegen/samplegen/v1/sample_config_v1.proto
hzyi-google: I must confess that I don't understand proto package names well enough to assert appending `v1` here is correct (hopefully will gain more knowledge during implementing this config), but my understanding is:
The `schema_version` field should always be there regardless of the version of sample configuration. When the generator loads the config yaml specified by users, it will check the `schema_version` field first, and then choose the correct version of config proto based on the value of this field to convert the yaml into -- and so that the generator needs to store all possible versions (i.e., supported and not deprecated) of schemas written in protos here, hence the package `v1` here. And I think it should be like:

```
com.google.api.codegen.samplegen.v1.sample_config_v1.proto 
 ^^^^^ load this when the yaml has schema_version field value set to 1.0.0,

com.google.api.codegen.samplegen.v1p2.sample_config_v1p2.proto
^^^^^ load this when the yaml has schema_version field value set to 1.2.0
```

Also note that the name of proto file is largely ignored, so package name is probably _the_ identifier here.

PR #2825, opened 2019-06-10
node: use Buffer as a type for protobuf bytes

alexander-fenster:
Documentation incorrectly shows `string` as a Node.js type for protobuf `bytes`. Changing that to `Buffer`. It won't be a breaking change for the generated clients since (a) `string` never worked in this case, and (b) it's just the documentation and samples, actual code does not change.


alexander-fenster:
@bdhess Please take a look. Thank you!

bdhess:
lgtm; thanks for knocking this out!

alexander-fenster:
@bcoe Nope, and (more importantly) there is no code path, it's only in the comments and generated samples (which is a code but is never executed). The customers just found the samples that used `string` where `Buffer` is needed, and those samples did not work.

PR #2827, opened 2019-06-10
SampleGen: generate multi-line descriptions in sample metadata

hzyi-google:
To not break when a a sample has a long long description, like [speech](https://github.com/googleapis/googleapis/blob/4145cd9e644de069b17a17cefd3364d351214b18/google/cloud/speech/v1p1beta1/samples/speech_transcribe_diarization_beta.yaml#L5)


hzyi-google:
cc/ @nnegrey

PR #2828, opened 2019-06-11
SampleGen: parse the new sample config

hzyi-google:
* Accept a command line flag `sample_yamls`, which passes in a list of sample config files that follows the schema of `SampleConfigProto`.

* Parse the config files into a `SampleConfig`, and store the `SampleConfig` in `GapicProductConfig`.

* Do not generate anything based on the `SampleConfig` at this moment.

* Some minor changes to `SampleConfigProto`.


src/main/proto/com/google/api/codegen/samplegen/v1/sample_config_v1.proto
```diff
@@ -24,22 +24,25 @@ message SampleSpec {
   // value in the emitted samples.
   string id = 1;
 
+  // The title of the sample.
+  string title = 2;
```
hzyi-google: We need `title` as well.

cc/ @vchudnov-g

src/main/proto/com/google/api/codegen/samplegen/v1/sample_config_v1.proto
```diff
@@ -9,10 +9,10 @@ option java_package = "com.google.api.codegen.samplegen.v1";
 message SampleConfigProto {
   string schema_version = 1;  // current version is "1.2"
 
-  repeated SampleSpec samples = 2;
+  repeated SampleSpecProto samples = 2;
```
hzyi-google: Renamed a bunch of top-level messages to `XxxProto` to be consistent with gapic_config.yaml. This also makes generation code more readable.

hzyi-google:
PTAL

src/main/proto/com/google/api/codegen/samplegen/v1/sample_config_v1.proto
vchudnov-g: Yeah, I was wondering about that....

src/main/proto/com/google/api/codegen/samplegen/v1/sample_config_v1.proto
vchudnov-g: Since we're getting rid of `gapic_config.yaml`, I don't see consistency with that as being important.

Readability is a more important concern. I had suggested not having the `Proto` suffix because it's redundant in the `.proto` file (all messages are protocol buffers there), but I can see that if the name transfers over to the generated code it would make for more readable code on this side.....

src/main/java/com/google/api/codegen/config/SampleConfig.java
```diff
@@ -50,14 +99,110 @@ public static Builder newBuilder() {
...
+    public abstract Builder callingPattern(String val);
+
     public abstract Builder valueSet(SampleValueSet val);
 
     public abstract Builder type(SampleSpec.SampleType val);
 
     public abstract SampleConfig build();
   }
+
+  public static ImmutableList<SampleConfig> create(
```
vchudnov-g: I would rename this to `createList` or `createMultiple` to avoid confusion with the overload. They're related but do different things, so they should have different names,

src/main/java/com/google/api/codegen/config/SampleConfig.java
```diff
@@ -50,14 +99,110 @@ public static Builder newBuilder() {
...
+            ImmutableList.<RequestFieldProto>builder().addAll(sampleSpec.getRequestList()).build())
+        .responseConfigs(
+            ImmutableList.<ResponseStatementProto>builder()
+                .addAll(sampleSpec.getResponseList())
+                .build())
+        .regionTag(sampleSpec.getRegionTag())
+        .callingPattern(
+            sampleSpec.getCallingPatternsCount() > 0
+                ? sampleSpec.getCallingPatternsList().get(0)
+                : "")
```
vchudnov-g: How about specifying `"default"` here, and everything downstream knows to use the explicitly named patterns? (Ie the inference happens here). Or even better, in the method above, as per my comment there.

src/main/java/com/google/api/codegen/gapic/GapicGeneratorApp.java
```diff
@@ -137,6 +149,22 @@ protected void process() throws Exception {
...
+    SampleConfigProto sampleConfigProto = null;
+    List<String> sampleConfigFileNames = options.get(SAMPLE_CONFIG_FILES);
+    if (sampleConfigFileNames.size() > 0) {
+      ConfigSource configSource =
+          loadConfigFromFiles(
+              sampleConfigFileNames,
+              SampleConfigProto.getDescriptor().getFullName(),
+              SampleConfigProto.getDefaultInstance());
+
+      // TODO(hzyi): Verify this works for repeated fields as well
```
vchudnov-g: Clarify?

src/main/java/com/google/api/codegen/gapic/GapicGeneratorApp.java
```diff
@@ -137,6 +149,22 @@ protected void process() throws Exception {
...
+      // TODO(hzyi): Allow users to put arbitrary top-level directives not
```
vchudnov-g: Just a thought: if we agree gapic-config should have the same semantics (I think it should, but we should decide as a team) then this logic can be put in a function called from `loadConfigFromFiles`, which will ignore any top level YAML field that is not a top-level proto message field

src/main/proto/com/google/api/codegen/samplegen/v1/sample_config_v1.proto
vchudnov-g: (Remind me: is the title use anywhere other than metadata?)

src/main/proto/com/google/api/codegen/samplegen/v1/sample_config_v1.proto
```diff
@@ -71,19 +74,24 @@ message SampleSpec {
   // generators should also generate the "default" calling-pattern
   // sample for this RPC if it was not included by any of the
   // previous matches.
-  repeated string calling_patterns = 6;
+  //
+  // Note: the language prefix and `default` calling patterns are not
```
vchudnov-g: Change this note to be a TODO to support them

src/main/java/com/google/api/codegen/config/SampleConfig.java
vchudnov-g: Could you add a test for this method?

src/main/proto/com/google/api/codegen/samplegen/v1/sample_config_v1.proto
hzyi-google: No. That's the only place.

src/main/java/com/google/api/codegen/config/SampleConfig.java
```diff
@@ -15,17 +15,66 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+import javax.annotation.Nullable;
 
 /** SampleConfig represents configurations of a sample. */
+// Note: This class was used as an intermediate data structure to hold some information
+// about samples. Now that we are flattening out sample configuration, this class will
+// mostly likely not be needed any more. However, to conform to the model/config/context
+// style used by gapic-generator, we will reuse this class to hold all the information
```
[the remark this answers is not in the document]
hzyi-google: Indefinitely.

src/main/java/com/google/api/codegen/config/SampleConfig.java
hzyi-google: IIRC We have always been using empty string to represent a default calling form. Is it OK to support `default` later?

src/main/java/com/google/api/codegen/gapic/GapicGeneratorApp.java
hzyi-google: YamlReader will merge multiple config yamls as long as they all have the same `type` and `version` specified. but I am not sure how it handles repeated fields.

src/main/java/com/google/api/codegen/gapic/GapicGeneratorApp.java
hzyi-google: We can I think.

src/main/proto/com/google/api/codegen/samplegen/v1/sample_config_v1.proto
hzyi-google: It wouldn't be hard to support `default` I think. But why is it necessary? Wouldn't leaving out this field be an appropriate way to tell the generator to use the `default` one?

I mean, if people do `int i;`, `i == 0` is a proper default value. Do people want to do `int i = int.default` at all?

src/main/java/com/google/api/codegen/config/SampleConfig.java
hzyi-google: Changed this to `createSampleConfigs`, and other one to `createSampleConfig`.

hzyi-google: It's kind of hard to add a unit test for this method, since we have to build a lot of fake `interfaceConfigs` and `methodConfigs`. Once we have the baselines in, as long as all the existing samples are still generated I think we should be good.

src/main/java/com/google/api/codegen/config/SampleConfig.java
```diff
@@ -50,14 +99,110 @@ public static Builder newBuilder() {
...
+
+    // Next, flat out calling forms
+    List<SampleSpecProto> flattenedSampleSpecs = new ArrayList<>();
+    for (SampleSpecProto spec : sampleSpecs) {
+      for (String pattern : spec.getCallingPatternsList()) {
+        sampleSpecs.add(spec.toBuilder().addCallingPatterns(pattern).build());
+      }
+    }
+
+    // We cannot auto-generate IDs for samples without user provided ones at this step.
```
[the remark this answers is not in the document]
hzyi-google: Comments updated. I agree that's probably the best way to disambiguate samples. However I don't want to commit to it before we implement it. When we implement (rewrite) this logic I'll make sure to document how sample ids are disambiguated.

hzyi-google:
PTAL

src/main/proto/com/google/api/codegen/samplegen/v1/sample_config_v1.proto
vchudnov-g: Yes, leaving out the field is an appropriate way. But we also need to support people wanting the default _as well as_ something else. eg `[default, longrunning_start_then_cancel]`

src/main/java/com/google/api/codegen/config/SampleConfig.java
vchudnov-g: oh, Java.....

vchudnov-g: I dislike architectures that discourage us from writing unit tests by making it onerous. I think you're right about coverage for this function via the baselines, though, so we can make do with that.

src/main/java/com/google/api/codegen/config/SampleConfig.java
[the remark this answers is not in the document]
vchudnov-g: sg

src/main/java/com/google/api/codegen/config/SampleConfig.java
vchudnov-g: We can support it later.

PR #2834, opened 2019-06-14
node: simplify inner call initialization

alexander-fenster:
This PR makes initialization of `this._innerApiCalls` just a little bit more readable by replacing `arguments` with a rest parameter syntax (`...args`). No changes in behavior intended.

Note: more changes to the generation code will follow, this one is just the first and the most minor one.

### DIFF
--- a/src/main/resources/com/google/api/codegen/nodejs/main.snip
+++ b/src/main/resources/com/google/api/codegen/nodejs/main.snip
@@ -240,18 +240,16 @@
           @end
         ];
         for (const methodName of {@stub.stubMethodsArrayName}) {
+          const innerCallPromise = {@stub.name}.then(
+            stub => (...args) => {
+              return stub[methodName].apply(stub, args);
+            },
+            err => () => {
+              throw err;
+            }
+          );
           this._innerApiCalls[methodName] = gax.createApiCall(
-            {@stub.name}.then(
-              stub =>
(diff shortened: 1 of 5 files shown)


JustinBeckwith:
👋 I have a petty ask.  Every time we make one of these changes to the generator, it comes with a corresponding 42 PRs by autosynth.  If you expect to line up a few of these ... can we do the work in a branch, and then bring it over to master when ready?  If you're looking to make these small and incremental to reduce risk to the generated libs, that's cool too.  Just looking to optimize for laziness 🙃

alexander-fenster:
Hi @JustinBeckwith! We're going have more of small changes during the upcoming week. I was actually going to be cautious and merge them one by one, but I totally get the idea of minimizing the number of synthtool PRs to merge.

This particular one does not block any of our work (it's just a cosmetic change) so I'm totally fine if it waits for some bigger change to come. For bigger changes such as the upcoming change of the logic of protobuf loading, I'd still ask you to go through all the 42 PRs :)

alexander-fenster:
@bcoe Yes, I verified that it works locally, and the CI passes. This particular piece of generated code is one of the most complicated code snippets in our libraries, so I'm just trying to make it look simpler :)

alexander-fenster:
This is going to be handled in #2888.

PR #2845, opened 2019-06-18
Add GAPIC Bazel Extensions for Node.js

vam-google:
The extensions provide minimum level of functionality. 

They generate the output identical to what artman provides (i.e. have feature parity with artman).

It does not try to build and run the generated tests from withing bazel (like it does for Java and Go). This functionality may be added later.

### DIFF
--- a/rules_gapic/README.md
+++ b/rules_gapic/README.md
@@ -50,7 +50,14 @@ The rules will call `gapic-generator` and do all the necessary pre- and post- ge
 
 4. **`php_gapic_library`** - PHP **macro**, which calls `php_gapic_srcjar` to generate and postprocess gapic library. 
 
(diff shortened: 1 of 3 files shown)


alexander-fenster:
I understand nothing here :)  so it's LGTM but I'll leave it to @eoogbe to approve on behalf of me.

vam-google:
@eoogbe Done,
@alexander-fenster I've added the readme documentation about it, I hope it can make it clearer. But in general, this whole PR is just the following two things:
- `nodejs_gapic.bzl` - call gapic generator to generate stuff (collecting necessary input parameters). Most of the body of the file is very simplistic shell script (inside the .bzl file). It  just calls `cp`, `zip/unzip`, `mv` and other basic commands. That is pretty much it.
- `nodejs_gapic_pkg.bzl` - does what artman does on top of gapic output (actually for Node.js it does very little, compared to other langues, which is very nice). Specifically puts the gapic output in a specific folder structure and copies the source .proto files there as well. 

The bazel rules code itself has Python syntax (I believe it was actually forked from Python 2.6).

PR #2846, opened 2019-06-18
[feat] SampleGen: comments on request object fields in C#

hzyi-google:


src/test/java/com/google/api/codegen/protoannotations/testdata/csharp_library.baseline
```diff
@@ -2081,13 +2083,19 @@ namespace Google.Example.Library.V1.Samples
             {
                 Shelf = new Shelf
                 {
+                    // Comment on a resource name
                     Name = new ShelfName("math").ToString(),
+                    // A super long comment
```
vchudnov-g: Not sure what the C# convention is, but my own personal preference across languages is to have a blank line between the preceding variable and the comment for the next variable—I think it's more readable:

```
// Comment for A
A="alex"

// Comment for B
B="babs"
```

Not a blocker for this PR, we can add this later. But worth thinking about. Opinions, @beccasaurus ?

PR #2851, opened 2019-06-20
go: URL encode x-goog-request-params

noahdietz:


src/main/java/com/google/api/codegen/transformer/go/GoImportSectionTransformer.java
```diff
@@ -59,7 +59,7 @@ public ImportSectionView generateImportSection(Map<String, TypeAlias> typeImport
   private boolean isStandardImport(String importPath) {
     // TODO(pongad): Some packages in standard library have slashes,
     // we might have to special case them.
-    if (importPath.equals("net/http")) {
+    if (importPath.equals("net/http") || importPath.equals("net/url")) {
```
pongad: I think I found a better way to detect this. We can probably look for a "." in the path, so that we'll catch "github.com/foo/bar", but not stdlib.

noahdietz: Nice! Done.

PR #2853, opened 2019-06-24
Python: Add client options.

busunkim96:
Tests will fail until there is a new release of api-core.


busunkim96:
@lukesneeringer Can I have some help with the remaining failure on gapic-generator-bazel-test?

PR #2856, opened 2019-06-26
Make it possible to pass metadata to the operations client

samphippen:
For googleads we inject `developer-token` and `login-customer-id` to all
Reques)ts. Previously to this patch, we had used a [monkey patch](https://git.io/fjrda)
to inject them in to the right place. However, ths patch was built
against an old version of both gax and the generated code and no longer
works for the newly generated google ads v2 protos.

At the core of the requirement is that both the main API client and,
where generated, the long running operations client, have injecatable
metadata so that we can provide them for LROs without patches.

This patch passes the metadata value passed in to the constructed client
along to the Long Running Operations client, which was not happening
before.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -178,7 +178,9 @@ anchor_run_ruby: &anchor_run_ruby
        name: Run Showcase Integration Tests
        command: |
          cd workspace/gapic-generator/showcase/ruby
-         bundle install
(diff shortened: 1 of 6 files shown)


samphippen:
@blowmage does this look right to you?

alexander-fenster:
The change to `snip` files normally require changes to baselines as well. I can update the baselines for you.

PR #2858, opened 2019-06-27
go: remove encoding on header request param key

noahdietz:
Removes URL encoding of the header request param keys from the snippets and only encodes the value instead.

Needing to encode the key will be decided later (if we actually need it at all). The chance a field name is non-ASCII is very very low.

Similar PR in https://github.com/googleapis/gapic-generator-go/pull/145


noahdietz:
This actually fixes bad behavior introduced in a previous CL detailed by https://github.com/googleapis/gapic-generator-go/issues/144

PR #2859, opened 2019-06-27
add feature to provide retry config via gRPC ServiceConfig

noahdietz:
This changes adds a new command line flag for GAPIC generation, `--grpc-service-config <filepath>`, that allows the caller to supply a [gRPC ServiceConfig](https://github.com/grpc/grpc-proto/blob/5ce8e3e598b805a1e0372062913f24b0715fdefc/grpc/service_config/service_config.proto) in proto-JSON format as the retry configuration for GAPICs generated with GAPIC config v2.

The logic for injecting the retry policies is 100% covered via unit tests.

Todo:
- [x] figure out how to add `io.grpc.serviceconfig` [proto](https://github.com/grpc/grpc-proto/blob/5ce8e3e598b805a1e0372062913f24b0715fdefc/grpc/service_config/service_config.proto) OR Java classes as a dependency
- [x] add baselines for languages other than Go
- [x] fix bazel build


noahdietz:
Need to figure out #2866 first

noahdietz:
@vam-google this is now ready for review! PTAL

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -485,6 +503,163 @@ private static GapicProductConfig createDummyInstance(
...
+   * config
+   */
+  static void findAndSetRetry(
+      InterfaceConfigProto.Builder ib,
+      boolean overwrite,
+      String method,
+      String rc,
+      String rp,
+      long timeout) {
+    int i = findMethod(ib, method);
```
alexander-fenster: Please no single-letter variable names :)

src/main/java/com/google/api/codegen/gapic/GapicGeneratorApp.java
```diff
@@ -149,6 +159,17 @@ protected void process() throws Exception {
       }
     }
 
+    // if gRPC Service Config is given, consume it
+    String gRPCServiceConfigPath = options.get(GRPC_SERVICE_CONFIG);
+    ServiceConfig gRPCServiceConfig = null;
+    if (!Strings.isNullOrEmpty(gRPCServiceConfigPath)) {
+      ServiceConfig.Builder b = ServiceConfig.newBuilder();
```
alexander-fenster: Same here: let's call it `builder`

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -122,7 +129,7 @@ public GapicProductConfig withPackageName(String packageName) {
   @Nullable
   public static GapicProductConfig create(
       Model model, ConfigProto configProto, TargetLanguage language) {
-    return create(model, configProto, null, null, null, language);
+    return create(model, configProto, null, null, null, language, null);
```
alexander-fenster: Disclaimer: I have almost zero Java experience. But in general, `create` function with 7 parameters, four of them being `null`, require either a builder, or at least showing `/* parameterName: */` in comments.

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -475,7 +487,13 @@ private static GapicProductConfig createDummyInstance(
           DiagCollector diagCollector,
           List<InterfaceConfigProto> interfaceConfigProtosList,
           ImmutableMap<String, Interface> protoInterfaces,
-          TargetLanguage language) {
+          TargetLanguage language,
+          ServiceConfig grpcServiceConfig) {
+    if (grpcServiceConfig != null) {
```
alexander-fenster: Indentation seems broken here.

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -485,6 +503,163 @@ private static GapicProductConfig createDummyInstance(
...
+    if (addRetryParams) {
+      ib.addRetryParamsDef(rpb);
+    }
+  }
+
+  /**
+   * finds the named MethodConfigProto, creates one if absent in the GAPIC interface, and sets retry
+   * config
+   */
+  static void findAndSetRetry(
```
alexander-fenster: private?

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -485,6 +503,163 @@ private static GapicProductConfig createDummyInstance(
...
+    }
+
+    mcp.setRetryParamsName(rp);
+    mcp.setRetryCodesName(rc);
+    mcp.setTimeoutMillis(timeout);
+    ib.setMethods(i, mcp);
+  }
+
+  /** returns the index of the named MethodConfigProto in the GAPIC interface and -1 if not found */
+  static int findMethod(InterfaceConfigProto.Builder service, String method) {
```
alexander-fenster: private?

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
noahdietz: According to the [java docs](https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html), _no modifier_ is not very permissive. I wanted these methods to be visible for testing purposes. Making them private makes them harder to test

jbolinger: You could add Guava's `@VisibleForTesting` so folks know that you did this intentionally.

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -485,6 +503,163 @@ private static GapicProductConfig createDummyInstance(
             .collect(Collectors.toMap(InterfaceConfigProto::getName, Function.identity())));
   }
 
+  static List<InterfaceConfigProto> injectRetryPolicyConfig(
```
hzyi-google: Hmmmm, injecting retry configs back to gapic config seems a bit counter-intuitive. This means we can't delete those fields from gapic_config proto, even though after this change, we do not anticipate users to continue to configure the retry fields through gapic config (if that is true).

Just thinking out loud, is it possible to inject retry policy configs directly to `MethodConfig`?

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
noahdietz: If you are referring to the function declaration, gFormat makes it this way. Not a lot I can do about it.

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
noahdietz: I will add the annotations! (implementation detail, it will come from `com.google.common.annotations.VisibleForTesting`)

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
alexander-fenster: Sorry - my bad!

src/main/proto/com/google/api/codegen/grpc/service_config.proto
```diff
@@ -0,0 +1,308 @@
...
+  // MethodConfig entries:
+  //
+  // method_config { name { service: "MyService" } ... }
+  // method_config { name { service: "MyService" method: "Foo" } ... }
+  //
+  // For a request for MyService/Foo, we will use the second entry, because it
+  // exactly matches the service and method name.
+  // For a request for MyService/Bar, we will use the first entry, because it
+  // provides the default for all methods of MyService.
+  message Name {
```
hzyi-google: Is it possible to pull this out to the top level? (to be consistent with gapic config definitions, and to make import statements easier)

Also is it possible to change the name of this message `Name`? Another class called `Name` is defined in this repo and is fairly commonly used. Changing this to some other name would save the potential trouble of having to refer to one of them with fully qualified name when used in the same file.

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
noahdietz: I agree this is not great. But IDEs show the variable names next to `null` values for this exact reason. See attached screenshot. I'm less inclined to add more comments for the internal method because of this, but happy to do so if you think its generally important.

Screenshot from 2019-07-16 14-33-52

src/test/java/com/google/api/codegen/CodegenTestUtil.java
```diff
@@ -72,4 +78,27 @@ public static ConfigProto readConfig(
...
+      DiagCollector diagCollector, TestDataLocator testDataLocator, String serviceConfigFileName) {
+    URL serviceConfigUrl = testDataLocator.findTestData(serviceConfigFileName);
+
+    String serviceConfigPath = Objects.requireNonNull(serviceConfigUrl).getPath();
+
+    ServiceConfig.Builder b = ServiceConfig.newBuilder();
+
+    try {
+      FileReader file = new FileReader(serviceConfigPath);
+      JsonFormat.parser().merge(file, b);
```
hzyi-google: Is it possible to use `MuitiYamlReader`, similar to gapic config, as well? https://github.com/googleapis/gapic-generator/pull/2859/files#diff-d39112fadd134cc4a174510601615043L65

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
noahdietz: Done!

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
noahdietz: this is "package-private" and I will add the VisibileForTesting annotation so that folks know this is on purpose

src/main/java/com/google/api/codegen/gapic/GapicGeneratorApp.java
noahdietz: Done!

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
noahdietz: Done!

src/test/java/com/google/api/codegen/testsrc/common/library_grpc_service_config.json
```diff
@@ -0,0 +1,52 @@
...
+    {
+      "name": [
+        { "service": "google.example.library.v1.MyProto", "method": "MyMethod" }
+      ],
+
+      "waitForReady": true,
+
+      "timeout": "60s"
+    }]
+}
```
hzyi-google: newline at EOF please :)

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
alexander-fenster: Makes us `vim` folks cry :)  But I'm generally fine, I don't have any authority and/or experience with Java here.

hzyi-google:
It's hard to tell what's happening to the baseline. IIUC nothing should change since this just involves reading retry configs from a different source. If not, it might be better to have a different PR that just makes a copy of the baseline files.

(also sorry for the huge merge conflicts, they seem to be coming from my commit :/)

src/main/proto/com/google/api/codegen/grpc/service_config.proto
noahdietz: Your requests are very reasonable and I would normally implement them, but I don't want to change this proto at all because it is a direct copy of another proto (as mentioned in this file's header comment). If we change it, it diverts from the original (which shouldnt change very much at all, but still).

hzyi-google: That makes sense.

noahdietz:
> IIUC nothing should change since this just involves reading retry configs from a different source.

You are correct, nothing changes with the baselines except for the values (and names) of the generated retry code.

> it might be better to have a different PR that just makes a copy of the baseline files.

Because your understanding is correct, is this a no-op? The `_with_grpc_service_config` baselines cover the difference in using this config source.

src/main/proto/com/google/api/codegen/grpc/service_config.proto
```diff
@@ -0,0 +1,308 @@
...
+  RetryThrottlingPolicy retry_throttling = 3;
+
+  message HealthCheckConfig {
+    // Service name to use in the health-checking request.
+    .google.protobuf.StringValue service_name = 1;
+  }
+  HealthCheckConfig health_check_config = 5;
+
+  // next available tag: 6
+}
```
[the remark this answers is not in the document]
noahdietz: Done.

src/test/java/com/google/api/codegen/CodegenTestUtil.java
```diff
@@ -72,4 +78,27 @@ public static ConfigProto readConfig(
...
     return (ConfigProto) configSource.getConfig();
   }
+
+  public static ServiceConfig readGRPCServiceConfig(
+      DiagCollector diagCollector, TestDataLocator testDataLocator, String serviceConfigFileName) {
+    URL serviceConfigUrl = testDataLocator.findTestData(serviceConfigFileName);
+
+    String serviceConfigPath = Objects.requireNonNull(serviceConfigUrl).getPath();
+
+    ServiceConfig.Builder b = ServiceConfig.newBuilder();
```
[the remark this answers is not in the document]
noahdietz: Done.

src/test/java/com/google/api/codegen/config/GapicConfigProducerTest.java
```diff
@@ -66,12 +73,141 @@ public void missingInterface() {
...
     Diag expectedError =
         Diag.error(
             SimpleLocation.TOPLEVEL,
             "interface not found: google.example.myproto.v1.MyUnknownProto. Interfaces: [google.example.myproto.v1.MyProto]");
     assertThat(model.getDiagReporter().getDiagCollector().hasErrors()).isTrue();
     assertThat(model.getDiagReporter().getDiagCollector().getDiags()).contains(expectedError);
   }
+
+  @Test
+  public void findMethod() {
```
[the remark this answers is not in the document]
noahdietz: Done.

src/test/java/com/google/api/codegen/config/GapicConfigProducerTest.java
```diff
@@ -66,12 +73,141 @@ public void missingInterface() {
...
+    MethodConfigProto.Builder mcb = MethodConfigProto.newBuilder();
```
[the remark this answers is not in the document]
noahdietz: Done.

src/test/java/com/google/api/codegen/testsrc/common/library_grpc_service_config.json
noahdietz: Done.

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
noahdietz: > This means we can't delete those fields from gapic_config proto

Ah I see your concern here.

>  is it possible to inject retry policy configs directly to MethodConfig?

Mmm I'm not following (my own unfamiliarity with the code base), where does the MethodConfig come into play during execution? Also, where is it defined? (If you have time, links would be helpful :smiley: )

The reason I went with essentially "spoofing" the GAPIC config is because there are N (unknown to me, might be known to you) downstream things that languages do with the GAPIC config after receiving it for generation. Instead of wading through that, I figured we could inject it with the new values and languages could do with them as they see fit.

src/test/java/com/google/api/codegen/CodegenTestUtil.java
noahdietz: Good idea. I would prefer to keep things in a JSON format however, because this is the format in which other systems that might use the gRPC ServiceConfig will expect it to be in. Thus, we aren't expecting anything different from other systems consuming the same config. (Disclaimer: these "other systems" might not exist yet, but the "how" they should function is well-defined by gRPC team).

I'm happy to look at making this more "formally" supported, but if possible, keeping it in JSON form makes life easier. 

WDYT?

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
hzyi-google: _(I figured that was the reason and I totally agree gapic-generator is super complicated)_

Correction: instead of injecting directly to [`MethodConfig`](https://github.com/googleapis/gapic-generator/blob/46f49e621c5aa04f03afcbf4abc7391648ab6ab7/src/main/java/com/google/api/codegen/config/MethodConfig.java#L49-L51), the right way to do this might be injecting retry configs to [`GapicInterfaceConfig`](https://github.com/googleapis/gapic-generator/blob/46f49e621c5aa04f03afcbf4abc7391648ab6ab7/src/main/java/com/google/api/codegen/config/GapicInterfaceConfig.java#L73-L77). And GapicInterfaceConfig would most likely pass the along to `MethodConfig`.

I think gapic-generator has this intermediate layer ([usually ***Config](https://github.com/googleapis/gapic-generator/tree/master/src/main/java/com/google/api/codegen/config)) that holds somewhat preprocessed information from gapic config yamls. (I don't exactly know why, but I guess the design was there for a bunch of reasons like being less proto-ish and more Java idiomatic, to support discogapics, to make config schema change easier etc). [`GapicProductConfig`](https://github.com/googleapis/gapic-generator/blob/master/src/main/java/com/google/api/codegen/config/GapicProductConfig.java) is the top-level Config object, and is the starting point for gapic generations.

To be more specific, retry codes are re-written into [`RetryCodesConfig`](https://github.com/googleapis/gapic-generator/blob/master/src/main/java/com/google/api/codegen/config/RetryCodesConfig.java), but 
retry setting definitions are not re-written into an intermediate object (we are using [protos](https://github.com/googleapis/gapic-generator/blob/46f49e621c5aa04f03afcbf4abc7391648ab6ab7/src/main/java/com/google/api/codegen/config/GapicInterfaceConfig.java#L77) directly, which starts to be annoying), and they are created and stored in `GapicInterfaceConfig` [here](https://github.com/googleapis/gapic-generator/blob/46f49e621c5aa04f03afcbf4abc7391648ab6ab7/src/main/java/com/google/api/codegen/config/GapicInterfaceConfig.java#L120-L152).

So instead of injecting into the config proto objects directly, probably alter the creation of `retryCodeConfig` and `retrySettingDefinition` mentioned in the above link and create those two from grpc service config protos?

Note, Ideally I think we should also change the type of [`retrySettingsDefinition`](https://github.com/googleapis/gapic-generator/blob/46f49e621c5aa04f03afcbf4abc7391648ab6ab7/src/main/java/com/google/api/codegen/config/GapicInterfaceConfig.java#L77) to some intermediate config objects and not depend on the protos (so that we can completely delete those fields from gapic config), but that would cause huge refactoring and probably should be done separately.


FYI the pipeline works like:
```
yamls -> parsed into protos -> re-written into the ***Config objects -> put into contexts
-> passed to transformers -> ... -> rendering
```

src/test/java/com/google/api/codegen/CodegenTestUtil.java
hzyi-google: That makes sense. 

BTW will service configs be checked in into googleapis/googleapis? Also does artman need to be updated to pick up the service configs and pass that to gapic-generator?

noahdietz: > Also does artman need to be updated to pick up the service configs and pass that to gapic-generator

most likely, yes! I need to explore this as i've never changed artman before. Any help/guidance is welcome!

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
noahdietz: Ok this is great! Thank you for the guidance. Everything looks like it'll still be in a similar place, I've just gotta refactor the logic a bit. The goal of removing the retry fields from the proto entirely is a great one, good call.

noahdietz: Ehh it might be more work than I first anticipated. I'll lyk when to take a look, might be a few commits coming in.

noahdietz:
@hzyi-google I've refactored it per your suggestion! PTAL

The baselines were actually improved because doing it this way prevented the generation of unused retry config definitions.

Furthermore, this change still uses the GAPIC defined Retry*Proto Java classes, but does not explicitly set the GAPIC MethodConfigProto `retry_codes_name` or `retry_params_name` fields anymore. So those could eventually be removed from the GAPIC MethodConfigProto I believe.

src/main/java/com/google/api/codegen/config/GapicInterfaceConfig.java
```diff
@@ -111,20 +111,33 @@ static GapicInterfaceConfig createInterfaceConfig(
...
 
-    RetryCodesConfig retryCodesConfig =
-        RetryCodesConfig.create(
-            interfaceConfigProto,
-            new ArrayList<>(interfaceInput.getMethodsToGenerate().keySet()),
-            protoParser);
+    RetryCodesConfig retryCodesConfig;
+    ImmutableMap<String, RetryParamsDefinitionProto> retrySettingsDefinition;
+    if (grpcGapicRetryMapping != null) {
+      // use the gRPC ServiceCongif retry mapping as the source
```
hzyi-google: typo: `ServiceConfig`

src/main/java/com/google/api/codegen/config/GrpcGapicRetryMapping.java
```diff
@@ -0,0 +1,144 @@
...
+
+  public abstract ImmutableMap<String, RetryParamsDefinitionProto> paramsDefMap();
+
+  public static Builder builder() {
+    return new AutoValue_GrpcGapicRetryMapping.Builder();
+  }
+
+  /** converts a float to a double rounding to the nearest hundredth */
+  private static double floatToDouble(float f) {
+    return Math.round(f * 100) / 100.0;
```
hzyi-google: Converting `float` to `double` gets extra precision. Why is rounding needed?

src/main/java/com/google/api/codegen/config/GrpcGapicRetryMapping.java
```diff
@@ -0,0 +1,144 @@
...
+import com.google.protobuf.util.Durations;
+import java.util.HashMap;
+import java.util.Map;
+
+@AutoValue
+public abstract class GrpcGapicRetryMapping {
+
+  public static GrpcGapicRetryMapping create(
+      ServiceConfig config, ImmutableMap<String, Interface> protoInterfaces) {
+    Map<String, String> methodCodesMap = new HashMap<>();
```
hzyi-google: nit: you can make `methodCodesMap` type of `ImmutableMap.Builder<String, String>`, and when finished with adding elements to it just call `methodCodesMap.build()` and you have the `ImmutableMap`. This way you don't have to make a copy.

src/main/java/com/google/api/codegen/config/GrpcGapicRetryMapping.java
```diff
@@ -0,0 +1,144 @@
...
+    Map<String, String> methodParamsMap = new HashMap<>();
+    Map<String, RetryCodesDefinitionProto> codesDefMap = new HashMap<>();
+    Map<String, RetryParamsDefinitionProto> paramsDefMap = new HashMap<>();
+
+    // build retry-to-interface mapping from gRPC ServiceConfig
+    int retryNdx = 1;
+    for (com.google.api.codegen.grpc.MethodConfig mc : config.getMethodConfigList()) {
```
hzyi-google: nit (very minor): Java generally prefers "real words" rather than abbreviations for variable names. e.g., `config` here would be preferred over `mc`. If you have time, changing some of these cases in this method would be really great. They are still correct as they are now for sure.

src/main/java/com/google/api/codegen/config/RetryCodesConfig.java
```diff
@@ -92,6 +92,41 @@ public static RetryCodesConfig create(
...
+    Map<String, ImmutableList<String>> codesDefMap = new HashMap<>();
+    Map<String, String> methodNamesMap = new HashMap<>();
+    ImmutableMap<String, String> methodCodesMap = retryMapping.methodCodesMap();
+    ImmutableMap<String, RetryCodesDefinitionProto> codesMap = retryMapping.codesDefMap();
+
+    methodCodesMap
+        .keySet()
+        .stream()
+        .filter(name -> name.startsWith(service))
+        .forEach(
```
hzyi-google: nit: Generally if there are multiple lines in the `forEach` block, for loops are preferred. The code works for sure.

src/main/java/com/google/api/codegen/config/GrpcGapicRetryMapping.java
noahdietz: I realize this is weird, but without rounding it, the value ends up being printed as `1.30000000` or something of the like. I realize its not that big of a deal, but it looked weird. It's just an "aesthetic" issue. Happy to take the rounding out if you think its unnecessary :)

src/main/java/com/google/api/codegen/config/GrpcGapicRetryMapping.java
noahdietz: I started with that, but ran into the issue of duplicate entries because the `ImmutableMap.Builder` doesn't have a 1. `putIfAbsent` 2. `containsKey` 3. `get`. I could've used another map as the "reference map" but felt that would be weird. 

Open to suggestions though. WDYT?

src/main/java/com/google/api/codegen/config/GrpcGapicRetryMapping.java
noahdietz: Yes, definitely. Sorry that's the gopher in me using short variable names. I will change them everywhere :smile_cat:

src/main/java/com/google/api/codegen/config/RetryCodesConfig.java
noahdietz: I will change it! Thanks for the tip!

src/main/java/com/google/api/codegen/config/GapicInterfaceConfig.java
noahdietz: Done.

src/main/java/com/google/api/codegen/config/GrpcGapicRetryMapping.java
noahdietz: Done.

src/main/java/com/google/api/codegen/config/RetryCodesConfig.java
noahdietz: Done.

src/main/java/com/google/api/codegen/config/GrpcGapicRetryMapping.java
hzyi-google: That makes sense.

src/main/java/com/google/api/codegen/config/GrpcGapicRetryMapping.java
hzyi-google: Hmmmmm that's very interesting. In that case can you add a comment here to clarify the reason that rounding is needed, just for future reference?

noahdietz: I removed it and let Java just convert from float to double. A value of `1.3` in the config is printed as `1.2999999523162842` in the GAPIC. 

This [SO answer](https://stackoverflow.com/questions/916081/convert-float-to-double-without-losing-precision) has some explanation.

I'll add a comment! Thanks

noahdietz: Done.

noahdietz:
@hzyi-google can you take another quick look at my recent changes? I fixed a scenario where there are multiple different `MethodConfig`s that omit `retry_policy` and have diff `timeout` values. Also, a default `no_retry` config for methods unknown or unspecified methods.

src/main/java/com/google/api/codegen/config/GrpcGapicRetryMapping.java
```diff
@@ -0,0 +1,163 @@
...
+          String fullName = fullyQualifiedName(serviceBase, name.getMethod());
+          methodCodesMap.put(fullName, codesBuilder.getName());
+          methodParamsMap.put(fullName, paramsBuilder.getName());
+          continue;
+        }
+
+        Interface interProto = protoInterfaces.get(serviceBase);
+        for (Method method : interProto.getMethods()) {
+          String fullName = method.getFullName();
+          methodCodesMap.putIfAbsent(fullName, codesBuilder.getName());
```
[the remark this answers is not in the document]
noahdietz: Yes, it is important in this context because this logic applies a Service-wide RetryPolicy to all of its Methods. However, a RetryPolicy for specifically named Method takes precedence over a Service-wide one. Thus, if a Method's RetryPolicy already exists in the mapping, we assume it was added by a named Method RetryPolicy.

Its like `policyA applies to google.cloud.langauge.v1.LangaugeService` while `policyB applies to google.cloud.language.v1.LangaugeService.AnnotateText`.

There is a comment at the top of the parent `for` loop describing this:
```
// apply specific method config (overwrites) or apply service config to all methods (does
// not overwrite method-specific policies)
```

src/main/java/com/google/api/codegen/config/GrpcGapicRetryMapping.java
```diff
@@ -0,0 +1,163 @@
...
+    paramsDefMap.putIfAbsent(defaultParamsBuilder.getName(), defaultParamsBuilder.build());
+
+    RetryCodesDefinitionProto.Builder defaultCodesBuilder = RetryCodesDefinitionProto.newBuilder();
+    defaultCodesBuilder.setName("no_retry_codes");
+    codesDefMap.putIfAbsent(defaultCodesBuilder.getName(), defaultCodesBuilder.build());
+
+    // build retry-to-interface mapping from gRPC ServiceConfig
+    int retryNdx = 1;
+    int noRetryNdx = 1;
+    for (com.google.api.codegen.grpc.MethodConfig methodConfig : config.getMethodConfigList()) {
```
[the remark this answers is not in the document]
noahdietz: Ack. See response to said comment.

src/main/java/com/google/api/codegen/config/GrpcGapicRetryMapping.java
```diff
@@ -0,0 +1,163 @@
...
+  /**
+   * Converts a float to a double rounding to the nearest hundredth.
+   *
+   * <p>This is necessary because when a float is converted to a double the double is more precise.
+   * When that same double is printed (e.g. in a template), the more precise value is used. For
+   * example, a float of 1.3 is printed as 1.2999999523162842. This is undesirable because this
+   * multiplier does not need to be this precise (it was originally a float) and it is not the
+   * expected value.
+   */
+  private static double floatToDouble(float f) {
```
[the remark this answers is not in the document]
noahdietz: Your understanding of my hack here is entirely correct! And your explanation/research is greatly appreciated.

I will try to figure out how we can do the formatting prior to printing.

src/test/java/com/google/api/codegen/protoannotations/testdata/go_library_with_grpc_service_config.baseline
```diff
@@ -0,0 +1,4705 @@
+============== file: cloud.google.com/go/library/apiv1/UpdateBook_smoke_test.go ==============
```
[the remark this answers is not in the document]
noahdietz: > Do we really need all these new baseline files?

Probably not, especially because this feature is mainly to support languages that don't have a microgenerator and need to continue to use this generator OR to support a migration step for particularly complicated APIs.

> My assumption was that the feature allowed to configure retries from a different source, but the actual result...should remain intact.

Correct.

> The problem with having the whole new set of baseline tests is that they go out of control

I totally agree and they have been a hassle to keep up-to-date with master.

I will keep the Java and PHP baselines because Java will probably stay on this generator for a while and PHP won't have a microgenerator (AFAIK), while all other languages have one.

WDYT?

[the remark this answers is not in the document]
noahdietz: > this PR adds 82,868 new lines of code

I'm trying to bump my contributor ranking :wink: just kidding of course

src/main/java/com/google/api/codegen/transformer/RetryDefinitionsTransformer.java
```diff
@@ -93,6 +96,28 @@
...
+          GrpcGapicRetryMapping retryMapping, String service) {
+    Map<String, RetryParamsDefinitionProto> paramsMap = new HashMap<>();
+    ImmutableMap<String, String> methodParamMap = retryMapping.methodParamsMap();
+    ImmutableMap<String, RetryParamsDefinitionProto> paramsDefMap = retryMapping.paramsDefMap();
+
+    methodParamMap
+        .keySet()
+        .stream()
+        .filter(name -> name.startsWith(service))
+        .forEach(
```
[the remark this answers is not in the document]
noahdietz: Honestly, I discovered this API while working on it and I assumed it was "the right thing to do". But you're right, the `for` loop is more than sufficient. I'll take a look at refactoring such occurrences, thank you.

src/main/java/com/google/api/codegen/config/GrpcGapicRetryMapping.java
```diff
@@ -0,0 +1,163 @@
...
+    defaultParamsBuilder.setTotalTimeoutMillis(0);
+    defaultParamsBuilder.setName("no_retry_params");
+    paramsDefMap.putIfAbsent(defaultParamsBuilder.getName(), defaultParamsBuilder.build());
+
+    RetryCodesDefinitionProto.Builder defaultCodesBuilder = RetryCodesDefinitionProto.newBuilder();
+    defaultCodesBuilder.setName("no_retry_codes");
+    codesDefMap.putIfAbsent(defaultCodesBuilder.getName(), defaultCodesBuilder.build());
+
+    // build retry-to-interface mapping from gRPC ServiceConfig
+    int retryNdx = 1;
```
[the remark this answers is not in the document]
noahdietz: Done.

src/main/java/com/google/api/codegen/transformer/RetryDefinitionsTransformer.java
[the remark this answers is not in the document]
noahdietz: Done (in two spots).

src/main/java/com/google/api/codegen/config/GapicMethodConfig.java
```diff
@@ -73,7 +73,8 @@ public Method getMethod() {
       Method method,
       ProtoMethodModel methodModel,
       RetryCodesConfig retryCodesConfig,
-      ImmutableSet<String> retryParamsConfigNames) {
+      ImmutableSet<String> retryParamsConfigNames,
```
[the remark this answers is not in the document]
noahdietz: I didn't change this code, I just added a param following it (and a `,` to the end of this line). Do you want me to refactor this method signature as you said? Just want to make double sure that's what you're asking before I go and do it.

rules_gapic/gapic.bzl
```diff
@@ -36,6 +36,7 @@ def _gapic_srcjar_impl(ctx):
         _set_args(attr.language, "--language=", arguments, required = True)
         _set_args(attr.service_yaml, "--service_yaml=", arguments, inputs)
         _set_args(attr.package_yaml2, "--package_yaml2=", arguments, inputs)
+        _set_args(attr.grpc_service_config, "--grpc_service_config=", arguments, inputs)
```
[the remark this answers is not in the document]
noahdietz: Done.

repositories.bzl
```diff
@@ -70,12 +70,57 @@ def com_google_api_codegen_repositories():
...
+
     _maybe(
         native.bind,
         name = "zlib",
         actual = "@net_zlib//:zlib",
     )
 
+    _maybe(
+        native.maven_jar,
+        name = "guava_maven",
```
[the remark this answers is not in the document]
noahdietz: > but will it fail if compiled with newer guava (i.e. if you put on L97 "@com_google_guava_guava__com_google_api_codegen//jar" instead, will it break)?

Just tried this and the build still works! Gonna roll with this for now.

> Is there a better way to parse protobuf jsons?

No, not that I can find. It is the only lib I can find that will read JSON-pb format into the corresponding Message builder.

repositories.bzl
```diff
@@ -70,12 +70,57 @@ def com_google_api_codegen_repositories():
...
+    _maybe(
+        native.bind,
+        name = "guava",
+        actual = "@guava_maven//jar",
+    )
+
+    _maybe(
+        native.maven_jar,
+        name = "gson_maven",
+        artifact = "com.google.code.gson:gson:2.7",
```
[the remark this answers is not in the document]
noahdietz: Done.

src/test/java/com/google/api/codegen/protoannotations/testdata/go_library_with_grpc_service_config.baseline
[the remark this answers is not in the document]
noahdietz: Done.

noahdietz:
> 

@vam-google I included something like this. It does not test the conversion of JSON file-to-Java class (test utils do that reading), but it exercises all of the retry configuration and assignment logic. You can see that [here](https://github.com/googleapis/gapic-generator/pull/2859/files#diff-bad829ba2e813547d0e050fc65417a57R83). It does not use the APIs directly, it activates the ServiceConfig code path.

If you'd like to see more tests, please lmk, happy to add more.

src/main/java/com/google/api/codegen/config/GrpcGapicRetryMapping.java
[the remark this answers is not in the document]
noahdietz: OK, so this is actually the only place the value is assigned before being printed (it's just passed around between here and the snippets). The snippets call `{@def.params.getRetryDelayMultiplier}` which is a method on the `RetryParamsDefinitionProto`. I could add some method to the snippets and then change all of the snippets to use that for formatting the value from the definition? This seems overly complicated...

Or I could do something like this, but I guess it's still changing the value too early like you said not to do...
```java
private static double floatToDouble(float f) {
    String formatted = new DecimalFormat(".##").format((double) f);
    return Double.parseDouble(formatted);
}
```

WDYT?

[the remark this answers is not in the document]
AlanGasperini: what about consistently using floats/doubles everywhere?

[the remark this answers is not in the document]
noahdietz: These values represent fields from protos that are defined as `float` (gRPC ServiceConfig) and `double` (GAPIC Config). So we _could_ change one of them, it would really have to be the GAPIC Config to not diverge from gRPC, but that would result in cascading changes elsewhere in the code base. I'm not sure it's worth the effort :smile:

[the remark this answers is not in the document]
AlanGasperini: Can we use 1.25 instead of 1.3? :)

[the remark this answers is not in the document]
noahdietz: > Can we use 1.25 instead of 1.3? :)

No, because the value is actually `1.3` in the config in this case, but conversion to `double` does the above mangling^^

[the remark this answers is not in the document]
vam-google: **tl;dr;** Some brief history of gapic-generator architecture is below. I helped me, when I finally understood what was going on in gapic-generator, and is kind of relevant here as we decide the proper place where the toString conversion should go.

**tl; dr; 2** If you think it is more complicated than it should be, then you are not alone... I believe that the "thing" described below is the main source of gapic-generator complications and why people don't like it. Fixing that should be doable though (different topic).

**The very biref history of gapic-generator's architecture**
On a very high level the architecture of the gapic-generator tries to follow MVC (Model-View-Controller) model (if it applies well here and if it was a good idea is a different story...):
1. Consume all config values from everywhere (protos, yamls, command line args, stuff...). We can think about those as the **"Model"**.
2. The gathered config values (see item 1) are used for two main purposes:
    1. Control generation flow (i.e. determine which files should be generated and how). The **"Controller"** part (well, the term is applied here very loosely, but close enough)
    2. Determine what exactly should be put in those generated files (i.e. the **actual text**).
3. The actual text part is determined by a combination of in `.snip` files and corresponding `*View` objects (the **"View"** part). GAPIC simply takes a snip file, a View object and applies one another to get a text file output.
     1. Notice that the methods of the view classes (like `RetryParamsDefinitionView` in our case) almost solely return either `String`s or collections of `Strings`. This is so because they are all about "text".
     2. Something like `{@def.params.getMaxRetryDelayMillis}` in a snip file literally means a call to `def.params().getMaxRetryDelayMillis()`, where `def` is an object of a `*View` type (`RetryParamsDefinitionView` in our case).
4. To create the `*View` objects, a corresponding `*Transformer` object is used (i guess transformers loosely fall in to the Controller part). It will be the `RetryDefinitionsTransformer` class in our case.

**How it applies to our concrete problem**
When template parameters are applied in snip files, those are taken from `RetryParamsDefinitionView` object. But that object has nested objects in it (`RetryDefinitionsTransformer`), and those are returned directly. Which is **fine if you do not need any special string representation logic** in place for that object. But looks like we have it here (to print the float properly). 

To solve it "idiomatically" for gapic-generator, we need a nested View object to be returned. Nested view objects are very common in gapic-generator (for example `ServiceDocView` has `exampleApiMethod()` which returns a nested view `ApiMethodView`).

So we need to create a class, something like `RetryParamsDefinitionProtoView`, and control there the String representation of its fields.

```java
public class RetryParamsDefinitionProto {
  // For fields which don't have custom logic, just propagate them as is
  public String getRetryDelayMultiplier() {
    return new DecimalFormat(".##").format(proto.getRetryDelyMultiplier());
  }
  // ...
}
```

Then, when `RetryParamsDefinitionView` is created in `RetryDefinitionsTransformer` the view object is put (instead of RetryParamsDefinitionProto) such tat `RetryParamsDefinitionView.params()` returns `RetryParamsDefinitionProtoView` instead of `RetryParamsDefinitionProto` and when called from snip file that will output proper string.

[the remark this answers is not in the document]
vam-google: **tl;dr;** Some brief history of gapic-generator architecture is below. I helped me, when I finally understood what was going on in gapic-generator, and is kind of relevant here as we decide the proper place where the toString conversion should go.

**tl; dr; 2** If you think it is more complicated than it should be, then you are not alone... I believe that the "thing" described below is the main source of gapic-generator complications and why people don't like it. Fixing that should be doable though (different topic).

**The very biref history of gapic-generator's architecture**
On a very high level the architecture of the gapic-generator tries to follow MVC (Model-View-Controller) model (if it applies well here and if it was a good idea is a different story...):
1. Consume all config values from everywhere (protos, yamls, command line args, stuff...). We can think about those as the **"Model"**.
2. The gathered config values (see item 1) are used for two main purposes:
    1. Control generation flow (i.e. determine which files should be generated and how). The **"Controller"** part (well, the term is applied here very loosely, but close enough)
    2. Determine what exactly should be put in those generated files (i.e. the **actual text**).
3. The actual text part is determined by a combination of in `.snip` files and corresponding `*View` objects (the **"View"** part). GAPIC simply takes a snip file, a View object and applies one another to get a text file output.
     1. Notice that the methods of the view classes (like `RetryParamsDefinitionView` in our case) almost solely return either `String`s or collections of `Strings`. This is so because they are all about "text".
     2. Something like `{@def.params.getMaxRetryDelayMillis}` in a snip file literally means a call to `def.params().getMaxRetryDelayMillis()`, where `def` is an object of a `*View` type (`RetryParamsDefinitionView` in our case).
4. To create the `*View` objects, a corresponding `*Transformer` object is used (i guess transformers loosely fall in to the Controller part). It will be the `RetryDefinitionsTransformer` class in our case.

**How it applies to our concrete problem**
When template parameters are applied in snip files, those are taken from `RetryParamsDefinitionView` object. But that object has nested objects in it (`RetryDefinitionsTransformer`), and those are returned directly. Which is **fine if you do not need any special string representation logic** in place for that object. But looks like we have it here (to print the float properly). 

To solve it "idiomatically" for gapic-generator, we need a nested View object to be returned. Nested view objects are very common in gapic-generator (for example `ServiceDocView` has `exampleApiMethod()` which returns a nested view `ApiMethodView`).

So we need to create a class, something like `RetryParamsDefinitionProtoView`, and control there the String representation of its fields.

```java
public class RetryParamsDefinitionProto {
  // For fields which don't have custom logic, just propagate them as is
  public String getRetryDelayMultiplier() {
    return new DecimalFormat(".##").format(proto.getRetryDelyMultiplier());
  }
  // ...
}
```

Then, when `RetryParamsDefinitionView` is created in `RetryDefinitionsTransformer` the view object is put (instead of RetryParamsDefinitionProto) such tat `RetryParamsDefinitionView.params()` returns `RetryParamsDefinitionProtoView` instead of `RetryParamsDefinitionProto` and when called from snip file that will output proper string.

src/main/java/com/google/api/codegen/config/GrpcGapicRetryMapping.java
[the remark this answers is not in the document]
vam-google: Ok, that  makes perfect sense, and also becomes a natural target for refactoring this big method: split the service-wide logic and method-wide logic (i.e. method-wide porition goes into a private method and is called here instead). If it is still big enough, we can put the internals of the for loop itself in another private method (so ther ewill be this method, calling loop internals  private method, which in its turn calls the method-wide logic private method).

The above is just a suggestion (please disregard it if you have other refactoring ideas or a solid reason to keep this method as big as it  already is).

repositories.bzl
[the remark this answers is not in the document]
vam-google: I'm a bit confused here. The comment says build still works without this, but it is still here. I would not be surprised if it broke, but otherwise, please clean this up (the idea is to keep the dependencies as clean as possible, because they become a nightmare, when gapic-generator is used in wider scope and/or have to be kept up to date with its dependencies).

src/main/java/com/google/api/codegen/config/GrpcGapicRetryMapping.java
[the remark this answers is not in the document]
noahdietz: I'm not sure I follow the suggestion.
> split the service-wide logic and method-wide logic (i.e. method-wide porition goes into a private method and is called here instead)

The Method-specific logic should use `put()` to overwrite/take precedence, while the Service-wide logic should use `putIfAbsent()` so as to **not** overwrite an existing Method-specific policy. So it's not clear to me how that helper would be reused. Am I missing something?

If it is about reducing the line size of this function, I could also move the RetryPolicy-to-Retry(Params | Codes)Definition building logic into private helpers.

src/main/java/com/google/api/codegen/config/GrpcGapicRetryMapping.java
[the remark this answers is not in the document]
noahdietz: > So we need to create a class, something like RetryParamsDefinitionProtoView, and control there the String representation of its fields.

Just to be clear, is that the recommendation? Or is this just an example of how it would be done "idiomatically" given the history?

repositories.bzl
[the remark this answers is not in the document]
noahdietz: This comment is on an outdated diff. There is no `guave_maven` dependency in `repositories.bzl`, just the `@com_google_guava_guava__com_google_api_codegen//jar"` you suggested I use [here](https://github.com/googleapis/gapic-generator/pull/2859/files#diff-515bc54a0cbb4b12fb4a7c465758b011R88-R92).

src/main/java/com/google/api/codegen/config/GrpcGapicRetryMapping.java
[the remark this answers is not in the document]
noahdietz: I have refactored it. I pulled the Retry(Params | Codes)Definition building logic into private helpers and moved the `for` loop that applies the retry policy names to a method/service name into a private helper to make it read a bit better. 

@vam-google WDYT?

src/main/java/com/google/api/codegen/config/GrpcGapicRetryMapping.java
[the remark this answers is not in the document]
vam-google: @noahdietz At this point I'm wondering why RetryPolicy.backoff_multiplier is declared as float (they could just declare it as double and make everybody happy). But if we can't change that, we do need to keep string representation and exact values separate. 

In general, we need to take double values under control. Probably the least verbose way we can do it is changing `RetryParamsDefinitionView` class by adding the `RetryParamsDefinitionProtoView` in it:

```java
@AutoValue
public abstract class RetryParamsDefinitionView {

  // ... Unchanged methods ...
  // Replace returned value from RetryParamsDefinitionProto to RetryParamsDefinitionProtoView
  public abstract RetryParamsDefinitionProtoView params();
  
  // ... Unchanged methods ... 

  // Add the view class as public static internal (to minimize spawning new classes and files)
  public static class RetryParamsDefinitionProtoView {
    private final RetryParamsDefinitionProto params;

    public RetryParamsDefinitionProtoView(RetryParamsDefinitionProto params) {
      this.params = params;
    }

    public String getName() {
      return params.getName();
    }

    public long getInitialRetryDelayMillis() {
      return params.getInitialRetryDelayMillis();
    }

    public String getRetryDelayMultiplier() {
      return new DecimalFormat(".####").format(params.getRetryDelayMultiplier());
    }

    public long getMaxRetryDelayMillis() {
      return params.getMaxRetryDelayMillis();
    }

    public long getInitialRpcTimeoutMillis() {
      return params.getInitialRpcTimeoutMillis();
    }

    public String getRpcTimeoutMultiplier() {
      return new DecimalFormat(".####").format(params.getRpcTimeoutMultiplier());
    }

    public long getMaxRpcTimeoutMillis() {
      return params.getMaxRpcTimeoutMillis();
    }

    public long getTotalTimeoutMillis() {
      return params.getTotalTimeoutMillis();
    }
  }
  
  @AutoValue.Builder
  public abstract static class Builder {
    // ... Unchanged methods in Builder class ...

    // Change input parameter value from RetryParamsDefinitionProto to RetryParamsDefinitionProtoView
    public abstract Builder params(RetryParamsDefinitionProtoView val);
  }
}
```

Then in `RetryDefinitionsTransformer` change the line 127 to:
```java
      view.params(
          new RetryParamsDefinitionView.RetryParamsDefinitionProtoView(retryCodesDef.getValue()));

```

That should be it (no more changes required anywhere, including snip files).

src/main/java/com/google/api/codegen/config/GrpcGapicRetryMapping.java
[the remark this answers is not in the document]
vam-google: @noahdietz Looks good to me. If we did not need the `putIfAbsent` stuff, then it could look prettier (instead of passing each map separately, we could pass one big builder of the final object (which aggreates the three maps builder), thus keeping the list of parameters of the private methods shorter). But it is all ok.

My concern was not that much about the size (sometimes it is just ok to have big methods, actually I think the refactored version is even slightly bigger then original). It was more about ensuring that each method does one concrete thing.

repositories.bzl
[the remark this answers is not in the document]
vam-google: @noahdietz Oh, sorry, you are right, I looked at a wrong dif. LGTM

src/main/java/com/google/api/codegen/config/GrpcGapicRetryMapping.java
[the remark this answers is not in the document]
noahdietz: Thanks a lot @vam-google, this worked great for Java. I had to tweak the formatting methods to return `""` when the value is `0` because it was printing `.0`. Empty string means the snippet won't print the setter for that value, which is the preferred behavior, vs. printing the setter with `.0` in it.

One issue however is the that this doesn't work with languages that load client config from a JSON file: php, python, nodejs and ruby. This View we've changed never gets called, and instead, the JSON representation of the value is just printed, meaning we get `1.2999999523162842` instead of `1.3`. I'm trying to find the section that handles this case...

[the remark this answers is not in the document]
noahdietz: Correction, a different method is being called in the [clientconfig/common.snip](https://github.com/googleapis/gapic-generator/blob/8de8e350194d1a72ac171a57e368e7a9e7e416a8/src/main/resources/com/google/api/codegen/clientconfig/common.snip#L62) that generates the JSON client config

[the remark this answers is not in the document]
noahdietz: @hzyi-google mentioned it before, `RetryParamsDefinitionProto` is one of the few config classes using the Model directly, rather than an intermediate class (like RetryCodesConfig). This makes it hard to build on for things like this. The concrete model is being passed directly to the snippet.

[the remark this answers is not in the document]
vam-google: Grrr... Such a mess =(... we have such a ridiculous mess in gapic-generator...

So when I look to it closer, we actually already have:
- RetryCodesDefinitionView
- RetryConfigDefinitionView
- RetryParamsDefinitionView

We already have 3 classes which do same thing... People just add stuff completely disregarding what is already there. Such a mess. The "idiomatic" way would be fixing that mess and introducing one view which covers all cases. 

**For the clientconfig/common.snip:**
That should not even be part of gapic generator (it should not generate those type of files)... So much garbage =(. To process those the `LegacyGapicGenrator` is used  which does not work with views...

**What to do**
Ok, at this point it is just me torturing you and forcing to the yak shaving... Scrap it, it is an overwhelming mess that is not your fault, so expecting you to fix it as part of this pr will be not ok. 

Probably the best thing we can do here without starting the yak shaving thing will be changing this method implementation to something like :
```java
return Double.valueOf(Float.toString(1.3F)).doubleValue()
```
This converts `float` to `double` via their's string representation "without loosing" or explicitly specifying precision. This is fundamentally wrong, but kind of ok for us, because the values themselves come from strings and go to strings (the gRPC service config, the source of truth of that value is actually a string).

Sorry, I did not realize that retries have many "views" (this must be fixed) and that the `LegacyGapicGenerator` is still in use.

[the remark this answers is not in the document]
noahdietz: Thanks for all of your help, @vam-google. I went with your final suggestion here.

@AlanGasperini what Vadym mentions might be nice technical debt to target in a fix-it week whenever one of those things comes around again.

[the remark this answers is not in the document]
AlanGasperini: sgtm

PR #2860, opened 2019-06-28
Allow samplegen without sample configs

vchudnov-g:
We are not yet requiring sample configs because we only parse them but
do not use the results (and users have not yet converted from the
gapic configs). PR #2828 accidentally made having these sample configs
parsed a requirement for samplegen. This breaks users.

This PR ignores whether the sample configs have been parsed in order
to re-enable samplegen for the time being.


alexander-fenster:
Cc: @beccasaurus

PR #2862, opened 2019-06-28
Reconstruct PackagingConfig from other sources (gapic_yaml and protos)

vam-google:
This makes `--package_yaml2` command line argument fully optional (for all languages). `--package_yaml2` is created by artman from `artman_yaml.` Not needing `--package_yaml2` as an argument effectively removes gapic-generator's "indirect dependency" on artman.

This is a prerequisite for Bazel generation for all languages which relay on `PackagingConfig` data (Ruby specifically, and all of the samples in all of the languages).


src/main/java/com/google/api/codegen/config/PackagingConfig.java
```diff
@@ -28,12 +30,18 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import javax.annotation.Nullable;
 import org.yaml.snakeyaml.Yaml;
 
 /** Represents the packaging config necessary to run package generation for an API. */
 @AutoValue
 public abstract class PackagingConfig {
+  private static final Pattern COMMON_PKG_PATTERN =
+      Pattern.compile("(?i)(?<org>(\\w+\\.){0,2})(?<name>\\w+)\\.(?<ver>\\w*\\d+[\\w\\d]*)(\\.|$)");
```
hzyi-google: This doesn't seem to work for grafeas Java gapic?
https://github.com/googleapis/googleapis/blob/master/grafeas/v1/grafeas_gapic.yaml

hzyi-google: Talked offline: different issue.

vam-google: As we discussed offline, this should not matter.

PR #2864, opened 2019-07-02
[fix] Generate helper methods for all resource names of a referenced oneof resource name

hzyi-google:
*[EDIT: please no internal tickets on Github]*

Gapic-generator currently generates helper methods for all the following resource names:
1. resource names defined under the interface
2. resource names defined top-level, and referenced in `field_name_patterns` of any methods of the interface

For 2., if the resource name is a oneof resource name, only helper methods of the first resource name of the oneof resource name is generated. This PR generates all of them.

The change works for proto annotations too.


hzyi-google:
Python failure seems unrelated. @noahdietz and @software-dov are looking into it.

hzyi-google:
PTAL

PR #2870, opened 2019-07-11
C# Add method to format resource-names

chrisdunelm:
Updates https://github.com/googleapis/google-cloud-dotnet/issues/2443


chrisdunelm:
Thought it was worth doing this, as I was already working on this code.

chrisdunelm:
Yes, this is exactly the code that the micro-generator generates :)
(Unless I've made a mistake...)

PR #2871, opened 2019-07-11
SampleGen: use sample config v1.2 in Ruby

hzyi-google:


src/test/java/com/google/api/codegen/gapic/testdata/ruby/ruby_samplegen_config_migration_library.baseline
```diff
@@ -5402,7 +5402,7 @@ gem "google-cloud-library", path: "../../"
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# DO NOT EDIT! This is a generated sample ("RequestStreamingClient",  "empty_response_type_with_response_handling")
+# DO NOT EDIT! This is a generated sample ("RequestStreamingClient",  "babble_about_book_1")
```
hzyi-google: We don't have `sample_value_set` and more, so use sample ID instead.

src/test/java/com/google/api/codegen/gapic/testdata/ruby/ruby_samplegen_config_migration_library.baseline
```diff
@@ -6505,7 +6505,7 @@ def sample_get_shelf
   message = {}
   string_builder = {}
 
-  response = library_client.get_shelf(formatted_name, options_)
+  response = library_client.get_shelf(formatted_name, options_, message: message, string_builder: string_builder)
```
hzyi-google: The previous generated code was wrong - this change fixes it.

tl;dr: message and string_builder should be passed in.
implementation detail: we were not always using the correct `initCodeView` (we were using the once generated for incode samples, which most of the time, is the same as the standalone samples). See changes in Ruby standalone snippet for reference.

src/test/java/com/google/api/codegen/gapic/testdata/ruby/ruby_library.baseline
```diff
@@ -6331,7 +6331,7 @@ def sample_publish_series
   books = []
   series_uuid = {}
 
-  response = library_client.publish_series(shelf, books, series_uuid, edition: edition)
+  response = library_client.publish_series(shelf, books, series_uuid)
```
hzyi-google: The originally generated code was wrong (there isn't a `edition` variable).

It was because that we were using the wrong initCodeView (we used the one for incode sample). See also: the changes in [ruby/standalone_sample.snip](https://github.com/googleapis/gapic-generator/pull/2871/files#diff-5bef66642b3e7df4a108d12f2c335766L19).

hzyi-google:
PTAL

src/main/java/com/google/api/codegen/config/SampleContext.java
```diff
@@ -0,0 +1,72 @@
...
+import com.google.auto.value.AutoValue;
+
+/** The context of transforming a sample into a view model. */
+@AutoValue
+public abstract class SampleContext {
+
+  /**
+   * The unique sample ID generated by the generator.
+   *
+   * <p>If the user provided a sample ID, the user provided ID will be used as a base ID.
```
vchudnov-g: I don't think this additional description of _how_ we create a unique ID belongs here but rather in the comments to `SampleFileRegistry`, which is the class that implements the algorithm you describe. The comments here should just say it's unique and refer folks to the other class for details. The reasoning is that this way we keep the description close to the implementation, and can make sure they're accurate whenever we tweak anything.

src/main/java/com/google/api/codegen/transformer/SampleFileRegistry.java
```diff
@@ -35,48 +38,71 @@
...
+      userProvidedIdCount.put(idWithCallingPattern, 1);
+      return idWithCallingPattern;
+    }
+
+    int suffix = usedSuffixes.getOrDefault(idWithCallingPattern, 0) + 1;
+    usedSuffixes.put(idWithCallingPattern, suffix);
+    return idWithCallingPattern + suffix;
+  }
+
+  @Deprecated
```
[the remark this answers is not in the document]
hzyi-google: Yes! Will do.

src/main/java/com/google/api/codegen/transformer/SampleTransformer.java
```diff
@@ -222,6 +224,88 @@ public void generateSamples(
     return methodSampleViews;
   }
 
+  // entry point for generating standalone samples using sample config.
```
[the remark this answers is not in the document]
hzyi-google: By "entry point" I meant that when generating samples using sample config (instead of gapic config), `LangGapicSamplesTransformer` should call this method rather than other overloads.

src/main/java/com/google/api/codegen/config/SampleContext.java
hzyi-google: SGTM

src/main/java/com/google/api/codegen/config/SampleConfig.java
```diff
@@ -126,11 +137,16 @@ public static Builder newBuilder() {
...
+      @Nullable SampleConfigProto sampleConfigProto,
+      final Map<String, InterfaceConfig> interfaceConfigMap) {
+    if (sampleConfigProto == null) {
+      sampleConfigProto = SampleConfigProto.getDefaultInstance();
+    }
+
     // First, apply region tag as IDs if IDs are not given
     List<SampleSpecProto> sampleSpecs = new ArrayList<>();
-    for (SampleSpecProto spec : sampleSpecs) {
+    for (SampleSpecProto spec : sampleConfigProto.getSamplesList()) {
```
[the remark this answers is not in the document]
hzyi-google: Nice catch! I don't think I have handled this. Will add it according to https://github.com/googleapis/gapic-generator/pull/2871#discussion-diff-303647292R28

PR #2881, opened 2019-07-19
fix(nodejs): drop dependency on through2

JustinBeckwith:
The gapic generated tests for nodejs often take a dependency on the `through2` module.  Oftentimes... it's not actually needed!  This is in pursuit of dropping the total number of dependencies in our projects to improve security and reliability.


alexander-fenster:
Thanks @JustinBeckwith! I have updated one baseline test that was still failing.

PR #2882, opened 2019-07-19
Java: Remove unsupported retry samples

### LINKED ISSUE #2880: Java: Remove incorrect documentation that configures retry through StreamingCallSettings
https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-video-intelligence/src/main/java/com/google/cloud/videointelligence/v1p3beta1/stub/StreamingVideoIntelligenceServiceStubSettings.java#L64-L68

This method doesn't exist.

hzyi-google:
gapic-generator generates a sample code in `XapiStubSettings.java` to illustrate the builder of the Setting classes are recursive. For a service, it does so by changing the retry settings of one of the methods in this service.

This sample code is wrong when the service only contains bidi/client streaming methods, because the respective `StreamingCallSettings` classes do not have retry settings. Example: https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-video-intelligence/src/main/java/com/google/cloud/videointelligence/v1p3beta1/stub/StreamingVideoIntelligenceServiceStubSettings.java#L64-L68

While the long-term solution should be adding retry support to [`StreamingCallSettings`](https://github.com/googleapis/gax-java/blob/master/gax/src/main/java/com/google/api/gax/rpc/StreamingCallSettings.java), we should remove the wrong sample code before the feature is added.

Fix #2880

### DIFF
--- a/src/test/java/com/google/api/codegen/gapic/testdata/java/java_my_streaming_proto.baseline
+++ b/src/test/java/com/google/api/codegen/gapic/testdata/java/java_my_streaming_proto.baseline
@@ -19,7 +19,7 @@ buildscript {
 
 apply plugin: 'java'
 
-description = 'GAPIC library for google-cloud-library-v1'
+description = 'GAPIC library for google-cloud-my_streaming_service-v1'
 group = 'com.google.cloud'
 version = '0.0.0-SNAPSHOT'
 sourceCompatibility = 1.7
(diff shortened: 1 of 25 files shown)


src/test/java/com/google/api/codegen/gapic/testdata/java/java_my_streaming_proto.baseline
```diff
@@ -0,0 +1,1434 @@
...
+ * <li>The default service address (mystreaming-example.googleapis.com) and default port (443)
+ * are used.
+ * <li>Credentials are acquired automatically through Application Default Credentials.
+ * <li>Retries are configured for idempotent methods but not for non-idempotent methods.
+ * </ul>
+ *
+ * <p>The builder of this class is recursive, so contained classes are themselves builders.
+ * When build() is called, the tree of builders is called to create the complete settings
+ * object.
+ *
```
hzyi-google: The example shouldn't be generated here.

eoogbe:
Just a quick note: your PR fixes itself. (`Fix #2882` is the same as this PR number.)

hzyi-google:
@eoogbe Please pretend you did not see that :p

hzyi-google:
PTAL

PR #2883, opened 2019-07-22
SampleGen: Migration all dynamic languages to use sample config

hzyi-google:


hzyi-google:
PTAL

cc/ @googleapis/samplegen

vchudnov-g:
Separate question: where are we at with sample id uniquification? Is that already coded or yet to be done?

src/main/java/com/google/api/codegen/transformer/DynamicLangGapicSamplesTransformer.java
```diff
@@ -0,0 +1,236 @@
...
+      InterfaceContext context,
+      String sampleFileName,
+      OptionalArrayMethodView method,
+      MethodSampleView sample) {
+    String outputPath =
+        Paths.get(
+                pathMapper.getSamplesOutputPath(
+                    context.getInterfaceModel().getFullName(),
+                    productConfig,
+                    method.name()), // we don't need the last one?
```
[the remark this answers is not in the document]
hzyi-google: Oh whoops. Debugging comment. I'll remove it.

src/main/java/com/google/api/codegen/transformer/DynamicLangGapicSamplesTransformer.java
```diff
@@ -0,0 +1,236 @@
...
+                    .sampleType(SampleSpec.SampleType.STANDALONE)
+                    .callingForm(form)
+                    .sampleConfig(sampleConfig)
+                    .initCodeOutputType(initCodeOutputType)
+                    .build();
+
+            OptionalArrayMethodView methodView =
+                apiMethodTransformer.generateApiMethod(methodContext, sampleContext);
+
+            MethodSampleView methodSampleView = methodView.samples().get(0);
```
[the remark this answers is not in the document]
hzyi-google: Not for standalone samples. But in some languages one method can have more than one in-code samples. So, :man_shrugging:

src/main/java/com/google/api/codegen/transformer/SampleFileRegistry.java
```diff
@@ -112,6 +112,10 @@ public String getSampleFileName(MethodSampleView sample, String method) {
     if (userProvidedIdCount.get(regionTag) == 1) {
       fileName = namer.getApiSampleFileName(regionTag);
     } else {
+      // method names can be in snake_case, camelCase or ParscalCase depending on the language
+      Name methodName =
```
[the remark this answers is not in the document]
hzyi-google: That's a good point. But this method is only needed by generating samples using gapic config and we are deprecating that, so it's probably not worth it :p

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSamplesTransformer.java
```diff
@@ -18,183 +18,74 @@
...
-    return ImmutableList.of(STANDALONE_SAMPLE_TEMPLATE_FILENAME);
+    super(
+        STANDALONE_SAMPLE_TEMPLATE_FILENAME,
+        pathMapper,
+        fileHeaderTransformer,
+        apiMethodTransformer,
+        new PhpFeatureConfig(),
+        p -> new PhpSurfaceNamer(p.getPackageName()),
+        p ->
+            new ModelTypeTable(
```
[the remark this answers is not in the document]
hzyi-google: Just tried, the formatter doesn't like it :(

src/main/java/com/google/api/codegen/viewmodel/CallingForm.java
```diff
@@ -153,15 +153,25 @@ static RpcType fromMethodContext(MethodContext context) {
               .put(PHP, RpcType.UNARY, ImmutableList.of(Request))
               .put(PHP, RpcType.LRO, ImmutableList.of(LongRunningRequest, LongRunningRequestAsync))
               .put(PHP, RpcType.PAGED_STREAMING, ImmutableList.of(RequestPaged, RequestPagedAll))
-              .put(PHP, RpcType.CLIENT_STREAMING, ImmutableList.of(RequestStreamingClient))
+              .put(
+                  PHP,
```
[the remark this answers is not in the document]
hzyi-google: The formatter doesn't like this either :(

hzyi-google:
The sample id quantification is done here: https://github.com/googleapis/gapic-generator/blob/df80c1509b34613dd67623f03b846fbe9b1f2feb/src/main/java/com/google/api/codegen/transformer/SampleFileRegistry.java#L64

And I forgot to change the PHP test to use sample configs :( I'll update that

src/main/java/com/google/api/codegen/transformer/DynamicLangGapicSamplesTransformer.java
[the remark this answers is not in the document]
vchudnov-g: OK, makes sense to leave it, then, as long as we have incode samples (which we're not in the micros, for now)

src/main/java/com/google/api/codegen/transformer/SampleFileRegistry.java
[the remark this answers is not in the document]
vchudnov-g: sg

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSamplesTransformer.java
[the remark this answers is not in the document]
vchudnov-g: drats! Not a big fan of how this looks. Oh, well.

src/main/java/com/google/api/codegen/transformer/DynamicLangGapicSamplesTransformer.java
```diff
@@ -0,0 +1,258 @@
...
+    for (InterfaceContext interfaceContext : interfaceContexts) {
+      for (MethodModel method : interfaceContext.getSupportedMethods()) {
+        MethodContext methodContext = interfaceContext.asRequestMethodContext(method);
+        String interfaceName =
+            interfaceContext.getInterfaceConfig().getInterfaceModel().getFullName();
+        String methodName = method.getSimpleName();
+        MoreObjects.firstNonNull(
+                sampleConfigTable.get(interfaceName, methodName), ImmutableList.<SampleConfig>of())
+            .stream()
+            .forEach(configNamesCount.put(c, namer.getMatchingCallingForms(methodContext, c.callingPattern())))
```
vchudnov-g: I'm confused by this sequence. Could you add some clarifying comments? In particular, you're getting a list of matching calling forms, and then you're calling a very similar but separate function to see whether there are any matching calling forms, and then you're putting something again? 

A loop might be simpler
(see, eg, https://blog.jooq.org/2015/12/08/3-reasons-why-you-shouldnt-replace-your-for-loops-by-stream-foreach/)

vchudnov-g: This should be a lambda inside the `forEach`, right? Otherwise `c` is not defined.

src/main/java/com/google/api/codegen/transformer/DynamicLangGapicSamplesTransformer.java
```diff
@@ -0,0 +1,258 @@
...
+            .filter(c -> hasMatchingCallingForm(c.callingPattern(), namer, methodContext))
+            .forEach(c -> configNamesCount.put(c));
```
vchudnov-g: What does it mean to `put` into a `Map` without a value?

src/test/java/com/google/api/codegen/gapic/testdata/php/php_library.baseline
```diff
@@ -1505,7 +1505,7 @@ function samplePublishSeries()
     $seriesUuid = new SeriesUuid();
 
     try {
-        $response = $libraryServiceClient->publishSeries($shelf, $books, $seriesUuid, ['edition' => $edition]);
+        $response = $libraryServiceClient->publishSeries($shelf, $books, $seriesUuid);
```
vchudnov-g: What's the significance of this changing? Was it not right before?

src/test/java/com/google/api/codegen/gapic/testdata/php/php_library.baseline
```diff
@@ -1896,7 +1896,7 @@ function sampleGetShelf()
     $stringBuilder = new StringBuilder();
 
     try {
-        $response = $libraryServiceClient->getShelf($formattedName, $options);
+        $response = $libraryServiceClient->getShelf($formattedName, $options, ['message' => $message, 'stringBuilder' => $stringBuilder]);
```
vchudnov-g: ditto

src/main/java/com/google/api/codegen/transformer/DynamicLangGapicSamplesTransformer.java
hzyi-google: Sorry I run into some problem when add PHP test, but I had to leave yesterday so pushed some WIP code here so that I can catch up today WFH. I'll let you know when it's ready.

hzyi-google: same as above

src/main/java/com/google/api/codegen/transformer/DynamicLangGapicSamplesTransformer.java
hzyi-google: same as above

src/test/java/com/google/api/codegen/gapic/testdata/php/php_library.baseline
hzyi-google: same as above

src/test/java/com/google/api/codegen/gapic/testdata/php/php_library.baseline
hzyi-google: same as above.

hzyi-google: This change is necessary. This is quite similar to Ruby: we were not using the correct `initCode` in the template. We actually constructed the variable `$stringBuilder` a few lines above so it should be passed in.

src/test/java/com/google/api/codegen/gapic/testdata/php/php_library.baseline
hzyi-google: Right. This is also due to that we are not using the correct `initCode`. See https://github.com/googleapis/gapic-generator/pull/2883/files#diff-70645adb9110db92a78ae7061153d8ddL168.

(Also we don't have `edition` at all :p)

hzyi-google:
@vchudnov-g I discovered that de-duplicating logic wasn't correct when users specify `calling_patterns` with `.*` which matches to more than one `CallingForm`. Now everything should be back to normal.

Also `SampleFileRegistry` is now taking a map as a parameter in its constructor, which is sort of confusing. Given we have more languages to migrate and we need to do some cleanup anyway, I decided to leave it as it is for now. Issue filed: https://github.com/googleapis/gapic-generator/issues/2893

hzyi-google:
PTAL

src/main/java/com/google/api/codegen/transformer/SampleFileRegistry.java
```diff
@@ -54,37 +55,53 @@ public SampleFileRegistry(SurfaceNamer namer, List<MethodSampleView> allSamples)
...
-    Preconditions.checkState(count != null && count > 0, "Sample not registered.");
 
-    if (count == 1) {
-      return userProvidedId;
+    // Find a unique id based on base id and save the unique id.
+    String calculatedId = calculatedBaseId;
+    while (userProvidedIdCount.containsKey(calculatedId)) {
+      int suffix = firstNonNull(usedSuffixes.get(calculatedBaseId), 0) + 1;
+      usedSuffixes.put(calculatedId, suffix);
+      calculatedId = calculatedBaseId + suffix;
```
vchudnov-g: should we have an underscore before the numeric suffix, so it reads better in snake case?

hzyi-google: Sure.

PR #2885, opened 2019-07-24
Drop "non_empty = True" to fix a bazel error.

g-easy:
This change is needed to fix a build-time error that started happening with
bazel version 0.27.

See also: https://github.com/bazelbuild/bazel/issues/5818


g-easy:
Done.

The error looks like:
```
File ".../external/com_google_api_codegen/rules_gapic/php/php_gapic_pkg.bzl", line 59
		rule(attrs = {"deps": attr.label_list...)}, <2 more arguments>)
	File ".../external/com_google_api_codegen/rules_gapic/php/php_gapic_pkg.bzl", line 61, in rule
		attr.label_list(allow_files = True, mandatory = Tr..., ...)
'non_empty' is no longer supported. use allow_empty instead. You can use --incompatible_disable_deprecated_attr_params=false to temporarily disable this check.
```

PR #2886, opened 2019-07-25
SampleGen: Adjust C# region tags location

### LINKED ISSUE #2873: SampleGen: C# region tag location
Now, C# samples go like:
```c#
// License
// Generated sample, do not edit!

use CommandLine;

// [START region_tag]
namespace Com.Google.Cloud.Language.Samples
{
    import ...;
    import ...;

    public class SampleAnalyzeText
    {
        public static void SampleAnalyzeText(...)
        {
            ...
        }
        // [END region_tag]

        public static void Main()
        {
            // process command line arguments

            SampleAnalyzeText(param1, param2);
        }
    }
}
```
Start tag and end tag are not at the same level, which will cause samples on devsite to miss closing braces.

Found in: #2872

hzyi-google:
Fix #2873

### DIFF
--- a/src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
+++ b/src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
@@ -44,8 +44,9 @@ namespace Google.Example.Library.V1.Samples
             // Shelf deleted
             Console.WriteLine("Shelf deleted.");
         }
-        // [END sample]
     }
+
+    // [END sample]
 
     public class Main {
         public static void Main(string[] args)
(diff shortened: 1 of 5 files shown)


src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
```diff
@@ -44,11 +44,14 @@ namespace Google.Example.Library.V1.Samples
             // Shelf deleted
             Console.WriteLine("Shelf deleted.");
         }
-        // [END sample]
+    }
+
+    // [END sample]
 
+    public class DeleteShelfAsyncFlattenedAsyncEmptyResponseTypeWithResponseHandlingMain {
```
chrisdunelm: The naming is a little crazy!

hzyi-google: Yes! That's to test all samples have unique class names when users happen to provide the same name for two samples. It should never happen in production :)

PR #2888, opened 2019-07-25
node: make the generated Node.js libraries work with grpc-fallback and browser

### LINKED ISSUE #2835: Extraneous protobuf require breaking builds
We have an issue over here where a bonus protofbuf require is breaking eslint:
https://github.com/googleapis/nodejs-talent/pull/69

alexander-fenster:
Fixes #2835 and a whole lot of random client libraries issues where folks try to use our libraries in weird environments.

**Serious (though not breaking) changes to the client library code!**

In this pull request, the following changes are introduced to Node.js library generation:

1. `.proto` files are now compiled statically into JSON (using the `compileProtos` script introduced in https://github.com/googleapis/gax-nodejs/pull/547). This allows us to stop parsing `.proto` files during the runtime, and also allows us to bundle the generated client library for browser. The list of proto files to be compiled is dumped by a generator into a JSON file (one file per service) so that the `compileProtos` script knew which proto files to compile. In the real life usage, `artman` or the upcoming micro-generator will call `compileProtos` (we cannot do it from here because we don't have a Node.js environment). Right now, it's taken care of in `artman`.

2. The generated client library how has a special entry point for a browser, `src/browser.js`, which changes the behavior of `.proto` files loading (in browser's `google-gax` version there is no filesystem access, hence the protos are passed to `google-gax` as JSON objects, while in Node.js `@grpc/proto-loader` requires them to be files).

3. `protobufjs` is now exported from `google-gax` (https://github.com/googleapis/gax-nodejs/pull/544), so there is no need to have it as a dependency of the client library.

4. A gRPC-fallback mode for Node.js is introduced: by passing `fallback: true` as an option for the client constructor, the users asks to use the fallback version of `google-gax`.

This is a generator-related part of the grpc-fallback internship project by @ramya-ramalingam and @LibanOdowa.


alexander-fenster:
This actually works - needs to release gax+browser and then it will be ready to takeoff. The only thing I want to wait for here is a small change to gax to make grpc-fallback usable in Node.js, then we'll be able to have an option for the client that enables or disables fallback mode regardless if we're in browser or not.

alexander-fenster:
Please don't merge until 8/19 (when I get back from vacation).

alexander-fenster:
@lukesneeringer This is the PR I was asking you to review. It makes some changes under the hood: instead of loading `.proto` files on the fly, it now uses the JSON representation of proto files (compiled by `pbjs` from `protobufjs`), it allows gRPC fallback usage for both browser (automatically with `webpack`) and Node.js (by passing an option), and it also removes `protobufjs` from dependencies (since `google-gax` depends on `protobufjs`, we can just reuse it from there).

To get an idea of how different client (regular, gRPC fallback, or browser) work, take a look at the showcase test code in `showcase/nodejs/index.js` (here is its output in [CircleCI](https://circleci.com/gh/googleapis/gapic-generator/36802)), see `Run Showcase integration tests`). For browser tests, I use `karma` test runner and a headless Chromium.

PR #2894, opened 2019-07-27
SampleGen: csproj for Sample package

hzyi-google:
- Generates a `.csproj` file for `Google.Cloud.Api.V1.Samples` package, which references the gapic project and `CommandLine` package.

### DIFF
--- a/src/main/java/com/google/api/codegen/transformer/csharp/CSharpBasicPackageTransformer.java
+++ b/src/main/java/com/google/api/codegen/transformer/csharp/CSharpBasicPackageTransformer.java
@@ -41,7 +41,8 @@ public class CSharpBasicPackageTransformer implements ModelToViewTransformer<Pro
       "csharp/gapic_snippets_csproj.snip";
   private static final String UNITTEST_CSPROJ_TEMPLATE_FILENAME =
       "csharp/gapic_unittest_csproj.snip";
-  private static final String SAMPLE_CSPROJ_TEMPLATE_FILENAME = "csharp/gapic_samples_csproj.snip";
+  private static final String SAMPLE_CSPROJ_TEMPLATE_FILENAME =
+      "csharp/gapic_samples_csproj.snip";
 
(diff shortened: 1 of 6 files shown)


hzyi-google:
@chrisdunelm @googleapis/samplegen 

I wasn't able to build the samples project emitted by artman directly due to the following error message: 

`CSC : error CS2017: Cannot specify /main if building a module or library`

However, when I copy the entire `Samples` directory to a local clone of `google-cloud-dotnet`, everything works fine.

Is there anything else done during the packaging process?

hzyi-google:
@munkhuushmgl

src/main/java/com/google/api/codegen/transformer/csharp/CSharpBasicPackageTransformer.java
```diff
@@ -41,6 +41,7 @@
       "csharp/gapic_snippets_csproj.snip";
   private static final String UNITTEST_CSPROJ_TEMPLATE_FILENAME =
       "csharp/gapic_unittest_csproj.snip";
+  private static final String SAMPLE_CSPROJ_TEMPLATE_FILENAME = "csharp/gapic_samples_csproj.snip";
```
vchudnov-g: nit: for consistency, have this line and the previous formatted similarly

hzyi-google:
So C# generates project files [here](https://github.com/googleapis/google-cloud-dotnet/blob/abbbf43d00bec4335f091c497a47ca4ee252a675/tools/Google.Cloud.Tools.ProjectGenerator/Program.cs#L429), and therefore this generated file will at most be used for our own testing. I'll merge it and submit a separate PR to google-cloud-dotnet that will actually be generating the csproj file for Sample package.

src/main/java/com/google/api/codegen/transformer/csharp/CSharpBasicPackageTransformer.java
hzyi-google: Nope, formatter doesn't allow me to do that (so annoying!)

PR #2900, opened 2019-07-31
py,go,java: remove send/recv msg size limit

### LINKED ISSUE #2895: Add config option for maxInboundMsgSize
For some client we need to set maxInboundMsgSize, right now we are using synth to update code after generation. Can it be added as part of gapic config? Please see this [comment](https://github.com/googleapis/google-cloud-java/pull/5883#pullrequestreview-267294991)

noahdietz:
Fixes #2895 for python, Go and Java

Node.js (grpc-js) does not check size limits, therefore it is already unbounded.

### DIFF
--- a/src/main/resources/com/google/api/codegen/py/transport.snip
+++ b/src/main/resources/com/google/api/codegen/py/transport.snip
@@ -50,6 +50,10 @@
                 channel = self.create_channel(
                     address=address,
                     credentials=credentials,
+                    options={
+                        'grpc.max_send_message_length': -1,
+                        'grpc.max_receive_message_length': -1,
+                    },
                 )
 
             self._channel = channel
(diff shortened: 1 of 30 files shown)


hzyi-google:
Java doesn't seem to support `maxOutboundMessageSize` :(

noahdietz:
Oops! I assumed :man_facepalming: Fixed!

noahdietz:
The latter changes to the Go snippets were essentially copied from https://github.com/googleapis/gapic-generator-go/pull/165

busunkim96:
I opened https://github.com/grpc/grpc/pull/19810 to see if the restriction could be lifted in gRPC.

src/main/resources/com/google/api/codegen/py/transport.snip
```diff
@@ -50,6 +50,10 @@
                 channel = self.create_channel(
                     address=address,
                     credentials=credentials,
+                    options={
+                        'grpc.max_send_message_length': -1,
+                        'grpc.max_receive_message_length': -1,
+                    },
```
tseaver: This line needs to have `.items()` tacked on (`options` is expected to be a sequence of two-tuples, not a dict).

PR #2901, opened 2019-07-31
SampleGen: print a descriptive line in Java LRO samples

### LINKED ISSUE #2898: Java Generated sample doesn't print waiting for output
@googleapis/samplegen 

Generating samples in both Python and Java for asynchronous tasks. 

Python will generate a line: `print(u'Waiting for operation to complete...')`
Where Java doesn't, would be nice to include to let the user know something is happening.

hzyi-google:
Fix #2898


src/main/resources/com/google/api/codegen/java/standalone_sample.snip
```diff
@@ -176,12 +176,15 @@
 
 @private asyncOperationMethodSampleCode(apiMethod, sample)
   {@initCode(sample.sampleInitCode)}
+  OperationFuture<{@apiMethod.responseTypeName}, {@apiMethod.operationMethod.metadataTypeName}> \
+    future = {@apiMethod.apiVariableName}.{@apiMethod.name}(\
```
nnegrey: @beccasaurus, for the variable name `future`, I feel I typically see `operation` used more. Thoughts?

hzyi-google: A quick search in java-docs-samples seems to give me more `future` than `operation` for LRO methods of Gapic clients.

nnegrey: Many uses have been abbreviated to just `op` or some use `response` (which I think we shouldn't use.

beccasaurus: The GAPIC repo seems to use `future` almost exclusively.

Looking at java-docs-samples, I see op, operation, future, response

Let's not do response or op

Operation vs Future?

Well, it's an instance of `OperationFuture` lol. Both!

I could go either way, I'd say `future` because it's consistently used in the library. But `operation` would be fine and consistent with most languages!

hzyi-google: @nnegrey @beccasaurus 
If y'all agree with `future` can either of you approve this PR?
If there is still a strong preference of `operation` please let me know too :)

PR #2907, opened 2019-08-05
SampleGen: Generate samples from sample configs in Java and C#

hzyi-google:


src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_samplegen_config_migration_library.baseline
```diff
@@ -15,7 +15,7 @@
 
 // Generated code. DO NOT EDIT!
 
-// This is a generated sample ("Request", "empty_response_type_with_response_handling")
+// This is a generated sample ("Request", "delete_shelf_request_empty_response_type_with_response_handling")
```
hzyi-google: Again, there is no sample value set ids any more, so we use sample ids.

hzyi-google:
cc/ @googleapis/samplegen

src/main/java/com/google/api/codegen/transformer/csharp/CSharpStandaloneSampleTransformer.java
```diff
@@ -56,85 +80,22 @@
...
-
-  @Override
-  public List<String> getTemplateFileNames() {
-    return Collections.singletonList(STANDALONE_SAMPLE_TEMPLATE_FILENAME);
+    super(
+        STANDALONE_SAMPLE_TEMPLATE_FILENAME,
+        pathMapper,
+        fileHeaderTransformer,
+        csharpApiMethodTransformer,
+        p -> new CSharpFeatureConfig(),
```
vchudnov-g: `p` is not very informative. How about `pkg` or `pkgName`?

src/main/java/com/google/api/codegen/transformer/java/JavaApiMethodTransformer.java
```diff
@@ -0,0 +1,260 @@
+/* Copyright 2018 Google LLC
```
vchudnov-g: 2019

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSamplesTransformer.java
```diff
@@ -14,150 +14,68 @@
...
+import com.google.api.codegen.viewmodel.ClientMethodType;
+import com.google.common.collect.ImmutableMap;
 
-/**
- * A transformer to generate Java standalone samples for each method in the GAPIC surface generated
- * from the same ApiModel.
- */
-public class JavaGapicSamplesTransformer implements ModelToViewTransformer<ProtoApiModel> {
-  private final GapicCodePathMapper pathMapper;
+/** A transformer that generates C# standalone samples. */
```
vchudnov-g: s/C#/Java/

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSamplesTransformer.java
```diff
@@ -14,150 +14,68 @@
...
-    }
-
-    return surfaceDocs;
+  public JavaGapicSamplesTransformer(GapicCodePathMapper pathMapper) {
+    super(
+        STANDALONE_SAMPLE_TEMPLATE_FILENAME,
+        pathMapper,
+        fileHeaderTransformer,
+        apiMethodTransformer,
+        p -> JavaFeatureConfig.create(p),
```
vchudnov-g: `pkg`

src/main/java/com/google/api/codegen/viewmodel/CallingForm.java
```diff
@@ -132,6 +133,42 @@ static RpcType fromMethodContext(MethodContext context) {
   private static final Table<TargetLanguage, RpcType, ImmutableList<CallingForm>>
       CALLING_FORM_TABLE =
           ImmutableTable.<TargetLanguage, RpcType, ImmutableList<CallingForm>>builder()
+              .put(
```
vchudnov-g: What's the relationship between this table and the `CALLING_FORM_CLIENT_METHOD_TYPE_MAP` you have in the other files above? This one maps RPC types to CallingForms, the other one maps ClientMethodTypes to CallingForms. Why do we need both? Is this something we could simplify?

src/main/java/com/google/api/codegen/java/JavaGapicSamplePathMapper.java
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  // TODO(hzyi): Depreate this method.
+  // The method was originally added to the interface `GapicCodePathMapper` because `method` was
+  // needed. However, the method name is no longer needed in any of the seven languages we support,
+  // and therefore we can reuse `getOutputPath` to calculate output path for sample generation.
+  // This means we cannot use one instance of `GapicCodePathMapper` for both codegen and samplegen,
+  // but that is fine.
+  public String getSamplesOutputPath(String elementFullName, ProductConfig config, String method) {
+    throw new UnsupportedOperationException("Deprecated: use getOutputPath instead.");
```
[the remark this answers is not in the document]
hzyi-google: Oh because this class implements `GapicCodePathMapper` which has a `getSampleOutputPath`. We'll need to remove this method in the interface and all of it's subclasses all at once, which might be too many changes for this PR :)

src/main/java/com/google/api/codegen/transformer/SampleFileRegistry.java
```diff
@@ -111,6 +111,10 @@ public String getSampleClassName(MethodSampleView sample, String method) {
     Preconditions.checkState(
         userProvidedIdCount.get(regionTag) != null && userProvidedIdCount.get(regionTag) > 0,
         "Sample not registered.");
+    // method names can be in snake_case, camelCase or ParscalCase depending on the language
```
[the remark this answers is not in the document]
hzyi-google: We use `UpperCamel`. I can change it over.

src/main/java/com/google/api/codegen/transformer/StaticLangGapicSamplesTransformer.java
```diff
@@ -0,0 +1,278 @@
...
+    return sampleFileViews.build();
+  }
+
+  private List<ViewModel> generateSamplesFromSampleConfigs(
+      List<InterfaceContext> interfaceContexts,
+      GapicProductConfig productConfig,
+      SurfaceNamer namer,
+      ImmutableTable<String, String, ImmutableList<SampleConfig>> sampleConfigTable) {
+
+    // Loop through sample configs and and map each sample id to its matching calling forms.
```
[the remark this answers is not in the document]
hzyi-google: Ack (but I didn't get the joke ;-/)

src/main/java/com/google/api/codegen/viewmodel/CallingForm.java
hzyi-google: `RpcType` is the proto/grpc layer and contains six values: `Unary`, `PagedStreaming`, `LongRunning`, `GrpcClientStreaming`, `GrpcServerStreaming`, `GrpcBidiStreaming`.

For Java and C#, one RPC method would have multiple overloads, and the type of the overload is determined by `ClientMethodType`. For example, a unary RPC would have at least three public method in Java client libraries: `RequestObjectMethod`, `FlattenedMethod, CallableMethod`. For all other languages there aren't multiple overloads, but because we generate all of them in one generator they are forced to be compatible with Java and C#.

`CallingForm` has some overlaps with `ClientMethodType` but not completely equivalent. For example, For Node.js LROs, there is just one client method type: `LongRunningOptionalArrayMethod`, but we have multiple calling forms because there are multiple patterns to call an LRO method in Node.js. But for Java and C#, `ClientMethodType` and `CallingForm` are _mostly_ one to one.

I see this as an unavoidable problem of having one generator work for seven languages - the terminologies have to be general, but they don't always make much sense when it comes to a specific language :(

Making it less bad, the enum `RpcType` is a private enum inside `CallingForm` and thus invisible outside, which might make it less confusing for maintainers of this generator :/

PR #2908, opened 2019-08-07
C# fix flattened required wrapper types

chrisdunelm:


src/main/resources/com/google/api/codegen/csharp/gapic_client.snip
```diff
@@ -1173,7 +1174,11 @@
                     @if param.typeName == "string"
                         {@param.nameAsMethodName} = gax::GaxPreconditions.CheckNotNullOrEmpty({@param.name}, nameof({@param.name})),
                     @else
-                        {@param.nameAsMethodName} = gax::GaxPreconditions.CheckNotNull({@param.name}, nameof({@param.name})),
+                        @if param.typeName.endsWith("?")
+                            {@param.nameAsMethodName} = {@param.name} ?? throw new sys::ArgumentNullException(nameof({@param.name})),
```
jbolinger: Why does this make a difference? It looks like `gax::GaxPreconditions.CheckNotNull` does exactly the same thing?

chrisdunelm: The gax method has a `class` constraint on the type parameter. In this case the type is a nullable primitive, which is not a class (it's a struct).
We could (and probably will) add a new method to `GaxPreconditions` to allow this test on a struct, but that'll take time. This can be changed sometime in the future, as it won't cause a breaking change.

PR #2924, opened 2019-08-13
SampleGen: Sanity checks to sample configs before parsing them into protos

hzyi-google:
- If file does not exist: throw a `RuntimeException`
- If `type` or `config_schema_version` mismatch: ignore


hzyi-google:
PTAL

src/test/java/com/google/api/codegen/testsrc/common/samples/fake.sample.yaml
```diff
@@ -0,0 +1,2 @@
+---
+type: this.is.not.a.sample.config
```
beccasaurus: Winning.

src/main/java/com/google/api/codegen/util/SampleConfigSanitizer.java
```diff
@@ -0,0 +1,57 @@
...
+import org.yaml.snakeyaml.Yaml;
+
+public class SampleConfigSanitizer {
+
+  private SampleConfigSanitizer() {
+    // utility class
+  }
+
+  private static final String TYPE_FIELD_NAME = "type";
+  private static final String SCHEMA_VERSION_FIELD_NAME = "config_schema_version";
```
vchudnov-g: We resolved to use `schema_version` instead of `config_schema_version`. IIRC, we don't need `config_schema_version` for anything.

src/main/java/com/google/api/codegen/util/SampleConfigSanitizer.java
```diff
@@ -0,0 +1,57 @@
...
+  public static List<String> sanitize(List<String> sampleConfigFileNames) {
+    return sampleConfigFileNames
+        .stream()
+        .filter(SampleConfigSanitizer::isValidSampleConfig)
+        .collect(Collectors.toList());
+  }
+
+  @SuppressWarnings("unchecked")
+  private static boolean isValidSampleConfig(String sampleConfigFileName) {
+    try (Reader reader = new FileReader(sampleConfigFileName)) {
```
vchudnov-g: A YAML file can have multiple YAML docs, so you should iterate over those.

src/main/java/com/google/api/codegen/util/SampleConfigSanitizer.java
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  private static final String TYPE_FIELD_NAME = "type";
+  private static final String SCHEMA_VERSION_FIELD_NAME = "config_schema_version";
+  private static final String VALID_TYPE_VALUE =
+      "com.google.api.codegen.samplegen.v1p2.SampleConfigProto";
+  private static final String VALID_SCHEMA_VERSION = "1.2.0";
+
+  // Filter out sample configure files without proper type definition.
+  public static List<String> sanitize(List<String> sampleConfigFileNames) {
```
vchudnov-g: As per [this comment thread](https://docs.google.com/a/google.com/document/d/1g6M5gl2gdrefQGyCW_MHAqZPRBiobCNJ4XChJJ4ufq4/edit?disco=AAAADYbhLxU) the semantics should be that 

* if files are explicitly specified, each such file must be a yaml file that contains at least one sample config doc (otherwise fail)
* if directories are specified, in each such directory we ignore non-yaml files and files that don't include any sample config docs
* across all the sample files and directories specified, we must have obtained at least one sample config doc

Thus, 
* specifying just `--samples=/tmp/non-sample.yaml` will fail
* if we were to support globbing (which is not in the µ-samplegen spec and we should not support it until it is): if `/tmp/foo/`  has both sample-configs and non-sample-config YAML files, specifying `--samples=/tmp/foo/*yaml` will fail, but `--samples=/tmp/foo/` will pass

FYI @software-dov

src/main/java/com/google/api/codegen/util/SampleConfigSanitizer.java
hzyi-google: test samples in `gapic-generator` use `config_schema_version`...?

src/main/java/com/google/api/codegen/util/SampleConfigSanitizer.java
hzyi-google: That's done in artman. artman resolves the input and find all the yaml files and pass that to gapic-generator.

hzyi-google: Also that's for micro-generators.

src/main/java/com/google/api/codegen/util/SampleConfigSanitizer.java
vchudnov-g: At some point there was a discussion that we should just use `schema_version`. Let's look for `schema_version` first, and if that is not present, we look for `config_schema_version`. Users should be encouraged to use `schema_version`. Once all the current configs use this, we can remove support for `config_schema_version`.

src/main/java/com/google/api/codegen/util/SampleConfigSanitizer.java
vchudnov-g: Hmmm... that's OK for now. When artman goes away, we'll need to incorporate that logic into gapic-generator, since that's part of the µgenerator specs. Could you file a FR for this?

src/main/java/com/google/api/codegen/util/SampleConfigSanitizer.java
hzyi-google: I'd rather change all to `schema_version` all at once. It'd be confusing to have two.

src/main/java/com/google/api/codegen/util/SampleConfigSanitizer.java
hzyi-google: I don't think the monolith need to support this, because the CLI would be so different anyway.
However, I do agree that it should be supported when monolith works as a protoc Java plugin.

src/main/java/com/google/api/codegen/util/SampleConfigSanitizer.java
hzyi-google: Added check for each doc in a file. If any of the docs in one file does not have valid `type` or `schema_version`. This is the best effort for now -- the `YamlReader` in api-compiler doesn't support multi-doc yaml files anyway.

(Honestly I don't think supporting multi-doc yamls is necessary, considering there are two top-level fields: `schema_version` and `type` whose values don't change that much. User can already configure any number of samples in one file)

hzyi-google:
PTAL

src/main/java/com/google/api/codegen/util/SampleConfigSanitizer.java
vchudnov-g: Yes, that's what I had in mind. In the protoc world, this is the responsibility of the generators, not  bazel or anything else. Could you file the FR and link it here?

src/main/java/com/google/api/codegen/util/SampleConfigSanitizer.java
vchudnov-g: Speaking to Rebecca yesterday, it seems we do not have to worry about backwards compatibility yet. Yay! So let's just use `schema_version` for now.

src/main/java/com/google/api/codegen/util/SampleConfigSanitizer.java
vchudnov-g: Where do you check the third requirement?

src/main/java/com/google/api/codegen/util/SampleConfigSanitizer.java
```diff
@@ -0,0 +1,62 @@
...
+
+import java.io.FileReader;
+import java.io.IOException;
+import java.io.Reader;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import org.yaml.snakeyaml.Yaml;
+
+public class SampleConfigSanitizer {
```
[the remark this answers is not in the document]
hzyi-google: I originally went with `Reader`, but decided to change the name because this class does not really do the yaml reading (`MultiYamlReader` still does that, though it's not doing quite good of a job but that's still the truth for now). This class just check the input and filter out whatever is not recognized as a sample config file, thus the name `Sanitizer`. 

I want to ultimately get rid of `MultiYamlReader` for samples and use this class to read the sample configs, and probably we can rename the class when this is done.

src/main/java/com/google/api/codegen/util/SampleConfigSanitizer.java
hzyi-google: For now for both cases the generator will emit no samples.

hzyi-google: issue: https://github.com/googleapis/gapic-generator/issues/2930

src/main/java/com/google/api/codegen/util/SampleConfigSanitizer.java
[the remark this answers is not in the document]
vchudnov-g: Even if we keep `MultiYamlReader` (I'm curious why you want to replace it), I think `Sanitizer` is not an informative name. I guess you're right in that this returns the file names, not contents, so maybe something with "Filter" in the name. But whatevs, it's a nit we can worry about separately.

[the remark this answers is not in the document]
hzyi-google: Because `MultiYamlReader` uses `YamlReader` from `api-compiler`, which doesn't support reading yamls with anchors.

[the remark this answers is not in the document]
vchudnov-g: Excellent reason!

Issue #2928, opened 2019-08-14
Python Samples style change


hzyi-google:
- [ ] Remove the calling_form sample ID combination in the `DO NOT EDIT` warning
- [x] Imports should have an empty line above them from the comment
- [x] Remove the blank line between region tag and first import
- [x] Comment blocks shouldn't have a newline at the end
- [x] Remove unused `import sys`

cc/ @nnegrey Please let me know if I missed a thing or two :stuck_out_tongue_closed_eyes:

vchudnov-g:
We haven't decided to do the first item yet. I'm not opposed, but let's discuss metadata in general and where we want it to live.

hzyi-google:
2 - 5 are fixed. We don't have any immediate work item for 1. Close this issue.

PR #2927, opened 2019-08-13
SampleGen: Remove newline at the end of a comment block

hzyi-google:
Fix #2578

PR #2935, opened 2019-08-15
SampleGen: Python style changes

hzyi-google:
Fix 2, 3, 4, 5 of #2928 

1 is not touched because it is not clear whether 1 should be fixed from the changelist comments (i.e., whether the combination of region tag and calling form should be kept or deleted).


hzyi-google:
PTAL

vchudnov-g:
4 is fixed by #2927

PR #2941, opened 2019-08-19
SampleGen: blankline at the end of a for loop block

hzyi-google:


hzyi-google:
@googleapis/samplegen

PR #2942, opened 2019-08-19
SampleGen: Add a sample test case that exposes bugs when constructing map fields

hzyi-google:
This PR exposes bugs in (at least) 3 languages when constructing map fields but doesn't actually fix any of them :(

- In Java: the key should be boxed type (`Integer`)
- In C#: the format is broken
- In Python: we need to use objects instead of dictionaries when used as a map value (to be verified)
- In Node.js: there seems to be an extra comma? 

Issue tracking the bugs: https://github.com/googleapis/gapic-generator/issues/2943


hzyi-google:
cc/ @googleapis/samplegen

hzyi-google:
The link is in PR description.

I initially wanted to just fix the Python bug which was reported by DPEs, but during fixing them the bugs for other languages uncover themselves. After struggling with C# format for a while I figured that I wouldn't be able to fix all of them in one PR... So decided to do this.

vchudnov-g:
Oh, cool! :+1:

PR #2944, opened 2019-08-21
Add py_gapic_library Bazel rule

eoogbe:
Usage:

```bzl
py_gapic_library(
    name = "language_py_gapic",
    src = ":language_proto_with_info",
    gapic_yaml = "language_gapic.yaml",
    service_yaml = "//google/cloud/language:language_v1.yaml",
)
```


rules_gapic/python/py_gapic.bzl
```diff
@@ -0,0 +1,152 @@
...
+    output_dir_path = "%s/%s" % (output_main.dirname, output_dir_name)
+
+    # Note the script is more complicated than it intuitively should be because of limitations
+    # inherent to bazel execution environment: no absolute paths allowed, the generated artifacts
+    # must ensure uniqueness within a build.
+    script = """
+    unzip -q {gapic_srcjar} -d {output_dir_path}
+    {formatter} -q {output_dir_path}
+    pushd {output_dir_path}
+    rm nox.py setup.py
```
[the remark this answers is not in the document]
eoogbe: Done

rules_gapic/python/py_gapic.bzl
```diff
@@ -0,0 +1,152 @@
...
+    main_file = ":%s.srcjar" % srcjar_name
+    main_dir = "%s_main" % srcjar_name
+
+    unzipped_srcjar(
+        name = main_dir,
+        srcjar = main_file,
+        extension = ".py",
+    )
+
+    native.py_library(
```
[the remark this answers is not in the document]
eoogbe: It works

rules_gapic/python/py_gapic.bzl
```diff
@@ -0,0 +1,170 @@
...
+        "formatter": attr.label(
+            doc = "Formats the output to a Python-idiomatic style.",
+            default = Label("@pypi_black//:black"),
+            executable = True,
+            cfg = "host",
+        ),
+    },
+    outputs = {
+      "main": "%{name}.srcjar",
+      "test": "%{name}-test.srcjar",
```
vam-google: Sorry I should have mentioned it in first pass (when asked to add pkg). What about the `smoke-test` part? Not all clients have it, but those which do, usually require special handling. 
The reason why we want to separate unit tests and smoke tests is that we want to run unit tests but not smoke tests, because those require special environment setup and in general much harder to control than unit tests.

You can use pubsub or some other complicated clients for an example of a client which has smoke tests configured.

rules_gapic/python/py_gapic.bzl
```diff
@@ -0,0 +1,170 @@
...
+    output_dir_path = "%s/%s" % (output_main.dirname, output_dir_name)
+
+    # Note the script is more complicated than it intuitively should be because of limitations
+    # inherent to bazel execution environment: no absolute paths allowed, the generated artifacts
+    # must ensure uniqueness within a build.
+    script = """
+    unzip -q {gapic_srcjar} -d {output_dir_path}
+    {formatter} -q {output_dir_path}
+    pushd {output_dir_path}
+    zip -q -r {output_dir_name}-pkg.srcjar nox.py setup.py setup.cfg docs MANIFEST.in README.rst LICENSE
```
vam-google: minor: please consider using wildcards for things like this (something like `*.py` instead of `nox.py setup.py`). 

The reason for this is because usually it is easier to keep up with changes form generator (if they add a new file there is a lower chance that we will need to update the rule as well). The flipside of it is that we may mistakenly grab files, which should not be grabbed, but it is less likely (usually we want everything).

rules_gapic/python/py_gapic.bzl
eoogbe: Done

rules_gapic/python/py_gapic.bzl
eoogbe: With wildcards, it grabs the wrong files

PR #2946, opened 2019-08-21
SampleGen: Correctly initialize map objects in Java and Node.js

hzyi-google:
- Java: use boxed types
- Node.js: remove the comma when there is just one entry


hzyi-google:
Java generic types (those inside `<>`) cannot be primitive types.

PR #2951, opened 2019-08-27
bazel: make grpc_service_config attr a label

noahdietz:
After implementing https://github.com/googleapis/gapic-generator-go/pull/181 I think the attribute I added for `grpc_service_config` support was incorrect. This PR should fix it.


noahdietz:
> Can you please clarify that running it works?

I verified that it works. Of course in order to do that, I had to plumb the config file through the `go_gapic_library` macro to `gapic_srcjar`, so I included that change here.

noahdietz:
@vam-google if you get chance can you do another once-over on this PR since I added kwargs to the go_gapic_library && gapic_srcjar ?

PR #2964, opened 2019-09-01
node: minor fixes to Node.js generator

alexander-fenster:
The generated list of protos for Node.js does not have a newline at the end of file. Fixing that.

Also, adding license header to the generated `webpack.config.js`.


hzyi-google:
Generator changes LGTM.

PR #2965, opened 2019-09-03
SampleGen: Set up samplegen baseline tests against showcase

hzyi-google:
- Add a new class `GapicCodegenTestConfig` with builders to replace the helper method with 7 parameters that creates test configs
- Set up baseline tests for samplegen against showcase proto source
  - only one simple sample config
  - only in Java
  - rely solely on proto annotations, not gapic config

- Next:
  - set up CI to run this sample via sample-tester
  - more languages/features to test


hzyi-google:
Please hold on reviewing... I think I broke bazel. I'll get to it soon.

PR #2966, opened 2019-09-03
SampleGen: Correctly parse initial values 

### LINKED ISSUE #2961: SampleGen: sample metadata usage rendering wrong with "[String Values in Brackets]"
### Steps to Reproduce
 - Set the value of any input_parameter to a string wrapped in brackets  
   > e.g. `value: "[Google Cloud Project ID]"`
 - Generate code samples (tested with Python)

### Expected Result
 - Usage metadata should look like this:  
   ```yaml
   usage: python3 samples/v3beta1/translate_foo.py [--project "[Google Cloud Project ID]"]
   ```

### Actual Result
 - The first bracket is missing!!  
   ```yaml
   usage: python3 samples/v3beta1/translate_foo.py [--project "Google Cloud Project ID]"]
   ```

I have tools which treat [bracket wrapped] values in a special way (I've been using this as a convention for required values)

@googleapis/samplegen 

----

Here's a Translation v3beta1 code sample for repro:

```yaml
type: com.google.api.codegen.samplegen.v1p2.SampleConfigProto
schema_version: 1.2.0
samples:
- region_tag: translate_list_language_names_beta
  title: List Supported Language Names and Codes
  description: Listing supported languages with target language name
  service: google.cloud.translation.v3beta1.TranslationService
  rpc: GetSupportedLanguages
  request:
  - field: display_language_code
    value: en
    comment: |
      The language to use to return localized, human readable names of supported languages.
    input_parameter: language_code
  - field: parent%project
    value: "[Google Cloud Project ID]"
    input_parameter: project
  - field: parent%location
    value: global
  response:
  - comment:
    - List language codes of supported languages
  - loop:
      collection: "$resp.languages"
      variable: language
      body:
      - print:
        - "Language Code: %s"
        - language.language_code
      - print:
        - "Display Name: %s"
        - language.display_name
```

hzyi-google:
Fix #2961

### DIFF
--- a/src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
+++ b/src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
@@ -264,11 +264,28 @@ public class FieldStructureParser {
         return tokenStr;
       }
     }
+
+    // Scanner parsed more than one token. Prepend the parse text to the rest of value and return
+    if (token == Scanner.IDENT || token == Scanner.INT) {
+      return scanner.tokenStr() + scanner.input().substring(scanner.pos());
+    }
+
+    // There can't be nothing on the RHS of '='
+    if (token == Scanner.EOF) {
+      throw new IllegalStateException("unexpected EOF when parsing values");
+    }
+
+    // Empty object literal
     if (token == '{') {
       if (scanner.scan() == '}') {
         return EMPTY_OBJECT;
       }
     }
-    return scanner.tokenStr() + scanner.input().substring(scanner.pos());
+
+    // Scanner parsed one token. Prepend it to the rest of value and return
+    return new StringBuilder()
+        .append((char) token)
+        .append(scanner.input().substring(scanner.pos()))
+        .toString();
   }
 }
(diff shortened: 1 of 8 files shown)


src/main/java/com/google/api/codegen/metacode/FieldStructureParser.java
```diff
@@ -264,11 +264,28 @@ private static String parseValue(Scanner scanner) {
         return tokenStr;
       }
     }
+
+    // Scanner parsed more than one token. Prepend the parse text to the rest of value and return
+    if (token == Scanner.IDENT || token == Scanner.INT) {
+      return scanner.tokenStr() + scanner.input().substring(scanner.pos());
+    }
+
+    // There can't be nothing on the RHS of '='
```
vchudnov-g: s/nothing/anything/

src/test/java/com/google/api/codegen/testsrc/common/samples/publish_series.sample.yaml
```diff
@@ -11,7 +11,7 @@ samples:
   region_tag: "canonical"
   request:
   - field: shelf.name
-    value: Math
+    value: "[Math]"
```
vchudnov-g: I was confused by the brackets until I saw the config changed. I might have gone for the uuid field to make it less surprising, but no biggie. <shrug/>

beccasaurus: +1 that makes sense from a real user perspective ^ (uuid)

hzyi-google: The uuid is not handled by the same chunk of code though... `value` is the only thing that hates `[`.

Issue #2972, opened 2019-09-06
gapic-generator does not understand projects


lukesneeringer:
Many GCP APIs have the (shared) concept of a _project_, which is at the top of the hierarchy for most GCP resources.

Accordingly, the GAPIC config often has a block that looks like this:

```yaml
  collections:
  - entity_name: project
    language_overrides:
    - language: csharp
      common_resource_name: Google.Api.Gax.ResourceNames.ProjectName
```

Generating C# with this block crashes with the error described here: https://github.com/googleapis/google-cloud-dotnet/issues/3423

Generating C# with this block removed (which would be a problem for them; the override matters) crashes with a different error:

```
ERROR: toplevel: Reference to unknown type "cloudresourcemanager.googleapis.com/Project" on field google.cloud.phishingprotection.v1beta1.ReportPhishingRequest.google.cloud.phishingprotection.v1beta1.ReportPhishingRequest.parent
```

**P0:** We _either_ need generating with this YAML block to work correctly in all languages, _or_ we need to be able to remove the block completely and have the generator do the correct thing whenever it encounters a project (if choosing between them, removal is better,  but the P0 is to get this working)...

**P1:** ...or org, folder, or billing account.

lukesneeringer:
Fixed by #2979.

PR #2973, opened 2019-09-07
[Do not merge] initial commit to add project as a well-known resource name config

hzyi-google:
Fix https://github.com/googleapis/gapic-generator/issues/2972

Please do not submit for now because this PR contains a bunch of hacks which I intend to clean up/add more tests when I have more time to get a better understanding about resource name configurations in both v1 and v2 gapic config.

This PR shall make generating GAPICs for all languages but Java for APIs annotated with `cloudresourcemanager.googleapis.com/Project` work.

The reason why Java gapic generation won't work is that Java resource name classes are generated by a separate plugin. We need to update that to understand the `cloudresourcemanager.googleapis.com/Project` annotation to unblock Java.

Some more thoughts: for one second I believed that add annotations to the actual message of `Project` is the better solution, but I realized it isn't going to work: the message `Project` will most likely live in the API of `cloudresourcemanager`, and the generated Java/C# resource name classes (`ProjectName`) will have its package/namespace derived from the proto package of the `cloudresourcemanager` API, rather than the actual API that we are generating GAPICs from. This would have been a backward-incompatible change.

### DIFF
--- a/src/main/java/com/google/api/codegen/config/GapicProductConfig.java
+++ b/src/main/java/com/google/api/codegen/config/GapicProductConfig.java
@@ -195,7 +195,9 @@ public abstract class GapicProductConfig implements ProductConfig {
     }
 
     List<ProtoFile> sourceProtos =
-        model.getFiles().stream()
+        model
+            .getFiles()
+            .stream()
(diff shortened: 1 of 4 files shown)


hzyi-google:
cc/ @vam-google @lukesneeringer FYI

PR #2979, opened 2019-09-16
Fix Resource Names annotations

vam-google:
This is WIP

Add support for common resource names.
This does not fix all of the existing problems with the resource names.

Related fixes for `artman` and `protoc-java-resource-names-plugin` are pending.

**Config changes**
[library.proto](https://github.com/googleapis/gapic-generator/pull/2979/files#diff-a44629aaf921c03fdf47901df114cebf)  - relevant for gapic v1 and gapic v2 test cases
[library_gapic.yaml](https://github.com/googleapis/gapic-generator/pull/2979/files#diff-9f9ffb93516a846c5e39b7afc9e0c5de
) - relevant for gapic v1 test cases
[library_v2_gapic.yaml](https://github.com/googleapis/gapic-generator/pull/2979/files#diff-fc97e801b7222b7481527c795281e442) - relevant for gapic v2 test cases (basically just a fix of a bug)

The test cases try to do same in v1 and v2 worlds and check if the output produces same (or at least similar and binary compatible result).

The tests try to test:
- simple common resource name case (using `cloudresourcemanager.googleapis.com/Folder` as example):
  - referenced as `type` (`GetBookFromAnywhereRequest.folder`)
- common resource name containing the "calculated" parent in its pattern (using `locations.googleapis.com/Location` as an example) - notice **the difference** in generated code foe ach of the two cases below:
   - referenced as `chilt_type` (`GetBookFromArchiveRequest.parent`) - see `location_parent_oneof` in `library_gapic.yaml` for an analogous config in v1
   - referenced as `type` (`GetBookFromAnywhereRequest.place`) 
- language overrides (common_resource_name in C#) - (it was checked during real library generation, but no specific tests case added  here, will add soon)

**Output Changes**
The config changes produced output changes in the `<lang>_library.baseline` files for both of the sets of test cases (v1 and v2). Lang owners, please check that each of the `<lang>_library.baseline` file matches correctly the changes in the input (i.e. in library.proto and gapic yamls).


src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
```diff
@@ -6302,27 +6302,29 @@ namespace Google.Example.Library.V1.Snippets
         /// <summary>Snippet for GetBookFromArchiveAsync</summary>
         public async Task GetBookFromArchiveAsync()
         {
-            // Snippet: GetBookFromArchiveAsync(ArchivedBookName,CallSettings)
-            // Additional: GetBookFromArchiveAsync(ArchivedBookName,CancellationToken)
+            // Snippet: GetBookFromArchiveAsync(ArchivedBookName,LocationParentNameOneof,CallSettings)
```
lukesneeringer: Are these due to a change in the config, or a C# behavior change?

src/test/java/com/google/api/codegen/testsrc/common/common_resources.proto
```diff
@@ -0,0 +1,68 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This file contains stub messages for common resources in GCP.
+// It is not intended to be directly generated, and is instead used by
+// other tooling to be able to match common resource patterns.
+syntax = "proto3";
+
+package google.cloud;
```
noahdietz: This doesn't have any of the language pkg options. Are those necessary?

src/test/java/com/google/api/codegen/testsrc/common/common_resources.proto
```diff
@@ -0,0 +1,68 @@
...
+}
+
+message Location {
+  option (google.api.resource) = {
+    type: "locations.googleapis.com/Location"
+    pattern: "projects/{project}/locations/{location}"
+  };
+
+  string name = 1;
+}
```
hzyi-google: nit: newline please :)

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -807,14 +826,16 @@ private static GapicProductConfig createDummyInstance(
     // pattern field has been removed. Therefore, throw an error if a single resource config is
     // found that does not exist in an annotation.
     for (String key : singleResourceNameConfigsFromGapicConfig.keySet()) {
-      if (!annotationResourceNameConfigs.containsKey(key)) {
+      String annotationsStyleKey = Name.from(key).toUpperCamel();
```
hzyi-google: Can you add a comment here about why converting to upper case is needed?

src/main/java/com/google/api/codegen/transformer/java/JavaModelTypeNameConverter.java
```diff
@@ -273,14 +274,23 @@ private static String getResourceNamePackage(ResourceNameConfig resourceNameConf
...
       case FIXED:
       case SINGLE:
       case ONEOF:
-        return getJavaPackage(resourceNameConfig.getAssignedProtoFile());
+        ProtoFile assignedProtoFile = fieldConfig.getResourceNameConfig().getAssignedProtoFile();
+        FieldModel fm = fieldConfig.getField();
+        if (fm instanceof ProtoField) {
+          ProtoFile fieldProtoFile = ((ProtoField) fm).getProtoField().getFile();
+          if (!assignedProtoFile.getFullName().equals(fieldProtoFile.getFullName())) {
+            assignedProtoFile = fieldProtoFile;
```
hzyi-google: This seems to be using the package with code that references a resource instead of defines a resource to generate Java package for the resource name classes, which is absolutely correct. However, two questions:

1) should this always be used?
2) I _think_ the way `assignedProtoFile` is set up makes sure it is always inside the package of the API we are generating GAPICs for. Is that not the case?

hzyi-google:
Java generated code lgtm.

src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
vam-google: This is due to the change in gapic yaml and library.proto.

Relevant config changes:
[library.proto](https://github.com/googleapis/gapic-generator/pull/2979/files#diff-a44629aaf921c03fdf47901df114cebf)  - relevant for gapic v1 and gapic v2 test cases
[library_gapic.yaml](https://github.com/googleapis/gapic-generator/pull/2979/files#diff-9f9ffb93516a846c5e39b7afc9e0c5de
) - relevant for gapic v1 test cases
[library_v2_gapic.yaml](https://github.com/googleapis/gapic-generator/pull/2979/files#diff-fc97e801b7222b7481527c795281e442) - relevant for gapic v2 test cases (basically just a fix of a bug)

The test cases try to do same in v1 and v2 worlds and check if the output produces same (or at least similar and binary compatible result).

The tests try to test:
- simple common resource name case (using "cloudresourcemanager.googleapis.com/Folder" as example):
  - referenced as `type` (`GetBookFromAnywhereRequest.folder`)
- common resource name containing the "calculated" parent in its pattern (using `locations.googleapis.com/Location` as an example) - notice **the difference** in generated code foe ach of the two cases below:
   - referenced as `chilt_type` (`GetBookFromArchiveRequest.parent`) - see `location_parent_oneof` in `library_gapic.yaml` for an analogous config in v1
   - referenced as `type` (`GetBookFromAnywhereRequest.place`) 
- language overrides (common resource_name) - TBD

lukesneeringer: Okay, great. Thanks.

alexander-fenster:
What's up with Java CI?

vam-google:
@alexander-fenster it depends on generating the whole client via artman, and since there are pending changes from artman and resource names plugin I do not expect them to pass. But for this particular one, I'm investigating to make sure it is not a regression (and fix if it is).

src/main/java/com/google/api/codegen/transformer/java/JavaModelTypeNameConverter.java
vam-google: 1 & 2: If I understand your question correctly: I think `assignedProtoFile` cannot be used properly for our purpose (effectively that property should be removed, but I'm trying to keep this change as small as possible, since there are already to many "moving" parts everywhere and very easy to break already working things). Basically `assignedProtoFile` assigns annotation definition to the file it was defined in, as result it assumes `1:1` relationship between resource name definition and generated `<Resource>Name` classes.

Unfortunately the above statement is not true for java. If we have a resource name declared in one place, but used under two different packages (for example if API consists of 2 different interfaces under different packages) gapic-generator generates 2 `<Resource>Name`  classes from one resource name definition: `package1.<Resource>Name` and `package2.<Resource>Name`. This means that there is actual `1:m` relationship between resource name definition and the generated classes, but `assignedProtoFile` returns a single file per single resource (so the whole class model chosen here does not allow to express the relationship).

hzyi-google: Okay that makes sense. In this case why is it necessary to check whether assignedProtoFile.getFullName() and filedProtoFile.getProtoField().getFile() are equal? It seems that the latter can always be used if `fm instanceof ProtoField` is true.

src/test/java/com/google/api/codegen/testsrc/common/common_resources.proto
chrisdunelm: This file is about to change to use the new file-level annotation (rather than using proto messages).
Should this PR be updated to this new format?

vam-google: Could be the right thing to do. It will delay this PR and resource-name plugin one, because same thing will need to be implemented in two places, probably in quite different way, because both gapic-generator and resource name plugin try to do same thing but in a different way =(.

lukesneeringer: Please submit this PR as-is and let's make a new one to add support for the file level annotation.

I will not change `common_resources.proto` until gapic-generator supports it.

vam-google: This is the `commor_resources.proto` file, same as at https://github.com/googleapis/googleapis/blob/master/google/cloud/common_resources.proto. 
Here is is used only for tests (as from point of view of gapic-generator, it is just a regular proto file, gapic-generator is technically not aware of common_resources concept). 
In prod environment the file from googleapis will be used.

I guess the file does not have the other options because it simply does not need them (even the messages are fake, they exist only to declare annoations in them).

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
vam-google: Added. In brief: this is because `singleResourceNameConfigsFromGapicConfig` gets resource names from gapic_yaml, where they are in `lower_snake_case`, and `annotationResourceNameConfigs` has resources from annotations, where they are in `UpperCamelCase` so we need to convert the keys to common format before comparing them.

PR #2994, opened 2019-10-08
[bazel] Make Bazel extensions for Java GAPIC compatible with recent common_resources feature

vam-google:
Also refactor java rules, most importantly simplify gradle package generation, simplify the rules surface.
This PR relies on corresponding fix in protoc-java-resource-names plugin and contain braking changes for the current googleapis configuration (will be updated in separate CL).


rules_gapic/java/java_gapic_pkg.bzl
```diff
@@ -161,182 +155,90 @@ java_gapic_srcs_pkg = rule(
...
-        classifier = None):
+        project_deps = None,
+        test_project_deps = None,
+        **kwargs):
     resource_target_name = "%s-resources" % name
-    template_label = Label("//rules_gapic/java:resources/gradle/%s.gradle.tmpl" % pkg_type)
+
+    static_substitutions = dict(_PROPERTIES)
+    static_substitutions["{{name}}"] = name
+    #    static_substitutions.update(extra_substitutions)
```
noahdietz: Should we delete this commented code, comment it with a reason or uncomment?

rules_gapic/java/resources/gradle/client.gradle.tmpl
```diff
@@ -7,8 +7,8 @@ buildscript {
 apply plugin: 'java'
 
 description = 'GAPIC library for {{name}}'
-group = '{{group}}'
-version = '{{version}}'
+group = 'com.google.cloud'
```
noahdietz: I might be missing something in how this is used but hard coding "cloud" into the gradle template seems off to me. Why hard code it?

rules_gapic/java/java_gapic_pkg.bzl
```diff
@@ -354,6 +256,84 @@ def _java_gapic_gradle_pkg(
...
+    return _is_java_dependency(dep) and hasattr(dep[JavaInfo], "source_jars") and dep.label.package != "jar"
+
+def _is_proto_dependency(dep):
+    return ProtoInfo in dep
+
+def _is_gapic_pkg_dependency(dep):
+    files_list = dep.files.to_list()
+    if not files_list or len(files_list) != 1:
+        return False
+    return files_list[0].extension in (".tar.gz", "gz", "tgz")
```
noahdietz: This list of file compression extensions appears twice in this file. Should we pull it out into a variable like `_SUPPORTED_EXTENSIONS`?

This one does't have a `.` prefix on `gz` and `tgz` like the other list does, is that a real difference?

rules_gapic/java/resources/gradle/client.gradle.tmpl
vam-google: Yes, this is absolutely off (wrong). Unfortunately we already do it even deeper (inside the generator itself, when we generate same files via artman+gapic-generator bundle). For java it is here: [build_base.gradle.snip](https://github.com/googleapis/gapic-generator/blob/master/src/main/resources/com/google/api/codegen/metadatagen/java/grpc/build_base.gradle.snip#L11). 

Gapic-generator is contaminated with stuff like this, for example python setup snip hardcodes google-cloud-python in it: [setup.py.snip](https://github.com/googleapis/gapic-generator/blob/master/src/main/resources/com/google/api/codegen/py/setup.py.snip#L60). Other languages have similar stuff.

rules_gapic/java/java_gapic_pkg.bzl
vam-google: Oh, I think `.` is my mistake, `extension` property does not have the dot, so it should not be in the string literal. 

You are right that there is a lot of duplication here. I cleaned up this whole logic.

rules_gapic/java/java_gapic_pkg.bzl
vam-google: I forgot to remove this. Removing now.

vam-google:
@noahdietz PTAL

PR #2996, opened 2019-10-09
Add sample_type field to permit in-code samples.

vchudnov-g:
This is an optional field in the sample config. If it is not provided, generators should emit standalone samples only for the given config. (This preserves backward compatibility with uses of the current schema.)


src/main/proto/com/google/api/codegen/samplegen/v1p2/sample_config_v1p2.proto
```diff
@@ -86,6 +86,15 @@ message SampleSpecProto {
...
   repeated ResponseStatementProto response = 9;
+
+  // A list of the types samples to generate from this
+  // configuration. The elements can be any of "standalone" and
+  // "incode/XXX", where XXX is an arbitrary run of alphanumeric
+  // characters (these are used to lexicographically sort multiple
+  // in-code samples for the same RPC before including them in the
+  // library source code). If not specified or empty, defaults to
+  // ["standalone"]
+  repeated string sample_type = 10
```
hzyi-google: This is not compiling - please add a comma after `10`

vchudnov-g: Oops! Done

src/main/proto/com/google/api/codegen/samplegen/v1p2/sample_config_v1p2.proto
```diff
@@ -86,6 +86,15 @@ message SampleSpecProto {
...
   repeated ResponseStatementProto response = 9;
+
+  // A list of the types samples to generate from this
+  // configuration. The elements can be any of "standalone" and
+  // "incode/XXX", where XXX is an arbitrary run of alphanumeric
+  // characters (these are used to lexicographically sort multiple
+  // in-code samples for the same RPC when including them in the
+  // library source code). If not specified or empty, defaults to
+  // ["standalone"]
+  repeated string sample_type = 10;
```
software-dov: We can provide the default in the proto file using custom extensions.
```proto
import "google/protobuf/descriptor.proto";

message MyDefault {
  repeated string default = 1;
}

extend google.protobuf.FieldOptions {
  MyDefault my_default = 123456;
}

message Message {
  repeated string sample_type = 1 [(my_default) = {default: ["standalone"]}];
}
```

Something like that.

hzyi-google: This is fascinating, but I am afraid the tool that gapic-generator uses to convert YAMLs to protobufs isn't smart enough to handle it :(

software-dov:
Is this review still open?

vchudnov-g:
Merged with master. I had to tweak the circle config file to make CircleCI pass on my fork. For some reason, the `environment:` values for the job are not getting passed to each individual step. See, for example, the "Test speech" output for Ruby [here](https://circleci.com/gh/vchudnov-g/toolkit/278): `$LANGUAGE` (still set using `environment:`) does not get substituted, but `$RUBY_GEN_DIR`, which I now set via `>> $BASH_ENV`, does.

I don't know why this was passing on the main repo even though it was failing on my fork.

PTAL @hzyi-google @software-dov  @jbolinger

PR #2998, opened 2019-10-09
Parse resource name annotations from imported protos too

hzyi-google:
So far we only parse resource name related annotations from source protos (proto files that we generate code for) but not imports. However this does not work for IAM. IAM request messages declare `resource_reference.type = "*"`, but the request object is an import instead of a source proto for a certain API.

Also slightly tweak the method `addTag` (add `method_signature` and change `resource_reference.type` to `"*"`) to mimic IAM methods.

### DIFF
--- a/src/main/java/com/google/api/codegen/config/ResourceNameMessageConfigs.java
+++ b/src/main/java/com/google/api/codegen/config/ResourceNameMessageConfigs.java
@@ -52,6 +52,7 @@ public abstract class ResourceNameMessageConfigs {
   static ResourceNameMessageConfigs createFromAnnotations(
       DiagCollector diagCollector,
       List<ProtoFile> protoFiles,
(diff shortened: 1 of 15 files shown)


hzyi-google:
PTAL

src/main/java/com/google/api/codegen/config/ResourceNameMessageConfigs.java
```diff
@@ -77,6 +78,7 @@ static ResourceNameMessageConfigs createFromAnnotations(
                   message,
                   resourceField);
           if (Strings.isNullOrEmpty(entityName)) continue;
+          if (!resourceNameConfigs.containsKey(entityName)) continue;
```
vam-google: Please always use braces after `if`s (I see you made it consistent with the preivous if statement, please update the above if statement as well).

PR #3001, opened 2019-10-11
Support file-level resource_definition annotations

hzyi-google:
Adding support for generating the resource name classes in Java resource name plugin is WIP.


src/main/java/com/google/api/codegen/util/ProtoParser.java
```diff
@@ -148,6 +152,11 @@ public ResourceDescriptor getResourceDescriptor(MessageType element) {
     return getProtoExtension(element, ResourceProto.resource);
   }
 
+  public List<ResourceDescriptor> getFileLevelResourceDescriptors(ProtoFile protoFile) {
```
vam-google: It looks like this method is called only from another method in this class. Is there a reason to have it as `public`?

src/main/java/com/google/api/codegen/util/ProtoParser.java
```diff
@@ -175,32 +184,39 @@ private HttpRule getHttpRule(Method method) {
...
-                    "Multiple ResourceDescriptor defintions with the type"
-                        + " %s are defined in proto file %s. Values for type must be unique.",
-                    definition.getType(),
-                    protoFile.getFullName()));
-            continue;
-          }
-          ResourceDescriptorConfig config =
-              ResourceDescriptorConfig.from(definition, message.getFile());
-          mapBuilder.put(config.getUnifiedResourceType(), config);
+      protoFile
```
vam-google: (stylistic and subjective, a suggestion, not a requirement): this code looks more complicated than it should.

I think the reason why you had to introduce an intermediate collection (`resourceDescriptors`) is because we needed to iterate in same way over 2 sets of data, retrieved differently: 
set1: fileLevelResourceDescriptors (new in this PR)
set2: message-leel resource descriptors (old stuff)

To do that we here introduce an intermediate collection (`resourceDescriptors`) and then mix two styles (for loop and java8 streams) to process that. 

WDYT about just moving the processing logic in a private method and simply iterate over 2 sets of data twice, calling that private method (in either way: java8 streams or for loop)? It will achieve same result, will keep code style more consistent, kind of more efficient (no need to have intermediate collection) and easier to understand (it will be immediately clear that we just do same thing over file descriptors and then over message descriptors).

src/main/java/com/google/api/codegen/util/ProtoParser.java
hzyi-google: Nope. Too happy copying and pasting.

src/main/java/com/google/api/codegen/util/ProtoParser.java
hzyi-google: That makes sense! I'll update it.

PR #3017, opened 2019-11-05
samplegen: generate program.cs for c# samples

hzyi-google:


hzyi-google:
PTAL

src/main/java/com/google/api/codegen/transformer/StaticLangGapicSamplesTransformer.java
```diff
@@ -96,13 +98,23 @@ protected StaticLangGapicSamplesTransformer(
     ImmutableTable<String, String, ImmutableList<SampleConfig>> sampleConfigTable =
         productConfig.getSampleConfigTable();
 
+    List<ViewModel> sampleFiles;
+
     // We don't have sample configs. Continue to use gapic config.
     if (sampleConfigTable.isEmpty()) {
-      return generateSamplesFromGapicConfigs(interfaceContexts, productConfig, namer);
+      sampleFiles = generateSamplesFromGapicConfigs(interfaceContexts, productConfig, namer);
```
vchudnov-g: Could you move the comment right above the`if`to inside the `if` clause?

src/main/java/com/google/api/codegen/transformer/csharp/CSharpStandaloneSampleTransformer.java
```diff
@@ -29,13 +30,15 @@
...
 import com.google.api.codegen.viewmodel.ClientMethodType;
+import com.google.api.codegen.viewmodel.SampleEntryPointView;
 import com.google.common.collect.ImmutableMap;
 import java.util.function.Function;
 
 /** A transformer that generates C# standalone samples. */
 public class CSharpStandaloneSampleTransformer extends StaticLangGapicSamplesTransformer {
 
   private static final String STANDALONE_SAMPLE_TEMPLATE_FILENAME = "csharp/standalone_sample.snip";
+  private static final String PROGRAM_TEMPLATE_FILENAME = "csharp/gapic_samples_program.snip";
```
vchudnov-g: Should we call this `..._runner` rather than `..._program`?

src/main/java/com/google/api/codegen/transformer/csharp/CSharpStandaloneSampleTransformer.java
```diff
@@ -115,6 +118,21 @@ protected ClientMethodType fromCallingForm(CallingForm callingForm) {
...
 
+  @Override
+  protected SampleEntryPointView generateSampleEntryPoint(InterfaceContext interfaceContext) {
+    SampleEntryPointView.Builder entryPoint = SampleEntryPointView.newBuilder();
+    String sampleDirectory =
+        pathMapper.getOutputPath(
+            interfaceContext.getInterfaceModel().getFullName(),
+            interfaceContext.getProductConfig());
+    String outputPath =
+        new StringBuilder().append(sampleDirectory).append("/Program.cs").toString();
```
vchudnov-g: Is hard-coding the forward slash a problem? Where does it get converted to the system-specific path separator?

vchudnov-g: Here, too, `Program` strikes me as too generic a name. How about `Runner`. @jskeet , what do you think?

hzyi-google: `Program` is a commonly used name for entry point class in C#.

hzyi-google: I think the rendering engine explicitly uses forward slashes as it's commonly used in gapic-generator: e.g., https://github.com/googleapis/gapic-generator/blob/85b64b225dbd504c2aad519d0ea2e15ee43395a6/src/main/java/com/google/api/codegen/gapic/CommonGapicCodePathMapper.java#L5

src/main/java/com/google/api/codegen/transformer/csharp/CSharpStandaloneSampleTransformer.java
hzyi-google: same as below, `program` is commonly used as the name of an entry point class.

PR #3018, opened 2019-11-05
C# samplegen: remove commented-out lines showing default values of parameters

hzyi-google:
Fix https://github.com/googleapis/google-cloud-dotnet/pull/3650#discussion_r337877664


SurferJeffAtGoogle:
I'd like to see the default values be actual default values for the parameters, so the reader still sees a clue of what the default values should look like.

beccasaurus:
+💯 @amanda-tarafa [requested the same](https://github.com/googleapis/google-cloud-dotnet/pull/3650#discussion_r354413864) just this morning :)

PR #3019, opened 2019-11-05
Java samplegen: replace comment-out default lines with an overload of the sample method

hzyi-google:
Fix https://github.com/googleapis/google-cloud-java/pull/6543


hzyi-google:
cc/ @kurtisvg

PR #3020, opened 2019-11-07
Python: Include custom static files in generated Python docs

busunkim96:
Allow extra static files in the generated docs so the Python 2 sunset banner can be added.

See https://github.com/googleapis/google-cloud-python/pull/9036

image


busunkim96:
I have reached the end of my copy-pasta journey. @software-dov would you be able to take a look?

software-dov:
I feel like I'm the least qualified person to take a look at this review; I'm not great with the monolith, and I've looked at CSS classes for maybe 20 minutes of my life.

The screenshot looks great, and that's good enough for me.

busunkim96:
Chris pointed out that we can probably do most of this with synthtool templates. I'lll back out most of the changes here and just keep the change to `docs/conf.py`.

PR #3022, opened 2019-11-07
resource names: assign the correct proto file for package generation

hzyi-google:


src/main/java/com/google/api/codegen/config/ResourceDescriptorConfig.java
```diff
@@ -135,13 +135,15 @@ private String getUnqualifiedTypeName() {
                       .setEntityId(nameMap.get(p).toUpperCamel())
                       .setEntityName(overrideConfig.getEntityName())
                       .setCommonResourceName(overrideConfig.getCommonResourceName())
+                      .setAssignedProtoFile(protoFile)
```
vam-google: I believe this will break common resources, as they are defined in a common file, but must generate classes in service-specific namespace, so using protoFile to determine package of the generated class would not work, because common_resources namespace does not match service namespace.

hzyi-google: So this is why I also had the change in `ProtoParser` and `GapicProductConfig`, which would always assign one of the source protos of the API to all the resource name configs, including common resource names.
There are actually references to common resource names in the baseline files. They remain unchanged after this PR (which is WAI) so they did not show in the diffs. (I know this because I broke them during implementing the PR, and hence I added the change in the other two files to fix that)

hzyi-google:
Replaced by #3027

PR #3024, opened 2019-11-07
bazel: add grpc_service_config attr to all gapic rules

noahdietz:
I tested locally against googleapis/googleapis using local_repository to link my changes into the googleapis WORKSPACE

Note: looks like there might be some formatting changes included too

### DIFF
--- a/rules_gapic/csharp/csharp_gapic.bzl
+++ b/rules_gapic/csharp/csharp_gapic.bzl
@@ -108,7 +108,7 @@ _csharp_gapic_postprocessed_srcjar = rule(
     },
 )
 
-def csharp_gapic_srcjar(name, src, gapic_yaml, service_yaml, package, **kwargs):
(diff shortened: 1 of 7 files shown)


rules_gapic/csharp/csharp_gapic.bzl
```diff
@@ -108,7 +108,7 @@ _csharp_gapic_postprocessed_srcjar = rule(
     },
 )
 
-def csharp_gapic_srcjar(name, src, gapic_yaml, service_yaml, package, **kwargs):
+def csharp_gapic_srcjar(name, src, gapic_yaml, service_yaml, package, grpc_service_config = None, **kwargs):
```
vam-google: Speaking about formatting changes: when the arguments list get that long, we should probably switch to the arg-per-line style. Not very important, but please consider reformatting this to that style.

noahdietz: Done!

PR #3041, opened 2019-11-25
Ruby standalone sample syntax updates 

beccasaurus:
Refactors generated Ruby samples to render using syntax suggested from this code review PR:
 - https://github.com/googleapis/google-cloud-ruby/pull/4264

[WIP] to first get +1 on https://github.com/googleapis/google-cloud-ruby/pull/4264 for these changes (which shows rendered Ruby code samples)

@googleapis/samplegen

### DIFF
--- a/src/main/resources/com/google/api/codegen/ruby/response.snip
+++ b/src/main/resources/com/google/api/codegen/ruby/response.snip
@@ -22,6 +22,7 @@
 @end
 
 @private processCommentView(view)
+
   {@toComments(view.lines)}
 @end
 
--- a/src/main/resources/com/google/api/codegen/ruby/standalone_sample.snip
+++ b/src/main/resources/com/google/api/codegen/ruby/standalone_sample.snip
@@ -17,7 +17,6 @@
       @#   bundle exec ruby {@sampleFile.outputPath} {@commandlineArgumentList(sample.sampleInitCode.argDefaultParams)}
 
       @# {@sample.descriptionLine}
-
       {@standaloneSample(apiMethod, sample)}
 
     @end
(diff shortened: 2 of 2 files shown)


beccasaurus:
@hzyi-google from syncing up with @blowmage we know that this functionality is planned for the Ruby µgenerator. Closing pending support added to µgenerator.

hzyi-google:
@beccasaurus it seems this PR contains only syntax fixes, how come the functionality is planned for the micro generator? Does the micro generator support samplegen already?

beccasaurus:
@hzyi-google Yep, it does support samplegen (although it does not yet support generating standalone files). Full samplegen support will be in place before any libraries are switched over to Ruby's µgen.

Now, that said! Shortly after I closed this, I talked to @alangasperini and @vchudnov-g and we decided it would be smart to get LGTM's from Ruby folks on the monolith's sample output because the µgen's output should be nearly identical (except that the new Ruby µgen produces some breaking changes)

With that in mind... it seems useful to get these changes in!

There was only 1 remaining comment from Ruby folks regarding the commented out variable names. Here is an example of the desired sample output from @blowmage:
https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/2b05b1b2e37c4e0867e348f78993b9a300ae0d9c/translate/translate_v3_translate_text.rb#L30-L35

With all of this in mind, I'm going to (1) re-open (2) give this PR description a tasklist listing the syntax changes (3) make the final requested change (4) ask @blowmage @dazuma /cc @frankyn for LGTM on these changes and we'll get them committed here into the monolith!

PR #3048, opened 2019-12-12
[bazel] Add BUILD file generator and tests.

vam-google:
This is the generator used to populate current googleapis directory with `BUILD.bazel` files. It is designed to not have any external dependencies (pure JDK is enough) for future easier integration to more appropriate places (sanitization logic).

This is one of the reasons, why it heavily relies on regexps for parsing all types of files and does not use any of the actual parsers (proto, yaml, json).

The goal of this generator is to remain simple and portable (no dependencies). It is highly unlikely that it will have to be heavily extended in the future, so just relying on regex should be enough for now. We need only tiny portion of information from each file (like a couple of properties), so using actual parsers is an overkill and it will complicate the implementation drastically.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -14,5 +14,10 @@
       "extends": "packages:linters",
       "groupName": "linters"
     }
+  ],
+  "ignoreDeps": [
+    "com.google.protobuf:protobuf-java",
+    "com.google.protobuf:protoc",
+    "com.google.cloud:google-cloud-pubsub"
   ]
 }


rules_gapic/bazel/src/main/java/com/google/api/codegen/bazel/ApisVisitor.java
```diff
@@ -0,0 +1,172 @@
...
+    File outDir = new File(outDirPath);
+
+    if (!outDir.exists()) {
+      if (!outDir.mkdirs()) {
+        System.out.println("WARNING: Could not create directory: " + outDir.toString());
+        return FileVisitResult.CONTINUE;
+      }
+    }
+
+    // System.out.println(bp);
```
noahdietz: Delete this please.

rules_gapic/bazel/src/main/java/com/google/api/codegen/bazel/ArgsParser.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.Scanner;
+
+class ArgsParser {
+  private final Map<String, String> parsedArgs = new HashMap<>();
+
+  ArgsParser(String[] args) {
+    for (String arg : args) {
+      String[] argNameVal = arg.split("=");
+      if (argNameVal.length != 2) {
+        System.out.println("WATNING: Ignoring unrecognized argument: " + arg);
```
noahdietz: Typo: s/`WATNING`/`WARNING`

rules_gapic/bazel/src/main/java/com/google/api/codegen/bazel/BUILD.bazel.raw_api.mustache
```diff
@@ -0,0 +1,169 @@
+# This file was automatically generated by BuildFileGenerator
+
+# This is an API workspace, having public visibility by default makes perfect sense.
+package(default_visibility = ["//visibility:public"])
+
+##############################################################################
+# Common
+##############################################################################
+load("@com_google_googleapis_imports//:imports.bzl", "proto_library_with_info")
```
noahdietz: I think this is missing `load("@rules_proto//proto:defs.bzl", "proto_library")`

rules_gapic/bazel/src/main/java/com/google/api/codegen/bazel/BUILD.bazel.raw_api.mustache
```diff
@@ -0,0 +1,169 @@
...
+    name = "{{name}}_proto",
+    srcs = [
+        {{proto_srcs}}
+    ],
+    deps = [
+        {{proto_deps}}
+    ],
+)
+
+proto_library_with_info(
```
noahdietz: Is it necessary to import and set up this target when it isn't used in this "raw" build file? If not, let's remove it and the load statement for it.

rules_gapic/bazel/src/main/java/com/google/api/codegen/bazel/BazelBuildFileView.java
```diff
@@ -0,0 +1,224 @@
...
+
+    tokens.put("proto_srcs", joinSetWithIndentation(bp.getProtos()));
+    tokens.put("version", bp.getVersion());
+    tokens.put("package", bp.getProtoPackage());
+
+    String packPrefix = bp.getProtoPackage().replace(".", "/") + '/';
+    Set<String> actualImports = new TreeSet<>();
+    for (String imp : bp.getImports()) {
+      if (imp.startsWith(packPrefix) && imp.indexOf('/', packPrefix.length()) == -1) {
+        // Ignore imports form same package, as all protos in same package are put in same
```
noahdietz: Typo: s/`form`/`from`

rules_gapic/bazel/src/main/java/com/google/api/codegen/bazel/BuildFileGenerator.java
```diff
@@ -0,0 +1,22 @@
+package com.google.api.codegen.bazel;
+
+import java.io.IOException;
+import java.nio.file.Files;
+
+public class BuildFileGenerator {
+  public static void main(String[] args) throws IOException {
```
hzyi-google: Can we have some example usage here?

hzyi-google:
meta question: while this bazel file generator will certainly be very useful, it seems very loosely related to gapic-generator. Would it be possible to put the code in `//third_party/googleapis`? With bazel you don't have any non-jvm dependencies and the generator itself is bazel-aware, so it shouldn't be hard to move the code around. It'll potentially make our tool easier to use - most API producers will be able to run the generator to generate the bazel files and use the generated bazel files to generate libraries from `third_party/googleapis`, without needing to talk to us or go to github. WDYT?

rules_gapic/bazel/src/main/java/com/google/api/codegen/bazel/ApiVersionedDir.java
```diff
@@ -0,0 +1,212 @@
...
+    // Parse proto package
+    Matcher m = PROTO_PACKAGE.matcher(fileBody);
+    if (m.find() && (protoPackage == null || name == null || version == null)) {
+      protoPackage = m.group("protoPackage");
+      String[] tokens = protoPackage.split("\\.");
+      if (tokens.length >= 1 && name == null) {
+        name = tokens[tokens.length - 1];
+        assemblyName = name;
+      }
+      if (tokens.length >= 2) {
```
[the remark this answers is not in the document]
vam-google: Done

rules_gapic/bazel/src/main/java/com/google/api/codegen/bazel/ApiVersionedDir.java
```diff
@@ -0,0 +1,212 @@
...
+
+    // Parse declared service names
+    m = SERVICE.matcher(fileBody);
+    while (m.find()) {
+      services.add(m.group("service"));
+    }
+  }
+
+  void parseJsonFile(String fileName, String fileBody) {
+    if (fileBody.contains("methodConfig") && fileBody.contains("retryPolicy")) {
```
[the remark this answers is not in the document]
vam-google: Ok. This wasn't used yet (since we decided to not include grpc_service_configs yet). I left it to check only for the methodConfig portion (I guess it should be good enough).

rules_gapic/bazel/src/main/java/com/google/api/codegen/bazel/ApisVisitor.java
vam-google: Deleted

rules_gapic/bazel/src/main/java/com/google/api/codegen/bazel/ArgsParser.java
vam-google: Oops =). Corrrected

rules_gapic/bazel/src/main/java/com/google/api/codegen/bazel/BUILD.bazel.raw_api.mustache
vam-google: You are right. It is optional now, but will be required for bazel 2.0 (that is why build, which used the gnerated files, did not fail). Added.

rules_gapic/bazel/src/main/java/com/google/api/codegen/bazel/BUILD.bazel.raw_api.mustache
vam-google: I think you are right, it is not needed and is even harmful. Removed those.

rules_gapic/bazel/src/main/java/com/google/api/codegen/bazel/BazelBuildFileView.java
vam-google: corrected

rules_gapic/bazel/BUILD.bazel
```diff
@@ -0,0 +1,18 @@
...
+    visibility = ["//visibility:public"],
+)
+
+java_test(
+    name = "build_file_generator_test",
+    srcs = glob(["src/test/java/**/*.java"]),
+    deps = [":build_file_generator"],
+    data = glob(["src/test/data/**/*.*"]),
+    test_class = "com.google.api.codegen.bazel.BuildFileGeneratorTest",
+)
```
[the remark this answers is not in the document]
vam-google: Added new line

rules_gapic/bazel/src/main/java/com/google/api/codegen/bazel/ApisVisitor.java
```diff
@@ -0,0 +1,172 @@
...
+  public static String readFile(String path) throws IOException {
+    return new String(Files.readAllBytes(Paths.get(path).normalize()), StandardCharsets.UTF_8);
+  }
+
+  private static String readFile(Path path) throws IOException {
+    return new String(Files.readAllBytes(path), StandardCharsets.UTF_8);
+  }
+
+  public void setWriter(boolean writer) {
+    this.writer = writer;
```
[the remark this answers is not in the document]
vam-google: Renamed it to `writerMode`. This field determines if the AisVisitor is in writer or reader mode. The way it works: it first walks the file tree to read stuff and then walks same file tree to write stuff. 

The reason why two passes are needed is because the the order in which siblings of a directory are visited is not defined, and some of the properties depend on the "full picture" (i.e. it needs to know all files in dir to pick the right one, like if there are service.yaml and service_v1.yaml for v1 dir it should pick v1). Only doing it in two passes we can guarantee that the "writer" has "full picture" of the previous read directory and pick the correct files and propeerties.

rules_gapic/bazel/src/main/java/com/google/api/codegen/bazel/ApisVisitor.java
```diff
@@ -0,0 +1,172 @@
...
+    if (fileName.endsWith(".yaml")) {
+      if (!fileName.endsWith(".legacy.yaml")) {
+        String fileBody = readFile(file);
+        bp.parseYamlFile(fileName, fileBody);
+        bap.parseYamlFile(fileName, fileBody);
+      }
+    } else if (fileName.endsWith(".proto")) {
+      bp.parseProtoFile(fileName, readFile(file));
+    } else if (fileName.endsWith(".bazel")) {
+      // merge files?
```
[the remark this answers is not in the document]
vam-google: Done

rules_gapic/bazel/src/main/java/com/google/api/codegen/bazel/ArgsParser.java
```diff
@@ -0,0 +1,57 @@
...
+      }
+      parsedArgs.put(argNameVal[0], argNameVal[1].trim());
+    }
+
+    List<String> required = Arrays.asList("--src", "--dest");
+    if (!parsedArgs.keySet().containsAll(required)) {
+      System.out.println(
+          "ERROR: Not all of the required arguments are spcified. The required arguments are: "
+              + required);
+    }
```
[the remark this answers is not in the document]
vam-google: Made it throw Exception, which will terminate program with the message.

rules_gapic/bazel/src/main/java/com/google/api/codegen/bazel/ArgsParser.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Scanner;
+
+class ArgsParser {
+  private final Map<String, String> parsedArgs = new HashMap<>();
+
+  ArgsParser(String[] args) {
+    for (String arg : args) {
+      String[] argNameVal = arg.split("=");
```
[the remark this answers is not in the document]
vam-google: I think for this particular case I would do same even if there was no goal to avoid third-party dependencies (it is a simple call, and the tool does not do sophisticated comman line interface (only two args) so doing it manually justified).

rules_gapic/bazel/src/main/java/com/google/api/codegen/bazel/BuildFileGenerator.java
vam-google: Added

vam-google:
@hzyi-google I think your proposal is very reasonable. My reasons to put it here is:
- This allows to keep rules and build files generator as a self-contained instance (similarly like gapic-generator is capable of generating gapic yaml)
- I would like to avoid adding java-specific code (or any programming code actually) to googleapis. Lets keep googleapis a repo for configs and API definitions, but not for program source code.

This is still in "testing" mode, and I'm not sure what is the proper lace for it (i guess the best place is the sanitization logic). Putting it here as one of the easiest (but still reasonable) places for it. 

If you don't have a strong opinion abou it, I would like to keep this code here. WDYT?

hzyi-google:
Okay that's fair.

I'll be away for a while so please feel free to merge after @noahdietz signs off.

vam-google:
This PR was accidentally closed. Please see the reopened PR #3055.

PR #3049, opened 2019-12-12
chore: ignore a few dependencies

hzyi-google:
- PubSub was used only for benchmarking (which we don't not measure now)
- The bot isn't smart enough to correctly update protobuf's version (not even humans can understand a gradle script, please don't put that burden on a bot)

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -14,7 +14,7 @@
       "extends": "packages:linters",
       "groupName": "linters"
     }
-  ]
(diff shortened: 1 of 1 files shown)


alexander-fenster:
@hzyi-google I reported it to the Renovate bot author just in case. https://github.com/googleapis/gapic-generator/pull/3042#issuecomment-564877469

PR #3055, opened 2019-12-19
[bazel] Add BUILD file generator and tests

vam-google:
(It is a copy of https://github.com/googleapis/gapic-generator/pull/3048, which was accidentally wiped out and closed. The state of this PR corresponds to the state of the original one on the moment of closure)

This is the generator used to populate current googleapis directory with BUILD.bazel files. It is designed to not have any external dependencies (pure JDK is enough) for future easier integration to more appropriate places (sanitization logic).

This is one of the reasons, why it heavily relies on regexps for parsing all types of files and does not use any of the actual parsers (proto, yaml, json).

The goal of this generator is to remain simple and portable (no dependencies). It is highly unlikely that it will have to be heavily extended in the future, so just relying on regex should be enough for now. We need only tiny portion of information from each file (like a couple of properties), so using actual parsers is an overkill and it will complicate the implementation drastically.


rules_gapic/bazel/src/main/java/com/google/api/codegen/bazel/BUILD.bazel.raw_api.mustache
```diff
@@ -0,0 +1,162 @@
+# This file was automatically generated by BuildFileGenerator
+
+# This is an API workspace, having public visibility by default makes perfect sense.
+package(default_visibility = ["//visibility:public"])
+
+##############################################################################
+# Common
+##############################################################################
+load("@com_google_googleapis_imports//:imports.bzl")
```
noahdietz: Missing `load("@rules_proto//proto:defs.bzl", "proto_library")`

vam-google: Wow! Very nice catch!

PR #3059, opened 2019-12-20
Raise deprecation warning when Python 2.7 is detected

busunkim96:


busunkim96:
@software-dov Ah, thank you for reviewing! Would you be able to point me to which tests I should be looking at?

software-dov:
In the test-gapic-generator, it looks like the two failures are `python_no_path_templates` and `python_multiple_services`, both of which are baseline tests.

For python-test, it looks like the generator is failing to run. We should get a look at `/tmp/workspace/reports/generations/smoketest.log` from the first container.

Not sure about the bazel tests.

PR #3067, opened 2020-01-06
Support empty method signatures

hzyi-google:
When an rpc has a `method_signature = ""`, make sure gapic generator generates an overload with no parameters in Java and C#.

Add a test to demonstrate that, and slightly tweak the format of templates to make generated code look prettier.

Also update Java `initcode.snip` to make code that instantiates request objects indent correctly when the code takes up multiple lines.

### DIFF
--- a/src/main/resources/com/google/api/codegen/csharp/gapic_snippets.snip
+++ b/src/main/resources/com/google/api/codegen/csharp/gapic_snippets.snip
@@ -67,7 +67,7 @@
 @private flattenedMethodSync(snippetMethod)
     public void {@snippetMethod.snippetMethodName}()
     {
-        // Snippet: {@snippetMethod.method.name}({@snippetArgTypes(snippetMethod.method.methodParams)}{@if(snippetMethod.method.methodParams, ",CallSettings", "")})
+        // Snippet: {@snippetMethod.method.name}({@snippetArgTypes(snippetMethod.method.methodParams)}{@if(snippetMethod.method.methodParams, ",", "")}CallSettings)
         {@preparations(snippetMethod, FALSE)}
         @if snippetMethod.method.hasReturnValue
             {@snippetMethod.callerResponseTypeName} response = {@sampleMethodCall(snippetMethod)};
@@ -81,7 +81,7 @@
 @private flattenedMethodAsyncCallSettings(snippetMethod)
     public async Task {@snippetMethod.snippetMethodName}()
     {
-        // Snippet: {@snippetMethod.method.name}({@snippetArgTypes(snippetMethod.method.methodParams)}{@if(snippetMethod.method.methodParams, ",CallSettings", "")})
+        // Snippet: {@snippetMethod.method.name}({@snippetArgTypes(snippetMethod.method.methodParams)}{@if(snippetMethod.method.methodParams, ",", "")}CallSettings)
         // Additional: {@snippetMethod.method.name}({@snippetArgTypes(snippetMethod.method.methodParams)},CancellationToken)
         {@preparations(snippetMethod, TRUE)}
         @if snippetMethod.method.hasReturnValue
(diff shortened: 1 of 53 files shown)


src/test/java/com/google/api/codegen/discogapic/testdata/java/java_simplecompute.v1.json.baseline
```diff
@@ -8890,8 +8890,8 @@ public class AddressClient implements BackgroundResource {
   public final AggregatedListAddressesPagedResponse aggregatedListAddresses(ProjectName project) {
     AggregatedListAddressesHttpRequest request =
         AggregatedListAddressesHttpRequest.newBuilder()
-        .setProject(project == null ? null : project.toString())
-        .build();
+            .setProject(project == null ? null : project.toString())
```
vam-google: It is ok to keep it here, but please try to not mix unrelated refactoring/formatting with new features. For cases, when it happens, please mention it in the PR description.

src/main/resources/com/google/api/codegen/csharp/gapic_snippets.snip
```diff
@@ -67,7 +67,7 @@
 @private flattenedMethodSync(snippetMethod)
     public void {@snippetMethod.snippetMethodName}()
     {
-        // Snippet: {@snippetMethod.method.name}({@snippetArgTypes(snippetMethod.method.methodParams)},CallSettings)
+        // Snippet: {@snippetMethod.method.name}({@snippetArgTypes(snippetMethod.method.methodParams)}{@if(snippetMethod.method.methodParams, ",", "")}CallSettings)
```
vam-google: Probably the logic like this should be in the view object itself or in the `@snippetArgTypes()` "snip" method (so you can avoid repeating the `@if` statement everywhere).

src/main/resources/com/google/api/codegen/java/grpc_test.snip
```diff
@@ -82,8 +82,10 @@
   public void {@test.name}() {
     {@addResponse(test)}
 
-    {@initCode(test.testCaseInitCode)}
+    @if test.testCaseInitCode.lines
```
vam-google: Why is this needed?

src/main/resources/com/google/api/codegen/java/main.snip
```diff
@@ -452,9 +452,13 @@
       {@pathTemplateCheckLine(pathTemplateCheck)}
     @end
     {@apiMethod.serviceRequestTypeName} request =
-        {@apiMethod.serviceRequestTypeName}.newBuilder()
-        {@builderSetCalls(apiMethod.requestObjectParams)}
-        .build();
+        @if apiMethod.requestObjectParams
+          {@apiMethod.serviceRequestTypeName}.newBuilder()
+              {@builderSetCalls(apiMethod.requestObjectParams)}
```
vam-google: Can `@builderSetCalls` method return nothing if `apiMethod.requestObjectParams` is empty? I is subjective, but it seems like moving the check to the `@builderSetCalls` snip method is cleaner, than having the if-else statement here. Also it can reduce code duplication (seems same check is done on L613)

noahdietz:
LGTM whenever Vadym approves.

src/test/java/com/google/api/codegen/discogapic/testdata/java/java_simplecompute.v1.json.baseline
hzyi-google: Edited the PR description.

src/main/resources/com/google/api/codegen/csharp/gapic_snippets.snip
hzyi-google: So `@snippetArgTypes()` are reused in another overload which has an additional argument `CancellationToken` (see [here](https://github.com/googleapis/gapic-generator/blob/cd919793e439d2d21505900aa08ef9f99519d8fa/src/main/resources/com/google/api/codegen/csharp/gapic_snippets.snip#L85)). Moving `CallingSettings` into the view object will cause huge refactoring to C# templates. Given the state of the C# micro generator it is probably not worth it...

src/main/resources/com/google/api/codegen/java/grpc_test.snip
hzyi-google: To avoid the extra blank line if `test.testCaseInitCode` is empty.

src/main/resources/com/google/api/codegen/java/main.snip
hzyi-google: I did this because even if `@builderSetCalls` returns nothing, there will be an empty newline in the caller method. This is some uncommon settings about snippet that I never understand... That's also why I wrapped `@builderSetCalls` in an if-else statement.

hzyi-google:
PTAL

PR #3069, opened 2020-01-08
fix: use string literals in samples for an any resource name if there are no other resources

hzyi-google:
It's hard to demonstrate the change in baseline since we do have other resources. See [this](https://github.com/hzyi-google/sample-staging/blob/ae03c7d75df202a060ab25c1b114dcb7ed5e5e5f/20200108/java/gapic-google-cloud-containeranalysis-v1/src/main/java/com/google/cloud/devtools/containeranalysis/v1/ContainerAnalysisClient.java#L54-L56) for what the generated code looks like.


hzyi-google:
Yes, this is to fix containeranalysis

PR #3072, opened 2020-01-16
Use PagedCallableMethod as example if there is not other method

hzyi-google:


hzyi-google:
It's a bit hard to demonstrate this behavior in baseline since this case happens when there is only one method in an API, but our baseline has a thousand method...

PR #3076, opened 2020-01-28
Remove unsupported methods from baseline tests

hzyi-google:
Delete the method `ListStrings` from the baseline library API and change the paging field to `Book book` for method `FindRelatedBooks` (No code changes).

As per https://aip.dev/client-libraries/4233 the resource to be paged on must be non-primitive. This voids two methods `ListStrings` and `FindRelatedBooks` in the baseline library API, which were used to test the functionality of transforming an auto-paging resource name string field to a resource name object in Java and C#. Since we don't support the feature any more, we can safely delete them from baseline files.

We should error when the proto does not satisfy the requirements listed in the AIP above though, but since we have linters on protobuf changes taking care of that already I want to postpone that later ([issue](https://github.com/googleapis/gapic-generator/issues/3077) filed). Deleting the two methods eliminates the extra burden in implementing multi-pattern resources referenced by `child_type`. Also I do not know of any real APIs that need this feature, but please let me know if there are.


noahdietz:
(Haven't looked at the PR yet).

I know of three APIs that paginate on a primitive type, all on `string`.

* [PubSub](https://github.com/googleapis/googleapis/blob/master/google/pubsub/v1/pubsub_gapic.yaml#L540-L556)
* [Logging](https://github.com/googleapis/googleapis/blob/master/google/logging/v2/logging_gapic.yaml#L486-L499)
* [Firestore](https://github.com/googleapis/googleapis/blob/master/google/firestore/v1/firestore_gapic.yaml#L206-L219)  (v1 and v1beta1)

In all cases, the `repeated string` is the only `repeated` field in the response, so it's not a case of bad auto-pagination. Technically, these APIs violate the relevant AIPs, but we'd have to work with them to change their APIs, which seems unlikely to happen.

jskeet:
Humbug. That's annoying. Thanks for spotting it ahead of time, Noah.

cc @lukesneeringer - we'll need to work out what to do about these...

hzyi-google:
@noahdietz Thanks for pointing that out.

Now that we still need this functionality for some APIs I will keep the tests there, and close this PR.

PR #3078, opened 2020-01-28
Fix generated code example for changing the timeout. 

kamalaboulhosn:
The original code just gets the Settings builder and sets the total timeout on it without calling setRetrySettings on the method setting's builder.


kamalaboulhosn:
@hzyi-google I don't have permission to merge this in. Would you be able to do it? Thanks!

hzyi-google:
@kamalaboulhosn done! Thank you for contributing!

PR #3079, opened 2020-01-29
bazel: Add support for prebuilt PHP binaries

vam-google:
Also add linux prebuilt PHP binaries in the repository.

This allows bazel build to not rebuild PHP from sources each time. In case if the prebuilt versions fail on the build machine it falls back to building PHP from sources. 

*Also adds a tiny fix for BazelBuildFileView.java (which is not related, but I decided that the fix was not worth its own PR).


alexander-fenster:
I don't like this :) Is there any way without putting a PHP binary here? Maybe there is any official location to take it from?

vam-google:
@alexander-fenster I don't like it either, bu tplease think aobut it as an optimization hack. Note, it is optional (it is not required to provide any binary during the build, so it will be built form sources automatically on demand). There are not official binary distributions for PHP, and I don't want to depend on anything non-official.

PR #3081, opened 2020-01-30
Generate method overloads when a resource field has multiple resource names configured

hzyi-google:
- When a field has `child_type` reference and the referenced resource has multiple patterns, allow this field to be associated with multiple resources if all these resources combined have exactly all the parent patterns of the resource referenced by `child_type`.
- Generate method overloads if the field with multiple resource names are in a flattening config.
- If there are multiple fields with multiple resource names, generate all possible combinations.
- Always generate an overload with all resource names treated as strings.
- In Java, repeated fields with resource name configs are always treated as `List<String>` to avoid multiple methods having the same signature after type erasure.
- The `*` annotation is not yet supported.

See changes to `ArchiveBooks`, `LongRunningArchiveBooks` and `StreamingArchiveBooks` for testing generating overloads for unary, longrunning and streaming methods.

See changes to `MoveBooks` for testing generating all the combinations.


src/test/java/com/google/api/codegen/gapic/testdata/java/java_library.baseline
```diff
@@ -7586,9 +7587,9 @@ public class LibraryClient implements BackgroundResource {
             .setRequiredSingularString(requiredSingularString)
             .setRequiredSingularBytes(requiredSingularBytes)
             .setRequiredSingularMessage(requiredSingularMessage)
-            .setRequiredSingularResourceName(requiredSingularResourceName)
-            .setRequiredSingularResourceNameOneof(requiredSingularResourceNameOneof)
-            .setRequiredSingularResourceNameCommon(requiredSingularResourceNameCommon)
+            .setRequiredSingularResourceName(requiredSingularResourceName == null ? null : requiredSingularResourceName.toString())
```
chrisdunelm: Is it correct that a required resource-name is allowed to be `null`?

hzyi-google: This has been how java delegates the overload with resource name to the non-flattened method. 

For example: https://github.com/googleapis/java-tasks/blob/master/google-cloud-tasks/src/main/java/com/google/cloud/tasks/v2/CloudTasksClient.java#L1532-L1540

Actually I think protobuf Java does not allow setting null values on any field, so the code will still fail on the client side. I guess we just pass the error handling to protobuf :)

PR #3087, opened 2020-02-06
feat (Java): generate tests using the non-flattening overload if an rpc does not have method_signature config

hzyi-google:
Also validated locally that the change would fix accessapproval.


hzyi-google:
Update: showcase tests caught a bug in the generated unit tests that the unit tests don't know how to handle oneof. For now I changed the request and assertion to populate only required fields.

hzyi-google:
Update 2: Updated packagegen to fix the diamond dependency problem between protobuf and gax for showcase.

PR #3098, opened 2020-02-26
Python: remove experimental notices for streaming

hongalex:
The PubSub python client has been GA since August 2019, so the notice about streaming should be removed.

I believe some of the other references are related to samplegen/tests, so I removed the reference since I believe the experimental notice was placed because of streaming.
Let me know if you disagree and I'll revert those changes.


busunkim96:
@hongalex Thank you for doing this! @hzyi-google Could you also give this a quick review?

PR #3099, opened 2020-02-26
Update generated Node.js samples

fhinkel:
* use async/await
* fix paginated samples
* add Buffer.from()


alexander-fenster:
@fhinkel Please let me know if you need a review on this!

PR #3100, opened 2020-02-28
feat: adds reserved name support for PHP and reserved name tests

bshaffer:
A new API coming out soon has a proto class `Namespace`, which is breaking for PHP client libraries.

See https://github.com/protocolbuffers/protobuf/pull/7268/files, which ensures the protobuf c library follows the same rules

### DIFF
--- a/src/test/java/com/google/api/codegen/gapic/testdata/php/php_library.baseline
+++ b/src/test/java/com/google/api/codegen/gapic/testdata/php/php_library.baseline
@@ -5158,6 +5158,7 @@ use Google\Auth\FetchAuthTokenInterface;
 use Google\Example\Library\V1\MethodRequest;
 use Google\Example\Library\V1\MethodResponse;
 use Google\Example\Library\V1\MyProtoGrpcClient;
+use Google\Example\Library\V1\PBNamespace;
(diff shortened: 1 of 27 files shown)


src/test/java/com/google/api/codegen/gapic/testdata/php/php_library.baseline
```diff
@@ -5421,6 +5422,57 @@ class MyProtoGapicClient
...
+     *     @type SubMessage $myfield
+     *     @type SubMessage $secondfield
+     *     @type RetrySettings|array $retrySettings
+     *          Retry settings to use for this call. Can be a
+     *          {@see Google\ApiCore\RetrySettings} object, or an associative array
+     *          of retry settings parameters. See the documentation on
+     *          {@see Google\ApiCore\RetrySettings} for example usage.
+     * }
+     *
+     * @return \Google\Example\Library\V1\Namespace
```
dwsupplee: It looks like this should be `PBNamespace`?

bshaffer: Good catch! This is now fixed

PR #3115, opened 2020-03-05
fix: honor resource config on name field in method flattening

hzyi-google:
We generate flattening overload methods where resource name strings are replaced by resource name types. Previously we only check `resource_reference` to see if a field is a resource name, this is not enough as the field `name` in a resource message is also a resource name field, but does not have `resource_reference` annotation.

I found this while generating the method [CreateSubscription](https://github.com/googleapis/googleapis/blob/638253bf86d1ce1c314108a089b7351440c2f0bf/google/pubsub/v1/pubsub.proto#L375) for PubSub. The pattern is strongly discouraged for new APIs; but we can't break PubSub.

### DIFF
--- a/src/test/java/com/google/api/codegen/protoannotations/testdata/java_library.baseline
+++ b/src/test/java/com/google/api/codegen/protoannotations/testdata/java_library.baseline
@@ -4288,6 +4288,7 @@ import com.google.api.resourcenames.ResourceName;
 import com.google.common.base.Function;
 import com.google.common.collect.Iterables;
 import com.google.common.util.concurrent.MoreExecutors;
+import com.google.example.library.v1.Book;
 import com.google.example.library.v1.TestOptionalRequiredFlatteningParamsRequest;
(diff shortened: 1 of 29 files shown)


src/test/java/com/google/api/codegen/protoannotations/testdata/java_library.baseline
```diff
@@ -9192,6 +9193,121 @@ public class LibraryClient implements BackgroundResource {
...
+   *
+   * @param name The resource name of the book.
+   * Book names have the form `bookShelves/{shelf_id}/books/{book_id}`.
+   * Message field comment may include special characters: &lt;&gt;&amp;"`'{@literal @}.
+   * @param author The name of the book author.
+   * @param title The title of the book.
+   * @param rating For testing enums.
+   * @throws com.google.api.gax.rpc.ApiException if the remote call fails
+   */
+  public final void saveBook(BookName name, String author, String title, Book.Rating rating) {
```
hzyi-google: This is the only baseline change that matters... everything else is caused by adding an RPC.

PR #3122, opened 2020-03-06
bazel: Update `gapic_assembly_pkg` rules surface

vam-google:
Add `assembly_name` attribute. This allows us to simplify the `name` attribute value and as result simplify and unify command used to generate the final package.

For example
coommand before:
```
bazel build //google/cloud/language/v1:google-cloud-language-v1-csharp
```

command after:
```
bazel build //google/cloud/language/v1:csharp
```
The changes are **backward-compatible**

### DIFF
--- a/rules_gapic/csharp/csharp_gapic_pkg.bzl
+++ b/rules_gapic/csharp/csharp_gapic_pkg.bzl
@@ -74,11 +74,14 @@ _csharp_gapic_src_pkg = rule(
     implementation = _csharp_gapic_src_pkg_impl,
 )
 
-def csharp_gapic_assembly_pkg(name, deps, **kwargs):
+def csharp_gapic_assembly_pkg(name, deps, assembly_name = None, **kwargs):
+    package_dir = name
+    if assembly_name:
+        package_dir = "google-cloud-%s-%s" % (assembly_name, name)
     _csharp_gapic_src_pkg(
         name = name,
         deps = deps,
-        package_dir = name,
+        package_dir = package_dir,
         **kwargs
     )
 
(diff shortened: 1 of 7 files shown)


rules_gapic/csharp/csharp_gapic_pkg.bzl
```diff
@@ -74,11 +74,14 @@ _csharp_gapic_src_pkg = rule(
     implementation = _csharp_gapic_src_pkg_impl,
 )
 
-def csharp_gapic_assembly_pkg(name, deps, **kwargs):
+def csharp_gapic_assembly_pkg(name, deps, assembly_name = None, **kwargs):
+    package_dir = name
+    if assembly_name:
+        package_dir = "google-cloud-%s-%s" % (assembly_name, name)
```
noahdietz: Is it possible that Ads wants to use this new feature, but we are encoding "cloud" into the package names?

jpoehnelt:
>This seems fine to me, just one concern about cross-org usage i.e. non-Cloud GAPICs (Ads).

Seconded

vam-google:
@noahdietz @jpoehnelt This change is backward compatible. The `assembly_name` field is optional and if someboyd does not specify it, the name is used as is (i.e. wthout enforced "cloud" prefix in it). So Ads, Maps and others should be good.

rules_gapic/csharp/csharp_gapic_pkg.bzl
vam-google: Ads may keep using it in the backward-compatible way, i.e. omitting the `assembly_name` prefix such that the `name` itself will be used for package_dir (and they can specify anything there with or without cloud prefix).

PR #3128, opened 2020-03-06
fix: bring setup.py versions inline with nox versions

jpoehnelt:
This updates the classifiers with Python 3.7. This should be a safe default as we currently run unittests with nox against 3.7.

https://github.com/googleapis/gapic-generator/blob/06463219f95bedb0f9c788dffa888c5ebfc199cc/src/main/resources/com/google/api/codegen/py/nox.py.snip#L34

Additionally removes python 3.4 for which [maintenance ended in March 2019](https://www.python.org/downloads/release/python-3410/). 

Python 3.8 is still blocked by protobuf https://github.com/googleapis/google-cloud-python/issues/9878.


alexander-fenster:
@software-dov PTAL :)

jpoehnelt:
rebased the branch, PTAL :)

jpoehnelt:
@software-dov PTAL 🏓

PR #3138, opened 2020-03-11
Assign the proto file where the interface is defined when generating samples and tests for arbitrary resource names

hzyi-google:
Follow-up of #3027

When picking a `SingleResourceNameConfig` to use to generate samples for `AnyResourceNameConfig`, we need to assign the correct proto file to the `SingleResourceNameConfig` so that we know the correct full classpath of the generated resource name class.

#3027 uses the package where the method is defined, which fixed a common case of IAM methods where request objects are imported from the IAM package.

That did not cover the case for PubSub where the IAM methods are added by `reroute_to_grpc_interface` in gapic config. In this case the method is also defined in IAM package.

This PR change the behavior to assigning the proto file where the interface we are generating is defined, which should cover both cases.

Caught this when trying to generate PubSub from gapic v2. Had the following error:

```
ERROR: /tmpfs/src/piper/google3/third_party/googleapis/google/pubsub/v1/BUILD.bazel:69:1: Building google/pubsub/v1/libpubsub_java_gapic_test.jar (1 source jar) failed (Exit 1)
/com/google/cloud/pubsub/v1/TopicAdminClientTest.java:438: error: cannot find symbol
    ResourceName resource = com.google.iam.v1.ProjectName.of("[PROJECT]");
                                             ^
  symbol:   class ProjectName
  location: package com.google.iam.v1
```

### DIFF
--- a/src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
+++ b/src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
@@ -16556,7 +16556,6 @@ namespace Google.Example.Library.V1
         /// </summary>
         /// <param name="resource">
         /// REQUIRED: The resource which the label is being added to.
-        /// In the form "shelves/{shelf_id}/books/{book_id}".
         /// </param>
         /// <param name="label">
         /// REQUIRED: The label to add.
@@ -16583,7 +16582,6 @@ namespace Google.Example.Library.V1
         /// </summary>
         /// <param name="resource">
         /// REQUIRED: The resource which the label is being added to.
-        /// In the form "shelves/{shelf_id}/books/{book_id}".
         /// </param>
         /// <param name="label">
         /// REQUIRED: The label to add.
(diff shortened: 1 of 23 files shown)


src/test/java/com/google/api/codegen/gapic/testdata/csharp/csharp_library.baseline
```diff
@@ -16556,7 +16556,6 @@ namespace Google.Example.Library.V1
         /// </summary>
         /// <param name="resource">
         /// REQUIRED: The resource which the label is being added to.
-        /// In the form "shelves/{shelf_id}/books/{book_id}".
```
vam-google: Why are these deleted all over the place? (i.e. please explain the connection between the code change and this change in the generated comments)

hzyi-google: I changed the pattern of the field `resource` from `Book` to `*` to repro this issue: https://github.com/googleapis/gapic-generator/pull/3138/files#diff-4c8eaf76addf344f8abf6b2800029f9eR51

And the comment is no longer correct... so I deleted the comment, hence the change is baseline.

hzyi-google:
This isn't a breaking change; the change only affects how gapic generator creates a resource name for an AnyResourceName, which is only in samples and request object construction part of unit tests. Nothing on the API surface changes.

PR #3143, opened 2020-03-16
PHP multi-pattern resource name support without gapic config v2

hzyi-google:
Generates a resource name helper for each pattern of a multi-pattern resource name defined in proto annotations, so that PHP generation does not have to rely on gapic v2.

To sum up, when a library is generated from gapic v2 + proto annotations, we generate the following resource name helpers in PHP:
1. one for every single-pattern resource name
2. the first pattern of a multi-pattern resource, as if it is a single-pattern resource (for backward-compatibility when a single-pattern resource gains a new pattern, and becomes a multi-pattern one)
3. one for every pattern of a multi-pattern resource name,  marked as `@experimental` so that we can change the name if needed
4. one for each entry in gapic config v2 (for backward-compatibility with libraries generated from gapic config v1), marked as `@deprecated` so that we will remove them after we remove gapic v2.

Please see the generated outcome for multi-pattern resource `Book` as well as `Reader`.


hzyi-google:
PTAL

PR #3144, opened 2020-03-19
feat: take source protos in all sub-packages

hzyi-google:
Mainly for ads need but may be useful for all APIs in general. All sub-packages of the `package` flag passed to gapic-generator from either artman or bazel will be considered source protos to generate gapics for. This is needed beause ads groups their messages and services into different sub-directories: https://github.com/googleapis/googleapis/tree/master/google/ads/googleads/v3.

We mainly need to know what messages belong to this GAPIC to know what resources we should generate helpers for and are legal to be referenced for now, but enabling generating gapics from multiple packages is useful in general.

Tested with Ads build locally and both generation and tests pass.


hzyi-google:
Confirmed that 3p/googleapis build succeeded.

PR #3146, opened 2020-03-23
feat: allow static substitution for group name

jpoehnelt:
The build.gradle file `group = 'com.google.cloud'` does not work for other groups without postprocessing. The approach here is to allow the macro to pass the substitutions dict. Alternatively, I could bring the group name up as an argument to the macro.


jpoehnelt:
@vam-google PTAL

PR #3147, opened 2020-03-25
Revert "feat: allow static substitution for group name"

jpoehnelt:
This caused a regression. I'll look into adding a test before adding again.


jpoehnelt:
@vam-google @noahdietz ptal

jpoehnelt:
There is a path that the default substitution variable was not set and the value rendered as {{group}}. My GAPIC github action caught this in a pr: https://github.com/googlemaps/java-routes/pull/6/files

PR #3148, opened 2020-03-25
bazel: go build gen check for cloud in pkg name

noahdietz:
The assumption that Cloud APIs will have the Cloud-specific auth scope was not entirely true. I'm not sure if the API I encountered was incorrect or what, but we can also check for `cloud` in the proto package name. This should give us coverage of APIs that either have the Cloud-specific auth scope or are actually placed in the Cloud uber-package.

### DIFF
--- a/rules_gapic/bazel/src/main/java/com/google/api/codegen/bazel/BazelBuildFileView.java
+++ b/rules_gapic/bazel/src/main/java/com/google/api/codegen/bazel/BazelBuildFileView.java
@@ -94,7 +94,7 @@ class BazelBuildFileView {
     goPkg = goPkg.replaceFirst("cloud\\/", "");
 
     String goImport = "";
-    if (isCloud) {
+    if (isCloud || protoPkg.contains("cloud")) {
       goImport = "cloud.google.com/go/";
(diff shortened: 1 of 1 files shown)


rules_gapic/bazel/src/main/java/com/google/api/codegen/bazel/BazelBuildFileView.java
```diff
@@ -94,7 +94,7 @@ private String assembleGoImportPath(boolean isCloud, String protoPkg, String goP
     goPkg = goPkg.replaceFirst("cloud\\/", "");
 
     String goImport = "";
-    if (isCloud) {
+    if (isCloud || protoPkg.contains("cloud")) {
```
vam-google: This looks weird. `isCloud` should define if it is a cloud or no. Here it does it partially, and it can be overriden by protoPkg value (which also an argument to this function).

Please make sure that isCloud completely defines the cloud thing. (i.e. it an be as straightforward as moving `protoPkg.contains("cloud")` from here to the place where `assembleGoImportPath` is defined.

noahdietz: Ah, good call. I did just that! Done.

PR #3150, opened 2020-03-27
buildgen: tighten gopkg version regexp

noahdietz:
The original regexp was naive and didn't account for the presence of a `v` in earlier in the path than the version. For example `bigquery/reservation/v1;reservation` would become `bigquery/reservapiation/v1;reservation`.

Asserting that the matching group starts with `\\/` and restricting the capture group to `a-z0-9` (rather than anything) fixes this.

### DIFF
--- a/rules_gapic/bazel/src/main/java/com/google/api/codegen/bazel/BazelBuildFileView.java
+++ b/rules_gapic/bazel/src/main/java/com/google/api/codegen/bazel/BazelBuildFileView.java
@@ -96,7 +96,7 @@ class BazelBuildFileView {
     String goImport = "";
     if (isCloud) {
       goImport = "cloud.google.com/go/";
-      goPkg = goPkg.replaceFirst("\\/v([a-z1-9]+);", "\\/apiv$1;");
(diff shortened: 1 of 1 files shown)


rules_gapic/bazel/src/main/java/com/google/api/codegen/bazel/BazelBuildFileView.java
```diff
@@ -96,7 +96,7 @@ private String assembleGoImportPath(boolean isCloud, String protoPkg, String goP
     String goImport = "";
     if (isCloud) {
       goImport = "cloud.google.com/go/";
-      goPkg = goPkg.replaceFirst("v(.+);", "apiv$1;");
+      goPkg = goPkg.replaceFirst("\\/v([a-z1-9]+);", "\\/apiv$1;");
```
vam-google: what about 0? Version probably can' start with 0, but v10 is theoretically possible.

noahdietz: Good idea. Done!

noahdietz:
> Is circleci stuck?

Nope, I just merged `master` into this branch so it's running again.

PR #3151, opened 2020-03-28
feat: trace up only one level when calculating parent types with singleton resource names 

hzyi-google:
parent type of `projects/{project}/foo/bar/{bar}` would be `projects/{project}/foo`.
parent type of `projects/{project}/foo` would be `projects/{project}`.

This is to support dialogflow.

### DIFF
--- a/src/main/java/com/google/api/codegen/config/ResourceDescriptorConfig.java
+++ b/src/main/java/com/google/api/codegen/config/ResourceDescriptorConfig.java
@@ -271,15 +271,16 @@ public abstract class ResourceDescriptorConfig {
   @VisibleForTesting
   static String getParentPattern(String pattern) {
     List<String> segments = getSegments(pattern);
-    int index = segments.size() - 2;
-    while (index >= 0 && !isVariableBinding(segments.get(index))) {
+    int index = segments.size() - 1;
+    if (isVariableBinding(segments.get(index))) {
+      index -= 2;
+    } else {
       index--;
     }
-    index++;
-    if (index <= 0) {
+    if (index < 0) {
       return "";
     }
-    return String.join("/", segments.subList(0, index));
+    return String.join("/", segments.subList(0, index + 1));
   }
 
   private static Map<String, Boolean> getParentPatternsMap(ResourceDescriptorConfig resource) {
--- a/src/main/java/com/google/api/codegen/config/ResourceDescriptorConfig.java
+++ b/src/main/java/com/google/api/codegen/config/ResourceDescriptorConfig.java
@@ -270,16 +270,19 @@ public abstract class ResourceDescriptorConfig {
 
   @VisibleForTesting
   static String getParentPattern(String pattern) {
+    Preconditions.checkArgument(!pattern.equals(""), "resource pattern can't be an empty string.");
+    if (pattern.equals("*")) {
+      return "*";
+    }
+
     List<String> segments = getSegments(pattern);
     int index = segments.size() - 1;
-    if (index < 0) {
-      return "";
-    }
     if (isVariableBinding(segments.get(index))) {
       index -= 2;
     } else {
       index--;
     }
+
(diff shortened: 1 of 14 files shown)


src/main/java/com/google/api/codegen/config/ResourceDescriptorConfig.java
```diff
@@ -271,15 +271,16 @@ private String getUnqualifiedTypeName() {
   @VisibleForTesting
   static String getParentPattern(String pattern) {
     List<String> segments = getSegments(pattern);
-    int index = segments.size() - 2;
-    while (index >= 0 && !isVariableBinding(segments.get(index))) {
+    int index = segments.size() - 1;
```
jskeet: Is this method ever called for a pattern of "*"? If so, it should return "*" at that point.

src/main/java/com/google/api/codegen/config/ResourceDescriptorConfig.java
```diff
@@ -271,15 +271,16 @@ private String getUnqualifiedTypeName() {
...
+    if (isVariableBinding(segments.get(index))) {
+      index -= 2;
+    } else {
       index--;
     }
-    index++;
-    if (index <= 0) {
+    if (index < 0) {
```
jskeet: Does this report an error?

src/main/java/com/google/api/codegen/config/ResourceDescriptorConfig.java
hzyi-google: That's a good point!

src/main/java/com/google/api/codegen/config/ResourceDescriptorConfig.java
hzyi-google: Done

src/main/java/com/google/api/codegen/config/ResourceDescriptorConfig.java
```diff
@@ -272,14 +272,16 @@ private String getUnqualifiedTypeName() {
...
     if (isVariableBinding(segments.get(index))) {
       index -= 2;
     } else {
       index--;
     }
-    if (index < 0) {
-      return "";
-    }
+    Preconditions.checkArgument(
+        index >= -1, "malformatted pattern, can't calculate parent: %s", pattern);
```
jskeet: I suspect this should be `index >= 0` - as otherwise we get an *empty* pattern, which still isn't a useful parent.
(Basically, taking the root of "foo" for "foos/{foo}" is invalid, I'd say. The empty pattern can't represent a resource name, so it's not useful.)

hzyi-google: That is correct from the public-facing point of view - but this is an internal method. We need to call this method on every resource pattern to find if it has a parent, if so, which resource does the parent belong to. Therefore we can't error if the calculated parent is empty (for instance "projects/{project}". I guess we are sort of using an empty string as a way to express "this pattern does not have a parent pattern".

jskeet: Ah, okay - in C# I believe we only find parents when we're asked to :)

hzyi-google: Hi Jon, can you approve the PR if you don't think there are other issues?

jskeet: There's still the matter of the parent - of `"*"` - which wasn't clear in my previous comment due to Markdown, unfortunately. Basically I'd like to see a test showing that if you pass in `"*"`, you get `"*"` back - and code to make that work.

hzyi-google: Oh that solves all the puzzle now. I will fix `"*"` shortly.

PR #3158, opened 2020-04-02
fix: align text and fill in empty copyright owner

jpoehnelt:


jpoehnelt:
@hzyi-google ptal

src/main/resources/com/google/api/codegen/LICENSE.snip
```diff
@@ -187,7 +187,7 @@
           same "printed page" as the copyright notice for easier
           identification within third-party archives.
     
-       Copyright [yyyy] [name of copyright owner]
+       Copyright 2020 Google LLC
```
hzyi-google, line 190: Thanks for contributing!

However I think the brackets are intentional as a few lines above, it says
```
To apply the Apache License to your work, attach the following
          boilerplate notice, with the fields enclosed by brackets "[]"
          replaced with your own identifying information. (Don't include
```

We've already applied such rule to all generated source code.

jpoehnelt: It doesn't make it through rules_gapic. I just tested py_gapic_assembly_pkg.

hzyi-google: So we actually don't need it at all - as long as the generated source code have licenses as comments on the top, we don't need a separate license template.

jpoehnelt: I'll revert the brackets, would be nice to keep the text aligned in the license template.

hzyi-google: Sounds good. Thanks!

jpoehnelt: new commit with those changes reverted

hzyi-google:
LGTM

(will merge after tests pass)

PR #3161, opened 2020-04-03
fix(python): add PROJECT_ID for python-system-test

busunkim96:


hzyi-google:
@busunkim96 looks like the system test passed!

PR #3162, opened 2020-04-06
feat: add the last segment when generating names for singleton resource pattern

hzyi-google:
This is to support [SecurityCenter](https://github.com/googleapis/googleapis/blob/master/google/cloud/securitycenter/v1/security_marks.proto#L33-L38).

`SecurityMarks` is a multi-pattern singleton resource:

```
option (google.api.resource) = {
    type: "securitycenter.googleapis.com/SecurityMarks"
    pattern: "organizations/{organization}/assets/{asset}/securityMarks"
    pattern: "organizations/{organization}/sources/{source}/findings/{finding}/securityMarks"
};
```

Currently we generate an ID for a pattern by concatenating all binding values. This leads to slightly unclear helper functions in Java and PHP for the `SecurityMarks` resource:

```
// PHP
client.organizationAssetName("my-org", "ast"); // returns organizations/my-org/assets/ast/securityMarks

// Java
SecurityMarksName.ofOrganizationAssetName("org", "ast");
```
This PR adds the last segment to the resource name pattern ID if the last segment is not a binding value. This will lead to more descriptive function names:

```
// PHP
client.organizationAssetSecurityMarksName("my-org", "ast"); // returns organizations/my-org/assets/ast/securityMarks

// Java
SecurityMarksName.ofOrganizationAssetSecurityMarksName("org", "ast");
```

### DIFF
--- a/src/main/java/com/google/api/codegen/config/ResourceNamePatternConfig.java
+++ b/src/main/java/com/google/api/codegen/config/ResourceNamePatternConfig.java
@@ -98,7 +98,7 @@ public class ResourceNamePatternConfig {
     String patternId = getBindingVariables().stream().collect(Collectors.joining("_"));
 
     String[] segments = pattern.split("/");
-    checkState(segments.length > 0, "internal: pattern %s is fixed pattern.", pattern);
+    checkState(segments.length > 1, "internal: pattern %s is fixed pattern.", pattern);
     String lastSegment = segments[segments.length - 1];
 
     // Singleton resource. Append the last segment.
--- a/src/main/java/com/google/api/codegen/config/ResourceNamePatternConfig.java
+++ b/src/main/java/com/google/api/codegen/config/ResourceNamePatternConfig.java
@@ -14,9 +14,11 @@
  */
 package com.google.api.codegen.config;
 
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkState;
+
 import com.google.api.codegen.util.Name;
 import com.google.api.pathtemplate.PathTemplate;
-import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableSet;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
@@ -84,7 +86,6 @@ public class ResourceNamePatternConfig {
    * pattern with underscores.
    */
   public String getPatternId() {
-    // TODO(hzyi): support singleton resource name patterns
     if (isFixedPattern()) {
       String name = pattern.replaceAll("^[^a-zA-Z]+", "");
       name = name.replaceAll("[^a-zA-Z]$", "");
@@ -94,7 +95,24 @@ public class ResourceNamePatternConfig {
 
     // PathTemplate uses an ImmutableMap to keep track of bindings, so we
     // can count on it to give us the correct order of binding variables
-    return getBindingVariables().stream().collect(Collectors.joining("_"));
+    String patternId = getBindingVariables().stream().collect(Collectors.joining("_"));
+
+    String[] segments = pattern.split("/");
+    checkState(segments.length > 0, "internal: pattern %s is fixed pattern.", pattern);
+    String lastSegment = segments[segments.length - 1];
+
+    // Singleton resource. Append the last segment.
+    if (!lastSegment.startsWith("{") || !lastSegment.endsWith("}")) {
+
+      checkArgument(
+          !lastSegment.startsWith("{") && !lastSegment.endsWith("}"),
+          "bad format: segment %s of resource pattern %s has unmatching braces",
+          lastSegment,
+          pattern);
+
+      patternId = patternId + '_' + Name.anyLower(lastSegment).toLowerUnderscore();
+    }
+    return patternId;
   }
 
   /**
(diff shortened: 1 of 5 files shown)


dwsupplee:
@hzyi-google Thanks for this. Should we expect this to be breaking or additive?

hzyi-google:
@dwsupplee This will be breaking for SecurityCenter I think. However I think the breaking changes are on `@experimental` methods so there shouldn't be any need to do major version bumps.

hzyi-google:
@alexander-fenster CI question...

I'll need https://github.com/googleapis/protoc-java-resource-names-plugin/pull/83 to make `java-openjdk8-test` pass. However I can't release a new artman if I don't merge this PR...

alexander-fenster:
@hzyi-google Let me know which checks to exclude for which repos, and I can temporary disable them.

hzyi-google:
@alexander-fenster thanks! I'll ping you after I get an approval on this PR.

dwsupplee:
I like adding `securityMarks` to the name as it helps provide context to users. With the existing naming it isn't clear where or why you would use this helper.

jskeet:
> I like adding securityMarks to the name as it helps provide context to users. With the existing naming it isn't clear where or why you would use this helper.

But it removes consistency. If everywhere there's `FromFooBar` you pass in a Foo and a Bar, isn't it confusing to now have `FromFooBarBaz` that *doesn't* take a Foo, Bar and Baz - just Foo and Bar?

dwsupplee:
I don't think I would find that too confusing because the signature and docs would provide me all the detail I need about what I can pass in. Also, we already have had curated entity names for formatting methods that don't match patterns so doing something like this isn't completely new.

There are lots of trade-offs with the approach we have taken with the helpers in the GAPICs, especially now that we plan to no longer use curated entity names. I lean towards this trade-off being a decent compromise to help guide users given the circumstances.

jskeet:
I know this will be (another) breaking change for Dialogflow. Do we know what other APIs it would affect for existing patterns?

> I don't think I would find that too confusing because the signature and docs would provide me all the detail I need about what I can pass in

But not why it's got a confusing name. If I'm looking for something that only accepts a Foo and a Bar, and I see a method that I *expect* to take more than I want to pass, I may well not even see that documentation.

I suspect this problem is worse in C# than other languages precisely because we chose the factory method name based on "that describes what things you pass in". One option would be to make the change for other languages but not for C#.  That's a sort of "last resort" option though, I think.

Has @chrisdunelm looked at this?

dwsupplee:
> But not why it's got a confusing name. If I'm looking for something that only accepts a Foo and a Bar, and I see a method that I expect to take more than I want to pass, I may well not even see that documentation.

Yes, it is unfortunate we have decided to not use curated entity names and I agree it is confusing (although, I understand the need for the move). There are arguments not doing this change makes it confusing as well, just in another way. I'm not sure there is a clear cut "ideal" solution here, really.

chrisdunelm:
I have a slight preference to not make this change.

In the Java case, it doesn't appear too confusing currently as the required context is provided by the resource-name class on which the static method is being called - `SecurityMarksName` in this example. I appreciate that the naming might not always be this consistent, but it's likely to be in the majority of cases.

I agree PHP is currently less than ideal, but agree with Jon that it's good to be consistent with the naming of the method is based on what is passed in.

Although keeping languages consistent with each other isn't an absolute requirement, it is certainly helpful, so I'd rather see all languages keep the same naming convention for these methods.

hzyi-google:
While this PR is mostly for PHP, I'll drop my $0.02 here:

<h2>Compatibility</h2>

All changed public methods have been guarded by `@experimental` in PHP or `@betaApi` in Java. So we don't have breaking change concern.

<h2>Per-language Naming</h2>

**PHP: prefer to change**

I think adding the last segment would result in a better method name in PHP, and may even be necessary in some possible cases in the future. Unlike Java and C# where each resource name has a separate class, PHP resource name helpers are generated on Gapic client surface. Thus even if appending the last segment breaks the consistency where the method name tells about formatting arguments, it is more clear in the sense that what resource name this method returns.

Also, in some extreme case (we don't have this problem now but it's a possible scenario), PHP generation will fail if we have such two patterns, because under current strategy they will have the same formatting method name.
```
projects/{project}
projects/{project}/foo
```

**C#: prefer not to change**

I think the existing method names are clear and consistent. C# doesn't need to change anything.

**Java: slightly prefer to change**

I agree that the existing names are not confusing and usable (because resource names are in separate classes), but changing to append the last segment would be slightly preferable. The main reason Java is different from C# is that Java used the `of` convention, and Java always has a `Name` suffix for these helpers. For example I really think

 `SecurityMarksName.ofProjectAssetSecurityMarksName`

makes more sense than

`SecurityMarksName.ofProjectAssetName`,

because the latter would make me think I am creating an `AssetName`. While in C#, the `From` prefix makes it clear that the remaining parts of the method name are strings to be used to create the resource name.

<h2>Cross-language Consistency</h2>

The way different languages treat multi-pattern resource names is already different - For example, Ruby and Node.js decided to take a hash rather than generate one helper for each pattern, and Python decided to only generate helpers for the first pattern. I would say this is less of a concern now (and we have different generators for different languages anyway).

jskeet:
Thanks for the great summary, Hanzhen.

I'm "just about okay" with the divergence here - and I suspect it's possible our users will never even notice :)

I don't know who the relevant decision maker is though.

hzyi-google:
@chingor13 could you please see if Java baseline changes are fine?

hzyi-google:
Discussed offline: this is OK in Java.

hzyi-google:
@alexander-fenster cl 305738852 tested the change with third_party presubmit. Could you please turn off java-openjdk8-test temporarily?

PR #3163, opened 2020-04-07
feat: support '*' in resource definition

hzyi-google:
This is to support Asset.

If referencing a resource with only one `*` pattern, treats it the same as `resource_reference.type = "*"`. That is, do not generate any resource name class for it, and use `ResourceName` interface on the client surface.

### DIFF
--- a/src/main/java/com/google/api/codegen/config/ResourceNameMessageConfigs.java
+++ b/src/main/java/com/google/api/codegen/config/ResourceNameMessageConfigs.java
@@ -174,32 +174,6 @@ public abstract class ResourceNameMessageConfigs {
                 "unknown resource name type: " + resource.getResourceNameType());
         }
       }
-
-      // if (!childType.isEmpty()) {
-      //   List<ResourceDescriptorConfig> parents =
-      //       childParentResourceMap.getOrDefault(childType, Collections.emptyList());
-      //   for (ResourceDescriptorConfig parentResourceDescriptor : parents) {
-      //     String derivedEntityName = parentResourceDescriptor.getDerivedEntityName();
-      //     ResourceNameConfig parentResource = resourceNameConfigs.get(derivedEntityName);
-      //     Preconditions.checkArgument(
-      //         parentResource != null, "Referencing non-existing parent resource: %s", childType);
-      //     fieldEntityMap.put(field.getSimpleName(), parentResource.getEntityId());
-      //   }
-      //   continue;
-      // }
-
-      // String unqualifiedResourceType = ResourceDescriptorConfig.getUnqualifiedTypeName(type);
-      // ResourceNameConfig resourceNameConfig =
-      //     resourceNameConfigs.get(unqualifiedResourceType + "Oneof");
-      // if (resourceNameConfig != null
-      //     && resourceNameConfig.getResourceNameType() == ResourceNameType.ONEOF) {
-      //   fieldEntityMap.put(field.getSimpleName(), unqualifiedResourceType + "Oneof");
-      //   continue;
-      // }
-      // resourceNameConfig = resourceNameConfigs.get(unqualifiedResourceType);
-      // Preconditions.checkArgument(
-      //     resourceNameConfig != null, "Referencing non-existing resource: %s", type);
-      // fieldEntityMap.put(field.getSimpleName(), unqualifiedResourceType);
     }
   }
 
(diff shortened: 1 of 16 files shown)


hzyi-google:
This wouldn't be a breaking change. The design is as follows. I'll add baseline tests (and implementation if needed) to demo the case where a resource has both `*` and a normal pattern, but in a separate PR because
 1) This PR is enough to unblock Asset
 2) Too much changes in baseline would be confusing and hard to review.

Say we have:
```protobuf
option (google.api.resource_definition) = {
  type: ".../Asset",
  pattern: "*"
}

message CreateAssetRequest {
  string name = 1 [ (...).type = ".../Asset"];
}

rpc CreateAsset(CreateAssetRequest) returns (CreateAssetResponse) {
  option (...method_signature) = "name";
}
```
We will generate:

```java
public CreateAssetResp createAsset(CreateAssetRequest request);
public CreateAssetResp createAsset(ResourceName name);
```

Later if `Asset` gets a new pattern:

```protobuf
option (google.api.resource_definition) = {
  type: ".../Asset",
  pattern: "*",
  pattern: "projects/{}/assets/{}"
}
```

We will generate:
```java
public CreateAssetResp createAsset(CreateAssetRequest request);
public CreateAssetResp createAsset(ResourceName name);
public CreateAssetResp createAsset(AssetName name);
```
and `AssetName` is a single-pattern resource name.

Later if `Asset` gets another pattern:
```protobuf
option (google.api.resource_definition) = {
  type: ".../Asset",
  pattern: "*",
  pattern: "projects/{}/assets/{}",
  pattern: "organization/{}/assets/{}"
}
```

We will generate the same methods on Gapic surface, but `AssetName` becomes a two-pattern resource name (from single-pattern to multi-pattern is not a breaking change).

jskeet:
Okay, that sounds reasonable.

I don't understand what's broken in Asset at the moment for Java, given that C# generates it fine.

@chrisdunelm Is this actually just the C# existing behaviour?

chrisdunelm:
I haven't looked at this in enourmous detail, but I'm fairly sure this is exactly what is already implemented in C#.

hzyi-google:
@jskeet because Java hasn't implemented specifying `*` for resource name patterns yet :p for now Java thinks `*` is a fix-valued pattern and the generation blew up. I found this while migrating the library to annotations (it's still in gapic v1 now).

Asset is the first API of such cases :)

jskeet:
Okay. With that extra context I'll have another look tomorrow

src/main/java/com/google/api/codegen/config/ResourceNameMessageConfigs.java
```diff
@@ -134,36 +140,40 @@ private static void loadFieldEntityPairFromResourceReferenceAnnotation(
...
+          case ANY:
+            fieldEntityMap.put(field.getSimpleName(), "*");
+            break;
+          case SINGLE:
+            fieldEntityMap.put(field.getSimpleName(), unqualifiedName);
+            break;
+          case ONEOF:
+            fieldEntityMap.put(field.getSimpleName(), unqualifiedName);
+            if (((ResourceNameOneofConfig) resource).hasAnyResourceNamePattern()) {
+              fieldEntityMap.put(field.getSimpleName(), "*");
```
jskeet, line 169: I think it might be clearer to move the earlier `put` call to an `else` block - so that every branch of this switch statement calls `fieldEntityMap.put` exactly once. (You could even extract this to a method which just returns "the value to put in the map" so you only have a single `fieldEntityMap.put` call.)

hzyi-google: I might be missing your point Jon, but the former `put` needs to be executed regardless of whether the resource has a `*` pattern, so I can't move it to the else block.

jskeet: You've currently got (pseudo-code)

```text
put (key, value1)
if (condition)
  put (key, value2)
```

I'm suggesting that's better as:

```text
if (condition)
  put (key, value2)
else
  put (key, value1)
```

So it's easy to see that in every path, the key is put exactly once. What's the benefit of putting `unqualifiedName` as the value and *then* putting `"*"` as the value? That's just overwriting it, right?

hzyi-google: Ahh I see the confusion here. `fieldEntityMap` is a `Multimap`- it supports duplicate keys.

https://github.com/googleapis/gapic-generator/pull/3163/files/74feb00d6b94ff04fed06c784bfcd826529246b3#diff-a0f27b8e4c41d2ebf887482b72314a58R82

jskeet: Ah, that would definitely explain it. Thanks for clarifying.

PR #3165, opened 2020-04-08
feat: Support extra plugin_args for php bazel rules

vam-google:


alexander-fenster:
So...... do we have tests that cover that? :)

vam-google:
@alexander-fenster I've ran the whole build on googleapis and it passed.

Issue #3167, opened 2020-04-08
Java: breaking change when a string becomes a resource name


yihanzhen:
In Java, if a method has fields referencing resource names, we generate overloads that cover all combinations of resource name types as well as an extra one where all fields are treated as strings. This may cause breaking changes if a ordinary string becomes a resource name string:

For example, if we used to have:

```protobuf
rpc CreateFoo(CreateFooRequest) returns (CreateFooResponse) {
  option (google.api.method_signature) = "foo,bar";
}

message CreateFooRequest {
  string foo = 1 [
    (google.api.resource_reference).type = "foo.googleapis.com/Foo"];

  string bar = 2;
}
```

we generate:
```java
CreateFooResponse createFoo(CreateFooRequest request);
CreateFooResponse createFoo(FooName foo, String bar);
CreateFooResponse createFoo(String foo, String bar);
```

Now we want to make `bar` a resource name, or we realized we forgot to annotate `bar`, we change the request message to:

```protobuf
message CreateFooRequest {
  string foo = 1 [
    (google.api.resource_reference).type = "foo.googleapis.com/Foo"];

  string bar = 2  [
    (google.api.resource_reference).type = "foo.googleapis.com/Bar"];
}
```

And we will generate:
```java
CreateFooResponse createFoo(CreateFooRequest request);
CreateFooResponse createFoo(FooName foo, BarName bar);
CreateFooResponse createFoo(String foo, String bar);
```

This is problematic: the signature of the second method is changed and thus causing a breaking change.

Proposing: we generate overloads that covers combinations where each field can be a resource name type or a string.

So for the second case above, we generate:

```
CreateFooResponse createFoo(CreateFooRequest request);
CreateFooResponse createFoo(FooName foo, BarName bar);
CreateFooResponse createFoo(FooName foo, String bar);
CreateFooResponse createFoo(String foo, BarName bar);
CreateFooResponse createFoo(String foo, String bar);
```

miraleung:
Discussion resolved in #3177.

PR #3177, opened 2020-04-16
[gapic-gen] fix: generate string overloads for resource names method params

miraleung:
Please see https://github.com/googleapis/gapic-generator/issues/3167 for
more details.

### DIFF
--- a/src/main/java/com/google/api/codegen/config/FlatteningConfig.java
+++ b/src/main/java/com/google/api/codegen/config/FlatteningConfig.java
@@ -280,6 +280,13 @@ public abstract class FlatteningConfig {
     if (hasAnyResourceNameParameter(flatteningConfigs)) {
       flatteningConfigs.add(withResourceNamesInSamplesOnly(flatteningConfigs.get(0)));
     }
+
+    // Generate an overload where only one of the resource names are treated as strings,
+    // if there are two or more resource names in the method surface.
+    if (!flatteningConfigs.isEmpty()) {
+      flatteningConfigs.addAll(slidingStringResourceNameConfigs(flatteningConfigs));
+    }
+
     return flatteningConfigs
         .stream()
         .map(ImmutableMap::copyOf)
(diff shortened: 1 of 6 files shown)


hzyi-google:
> we declare adding/removing resource name reference to a flattened field as a breaking change

This needs to run through other languages as well.

miraleung:
This currently breaks only vision beta libraries, so if they are OK with these breakages and other languages are fine with this, then that declaration SGTM.

vam-google:
@miraleung I guess braking beta libraries should be Ok, especially if not breaking them is proven to be problematic. 

For the other langauges part, I've clarified with @hzyi-google what he meant by his comment. I guess we need other language owners agree with the "proposal". 

@software-dov 
@chrisdunelm 
@viacheslav-rostovtsev
@noahdietz 
@alexander-fenster  

What do you think if we  **mark adding resource name to a field a potentially breaking change**?

chrisdunelm:
I agree that we should not generate all combinations of string/resource-name parameters. We've definitely need to avoid the 2^n issue.
Simply "declare adding/removing resource name reference to a flattened field as a breaking change" does avoid this, and requires no code changes in any generator.
However, it's less than ideal, and may cause difficulty for API producer teams; as an API grows in size and number of contributors it becomes more likely mistakes will be made.

We should look to see if there's a better solution here, but I agree that in the short-term at least we should make this a breaking change.
As others have said, for a beta release a breaking change like this is generally OK.

viacheslav-rostovtsev:
I agree with Chris above that generating all combinations is a bad solution -- it'll clutter all the APIs unnecessarily. Making it a breaking change seems OK and we can re-address this if it turns out a common re-occuring problem.

noahdietz:
I agree with Chris and Slava, but in practice I don't see it being all that disastrous. In my experience (albeit limited) there are few APIs that have multiple resource_references in a single request message, mainly because RPCs typically work with a single resource (name, parent, etc). The subset that have multiple resource_references and have several such fields included in method_signatures is probably smaller. Although, it might be better to not put ourselves in this position in the first place, because it is very easy to add method_signatures & resource_references and then we start seeing combinatorial explosion. 

Breaking alpha/beta APIs is acceptable if well communicated. I think this (making it a breaking change to add/remove a resource reference to a flattened field) will definitely become an issue if we do not address it soon. We might want to run it by Luke b.c I think other systems might depend on the resource annotations as well.

PR #3185, opened 2020-04-24
fix: correct gapic generator template logic

### LINKED ISSUE #3181: Incorrect code generated
Hi,
We have some issues with the code generated in `MockBigQueryWriteImpl.java` file which resulted in synth tool proposing this following PR: https://github.com/googleapis/java-bigquerystorage/pull/191/files

The element is never used before being popped so we would like to get it fixed to what we have on https://github.com/googleapis/java-bigquerystorage/blob/d810b819e386a69df1781d3ed4b47f75f63a320e/google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1alpha2/MockBigQueryWriteImpl.java#L90 

```
requests.add(value);
final Object response = responses.remove();
```

Please correct the generator code or point us to where we can correct it ourselves. Thanks!

stephaniewang526:
Fixes #3181

### DIFF
--- a/src/main/resources/com/google/api/codegen/java/mock_service_impl.snip
+++ b/src/main/resources/com/google/api/codegen/java/mock_service_impl.snip
@@ -78,11 +78,12 @@
   @@Override
   public StreamObserver<{@method.requestTypeName}> {@method.name}(
       final StreamObserver<{@method.responseTypeName}> responseObserver) {
-    final Object response = responses.remove();
     StreamObserver<{@method.requestTypeName}> requestObserver =
         new StreamObserver<{@method.requestTypeName}>() {
       @@Override
       public void onNext({@method.requestTypeName} value) {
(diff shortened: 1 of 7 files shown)


stephaniewang526:
Hi Vadym, thanks for the instructions. I tested this change in v1alpha2 (where we introduced write api changes) and was able to verify the following in the generated tar file `google-cloud-bigquery-storage-v1alpha2-java.tar.gz`:
```   
public StreamObserver<AppendRowsRequest> appendRows(
      final StreamObserver<AppendRowsResponse> responseObserver) {
    StreamObserver<AppendRowsRequest> requestObserver =
        new StreamObserver<AppendRowsRequest>() {
          @Override
          public void onNext(AppendRowsRequest value) {
            requests.add(value);
            final Object response = responses.remove();
...
```

PR #3187, opened 2020-04-30
fix: Fix BUILD file generator (accept dashes in proto options)

vam-google:
Dash is used in showcase protos


vam-google:
@yihanzhen Test

PR #3192, opened 2020-05-11
build_gen: include LRO GAPIC in go_gapic_library dep

noahdietz:
Go GAPICs directly depend on the manual LRO module (containing a Operation construction helper), the GAPIC (for LRO client creation, this dependency), and the go-genproto (for the protobuf/grpc code). This adds the autogen target to the go_gapic_library.

There is a discrepancy with the previous monolith implementation in how the microgenerator implementation presents a go_library that requires the direct depedency on `com_google_cloud_go//longrunning/autogen` in the go_gapic_library target. This discrepancy makes the existing `longrunning_go_gapic` dependency insufficient resulting in `com_google_cloud_go//longrunning/autogen` needing to be added. 

Ultimately, we need to realign the microgenerator implementation with the previous monolith implementation of go_gapic_library.

### DIFF
--- a/rules_gapic/bazel/src/main/java/com/google/api/codegen/bazel/BazelBuildFileView.java
+++ b/rules_gapic/bazel/src/main/java/com/google/api/codegen/bazel/BazelBuildFileView.java
@@ -231,11 +231,13 @@ class BazelBuildFileView {
       if (protoImport.endsWith(":operations_proto")) {
         // Disable injection of unused longrunning GAPIC target as dependency.
         //
-        // TODO(ndietz) enable this dependency once issue is closed: https://github.com/googleapis/gapic-generator-go/issues/387
+        // TODO(ndietz) enable this dependency once issue is closed:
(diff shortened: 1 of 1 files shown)


rules_gapic/bazel/src/main/java/com/google/api/codegen/bazel/BazelBuildFileView.java
```diff
@@ -232,6 +232,7 @@ private String replaceLabelName(String labelPathAndName, String newLabelName) {
         goImports.add(replaceLabelName(protoImport, ":longrunning_go_gapic"));
```
vam-google, line 232: Probably remove or comment out with explanation why gapic does not work and what should be done to make it work.

noahdietz: Good idea, opened an issue for it too. Done.

PR #3200, opened 2020-05-19
all: grpc_service_config timeout settings

noahdietz:
When using grpc_service_config the max & initial rpc timeout should be the same as the total timeout. This essentially renders the hedging logic unused because the initial timeout is the limit. This means one RPC is made with the desired (total) timeout rather than several smaller ones that are backed off until the total is reached.

Previously, the max & initial timeout would be `0` which I believe is incorrect, because that means there is _no_ client-side timeout on the RPC, even if `total_timeout` was provided. So the server would have to close the connection before the client would. This could've caused client calls to hang.

Other changes:
* switch showcase to gapic v2 & grpc_service_config
* retry mapping: fix NPE when a Service in the config is not present in protos
* showcase: add blockTimeout test
* showcase: add valid block showcase test
* update gapic-showcase image for java to v0.10.1
* lengthen delay on flaky showcase java streaming tests

Note: the `ShowcaseTest.java` was tested locally against `gapic-showcase` as well as in CI.


src/main/java/com/google/api/codegen/config/GrpcGapicRetryMapping.java
```diff
@@ -151,11 +155,17 @@ private static void applyRetryPolicyToName(
 
   private static RetryParamsDefinitionProto.Builder retryPolicyToParamsBuilder(
       RetryPolicy retryPolicy, long timeout, String policyName) {
+    // If the timeout is 0, make the timeout multiplier 0 as well, otherwise
```
vam-google, line 158: Is it necessary to set them both to 0? My naive understanding of this suggests that if timeout is 0 it does not matter what the multiplier is, because it will be multiplied by 0, so the total result will always be 0 anyways (well, unless the multiplier is infinity, but it is not practically possible).

src/test/java/com/google/api/codegen/testsrc/showcase/showcase_gapic.yaml
```diff
@@ -20,139 +20,3 @@ language_settings:
     package_name: Google\Showcase\V1beta1
   nodejs:
     package_name: showcase.v1beta1
-# The configuration for the license header to put on generated files.
```
vam-google, line 23: Why is all this removed?

noahdietz: "all this" being the gapic config? Or the license header stuff? 

I converted the config to gapic v2, so I removed extraneous config, and parts of the license header config are unsupported in gapic v2 so I removed the entire thing. Most gapic v2 configs don't have them anways.

src/main/java/com/google/api/codegen/config/GrpcGapicRetryMapping.java
noahdietz: It was more of an aesthetic thing - if the timeout was 0, make the multiplier 0 too. I will revert this and have it just default to 1 regardless.

noahdietz: Done.

PR #3190, opened 2020-05-06
chore: add tests for complex resource IDs

miraleung:
Update tests and baseline protos to support  non-slash resource name separators. That is, "-", "~", "_", ".". Tests will pass once [this PR](https://github.com/googleapis/api-common-java/pull/125) is submitted and a new release is cut.

Relevant issue [here](https://github.com/googleapis/gapic-generator/issues/3164).

PR #3204, opened 2020-05-27
[gg] update api-common to 1.9.2, auto-value to 1.7.2, and guava to 29.0-jre

miraleung:
Dep update so we can use the complex resource name changes in the 1.9.1 release of api-common-java.


miraleung:
Note that the test changes are consistent with the changes in #3190.

PR #3206, opened 2020-05-28
fix: nodejs build_gen uses grpc_service_config

noahdietz:
fix: nodejs build_gen uses grpc_service_config

fix: ignore grpc_service_config unless using gapic_v2

fix: use `None` instead of `""` when grpc_service_config is not present

cc: @alexander-fenster


noahdietz:
> though nodejs has already migrated to microgenerator completely (including bazel presubmits)

build_gen still generates the `gapic_yaml` attribute in `nodejs_gapic_library`, which is unused because the implementation of that rule is provided via gapic-generator-typescript. This fixes that.

PR #3215, opened 2020-06-01
fix: build_gen use lro gapic as lro dep

noahdietz:
Go back to using the local (to googleapis) `//google/longrunning:longrunning_go_gapic` target as the LRO gapic dep for other `go_gapic_library` targets. https://github.com/googleapis/gapic-generator-go/pull/405 fixes the shading issue that required the remote dep `@com_google_cloud_go//longrunning/autogen:go_default_library` be provided.


noahdietz:
> LGTM,
> 
> Does it mean we technically need to regenerate all BUILD.bazel files?

I have a CL ready that updates all of the targets.

PR #3218, opened 2020-06-02
[gg] chore: remove artman client generation from circleci

miraleung:


miraleung:
Closing since it looks like the artman-generated clients are still being used.

PR #3222, opened 2020-06-05
feat: add proto3_optional to proto_custom_library

noahdietz:
This adds the proto3_optional enablement flag to proto_custom_library
which in turn enables it in proto_library_with_info and micro-Bazel
GAPIC targets.


noahdietz:
`bazel build //...` passes in googleapis with this change (and with no uses of proto3_optional in any of the protos)

PR #3239, opened 2020-06-23
feat: add opt_file_args to pass the file arguments to the protoc plugin

viacheslav-rostovtsev:
Currently opt_args parameter allows passing the string arguments to the protoc plugin via the `-<plugin>_opt=` protoc command line argument. This does not allow passing the arguments that point to the files. 
This PR introduces the new parameter to the `proto_custom_library` rule: `opt_file_args`
It works similarly to the `plugin_file_args` to allow passing the file-arguments to the protoc plugins.


alexander-fenster:
@viacheslav-rostovtsev just FYI, I renamed the PR to match the conventional commit standards which is needed for a release bot.  (I'll leave it to @vam-google to review)

Issue #3249, opened 2020-07-10
Broken layout of `java_gapic_assembly_gradle_pkg` output


aohren:
Up until recently, the `java_gapic_assembly_gradle_pkg` rule would product a .tar.gz packaging with the following directory hierarchy:

[Gradle build files + 3 top level directories (for a `java_gapic_assembly_gradle_pkg` with `name = "googleads-java"`)]:
```
.
├── gapic-googleads-java
├── gradle
└── proto-googleads-java
```

Now, the above hierarchy is nested 1-level deep, and two new additional directories (`gapic-googleads-java` and `proto-googleads-java`) exist at root level (one directory up from the build files). The generates sources are also spread out now amongst these 4 different directories (where previously they were only in the 2 `gapic-googleads-java` and `proto-googleads-java` directories above:
```
.
├── gapic-googleads-java
│   └── src
├── googleads-java
│   ├── gapic-googleads-java
│   ├── gradle
│   └── proto-googleads-java
└── proto-googleads-java
    └── src
```

This has broken ability to consume these products downstream, and also makes the Gradle assembly unbuildable, since now there are source files outside of the gradle build directory hierarchy.

vam-google:
This is googelapis-wide issue. Something is completely broken here. May be related to the update to protobuf 3.12.1

vam-google:
Fixed

PR #3253, opened 2020-07-16
fix: Fix for java_gapic_assembly_gradle_pkg rule.

vam-google:
More details at https://github.com/googleapis/gapic-generator/issues/3249.

The Bazel's native pkg_tar rule behaves weirdly when package_dir parameter is specified (at least on some Linux machines it does not put all the files under the package_dir).

As a workaround for that bug we tis fix creates the custom implementation of the pkg_tar rule.


vam-google:
@aohren FYI

PR #3262, opened 2020-08-04
[gapic-gen] Update googleapis hash from 2019 to latest

miraleung:


noahdietz:
I think @vam-google should also verify. This is a cycle dependency technically and I wonder if there is any issue with updating the googleapis hash here.

miraleung:
@vam-google Clicked merge too soon, PTAL and LMK if I should revert this.

alexander-fenster:
@miraleung This repository is handled by a release-please automation and it expects commit messages to match [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/). The title could've been `fix: update googleapis hash ......` or something like that.  The current title won't be understood by the automation.

alexander-fenster:
Also, just FYI, the CI failure should be taken care of by https://github.com/googleapis/gapic-generator/pull/3263.

PR #3265, opened 2020-08-07
fix: Make java lowerCamel methos naming match protobuf java stubs output methods naming

vam-google:
Specifically, before this fix, for a field named `one_two3four` protobuf was generating java method named `setOneTwo3Four()`, while gapic generator was assuming it to be `setOneTwo3four()`.

This change is applied in the most limited scope (only for java methods naming, because this logic is applied only for methods  naming in protobuf). This should be a relatively safe change because protobuf in general does not let fields to differ only by the underscore characters `_`. I.e. one message cannot have two fields one named `one_two` and the other one `onetwo` or `oneTwo`. This restriction greatly reduces number of possible `snake_case` `camelCase` conversion combinations.

### DIFF
--- a/src/main/java/com/google/api/codegen/util/Name.java
+++ b/src/main/java/com/google/api/codegen/util/Name.java
@@ -243,7 +243,15 @@ public class Name {
   }
 
   public String toUpperCamelAndDigits() {
-    char[] upper = toUpperCamel().toCharArray();
+    return capitalizeDigitsAfterNumbers(toUpperCamel());
+  }
+
+  public String toLowerCamelAndDigits() {
+    return capitalizeDigitsAfterNumbers(toLowerCamel());
+  }
+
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/api/codegen/util/Name.java
```diff
@@ -243,7 +243,15 @@ public String toUpperCamel() {
...
   public String toUpperCamelAndDigits() {
-    char[] upper = toUpperCamel().toCharArray();
+    return capitalizeDigitsAfterNumbers(toUpperCamel());
+  }
+
+  public String toLowerCamelAndDigits() {
+    return capitalizeDigitsAfterNumbers(toLowerCamel());
+  }
+
+  private String capitalizeDigitsAfterNumbers(String camelCaseIdentifier) {
```
alexander-fenster, line 253: late to the party, but I bet it meant to be `capitalizeLettersAfterNumbers` :)

PR #3268, opened 2020-08-24
chore: update showcase version for sequence tests

noahdietz:
This updates the version of Showcase used for generation in CI to v0.12.0 and adds a test for the new SequenceService to assert the basic retry attempt logic and timeout-backoff behavior works as implemented.

### DIFF
--- a/showcase/java/src/test/java/com/google/api/showcase/ShowcaseTest.java
+++ b/showcase/java/src/test/java/com/google/api/showcase/ShowcaseTest.java
@@ -28,6 +28,7 @@ import com.google.api.gax.rpc.StreamController;
 import com.google.protobuf.Duration;
 import com.google.rpc.Code;
 import com.google.rpc.Status;
+import com.google.showcase.v1beta1.AttemptSequenceRequest;
 import com.google.showcase.v1beta1.BlockRequest;
 import com.google.showcase.v1beta1.BlockResponse;
 import com.google.showcase.v1beta1.EchoClient;
@@ -35,6 +36,10 @@ import com.google.showcase.v1beta1.EchoRequest;
 import com.google.showcase.v1beta1.EchoResponse;
 import com.google.showcase.v1beta1.EchoSettings;
 import com.google.showcase.v1beta1.ExpandRequest;
+import com.google.showcase.v1beta1.Sequence;
+import com.google.showcase.v1beta1.SequenceReport;
+import com.google.showcase.v1beta1.SequenceServiceClient;
+import com.google.showcase.v1beta1.SequenceServiceSettings;
 import io.grpc.StatusRuntimeException;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


noahdietz:
I'm not really sure why the ruby tests are failing with `TypeError: Sequence is not a module` - my hunch is that the service is named `SequenceService` and there is a message named `Sequence` and somehow this results in a naming conflict.

Not really sure what to do about that. We've migrated everything to gapic-generator-ruby, maybe we shouldn't make the tests required?

showcase/java/src/test/java/com/google/api/showcase/ShowcaseTest.java
```diff
@@ -258,4 +277,66 @@ public void block() {
...
+    assertThat(sequence.getName()).isNotNull();
+
+    seqClient.attemptSequence(
+        AttemptSequenceRequest.newBuilder().setName(sequence.getName()).build());
+
+    SequenceReport report = seqClient.getSequenceReport(sequence.getName() + "/sequenceReport");
+    assertThat(report.getAttemptsList()).isNotNull();
+    assertThat(report.getAttemptsList().size()).isEqualTo(3);
+    for (int i = 0; i < report.getAttemptsList().size(); i++) {
+      if (i == 0) {
```
vam-google, line 331: Why not just `int i = 1` in the for loop declaration?

showcase/java/src/test/java/com/google/api/showcase/ShowcaseTest.java
```diff
@@ -258,4 +277,66 @@ public void block() {
                 .build());
     assertThat(result.getContent()).isEqualTo("Hello, World!");
   }
+
+  @Test
+  public void attemptSequence() {
```
vam-google, line 282: The other test is `TimeoutBackoff`, can you be more specific about what this test tests?

showcase/java/src/test/java/com/google/api/showcase/ShowcaseTest.java
noahdietz: Good call, changed.

showcase/java/src/test/java/com/google/api/showcase/ShowcaseTest.java
noahdietz: I opted to remove this test because it wasn't doing anything more than the `TimeoutBackoff` one. It was the first one I added as I was developing it and I removed some others that were irrelevant to this PR, but neglected to remove this one.

noahdietz:
Just spoke to dazuma and he said that naming collisions are definitely possible, real cases in cloud libs were worked around with synthtool, and were ultimately fixed with the microgenerator. I guess I should do something to exclude this proto from Ruby generation in this CI.

noahdietz:
Fixed it by breaking out ruby generation, and before generating it, I moved the offending proto out of the target directory and remove the service from the 1P yaml. Hacky, but really the only way to exclude this proto from ruby showcase gen.

PR #3274, opened 2020-09-03
Update C# to use the micro-generator

chrisdunelm:


chrisdunelm:
@vam-google Having changed the C# bazel file to use the micro-generator, is there a test file I need to update in this PR too? The gapic-generator-bazel-test is failing, but I can't see enough detail in the logs to know what I might need to update to fix it!
Thanks.

vam-google:
The the build file generator test (`//rules_gapic/bazel:build_file_generator_test`)  test is failing, which is expected, since the build file template was updated so we need to update the baseline (expected output) file to match it: https://github.com/googleapis/gapic-generator/blob/master/rules_gapic/bazel/src/test/data/googleapis/google/example/library/v1/BUILD.bazel.baseline#L346

vam-google (approved):
LGTM, but please make sure that this change and the rest of the changes (like the changes in googleapis) go together (minimize the window when some of them are pushed and others are not)

alexander-fenster:
I just wanted to point out that this PR was merged but never released. Are you expecting this change to be in production already? @chrisdunelm @vam-google

vam-google:
I'm surprised that it is not released. Sure, please release it and integrate with googleapis

PR #3275, opened 2020-09-10
feat: REST GAPIC (REGAPIC) Support for Java

vam-google:
This PR depends on https://github.com/googleapis/gax-java/pull/1177 and provides MVP for generating rest gapic clients. It was tested on real GCE API and was able to produce a valid client.


rules_gapic/gapic.bzl
```diff
@@ -40,11 +40,13 @@ def _gapic_srcjar_impl(ctx):
         _set_args(attr.service_yaml, "--service_yaml=", arguments, inputs)
         _set_args(attr.package_yaml2, "--package_yaml2=", arguments, inputs)
         _set_args(attr.grpc_service_config, "--grpc_service_config=", arguments, inputs)
+        _set_args(attr.transport_protocol, "--transport_protocol=", arguments)
```
miraleung, line 43: Could we please document the valid values of `transport_protocol` somewhere?

rules_gapic/java/java_gapic.bzl
```diff
@@ -175,6 +176,17 @@ def java_gapic_library(
         "@javax_annotation_javax_annotation_api//jar",
     ]
 
+    if transport_protocol == "http":
```
miraleung, line 179: It looks like either HTTP or gRPC will be generated, not both. Would it make sense to make `transport_protocol` a boolean flag that indicates whether it's HTTP or not?

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTransformer.java
```diff
@@ -42,9 +43,11 @@
   private static final String SETTINGS_TEMPLATE_FILENAME = "java/settings.snip";
   private static final String STUB_SETTINGS_TEMPLATE_FILENAME = "java/stub_settings.snip";
   private static final String STUB_INTERFACE_TEMPLATE_FILENAME = "java/stub_interface.snip";
-  private static final String GRPC_STUB_TEMPLATE_FILENAME = "java/grpc_stub.snip";
+  private static final String STUB_TEMPLATE_FILENAME = "java/stub.snip";
```
miraleung, line 46: Since only one of HTTP or gRPC will be generated, would it make sense to split the template into `grpc_stub.snip` and `http_stub.snip`? As a reader, I would prefer duplicate code to parsing the diffs (which is less clear without doing a deep dive).

Ditto for `test.snip`.

miraleung (changes requested):

rules_gapic/gapic.bzl
vam-google: I updated the code to match it the main regapic spec. Those values are documented there (and this argument was changed from transport_protocol to just transport). The values are `grpc`, `rest`, `grpc+rest`. This implementation does not support `grpc+rest` because it is not needed for GCE (the main thing we need now), and investing time end effort into supporting the proper grpc+rest in monolith is probably not worth it.

rules_gapic/java/java_gapic.bzl
vam-google: I changed `transport_protocol` to `transport`. The name and the possible values for the argument now match the general REGAPIC spec.

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTransformer.java
vam-google: GRPC and REST stubs share too much (like 90-80% of the file is the same). Keeping them separate makes it much easier for rest and grpc stubs to drift from each other leading to troubles supporting it. 

I tried to make the .ship files match the philosophical view on the two different transports implementations - they are essentially the same thing with some transport-specific tweaks. Duplicating the .snip files for each transport views them differently - two transports are two separate independent things.

vam-google:
@miraleung PTAL

src/main/java/com/google/api/codegen/transformer/ApiCallableTransformer.java
```diff
@@ -212,39 +220,91 @@ private void setCommonApiCallableFields(
...
+                .getPath()
+                .stream()
+                .map(pathSegment -> normalizePathSegment(pathSegment.toString()))
+                .collect(Collectors.joining("/", "/", "")));
+
+        httpMethodView.pathParamSelectors(
+            populateMethodSelectors(namer, httpAttr.getPathSelectors()));
+        httpMethodView.queryParamSelectors(
+            populateMethodSelectors(namer, httpAttr.getParamSelectors()));
+        httpMethodView.bodySelectors(populateMethodSelectors(namer, httpAttr.getBodySelectors()));
```
vchudnov-g, line 275: nit: Suggest a `\n` before `populateMethodSelectors` to highlight the parallelism with the previous two calls.

src/main/java/com/google/api/codegen/transformer/java/JavaFeatureConfig.java
```diff
@@ -19,22 +19,26 @@
...
 import com.google.api.codegen.transformer.DefaultFeatureConfig;
 import com.google.auto.value.AutoValue;
 
 @AutoValue
 public abstract class JavaFeatureConfig extends DefaultFeatureConfig {
+  @Override
+  public abstract boolean resourceNameTypesEnabled();
 
   @Override
-  public abstract boolean enableStringFormatFunctions();
```
vchudnov-g, line 29: The deleted lines are just cleanup?

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceTransformer.java
```diff
@@ -635,6 +635,15 @@ private StaticLangRpcStubView generateRpcStubClass(
     stubClass.hasDefaultInstance(interfaceConfig.hasDefaultInstance());
     stubClass.hasLongRunningOperations(interfaceConfig.hasLongRunningOperations());
 
+    stubClass.transportProtocol(productConfig.getTransportProtocol());
+    if (productConfig.getTransportProtocol() == TransportProtocol.HTTP) {
```
vchudnov-g, line 639: nit: This is fine, but I'd consider a look-up table here.

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceTransformer.java
```diff
@@ -922,8 +931,17 @@ private void addRpcStubImports(InterfaceContext context) {
         typeTable.saveNicknameFor("com.google.api.client.http.HttpMethods");
         typeTable.saveNicknameFor("com.google.api.core.InternalApi");
         typeTable.saveNicknameFor("com.google.api.pathtemplate.PathTemplate");
-        typeTable.saveNicknameFor("com.google.api.gax.httpjson.ApiMessageHttpRequestFormatter");
-        typeTable.saveNicknameFor("com.google.api.gax.httpjson.ApiMessageHttpResponseParser");
+        String configSchemaVersion = context.getProductConfig().getConfigSchemaVersion();
+        if (configSchemaVersion != null && configSchemaVersion.startsWith("1.")) {
```
vchudnov-g, line 935: Is this also to distinguish between DiscoGAPIC and REGAPIC? If so, add a comment to make that explicit.

src/main/java/com/google/api/codegen/viewmodel/StaticLangRpcStubView.java
```diff
@@ -45,6 +46,19 @@
 
   public abstract String parentName();
 
+  public abstract TransportProtocol transportProtocol();
+
+  public String transportProtocolName() {
+    if (transportProtocol() == TransportProtocol.GRPC) {
+      return "gRPC";
+    }
+    return transportProtocol().toString();
```
vchudnov-g, line 55: This is used in the templates, right? Maybe we need to translate `HTTP` to `REST`, which is what we're exposing externally....

src/main/resources/com/google/api/codegen/java/stub.snip
```diff
@@ -58,6 +62,79 @@
           .build();
 @end
 
+@private httpMethodDescriptor(methodDescriptor)
```
vchudnov-g, line 65: nit: I'd lean towards s/`http`/`rest`/ but if you think this is better for consistency with existing internal code, that's fine.

src/test/java/com/google/api/codegen/gapic/GapicCodeGeneratorTest.java
```diff
@@ -64,19 +67,22 @@ public GapicCodeGeneratorTest(
             null,
             "library",
             null,
+            TransportProtocol.GRPC,
```
vchudnov-g, line 70: Could we also add equivalent tests for `HTTP`?

src/test/java/com/google/api/codegen/gapic/ProtocGapicPluginGeneratorTest.java
```diff
@@ -53,7 +53,7 @@ public void testGenerator() {
                 model.getFiles().stream().map(ProtoFile::getProto).collect(Collectors.toList()))
             // Only the file to generate a client for (don't generate dependencies)
             .addFileToGenerate("multiple_services.proto")
-            .setParameter("language=java")
+            .setParameter("language=java,transport=grpc")
```
vchudnov-g, line 56: can we also test for `transport=rest`?

src/test/java/com/google/api/codegen/testsrc/protoannotations/library.proto
```diff
@@ -86,7 +86,7 @@ service LibraryService {
 
   // Deletes a shelf.
   rpc DeleteShelf(DeleteShelfRequest) returns (google.protobuf.Empty) {
-    option (google.api.http) = { delete: "/v1/{name=bookShelves/*}" };
+    option (google.api.http) = { delete: "/v1/bookShelves/{name}" };
```
vchudnov-g, line 89: So we're supporting both `/{name=bookShelves/*}` and `/bookShelves/{name}`? It looks from https://google.aip.dev/127 that only the former is that standard. What am I overlooking?

vchudnov-g (changes requested):
Looking good!

(You submitted a commit while I was reviewing, so I'll do another pass after you respond)

rules_gapic/java/java_gapic.bzl
```diff
@@ -175,6 +176,17 @@ def java_gapic_library(
         "@javax_annotation_javax_annotation_api//jar",
     ]
 
+    if transport == "rest":
```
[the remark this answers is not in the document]
vam-google: I'm trying to keep it backward-compatible with the older version of the rules, when no transport argument meant grpc. If the implementation of gapic generator does not support rest+grpc I don't think it makes much sense to support that in the rules. The current implementation is both backward compatible and the simplest I could come up with. Doing removal would mean that `rest+grpc` and `grpc+rest` "rest" only which is confusing and more complicated behavior.

rules_gapic/java/java_gapic.bzl
```diff
@@ -183,16 +195,24 @@ def java_gapic_library(
...
         "@com_google_api_gax_java//gax:gax_testlib",
         "@com_google_code_gson_gson//jar",
-        "@io_grpc_grpc_java//auth:auth",
-        "@io_grpc_grpc_netty_shaded//jar",
-        "@io_grpc_grpc_java//stub:stub",
-        "@io_opencensus_opencensus_contrib_grpc_metrics//jar",
         "@junit_junit//jar",
     ]
 
+    if transport == "rest":
```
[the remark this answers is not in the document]
vam-google: Please lets keep it simple, not trying to support in the rules something which is not supported in the generator itself.

rules_gapic/java/java_gapic_pkg.bzl
```diff
@@ -206,9 +207,14 @@ def java_gapic_assembly_gradle_pkg(
         grpc_target_dep = ["%s" % grpc_target]
 
     if client_deps:
+        if transport == "rest":
```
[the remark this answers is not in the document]
vam-google: This `rest`, `grpc`, `rest+grpc`, `grpc+rest` gets propagated in many places (like the above and the bottom comments). By doing it this way I was trying to keep it as simple as possible and not trying to implement in the rules something which is not there in the generator.

rules_gapic/java/java_gapic_pkg.bzl
```diff
@@ -206,9 +207,14 @@ def java_gapic_assembly_gradle_pkg(
...
+            template_label = Label("//rules_gapic/java:resources/gradle/client_disco.gradle.tmpl")
+        else:
```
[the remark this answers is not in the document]
vam-google: Both approaches are possible, but having a third template is kind of a "lazy" approach, easy and straightforward, so most probably we will go with it.

src/main/java/com/google/api/codegen/GeneratorMain.java
```diff
@@ -149,6 +149,14 @@
           .argName("GRPC-SERVICE-CONFIG")
           .required(false)
           .build();
+  private static final Option TRANSPORT =
+      Option.builder()
+          .longOpt("transport") // Possible values: grpc, rest or grpc+rest (not supported yet).
+          .desc("Transport used by the generated clients.")
```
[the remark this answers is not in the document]
vam-google: Done

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -305,7 +306,7 @@ public static GapicProductConfig create(
       return null;
     }
 
-    TransportProtocol transportProtocol = TransportProtocol.GRPC;
+    // TransportProtocol transportProtocol = TransportProtocol.GRPC;
```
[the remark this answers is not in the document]
vam-google: Removed

src/main/java/com/google/api/codegen/config/GapicProductConfig.java
```diff
@@ -159,7 +159,8 @@ public static GapicProductConfig create(
       @Nullable String protoPackage,
       @Nullable String clientPackage,
       TargetLanguage language,
-      @Nullable ServiceConfig grpcServiceConfig) {
+      @Nullable ServiceConfig grpcServiceConfig,
+      TransportProtocol transportProtocol) {
```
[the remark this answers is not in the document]
vam-google: Added

src/main/java/com/google/api/codegen/gapic/GapicGeneratorApp.java
```diff
@@ -112,6 +113,10 @@
           "The filepath of the JSON gRPC Service Config file.",
           "");
 
+  public static final Option<String> TRANSPORT =
+      ToolOptions.createOption(
+          String.class, "transport", "The generated client transport.", "grpc");
```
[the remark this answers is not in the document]
vam-google: Added, but kept grpc as default.

src/main/java/com/google/api/codegen/gapic/GapicGeneratorApp.java
```diff
@@ -208,6 +213,16 @@ protected void process() throws Exception {
     }
 
     String clientPackage = Strings.emptyToNull(options.get(CLIENT_PACKAGE));
+    String transport = options.get(TRANSPORT).toLowerCase();
+
+    TransportProtocol tp;
+    if (transport.equals("grpc")) {
```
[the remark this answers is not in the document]
vam-google: Please, lets keep it as is. It is very unlikely that we will ever support multiple transports in the monolith.

src/main/java/com/google/api/codegen/transformer/ApiCallableTransformer.java
```diff
@@ -212,39 +220,91 @@ private void setCommonApiCallableFields(
...
+          if (fieldConfig.getResourceNameConfig() != null
+              && fieldConfig.getResourceNameConfig().equals(nameConfig)) {
+            httpMethodView.resourceNameFieldName(
+                context
+                    .getNamer()
+                    .privateFieldName(Name.anyCamel(fieldConfig.getField().getNameAsParameter())));
+          }
         }
+        return httpMethodView.build();
+      } else if (context.getMethodModel() instanceof ProtoMethodModel) {
```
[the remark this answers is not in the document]
vam-google: Your understanding is correct. Added the comment.

src/main/java/com/google/api/codegen/transformer/ApiCallableTransformer.java
vam-google: Added

src/main/java/com/google/api/codegen/transformer/java/JavaFeatureConfig.java
vam-google: This is GitHub diff tool doing poor job =(. Nothing was removed, just shifted a bit. This method is still there on line 29 in the "new" version of the file, shown as addition.

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceTransformer.java
vam-google: This is how these types of differences were reflected in the other parts of gapic-generator (i.e. if statements like this are pretty common). Not very pretty, but given there are only two options, lookup table seems not necessary.

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceTransformer.java
vam-google: This is correct. A hacky way of doing it, but it is quite reliable way to distinguish (DIREGAPIC is always 2.0 gapic yaml or no yaml, while discogapic is always gapic yaml 1.0).

src/main/java/com/google/api/codegen/viewmodel/StaticLangRpcStubView.java
vam-google: Agree. Changed to REST

src/main/resources/com/google/api/codegen/java/stub.snip
vam-google: Yes, this is for consistency and compatibility with discogapic. I can't make big changes here untill discogapic is removed.

src/test/java/com/google/api/codegen/gapic/GapicCodeGeneratorTest.java
vam-google: This test is for gapic yaml v1 generation, which is not applicable for DIREGAPIC.

src/test/java/com/google/api/codegen/gapic/ProtocGapicPluginGeneratorTest.java
vam-google: We technically do, on line 76. But seriously speaking, these tests are useless, because gapic-generator is never executed as plugin in production, so they are testing dead logic.

src/test/java/com/google/api/codegen/testsrc/protoannotations/library.proto
vam-google: Both works, even if they are not listed in AIP127, so testing both.

vam-google:
@vchudnov-g @miraleung PTAL

rules_gapic/java/java_gapic.bzl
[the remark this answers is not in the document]
miraleung: +1 to backwards-compatibility.

src/main/java/com/google/api/codegen/gapic/GapicGeneratorApp.java
```diff
@@ -112,6 +113,14 @@
           "The filepath of the JSON gRPC Service Config file.",
           "");
 
+  public static final Option<String> TRANSPORT =
+      ToolOptions.createOption(
+          String.class,
+          "transport",
+          "List of transports to use ('rest' or 'grpc') separated by '+'. NOTE: For now"
```
miraleung, line 120: Should this just be "Transport to use" since it's not a List option?

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTransformer.java
miraleung: SGTM.

miraleung (approved):

src/main/java/com/google/api/codegen/transformer/ApiCallableTransformer.java
vchudnov-g: Don't see it. I think maybe you added the `\n` in the previous line rather than just inside the outer parenthesis?

src/main/java/com/google/api/codegen/transformer/java/JavaFeatureConfig.java
vchudnov-g: oops. I missed that!

src/main/java/com/google/api/codegen/transformer/java/JavaGapicSurfaceTransformer.java
vchudnov-g: +1

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceTransformer.java
vchudnov-g: sgtm

src/main/java/com/google/api/codegen/transformer/java/JavaSurfaceTransformer.java
vchudnov-g: Yeah. The comment is very helpful! (As in the other file where you also added it.) Thanks!

src/main/resources/com/google/api/codegen/java/stub.snip
vchudnov-g: ok

src/test/java/com/google/api/codegen/gapic/ProtocGapicPluginGeneratorTest.java
vchudnov-g: But we are providing the functionality in the generator, even if we ourselves don't use it, so it should be tested. Line 76 tests the failing case.

All that said, it's only worth doing if it exercises additional code. I think this will test that the high-level transport selection doesn't error out, right? So it might be worth including. Can you simply copy this existing test case and change the transport (and maybe check for the presence of a REST-specific string, if that's simple).

src/test/java/com/google/api/codegen/testsrc/protoannotations/library.proto
vchudnov-g: But if it's not in the AIP, we want to NOT support it. We don't want people annotating their gRPC APIs incorrectly, so we need to fail when they do. If the unapproved pattern is something that the converter is outputting into the synthetic protos, let's fix that. At the very least, could you file a bug in the converter repo and one in this repo, and add a TODO here to restrict to the supported pattern?

src/main/java/com/google/api/codegen/GeneratorMain.java
```diff
@@ -149,6 +149,17 @@
           .argName("GRPC-SERVICE-CONFIG")
           .required(false)
           .build();
+  private static final Option TRANSPORT =
+      Option.builder()
+          .longOpt("transport")
+          .desc(
+              "List of transports to support. Valid transport names ('grpc' or 'rest') are"
+                  + " separated by '+'. Default is 'grpc'. NOTE: for now, GAPICs support only"
```
vchudnov-g, line 157: This comment as phrased makes sense with the code changes I suggested in the last pass, where (a) the default list is `grpc+rest` and (b) because we're not yet implementing dual transport, we then look at only the first element in the list as per my #[1] comment: `all_transports = all_transports[0:1]`

vchudnov-g (commented):
I have a concern about AIP127; help me understand that. And then the help text in `GeneratorMain` I doesn't quite match the current implementation.

src/test/java/com/google/api/codegen/testsrc/protoannotations/library.proto
vam-google: This would require non-trivial unnecessary conversions in the converter which I really would like to avoid (one more thing to break potentially). This works out of the box in Java and PHP  and I'm pretty sure it will work in other langauges as well. If anything, I believe it is AIP description itself is incomplete, not the pattern used here is wrong. Note, that AIP does not really attempt to provide a full spec of how the template strings may look like.

Lets discuss it in person, I feel strong about not making the unnecessary conversion here (we are converting something simple to something more complicted, in a situation when it really looks unnecessary, to follow AIP which is not attempting to rigorously define how these template strings should look like).

rules_gapic/java/java_gapic.bzl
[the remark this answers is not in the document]
vam-google: Note, all this is essentially a throw away code, because the final destination of it will be in java microgenerator. Please lets not try to polish throwaway code or add features which most likely will never be fully implemented here (like the http+grpc option).

src/test/java/com/google/api/codegen/testsrc/protoannotations/library.proto
vam-google: Btw, Looks like the annotation itself officially supports the plain syntax: https://github.com/googleapis/googleapis/blob/master/google/api/http.proto#L96

rules_gapic/java/java_gapic.bzl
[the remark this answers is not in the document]
vchudnov-g: It's a very simple change that increases clarity, IMO. My preference would be to add it. But as I said, I'm not going to block on this.

src/main/java/com/google/api/codegen/GeneratorMain.java
vchudnov-g: Please update the comment if you're not handling the list option: you're not handling the "+" separated list unless you make changes like the ones I suggested,.

src/main/java/com/google/api/codegen/gapic/GapicGeneratorApp.java
vchudnov-g: Again, change the NOTE to be accurate: for now, we only support a single transport.

src/main/java/com/google/api/codegen/transformer/ApiCallableTransformer.java
vchudnov-g: ping

src/test/java/com/google/api/codegen/gapic/ProtocGapicPluginGeneratorTest.java
vchudnov-g: ping

src/test/java/com/google/api/codegen/testsrc/protoannotations/library.proto
vchudnov-g: Short version: OK, but add a note.

Long version:

In our discussion today we said that the HTTP annotation is independent of resource name formatting. It just says how to substitute the fields into an HTTP pattern, not how that pattern should look like,

I think that's wrong, though: from the AIP:
```
URIs must use the {foo=bar/*} syntax to represent a variable that should be populated in the request proto. When extracting a resource name, the variable must include the entire resource name, not just the ID component.
```

That said: it's OK for the monolith to accept this input for now. I don't think microgenerators should. I would put a TODO here noting that this is non-standard, even though we won't fix it in the monolith

(And the converter emits protos in the correct format, right?)

vchudnov-g (approved):
Please address the doc comments and the comments I pinged you on.

(BTW: I think if you reply to my review with a review of your own, then it will show up in my dashboard. I think that's why I didn't see this earlier.)

src/test/java/com/google/api/codegen/testsrc/protoannotations/library.proto
vchudnov-g: To be clear, this isn't blocking this PR.

On the generation side, I checked and you're right, this pattern is intended to be permissible. The relevant AIP here is [AIP 4231](https://google.aip.dev/client-libraries/4231#complex-resource-id-path-segments) (Could you add a reference to that in the converter design doc, please?). I'll work with Luke to make the wording on AIP 127 less misleading (to me).

PR #3294, opened 2020-10-13
fix: use separate value generator in test generation

yihanzhen:
Sometimes a library has two requests methods with a field of  the same name but different types. The `TestValueGenerator` keys on the short name of fields, and reuses a generated value if it has generated a value for this key (short field name). Therefore in such cases it may try to assign a string literal to a integer type.

Change to create a different `TestValueGenerator` for every single test case generation to avoid such cases.


vam-google (approved):
LGTM, thanks Hanzhen for fixing the stuff you are not longer responsible for =)

PR #3299, opened 2020-10-15
fix: use remote lro Go gapic instead of local

noahdietz:
Using the `googleapis` local LRO GAPIC target instead of the remote version in `cloud.google.com/go` is causing some Go linker issues. We've found that if we use the remote dep instead of the local target, the Go linker doesn't complain about it. I believe this is because the remote version depends on some of the remote dependencies (in go-genproto), rather than the local `googleapis` versions of them e.g. `@go_googleapis//google/api:annotations_go_proto` (remote) vs. `//google/api:annotations_go_proto` (local).


vam-google (approved):
LGTM

PR #3301, opened 2020-10-23
DIREGAPIC for PHP support

bshaffer:
Work in progress - first pass at DIREGAPIC generation for PHP


src/main/java/com/google/api/codegen/transformer/TestCaseTransformer.java
```diff
@@ -233,8 +236,9 @@ public TestCaseView createTestCaseView(
         .build();
   }
 
-  private String getMethodDescriptorName(MethodContext context) {
-    if (context.getProductConfig().getTransportProtocol().equals(TransportProtocol.HTTP)) {
+  private String getMethodDescriptorName(MethodContext context, TargetLanguage targetLanguage) {
+    if (context.getProductConfig().getTransportProtocol().equals(TransportProtocol.HTTP)
+        && targetLanguage != TargetLanguage.PHP) {
```
bshaffer, line 241: This is an unfortunate hack to get around the issue with Java doing something different than PHP. If there's a better way to do this, I'm all ears!

src/main/java/com/google/api/codegen/viewmodel/DynamicLangXApiView.java
```diff
@@ -128,6 +128,8 @@ public boolean hasMissingDefaultOptions() {
   @Nullable
   public abstract String grpcTransportImportName();
 
+  public abstract boolean isRestOnlyTransport();
```
vchudnov-g, line 131: As per go/actools-regapic-design, in the final product, generated GAPICs must be able to support multiple transports if supported by the API. For Java, we'll support this in the microgenerator; the monolith generates single-transport GAPICs. For PHP, the situation is likely similar, though on a longer timescale. All this to say that I would appreciate structuring the code in terms of "supports REST" rather than "REST only", but only where this doesn't require undue work. For everything else, please add a TODO to the effect of  "we're generating single-transport GAPICs for now, though in the fullness of time we'll generate multi-transport GAPICs."

src/main/resources/com/google/api/codegen/php/client_impl.snip
```diff
@@ -348,24 +361,42 @@
...
-     *           $transportConfig = [
-     *               'grpc' => [...],
-     *               'rest' => [...]
-     *           ];
-     *           See the {@@see \Google\ApiCore\Transport\GrpcTransport::build()} and
-     *           {@@see \Google\ApiCore\Transport\RestTransport::build()} methods for the
-     *           supported options.
+     @if xapiClass.isRestOnlyTransport
+         *     @@type string|TransportInterface $transport
+         *           The transport used for executing network requests. Defaults to `rest`.
```
vchudnov-g, line 366: given `RestOnlyTransport`, should this `Defaults to 'rest'` be `At the moment, only supports 'rest'`?

src/main/java/com/google/api/codegen/viewmodel/DynamicLangXApiView.java
bshaffer: It doesn't make sense to say `supports REST` because PHP GAPICs already support REST, and we do not need any special handling (REGAPIC). 

I could change this to `isDireGapic()`, `isDiscovery()`, or something along these lines, but it's very important to distinguish from existing REST support in OnePlatform GAPICs and the support for DIscovery REST-only APIs.

src/main/resources/com/google/api/codegen/php/client_impl.snip
bshaffer: I like that better!

src/main/java/com/google/api/codegen/transformer/TestCaseTransformer.java
vam-google: For this particular method, probably just make it Java-only (instead everything but PHP). I checked why this method even exists, and it turns out it should not even be there (it is  java-specific logic which should have been refactored long time ago).

src/main/java/com/google/api/codegen/viewmodel/DynamicLangXApiView.java
vam-google: Looks like the gist of the problem is we are either missing `GRPC_HTTP` value in the `TransportProtocol` enum or we should convert all usages of `TransportProtocol` to be a list value, which may contain multiple entries (to represent grpc+rest).

In that case you would not need `isRestOnlyTransport` the `TransportProtocol` value will contain the information you need. This is basically technical debt introduced for java started biting PHP implementation.

src/test/java/com/google/api/codegen/testsrc/protoannotations/library_rest_v2_gapic.yaml
```diff
@@ -0,0 +1,602 @@
+type: com.google.api.codegen.ConfigProto
+config_schema_version: 2.0.0
```
vam-google, line 2: This is a v2 gapic yaml, it should be a tiny one. All of the method-level configs should be delivered as proto annotation. I see a lot of samples config here, but it is basically deprecated at this point and should not be used for new code.

vam-google (commented):
Mostly LGTM with few comments

src/main/java/com/google/api/codegen/viewmodel/DynamicLangXApiView.java
vchudnov-g: Yes, the eventual goal is to allow list values, so that `grpc+rest` works. I would love it if support for the flags could be made to work now, and I would love it even more if the implementation supported multi-transport.

One question, then, about PHP: since PHP already supports REST, REGAPIC is not replacing the existing support, but just adding to it, right? Am I missing something?

bshaffer: I still think, rather than focus on the `TransportProtocol` as the means to derive this logic, we should remove `TransportProtocol` as an enum/option/parameter and instead use `ClientType`. This would allow us to have `ClientType.GAPIC`, `ClientType.REGAPIC`, and `ClientType.DIREGAPIC` (I would prefer `ClientType.DISCOVERY` and `ClientType.ONE_PLATFORM`, but if you feel differing between GAPIC and REGAPIC is important we can do the former).

src/test/java/com/google/api/codegen/testsrc/protoannotations/library_rest_v2_gapic.yaml
bshaffer: I am not familiar with the difference, but I'm happy to make the change. Can you point to an example of what I should be using instead?

src/main/java/com/google/api/codegen/viewmodel/DynamicLangXApiView.java
bshaffer: We could also do, instead, something along the lines of `hasGrpcSupport`, which serves the same purpose as `isRestOnly`, but makes it a little clearer. WDYT?

rules_gapic/php/php_gapic_repositories.bzl
```diff
@@ -72,7 +72,7 @@ def php_gapic_repositories():
     _maybe(
         http_file,
         name = "php_cs_fixer",
-        urls = ["https://github.com/FriendsOfPHP/PHP-CS-Fixer/releases/download/v2.9.1/php-cs-fixer.phar"],
+        urls = ["https://github.com/FriendsOfPHP/PHP-CS-Fixer/releases/download/v2.16.2/php-cs-fixer.phar"],
```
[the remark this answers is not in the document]
bshaffer: I can also downgrade this... I was experimenting with the latest version because the previous version wasn't running on PHP 7.4. IIRC, the latest version ALSO doesn't run on PHP 7.4, so this change is not necessary.

src/main/java/com/google/api/codegen/viewmodel/DynamicLangXApiView.java
bshaffer: Changed to `supportsGrpcTransport`

bshaffer:
Closing in favor of https://github.com/googleapis/gapic-generator/pull/3305

PR #3305, opened 2020-11-10
feat: add DIREGAPIC support for PHP

bshaffer:
Resubmission of https://github.com/googleapis/gapic-generator/pull/3301 from origin so all tests pass

 - Adds `supportsGrpcTransport`, which returns false when `transport_protocol` is set to `REST`
 - Modifies PHP GAPIC to:
   - ensure documentation for `transport` and `transportConfig` service clients options omit `grpc`
   - omit `gcpApiConfigPath` config in `getClientDefaults`
   - omit importing any grpc-related classes
   - Adds `defaultTransport` and `getSupportedTransport` functions to override the transport verification functionality in `google/gax`
 - Adds new baseline test for DIREGAPIC PHP
 - Updates PHP bazel rules to allow for REST transport


src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
```diff
@@ -426,6 +435,9 @@ private RestPlaceholderConfigView generateRestPlaceholderConfigView(
     List<GrpcStreamingDetailView> result = new ArrayList<>();
 
     for (MethodModel method : context.getGrpcStreamingMethods()) {
+      if (!supportsGrpcTransport()) {
+        throw new RuntimeException("Streaming methods invalid for REST-only transport");
```
vchudnov-g, line 439: tiny nit: What about: "Streaming methods only valid for gRPC transport"

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
```diff
@@ -453,7 +465,9 @@ private void addApiImports(GapicInterfaceContext context) {
     typeTable.saveNicknameFor("\\Google\\ApiCore\\CredentialsWrapper");
     typeTable.saveNicknameFor("\\Google\\ApiCore\\GapicClientTrait");
     typeTable.saveNicknameFor("\\Google\\ApiCore\\PathTemplate");
-    typeTable.saveNicknameFor("\\Google\\ApiCore\\RequestParamsHeaderDescriptor");
+    if (supportsGrpcTransport()) {
```
vchudnov-g, line 468: tiny nit: Does order matter here? If not, I would move this to be the least one, for readability.

vchudnov-g (commented):
Looking good! Only nits so far, but please wait for @vam-google to TAL.

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
```diff
@@ -191,14 +195,19 @@ private ViewModel buildGapicClientViewModel(GapicInterfaceContext context) {
     apiImplClass.clientConfigPath(namer.getClientConfigPath(context.getInterfaceConfig()));
     apiImplClass.clientConfigName(namer.getClientConfigName(context.getInterfaceConfig()));
     apiImplClass.interfaceKey(context.getInterface().getFullName());
-    String grpcClientTypeName =
-        namer.getAndSaveNicknameForGrpcClientTypeName(
-            context.getImportTypeTable(), context.getInterfaceModel());
-    apiImplClass.grpcClientTypeName(grpcClientTypeName);
+    if (supportsGrpcTransport()) {
```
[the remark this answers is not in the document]
bshaffer: For PHP, this block generates a GAPIC that supports both.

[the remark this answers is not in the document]
bshaffer: I've added comments to clarify, let me know if this is sufficient!

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
bshaffer: I agree! Order doesn't matter! Moved for readability.

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
```diff
@@ -191,15 +195,22 @@ private ViewModel buildGapicClientViewModel(GapicInterfaceContext context) {
...
+      String grpcClientTypeName =
+          namer.getAndSaveNicknameForGrpcClientTypeName(
+              context.getImportTypeTable(), context.getInterfaceModel());
+      apiImplClass.grpcClientTypeName(grpcClientTypeName);
+
+      apiImplClass.stubs(grpcStubTransformer.generateGrpcStubs(context));
+    } else {
+      // PHP generates a client that only supports REST
+      apiImplClass.grpcClientTypeName("");
+      apiImplClass.stubs(new ArrayList<>());
```
vam-google, line 209: nit: probably `Collections.emptyList()` is what you need here.

src/main/java/com/google/api/codegen/transformer/php/PhpGapicSurfaceTransformer.java
```diff
@@ -426,6 +437,9 @@ private RestPlaceholderConfigView generateRestPlaceholderConfigView(
     List<GrpcStreamingDetailView> result = new ArrayList<>();
 
     for (MethodModel method : context.getGrpcStreamingMethods()) {
+      if (!supportsGrpcTransport()) {
+        throw new RuntimeException("Streaming methods only valid for gRPC transport");
```
vam-google, line 441: How will this behave for existing APIs? There are many APIs that PHP generates the multitransport client and those have streaming methods. I believe in that case PHP still creates a client, just the streaming methods are no-op, but the generation itself does not fail and all the non-streaming methods are present in the generated surface. Will make gapic-generator throwing exception for such APIs and failing the whole generation process as a result?

Also, please do not use `RuntimeException` directly. It is an exception designed to be a base for the other specific exceptions. Please pick something more specific (UnsupportedOperation or something like that)

vam-google (commented):
Overall looks good. Just a few comments. 
I'm also worried a bit about the addition of one more huge test file `library_rest.proto`. Why can't we use one of the existing `library.proto`. Is it because of streaming methods?

vam-google (approved):
LGTM

PR #3311, opened 2020-11-19
 build(python): default to microgenerator rules

busunkim96:
Keep `py_gapic_library2` alias for now. This will be switched to `py_gapic_library` once `repository_rules.bzl` in googleapis is cleaned up. There are still a handful (<5) libraries using the monolith in Python but any new libraries should use the microgenerator.


software-dov (approved):

busunkim96:
@software-dov @vam-google  Could I get help with the coverage drop? (Or is it okay to ignore?)

busunkim96:
Friendly ping @vam-google

PR #3312, opened 2020-12-06
chore: Update `BuildFileGenerator` to match GAPIC generators suface used in gogoleapis

vam-google:
The generated `BUIDL.bazel` files now use java microgenerator rules surface.
Python rules were "rolled back" to use regular names instead of aliases, because googleapis will be using (once the pending CL is merged) python mocrogenerator by default.

Also changed the order of parameters in C# rule to match default buildozer formatting.


busunkim96 (approved):

miraleung (approved):
LGTM for Java

PR #3313, opened 2020-12-07
feat: generate GAPIC metadata file for Node.js by default

alexander-fenster:
Since we have [enabled](https://github.com/googleapis/googleapis/commit/d189e871205fea665a9648f7c4676f027495ccaf) GAPIC metadata generation for all Node.js builds, let's have this setting enabled by default.

Also setting the default `package_name` since it is always defined for all the libraries (it's normally renamed manually to something like `@google-cloud/library` but having some reasonable default is good).


vam-google (approved):
LGTM

PR #3317, opened 2020-12-09
fix: don't change PHP functions which are reserved words 

bshaffer:
with the exception of `__halt_compiler`, which is so unlikely to be generated as a function name that we don't need to write a case for it, none of the PHP keywords are invalid as class function names. We can safely remove wrapping them.

### DIFF
--- a/showcase/php/tests/ShowcaseIntegrationTest.php
+++ b/showcase/php/tests/ShowcaseIntegrationTest.php
@@ -48,7 +48,7 @@ class ShowcaseIntegrationTest extends TestCase
      */
     public function testUnary($client)
     {
-        $response = $client->echo_([
+        $response = $client->echo([
             'content' => '"Wales snail hail fails!" wails Gail'
         ]);
         $this->assertEquals(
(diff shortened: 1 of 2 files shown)


miraleung (approved):

vam-google (approved):
Can it result in a breaking change for the existing PHP gapic clients which already have the function names changed even though it was not necessary.

vam-google:
@bshaffer LGTM with a question

bshaffer:
@vam-google I searched for the function prefix `_(` in `google-cloud-php` and there were no results. I believe this is sufficient to ascertain there are no conflicts!

vchudnov-g (commented):
lgtm, but I'm no PHP expert

bshaffer:
@jdpedrie caught that this is only the case for PHP 7.0+

Versions of PHP 5.6 and below will throw fatal errors for functions named with reserved words

bshaffer:
Note: Removing the escaping of functions for PHP Reserved Words for all GAPICs exposes the very low-likelyhood that we might generating a GAPIC with a PHP Fatal Error on PHP < 7 between now and when we deprecate support for PHP < 7. 

The plan is to deprecate support in early 2021. If such a fatal error occurs, we will be able to wrap the functions in reserved words for all but DIREGAPIC client libraries, which will all require PHP > 7 from the start.

PR #3319, opened 2020-12-18
fix: preserve the value of opt_args parameter

alexander-fenster:
Needed for `py_gapic_library` rule.


vam-google (approved):
LGTM

busunkim96:
Thanks @alexander-fenster!

PR #3322, opened 2020-12-29
Revert "fix: don't change PHP functions which are reserved words "

bshaffer:
Reverts googleapis/gapic-generator#3317

This PR requires a minimum version of PHP 7.0 in `google/cloud-compute`. We will require this for the `beta`, but for the `alpha` we can use the function names `list_()`.


vam-google (approved):
LGTM

PR #3325, opened 2021-01-07
update php paged map field snippets

bshaffer:
Update the sample code for mapped fields to include the key for clarity.


vam-google (approved):
LGTM

PR #3333, opened 2021-01-21
fix: PHP Use a deterministic ordering for rest placeholders

chrisdunelm:
Deterministic outputs are required for automated testing of the micro-generator when comparing outputs between generators


bshaffer (approved):
This is going to be a doozy when these clients get regenerated, but I'm ok with that.

/cc @dwsupplee @jdpedrie

chrisdunelm:
Thanks @bshaffer :)

chrisdunelm:
@noahdietz @miraleung if one of you could possibly review this (tiny!) PR, that'd be great :) I think it needs an actools reviewer before I can merge...
Thank you!

miraleung (approved):

